PR #1391, opened 2026-05-30
fix(i18n): metadata labels follow language switch without refresh (#1319)

xuyushun441-sys:
## What

Resolves the framework half of #1319: switching the UI language (zh-CN ↔ English) left the app **half-translated** — static strings flipped instantly but server-resolved metadata labels (object/field/view labels, action-dialog text) stayed in the old language until a hard refresh.

Root cause: metadata labels are resolved **server-side** and cached per fetch, with no locale awareness in the request or the protocol ETag, so a language switch never refetched them.

## Changes

- **`@objectstack/client`** — add `setLocale`/`getLocale`; send `Accept-Language` on every request so the server resolves labels in the active locale.
- **`@objectstack/objectql` + `@objectstack/spec`** — make the protocol ETag **locale-aware** so cached metadata no longer collides across languages.
- **`@objectstack/client-react`** — metadata hooks refetch when the active locale changes (`ObjectStackLocaleContext`, render-phase `setLocale` sync).
- **`@objectstack/rest`** — honor the request locale when resolving metadata labels.
- **`apps/account`** — thread the router locale through `__root.tsx` so a language switch relabels in place.

## Why this matters

This is the platform half of a two-repo fix. The companion renderer change ships in **objectui** (`createAuthenticatedFetch` Accept-Language injection + a render-phase cache clear and `key={language}` metadata-subtree remount), because objectui consumes `@objectstack/client` from the registry and can't depend on the unpublished `setLocale`. Together they satisfy the issue's acceptance criterion: **switch language without a page refresh**.

## Testing

- `packages/client` — `client.test.ts` covers `setLocale`/`getLocale` + `Accept-Language` header.
- `packages/objectql` — `protocol-meta.test.ts` covers the locale-aware ETag.
- All touched packages type-check clean.

## Notes for reviewers

- This branch was at `origin/main`; the PR contains a single commit. Unrelated in-progress seed-loader/system-names work in the working tree was deliberately **left out**.
- A changeset is included (`client`/`client-react` minor; `objectql`/`rest`/`spec` patch).
- Follow-up milestones from the issue are also done on the objectui side: **M2** dev-mode missing-key warnings, **M3** date/number/RTL (verified already complete).

### DIFF
--- a/.changeset/locale-aware-metadata.md
+++ b/.changeset/locale-aware-metadata.md
@@ -1,0 +1,16 @@
+---
+"@objectstack/client": minor
+"@objectstack/client-react": minor
+"@objectstack/objectql": patch
+"@objectstack/rest": patch
+"@objectstack/spec": patch
+---
+
+Make metadata labels follow the active UI language without a page refresh (#1319).
+
+The client now carries the active locale on every request (`Accept-Language`,
+`setLocale`/`getLocale`), the protocol ETag is locale-aware so cached metadata
+no longer collides across languages, and the `client-react` metadata hooks
+refetch when the locale changes. The `apps/account` console wires its router
+locale through so a language switch relabels server-resolved object/field/view
+labels in place instead of leaving the UI half-translated until reload.
--- a/apps/account/src/routes/__root.tsx
+++ b/apps/account/src/routes/__root.tsx
@@ -2,6 +2,7 @@
 
 import { createRootRoute, Outlet, useLocation, useNavigate } from '@tanstack/react-router';
 import { useEffect, useMemo, useState } from 'react';
+import { useObjectTranslation } from '@object-ui/i18n';
 import { ObjectStackProvider } from '@objectstack/client-react';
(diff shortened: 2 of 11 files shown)


xuyushun441-sys:
### Re: failing **Vercel** check (`spec` / `apps/docs` deployment)

This failure is **pre-existing and not caused by this PR** — the same Vercel `spec` docs deployment is currently failing on `main` (`Deployment failed.`) and on other open PRs (e.g. #1390). It's an environment/deploy-level breakage independent of these changes.

I reproduced the docs build's spec-generation steps locally against this branch:

```
pnpm --filter @objectstack/spec gen:schema   # ✅ 1684 schemas generated
pnpm --filter @objectstack/spec gen:docs     # ✅ Done!
```

Both pass. The only spec change here is one optional `locale` field on `GetMetaItemCachedRequestSchema`, which flows through schema/doc generation cleanly. No code change is warranted for this check; it should go green once the docs deployment is fixed on `main`.

PR #1392, opened 2026-05-30
fix(seed): bind seed records to users via os.user + fail loudly on unresolved refs

### LINKED ISSUE #1389: Seed data: no way to bind records to users (required lookup('user') makes objects un-seedable, fails silently)
## Summary

There is no documented or ergonomic way to reference **users** from declarative seed data (`defineDataset` / `defineStack({ data })`). Any `lookup('user')` field that is `required: true` makes its object **impossible to seed**, because the platform users that a seed record would point to do not exist until *after* first signup / auth provisioning.

This silently breaks templates: every record whose object has a required `user` lookup fails validation at boot and is dropped, leaving the table empty with **no error surfaced** to the developer.

## Repro

1. Define an object with a required user lookup:

   ```ts
   export const Resource = ObjectSchema.create({
     name: 'pm_resource',
     fields: {
       project: Field.lookup('pm_project', { required: true }),
       person:  Field.lookup('user',       { required: true }), // <-- problem
       allocated_hours_per_week: Field.number({ required: true }),
     },
   });
   ```

2. Provide seed data for it:

   ```ts
   export const resources = defineDataset(Resource, {
     mode: 'upsert',
     externalId: 'role',
     records: [
       { role: 'iOS Dev', project: { name: 'Mobile App Redesign' },
         person: '???', allocated_hours_per_week: 40 },
     ],
   });
   ```

3. `os compile` succeeds (the artifact contains the records).
4. `os dev` against a **fresh** environment seeds **0 rows** for that object — no log line, no validation error, no hint.

## Expected

A first-class, documented mechanism to bind seed records to users, plus a loud failure when a seed record can't resolve a required reference.

## Questions / discussion

1. **How should seed data reference users?** Options I can think of:
   - A reserved seed-only sentinel for the bootstrap/admin user, e.g. `person: { ref: '@admin' }` or `cel\`currentUser()\``, resolved at seed time to the first/admin user.
   - Allow seeding `user` records themselves in `data`, then referencing them by `externalId` (needs confirmation that the auth-provided `user` object is writable from a project stack's seed).
   - A documented "seed runs as / on behalf of" identity so `created_by` / required user lookups can default.

2. **Ordering / availability** — is the admin user guaranteed to exist at the point `data` seeding runs? If not, can seeding be deferred until after first-run auth provisioning?

3. **Fail loudly** — when a seed record fails validation (e.g. unresolved required lookup), the runtime should emit a warning/error per dropped record rather than silently inserting nothing. The silent-drop behavior is what made this hard to diagnose.

## Impact

Discovered while building the `project` template in `objectstack-ai/templates`: all five seed datasets produced **0 rows** in a fresh environment. The root cause for `pm_resource` specifically was a required `lookup('user')` (`person`) that seed data could not satisfy. As a workaround we made `person` optional and dropped the user reference from seed records, but a real binding mechanism is needed so templates can ship realistic, user-attributed sample data.

## Environment

- `@objectstack/*` `7.3.0`
- Driver: `SqlDriver(better-sqlite3)`, single-tenant local dev (`os dev`)

xuyushun441-sys:
## Summary

Fixes #1389 — _"Seed data: no way to bind records to users (required `lookup('user')` makes objects un-seedable, fails silently)"_.

Records seeded via `defineDataset` / `defineStack({ data })` can now bind to a platform user with `` cel`os.user.id` `` (and to the org with `` cel`os.org.id` ``). Previously this never resolved at boot, so any object with a **required** owner lookup was un-seedable and the records were **silently dropped**.

## What changed

**1. `os.user` / `os.org` now resolve (first-class binding)**
- New `SeedLoaderConfig.identity` (`@objectstack/spec`) carries the `os.user` / `os.org` subject into CEL evaluation.
- The seed loader threads it into the eval context; `os.org` falls back to `organizationId` when not explicitly set.
- Convention is the existing `os.user` / `os.org` — no new syntax (`currentUser()`, `@admin`, …).

**2. Admin availability / ordering guarantee**
- `AppPlugin` provisions a deterministic, **non-loginable** system user (`usr_system`, role `system`) **before** any seed runs and binds it as `os.user`, so identity-derived seeds resolve even on a fresh boot — before the first human sign-up.
- The human login admin stays a separate better-auth identity (ADR-0010 lock on `sys_user` respected — no credential is minted here). Analogous to Salesforce's "Automated Process" user.
- Exposed as the canonical `SystemUserId.SYSTEM` constant.

**3. Loud failure instead of silent drop**
- A record whose CEL value can't resolve (e.g. a required `` cel`os.user.id` `` with no identity) — or that fails to write — is now counted as an error, marks the load `success === false`, and logs an actionable message.
- Applied to both the inline seed run and the per-tenant replayer; boot logs a success/warn summary.

**4. Docs**
- `content/docs/guides/seed-data.mdx` gains a "Dynamic Values (CEL)" section covering the `os.user` binding, the `usr_system` ordering guarantee, and the loud-failure behavior.

## Reviewer notes
- A changeset (`minor` for `@objectstack/spec` + `@objectstack/runtime`) is included.
- `@objectstack/spec` must be rebuilt so its `dist` `.d.ts` exposes the new `identity` field + `SystemUserId` (vitest resolves spec from src, but `tsc` resolves via dist).

## Testing
- 4 new seed-identity tests (resolve `os.user.id`; loud failure when unbound; `os.org` → `organizationId` fallback; write-failure surfaced).
- Full runtime suite: **307 tests / 19 files passing**. `tsc --noEmit` clean on both `runtime` and `spec`.

### DIFF
--- a/.changeset/seed-identity-binding.md
+++ b/.changeset/seed-identity-binding.md
@@ -1,0 +1,23 @@
+---
+"@objectstack/spec": minor
+"@objectstack/runtime": minor
+---
+
+Seed data: first-class identity binding + loud failures (fixes #1389)
+
+Records seeded via `defineDataset` / `defineStack({ data })` can now bind to a
+platform user with `cel\`os.user.id\`` (and to the org with `cel\`os.org.id\``),
+which previously never resolved at boot.
+
+- **`os.user` / `os.org` now actually resolve.** The runtime provisions a
+  deterministic, non-loginable system user (`usr_system`, role `system`)
+  *before* any seed runs and binds it to `os.user`, so identity-derived seed
+  values resolve even on a fresh boot — before the first human sign-up. The
+  human login admin remains a separate better-auth identity and need not own
+  seed data. Exposed as the canonical `SystemUserId.SYSTEM` constant.
+- **New `SeedLoaderConfig.identity`** carries the `os.user` / `os.org` subject
+  into CEL evaluation (`@objectstack/spec`).
+- **Failures are loud, not silent.** A record whose CEL value can't resolve
+  (e.g. a required `cel\`os.user.id\`` with no identity) — or that fails to
+  write — is now counted as an error, marks the load unsuccessful, and logs an
+  actionable message, instead of being silently dropped.
--- a/content/docs/guides/seed-data.mdx
+++ b/content/docs/guides/seed-data.mdx
@@ -203,6 +203,84 @@ export const SeedData = [accountsSeed, contactsSeed];
 
 ---
 
+## Dynamic Values (CEL)
+
+Any field value may be a **CEL expression** evaluated at install time against a
+single per-load pinned `now`. This is the only correct way to author time-based
+or identity-derived seed values — a literal `new Date()` would ship the package
+author's clock to every customer and break build determinism.
+
+```typescript
+import { defineDataset, cel } from '@objectstack/spec';
+
+defineDataset(Opportunity, {
+  records: [{
+    name:            'Acme Q3 Renewal',
+    close_date:      cel`daysFromNow(45)`,
+    created_at:      cel`now()`,
+    owner_id:        cel`os.user.id`,   // the seed identity
+    organization_id: cel`os.org.id`,
+  }],
+});
+```
+
+Available in the seed CEL context:
+
+- **Functions:** `now()`, `today()`, `daysFromNow(n)`, `daysAgo(n)`,
+  `isBlank(v)`, `coalesce(v, fallback)`
+- **Scope:** `os.user`, `os.org`, `os.env`
+
+### Binding records to a user (`os.user`)
+
+Many objects have a **required** owner lookup — `owner_id`, `created_by`,
+`assigned_to`. To seed such a record, bind it to a user with `cel\`os.user.id\``.
+This is the single canonical convention; there is no `currentUser()`, `@admin`,
+or similar special syntax.
+
+```typescript
+defineDataset(Project, {
+  externalId: 'code',
+  records: [{
+    code:     'bootstrap',
+    name:     'Bootstrap Project',
+    owner_id: cel`os.user.id`,   // ← bound to the seed identity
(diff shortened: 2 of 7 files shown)


xuyushun441-sys:
## CI diagnosis — both failures are pre-existing on `main`, not introduced by this PR

I looked into the **Test Core** and **Build Docs** failures. Neither is caused by the seed changes in this PR — both are latent issues that live on `main`:

### Test Core — stale assertion in `packages/objectql/src/protocol-meta.test.ts`
`"should load records of different types"` asserts:
```ts
expect(registry.getItem('object', 'task')).toEqual(objDef)
```
The loaded record now carries extra `_packageId: "sys_metadata"` + `_provenance: "package"` annotations added by the package-bound-metadata feature (`loadMetaFromDb` → `registerObject(data, record.packageId || 'sys_metadata')`). The assertion was never updated for those fields. This test is **already failing on `main`** (main's Test Core conclusion = failure), and it has nothing to do with seed binding — running it with `-t` confirms the seed PR's new `describe` block never executes yet the failure persists.

### Build Docs — doc generator emits unescaped angle brackets into MDX
Turbopack fails on `./content/docs/references/ui/app.mdx` with `Expected a closing tag for <id>`. Root cause: the schema→MDX generator passes `.describe()` text through verbatim, so placeholders like `<id>` (e.g. `os projects bind <id>` in `cloud/marketplace.mdx:40`), `Promise<void>` and `Record<string, any[]>` (`memory.zod.ts`), `<pkg>`, `<h1>` are rendered as raw JSX-like tags and break the MDX parse. **Build Docs is `skipped` on `main`**, so the breakage is latent there; it only surfaced here because this PR touched a docs-adjacent path and triggered the job.

### Recommendation
Both belong in their own fix against `main` (update the provenance-aware test assertion; make the doc generator HTML-escape `<`/`>` in description text), owned by the respective subsystems — not on this merged seed PR. Flagging here for the record.

PR #1670, opened 2026-06-09
fix(i18n): translate Setup System Settings menu labels and pages

baozhoutao:
## Problem

> **i18n Finding** — System Settings has untranslated menu labels/pages.

The Setup app ("System Settings") rendered most of its left-nav menu in
English even for `zh-CN` / `ja-JP` / `es-ES` users. Investigation found
**two independent causes** — a missing-data gap *and* a structural bug at
the REST translation boundary.

### 1. Root cause — `getMetaItem` envelope was never unwrapped before translation

`GET /meta/app/setup` for the single-item app route **bypasses the HTTP
cache by design** (per-user RBAC filtering via `filterAppForUser`), so it
received the raw `getMetaItem` envelope `{ type, name, item, lock, … }`
and passed *the whole envelope* to `translateMetadataDocument`. But the
translatable document — and its `navigation` tree — lives nested at
`.item`. The resolver found no `navigation` on the envelope's top level,
so it walked nothing and **left every menu label in English**; only the
envelope's own top-level `label` got translated, which consumers reading
`.item` never see.

The cached read path passes the already-unwrapped document, so it was
never affected — which is exactly why the bug only showed on the
cache-bypassing app route.

A parallel bug: list endpoints (`{ items: [...] }` shape) were dropped
untranslated by a bare `Array.isArray(items)` guard.

### 2. Missing translation keys

Several Setup nav entries had **no bundle key in any locale**, so they
fell back to English regardless of the above:
- Configuration settings pages: `nav_settings_auth/storage/ai/knowledge/feature_flags`
- `group_integrations` (the Setup app's own empty integrations anchor)
- `nav_notification_{preferences,subscriptions,templates}` (contributed to
  `group_configuration` by `@objectstack/service-messaging`, which ships
  no translation bundle)
- Per-manifest settings **page** content (`auth`/`ai`/`knowledge`) missing
  for some locales.

## Changes

**`packages/rest/src/rest-server.ts`** — the structural fix
- Add `isMetaEnvelope()` and unwrap the inner document before calling
  `translateMetadataDocument` in both `translateMetaItem` and
  `translateMetaItems`.
- `translateMetaItems` now also handles the `{ items: [...] }` list-envelope
  shape (previously dropped untranslated).

**`packages/rest/src/rest.test.ts`** — 4 regression tests covering envelope
unwrap, the bare-document path, and both list shapes (`{items}` + raw array).

**`packages/platform-objects/src/apps/translations/{en,zh-CN,ja-JP,es-ES}.ts`**
— Setup nav menu labels: `nav_settings_auth/storage/ai/knowledge/feature_flags`,
`group_integrations`, and `nav_notification_{preferences,subscriptions,templates}`.

**`packages/services/service-settings/src/translations/{zh-CN,ja-JP,es-ES}.ts`**
— settings **page** content for `auth` (+ `ai`/`knowledge` for `ja-JP`).

`en` stays the source of truth; all keys mirror the manifest/nav-node names
so the locale → fallback → literal resolver chain stays intact.

## Verification

End-to-end against a running dev server (`zh-CN`):

- **API** — `GET /api/v1/meta/app/setup` with `Accept-Language: zh-CN`:
  `item.label` → **系统设置**, and **0 remaining English nav labels** across
  the whole tree (every group + leaf, including the `type:'url'` settings
  items and the notification entries).
- **UI** — the Console's System Settings sidebar renders fully localized:
  全部设置 · 品牌 · 认证 · 邮件 · 文件存储 · AI 与 Embedder · 知识库 ·
  功能开关 · 通知偏好 · 通知订阅 · 通知模板.
- **Tests** — `@objectstack/rest` 100/100 pass (96 existing + 4 new
  regression cases). `@objectstack/platform-objects` builds clean.

## Scope / follow-ups

- Covers the four shipped locales (`en`, `zh-CN`, `ja-JP`, `es-ES`).
- The notification nav labels currently live in the central
  `SetupAppTranslations` (alongside `nav_notifications`) since they merge
  into the Setup app by stable id. If `@objectstack/service-messaging`
  later grows its own translation bundle, they could move there.

### DIFF
--- a/packages/platform-objects/src/apps/translations/en.ts
+++ b/packages/platform-objects/src/apps/translations/en.ts
@@ -50,6 +50,7 @@ export const en: TranslationData = {
         group_access_control: { label: 'Access Control' },
         group_approvals: { label: 'Approvals' },
         group_configuration: { label: 'Configuration' },
+        group_integrations: { label: 'Integrations' },
         group_diagnostics: { label: 'Diagnostics' },
         group_advanced: { label: 'Advanced' },
 
@@ -88,6 +89,10 @@ export const en: TranslationData = {
         nav_settings_ai: { label: 'AI & Embedder' },
         nav_settings_knowledge: { label: 'Knowledge' },
         nav_settings_feature_flags: { label: 'Feature Flags' },
+        // Notifications (contributed by @objectstack/service-messaging)
+        nav_notification_preferences: { label: 'Notification Preferences' },
+        nav_notification_subscriptions: { label: 'Notification Subscriptions' },
+        nav_notification_templates: { label: 'Notification Templates' },
 
         // Diagnostics
         nav_sessions: { label: 'Sessions' },
--- a/packages/platform-objects/src/apps/translations/en.ts
+++ b/packages/platform-objects/src/apps/translations/en.ts
@@ -84,6 +84,9 @@ export const en: TranslationData = {
         nav_settings_mail: { label: 'Email' },
         nav_settings_branding: { label: 'Branding' },
         nav_settings_auth: { label: 'Authentication' },
+        nav_settings_storage: { label: 'File Storage' },
+        nav_settings_ai: { label: 'AI & Embedder' },
+        nav_settings_knowledge: { label: 'Knowledge' },
         nav_settings_feature_flags: { label: 'Feature Flags' },
 
         // Diagnostics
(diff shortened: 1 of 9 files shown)


baozhoutao:
Scope note: while verifying this PR I found that the System Settings pages (`/apps/setup/system/settings`, `.../branding`, `.../auth`, …) render **blank** in the console. That is a separate issue — the console SPA (objectui) has no `system/settings` route or manifest-driven settings UI; the `/api/settings` backend works and the menu labels (this PR) translate correctly. Tracked separately as objectstack-ai/objectui#1609 so this i18n PR stays focused.

Issue #1928, opened 2026-06-15
Formula guardrail: cel-js arithmetic silently returns null (double × int + bare identifiers)


os-zhuang:
## Summary

Two formula-authoring footguns let an AI write a reasonable-looking `Field.formula` that **passes `objectstack build` but silently evaluates to `null` at runtime** — the failure class our guardrails are meant to eliminate. Both were found during 9.5.x release-prep testing of `example-crm` (fixed there in #1927; this issue tracks the *systemic* prevention).

## 1. No `double × int` arithmetic overload (primary)

cel-js (`@marcbachmann/cel-js`) types a record field number as `double` and a bare integer literal as `int`, and has **no mixed arithmetic overload**:

```
record.amount / 100              → ERR  no such overload: dyn<double> / int   → null
record.amount * 2                → ERR  no such overload: dyn<double> * int    → null
record.amount * record.probability / 100.0   → OK  (float literal)
record.amount >= 4               → OK   (cel-js DOES have mixed *comparison* overloads)
```

So `amount / 100`, `price * 2`, `total - fee` (int literal) all silently null. Only float literals (`100.0`, `2.0`) work. Integer literals in arithmetic are ubiquitous → high-frequency footgun.

**Constraint:** cel-js **rejects** registering operator overloads — `env.registerFunction('_/_(double, int): double', …)` → `Invalid signature`. So there is no trivial overload-registration fix.

### Options
- **(a) Build-time lint** — in `validate-expressions.ts` (ADR-0032), flag a number-typed field used as an arithmetic operand against an `int` literal; message: "use a float literal (e.g. `/ 100.0`)". Cheapest, matches the advisory-warning guardrail strategy.
- **(b) Engine AST promotion** — in the CEL engine, walk the parsed AST and promote `int` literals that are **arithmetic operands** to `double`. Must NOT touch `int`-typed function args (e.g. `daysFromNow(30)`). More robust (fixes silently), more invasive, needs tests.
- (a) and (b) are complementary.

## 2. Bare identifiers in formula expressions (secondary)

Formula scope exposes only `{record, previous, input, os}` (see `buildScope`, `packages/formula/src/stdlib.ts`) — **no field flattening**. So:

```
cel`status == "converted"`        → null   (bare `status`)
cel`record.status == "converted"` → OK
```

[...]

PR #1975, opened 2026-06-16
docs(adr): ADR-0053 — Phase 2 detailed design (reference-timezone model)

os-zhuang:
## What

Adds a **Phase 2 — Detailed Design** section to [ADR-0053](https://github.com/objectstack-ai/framework/blob/docs/adr-0053-phase2-design/docs/adr/0053-date-and-datetime-semantics.md) now that **Phase 1 landed** ([#1968](https://github.com/objectstack-ai/framework/pull/1968) — `Field.date` is a tz-naive `YYYY-MM-DD` string at the driver boundary).

This is the **plan for review**, not a commitment to ship. Each slice is independently revertable and gated behind "reference timezone unset → UTC" (today's behavior). Refs ADR-0053, #1928.

## Highlights

- **Two-axis reframe** — `compute-tz` (`today()`/`daysFromNow`, analytics bucketing, cron day-boundaries) vs `render-tz` (datetime display): distinct sources, distinct risk. Conflating them is the chief risk of Phase 2.
- **Reference-timezone resolver** — precedence (user pref → tenant-scoped org default → UTC), the single `resolveExecutionContext` chokepoint, a new `ExecutionContext.timezone` field, and the three execution branches (interactive / scheduled job / flow).
- **Shared DST-safe primitive** reused from `preference-resolver`'s `Intl.formatToParts` pattern — no hand-rolled offset math.
- **Three decisions, with rationale:**
  - **D1** — `today()`/`daysFromNow()`/`daysAgo()` return a `Date` at **UTC-midnight of the reference-tz calendar day** (forced by cel-js hydration; explicit callout of the "local-midnight-as-instant" trap that would re-introduce the silent-miss bug). Side-effect: fixes the "keeps wall-clock time" defect for free.
  - **D2** — tz-aware analytics buckets **in-memory, uniformly**; do not emit dialect-specific `date_trunc … AT TIME ZONE` (SQLite/MySQL have no tz support → cross-driver bucket skew).
  - **D3** — wire report schedules onto the **existing** croner+tz adapter rather than delete the dormant `cron_expression`/`timezone` fields (the *enforce* resolution the spec-liveness gate wants).
- **`applyFormulaPlan` blind spot** — read-time formula fields evaluate with no `now`/ctx/tz; worth fixing regardless of timezone.
- **Per-slice implementation map** (6 slices, file-level), **open prerequisites** (cel-js `duration()` support; email rendering bypasses the formatter pipeline), and the **feature-flag rollback**.

## Follow-up

The 6 implementation slices are being filed as separate tracked issues (linked under #1928).

### DIFF
--- a/docs/adr/0053-date-and-datetime-semantics.md
+++ b/docs/adr/0053-date-and-datetime-semantics.md
@@ -179,3 +179,156 @@ rendering.
    Rejected as a complete fix: it helps the formula path but not the
    write/filter asymmetry, and it bakes in UTC midnight (Phase 2's tz-aware
    `today()` supersedes it).
+
+---
+
+## Phase 2 — Detailed Design (implementation plan)
+
+> **Status:** Phase 1 landed in `@objectstack/driver-sql` (PR #1968) — `Field.date`
+> is now a tz-naive `YYYY-MM-DD` string at the write/read boundary. This section
+> details Phase 2 (the reference-timezone model) for review. It is the plan, not
+> yet a commitment to ship; each slice below is independently revertable and gated
+> behind "reference timezone unset → UTC" (today's behavior).
+
+### Reframe: two axes, two sources, two risk profiles
+
+ADR-0053's "timezone model" is, in the code, **two distinct concerns** that must
+be solved separately. Conflating them in one change is the chief risk of Phase 2.
+
+| Axis | Question it answers | Source | Boundary |
+|------|--------------------|--------|----------|
+| **Compute-tz** | "what calendar day is it / which bucket does this fall in?" — `today()`/`daysFromNow()`, analytics bucketing, cron day-boundaries | the **execution context** (user / org / job) | threaded *into* evaluation |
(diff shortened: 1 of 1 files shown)


os-zhuang:
### Phase 2 implementation slices (tracked)

Filed from the per-slice map in this ADR section. Each is independently revertable and gated behind "reference tz unset → UTC":

1. #1978 — reference-timezone resolver + `ExecutionContext.timezone` *(foundational; blocks 3–5)*
2. #1979 — thread context into `applyFormulaPlan` *(independent; worth doing on its own)*
3. #1980 — tz-aware `today()`/`daysFromNow()`/`daysAgo()` *(produces the shared `partsInTz` util)*
4. #1981 — render datetime in reference tz (formatters + email)
5. #1982 — tz-aware analytics date bucketing *(depends on 1 + 3)*
6. #1983 — activate report-schedule cron + timezone *(self-contained; liveness enforce)*

Cron day-boundaries (`sys_job`) need no change — already tz-wired via croner.

PR #2021, opened 2026-06-18
feat(ownership): auto-provision canonical owner_id + hand seeded records to first admin

os-zhuang:
## Problem

Record ownership was **opt-in** and the boot order made it impossible to seed ownership to the eventual human admin:

- Data seed runs in `app-plugin.start()` — **before** any human user exists (the login admin is minted later, on `kernel:ready`). So `os.user` is bound to a deterministic, non-loginable `usr_system`, and seeded rows land owned by `usr_system` or `NULL`.
- The human admin's id is random (better-auth), so it can't be referenced from static seed.
- Apps "almost never declare `owner_id`" (the platform's own RLS comment), so they reinvent a custom `owner` lookup that the platform's ownership machinery (auto-stamp, RLS, "My" views) can't see.

Net effect: seeded demo data is owned by nobody a human can log in as → My-views / owner reports / owner notifications are empty out of the box, and author-written objects silently ship with no working ownership at all. This is the worst failure mode for AI authors: it compiles, seeds and renders, but the feature is silently dead.

## Fix — ownership correct-by-default

**1. `applySystemFields` (objectql)** auto-injects a canonical, **reassignable** `owner_id` lookup (→ `sys_user`) on user-authored business objects, next to the existing tenant/audit fields. Unlike the audit `*_by` lookups it is NOT readonly — ownership transfers. Withheld for `managedBy` / `sys_*` tables and for objects that opt out via `ownership: 'org' | 'none'` (Dataverse-style). Once present, the existing machinery engages automatically (insert auto-stamp at security step 3.5, owner RLS, owner-keyed views/reports).

> Safe default direction: forgetting the opt-out leaves a harmless spare nullable column; the old opt-IN model let authors ship broken ownership.

**2. `claimSeedOwnership` (plugin-security)** — invoked from `bootstrapPlatformAdmin` right after the first human is promoted to platform admin — transfers ownership of seeded rows (`owner_id` NULL or `usr_system`) to that admin. The ownership twin of org-scoping's `claimOrphanOrgRows`. Idempotent; skips `managedBy` / `sys_*`. **Authors write plain seed records (no `owner_id`); the platform performs the handoff** — nothing to remember or mistype.

**3.** Hardens `bootstrapPlatformAdmin` against a latent dts typecheck error (defensive read of the untyped `description` on seed permission sets, surfaced when the cache invalidated).

## Why this shape (AI-authored code)

The design optimizes for *correct-by-default, loud-on-mistake* because the code — schemas and seeds — is largely AI-written:
- **Zero-config default**: write a normal object → ownership just works. Nothing to declare/forget.
- **Platform-decided exclusions**: `managedBy` / `sys_*` auto-excluded; only the rare catalog/junction needs one line `ownership: 'org'`.
- **Handoff, not author-supplied owner**: seeds stay plain; no `cel\`os.user.id\`` to mistype.

## Tests
- `registry.test.ts`: +5 owner_id injection cases (default inject, managedBy/sys_ exclusion, `ownership` opt-out, no-overwrite).
- `claim-seed-ownership.test.ts`: +6 (NULL & usr_system claim, human-owned untouched, managed/sys_ skip, idempotent, system-target no-op).
- **Full framework suite green: 128/128 tasks.**

## Follow-ups (not in this PR)
- Multi-tenant: mirror the handoff in `claimOrphanOrgRows` / `ensureDefaultOrganization` so per-org seed clones are owned by the org's admin.
- A compile-time lint that flags author-declared custom `owner`-ish lookups (now redundant alongside the auto `owner_id`).
- App migration: HotCRM's custom `owner` → canonical `owner_id`.

### DIFF
--- a/.changeset/seed-ownership-handoff.md
+++ b/.changeset/seed-ownership-handoff.md
@@ -29,12 +29,14 @@ author-written objects silently shipped with no working ownership at all.
   not the author — performs the handoff, so there is nothing to remember or
   mistype.
 
-- **`usr_system` is now provisioned lazily (runtime)** — only when a seed
-  dataset actually embeds `cel`os.user.id``. Because the default model leaves
-  `owner_id` NULL and relies on the handoff above, a typical bundle never
-  references `os.user`, so the non-loginable `usr_system` placeholder is never
-  created. It survives purely as a backward-compatible fallback; `os.org` is
-  unaffected (derived from `organizationId` in the loader).
+- **`usr_system` is never minted (runtime + objectql).** The seed loader binds
+  `os.user` to a NULL identity, so `cel`os.user.id`` resolves to NULL at seed
+  time (the owning admin does not exist yet) and the row seeds NULL-owned — then
+  the handoff above fills it. The runtime's `ensureSeedIdentity` (the only code
+  that inserted a `usr_system` row) is removed. `SystemUserId.SYSTEM` survives
+  only as a reserved id so legacy DBs' exclusion guards / ownership handoff still
+  recognize a pre-existing row. `os.org` is unaffected (derived from
+  `organizationId`).
 
 Also hardens `bootstrapPlatformAdmin` against a latent dts typecheck error
 (defensive read of the untyped `description` on seed permission sets).
--- a/.changeset/seed-ownership-handoff.md
+++ b/.changeset/seed-ownership-handoff.md
@@ -1,6 +1,7 @@
 ---
 "@objectstack/objectql": minor
 "@objectstack/plugin-security": minor
+"@objectstack/runtime": minor
 ---
 
 feat(ownership): auto-provision a canonical `owner_id` and hand seeded records to the first admin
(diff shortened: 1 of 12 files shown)


os-zhuang:
### Follow-up: `usr_system` demoted to a lazy fallback

Pushed 3 more commits addressing *"if the human admin owns everything via handoff, what is `usr_system` even for?"*

Audit across the whole framework (non-test): `usr_system` is **created once** (to bind `os.user` during seed CEL eval) and **only ever excluded everywhere else** (3 guards that stop it from being mistaken for a human admin). No runtime code ever acts *as* `usr_system` — system operations use the `{ isSystem: true }` context flag, not that userId. And no real seed dataset embeds `cel\`os.user.id\``.

So its sole job had collapsed to "seed-time `os.user` binding." With the NULL-owner + handoff model, that job disappears. Change:

- **`ensureSeedIdentity` is now invoked lazily** — only when a seed dataset actually embeds `cel\`os.user.id\`` (detected via the serialized CEL `source`). A typical bundle leaves `owner_id` NULL, so `usr_system` is **never created**. It remains purely as a backward-compatible fallback for any seed that still references `os.user`. `os.org` is unaffected (derived from `organizationId`).
- Docs on `SystemUserId.SYSTEM` and `ensureSeedIdentity` updated to describe the lazy-fallback role.

Net: the default boot no longer creates a placeholder system user at all; seeded rows are NULL-owned → handed to the first real admin. Same end result, one fewer magic identity.

Build green (22 tasks); runtime 388 + plugin-security 97 tests pass.

os-zhuang:
### usr_system eliminated entirely (not just lazy)

Per feedback — don't create usr_system at all; scheme for seeds that use `cel\`os.user.id\``:

**Resolve `os.user.id` to NULL at seed time.** It semantically means "the user installing this seed" — who doesn't exist yet — so the honest value is *unknown → NULL → resolved later by the first-admin handoff*.

- Seed loader binds `os.user` to a NULL identity (`{ id: null }`) → `cel\`os.user.id\`` evaluates to `null` (no crash, no dropped record). Row seeds NULL-owned; `claimSeedOwnership` re-owns it to the promoted admin.
- `ensureSeedIdentity` (the only code that inserted a `usr_system` row) is **removed** — a fresh boot never creates the placeholder.
- `SystemUserId.SYSTEM` kept only as a reserved id so legacy DBs' exclusion guards / handoff still recognize a pre-existing row.
- Edge: `os.user.id` on a required non-owner field now fails loudly at validation (correct).

Tests green: objectql 665 · runtime 388 · plugin-security 97.

PR #2167, opened 2026-06-22
docs(adr): two agents bound by surface (revise ADR-0063) + ADR-0064 tool scoping

xuyushun441-sys:


xuyushun441-sys:
Implementation tracked in #2168.

PR #2192, opened 2026-06-22
fix(ai): keep platform agents in catalog regardless of alias-map timing (ADR-0063 follow-up)

xuyushun441-sys:
Follow-up to #2181 (ADR-0063/0064). Found during live verification on the EE rig: `GET /api/v1/ai/agents` returned only `{ask}` — the `build` agent was missing — because `listAgents()` filtered the catalog by the in-memory `registerAgentAlias` values, and the cloud studio package's `metadata_assistant`→`build` alias wasn't applied to the runtime's map at request time (bundle module-load ordering). A missed alias must never hide a real platform agent.

## Change
- `listAgents()` recognises a platform agent by the intrinsic package-protection envelope stamped on built-ins at registration (`_provenance`/`_lock`/`_packageId`, or a still-public `protection` block) via `isPlatformAgentRecord(...)`, with the alias-table values kept only as a belt-and-suspenders fallback.
- Stray tenant custom agents (ADR-0063 §2, withdrawn) still stay filtered out of the catalog.

Pairs with cloud-side robustness for the studio alias registration (objectstack-ai/cloud#419).

Verified on the EE rig after the fix; no spec/schema change (no liveness impact).

### DIFF
--- a/.changeset/ai-agent-catalog-platform-filter.md
+++ b/.changeset/ai-agent-catalog-platform-filter.md
@@ -1,0 +1,15 @@
+---
+"@objectstack/service-ai": patch
+---
+
+fix(ai): keep platform agents in the runtime catalog regardless of alias-map timing.
+
+`AgentRuntime.listAgents()` filtered the catalog to platform agents by the
+in-memory `registerAgentAlias` values, so a missed/late alias registration
+(e.g. a cloud package's module-load `registerAgentAlias('metadata_assistant',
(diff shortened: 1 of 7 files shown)


xuyushun441-sys:
Closing as a duplicate — the framework catalog/alias fix already landed via #2190 (commit 6fefb1aaf), which cloud#419 pins. This branch is an earlier divergent copy of the same fix; merging it would regress #2190's refinements.

PR #2367, opened 2026-06-27
docs(adr): ADR-0075 — configurable, opt-out-able post-login landing

os-zhuang:
## Design proposal — for review (docs only)

Makes the **post-login landing** a metadata-declared, opt-out-able capability instead of a hardcode.

### Problem
The console hardcodes where users land after sign-in: `CloudAwareRootRedirect` carries a literal `PREFERRED_APPS = ['cloud_control']` (→ `/apps/cloud_control`, else the `/home` launcher). A deployment can neither **opt out** of the outer `/home` (e.g. our cloud service doesn't want it) nor **land somewhere custom** without forking the SPA. App `isDefault` is a badge that doesn't route.

### Key framing (from review)
The landing is a **product-definition decision made at build/dev time** by whoever builds a product on the framework (declared in metadata/source) — **not** a runtime, per-tenant, admin-UI preference.

### Proposal (metadata-native, near-zero new schema)
Resolve from existing app metadata:
- **Wire `isDefault` into root routing** — land in the `isDefault` app; its existing `homePageId` selects the landing page inside it. "Custom home" = your app + its `homePageId` (an ADR-0065 page). No settings namespace, no `page:<name>` top-level concept.
- **Fallback** (nobody declares `isDefault`): single visible app → it; else today's `/home` launcher (unchanged).
- Replace `CloudAwareRootRedirect`'s hardcode with a generic `LandingResolver`. Cloud already has `cloud_control` `isDefault: true`, so behavior is preserved and the hardcode is deleted.
- An env var stays only as a secondary **ops escape hatch**; metadata is the source of truth.

### Explicitly NOT doing
No `console.default_landing` Settings-Manifest namespace, no per-tenant `sys_setting`, no Settings UI, no per-tenant runtime landing. (An earlier draft proposed the Settings route — dropped per review.)

### Back-compat
No `isDefault` app → unchanged `/home`. Cloud → unchanged (still lands on `cloud_control`), minus the hardcode. Likely **no framework spec change** beyond documenting `isDefault`'s routing semantics; the work concentrates in objectui.

**Seeking design sign-off before implementing.**

### DIFF
--- a/docs/adr/0075-post-login-landing-resolution.md
+++ b/docs/adr/0075-post-login-landing-resolution.md
@@ -1,0 +1,152 @@
+# ADR-0075: Post-login landing resolution (configurable, opt-out-able home)
+
+- Status: Proposed
+- Date: 2026-06-27
+- Deciders: platform
+- Related: ADR-0007 (Settings Manifest + K/V Store + Resolver), ADR-0065 (server-driven UI / pages)
+
+## Context
+
+After sign-in the console must land the user *somewhere*. Today that choice is
+**hardcoded in the console SPA**:
+
+- `objectui/apps/console/src/components/CloudAwareRootRedirect.tsx` resolves `/`
+  with a literal `PREFERRED_APPS = ['cloud_control']`: if the `cloud_control`
+  app exists (cloud deployments) it redirects to `/apps/cloud_control`,
+  otherwise it falls back to `/home`.
+- `/home` renders `DefaultHomePage` (`@object-ui/app-shell` —
+  `console/home/HomePage.tsx`): a generic workspace launcher ("Your apps",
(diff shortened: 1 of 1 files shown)


os-zhuang:
Closing without merging — per review, this change is too small to warrant a standalone ADR. It reuses existing app metadata (`isDefault` + `homePageId`), the only real change being a generic landing resolver in objectui replacing the `PREFERRED_APPS=['cloud_control']` hardcode. The one durable note (`isDefault` gains routing semantics — a back-compat contract change — and why it's metadata-declared rather than a Settings UI) will live in the implementation PR description + changeset + a code comment. Superseded by the forthcoming objectui implementation PR.

PR #2421, opened 2026-06-28
docs(adr): ADR-0078 — segment the ObjectStackProtocol god-interface (ISP); keep the engine a pure primitive

os-zhuang:
## What

**ADR-0078** records a deep kernel-architecture assessment and the decision that follows. It builds on [ADR-0076](https://github.com/objectstack-ai/framework/blob/main/docs/adr/0076-objectql-core-tiering.md) (which handled packaging) and addresses the **deeper contract shape** 0076 did not.

## Findings (verified against source)

- **The engine is the crown jewel and it's correct** — `engine.ts` (~3.3k LOC) hard-codes **zero** governance; RLS/RBAC/field-masking/owner/tenant are all pluggable via `registerMiddleware` + hooks. Driver port minimal. **Protect it; don't add protocol/governance to it.**
- **The real debt is `ObjectStackProtocol`: a 70-method, 11-domain god-interface** (Data 9, Metadata 8, Feed 13, Notifications 7, Realtime 6, Packages 6, Views 5, …). **60/70 optional; no consumer uses >11%.** It aggregates domains that already have their own services, which forces a ~5.6k-LOC facade and makes the contract un-versionable. It's also the root cause of the `metadata-protocol` naming confusion (the package carries the thin data facade too).

## Decision

- **D1** Ratify the engine as a pure primitive (protect over tidiness).
- **D2** Segment the contract per ISP → `DataProtocol` + `MetadataProtocol` + optional capability protocols (Analytics/Feed/Realtime/Notification/View/…); keep a composed alias `ObjectStackProtocol = DataProtocol & MetadataProtocol & Partial<…>` for back-compat. **Spec/type-level, incremental, low-risk.**
- **D3** Impl + package naming follow the contract; capability domains served by their existing services, not re-bundled.
- **D4** Contract segmentation can start now (incremental); the impl restructure + rename ride the **cross-repo window with ADR-0076 Step 2** (cloud constructs `ObjectQLPlugin` directly in ~8 sites).

## Scope

Docs only — adds `docs/adr/0078-protocol-segmentation.md` and a forward-pointer in ADR-0076. No code. Empty changeset (no release impact).

### DIFF
--- a/.changeset/adr-0078-protocol-segmentation.md
+++ b/.changeset/adr-0078-protocol-segmentation.md
@@ -1,0 +1,4 @@
+---
+---
+
+docs(adr): ADR-0078 protocol segmentation. Docs only — no release impact.
--- a/docs/adr/0076-objectql-core-tiering.md
+++ b/docs/adr/0076-objectql-core-tiering.md
@@ -1,6 +1,7 @@
 # ADR-0076: objectql is the data engine — relocate metadata management (protocol) out of it; enforce the boundary; defer the engine repo-split
 
 **Status**: Proposed (2026-06-28, rev. 4 — data-driven; centerpiece corrected to protocol relocation) — v12 assessment
+**Follow-up**: [ADR-0078](./0078-protocol-segmentation.md) — segments the `ObjectStackProtocol` god-interface (the deeper contract shape this ADR did not address; engine stays a pure primitive).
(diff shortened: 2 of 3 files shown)


os-zhuang:
Folding into ADR-0076 instead of a separate ADR (per maintainer: fewer ADRs).

PR #2595, opened 2026-07-04
feat(spec,lint): adaptive record surface + semantic span + navigation.size (#2578)

os-zhuang:
Framework / spec side of #2578. All metadata is AI-authored, so the design goal is to make AI unable to get presentation wrong — presentation (surface / width / columns) is runtime-derivable, so **none of it is an authored key**.

## What

- **`deriveRecordSurface`** (`spec/src/data/record-surface.ts`) — the default record surface (page vs drawer/modal) derived from authorable field count; mobile always pages. A **derivation, not an authored key**: a `recordSurface` object key fails ADR-0085 §2's admission test (field count is machine-inferable). Explicit override remains the assigned-page path.
- **`FormField.span`** (`auto` | `full`) — relative field width, decoupled from the per-surface derived column count; replaces the fragile absolute `colSpan` (retained + clamped by the renderer).
- **`NavigationConfig.size`** (`auto` | sm | md | lg | xl | full) — T-shirt overlay size aligned with `FormView.modalSize`; **deprecates the pixel `width` / `drawerWidth`** (a pixel width can't be authored blind — the author doesn't know the client viewport). `auto` ⇒ renderer derives size from field count + viewport, AI writes nothing.
- **`validateFormLayout`** lint — `form-field-unknown` + `absolute-colspan-discouraged` (both warnings with fix hints; ADR-0078/0019, same bar for AI and hand authors).
- ADR-0085 §2 clarification (records `recordSurface` as rejected, to prevent re-proposal) + changeset.

## Tests
- spec **6677** tests + `check:api-surface` (regenerated, additive) — green
- lint **121** tests — green

## Companion
Renderer side lives in **objectstack-ai/objectui** (companion PR). Build-independent: objectui ships a local mirror of `deriveRecordSurface` pending this spec's release, then consolidates to the import.

Refs #2578


os-zhuang:
Renderer companion: objectstack-ai/objectui#2237

PR #2680, opened 2026-07-07
perf(seed,import): route bulk writes through the engine's batch insert path

### LINKED ISSUE #2678: perf/reliability: seed + data-import bypass the engine's bulk-insert path → N round-trips, N summary recomputes, no transient retry
## Summary

The two bulk-write paths — the **seed loader** (`defineDataset` replay) and the user-facing **data import** (CSV / bulk records) — write records **one at a time** through the single-record engine path, even though ObjectQL's engine **already has an efficient batched branch** (`driver.bulkCreate` + parent-deduplicated summary recompute). Over a remote DB (turso) this makes both:

- **slow** — N round-trips + N summary recomputes instead of ~N/batch, and
- **fragile** — each round-trip can hit a transient network error that is caught → the row is dropped → **no retry**.

Surfaced while debugging the 2026-07-06 HotCRM first-boot incident: a HotCRM seed's `crm_contact` inserts intermittently dropped rows on transient turso errors → the seed reported errors → the runtime's seed-once stamp never landed → per-env kernel churn (re-seed every boot). Non-deterministic (network-sensitive): the same seed inserted 5/5 contacts on one env and 1/5 on another.

### Seed loader — per record
`packages/metadata-protocol/src/seed-loader.ts` — `loadDataset()` loops per record calling `this.engine.insert(objectName, record, opts)` with a **single record**. No array / bulk form is ever constructed.

### Data import — per record
Entry points (`packages/rest/src/rest-server.ts`):
- `POST /data/:object/import` (sync, ≤5k rows) — ~L3524
- `POST /data/:object/import/jobs` (async, 5k–50k, persists `sys_import_job`) — ~L3598
- `POST /data/:object/createMany` (optional bulk endpoint, may be unwired) — ~L5749

Both sync + async delegate to `packages/rest/src/import-runner.ts` `runImport()`, which loops per row calling `p.createData({...})` / `p.updateData({...})` with a **single record** (~L201–274). It does **not** construct arrays or use the engine's bulk branch.

### The engine ALREADY has the efficient bulk primitive
`packages/objectql/src/engine.ts`:
- `insert(object, data: any | any[])` — the `Array.isArray(hookContext.input.data)` branch (~L2158) routes to `driver.bulkCreate(...)` — **one round-trip per batch**.
- `recomputeSummaries()` (~L1746–1782) **deduplicates parent ids across the whole array** and recomputes **once per unique parent** (1000 children of the same parent = 1 rollup recompute, not 1000). Called once per `insert()` call (~L2206).

Because neither seed nor import passes arrays, both get **N single-record `insert()` calls → N round-trips + N `recomputeSummaries()` invocations**.

### No transient-error retry (silent row loss)
- `import-runner.ts` ~L260: `catch → results.push({ ok:false, action:'failed', ... })` → continue. No retry.
- `seed-loader.ts` ~L413: `catch → errors.push(...)` → continue. No retry.

A transient driver/network blip (`fetch failed` / timeout / connection reset) on one row silently drops that row. Import returns a per-row report, so it's visible, but there's no automatic recovery.

## Impact

Import 1000 rows into an object with a rollup summary, over a remote turso DB (~100ms/round-trip):
- 1000 inserts + up to 1000 summary recomputes = **2000+ round-trips ≈ 3+ minutes**, plus a handful of transient failures that drop rows.
- With K rollup parents: 1000 inserts + K×1000 aggregate queries.

The seed variant additionally triggered **kernel churn** in the HotCRM incident (dropped seed rows → `errs>0` → the cloud seed-once stamp never lands → re-seed every cold boot).

## Proposed fix

A **shared batched-write helper** used by BOTH `seed-loader` and `import-runner`:

1. **Batch** rows (e.g. 100–500/batch) and call the engine's array/bulk `insert` → `driver.bulkCreate` (one round-trip per batch). The engine already dedups summary recomputes for arrays.
2. **Retry TRANSIENT driver errors** (network / timeout / connection reset) per batch with backoff. Do **not** retry logical errors (validation / constraint).
3. On a batch that fails on a **logical** error, **degrade to per-row within that batch** so one bad row doesn't drop the whole batch (bad row reported, others succeed).
4. (Optional) collapse summary recompute to once per parent at the end of the whole load.

Related runtime-side follow-up (cloud, tracked separately): the seed-once stamp guard is `errs===0` — a single transient error blocks the stamp forever → churn; relax to stamp-on-progress + surface residual errors via `reportObserved`.

## Acceptance criteria

- Seed + import route through the shared bulk helper; a 1000-row load over a `bulkCreate`-capable driver issues **~ceil(N/batch)** writes, not N.
- Summary-recompute count is bounded by **unique parents**, not row count.
- A transient driver error is **retried** and does not drop the row; a logical error is reported per-row **without failing the batch**.
- Tests: bulk-path write-count assertion, transient-retry, logical-error per-row degradation, summary-dedup.

## Refs
- HotCRM new-workspace "打开环境一直失败" incident — cloud#752 / #754 / #756, framework#2661.

os-zhuang:
## Summary

Fixes #2678.

The seed loader (`defineDataset` replay) and the data-import runner both wrote records **one at a time** through the single-record engine path, even though ObjectQL's engine already has an efficient batched branch (`driver.bulkCreate` + parent-deduplicated summary recompute) for the array form of `insert()`. Over a remote DB this meant N round-trips + N summary recomputes instead of ~N/batch, and a transient network error on any one row was silently dropped with no retry (the 2026-07-06 HotCRM first-boot incident).

- **New shared helper** `bulkWrite` / `withTransientRetry` (`packages/core/src/utils/bulk-write.ts`): batches rows through a caller-supplied batch-write function, retries a whole-batch transient failure (network blip / timeout) with exponential backoff, and degrades to per-row writes (each itself retried) when a batch fails for a non-transient reason — so one bad row can't drop the other N-1.
- **`SeedLoaderService.loadDataset()`** (`packages/metadata-protocol/src/seed-loader.ts`) buffers insert-mode records and flushes them in batches of 200 via the engine's array-form `insert()`. Datasets with a self-referencing field (e.g. `employee.manager_id -> employee`) keep the historical strictly-sequential per-record path, since a later record may need an earlier one's freshly-assigned id via `insertedRecords` — batching would defer that write past the point it's needed. Update-mode writes are wrapped in the same transient retry.
- **`runImport()`** (`packages/rest/src/import-runner.ts`) buffers create-resolved rows and flushes them via `protocol.createManyData()` when the protocol supports it (falling back to the original per-row `createData()` call otherwise). Per-row result ordering is preserved via an index-addressed results array even though creates are now written out of their original per-row sequence.
- **`Protocol.createManyData`** (`packages/metadata-protocol/src/protocol.ts`) now forwards `context` to `engine.insert()`, matching `createData` — previously it silently dropped tenant/security context, which is presumably why the issue called this endpoint "may be unwired".

## Test plan

- [x] New unit tests for `bulkWrite`/`withTransientRetry`/`defaultIsTransientError` (`packages/core/src/utils/bulk-write.test.ts`): bulk-path write-count assertion, transient-retry, logical-error per-row degradation, index correlation across batches.
- [x] New integration-style tests for `runImport()`'s batching (`packages/rest/src/import-runner-bulk.test.ts`): batch write-count, transient retry, logical degrade, no-`createManyData` fallback, result ordering with interleaved updates.
- [x] Updated existing mocks/assertions in `packages/runtime/src/seed-loader.test.ts` and `packages/runtime/src/http-dispatcher.test.ts` that asserted the old N-calls-per-record behavior, now asserting the batched call.
- [x] `pnpm test` (via turbo) green across `@objectstack/core`, `@objectstack/metadata-protocol`, `@objectstack/objectql`, `@objectstack/rest`, `@objectstack/runtime` — 461+797+216+15+... tests passing, no regressions.
- [x] `pnpm lint` clean on all changed files.

## Notes / follow-ups

- Summary-recompute count is bounded by unique parents *per batch* (not across the whole load) — collapsing it to a single end-of-load pass is called out as optional in the issue and is a larger engine-level change; left out of scope here.
- Import's `findExisting`/reference-resolution reads are unchanged (still per-row) — the issue's evidence and acceptance criteria are scoped to the write path.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MggPkEzsiBpWCQBAVVPD1V)_

### DIFF
--- a/.changeset/seed-import-bulk-insert.md
+++ b/.changeset/seed-import-bulk-insert.md
@@ -2,6 +2,7 @@
 '@objectstack/core': minor
 '@objectstack/metadata-protocol': minor
 '@objectstack/rest': minor
+'@objectstack/driver-sql': patch
 ---
 
 Seed loader and data-import now route bulk writes through the engine's array-form `insert()` (one round-trip per batch, with parent-deduplicated summary recompute) instead of one `insert()`/`createData()` call per record, and both retry transient driver errors instead of silently dropping the row (#2678).
@@ -12,3 +13,5 @@ A new shared helper, `bulkWrite` (`@objectstack/core`), batches rows through a c
 - `runImport()` (`@objectstack/rest`) buffers create-resolved rows and flushes them via `protocol.createManyData()` when the protocol supports it, falling back to the original per-row `createData()` call otherwise. `Protocol.createManyData` (`@objectstack/metadata-protocol`) now forwards `context` to `engine.insert()` like `createData` already did, so tenant-scoped bulk creates work correctly.
 
 Previously, a 1000-row seed or import into an object with a rollup summary issued 1000+ round-trips and up to 1000 summary recomputes; a single transient network error on any one row silently dropped it with no retry (the 2026-07-06 HotCRM first-boot incident). A `bulkCreate`-capable driver now sees roughly `ceil(N/batch)` writes, and a transient error is retried before a row is ever reported as failed.
+
+**Fix (`@objectstack/driver-sql`):** `SqlDriver.bulkCreate()` never generated a client-side id for a row missing one, unlike `create()` — a latent gap that this change is the first to exercise at scale (a bulk-inserted row without a driver-native id default silently landed with `id: NULL`). `bulkCreate()` now mirrors `create()`'s id/`_id` normalization per row.
--- a/.changeset/seed-import-bulk-insert.md
+++ b/.changeset/seed-import-bulk-insert.md
@@ -1,0 +1,14 @@
+---
+'@objectstack/core': minor
+'@objectstack/metadata-protocol': minor
+'@objectstack/rest': minor
+---
+
+Seed loader and data-import now route bulk writes through the engine's array-form `insert()` (one round-trip per batch, with parent-deduplicated summary recompute) instead of one `insert()`/`createData()` call per record, and both retry transient driver errors instead of silently dropping the row (#2678).
+
+A new shared helper, `bulkWrite` (`@objectstack/core`), batches rows through a caller-supplied batch-write function, retries a whole-batch transient failure (network blip / timeout) with exponential backoff, and degrades to per-row writes (each itself retried) when a batch fails for a non-transient reason — so one bad row can't drop the other N-1. `withTransientRetry` wraps a single write (e.g. an update) with the same retry behavior.
+
+- `SeedLoaderService.loadDataset()` (`@objectstack/metadata-protocol`) buffers insert-mode records and flushes them in batches of 200 via the engine's array `insert()`. Datasets with a self-referencing field (e.g. `employee.manager_id -> employee`) keep the historical per-record write path, since a later record may need an earlier one's freshly-assigned id.
+- `runImport()` (`@objectstack/rest`) buffers create-resolved rows and flushes them via `protocol.createManyData()` when the protocol supports it, falling back to the original per-row `createData()` call otherwise. `Protocol.createManyData` (`@objectstack/metadata-protocol`) now forwards `context` to `engine.insert()` like `createData` already did, so tenant-scoped bulk creates work correctly.
+
+Previously, a 1000-row seed or import into an object with a rollup summary issued 1000+ round-trips and up to 1000 summary recomputes; a single transient network error on any one row silently dropped it with no retry (the 2026-07-06 HotCRM first-boot incident). A `bulkCreate`-capable driver now sees roughly `ceil(N/batch)` writes, and a transient error is retried before a row is ever reported as failed.
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -26,6 +26,9 @@ export * from './utils/env.js';
 // Export timezone-aware calendar utilities (ADR-0053 Phase 2)
 export * from './utils/datetime.js';
 
+// Export the shared batched-write helper (framework#2678)
+export * from './utils/bulk-write.js';
+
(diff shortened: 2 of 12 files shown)


os-zhuang:
Investigated the CI failures from the first push:

**Real regression (fixed, pushed in 6e3f21d):** `Dogfood Regression Gate` / `Test Core` failed on `showcase-invoice-seed-isolation.dogfood.test.ts` — a seeded invoice line's `product` lookup resolved to `null`. Root cause: `SqlDriver.bulkCreate()` (`packages/plugins/driver-sql/src/sql-driver.ts`) never assigned a client-side id to a row missing one, unlike `create()` — so a bulk-inserted `showcase_product` row landed with `id: NULL`, and nothing downstream could resolve a reference to it. This bulk path was rarely exercised with real multi-row data before this PR; routing seed data through it at scale surfaced the gap. Fixed by mirroring `create()`'s id/`_id` normalization per row in `bulkCreate()`. Verified against a `main`-branch worktree that the test passes there today (i.e. this is not a pre-existing failure) and now passes on this branch too, along with the full `driver-sql` (248 tests) and `dogfood` (191 tests) suites.

**Unrelated, pre-existing (not fixed here):** CodeQL flagged 2 high-severity "polynomial regex" alerts in `packages/objectql/src/validation/record-validator.ts` and `rule-validator.ts` (`EMAIL_RE`). Confirmed via `git diff main` that neither file is touched by this PR's diff — these are latent findings in unrelated, unmodified code. Filed separately as #2681 rather than expanding this PR's scope.

Continuing to watch CI.

---

PR #3122, opened 2026-07-17
docs(approvals): drop duplicate role-vs-position callout (unbreak role-word ratchet)

os-zhuang:
## Why main is red

The `ESLint` job of **Lint & Type Check** fails on `main` at its *Reserved-word ("role") docs ratchet* step, so **every open PR inherits a red check**. `pnpm lint` itself is clean — this is not an eslint rule failure.

```
• content/docs/automation/approvals.mdx: role-word count grew 5 → 9.
  New occurrences are banned (ADR-0090 D3).
```

Introduced by 25a19be72 (#3113). Red at 021ba4caf and every commit after.

## The occurrences are not drift — and must not be reworded

The obvious fix (swap `role` → the D3 vocabulary) is **wrong here**, so this PR does not do it:

- `role` is a **real, shipping approver type** — [`approval.zod.ts:14`](packages/spec/src/automation/approval.zod.ts#L14), sitting in the enum next to `position` with a comment explaining exactly why it stays.
- It resolves against **better-auth's org-membership tier** (`sys_member.role`: owner/admin/member) — which ADR-0090 D3 names as its **single documented exception**: *"the better-auth boundary — `sys_member.role` is third-party schema we do not own; it remains."*
- Writing `position` in these spots would make the docs **factually wrong**. Distinguishing the two *is the entire point of the passage* — and of the [`approval-role-not-membership-tier`](packages/lint/src/validate-approval-approvers.ts) lint rule, which exists precisely because authors confuse them and their approvals silently route to nobody.

## What the ratchet did catch

Real redundancy. The callout added under *"The request opens, the run pauses"* restates the callout ~75 lines above it — same fact, same page, and the **older one says more** (it names the lint rule and `sys_user_position`). **Removed the duplicate** (−3 occurrences). That's a docs improvement on its own merits.

That left `role:<r>` in the `approverId` filter list (−1 short): a genuine *new* fact at the same sanctioned boundary ([`approval-service.ts:231`](packages/spec/src/contracts/approval-service.ts#L231)). Rather than delete a true API detail to satisfy a counter, the **surviving callout is tightened** — it stated role-the-approver-type twice (*"The `role` approver type is…"* and *"authored as `type: 'role'`"*), now once. That frees the slot and the file lands back at **exactly 5**.

## Baseline deliberately NOT bumped

A high-water mark that only ratchets down is the point of the gate. `scripts/role-word-baseline.json` is untouched — the only changed file is the `.mdx`.

## Reviewer's judgment call

[...]

PR #3132, opened 2026-07-17
docs(approvals): drop reserved word "role" from lifecycle copy (ADR-0090 D3)

os-zhuang:
Fixes the red **Lint & Type Check** on `main`. Docs-only — hence `skip-changeset`.

## What broke

[#3113](https://github.com/objectstack-ai/framework/pull/3113) (`25a19be72`) added the approval-lifecycle walkthrough, including **4 new occurrences of the reserved word "role"** in `content/docs/automation/approvals.mdx` (5 → 9), without updating `scripts/role-word-baseline.json`. `pnpm check:role-word` has failed on **every push to main since** — the last 4 runs of `lint.yml` are red — so every open PR inherits a failing ESLint check and real failures are masked.

## Why not bump the baseline

Per **ADR-0090 D3** the word is *reserved-forbidden* and new occurrences are **banned**; `--update` exists to ratchet the baseline **down** when a file improves. So the baseline stays at **5** and the copy is reworded.

## The fix

**Three of the four were a duplicate.** The new callout under *The request opens, the run pauses* restated the already-baselined `position` vs `role` callout ~80 lines above it on the same page — which says the same thing in more detail (it names the `os lint` rule). Removed in favour of a cross-reference.

While rewording I also tightened the claim to match [`expandApprovers()`](https://github.com/objectstack-ai/framework/blob/main/packages/plugins/plugin-approvals/src/approval-service.ts): an unresolvable approver entry does **not** leave `pending_approvers` empty — it falls back to a literal `<type>:<value>` slot that no user matches, which is *why* the request stalls.

**The fourth** documented `approverId` accepting `role:<r>`. That literal is real, but it is one of several (`team:` / `department:` / `position:` / `role:` / `manager:`), so naming only the `role:` form was both off-vocabulary and needlessly narrow. Generalized to `<type>:<value>` with a `position:finance_manager` example, matching the page’s existing `finance_manager` sample.

**The surviving 5 are the documented D3 exception** — the better-auth boundary (`sys_member.role`), which the `role` approver type genuinely resolves against via `expandRoleUsers()`. Left untouched.

## Verification

- `pnpm check:role-word` → exit **0**, approvals.mdx back to **5**, `role-word-baseline.json` **unmodified**.
- `pnpm lint`, `check:doc-authoring`, `check:authz-resolver`, `check:release-notes` → all pass.
  <sub>(Local `eslint` OOMs at the default heap on node 25 — an environment artifact, not this change; it is clean under `--max-old-space-size=8192`, and CI runs node 20.)</sub>
- **Rendered the page in the docs site**: the `<Steps>` block keeps all 4 step headings after the callout removal, the reworded `approverId` sentence renders (`<type>:<value>` survives MDX as inline code), and the new `#3-the-approval-node` cross-reference resolves to the right heading — no console or server errors. Worth checking by hand since no CI job validates anchors.

### DIFF
--- a/content/docs/automation/approvals.mdx
+++ b/content/docs/automation/approvals.mdx
@@ -130,14 +130,10 @@ record is **locked** against edits while pending (`lockRecord`, default `true`),
 and the flow run parks until a decision arrives.
 
 Only `approvers` is required on the node; everything else has a default
-(`behavior: 'first_response'`, `lockRecord: true`, `maxRevisions: 3`).
-
-<Callout type="warn">
-**`type: 'role'` is not a position.** In an approver entry, `role` means the
-better-auth membership tier (`owner` / `admin` / `member`). Naming a business
-role like `sales_manager` there matches nobody and the request silently has no
-approvers — use `{ type: 'position', value: 'sales_manager' }`.
-</Callout>
+(`behavior: 'first_response'`, `lockRecord: true`, `maxRevisions: 3`). An
+approver entry that resolves to no users falls back to a literal `<type>:<value>`
+slot that no one matches, so the request opens and then stalls — see the
+approver-type warning under [The approval node](#3-the-approval-node).
 
 ### The approver finds it in their queue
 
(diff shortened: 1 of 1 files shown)


os-zhuang:
Closing as a duplicate of #3122, which fixes the same file for the same reason, was opened ~10 min earlier, and is already fully green.

I missed #3122 before opening this — my mistake. #3122 is also the **better** fix on the 4th occurrence:

- Both PRs delete the duplicate `type: 'role' is not a position` callout (−3), which restates the callout ~75 lines above it.
- For the 4th, **I rewrote `approverId` accepting `role:<r>` → `position:finance_manager`. That was wrong.** Per [`contracts/approval-service.ts`](https://github.com/objectstack-ai/framework/blob/main/packages/spec/src/contracts/approval-service.ts) the identities are literally *"their user id, email, and `role:<r>` entries"* — `role:<r>` is one of **a person's own identities** (their better-auth membership tier) and a genuine ADR-0090 D3 exception. A `position:` literal is not an identity at all; it only appears as a fallback when position expansion matches nobody. So my version contradicted the sentence's own "cover a person's identities in one call" clause.
- #3122 instead trims the surviving callout's duplicated mention of the role approver type (−1) — removing real redundancy without laundering a real boundary literal.

No action needed here; #3122 covers it.

PR #3138, opened 2026-07-17
fix(spec): regenerate drifted skill references + react-blocks contract, and gate them in CI

os-zhuang:
Two committed generated surfaces under `skills/` were drifted from `packages/spec` and completely unguarded by CI. Same class as #3134, but on surfaces that ship to third parties via `npx skills add objectstack-ai/framework` — so the drift is served straight to consumers' agents.

## The drift was real, and user-facing

An independent oracle (parses every pointer out of the committed indexes, asserts the file exists — deliberately *not* the generator, so it isn't circular):

| | pointers | broken |
|---|---|---|
| main @ `ce468c832` | 113 | **6** |
| this PR | 126 | **0** |

The six dangling pointers handed agents paths to files that don't exist:

- `automation/workflow.zod.ts` — deleted by #1398 ("remove Workflow Rules"), index never regenerated
- `system/masking.zod.ts` — deleted from spec
- `data/dataset.zod.ts` — renamed to `data/seed.zod.ts` by #1620

Every change in the regenerated diff traces to a spec change; I checked each one. The `encryption.zod.ts` drops are because `field.zod.ts` no longer imports it; the `metadata-protection` / `protection` / `visibility` / `public-auth-features` additions are new spec imports (ADR-0010, ADR-0089, #2874); the `flow.zod.ts` and `manifest.zod.ts` description changes are edited JSDoc. Output is byte-identical across runs.

## Regenerating alone would have shipped a regression

`SKILL_MAP` still named `data/dataset.zod.ts`, and the generator **warned and skipped** the missing file. A plain regen would therefore have quietly dropped seed coverage from `objectstack-data` — whose own SKILL.md advertises `defineSeed()` authoring — and the new gate would have locked that in as "correct".

So: point the map at `data/seed.zod.ts`, and make a dangling entry a **hard failure in both modes**. The map is authored config; a dangling entry is always a bug in it, never a spec change to absorb. That warn-and-continue is the same false-green vector the gate exists to close.

## Gate mechanics

Follows #3134's `build-docs.ts`: every write goes through `emit()`, every wholesale-regenerated folder through `manageDir()`, nothing touches disk until `flush()` — so write and `--check` share **all** generation logic and differ only in disposition. Extracted to `scripts/lib/generated-output.ts` rather than copy-pasted a third time.

`manageDir()` takes an ownership predicate, because a skill's `references/` folder is only *partially* generated: `_index.md` is ours, but `react-blocks.md` (sibling generator) and hand-written notes must survive. Verified they do.

New scripts: `check:skill-refs`, `check:react-blocks`.

## Proven red before being trusted (13/13)

A gate that cannot fail is worse than no gate. Each was reproduced going red, then green again after restore:

| case | result |
|---|---|
| stale content | ✅ exit 1 `~ out of date` |
| missing file | ✅ exit 1 `+ missing` |
| **stale untracked leftover** (subfolder + loose `.zod.ts`) | ✅ exit 1 `- stale` |
| ↳ same state under `git diff --exit-code` | **exit 0 — blind.** This is why `manageDir` exists |
| dangling `SKILL_MAP` entry (the bug that shipped) | ✅ exit 1, in check **and** write mode |
| skill dir missing from disk | ✅ exit 1 |
| spec change the contract hadn't absorbed | ✅ exit 1 `~ out of date` |
| react-blocks: schemas fail to load (swallowed `z.toJSONSchema` error) | ✅ exit 1 — and write mode refused to clobber the good contract |

That last guard is why it checks *data-kind* props rather than `props.length`: `mergeProps` puts the hand-authored overlay first, so a block whose schema failed to load still reports props. The data props are what vanish.

## Placement: lint.yml, deliberately

In lint.yml's **"TypeScript Type Check"** (no paths filter + required), not ci.yml's `check-generated`. That job is filter-gated on:

```
skilldocs:
  - 'skills/**'
  - 'content/docs/guides/skills.mdx'
  - 'packages/spec/scripts/build-skill-docs.ts'
```

— **not** `packages/spec/src/**`. It would skip exactly the spec-only PRs that cause this drift (the #3076 failure mode), and it isn't a required check, so it could be merged past even when it does run. Required checks confirmed as `["TypeScript Type Check","Build Core","Test Core","Dogfood Regression Gate"]`.

Both gates need **no `^build`** — verified empirically by removing all 23 workspace `dist/` dirs and re-running (turbo's `^build` on `gen:skill-refs` is vestigial; `pnpm --filter` bypasses turbo anyway). So they run *before* the workspace build and fail in seconds.

## Note on #3134

#3134 is open and touches `lint.yml` + `packages/spec/package.json` too, inserting `check:docs` in the same step gap. Expect a trivial adjacent-block conflict; both blocks coexist. I based on `main` rather than on that branch so this stands alone. Follow-up once it lands: migrate `build-docs.ts` onto `scripts/lib/generated-output.ts`.

`check:react-conformance` was checked for overlap — there is none. It compares spec schemas against the objectui frontend manifest (opt-in via `MANIFEST=`, exits 0 without it) and never checks generator output against committed files.

## Verification

- `check:skill-refs` / `check:react-blocks` green on a clean tree; 13/13 red-proofs pass
- `tsc --noEmit` clean · spec tests **254 files / 6890 passed**
- oracle: 126 pointers, 0 broken · generators deterministic across 3 runs


os-zhuang:
Merged `main` in (no force-push, per AGENTS.md §3). #3134 landed while this was in review, and as predicted it conflicted in `lint.yml` — a trivial adjacent-block one. Resolved by keeping **both** gates; they're the same class on different surfaces:

```
- Type check (@objectstack/spec)
- Check generated reference docs are in sync with the spec    # check:docs      (#3134)
- Check generated skill references are in sync with the spec  # check:skill-refs (this PR)
- Check the react-blocks contract is in sync with the spec    # check:react-blocks (this PR)
- Build workspace packages
```

I also re-checked my placement argument against the *post-#3134* filter rather than the one I originally read, and updated the comment to match. #3134 renamed the job to `check-generated` and added spec paths to the filter — but only `migrations/**`, `conversions/**`, and `protocol-version.ts`. No schema dirs. So a PR touching `src/data/**` or `src/ui/**` — exactly what drives these two artifacts — still never triggers that job, and it's still not required. lint.yml remains the correct home.

Re-verified against the merged tree:

- `check:skill-refs` ✅ · `check:react-blocks` ✅ · `check:docs` ✅ (confirmed I don't break #3134's gate)
- oracle: 126 pointers, 0 broken
- all 4 required checks were green on `02624629e` before the merge

Follow-up from the merge, worth doing now that #3134 is in: `build-docs.ts` carries its own inline copy of `emit()`/`manageDir()`, while this PR extracted the same logic to `scripts/lib/generated-output.ts`. Migrating it onto the shared helper would leave one implementation of the convention instead of two.

Issue #3139, opened 2026-07-17
build-skill-references.ts: import regex is blind to multi-line imports, under-reporting skill schema deps


os-zhuang:
Found while fixing generated-skill drift in #3138 — out of scope there (it is pre-existing and identical before/after regeneration, so it does not affect that diff), filed per Prime Directive #10.

## Problem

`packages/spec/scripts/build-skill-references.ts` resolves the transitive schema closure with a **single-line** regex:

```ts
const re = /^import\s+.*\s+from\s+['"](\.[^'"]+)['"]/gm;
```

`.` does not match newlines, so a multi-line import is never seen:

```ts
import {
  FooSchema,
} from './foo.zod';        // ← invisible to the resolver
```

The dep is silently omitted from the skill's "Transitive dependencies" section. No warning — the file simply never enters the queue.

## Impact

18 non-test multi-line local imports in `packages/spec/src` are currently invisible, e.g.:

```
api/feed-api.zod.ts       -> ../data/feed.zod
api/feed-api.zod.ts       -> ../data/subscription.zod
api/protocol.zod.ts       -> ./analytics.zod
api/protocol.zod.ts       -> ../kernel/package-registry.zod
automation/bpmn-mapping.ts -> ./control-flow.zod
integration/connector/*.zod.ts -> ../connector.zod   (×4)
```

(217 including `.test.ts`, which are irrelevant here.)

The effect is under-reporting, not a broken pointer: `skills/*/references/_index.md` ships to third parties via `npx skills add`, and an agent following the index gets an incomplete picture of the schema graph. Whether any of the 18 is reachable from a `SKILL_MAP` entry's closure is worth checking — that determines real-world impact.

## Second, related blind spot

`extractLocalImports` appends `.ts` only when the specifier does not already end in `.ts`:

```ts
if (!resolved.endsWith('.ts')) resolved += '.ts';
```

So a `.js`-suffixed ESM specifier resolves to `foo.js.ts`, never exists, and is dropped — e.g. `contracts/data-engine.ts -> ../data/index.js` and `conversions/apply.ts -> ./types.js`.

## Suggested fix

Replace the regex with a real import scan (TS AST, or at minimum a multi-line-tolerant pattern plus `.js` → `.ts` specifier normalization). Note this will legitimately grow the "Transitive dependencies" section of several indexes — a gate now guards those files (#3138), so the regen must land with the fix.

## Why this is safe to do now

[...]

PR #3143, opened 2026-07-17
fix(spec): regenerate drifted skills output and gate it in CI

os-zhuang:
`skills/` is generated from the spec by `gen:skill-refs` and `gen:react-blocks` and committed, but **neither generator had a `check:` sibling and nothing in CI regenerated them**. Both existing gates miss this surface:

- `check:docs` (#3134) covers only `content/docs/references/`
- `check:skill-docs` covers only SKILL.md frontmatter → README + `skills-reference.mdx`

So it drifted: **8 committed files were stale on main** (verified on `c22e2c3fc`).

This bites harder than the content/docs case — `skills/` is **published to third parties** via `npx skills add objectstack-ai/framework --all`, so stale output ships to users.

## 1. A real bug, not just drift

Two published skills pointed agents at `data/dataset.zod.ts` — renamed to `data/seed.zod.ts` by #1620 ("complete Dataset->Seed rename"). The generator **warned and skipped, staying green**, so `objectstack-data` and `objectstack-platform` silently lost their seed schema pointer and told agents to `Read` a path that does not exist.

Fixed at the source (`SKILL_MAP`), not in the output. A missing mapped file is now a **hard error** rather than a warning — warn-and-skip is exactly what let this half-rename hide.

The rest of the diff is legitimate spec drift: a `drawerWidth` deprecation, a new `relationshipValueField` prop, `columns` becoming optional.

## 2. `--check` on both generators

Follows the convention #3134 established in `build-docs.ts`: every write goes through `emit()`, every regenerated folder through `manageDir()`, and nothing touches the tree until `flush()` — so check and write run **byte-for-byte identical generation logic** and differ only in final disposition. Not `git diff --exit-code`, which only sees tracked files and would miss a new or leftover page.

The sink is extracted to `scripts/lib/generated-output.ts` since both generators need it. `manageDir()` takes a predicate because a skill's `references/` folder is a **selective** wipe, not a wholesale one — hand-written siblings (`data-hooks.md`, `plugin-hooks.md`) and the *other* generator's `react-blocks.md` live there and must survive. `build-docs.ts` keeps its own inline copy for now; it predates the extraction and is under active change (#3126), so migrating it belongs in its own PR.

## 3. Wired into "TypeScript Type Check", not ESLint

`main` requires exactly 4 status checks (TypeScript Type Check, Build Core, Test Core, Dogfood Regression Gate). **ESLint is not one of them**, so a gate placed there runs but cannot block a merge. (Re-verified: the one ruleset is `enforcement: disabled` and adds no checks.) Same job as `check:docs`, same reason: no paths filter, required, and both read `src/` via tsx with no build — they fail in ~2s.

### Why not the existing "Check Generated Artifacts" job?

That job in `ci.yml` looks like the natural home — it already runs `check:skill-docs`. But it is the wrong place twice over, and its own filter comments document this class of bug biting twice before:

- **It is paths-filtered.** The `generated` filter watches `skills/**` and `build-skill-docs.ts`, but **not `packages/spec/src/**`**. Every file in this PR's drift traces to a `packages/spec/src/**` change (the `seed.zod.ts` rename, `view.zod.ts`'s `drawerWidth` deprecation, the new `relationshipValueField`). A spec-only PR changes the generated output without touching `skills/`, so the filter would not match and **the job would skip — which is exactly how this drift shipped.**
- **It is not a required check.** Only the 4 checks above are, so even when it does run it cannot block a merge.

`check:docs` landed in the typecheck job for precisely this reason (#3134), and its comment says so. These gates follow it.

## Proving the gate can go red

A gate that cannot fail is theatre, so each failure mode was exercised — all 14 as expected:

| case | expect | got |
|---|---|---|
| clean tree (both gates) | 0 | ✅ |
| tampered content | 1 | ✅ |
| deleted generated file | 1 | ✅ |
| leftover file in `references/` | 1 | ✅ |
| leftover file in a sub-folder | 1 | ✅ |
| `SKILL_MAP` → missing skill dir | 1 | ✅ |
| `SKILL_MAP` → renamed/missing zod file | 1 | ✅ |
| **vacuous:** empty `SKILL_MAP` | 1 | ✅ |
| **vacuous:** empty `REACT_BLOCKS` | 1 | ✅ |
| tampered contract json | 1 | ✅ |
| deleted `react-blocks.md` | 1 | ✅ |
| leftover file in `contracts/` | 1 | ✅ |
| **negative control:** hand-written `.md` in `references/` | **0** | ✅ |

The vacuous guards matter because a run that finds no input emits nothing, and "nothing differs" would read as success — the gate would pass while checking nothing at all. Both now fail loudly instead of greenly. The negative control proves the selective-wipe predicate is not over-broad.

Also verified: the original repro (`gen:skill-refs && gen:react-blocks && git status skills/`) now produces zero drift; `check:docs`, `check:role-word` (bidirectional ratchet, scopes `skills/`), `check:doc-authoring`, and `check:nul-bytes` all stay green.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -144,6 +144,26 @@ jobs:
       - name: Check generated reference docs are in sync with the spec
         run: pnpm --filter @objectstack/spec check:docs
 
+      # Generated-skills gate: skills/ is generated from the spec by
+      # `gen:skill-refs` + `gen:react-blocks` and committed, and nothing checked
+      # it — both had a `gen:` script but no `check:` sibling, so `check:docs`
+      # above (which only covers content/docs/references/) and `check:skill-docs`
+      # (only SKILL.md frontmatter → README + skills-reference.mdx) both miss it.
+      # It drifted: 8 files were stale on main, and skills/objectstack-{data,
+      # platform}/references/_index.md still pointed agents at
+      # data/dataset.zod.ts — renamed to data/seed.zod.ts back in #1620.
+      #
+      # This bites harder than the docs case: skills/ is PUBLISHED to third
+      # parties via `npx skills add objectstack-ai/framework --all`, so stale
+      # output ships to users. Same job as check:docs and for the same reason —
+      # it is a required status check with no paths filter, so the gate cannot go
+      # dormant. Both read src/ via tsx and need no build, so they fail in ~2s.
+      - name: Check generated skill references are in sync with the spec
+        run: pnpm --filter @objectstack/spec check:skill-refs
+
+      - name: Check generated react-blocks contract is in sync with the spec
+        run: pnpm --filter @objectstack/spec check:react-blocks
+
       # Example apps are AI-authoring reference templates; a red typecheck is a
       # bad signal to copy from. tsup transpiles them without a full typecheck,
       # so build alone will not catch type drift — typecheck them explicitly.
--- a/packages/spec/package.json
+++ b/packages/spec/package.json
@@ -190,6 +190,7 @@
     "gen:docs": "tsx scripts/build-docs.ts",
     "check:docs": "pnpm gen:schema && tsx scripts/build-docs.ts --check",
     "gen:skill-refs": "tsx scripts/build-skill-references.ts",
+    "check:skill-refs": "tsx scripts/build-skill-references.ts --check",
     "gen:skill-docs": "tsx scripts/build-skill-docs.ts",
     "check:skill-docs": "tsx scripts/build-skill-docs.ts --check",
     "analyze": "tsx scripts/analyze-bundle-size.ts",
@@ -205,6 +206,7 @@
     "test:coverage": "vitest run --coverage",
     "check:liveness": "tsx scripts/liveness/check-liveness.mts",
     "gen:react-blocks": "tsx scripts/build-react-blocks-contract.ts",
+    "check:react-blocks": "tsx scripts/build-react-blocks-contract.ts --check",
     "check:react-conformance": "tsx scripts/check-react-blocks-conformance.ts"
   },
   "keywords": [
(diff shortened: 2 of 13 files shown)


os-zhuang:
Heads-up: we collided. #3138 was landed on `main` as `6cc73e33c` a few minutes ago covering the same ground — two agents picked up this task in parallel. Flagging rather than closing, since this is your branch to call.

The generated output is fully redundant now. I diffed all 8 of this PR's `skills/` files against `main`:

```
identical: 8   differing: 0
```

Byte-identical — including the same `SKILL_MAP` fix (`data/dataset.zod.ts` → `data/seed.zod.ts` in both objectstack-data and objectstack-platform). Reassuring rather than annoying: two independent runs converging byte-for-byte is decent evidence the regeneration is right and the generator is deterministic.

We also independently reached the same three conclusions, which I'd take as corroboration of the design:

- `manageDir()` needs an ownership **predicate** — `references/` is only partially generated (`_index.md` is the generator's; `react-blocks.md` from the sibling generator and hand-written notes must survive a wipe).
- The generator's **warn-and-skip on a missing SKILL_MAP entry** is the actual false-green vector — `--check` alone would have locked the dropped seed pointer in as correct. It has to be fatal.
- ci.yml's `check-generated` is the plausible-looking wrong home: filter-gated without the schema dirs, and not required. lint.yml's `TypeScript Type Check` is the one with teeth.

What's on `main` now: `check:skill-refs` + `check:react-blocks`, sharing `scripts/lib/generated-output.ts`, wired into `TypeScript Type Check` after #3134's `check:docs`. Each was proven red before being trusted (13 cases — stale content, missing file, stale *untracked* leftover, dangling SKILL_MAP entry, missing skill dir, unabsorbed spec change), including the check that `git diff --exit-code` returns 0 on the stale-leftover case, which is why `manageDir` exists at all.

If anything in your implementation is better than what landed, it's worth salvaging as a follow-up on top of `main` — otherwise this can just close. One genuine leftover either way: `build-docs.ts` still carries its own inline `emit()`/`manageDir()` copy and should migrate onto the shared helper. Also filed #3139 for the resolver's multi-line-import blind spot.

Issue #2698, opened 2026-07-09
feat(mcp): OAuth 2.1 authorization for /api/v1/mcp (self-serve connect from claude.ai / Claude Code / Desktop)


os-zhuang:
## Context

The open edition's AI surface is MCP-only (BYO-AI via `@objectstack/mcp`, see `content/docs/ai/agents.mdx`). Today the HTTP MCP endpoint (`/api/v1/mcp`) authenticates **only** with API keys (`x-api-key` / `Authorization: ApiKey`, hashed at rest, scoped, fail-closed — `packages/core/src/security/api-key.ts`). That works for headless agents and developers, but it blocks the mainstream human-connected clients:

- **claude.ai custom connectors** and the Anthropic connector directory require the MCP-spec authorization flow (OAuth 2.1).
- **Claude Desktop / Claude Code** offer a much smoother connect experience over OAuth (browser login as yourself) than hand-copying header config.
- Every user currently needs an admin-minted API key before they can connect anything.

The MCP authorization spec is deliberately **decentralized**: the client discovers the authorization server via `.well-known` metadata on the MCP host itself, self-registers via Dynamic Client Registration (RFC 7591), and runs an authorization-code + PKCE flow. This means **self-hosted / private deployments work out of the box** — each deployment is its own authorization server; nothing is pre-registered with Anthropic or with any central ObjectStack service.

## Goal

Add spec-compliant OAuth 2.1 authorization to the MCP endpoint so any MCP client can connect self-serve, while keeping API keys as the headless track.

## Design decisions (rationale lives here + changeset; no separate ADR per repo convention)

1. **Each deployment is its own authorization server.** Back the flow with the embedded better-auth instance (it already owns login/sessions). No central authority, no dependency on cloud — the community/self-hosted edition is the primary target, not an afterthought.
2. **Discovery endpoints**: serve `.well-known/oauth-protected-resource` (resource metadata pointing at the AS) and `.well-known/oauth-authorization-server` (AS metadata) from the same origin as `/api/v1/mcp`.
3. **Dynamic Client Registration is mandatory.** Since every deployment is a distinct AS, clients cannot ship pre-registered client IDs. DCR is the enabler for "one generic client config works against any deployment".
4. **PKCE required; TLS required** (localhost exempt per OAuth 2.1). Intranet deployments with an internal CA work for local clients (Claude Code/Desktop); claude.ai web connectors additionally need the endpoint publicly reachable — a network decision, out of our hands, but must be documented.
[...]

PR #3182, opened 2026-07-18
docs(adr): ADR-0100 — serve-side MCP default wiring; register the MCP tool-execution surface in the authz matrix (#3167)

### LINKED ISSUE #3167: decision: default wiring for the serve-side MCP server (@objectstack/mcp) in dev — identity admission first
**Decision issue** — direction first, implementation split later. Counterpart of the consume-side work that closed #2977/#3056: those made ObjectStack *call* external MCP servers; this is about ObjectStack *being* one.

## Current state (verified against main)

An odd half-wired shape:

- The runtime dispatcher already serves an MCP **HTTP surface per-request at `/api/v1/mcp`, default-ON** (`OS_MCP_SERVER_ENABLED=false` opts out), plus the public `GET /api/v1/mcp/skill` SKILL.md download and a connect UI (`packages/mcp/src/connect-ui.ts`).
- But the service behind it — **`MCPServerPlugin`** (registers the `'mcp'` service; bridges ToolRegistry, MetadataService, DataEngine, and Agents; optionally starts a long-lived **stdio** transport via `autoStart` / `OS_MCP_SERVER_ENABLED=true`) — **is wired nowhere by default**: not in `os dev`/`serve`, not in the showcase, not in the scaffold. Its only in-repo consumer is an `app-todo` e2e.

Net effect: an AI-native platform whose official "let AI operate the running app" channel is a route that answers but a capability that is never registered. Discoverability is zero exactly where the platform's pitch is strongest — the dev loop, where a coding agent (Claude Code, Cursor, any MCP client) could read the schema, inspect seed data, and exercise actions of the app it is building.

## Proposed direction (to be ratified or amended)

**Dev default-on, prod explicit.** `os dev` registers `MCPServerPlugin` by default (escape hatch: `OS_MCP_SERVER_ENABLED=false`); `os serve`/production keeps today's explicit opt-in. The value asymmetry is stark: in dev the MCP endpoint completes the AI-builds-apps loop; in prod it is a high-authority execution surface that must be a deliberate choice.

### 1. Identity admission — hard precondition (ADR-0096)

MCP **tool execution is an execution surface**, the same class as GraphQL and realtime, which ADR-0096 registered in the authz conformance matrix (`packages/qa/dogfood/test/authz-conformance.matrix.ts`) with CI tripwires. Today the matrix covers MCP only incidentally (the data-write ingress readonly strip names the MCP dispatcher as one funneled caller); **there is no row for the MCP tool-execution principal**: when an MCP client calls `tools/call`, *whose* `ExecutionContext` runs it? Dev-admin? A token-resolved user? An explicit guest? Fall-open on a missing principal = full authority — the exact failure mode ADR-0096 exists to prevent. Registering the row (and the answer) is the gate for any default-on wiring; nothing else in this issue proceeds before it.

### 2. The dev/prod boundary definition

What *is* "dev" for this purpose — the `os dev` command itself (recommended: wiring follows the command, not the env), `NODE_ENV`, or a config flag? And the off-switch semantics (`OS_MCP_SERVER_ENABLED=false` today disables the HTTP surface — should it also gate the plugin registration?).

### 3. Transport & mount shape

The per-request HTTP surface already exists; the open questions are (a) whether dev auto-start should also cover the **stdio** transport (local AI tools spawning the process directly — likely no: `os dev` owns the process), and (b) whether the connect UI / skill download need dev-specific affordances (e.g. printing the MCP URL + connect instructions on dev boot, which the dispatcher's skill/connect-ui pieces suggest was the intent).

### 4. Default exposure scope

Auto-generating tools for *every* object's CRUD explodes the tool list on real apps (hundreds of tools makes MCP clients unusable) and widens the misuse surface. Decide the default: e.g. only `apiEnabled` objects + explicitly marked actions/flows, with a metadata-declared exposure config as the future story (consistent with the ADR-0097 metadata-first philosophy — what is exposed should itself be authorable).

## Optional extension (after the above)

With a stable in-repo serve-side endpoint, the showcase's declarative `provider: 'mcp'` connector gains a second demo target: **the platform connecting to itself**. Previously rejected for boot-ordering (automation `start()` runs before HTTP listens); the #3049 degrade+retry makes the self-connection heal seconds after boot, but the dogfood gate must not become timing-sensitive — treat as a separate, carefully-gated follow-up.

## Refs
- ADR-0096 (execution-surface identity admission — the matrix + tripwire mechanism this must register with)
- ADR-0097 / #2977 / #3056 (consume side, closed), ADR-0024 §4 (trust model)
- `packages/mcp` (`MCPServerPlugin`, `MCPServerRuntime`, connect-ui, skill), `packages/runtime/src/http-dispatcher.ts` (`/mcp` routes)

os-zhuang:
Decision-issue response to #3167 (*default wiring for the serve-side MCP server — identity admission first*). #3167 makes decision #1 (identity admission) a hard gate: "nothing else in this issue proceeds before it." This PR lands the decision record **and** that gate, with an end-to-end proof.

## Two findings that shaped the response

1. **The issue's premise is stale against `main`.** #3167 says `MCPServerPlugin` is "wired nowhere by default." Verified against `main`, MCP is already a **default-on core capability**: `isMcpServerEnabled()` defaults ON (`packages/types/src/env.ts`), `packages/cli/src/commands/serve.ts` pushes the `mcp` capability into `os serve`/`os dev` unless opted out, and the dispatcher route gate reads the same helper. The HTTP tool surface is also already **fail-closed on identity** — `handleMcp` returns 401 for a principal-less caller (no dev-admin fallback, no fall-open), and `buildMcpBridge` threads the caller `ExecutionContext` into `callData(..., ec)` on every verb (OAuth callers run at the ADR-0090 D10 ceiling∩user intersection).

2. **What was genuinely missing is exactly the gate.** The MCP tool-execution surface had **no row** in the ADR-0096 D4 authz conformance matrix — so "whose `ExecutionContext` runs a `tools/call`" lived only in ADR-0096 prose, not as a checked, red-CI-on-regression property.

## What this lands

- **`docs/adr/0100-serve-side-mcp-default-wiring-identity-admission.md`** — the four decisions grounded in the verified state:
  - **D1** identity admission — caller-principal (`user`) posture, fail-closed, registered as the gate.
  - **D2** the dev/prod boundary is **identity + the off-switch** (`OS_MCP_SERVER_ENABLED=false` gates registration *and* route), not a code-level default flip (an opt-in-for-prod amendment is documented for ratification).
  - **D3** HTTP-only default; **stdio stays explicit opt-in** (`os dev` owns the process; the stdio resource path carries the ADR-0096 E1 context-less read).
  - **D4** bounded default exposure — fixed generic tool set over `apiEnabled` objects, `sys_*` fail-closed, actions only when `ai.exposed`; metadata-declared exposure is the future story.
- **`authz-conformance.matrix.ts`** — two rows: enforced `mcp-tool-exec-identity` (fail-closed 401 gate + `buildMcpBridge` threading + OAuth scope gate; now carrying a real dogfood `proof` and marked high-risk) and experimental `mcp-stdio-resource-read` classifying the ADR-0096 **E1** context-less stdio resource read (opt-in, never networked) with its admission requirement — registered honestly, **not** patched with a discouraged `{ isSystem: true }` literal (ADR-0096 D2 retires those; the `systemContext(reason)` constructor doesn't exist yet).
- **`authz-conformance.test.ts`** — three source probes pinning both halves of the HTTP admission and the stdio fall-open, so dropping the gate/threading or networking the stdio read trips CI (STALE / UNCLASSIFIED) instead of surfacing in an adversarial review later.
- **`showcase-mcp-identity.dogfood.test.ts`** — the end-to-end proof: boots the showcase with `MCPServerPlugin` registered and asserts on the real HTTP stack that anonymous `POST /mcp`→**401** (not 404/501 — the surface is served and the service is registered; the denial is identity), an authenticated member **clears** the gate, and `GET /mcp/skill`→**200** (served default-on while tool-exec is gated).

## Scope / safety

Docs + a **private** test package (`@objectstack/dogfood`) only — **no runtime behavior change**, no changeset. The identity path, OAuth admission (#2698), and `ai.exposed` + capability action gate are pre-existing and unchanged; this PR *registers and proves* them. (`@objectstack/mcp` added to the dogfood dev deps; lockfile updated.)

## Validation

Validated locally: `pnpm --filter @objectstack/dogfood test` for `authz-conformance` + `showcase-mcp-identity` → **10/10 pass** (7 conformance ratchet + 3 dogfood-boot). The three new probes each match their source anchor exactly once and are a clean bijection with the two rows' `covers`; the `proof:`-exists and high-risk assertions pass.

## Open for ratification
- **D2**: keep default-on-everywhere behind the off-switch (recommended), or adopt the `NODE_ENV`-aware opt-in-for-prod amendment.

Closes #3167 pending ratification of D2.


os-zhuang:
Superseded by #3202, which already registered the MCP execution-surface identity admission in the ADR-0096 matrix (`mcp-http-identity` + `mcp-stdio-authority` rows + probes) for #3167. Closing as redundant — one of the probe keys is identical to #3202's, so this branch would double-classify the ledger. No separate decision ADR needed.

---

PR #3363, opened 2026-07-20
fix(spec): enforce ViewFilterRuleSchema operator enum + typed sort

akarma-synetal:
## Summary
- `operator`: `z.string()` → `z.enum([19 valid operators])` so SchemaForm auto-renders a Select dropdown
- `sort`: `z.union([z.string(), z.array(...)])` → `z.array(...)` so RepeaterField renders sort sub-fields correctly
- Updated fixtures and share-link object to use valid operators
- All 6807 spec tests pass

## Test plan
- [x] `pnpm --filter @objectstack/spec test` — 6807 passed, 0 failed
- [x] Verify Edit View Config in ObjectUI console shows operator dropdown + smart value input (paired with objectui PR)

Made with [Cursor](https://cursor.com)


os-zhuang:
Thanks for tackling this — the underlying need (a real operator enum so SchemaForm can render a dropdown and invalid operators are rejected) was sound, and it shipped in a reworked form via #3373 (now merged to `main`), tracked by #3368.

The rework kept your enum direction but adjusted a few things that made the original approach risky:

- **Legacy-alias normalization instead of a bare strict enum.** `operator` is now `z.preprocess(normalizeFilterOperator, z.enum(VIEW_FILTER_OPERATORS))`. Verified that under `z.toJSONSchema({ io: 'input' })` (the mode ObjectUI uses) this still surfaces the clean canonical enum for the dropdown, while folding stored shorthand/camelCase spellings (`gt`, `gte`, `isNull`, `notEquals`, `nin`, …) to canonical on parse. A strict enum with no aliases would have broken already-stored view metadata and ObjectUI's own `FilterBuilderWidget` writes (`gt`/`gte`).
- **`isNull` → `is_null`** (not `is_empty`) to preserve null-vs-empty semantics on `sys_share_link`.
- **Kept the legacy string `sort` form** (deprecation note only) and retained the objectui#2601 regression fixture — the `sort` removal is left to its own deprecation cycle rather than bundled here.
- **Relative-date operators** (`this_quarter`, …) were dropped from the operator vocabulary (they're date-range presets, unimplemented at the filter-rule level); fixtures/JSDoc updated instead of the tests being edited to pass.

It also pins `widget: 'filter-builder'` on the list-view `filter` field in `view.form.ts` (matching `dataset`/`page` forms) so the Studio editor gets the field-type-aware builder. Paired UI work landed in objectstack-ai/objectui#2766.

Since the change is now on `main` via #3373, this PR can be closed as superseded. Really appreciate the initial push here.

---

PR #3787, opened 2026-07-28
test(service-storage): give the attachment read-visibility harness real Filter Protocol semantics

os-zhuang:
## Problem

`packages/services/service-storage/src/attachment-read-visibility.test.ts` faked its engine with a matcher that understood implicit equality and `$in` — and nothing else. No `$and`, no `$or`, no `$not`.

Every assertion in the file could therefore only check the **shape** of the filter `computeParentVisibilityFilter` emits, never the rows that filter selects. That is a poor bargain for a predicate whose entire job is narrowing: it pairs a discriminator with an id list per branch, and any bug that widens `$or` — the class fixed in the SQL compiler by #3776 — loosens that pairing while leaving every shape assertion green.

## Changes

- **Real Filter Protocol semantics in the harness matcher**, mirroring [`memory-matcher.ts`](packages/plugins/driver-memory/src/memory-matcher.ts) and [`matches-filter.ts`](packages/formula/src/matches-filter.ts): every key inside one filter object ANDs, a `$or` array ORs its branches, and a branch's contents still AND. `$and` / `$not` handled too; field specs support `$eq`/`$ne`/`$in`/`$nin`.
- **Unimplemented operators now throw** instead of being silently ignored. A test double that quietly under-implements the protocol is precisely the failure mode this PR closes, so incompleteness is loud.
- **A row-level assertion** — the case the old harness structurally could not express. A fixture with a visible and an invisible record per parent type, plus a row whose `parent_object` belongs to one branch while its `parent_id` appears only in the *other* branch's id list. The emitted scope must return `['a1','a3']`; nothing but a genuine per-branch AND excludes the other three.
- **A conformance block for the matcher itself**, on the same 2x2 fixture and the same expectations as `memory-matcher-or-semantics.test.ts`, `matches-filter-or-semantics.test.ts` and `sql-driver-or-filter.test.ts`, so the four evaluators cannot drift apart — plus a case proving the matcher actually distinguishes a correctly paired scope from a widened one.

## Proof the gate can go red

I temporarily flattened each `$or` branch's keys into key-level ORs in the matcher and re-ran. **9 tests failed**, the row assertion loudest:

```
× the multi-parent $or admits only rows matching BOTH keys of one branch
  AssertionError: expected [ 'a1', 'a2', 'a3', 'a4', 'a5' ] to deeply equal [ 'a1', 'a3' ]
```

Every row in the fixture, where two should match. The injection was reverted; a `BUG INJECTION` grep over the branch returns nothing.

## Verification

- `@objectstack/service-storage`: **212/212 passing**, 17/17 files.
- Baseline check: the 58 failures seen on a fresh worktree reproduce identically at pristine `HEAD` and are stale-`dist` artifacts (`isFileIdToken` / `RAW_FILE_VALUES_CONTEXT_KEY` exist in `packages/spec/src` but not in the Jul-25 `dist`). Rebuilding `spec` + `platform-objects` clears all of them.
- The changed file is type-clean; `pnpm --filter @objectstack/service-storage build` succeeds with DTS.

Test-only — no runtime change, hence an empty-frontmatter changeset (releases nothing).

### DIFF
--- a/.changeset/attachment-read-visibility-real-filter-semantics.md
+++ b/.changeset/attachment-read-visibility-real-filter-semantics.md
@@ -1,0 +1,38 @@
+---
+---
+
+test(service-storage): give the attachment read-visibility harness real Filter Protocol semantics (#3774)
+
+Test-only — releases nothing.
+
+`attachment-read-visibility.test.ts` faked its engine with a matcher that
+understood implicit equality and `$in` and nothing else: no `$and`, no `$or`,
+no `$not`. Every assertion in the file could therefore only check the *shape*
+of the filter `computeParentVisibilityFilter` emits, never the rows that filter
+selects.
+
+That was a load-bearing gap. The middleware's multi-parent read scope is
+`{$or:[{parent_object, parent_id:{$in:[…]}}, …]}`, and driver-sql used to
+compile each `$or` branch's own keys with OR instead of AND (#3774, fixed in
+#3776) — widening exactly this shape into an in-tenant unauthorized read of
+other users' attachments. The suite stayed green throughout, because a fake
+that never evaluates `$or` cannot see a `$or` widening bug.
+
+Changes:
+
+- The harness matcher now implements the protocol for real, mirroring
+  `driver-memory/memory-matcher.ts` and `formula/matches-filter.ts`: every key
+  in a filter object ANDs, `$or` ORs its branches, and a branch's own contents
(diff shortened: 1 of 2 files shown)


os-zhuang:
Superseded by #3792 — same change, reopened from a branch whose commit messages follow this repo's neutral-prose practice for filter-widening work. Closing this one and deleting its branch.

PR #3834, opened 2026-07-28
fix(approvals,rest): tell the truth about approval writes — lock policy, dropped fields, and the batch create ingress (#3794, #3835)

baozhoutao:
Closes #3794 (the framework half — the Console half is objectstack-ai/objectui#2914) and #3835.

An approval flow reported record writability wrong in **both** directions: what the user *could* change said "locked", and what they *couldn't* said "updated successfully". Neither was wrong *behaviour* — the server did the right thing and told nobody. Pulling on the second thread turned up a third problem on the same write path, where the server was **not** doing the right thing (#3835).

## 1. `locks_record` on the approval request (#3794 problem 1)

The `beforeUpdate` lock hook decides on exactly one thing: the node config snapshot's `lockRecord` (`=== false` ⇒ the update goes through). Nothing exposed it, so a client saw only "a request is pending" and had to guess whether that means locked — and the Console guessed "locked", every time. A `lockRecord: false` node exists *precisely* so an approver can amend the record while deciding on it; painting "Locked for approval" over that hides the whole feature and approvers never try.

`rowFromRequest` now emits `locks_record`, read from the **same** snapshot the hook reads with the **same** default-true, so the two cannot disagree. It flows through `getRequest` / `listRequests` and therefore `GET /api/v1/approvals/requests`. Optional in the contract for version skew (absent ⇒ assume locked).

## 2. `droppedFields` on `POST /batch` (#3794 problem 2)

The engine strips writes to `readonly` (#2948) and `readonlyWhen`-locked (#3042) fields and completes the write without them. Every write path already reports what it dropped (#3431/#3455) — **except** the cross-object transactional batch, which never wired `onFieldsDropped` at all.

That is not a marginal path: it is the Console record form's save for a master-detail record, i.e. the one surface where a **user** edits a `readonlyWhen` field. They changed it, the form said "updated successfully", the value never moved, and nothing anywhere said so — the exact symptom in #3794.

The response now carries a top-level `droppedFields` list, each event tagged with the `index` of the operation that produced it (`results` entries are bare record echoes, with no envelope to hang a per-row list on). Omitted entirely when nothing was dropped, so the shape stays backward-compatible; the batch still commits either way — a strip is legal semantics, not an error.

## 3. …and that batch was not enforcing `readonly` on create at all (#3835)

[...]

PR #3844, opened 2026-07-28
fix(rest): route batch creates through the create ingress so `readonly` is enforced (#3835)

### LINKED ISSUE #3835: 跨对象事务批 `POST /batch` 绕过 create 侧的 `readonly` 入口剥离（#3043）
## 现象

`POST /api/v1/batch`（跨对象事务批，ADR-0034 / #1604）在事务里**直接调 `ql.insert` / `ql.update`**：

```ts
// packages/rest/src/rest-server.ts — /batch handler
if (op.action === 'create') {
    out.push(await ql.insert(op.object, data, { context: trxCtx, onFieldsDropped }));
}
```

而单条创建走的是 `protocol.createData`，它在进引擎**之前**先跑一次 `stripReadonlyForInsert`：

```ts
// packages/metadata-protocol/src/protocol.ts — createData
const data = stripReadonlyForInsert(
    this.engine.registry?.getObject(request.object),
    request.data,
    request.context,
);
```

这层入口剥离存在的原因是引擎的 INSERT 路径**对静态 `readonly` 免疫**（#3413：`stripReadonlyFields` 只在 update 分支跑）。也就是说：

- `POST /api/v1/data/:object` → 非系统调用者塞 `readonly` 列 → 被剥离；
- `POST /api/v1/batch`（`action: 'create'`）→ 同一个列 → **原样落库**。

同一个语义，两条写路径两个结果。

## 复现思路

对象 A 上声明一个 `readonly: true` 字段（例如 `approval_status`），以普通用户身份：

1. `POST /api/v1/data/A  {"approval_status":"approved", ...}` → 该字段被丢弃，取默认值（并在 `droppedFields` 里报出来）；
2. `POST /api/v1/batch  {"operations":[{"object":"A","action":"create","data":{"approval_status":"approved", ...}}]}` → 该字段被写进去。

## 影响

`readonly` 在 #2948/#3043 之后是**写路径约束**，不只是渲染提示。批量创建这条路留着口子，意味着任何能调 `/batch` 的客户端都能给审计戳、来源字段、审批状态镜像等只读列赋任意值——而这类字段常常正是别处逻辑信赖的。Console 的主记录表单（master-detail）走的就是 `/batch`，所以这不是一条冷门路径。

## 建议方向

按 PD #12「修生产者、别在消费者兜底」：把 create 侧的入口剥离**提到两条路共用的位置**，而不是在 `/batch` 里再抄一遍 `stripReadonlyForInsert`。两种可行落点：

- 让 `/batch` 的 create 分支走 `protocol.createData` 的同一层（而不是裸 `ql.insert`）；或
- 把静态 `readonly` 的 INSERT 剥离下沉进引擎（撤销 #3413 的 insert 豁免），入口层那次剥离随之退化为冗余并可移除。

第二种更接近「一个契约一处执行」，但要先确认 #3413 当初豁免 insert 的理由（系统种子 / seed 写入路径）在下沉后仍走得通——那批调用是 `isSystem`，本来就在剥离的豁免名单里。

## 来源

在 #3794 排查中发现（修复 PR：#3834，该 PR 只给 `/batch` 接上了 `onFieldsDropped` 的写可观测性，**没有**动这里的剥离行为——按 AGENTS.md PD #10 单独立此 issue，不扩大原范围）。

baozhoutao:
Closes #3835.

> **Stacked on #3834** — it touches the same `/batch` handler block and this branch is cut from it, so the diff shows both until #3834 lands. Merge #3834 first; I'll rebase this to a single commit.

## `readonly` meant two different things depending on which create endpoint you used

`POST /data/:object` runs the #3043 ingress strip, so a non-system caller cannot seed a read-only column — the field is dropped and reported. The cross-object transactional batch called `ql.insert` directly and skipped that ingress, and the engine's INSERT path is static-`readonly`-exempt **by design** (#3413 — the strip deliberately lives one layer up). So nothing enforced it on that route.

Measured on the showcase, `showcase_contact.lead_score` (`readonly: true`), same signed-in non-system user, identical payload:

| | before | after |
|---|---|---|
| `POST /data/showcase_contact` | `lead_score = null`, `droppedFields` reported | unchanged |
| `POST /batch` (`action: 'create'`) | **`lead_score = 999` written, nothing said** | `null`, `droppedFields` reported |

## Why route to the ingress instead of re-applying the strip here

Reading `stripReadonlyForInsert` turned up a **deliberate scope carve-out** that a second copy would have lost: platform objects (`sys_` / `managedBy`) are exempt on purpose, because their own guards must **reject** a forged value with 403 (ADR-0086's `managed_by`/`package_id`, #3004's `owner_id` anchor) — silently swallowing it would eat the very payload the guard exists to refuse. `isSystem` is exempt for the same class of reason.

That rules out the "sink the strip into the engine" option I floated in the issue: the engine has no notion of that boundary. So the batch's create ops now go through `p.createData` — the ingress itself. One create ingress, and a future change to its policy covers the batch for free (AGENTS.md PD #12: fix the producer, don't grow a second de-facto contract).

Mechanics that had to keep working, and do:

- **Transaction** — `trxCtx` is passed as the context, so the insert joins this batch's transaction and rolls back with it (asserted in a test: the ingress receives the trx context, not a bare one).
- **`{ $ref: <opIndex> }`** — `results` now holds the ingress's echoed record instead of the raw engine return; `$ref` resolves off `.id` either way. Verified live: an invoice + line batch produced a line whose `invoice` points at the invoice created in the same request.
- **`droppedFields`** — the ingress's events fold into the batch's per-op list (#3794), so a batch create reports strips the way an update does.

Update ops are untouched — the engine enforces `readonly` (#2948) and `readonlyWhen` (#3042) on its own update path.

## Tests

3 new in `rest-batch-endpoint.test.ts`: create ops go through the ingress with the transaction context; a forged `readonly` column never reaches the engine and comes back in `droppedFields` tagged with its op index; `$ref` still resolves through the ingress. The harness's protocol mock gained a `createData` that delegates to the same `ql.insert`, so existing assertions still read off `ql.insert`.

`rest`: 413 passing.

### DIFF
--- a/.changeset/approval-lock-and-batch-write-observability.md
+++ b/.changeset/approval-lock-and-batch-write-observability.md
@@ -1,0 +1,40 @@
+---
+"@objectstack/plugin-approvals": minor
+"@objectstack/spec": minor
+"@objectstack/rest": minor
+---
+
+fix(approvals,rest): tell the truth about what a write did — `locks_record` on the request, `droppedFields` on the cross-object batch (#3794)
+
+Two halves of the same complaint: in an approval flow the platform reported
+record writability wrong in *both* directions — what you could change said
+"locked", and what you couldn't said "updated successfully".
+
+**`locks_record` on the approval request.** The lock hook reads one thing to
+decide whether a pending approval blocks writes: the node config snapshot's
+`lockRecord` (`=== false` ⇒ the update goes through). Nothing exposed that, so a
+client had only "a pending request exists" and had to guess — and the Console
+guessed "locked", every time. A `lockRecord: false` node exists precisely so the
+approver can amend the record while deciding on it; painting "Locked for
+approval" over that hides the whole feature, and approvers never try. Request
+rows (`getRequest` / `listRequests`, and therefore `GET /api/v1/approvals/requests`)
+now carry `locks_record`, read from the same snapshot the hook reads, with the
+same default-true. Absent on rows from an older server ⇒ assume locked.
+
+**`droppedFields` on `POST /batch`.** The engine strips writes to `readonly`
+(#2948) and `readonlyWhen`-locked (#3042) fields and completes the write without
+them. Every write path already reported which fields it dropped (#3431/#3455) —
+except the cross-object transactional batch, which never wired
+`onFieldsDropped` at all. That path is the Console record form's save for a
+master-detail record, so it is exactly where a *user* edits a `readonlyWhen`
+field: they changed it, the form said "updated successfully", the value never
+moved, and nothing anywhere said so. The response now carries a top-level
+`droppedFields` list, each event tagged with the `index` of the operation that
+produced it (`results` entries are bare record echoes with no envelope to hang a
+per-row list on). Omitted entirely when nothing was dropped, so the shape stays
+backward-compatible; the batch still commits either way — a strip is legal
+semantics, not an error.
+
+The Console half of both fixes ships in objectui: the detail band now
+distinguishes "in approval (editable)" from "locked for approval", and the
+write-warning toast fires on batch saves.
--- a/.changeset/batch-create-readonly-ingress.md
+++ b/.changeset/batch-create-readonly-ingress.md
@@ -1,0 +1,39 @@
+---
+"@objectstack/rest": minor
+---
+
+fix(rest): a batch create goes through the same create ingress as a single create (#3835)
+
+`readonly` meant two different things depending on which create endpoint you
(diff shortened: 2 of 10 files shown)


baozhoutao:
折叠进 #3834 了——两个改动落在同一段 `/batch` 处理器上,拆成叠加 PR 只会让审核人多背一次合并顺序。这条提交已 cherry-pick 到 #3834 分支,#3835 由 #3834 一并关闭。

PR #3848, opened 2026-07-28
fix(objectql)!: one key for the empty group bucket — real `null`, on both aggregation paths (#3839)

### LINKED ISSUE #3839: 空分组桶的键两条路不一致:下推 SQL 给 `null`,内存兜底给 `"(null)"`(不限日期分桶)
同一个数据集,同一个查询,分组键为空的那个桶,**下推 SQL 给 `null`,内存兜底给字符串 `'(null)'`**。`engine.aggregate` 按查询逐次在两条路之间挑一条,所以同一个仪表盘会因为走了哪条路而拿到不同类型的桶键。

实测(better-sqlite3,两行:一行有值、一行 `at`/`stage` 均为 NULL):

```
--- 日期分桶 groupBy (dateGranularity: 'month') ---
  SQL      : [{"key":null,     "type":"null",   "total":2}, {"key":"2026-01","type":"string","total":1}]
  in-memory: [{"key":"(null)", "type":"string", "total":2}, {"key":"2026-01","type":"string","total":1}]

--- 普通 groupBy (['stage']) ---
  SQL      : [{"key":null,     "type":"null",   "total":2}, {"key":"won","type":"string","total":1}]
  in-memory: [{"key":"(null)", "type":"string", "total":2}, {"key":"won","type":"string","total":1}]
```

**比最初以为的宽**:这不是日期分桶专属。普通 `groupBy` 对空值列一样分歧 —— 度量总额两边都对(2 和 1),坏的只有键的**类型和字面量**。

## 两侧各自来自哪

- **内存侧**产出 `'(null)'`,两个来源:
  - `packages/objectql/src/in-memory-aggregation.ts:87` — `projectGroupValue`:`v == null ? '(null)' : String(v)`(普通 groupBy)
  - `packages/objectql/src/in-memory-aggregation.ts:211` — `bucketDateValue`:`if (value == null) return '(null)'`(日期分桶)
- **SQL 侧**产出 SQL NULL:分组列本身为 NULL,或 `strftime(...)` 对 NULL 输入返回 NULL。驱动不做转换,原样交出。

## 引擎按什么挑路

`engine.aggregate`(`packages/objectql/src/engine.ts`)下推的条件是:每个结构化 groupBy 项的 granularity 都被 `supports.queryDateGranularity` 声明,**且**不是非 UTC 时区。否则走 `driver.find()` + `applyInMemoryAggregation`。所以同一个数据集:

- SQLite + UTC + month → 下推 → 键是 `null`
- 同一个查询换成 `week`(SQLite 不声明)或换个非 UTC 时区 → 内存 → 键是 `'(null)'`
- driver-rest / driver-memory / Turso remote(`queryDateGranularity: {}`)→ 恒内存 → 恒 `'(null)'`

也就是说**换一个时区、换一个粒度、换一个驱动,同一份数据的空桶键就换一种形状**。

## 与 #3773 的关系:正交,且刻意未纳入

#3773/#3813 修的是"桶标签算错"。这一条不是算错——两边都正确地表示了"空",只是表示法不同。所以 `checkDateBucketParity`(`packages/verify/src/date-bucket-parity.ts`)的 fixture **刻意不含 null 实例**,并在注释里写明了理由:放进去会让那道门**为一个它不负责的原因而红**。

→ **修这条时,顺带把那条排除去掉**,让 fixture 带一个 null 实例,门就能长期守住收敛后的约定。落点在 `date-bucket-parity.ts` 的 `FIXTURE` 注释处。

## 往哪边收敛:未定,而且现有依据站不住

注释 `in-memory-aggregation.ts:25` 说 `'(null)'` 是「to remain consistent with the client `useReportData` hook」。核实结果:

- `useReportData` 确实存在,在 **objectui** 的 `packages/plugin-report/src/index.tsx`;
- 但该文件里 **`grep "(null)"` 无任何命中** —— 这个字面量不在它里面。

所以"客户端约定"这条依据需要重新核实,不能直接拿来定方向。两个方向都要先回答"谁在依赖现状":

1. **内存侧改成产出 `null`** —— 与 SQL 对齐,类型也更诚实(空就是空)。风险:任何按 `'(null)'` 字符串匹配的下游会静默失配。
2. **SQL 侧改成产出 `'(null)'`** —— 与现有内存约定对齐,前端不必区分。风险:把一个哨兵字符串塞进本可为 null 的数据列,且要在每个驱动上实现。

一个已知的相关点:`packages/services/service-analytics/src/strategies/cross-object-rebucket.ts:124` 用 `String(row[f] ?? '(null)')` 构造**内部**分组键,所以它在重新分桶时会把两种形状**合并**成同一个桶(不会因为形状不同而分裂);但输出仍然 `bucket[f] = row[f]`,原样保留,所以分歧照旧传到下游。

## 建议

先做**调查**再改:枚举 `'(null)'` 与"空桶键"的全部消费方(framework + objectui + cloud 三仓),确认现状到底有没有人依赖,再定方向。这条不阻塞任何东西 —— 度量值一直是对的,坏的只有一个空桶的标签形状。

os-zhuang:
Closes #3839.

## Outcome

A grouped row whose dimension value is empty now carries **real `null`** for that dimension, whichever way the aggregate ran. Downstream code can test the empty bucket with a plain `value == null` again: charts render their own empty label, a drill on that bucket builds `field = null` and returns the rows it should, and a dashboard no longer changes shape when the driver, the granularity or the reference timezone changes.

The reproduction from the issue, re-run against this branch:

```
--- 日期分桶 groupBy (dateGranularity: 'month') ---
  SQL      : [{"key":"2026-01","type":"string","total":1},{"key":null,"type":"null","total":2}]
  in-memory: [{"key":"2026-01","type":"string","total":1},{"key":null,"type":"null","total":2}]
  MATCH    : ✅ same key + type
--- 普通 groupBy (['stage']) ---
  SQL      : [{"key":null,"type":"null","total":2},{"key":"won","type":"string","total":1}]
  in-memory: [{"key":null,"type":"null","total":2},{"key":"won","type":"string","total":1}]
  MATCH    : ✅ same key + type
```

## Which way, and why that direction

The issue asked for the investigation before the change. Enumerating every consumer of `'(null)'` and of empty bucket keys across framework + objectui + cloud, the evidence is one-sided:

- **Nothing in production depends on the sentinel.** The only hard dependencies in all three repos are four assertions in two objectql test files. Zero production code, in any repo, compares against `'(null)'`.
- **The stated justification is dead.** `in-memory-aggregation.ts` said the string existed "to remain consistent with the client `useReportData` hook". That hook was removed with ADR-0021 — objectui `packages/plugin-report/src/index.tsx` is now just its epitaph — and the literal never appeared in it even when it existed.
- **Every consumer is already written against `null`, and the sentinel actively breaks them.** objectui's empty labels are all keyed off `== null` / `??` — `formatDimensionValue` → `—`, `useGroupedData` / `PivotTable` / `ReportViewer` → `(empty)`, kanban → localized "Uncategorized", `chart-series` drops the null series. The sentinel bypassed all of them and rendered a raw English debug string. Worse, `buildDatasetDrillFilter` and `ReportView` build the drill as `field = '(null)'`, so clicking the empty bucket returned **zero rows** instead of the empty-valued ones.
- **Everything else already emits real `null`**: the pushed-down SQL path, cloud's Turso remote transport, and framework's own newer semantic-layer evaluator (`preview-evaluator.ts`).
- **No type or schema blocks it** in either direction — the result row shape is untyped, and objectui types the group key as `Record<string, unknown>`.

Converging to `null` is therefore a bug fix, not just a cleanup.

## Changes

- `applyInMemoryAggregation` / `bucketDateValue` (objectql) key the empty bucket as `null`; `bucketDateValue` now returns `string | null`. A null instant and an unparseable one still share one bucket, because SQL cannot tell them apart either (`strftime('%Y-%m', 'not-a-date')` is NULL).
- The internal composite bucket id is JSON-encoded so the empty bucket stays distinct from a row whose value is the literal string `"null"` — `${null}` is `'null'`, so plain interpolation would have merged two real groups. Same in service-analytics' cross-object rebucketing.
- `bucketKeyToCalendarRange` (core) accepts `string | null`. Behavior unchanged — the empty bucket has no calendar span, so a drill on it opens the unscoped superset — but the call site no longer casts a lie.
- The driver output contract in spec now states the rule: a row with no value keys as `null`, never a sentinel. Propagating NULL through the bucket expression is the whole of it; a driver only breaks it by adding a `COALESCE`.

## Gates

As the issue requested, `checkDateBucketParity`'s deliberate exclusion is removed and its `FIXTURE` now carries a null instant.

Two latent bugs in that check had to be fixed first, or the new fixture would have been meaningless — and one of them would have made it **silently bless** the divergence:

1. It folded bucket labels through `String(value)`, which turns SQL NULL into `'null'` — a label a TEXT column can genuinely hold. A side spelling "empty" as a string could compare **equal** to one returning real NULL. The empty bucket is now keyed out of band.
2. It compared label sets with `JSON.stringify`, which is sensitive to key **insertion order**. Row order is not part of this contract and the two paths legitimately differ (SQL sorts its groups; the in-memory path emits first-seen order), so a driver with entirely correct buckets was reported as disagreeing — with an empty diff message, since `describeDiff` is keyed and could not name one. This actually fired on both real drivers once the null row was added. The comparison is now order-insensitive.

A new dogfood check, `empty-group-bucket-parity.test.ts`, covers the non-date half against real drivers — the issue's point that this was never date-specific — for both `driver-sql` and `driver-sqlite-wasm`, both groupBy shapes.

## Verification

- **Both gates proven able to fail.** With the sentinel temporarily restored and objectql rebuilt, all four new/extended real-driver cases go red with exactly the #3839 signature: `(null): sql=— in-memory=1, ‹empty bucket›: sql=1 in-memory=—`. Restored and green again.
- Two new negative controls pin the sentinel detection itself (a driver spelling empty as `'(null)'`, and as `'null'`).
- Suites green: objectql 1138, driver-sql 401, driver-sqlite-wasm 126, core 414, service-analytics 275, verify 7, dogfood 387.
- `pnpm build` clean with no generated-artifact drift; `pnpm lint` clean; `check:nul-bytes` / `check:doc-authoring` / `check:role-word` / `check:release-notes` all OK.

## Note for reviewers

The in-memory path also `String()`-coerces every non-null group value, so a numeric column's bucket keys come back as strings there and as numbers from SQL. That is a second, separate divergence in the same function with a much wider blast radius, and it is **not** touched here. Worth its own issue if you agree.

### DIFF
--- a/.changeset/empty-group-bucket-key-null.md
+++ b/.changeset/empty-group-bucket-key-null.md
@@ -1,0 +1,87 @@
+---
+"@objectstack/objectql": minor
+"@objectstack/verify": minor
+"@objectstack/core": patch
+"@objectstack/service-analytics": patch
+"@objectstack/spec": patch
+---
+
+fix(objectql)!: one key for the empty group bucket — real `null`, on both aggregation paths (#3839)
+
+A grouped row whose dimension value is empty now carries `null` for that
+dimension no matter which way the aggregate ran. Downstream code can test the
+empty bucket with a plain `value == null` again: charts render their own empty
+label, drill-through on that bucket builds `field = null` and returns the rows
+it should, and a dashboard no longer changes shape when the driver, the
+granularity or the reference timezone changes.
+
+### What was wrong
+
+`engine.aggregate` has two implementations of one feature. It pushes the
+aggregate down as SQL when the driver advertises every requested granularity and
+the reference timezone is UTC; otherwise it fetches rows and buckets them in JS.
+The two disagreed about how to spell "empty":
+
+```
+--- same dataset, same query, one row with a NULL value ---
+  pushed-down SQL : [{ "key": null,     "type": "null",   "total": 2 }, …]
+  in-memory       : [{ "key": "(null)", "type": "string", "total": 2 }, …]
+```
+
+The measures were always right — only the key's type and literal differed —
+which is why this went unnoticed for so long: every total reconciled. But the
+engine picks a path per query, so the same data produced a different bucket key
+on SQLite-plus-UTC-plus-`month` than on `week` (which SQLite does not advertise),
+a non-UTC timezone, or `driver-rest` / `driver-memory` / a remote Turso, all of
+which bucket in memory unconditionally.
+
+It was never date-specific either. A plain `groupBy: ['stage']` over a NULL
+column diverged the same way.
+
+Consumers are written against `null` — they check `== null` and supply their own
+empty label ('—', '(empty)', a localized "Uncategorized"). The sentinel defeated
+every one of them: it rendered a raw English debug string in the UI, and a drill
+on the empty bucket compiled to `field = '(null)'` and matched nothing.
+
+The in-memory path's comment justified the string as staying "consistent with
+the client `useReportData` hook". That hook was removed with ADR-0021, and the
+literal never appeared in it.
+
+### What changed
+
+- `applyInMemoryAggregation` and `bucketDateValue` (`@objectstack/objectql`) key
+  the empty bucket as `null`. `bucketDateValue` now returns `string | null`. A
+  null instant and an unparseable one still share one bucket, because SQL cannot
+  tell them apart either (`strftime('%Y-%m', 'not-a-date')` is NULL).
+- The internal composite bucket id is JSON-encoded, so the empty bucket stays
+  distinct from a row whose value is the literal string `"null"`.
+- `bucketKeyToCalendarRange` (`@objectstack/core`) accepts `string | null`. The
+  empty bucket has no calendar span, so a drill on it opens the unscoped
+  superset instead of an invented bound — unchanged behavior, honest signature.
+- The driver output contract in `@objectstack/spec` now states the rule: a row
+  with no value keys as `null`, never a sentinel. Propagating NULL through the
+  bucket expression is the whole of it; a driver only breaks it by adding a
+  `COALESCE`.
+
+### Gates
+
+`checkDateBucketParity` (`@objectstack/verify`) deliberately carried no null
+instant, because the divergence would have failed it for a reason it was not
+about. Its fixture now has one, so the convergence is held in place — including
+for out-of-tree drivers that run the check against themselves.
+
+Two fixes were needed to make that fixture meaningful:
+
+- The check folded bucket labels through `String(value)`, which turns SQL NULL
+  into `'null'` — a label a TEXT column can genuinely hold. A driver spelling
+  "empty" as a string could compare equal to one returning real NULL. The empty
+  bucket is now keyed out of band.
+- Label sets were compared with `JSON.stringify`, which is sensitive to key
+  insertion order. Row order is not part of this contract and the two paths
+  naturally differ (SQL sorts its groups; the in-memory path emits first-seen
+  order), so a driver with entirely correct buckets could be reported as
+  disagreeing — with an empty diff message, since nothing actually differed.
+  The comparison is now order-insensitive.
+
+A new dogfood check covers the non-date half against real drivers: same dataset,
+plain and date-bucketed `groupBy`, both paths, one key.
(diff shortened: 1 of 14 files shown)


os-zhuang:
Follow-up filed as #3849 — the `String()` coercion divergence flagged in the "Note for reviewers" section, now with evidence. It is worse than described: on a **boolean** column the two paths produce `0`/`1` (SQL) vs `"false"`/`"true"` (in-memory) — no overlap at all, so that is a semantic divergence, not just a type one. Deliberately out of scope here.

Issue #3855, opened 2026-07-28
[P3] Retire the three deprecated aliases — via the ADR-0087 D2 conversion layer, not by deleting the keys


os-zhuang:
Filed per Prime Directive #10 rather than expanding #3838 / #3854.

> **Revised.** The first version of this issue proposed deleting the keys from the Zod objects and adding `UNKNOWN_KEY_GUIDANCE` tombstones. That is wrong on both halves — see "The trap" below. The corrected plan is the conversion layer.

## Where the three aliases stand

The spec carries three fold-and-drop aliases. Each is accepted on input, lowered into its canonical key, and dropped from the parsed output:

| Alias | Canonical | Transform | Since |
|---|---|---|---|
| `action.execute` | `action.target` | `ui/action.zod.ts:683` | #3713 / #3742 |
| `field.conditionalRequired` | `field.requiredWhen` | `data/field.zod.ts:622` | #3754 / #3764 |
| `agent.knowledge.topics` | `agent.knowledge.sources` | `ai/agent.zod.ts:43` | #1878 / #1891 |

Two things have landed for all three: the conflict is unrepresentable on the parsed output (#3742 and siblings), and the discard is no longer silent (#3838, #3854). What is missing is a retirement date — all three say "will be removed in a future version" with no version named.

## The trap: deleting the key is a silent strip, not an error

**None of the three schemas is `.strict()`.**

- `FieldSchema` — `data/field.zod.ts:503` says so in a comment, and records that this exact trap already bit: `dataQuality` and `cached` outlived their keys by a release, stayed discoverable on the API surface and in the generated reference docs, and an author writing them got a clean parse with the key silently stripped (#3726, #3733 — the ADR-0104 failure class).
- `ActionSchema` (`ui/action.zod.ts:389`) and `AIKnowledgeSchema` (`ai/agent.zod.ts:35`) are bare `z.object`, so Zod's default `.strip()`.

So "delete the key from the Zod object" does not produce an error — it produces a **silent strip**. An author still writing `conditionalRequired` after the removal gets: parses clean, key vanishes, field never required. That is precisely the failure mode #3713 → #3743 → #3838 → #3854 exists to eliminate, reintroduced one layer down.

[...]

Issue #3743, opened 2026-07-28
[P3] action `target` + `execute` both declared: the alias is discarded silently — no author-facing warning


os-zhuang:
Follow-up to #3713 / #3742, filed per Prime Directive #10 rather than expanding that PR's scope.

## What #3742 fixed, and what it didn't

#3742 made the `target` / `execute` precedence consistent across all three readers and had the `ActionSchema` transform **drop** the alias from its parsed output, so the "two different scripts for one button" divergence is now unrepresentable.

What it did **not** change: when an author declares both slots with *different* values, `execute` is discarded **silently**. That is now documented, tested and consistent — a big improvement on "the wrong code ran" — but the author still gets no signal that one of the two handlers they wrote is being thrown away.

Per Prime Directive #12 ("reject it at authoring/publish so the error surfaces loudly"), that deserves a warning.

## Why it wasn't done in #3742

The natural home is the CLI compile lint pipeline, which already has the right shape for this — advisory findings that never fail the build (`lintFlowPatterns`, `lintLivenessProperties`, `lintViewRefs` in `packages/cli/src/commands/compile.ts`).

The problem is ordering. Every one of those lints runs on the **parsed** stack:

```ts
// packages/cli/src/commands/compile.ts
const lowering = lowerCallables(normalized);            // ~L107  (pre-parse)
const result = ObjectStackDefinitionSchema.safeParse(lowering.lowered);  // ~L136
…
const flowLint = lintFlowPatterns(result.data as Record<string, unknown>);  // ~L405  (post-parse)
```

By the time the existing lints see the stack, the transform has already consumed `execute` — so a lint added alongside them is structurally blind to the conflict it is meant to report.

The string/string case (exactly the one #3713 describes: `target: 'preferredHandler', execute: 'legacyHandler'`) *is* still visible at line ~136, because `lowerCallables` deliberately leaves a string pair alone for the spec transform to resolve. So the fix is available — it just needs a **new pre-parse lint hook**, not another entry in the post-parse block. That's a structural change to the compile pipeline, which is more than the P2 fix warranted.

## Suggested shape

1. Add a pre-parse lint pass in `compile.ts` between lowering and `safeParse`, operating on `lowering.lowered`.
2. First rule: `action.target` and `action.execute` both present as strings with **different** values → advisory warning naming both, stating that `target` wins and `execute` is discarded, and pointing at the one-line fix (delete `execute`). Equal values are harmless — stay quiet.
[...]

PR #3895, opened 2026-07-28
fix(spec): ratchet the authorable key surface — the one contract no witness watched (#3855)

os-zhuang:
Follow-up to #3883. Non-breaking — a new gate plus a corrected docblock.

## The gap

For a metadata-driven platform the third-party API is **what an author may write**: the keys inside each schema. Nothing guarded them.

| Witness | Records | Sees a removed key? |
|---|---|---|
| `api-surface.json` | exported `name (kind)` | ❌ a key inside a schema is not an export |
| `api-surface-signatures.json` | hash of each `defineX` factory's type | ❌ hashes `checker.typeToString()`, which prints a type **reference** (`z.input<typeof ActionSchema>`) and never expands it structurally |
| `spec-changes.json` `removed[]` | diff of `api-surface.json` | ❌ inherits the same blind spot |
| downstream-contract fixture | frozen third-party consumer | ❌ only what it happens to exercise |
| `json-schema.manifest.json` | whole schemas | ❌ `AIKnowledge` still exists after `topics` goes |

Verified concretely: **#3883 narrowed `defineAction`'s input by three keys and `api-surface-signatures.json` did not change** — its last commit is still the unrelated #3735.

And this is not only a "deliberate change slipped a gate" story. **#3733 was the accidental version**: `dataQuality` / `cached` outlived their keys, stayed discoverable on the API surface and in the reference docs, and an author writing them got a clean parse with the key silently stripped.

ADR-0059 §5 named the trigger:

> A signature-level snapshot is **deferred until a narrowing actually slips both** (evidence-gated, like ADR-0054's Phase 3).

It has now, twice.

## The gate

`authorable-surface.json` — a committed ratchet of all **8588** authorable keys, derived from the walk that already emits the JSON Schemas. One level deeper, no new introspection, one pass.

It distinguishes three states for free, because a tombstoned key (`retiredKey()`, from #3883) is `z.never()`, which Zod renders as `{ "not": {} }`:

```
ai/AIKnowledge:topics [RETIRED]
data/Field:conditionalRequired [RETIRED]
ui/Action:execute [RETIRED]
ui/Action:body
ui/Action:target
…
```

| State | Meaning |
|---|---|
| live | a normal property — the author may write it |
| `[RETIRED]` | present but unwritable, carrying its own upgrade prescription |
| absent | gone from the contract with nothing left to say |

## Three failure modes, each proven non-vacuous

I simulated all three rather than assuming they fire — the whole point of this PR is that a gate can look present and be dead.

**1. A key vanishes with no tombstone** — the silent-strip class.

```
❌ 1 authorable key(s) disappeared from the contract:
     - ui/Action:phantomLegacyKey

   These schemas are NOT .strict(), so Zod silently STRIPS an unknown key: an author
   who keeps writing one gets a clean parse and a setting that never takes effect —
   no error, nothing to grep, nothing pointing at the changelog (#3733, ADR-0104).
```
…followed by the full retirement protocol (tombstone → D2/D3 entry → `major` changeset).

**2. A key is tombstoned with no registered migration** — audible to whoever hits it, but the change documentation is still empty.

```
❌ 1 key(s) were tombstoned with no registered migration:
     - ai/AIKnowledge:indexes

   The tombstone makes the removal audible to whoever hits it, but the change
   documentation is the primary channel and it is still empty: spec-changes.json
   (ADR-0087 D4) is a projection of the conversion + migration registries…
```

This is the rule that mechanizes what I did by hand in #3883 — where nothing would have stopped me skipping it.

**3. An addition is left uncommitted** (`--check` mode).

```
❌ authorable-surface.json is out of date (1 key(s) not recorded).
     + ui/Action:aria
```
An unrecorded key is invisible to the ratchet forever after, since it can only detect the disappearance of something it once saw.

Failures `process.exit(1)` **before** the snapshot write, so a red run never clobbers the baseline.

## Where it runs

- **`check:docs`** (`gen:schema && build-docs --check`) already runs in lint.yml's typecheck job, which has **no paths filter and is a required check** — the job whose own comment says "the gate cannot go dormant". Removal detection lands there for free.
- **`check:authorable-surface`** as an explicit step in `Check Generated Artifacts`, with the `generated` paths filter widened to `packages/spec/src/**` + the script + the snapshot — per that filter's own rule that "a path missing here makes that gate dormant on exactly the PRs that can break it, so keep the two in lockstep".

## Scope, deliberately

Value-level narrowing (an enum losing a member, a union narrowing) stays **ungated**, per ADR-0059 §5's evidence gate — no such narrowing has slipped yet, and snapshotting it would churn on every internal tweak, which is exactly why the signature hash was scoped to factories in the first place. Noise kills a ratchet faster than a gap does.

What I did fix is the **over-promise**: `build-api-surface.ts` advertised its signature snapshot as answering *"did the accepted authoring shape narrow?"*. It does not, and believing it did is what let this gap sit. The docblock now states the scope and points at the new ratchet.

## Testing

- `@objectstack/spec` 6818 passed · spec `tsc --noEmit` clean
- Every ESLint-job and typecheck-job gate green locally, including the two the new snapshot could disturb: `check:docs`, `check:api-surface`, plus `check:spec-changes` / `check:upgrade-guide` / `check:liveness` / `check:skill-refs` / `check:react-blocks`
- The three simulations above, each reverted afterwards; the committed snapshot is byte-identical to a clean regeneration

---
_Generated by [Claude Code](https://claude.ai/code/session_01JqC9ckaLbmhmytt65Gxi96)_


os-zhuang:
One more piece of evidence, surfaced by the docs-drift check on this PR.

`content/docs/protocol/backward-compatibility.mdx` has been making this exact promise to third parties all along:

> **Zod schemas** are part of the public API surface. Adding optional properties is a MINOR change; **removing or renaming properties is a MAJOR change**.

That is precisely the rule this ratchet enforces — and until now nothing backed it. The guarantee was published, the enforcement was absent, and the two gates that looked like they covered it were watching the TypeScript export surface instead.

So this is the same `declared ≠ enforced` pattern the whole #3743 → #3855 line has been closing, except this instance was written into a consumer-facing compatibility guarantee rather than into metadata. The doc needs no edit — it is more true after this PR than before it.

---

Issue #3983, opened 2026-07-30
`share-link-routes.ts` is the fifth drifting envelope module — no body carries `success`, and one answers the `{ ok: true }` #3689 retired


os-zhuang:
Found by the repo-wide guard added in #3843 (PR #3972), not by hand. #3843's survey listed four modules; this is a fifth, and it was missed for a specific and repeatable reason: every guard before this one was **per-package**, so a module nobody thought to convert was never audited. Filing it rather than letting it ride along, the same call #3675's changeset made about storage's success bodies.

Mildly ironic: #3675's own changeset cites this file as an example of the *good* shape — "the nested-`error` shape the sibling services already use (`settings-routes.ts`, `share-link-routes.ts`)". That was true of its `error` object and not of anything else.

## The drift

`packages/plugins/plugin-sharing/src/share-link-routes.ts`, 6 response write sites:

```ts
res.status(status).json({ error: { code, message } });          // :52  — sendError: nested, but NO `success: false`
await res.status(201).json({ link });                          // :104 — bare success
await res.json({ links: link });                               // :127 — bare success
await res.status(200).json({ ok: true });                       // :139 — the private second word for `success`
await res.json({ record: …, link: { … } });                    // :209 — bare success
res.status(200).json({ data: rows ?? [] });                     // :265 — `data` present, flag absent
```

So this module carries **three** of the four dialects the #3636 → #3843 sequence has been retiring:

| dialect | retired by | still here |
|---|---|---|
| no `success` flag on any body | #3689 (storage), #3843 (four more) | ✅ all 6 |
| `ok: true` as a second success word | #3689 (storage's `{ ok: true, key }`) | ✅ `:139` |
| payload at the top level instead of under `data` | #3689, #3843 | ✅ 4 sites |
| `error` as a bare string | #3675 | ❌ already nested — the one it got right |

`:265` is the interesting one: it already answers `{ data }`, so it is one field short of conformant.

## Why it is not fixed in #3843

Converting it is breaking for share-link consumers and needs a consumer sweep of its own — public share links are an end-user-facing feature, and `GET /share/:token` is read by the anonymous share view, not just by the SDK. That is exactly the shape of change #3687 did for storage: producer + consumers, landed deliberately, with the console taught both shapes first. It is not a quiet ride along with someone else's PR.

[...]

PR #4049, opened 2026-07-30
fix(runtime)!: the /share-links dispatcher domain stops emitting a duplicate link/links beside data (#4038)

### LINKED ISSUE #4038: The /share-links dispatcher domain emits `data` AND a legacy `link`/`links` beside it — a producer-side shim that outlived its readers
Found while sweeping consumers for #3983. Not fixed there because deleting it needs a cloud-side check that is out of that PR's scope (Prime Directive #10).

## The shim

`packages/runtime/src/domains/share-links.ts` answers create and list with the payload under **two** keys:

```ts
// POST /share-links → create
return { handled: true, response: { status: 201, body: { success: true, data: link, link } } };

// GET /share-links?object&recordId → list
return { handled: true, response: { status: 200, body: { success: true, data: links, links } } };
```

Note these bypass `deps.success()` — the domain's other three routes (`revoke`, `resolve`, `messages`) go through it and emit a clean `{ success: true, data }`. Only these two hand-build a body, and the only reason to is the duplicate key.

That is a **producer-side compatibility shim**: the legacy `link` / `links` keys are there for readers written before the envelope existed. Prime Directive #12 is about deleting *consumer*-side sniffing by fixing the producer; this is the mirror image — a producer emitting two dialects at once so consumers never have to choose.

## Why it can go now (and why not yet)

#3983 removes the reason it exists. Before that PR, the sharing plugin's routes answered bare `{ link }` / `{ links }` while this domain answered enveloped, so every consumer had to tolerate both:

```ts
// objectui packages/components/src/share/ShareDialog.tsx
setLinks(body.links ?? body.data ?? []);
const newLink = created.link ?? created.data;
```

That `??` is reading the *plugin's* shape first and this domain's second. Once #3983 lands, both surfaces answer `data`, the fallback is dead code, and the duplicate key has no reader left in either repo.

**What still needs checking before deletion:** this domain is the *designed primary* surface for cloud's per-environment kernels (`registerShareLinkRoutes: false`, per the #2462 step-① note in the file header). Anything in `cloud` reading `body.link` / `body.links` off these two routes would break. I can't see that repo from this session, so the sweep has to happen there.

## Suggested shape of the fix

1. Sweep `cloud` for `body.link` / `body.links` reads against `POST|GET /share-links`.
2. Drop the duplicate keys and route both through `deps.success(...)` like the domain's other three routes — which also removes the two hand-built bodies.
3. Delete the now-dead `?? body.data` / `?? created.data` fallbacks in objectui's `ShareDialog` (they become unreachable, and leaving them implies a shape that no longer exists).
4. Add the two routes to the dispatcher's envelope coverage so a hand-built body can't come back.

Step 3 is the part worth doing even if 1–2 are deferred: a tolerance kept after its reason is gone reads as "both shapes are still live" to the next person, which is how #3833 happened.

## Scope note

`scripts/check-route-envelope.mjs` does **not** cover this file — it scans route *modules* that write via `res.json(...)`, and the dispatcher domains return `{ status, body }` objects for a central sender instead. So this drift is invisible to that guard by construction. Whether the guard should grow a second mode for the domain handlers is a bigger question than this issue; worth raising separately if the dispatcher turns out to have more of these.

os-zhuang:
Closes #4038. The producer-side other half of #3983 — and the cloud sweep that issue was waiting on came back **clean**, so this is smaller than filed.

## What the shim was

Create and list answered with the payload under **two** keys:

```ts
{ success: true, data: link,  link  }   // POST /share-links
{ success: true, data: links, links }   // GET  /share-links
```

That duplicate existed so readers predating the envelope kept working — which is exactly why objectui's `ShareDialog` reads `body.links ?? body.data`. It stayed alive because the *other* surface for these same paths (the sharing plugin's routes) answered bare, so consumers genuinely needed both.

#3983 moved that surface onto `data`. The first branch of every `??` then had no producer left, and the duplicate had no reader anywhere.

## The sweep #4038 was blocked on

| repo | result |
|---|---|
| **framework** | no consumer of these routes at all |
| **objectui** | `ShareDialog` already falls through to `body.data`; nothing else reads them |
| **cloud** | **clean** — see below |

`cloud` was the blocker, and I could reach it after all. It touches share-links in exactly two places (`objectos-runtime/src/artifact-kernel-factory.ts`, `service-cloud/src/default-environment-plugins.ts`), and both only **register** `SharingServicePlugin` into a per-environment kernel with `registerShareLinkRoutes: false` — precisely so *this* dispatcher serves the paths. Zero calls to the endpoints, zero body reads, zero `body.link` / `body.links` anywhere in the repo.

## Shape

| route | was | now |
|---|---|---|
| `POST /share-links` | `{ success, data: link, link }` | `{ success, data: link }` |
| `GET /share-links` | `{ success, data: links, links }` | `{ success, data: links }` |

`data` is unchanged on both — only the duplicate key is gone. Anything reading `body.data`, or going through `ObjectStackClient.unwrapResponse`, sees no difference at all. Only a raw reader of top-level `body.link` / `body.links` would move, and there isn't one.

The list route now goes through `deps.success(...)` like the domain's other three. **Create stays hand-built** — `deps.success` hardcodes status 200 and this route is a **201**, which is the same reason `/keys` hand-builds its own 201, and the same shape it uses. My own issue text said to route *both* through the helper; that was wrong about the status, and I've corrected it on the issue.

## Guard

`scripts/check-route-envelope.mjs` does not and **cannot** cover this file — it scans route modules that write via `res.json(...)`, while dispatcher domains return `{ status, body }` for a central sender. So this drift was invisible to that guard by construction, which is worth stating plainly: the guard's 7/0/1 is a claim about route *modules*, not about every REST body the platform emits.

Three tests in `domain-handler-registry.test.ts` cover it instead — two per-route, plus a general one asserting no success body carries a top-level key outside `success` / `data` / `meta`, so a *new* duplicate spelling fails too. Restoring the duplicates fails all three.

## What I deliberately did **not** do

#4038's step 3 said to delete objectui's `body.links ?? body.data` tolerance, and I'm declining that — correcting my own recommendation.

With both producers converged, that `??` is no longer **drift** (two live producers disagreeing, which was the actual bug); it's **version skew** (one producer changed, the consumer still supports the older one). `@object-ui/components` publishes standalone against peer deps, so dropping it turns "new console, older framework" from *working* into a **silently empty share list** — a bad failure mode to trade for removing one `??`. Deleting it wants a version gate, which is a larger call than this issue. Noted on #4038 rather than left implicit.

## Verification

| gate | result |
|---|---|
| `@objectstack/runtime` | **909 passed (65 files)** |
| `@objectstack/client` | 200 passed (14 files) |
| new tests are load-bearing | restoring the duplicates → **3 of 3 fail** |
| `check:route-envelope` + `--self-test` | 7 conformant / 0 ratcheted / 1 exempt |
| spec `tsc --noEmit` · `check:docs` · `check:skill-refs` · `check:react-blocks` | all clean |
| `check:error-code-casing` · `check:console-sha` · `check:nul-bytes` · `check:role-word` · `check:org-identifier` | all clean |
| `eslint --no-inline-config` on changed files | clean |

No `.objectui-sha` bump needed — this touches no frontend, so the console pin stays where #4037 left it (`check:console-sha` confirms it is still in sync).

---

### DIFF
--- a/.changeset/share-links-dispatcher-dual-key.md
+++ b/.changeset/share-links-dispatcher-dual-key.md
@@ -1,0 +1,54 @@
+---
+"@objectstack/runtime": patch
+---
+
+fix(runtime)!: the /share-links dispatcher domain stops emitting a duplicate `link`/`links` beside `data` (#4038)
+
+The producer-side other half of #3983. That PR moved the sharing plugin's routes
+onto the declared envelope; this removes the compatibility shim the dispatcher
+twin had been carrying *because* that surface answered bare.
+
+Create and list answered with the payload under **two** keys:
+
+```ts
+{ success: true, data: link,  link  }   // POST /share-links
+{ success: true, data: links, links }   // GET  /share-links
+```
+
+The duplicate existed so readers predating the envelope kept working — which is
+why objectui's `ShareDialog` reads `body.links ?? body.data`. Once #3983 made both
+surfaces answer `data`, that first branch had no producer left, and the duplicate
+had no reader in **any** repo:
+
+- **framework** — no consumer of these routes at all
+- **objectui** — `ShareDialog` already falls through to `body.data`
+- **cloud** — swept: it only *registers* `SharingServicePlugin` into per-environment
+  kernels with `registerShareLinkRoutes: false` so this dispatcher serves the paths.
+  It never calls them and never reads a body. That sweep is what #4038 was waiting
+  on, and it came back clean.
+
+## Shape
+
+| route | was | now |
+|---|---|---|
+| `POST /share-links` | `{ success, data: link, link }` | `{ success, data: link }` |
+| `GET /share-links` | `{ success, data: links, links }` | `{ success, data: links }` |
+
+`data` is unchanged in both — only the duplicate key is gone. Anything reading
+`body.data`, or going through `ObjectStackClient.unwrapResponse`, sees no
+difference. A raw reader of the top-level `body.link` / `body.links` must move to
+`body.data`.
+
+The list route now routes through `deps.success(...)` like the domain's other
+three. Create stays hand-built, because `deps.success` hardcodes status 200 and
+this route is a **201** — the same reason `/keys` hand-builds its own 201, and the
+same shape it uses.
+
+## Guard
+
+`scripts/check-route-envelope.mjs` does not and cannot cover this file: it scans
+route modules that write via `res.json(...)`, while dispatcher domains return
+`{ status, body }` for a central sender. So the drift was invisible to it by
+construction. Three tests in `domain-handler-registry.test.ts` cover it instead —
+two per-route, plus a general one asserting no success body carries a top-level
+key outside `success` / `data` / `meta`. Restoring the duplicates fails all three.
--- a/packages/runtime/src/domain-handler-registry.test.ts
+++ b/packages/runtime/src/domain-handler-registry.test.ts
@@ -357,6 +357,62 @@ describe('HttpDispatcher extracted domains (PR-4: share-links)', () => {
         expect(record?.secret).toBeUndefined();
     });
 
+    // ── #4038: one shape, no duplicate keys ──────────────────────────────
+    //
+    // Create and list used to answer `{ success: true, data: link, link }` /
+    // `{ …, data: links, links }` — the payload under BOTH the envelope's `data`
+    // and a legacy top-level key. That shim existed because the sharing plugin's
+    // routes (the other surface for these same paths) answered bare, so every
+    // consumer had to read `body.links ?? body.data`. #3983 moved that surface
+    // onto this shape and left the duplicate with no reader anywhere — framework,
+    // objectui, or cloud — so it is gone. These pin that: a body that grows a
+    // second spelling of its payload fails here.
+    const authed: any = { executionContext: { userId: 'u1' } };
+    const LINK = { id: 'l1', token: 't1', object_name: 'account', record_id: 'r1' };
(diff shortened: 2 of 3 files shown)


os-zhuang:
Two notes from CI on this PR.

**The Check Changeset failure is infrastructure, not the changeset.** The job died before it ran anything: `corepack` couldn't fetch pnpm from `registry.npmjs.org` — `read ECONNRESET` inside `Engine.ensurePackageManager`, exit 1 during setup. Re-queued the failed job; no change pushed, because there is nothing here to fix.

**The docs-drift advisory (18 files) needs no doc change, but one of them corroborates the premise.** I checked the ones that could plausibly document these bodies — `api/wire-format.mdx`, `protocol/kernel/http-protocol.mdx`, `api/index.mdx` all have no share-link content. `api/client-sdk.mdx` does:

```ts
const link = await client.shareLinks.create('crm_account', recordId, { … });
await client.shareLinks.revoke(link.token);
```

That documented `link.token` is exactly what #3983 fixed. Before it, `create()` against the **plugin** surface returned `{ link: … }`, so `unwrapResponse` handed back the wrapper and `link.token` was `undefined` — the published example only worked if the dispatcher happened to serve the route. It is accurate on both surfaces as of #3983, and this PR keeps it that way by removing the dispatcher's duplicate key rather than by teaching anything to read two shapes.

So the doc needed no edit — but it was independent evidence that `data` is meant to unwrap to the link row itself, which is what both PRs converged on.

---

Issue #3843, opened 2026-07-28
Envelope drift is not just service-storage: four more route modules emit bare bodies, two of them the pre-#3675 `{ error: '<string>' }`


os-zhuang:
Found while fixing #3689 (PR #3837). That issue scoped itself to `storage-routes.ts`, and #3675 covered `service-i18n` alongside it — but neither asked whether the same drift exists elsewhere. It does, and in two modules it is the *older* shape, the one #3675 already declared wrong.

## The survey

Counting `res.json(...)` call sites against the two envelope halves, on `main` at 88ef03e:

| Module | `res.json` calls | carry `success: true` | bare `{ error: '<string>' }` |
|---|---:|---:|---:|
| `packages/services/service-settings/src/settings-routes.ts` | 5 | 0 | 0 |
| `packages/services/service-datasource/src/admin-routes.ts` | 12 | 0 | **3** |
| `packages/rest/src/external-datasource-routes.ts` | 7 | 0 | **1** |
| `packages/rest/src/package-routes.ts` | 16 | 3 | **3** |
| *(fixed)* `service-storage/src/storage-routes.ts` | 2 (both helpers) | ✅ | 0 |

### What each one actually emits

**`settings-routes.ts`** — the shape #3675's changeset described: nested `error`, no `success` flag.

```ts
res.status(status).json({ error: { code, message, ...extra } });   // :50
await res.json({ manifests });                                     // :65  — bare success
await res.json({ values: result });                                // :118 — bare success
```

**`service-datasource/admin-routes.ts`** — `error` is a **string**, i.e. exactly the pre-#3675 shape, and the codes are lowercase snake (relevant to #3841):

```ts
res.status(503).json({ error: 'datasource_admin_unavailable' });   // :49
res.status(400).json({ error: 'datasource_admin_error', message }); // :52 — message a SIBLING of error
res.status(404).json({ error: 'not_found' });                      // :110
res.json({ datasources });  res.json({ drivers: … });  res.json({ tables });  // bare successes
```

A caller reading `body.error.message` gets `undefined` here and the real message from the dispatcher — the identical asymmetry #3675 opened on.

**`rest/external-datasource-routes.ts`** — same bare-string error, plus its own private success word, the `ok` that #3689 just retired from storage:

```ts
res.status(503).json({ error: 'external_service_unavailable' });        // :41
res.json({ ok: results.every((r: any) => r.ok), results });            // :100
```

[...]

Issue #3689, opened 2026-07-27
service-storage success bodies are three shapes, none carrying `success: true` — the other half of #3675


os-zhuang:
Found while fixing the error envelope in #3675 (PR #3687), and deliberately left out of it because — unlike the error fix and unlike #3636's success fix — this one is **not additive**.

## The drift

`storage-routes.ts` returns three different success shapes, none of which carries the `success` flag `BaseResponseSchema` declares and `ObjectStackClient.unwrapResponse` keys on:

| Route(s) | Body |
|---|---|
| `/upload/presigned`, `/upload/complete`, `/upload/chunked`, `…/chunk/:i`, `…/complete`, `…/progress` | `{ data: {…} }` |
| `GET /files/:fileId/url` | `{ url }` — bare, no `data` |
| `PUT /_local/raw/:token` | `{ ok: true, key }` |

Meanwhile `storage.zod.ts` declares every one of these as `BaseResponseSchema.extend({ data: … })`, i.e. `{ success: boolean, data: {…} }`. `PresignedUrlResponse` and friends are `z.infer`red from those schemas and used as the SDK's declared return types, so the declaration says `success: boolean` and the wire says nothing.

## Why it is not currently breaking

Checked before filing: the storage SDK methods `return res.json()` **raw** — they do not call `unwrapResponse` — so they hand back exactly what the server sent and nothing throws. TypeScript cannot catch it because `res.json()` is `any`. The one method that reads into the body, `getDownloadUrl`, does `(await res.json()).url` and matches the bare `{ url }` shape, so it works too.

In other words this is a live `declared ≠ actual` gap that happens to be harmless *because* the SDK never relies on the declaration. That is the same posture i18n was in before #3636 — right up until something did rely on it.

## Why it is not additive

Adding `success: true` to the `{ data }` routes **is** additive and safe (callers destructure `.data`; adding a sibling key changes nothing). The other two are not:

- `{ url }` → `{ success: true, data: { url } }` **moves the payload** and breaks `getDownloadUrl` plus objectui's `body?.url ?? body?.data?.url` reader.
- `{ ok: true, key }` → likewise, and `ok` overlaps in meaning with `success`, so it is a naming decision as much as a shape one.

So it needs a deliberate pass with the consumers in hand, not a find-replace.

## What needs deciding

[...]

Issue #3675, opened 2026-07-27
service-i18n error responses are bare `{ error }`, not the declared `{ success: false, error: {...} }` envelope


os-zhuang:
Found while ledgering the autonomously-mounted i18n surface for #3636. That PR fixed the **success** envelope; this is the other half of the same drift, left out deliberately so the fix stayed scoped to what `unwrapResponse` keys on.

## The drift

The `i18n` route group declares the `response_envelope` middleware (`plugin-rest-api.zod.ts`), and `BaseResponseSchema` (`contract.zod.ts`) defines the shape as `{ success, error?, meta? }`. The dispatcher's `/i18n` domain emits exactly that via its `error()` helper (`http-dispatcher.ts`):

```ts
body: { success: false, error: { message, code, details } }
```

`service-i18n` emits a bare object instead — five sites in `i18n-service-plugin.ts`:

| Line | Body |
|---|---|
| 170 | `{ error: error.message }` (500) |
| 179 | `{ error: 'Missing locale parameter' }` (400) |
| 185 | `{ error: error.message }` (500) |
| 195 | `{ error: 'Missing object or locale parameter' }` (400) |
| 218 | `{ error: error.message }` (500) |

So the same SDK method gets two different error shapes depending on which provider mounted the route — the exact asymmetry #3636 fixed on the success path, still live on the error path. A caller that reads `body.error.message` works against the dispatcher and gets `undefined` against service-i18n; one that reads `body.error` as a string works the other way round.

## Why it wasn't fixed in #3636

`ObjectStackClient.unwrapResponse` keys on the `success` flag, so only the success bodies were breaking the SDK's declared return type. Widening to error bodies would have been scope creep into a route-audit PR, and unlike the success fix it is not purely additive: `error` changes from `string` to `object`, so anything reading it as a string breaks.

## What needs deciding

1. **Align to the contract** — `{ success: false, error: { message, code } }` at all five sites. Correct per Prime Directive #12 (fix the producer, not the consumer), and matches the dispatcher. Needs a sweep for readers that assume `error` is a string; `storage-routes.ts` uses the same bare shape, so if this is done it should probably be done for both services at once rather than leaving a new inconsistency.
2. **Leave it** and document the bare shape as intentional for autonomously-mounted service routes — cheaper, but it means the declared `response_envelope` middleware is not what these routes actually do, which is a `declared ≠ enforced` gap of its own.

[...]

PR #4090, opened 2026-07-30
test(spec): envelopeViolations — the conformance check BaseResponseSchema cannot express

os-zhuang:
Closes a gap in my own prior work on the #3843 line: the assertion those suites lead with proves less than their comments claim.

## The overclaim

Every conformance suite opens with this, under a comment I wrote in #3972 and reused since:

```ts
// The contract itself, imported — not a restatement of it.
const parsed = BaseResponseSchema.safeParse(body);
expect(parsed.success, `body is not a BaseResponse: …`).toBe(true);
```

It isn't the contract. Demonstrably:

```ts
BaseResponseSchema.safeParse({ success: true })                     // ✅ passes
BaseResponseSchema.safeParse({ success: true, data: link, link })   // ✅ passes
```

`BaseResponseSchema` is `{ success, error?, meta? }` — it **declares no `data`** (each response type adds its own via `.extend({ data })`), and a plain `z.object` **strips** unknown keys rather than rejecting them.

So it catches the one drift it was added for — a missing or non-boolean `success`, the flag `unwrapResponse` keys on, which is what #3675 / #3689 / #3843 were about — and nothing else. That second body is precisely the duplicate-payload drift `/share-links` shipped until #4038 / #4049 removed it, and `safeParse` passed it the entire time.

What was actually catching it: assertions each suite hand-wrote alongside (`deadTopLevel`, `expectData`, the `ENVELOPE_KEYS` loop I wrote once in `domain-handler-registry.test.ts`). That works exactly as long as whoever writes the next suite remembers to.

## `envelopeViolations(body)`

Returns every departure as readable reasons; empty means conformant.

| rule | why |
|---|---|
| `success` must be a **boolean** | the flag `unwrapResponse` keys on; without it a body is handed to callers raw |
| a success body must carry `data` | `undefined` only — `null`, `[]`, `{}`, `0`, `''` are payloads, not absences |
| a failure body must carry `error` with string `code` + `message` | the nested form, not the pre-#3675 bare string |
| no top-level key outside `success`/`data`/`error`/`meta` | the **general form** of the duplicate-payload drift |

It deliberately does *not* check the shape of `data`. That's each route's own payload schema, and conflating the two is what let `SettingsNamespacePayload` describe a whole body before #3843 and only `data` after it.

Placed in `contract.zod.ts` beside the schema it completes, alongside the plain predicates `spec/api` already exports (`standardErrorCodeForHttpStatus`, `readServiceSelfInfo`). **No new package** — and `BaseResponseSchema` itself is untouched, because adding `data` or `.strict()` would hit every `.extend()` caller and every producer for no gain over fixing this in the assertions.

## Verification

Two directions, because either alone would be misleading.

**The pairing is load-bearing.** I reintroduced the exact `/share-links` drift (`{ success: true, data, link: data }`). The new assertion fails with `stray top-level key `link` — the payload belongs under `data``; the old `safeParse` line still passes. That contrast *is* the argument for the change.

**The rules aren't too strict.** Every existing suite runs unchanged and green — nothing legitimate is flagged:

| package | result |
|---|---|
| `@objectstack/spec` | **6951 passed** (incl. 14 new for the predicate) |
| `@objectstack/rest` | 505 passed |
| `@objectstack/runtime` | 914 passed |
| `@objectstack/client` | 200 passed |
| storage · settings · datasource · i18n · sharing | 220 · 189 · 154 · 62 · 225 |
| spec `tsc --noEmit` · `check:docs` · `check:skill-refs` · `check:react-blocks` | clean |
| `check:route-envelope` · `check:error-code-casing` · `check:console-sha` · `check:nul-bytes` · `check:role-word` · `check:org-identifier` | clean |
| `pnpm lint` | **0 errors in source** (see note) |

**Two local-artifact notes**, both of which look like failures and aren't:

- `pnpm lint` reports 61 errors — *all* inside `.cache/objectui-96ee72e85439/`, the gitignored console build artifact my earlier `.objectui-sha` bump left in this worktree. Zero in real source; CI checks out fresh and never sees it.
- The first `rest` run showed 8 anonymous-deny 401 failures. That's the stale-dist signature I flagged in #4037's method note — I'd rebuilt only `spec`, not its dependents. A full `turbo run build` cleared it (505/505). Same for the recurring `runtime` `datasource-autoconnect` failure: gitignored `.objectstack/data/memory-driver.json` accumulating rows across local runs, 914/914 once removed.

---

### DIFF
--- a/.changeset/envelope-violations-predicate.md
+++ b/.changeset/envelope-violations-predicate.md
@@ -1,0 +1,46 @@
+---
+"@objectstack/spec": minor
+---
+
+**`envelopeViolations` — the conformance check `BaseResponseSchema` cannot express.**
+
+Every conformance suite from the #3843 line leads with
+`BaseResponseSchema.safeParse(body)`, under a comment claiming it is "the contract
+itself, imported — not a restatement of it". That overclaimed, and the gap is
+demonstrable:
+
+```ts
+BaseResponseSchema.safeParse({ success: true })                        // passes
+BaseResponseSchema.safeParse({ success: true, data: link, link })      // passes
+```
+
+The schema declares no `data` — each response type adds its own via
+`.extend({ data })` — and a plain `z.object` strips unknown keys rather than
+rejecting them. So it catches the one drift it was added for (a missing or
+non-boolean `success`, the flag `unwrapResponse` keys on) and nothing else. The
+second body above is exactly the duplicate-payload drift `/share-links` shipped
+until #4038 / #4049 removed it, and `safeParse` passed it the whole time.
+
+`envelopeViolations(body)` returns every departure from the declared envelope as
+readable reasons, empty when conformant:
+
+- `success` must be a **boolean**
+- a success body must carry `data` (`undefined` only — `null`, `[]`, `{}`, `0`
+  and `''` are payloads, not absences)
+- a failure body must carry `error` with a string `code` and `message` — the
+  nested form, not the pre-#3675 bare string
+- no top-level key outside `success` / `data` / `error` / `meta`, which is the
+  general form of the duplicate-payload drift
+
+It deliberately does **not** check the shape of `data`: that is each route's own
+payload schema, and conflating the two is what let `SettingsNamespacePayload`
+describe a whole body before #3843 and only `data` after it.
+
+The ten conformance suites now assert it beside `safeParse`, and their comments
+say what each of the two actually proves. Reintroducing the `/share-links`
+duplicate key is caught by the new assertion and still passes the old one — which
+is the demonstration that the pairing is the point.
+
+Placed in `contract.zod.ts` beside the schema it completes, alongside the other
+plain predicates `spec/api` already exports (`standardErrorCodeForHttpStatus`,
+`readServiceSelfInfo`). No new package.
--- a/packages/plugins/plugin-sharing/src/share-link-envelope.conformance.test.ts
+++ b/packages/plugins/plugin-sharing/src/share-link-envelope.conformance.test.ts
@@ -34,7 +34,7 @@
  */
 
 import { describe, expect, it, vi } from 'vitest';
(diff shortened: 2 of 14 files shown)


os-zhuang:
**TypeScript Type Check went red on `check:api-surface`; fixed in `04debd7`.**

The gate is a ratchet on `@objectstack/spec`'s public exports, and the new predicate is an addition to it:

```
  ./api
    + envelopeViolations (function)

@objectstack/spec public API changed: 0 breaking (removed/narrowed), 1 added.
```

Correctly classified — additive, nothing removed or narrowed. Regenerated with `gen:api-surface`: one line, alphabetically beside `getAuthEndpointUrl` / `readServiceSelfInfo` / `standardErrorCodeForHttpStatus`, the predicates that module already exports.

Two things worth recording, because both are the kind of trap that reads as "my change broke something" when it isn't.

**The job runs ten steps, not the four I ran.** I found the gates by grepping a window of `lint.yml` that ended before the rest of the job. The full list:

`spec tsc --noEmit` · `check:docs` · `check:skill-refs` · `check:react-blocks` · workspace build · examples typecheck · downstream-contract typecheck · **`check:api-surface`** · `check:skill-examples` · `check:i18n` + `check:i18n-coverage`

All ten pass locally now. This is the same failure mode as #3972, where I missed `check:docs` because it lives in the typecheck job rather than the test job — I fixed the symptom then and not the habit, which was to enumerate the job rather than grep for a name.

**`gen:api-surface` cannot run under `OS_SKIP_DTS=1`.** It resolves the module symbol from `dist/index.d.ts`, so with declarations skipped it fails with `Could not resolve module symbol for . — Is the package built?`, which reads like a broken build rather than a missing flag. Rebuilding spec *with* declarations is a prerequisite for regenerating the snapshot, not an optional speed-up.

Nothing about the change itself moved — the predicate, the suites, and the comments are as reviewed.

---

Issue #4053, opened 2026-07-30
`GET /ai/agents` is SDK-addressable but unenveloped, and the SDK compensates — making it conformant without a coordinated change silently returns an empty agent list


os-zhuang:
Not a bug today. Filing it because the *fix* is the hazard: this route is one of the last unenveloped SDK-addressable surfaces, and converting it the way #3675 → #3843 → #3983 converted the others would break `client.ai.agents.list()` **silently** unless three repos move together.

Found while checking whether the dispatcher domains hide more envelope drift after #4038/#4049. They mostly don't — of 10 hand-built responses across 16 domain files, `auth.ts` (×2) is better-auth's own wire format and `mcp.ts` passes an upstream `Response` through, both legitimately outside the REST envelope, and `keys.ts` / `share-links.ts` hand-build only to get a **201** (`deps.success` hardcodes 200). This is the one left over.

## Current state — consistent, and that is why it is easy to miss

The payload is **declared**, and declared as the whole body:

```ts
// packages/spec/src/api/protocol.zod.ts:1078
export const AiAgentsResponseSchema = lazySchema(() => z.object({
  agents: z.array(AgentSummarySchema).describe('Agents this caller may chat with'),
}));
```

Both producers emit exactly that, bare — no `success`, no `data`:

- **framework** — `packages/runtime/src/domains/ai.ts:47`, the degraded fallback when no AI service is registered (open-source default; `service-ai` is Cloud/EE). Returns `{ status: 200, body: { agents: [] } }`.
- **cloud** — `service-ai`'s `buildAgentRoutes`. Its own test reads the body as the bare shape: `chatbot-features.test.ts:876` → `const body = resp.body as { agents: Array<…> }`.

So unlike `/share-links` before #3983, **the two surfaces agree**. Nothing is broken.

## The trap

The route is `disposition: 'sdk'` with `client: 'ai.agents.list'` (`cloud/packages/service-ai/src/ai-route-ledger.ts:196`), and the SDK reads it like this:

```ts
// packages/client/src/index.ts:3802
list: async (): Promise<AiAgentSummary[]> => {
  const res = await this.fetch(`${this.baseUrl}${route}/agents`);
  const body = await this.unwrapResponse<AiAgentsResponse>(res);
  return body?.agents ?? [];
},
```

[...]

PR #4124, opened 2026-07-30
fix(runtime): the /ai/agents degraded fallback answers in the declared envelope (#4053)

os-zhuang:
The framework half of #4053, and the guard's **last ratchet retires** — 0 ratcheted on both surfaces.

```
✓ Route modules      — 8 audited: 7 conformant, 0 ratcheted, 1 exempt
✓ Dispatcher domains — 16 audited: 11 helper-only, 5 declared, 0 ratcheted
```

## The finding: this did not need a three-repo batch

#4053 opened by saying the conversion "must land with cloud and the SDK together or `ai.agents.list()` silently returns `[]`". Checking the SDK properly showed that is true of **one** of the two possible conversions, not both:

```ts
private async unwrapResponse<T>(res: Response): Promise<T> {
  const body = await res.json();
  if (body && typeof body.success === 'boolean' && 'data' in body) return body.data as T;
  return body as T;
}
```

| conversion | what `unwrapResponse` returns | `list()` does `body?.agents` |
|---|---|---|
| **`data: { agents }`** (this PR) | `{ agents: [...] }` | ✅ works |
| `data: [...]` (flattened) | `[...]` | ❌ `undefined` → **`[]`** |

So with the relocation variant the SDK needs no change, and **each surface converts independently**. Cloud's `service-ai` still answers unenveloped and keeps working unchanged; objectui already reads all four shapes (objectui#2992).

## Why relocation is the right reading, not just the convenient one

#3983 set the precedent that `data` carries the payload **directly**, and following it here would have looked consistent. The distinction:

- `AiAgentsResponseSchema` is a **declared** payload schema.
- share-links' `{ links }` was an ad-hoc wrapper with no schema behind it.

So this is the **#3843 relocation** — a declared payload moving under `data` unchanged, exactly as `SettingsNamespacePayload` did — rather than a reshape. The convenient answer and the correct one coincide, which is worth stating explicitly so the next person doesn't read it as precedent-breaking.

## Why the wrong choice would have been invisible

`useAiSurfaceEnabled` gates the **entire AI surface** on `agents.length > 0`, deliberately: the route is access-filtered per caller, so it is the only signal that is both edition- and user-aware (ADR-0068). An empty catalog is therefore a **correct** answer:

| | what the user sees |
|---|---|
| seat-less user (correct) | AI surface hidden |
| CE deployment, no `service-ai` (correct) | AI surface hidden |
| flattened conversion (bug) | **AI surface hidden** |

No error, no 403, no failed request, no log. That is why the road not taken is pinned as a test rather than left as a comment: `{ success: true, data: [ASK, BUILD] }` asserts `[]`, so the cost is recorded.

## Changes

- `runtime/src/domains/ai.ts` — the fallback goes through `deps.success`
- `scripts/check-route-envelope.mjs` — `ai.ts` drops to `handBuilt: 1` (the remaining one is the AI-service passthrough) and loses its ratchet
- `packages/client/src/ai-agents-envelope.test.ts` — 5 new tests over both live shapes
- the dispatcher's own fallback test now asserts the envelope and that no top-level `agents` survives

## Verification

| gate | result |
|---|---|
| `@objectstack/runtime` | **914 passed** |
| `@objectstack/client` | **205 passed** (5 new) |
| `@objectstack/rest` · `@objectstack/spec` | 505 · 6987 |
| `check:route-envelope` + self-test | **0 ratcheted, both surfaces** |
| all ten TypeScript Type Check gates | clean |
| six `check:*` guards | clean |

**Stale-dist note again**, third time this session and the same shape: `check:api-surface` first reported `- TemporalCase`, `- TemporalRow` … "5 breaking" — none of them mine, and I added no exports in this PR. Restarting the branch onto a newer `main` leaves `packages/spec/dist` behind it; rebuilding spec clears it. Recording it because the message names *other people's* exports as breaking changes, which reads as something far worse than it is.

## What remains on #4053

Cloud's `service-ai` (`routes/agent-routes.ts:352` → `{ status: 200, body: { agents } }`). Genuinely optional now rather than blocking: it works as-is against both consumers, and converting it is a one-line change whenever that repo is being touched anyway. I've left #4053 open with that reduced to its actual scope.

---


os-zhuang:
**Conflict resolved — my wildcard-ledger fix is withdrawn in favour of #4133's, which supersedes it.**

While this PR was open, `554ff923f` ("the auth wildcard yields paths the auth service does not own, and unbreak the #4116 ledger", #4117/#4133) landed and fixed the same file. We had both removed the stale `${prefix}/storage/*` entry, so the two collided on that line.

Theirs is strictly better and I took it wholesale (`git checkout --theirs`, now byte-identical to main):

- it also fixes `${prefix}/auth/*` to actually **yield**, which mine did not touch — that entry moves from `ratchet` to `yields: true`
- its comment already credits #4087/#4112 for the storage removal, reaching the same conclusion from the other direction

So `scripts/check-wildcard-fallthrough.mjs` no longer appears in this PR's diff at all. What remains is only the six files this PR is actually about:

```
.changeset/ai-agents-fallback-envelope.md
packages/client/src/ai-agents-envelope.test.ts
packages/runtime/src/domain-handler-registry.test.ts
packages/runtime/src/domains/ai.ts
packages/runtime/src/http-dispatcher.test.ts
scripts/check-route-envelope.mjs
```

Worth noting the mechanism, since it is the third such collision on this branch: main moved seven commits during the merge, and the only textual conflict was the one place two people fixed the same defect. The two earlier ones (#4058's `/ai/agents` assertion, #4112's deletion of `domains/storage.ts` from my own guard table) merged *cleanly* and would have shipped broken — CI caught one, and #4112's author correctly updated my table for the other.

## Re-verified after the merge

Applied all three staleness lessons up front this time rather than rediscovering them: `pnpm install --frozen-lockfile`, full rebuild, and cleared `packages/runtime/.objectstack` + `.cache/`.

| | |
|---|---|
| full suite (`turbo run test`, Test Core's command) | **129/129 packages** |
| ESLint job — all **11** checks | clean |
| TypeScript Type Check job — all **10** gates | clean |

---

Issue #4085, opened 2026-07-30
`os serve <config>` cannot boot without a prebuilt `dist/objectstack.json` — dies with `Service 'manifest' is async - use await`


os-zhuang:
Found while fixing #4012 (PR #4084), where the regression test needed a temp project booted through the CLI. Unrelated to that fix, so filing rather than expanding scope (AGENTS.md Prime Directive #10).

## Symptom

`os serve objectstack.config.ts` fails during boot whenever `dist/objectstack.json` is absent beside the config:

```
  Loading objectstack.config.ts...
[StandaloneStack] artifact read FAILED: path='<cwd>/dist/objectstack.json' error=ENOENT: no such file or directory

  ✗ Service 'manifest' is async - use await
```

Exit code 1. The thrower is `packages/core/src/kernel.ts:106` — `getService` finds the service in the plugin loader as an unresolved `Promise` and refuses it.

## Repro — it is not about the fixture being minimal

The clean control is `examples/app-todo`, which boots fine normally:

```bash
cd examples/app-todo
mv dist/objectstack.json /tmp/            # the ONLY change
os serve objectstack.config.ts --port 44318
#   ✗ Service 'manifest' is async - use await     → exit 1
mv /tmp/objectstack.json dist/            # restore
os serve objectstack.config.ts --port 44318
#   ✓ Server is ready
```

So a known-good app fails purely because its compiled artifact is missing. Any freshly-authored project — `os init` and then `os serve` before a first `os compile` — hits the same wall.

## The earlier warning

With #4084's boot-diagnostics fix applied, the preceding boot WARN is now visible and points at the same root:

```
WARN [AppPlugin] failed to seed persisted package state {"error":"Service 'objectql' is async - use await"}
```

`AppPlugin` reaches for `objectql` (and later `manifest`) while those services are still async factories mid-resolution. With the artifact present something else resolves them first and the race never opens. Worth noting that this warning was itself invisible before #4084 — the failure gave no clue as to why.

## Why it matters

Config-boot is a documented, first-class path — `serve` prints `Loading objectstack.config.ts...` and its own not-found hint offers `objectstack build` as one option *among* others, not a prerequisite. The error message names neither the missing artifact nor the fix, so it reads as an internal fault rather than "run `os compile` first".

## What "fixed" looks like

[...]

Issue #4012, opened 2026-07-30
`os dev` and `os serve` swallow every plugin boot-phase log line — WARN-level boot diagnostics are invisible on both CLI entrypoints


os-zhuang:
Found while verifying ADR-0110 D5's boot inventory (#4011). The inventory turned out to have its own placement bug (fixed there), but proving that exposed a bigger, older one: **no plugin's boot-phase logger output reaches the console on either CLI entrypoint**, at any log level.

## Evidence

All on `main` (`d6bfb3d`), `examples/app-todo`, injected with a deliberately undeclared handler so a boot WARN is *guaranteed* to be emitted (verified below):

| Run | Lines | Plugin boot WARNs | `bootstrap` / `kernel:ready` traces |
|---|---|---|---|
| `os dev` (default level) | 31 | **0** | **0** |
| `os dev --log-level debug` | 2,360 | **0** | **0** |
| `os serve objectstack.config.ts` | 19 | **0** | **0** |
| Programmatic `kernel.bootstrap()` with a plain logger | — | ✅ present | ✅ present |

The debug run is the damning one: it happily streams **thousands** of *data-phase* kernel lines (`"event":"beforeFind"` × 2,524, `"event":"afterUpdate"` × 2,472 …) yet contains **zero** boot-phase lines — not `Triggering kernel:ready hook` (`kernel.ts:363`, a plain `logger.debug`), not `✅ Bootstrap complete`, not one plugin `logger.warn`. So the logger sink is live and verbose — but only from some point *after* boot. Everything a plugin says during init/start/kernel:ready goes nowhere.

The two `⚠` lines that DO appear in the default output (`⚠ Console dist not found…`, `⚠ Flows: 4 flow(s) declared but…`) are the CLI's own hand-rendered summary, not kernel logger output — which is presumably why the gap has been invisible: the summary *looks* like boot diagnostics exist.

## Why this is a real defect and not cosmetics

- `serve.ts`'s own `--log-level` help text states the contract this breaks: *"Defaults to … `warn` **so flow/hook execution failures surface (ADR-0032)**."* A flow or hook that fails during boot-phase binding surfaces nowhere.
- ADR-0110 D5's `[action-governance]` inventory — the checklist that justifies D3's no-opt-out refusal (#3987 removed the migration valve on exactly the argument that "the boot inventory names every offender") — is emitted correctly since #4011 but still invisible on `os dev`/`os serve`.
[...]

PR #4129, opened 2026-07-30
fix(cli): a config-booted app keeps its module code — action handlers reach the engine again (#4095)

### LINKED ISSUE #4095: Config-booted apps lose their `onEnable` hook — every script action handler goes unregistered and 404s at dispatch (`examples/app-todo`: all 8)
Found immediately after #4084 made boot-phase warnings visible. The ADR-0110 D5 `[action-governance]` inventory fired on the platform's own example app, and it was **right**: the handlers really are missing.

This is the D5 checklist doing exactly the job it was built for, on its first boot where anyone could see it.

## The finding

Booting `examples/app-todo` through `os serve` (or `os dev`) reports all eight of its `todo_task` script actions as unbound:

```
WARN [action-governance] declared script actions with NO handler — a button wired to
nothing (ADR-0078); add a `body`, or register a handler under the declared `target`
{"count":8,"actions":["todo_task:clone_task","todo_task:complete_task",
 "todo_task:defer_task","todo_task:delete_completed","todo_task:export_csv",
 "todo_task:mass_complete","todo_task:set_reminder","todo_task:start_task"]}
```

The example is **correctly authored**. `task.actions.ts` declares `complete_task` with `target: 'completeTask'`, `register-handlers.ts` registers `engine.registerAction('todo_task', 'completeTask', completeTask)`, and `objectstack.config.ts` exports `export const onEnable = async (ctx) => registerTaskActionHandlers(ctx.ql)`. That pairing is exactly what `resolveActionHandlerKeys` resolves, so a registered handler would reconcile.

## It is not a false positive — the action 404s

```
$ curl -sX POST http://localhost:44340/api/v1/actions/todo_task/complete_task \
       -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' -d '{}'
{"success":false,"error":{"code":"RESOURCE_NOT_FOUND",
 "message":"Action 'complete_task' on object 'todo_task' not found","httpStatus":404}}
```

Every "Mark Complete" / "Start" / "Clone" / "Export CSV" button in the example app is dead over HTTP.

## Root cause

`AppPlugin` runs, reaches the runtime-hook resolution at `packages/runtime/src/app-plugin.ts:563`, and finds nothing:

```
$ os serve objectstack.config.ts --dev --log-level debug | grep onEnable
DEBUG No runtime.onEnable function found {"appId":"com.example.todo"}
```

So the bundle `AppPlugin` receives has no `onEnable`, even though the config module exports one. Note that `serve.ts` carries explicit code for exactly this hazard:

```ts
// Preserve module-level named exports (e.g. `onEnable`, `onDisable`
// lifecycle hooks) that would otherwise be dropped when we unwrap
// `mod.default`. Without this AppPlugin can never invoke runtime hooks
// declared as `export const onEnable = ...` alongside the default
// `defineStack(...)` export.
```

…and `app-plugin.ts:559-565` carries a matching fallback for the same reason. Both guards exist, and the hook still does not arrive — so the loss is somewhere between them. The `[StandaloneStack] artifact read` path is the obvious suspect: `dist/objectstack.json` is pure JSON metadata and cannot carry a function, so if the bundle handed to `AppPlugin` is derived from the artifact rather than the loaded module, metadata (the 8 declarations) survives and code (the handlers) does not — which is precisely the observed split.

## Repro

```bash
cd examples/app-todo
os serve objectstack.config.ts --port 44340 --dev
# banner now shows the [action-governance] WARN naming all 8 (needs #4084)
# then, with a session token:
curl -sX POST localhost:44340/api/v1/actions/todo_task/complete_task \
     -H "Authorization: Bearer $TOKEN" -d '{}'   # → 404
```

## Why this matters

- **`declared ≠ enforced` at the call site** (Prime Directive #10): the app declares eight executable actions and the runtime serves none of them. The showcase demonstrates a capability that does not work.
- It is silent without #4084. The 404 only appears if someone clicks the button; nothing at boot said the handlers were missing, which is how this survived.
- ADR-0110 D3 refuses undeclared handlers at dispatch with no opt-out. The mirror case — declared, handler dropped by the host — currently fails the same way at the same place, with the operator having done nothing wrong.

## Scope check worth doing

I only verified the `os serve <config>` path on `examples/app-todo`. Worth confirming whether the compiled-artifact path (`os build` → `os start`), the `os dev` fast path, and `app-showcase` / `app-crm` behave the same, since they share `AppPlugin`.

os-zhuang:
Closes #4095.

## The loss

`os serve <config>` calls `createStandaloneStack()`, which reads `dist/objectstack.json` and returns a **ready-made `AppPlugin`** for the app. That satisfied serve's "does the host already wrap itself?" guard, so the `new AppPlugin(config)` built from the **loaded module** — the only one carrying the module's `onEnable` — was skipped. A JSON artifact cannot hold a function, so the app booted with all of its metadata and none of its code.

Two probes nail it. First, what `createStandaloneStack()` actually returns once the artifact resolves:

```
PLUGINS: 4
 - com.objectstack.runtime.default-datasource   isApp=false
 - com.objectstack.metadata                     isApp=false
 - com.objectstack.engine.objectql              isApp=false
 - plugin.app.com.example.todo   ctor=AppPlugin  isApp=true
   bundle.onEnable = undefined      ← the hook that registers the handlers
   bundle.actions  = 8              ← all eight declarations
```

Second, serve at the branch that discards it:

```
[PROBE] hasAppPluginAlready=true  configHasMetadata=true  config.onEnable=function
```

The hook survives the entire config chain — `mod.default || mod`, the named-export preservation, the docs graft, both `mergeBootConfig` calls — and is then dropped by the guard. Exactly the observed split: **metadata survives, code does not.**

`examples/app-todo` is correctly authored — it declares `target: 'completeTask'`, registers `todo_task:completeTask`, and exports `onEnable`. The ADR-0110 D5 inventory had been naming all eight actions since #4012 made boot warnings visible, and it was right.

#4110 has since landed on this same seam from the other side, and its own comment corroborates the diagnosis — *"the artifact path never hit it only because `createStandaloneStack` appends ITS AppPlugin after the engine"*. It fixed the ordering; the code loss is untouched by it (see "Re-verified against the new base" below).

## The fix

Serve grafts the module's executable members onto the app bundle already registered, instead of dropping them along with the wrap.

- **Only what `AppPlugin` executes travels** — `onEnable` and the `functions` map that string-named hook/job handlers resolve against. `onDisable` is deliberately excluded: it is declared in `packages/spec` but `grep` finds no kernel, runtime or service that ever calls it, so grafting it would wire a hook nothing runs.
- **The artifact stays the metadata source of truth.** Neither side is a superset — the artifact carries compile-time enrichment the config never has (ADR-0046 packaged docs, which serve already grafts the *other* way), so this is a graft, not a swap. Code only; never metadata.
- **Targeting is by `manifest.id`**, so a host composing several `AppPlugin`s can never have one app's handlers attached to another. With no id to match it falls back to the single app bundle present, and **refuses rather than guess** when there are several.
- **A bundle's own value always wins**, so a host that wrapped itself on purpose is untouched.
- **Orphaned code is now loud.** A config whose hooks find no bundle gets a boot warning naming the consequence ("they 404 at dispatch") instead of vanishing — that silent drop is what hid this for so long.

`isAppPluginLike` is shared between the guard and the graft, so the decision to skip the wrap and the compensation for skipping it cannot drift apart.

## Verification — the user-visible thing actually works

Live HTTP against a real `examples/app-todo` boot:

| | before | after |
|---|---|---|
| `POST /api/v1/actions/todo_task/complete_task` | `404 RESOURCE_NOT_FOUND — Action 'complete_task' on object 'todo_task' not found` | `{"success":true}` |
| `POST …/export_csv` | same 404 | `{"success":true,"data":"subject,status,priority,…\nLearn ObjectStack,completed,…"}` — real seeded rows |
| `[action-governance]` boot warning | 1, naming all 8 | **0** |

The intermediate state was informative too: with an empty body the handler returned `400 VALIDATION_ERROR — Update requires an ID`, i.e. the failure moved from the router into the handler's own logic.

## Tests

- **14 unit cases** on the graft, weighted toward the refusals — id mismatch, ambiguity, never-overwrite, metadata-must-not-travel, `onDisable`-must-not-travel, and the degenerate inputs a boot can hand it.
- **1 end-to-end case** boots a real stack through `bin/run-dev.js` with a fixture whose action is executable *only* if `onEnable` ran, and asserts the inventory reports nothing. It also asserts the boot reached its banner, so "no warning" cannot pass vacuously. It uses the shared `test/helpers/serve-process.ts` harness #4110 extracted rather than carrying a second copy.

Its `os compile` step is retained, with its justification inverted: it was originally a workaround for config-boot being unable to start without an artifact (#4085, now fixed), and it is kept because the artifact is what makes `createStandaloneStack()` contribute the artifact-derived `AppPlugin` this regression is about. Without it, serve wraps the config directly, `onEnable` runs for the trivial reason, and the test would pass while exercising nothing.

## Re-verified against the new base

`main` moved by 12 commits mid-review (#4110, #4133, #4117, #4120, #4114, #4136 …), including two that touch this exact area, so the whole thing was re-checked rather than assumed:

- Merged `main` in at `001c01e`.
- **The bug is still live.** With `packages/cli/src/commands/serve.ts` taken verbatim from `origin/main`, the e2e fails with `the action's handler went unregistered`. #4110 fixed the plugin **ordering**; the code **loss** is a different branch of the same `if`, and remains.
- Full CLI suite **88 files / 909 tests** green; `eslint src` clean; `check:wildcard-fallthrough` green (`7 yielding / 0 ratcheted / 5 exempt`).

I earlier carried a two-line fix here for the wildcard guard that was failing on `main` — **that is now dropped**: #4133 fixed the same stale `MOUNTS` entry independently and better (it records *why* the storage twin disappeared), so the merge takes main's version and my commit contributes nothing. Nothing in this PR touches `scripts/` any more.

## Relationship to #4085 — checked, and they were not the same bug

I verified this before #4085 was fixed: `os serve <config>` could not boot without `dist/objectstack.json`, and that reproduced **unchanged with this fix applied** — a separate defect on the other side of the same fork. #4110 has since fixed it, from the ordering angle, which confirms the split. The verdict and the lead I left are [on #4085](https://github.com/objectstack-ai/objectstack/issues/4085#issuecomment-5129835786).


os-zhuang:
**Blocked on a red `main`, not on this diff.** The `ESLint` check is failing, but the failing step inside that job is not ESLint — it is the new wildcard fall-through guard:

```
✗ wildcard fall-through guard (#4116)

  packages/adapters/hono/src/index.ts:all `${prefix}/storage/*`
    DECLARED but not found by the scan. Moved, renamed or deleted? Update MOUNTS.
```

Verified on **pristine `origin/main`** with this branch nowhere in the picture — same single failure:

```bash
git worktree add /tmp/main-check origin/main
cd /tmp/main-check && node scripts/check-wildcard-fallthrough.mjs
#   ✗ wildcard fall-through guard (#4116)  →  1 problem(s)
```

## Cause — a merge-order collision between two landed PRs

- #4112 (`4be9d99`) **retired** the `/storage` dispatcher bridge, deleting the `app.all(prefix + '/storage/*')` mount. `packages/adapters/hono/src/index.ts:351` now carries a comment describing the removal.
- #4122 (`3df8819`) landed **after** it and added `scripts/check-wildcard-fallthrough.mjs`, whose `MOUNTS` table still declares that mount at line 158:

```js
'packages/adapters/hono/src/index.ts:all `${prefix}/storage/*`': {
  ratchet: '#4117 — terminal, same shape as #4088; adapter has no in-repo consumer',
},
```

Each was green on its own base; together the declaration is stale. Since the guard runs on every PR, this currently blocks all of them, not just this one.

## Why I have not fixed it here

The remedy is the two-line deletion the guard's own message asks for, but it is in another change's guard and there is a real choice behind it — drop the stale `MOUNTS` entry, or have the guard tolerate a declared-then-deleted mount rather than fail. That belongs to whoever owns #4116/#4117, so I am not making that call inside an unrelated CLI fix. Happy to push the one-line `MOUNTS` deletion if that is the wanted shape — just say so.

This branch has `main` merged in as of `6a381db`, and everything under this PR's own scope is green against it: **87 files / 906 tests** in `@objectstack/cli`, `eslint src` clean, and the end-to-end proof re-verified (`complete_task` → `{"success":true}`).

---

PR #4139, opened 2026-07-30
fix(tooling): drop the retired /storage mount from the wildcard gate — main is red

os-zhuang:
**`main` is currently red, and every open PR inherits it.** Three lines.

## What happened

#4122 added the wildcard fall-through guard with a `MOUNTS` ratchet declaring:

```js
'packages/adapters/hono/src/index.ts:all `${prefix}/storage/*`': {
  ratchet: '#4117 — terminal, same shape as #4088; adapter has no in-repo consumer',
},
```

#4112 (*"retire the /storage dispatcher bridge — it never spoke the storage contract"*) had already removed that mount. What survives at `packages/adapters/hono/src/index.ts:351` is a comment about it:

```ts
// This used to be `app.all(prefix + '/storage/*')` feeding …
```

The two PRs landed in an order that left the declaration pointing at nothing, so the guard fails:

```
✗ wildcard fall-through guard (#4116)

  packages/adapters/hono/src/index.ts:all `${prefix}/storage/*`
    DECLARED but not found by the scan. Moved, renamed or deleted? Update MOUNTS.

1 problem(s).
```

That is the `ESLint` job, which runs `pnpm check:wildcard-fallthrough`. Reproduced on a pristine `origin/main` worktree, and on an unrelated PR branch (objectstack-ai/objectstack#4135, which touches only `packages/spec/src/ui/**`) — this is not caused by any open change.

## The fix

Remove the entry, which is exactly what the guard's own message asks for. Nothing is being un-ratcheted: the mount it described no longer exists, so there is no fall-through left to classify.

## Verification

```
✓ wildcard fall-through: 6 yielding / 1 ratcheted / 5 exempt (12 namespace-claiming mounts)
```

The script's own 15-case self-test (`--self-test`) still passes, and it is the first half of the CI command, so both halves are covered.

No changeset — tooling-only, no published package affected.

Refs #4116, #4122, #4112

### DIFF
--- a/.changeset/unbreak-wildcard-gate.md
+++ b/.changeset/unbreak-wildcard-gate.md
@@ -1,0 +1,4 @@
+---
+---
+
+fix(tooling): drop the retired `/storage` mount from the wildcard fall-through gate (#4116). #4122 added the guard declaring `packages/adapters/hono/src/index.ts:all ${prefix}/storage/*` in its `MOUNTS` ratchet; #4112 had already retired that mount, so the declaration pointed at nothing and the ESLint job failed on `main` — every open PR inheriting a red check it did not cause. Removing the entry is what the guard's own message asks for; nothing is un-ratcheted, because the mount it described no longer exists. Tooling only; releases nothing.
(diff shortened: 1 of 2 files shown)


os-zhuang:
Superseded by #4133, which landed while this was in CI and fixed the same red gate properly.

I deleted the stale `MOUNTS` entry; #4133 instead made the auth wildcard yield the paths the auth service does not own and **reclassified** the storage mount rather than dropping it. The gate now reports `7 yielding / 0 ratcheted / 5 exempt` on `main` — no ratchet left at all, which is a better end state than my version's `1 ratcheted`.

Verified green on current `main` (`554ff923f`) from an unrelated branch. Closing unmerged; nothing here is worth keeping.

PR #4166, opened 2026-07-30
fix(cli,service-storage): stop re-pointing the storage adapter every boot — and mean it when it warns (#4096)

### LINKED ISSUE #4096: StorageServicePlugin warns "storage adapter swapped (LocalStorageAdapter → LocalStorageAdapter) … files may be unreachable" on every clean boot
Newly visible after #4084 unblocked boot-phase logging. Every `os dev` / `os serve` boot of `examples/app-todo` prints:

```
WARN StorageServicePlugin: storage adapter swapped (LocalStorageAdapter → LocalStorageAdapter).
Existing files were NOT migrated and may be unreachable through the new adapter.
```

The adapter was swapped for **the same adapter class**, and the warning claims a data-reachability risk that did not occur. It fires on a clean boot where nothing changed.

## Mechanism

`packages/services/service-storage/src/storage-service-plugin.ts`:

- `applySettings()` reads the `storage` settings namespace and bails only when *every* value is empty:
  ```ts
  const hasAny = Object.values(values).some((v) => v !== undefined && v !== null && v !== '');
  if (!hasAny) return;
  const next = await this.buildAdapterFromValues(values);
  this.storage.swap(next);           // line 335 — unconditional
  ```
- `swap()` then invokes the `onSwap` callback registered at line 172, which warns unconditionally:
  ```ts
  const prevName = (prev as any)?.constructor?.name ?? 'unknown';
  const nextName = (next as any)?.constructor?.name ?? 'unknown';
  ctx.logger.warn(`StorageServicePlugin: storage adapter swapped (${prevName} → ${nextName}). ` +
    'Existing files were NOT migrated and may be unreachable through the new adapter.');
  ```

Once the settings service has persisted defaults for the `storage` namespace, `hasAny` is true on every subsequent boot, so a same-kind, same-config adapter is rebuilt and swapped in, and the warning fires — forever.

## Why it is worth fixing rather than ignoring

The warning is correct and valuable for a *real* adapter change (Local → S3), which is genuinely a data-reachability event. Crying wolf on every boot is how an operator learns to skim past it — and this one now sits in the boot-diagnostics block right where real warnings land.

## Suggested fix

Skip the swap when the rebuilt adapter is equivalent to the current one (same class **and** same resolved config), so `onSwap` only fires on an actual change. Comparing constructor names alone is not enough — Local→Local with a different `rootDir` *is* a real migration hazard and must still warn.

os-zhuang:
Closes #4096.

## The issue's premise was half right — the warning was telling the truth

#4096 reports this on every clean boot and calls it spurious:

```
WARN StorageServicePlugin: storage adapter swapped (LocalStorageAdapter →
LocalStorageAdapter). Existing files were NOT migrated and may be unreachable
through the new adapter.
```

I built the "nothing changed, so don't swap" fix first — and the warning **survived it**. A probe at the decision point said why:

```
values.local_root = "./.objectstack/data/uploads"
current  = local://./storage                        ← what the plugin was running
next     = local://./.objectstack/data/uploads      ← what settings said
```

The backing store really was moving on every boot. The adapter was starting in the **wrong directory** and being corrected seconds later, and the warning was an accurate report of a real misconfiguration nobody could act on.

## Root cause: option keys nothing reads

`serve` constructed the plugin with `{ driver: 'local', root }`. `StorageServicePluginOptions` declares **neither key** — it takes `{ adapter, local: { rootDir } }`. Both were dropped on the floor, so:

- the plugin fell back to its own `./storage` default;
- **`OS_STORAGE_ROOT` did nothing** — a documented env var with no effect;
- uploads landed in `./storage`, not the `.objectstack/data/uploads/` the code comment promises;
- the `storage` settings namespace then corrected the root on its first read (its manifest default at `storage.manifest.ts:38` *is* `./.objectstack/data/uploads`), swapping the adapter and warning — forever.

A plain object literal passed to a parameter without an index signature type-checks fine and does nothing at runtime. Nothing was going to catch this except reading it.

## Three defects, three fixes

1. **`serve` passes what the plugin reads** — `{ adapter: 'local', local: { rootDir } }`. Extracted as `resolveStorageCapabilityArg` so the option **shape** is pinned by tests, including an explicit `not.toHaveProperty('driver'|'root')`.
2. **A swap is skipped when nothing changed.** The plugin records what the running adapter points at and compares resolved *configurations*, instead of rebuilding whenever the settings namespace held any value — true on every boot once that namespace persists its own defaults.
3. **The warning means what it says.** It fires when the **backing store moved** (kind change, different root, different bucket/region/endpoint) — not merely when the adapter object was replaced. A credential rotation swaps so the new key takes effect and logs at `info`: same bucket, nothing stranded. A `swap()` from a caller that resolved no target **still warns** — ignorance must not silence the one message that exists to be impossible to miss.

Path spellings are normalised, so the platform writing one default two ways (`./.objectstack/data/uploads` in the settings manifest, `.objectstack/data/uploads` in the CLI) is not read as a migration between a directory and itself.

The fingerprint that decides "did anything change" includes S3 credentials and is **never logged**; the `location` that goes into the warning text deliberately excludes them, and a test asserts it.

## Verification

`examples/app-todo`, real boot:

| | before | after |
|---|---|---|
| Boot-diagnostics warnings | 4 | **3** — storage line gone |
| `adapter swapped` occurrences | 1 | **0** |
| `./storage` created | yes | **no** — uploads go to the intended root |

## Tests

- **19** unit cases on the target resolver and the swap/warn split, weighted toward the distinctions that matter: credential rotation (swap, don't warn), `basePath` change (URLs move, bytes don't), kind change, same bucket on a different endpoint, `./x` vs `x`, and the unknown-caller case that must still warn.
- **4** plugin-level cases pinning what a boot *does* and *says* — including that a quiet boot does not disarm the next genuine migration.
- **7** on the CLI option shape.

Confirmed to fail against the pre-fix code (3 plugin cases fail, one with the exact `adapter swapped (LocalStorageAdapter → LocalStorageAdapter)` text). `service-storage` 19 files / **241** tests; `cli` 89 files / **917** tests; `eslint` clean; `tsc` error count unchanged from `main` (42 pre-existing, none added).

## Left alone deliberately

`config.storage` authored with the `driver`/`root` dialect is still forwarded verbatim and still not read by the plugin — the same mismatch one layer up, so a host writing `{ driver: 's3', … }` silently gets local disk. Fixing it means deciding whether the plugin accepts that dialect or the config schema is wrong; a lenient alias here would fossilize the wrong contract (AGENTS.md Prime Directive #12), so it is filed separately rather than papered over. The helper's doc comment says so at the call site.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HrRNgrWaRtggzmrHpbomyh)_

### DIFF
--- a/.changeset/storage-adapter-swap-verdict.md
+++ b/.changeset/storage-adapter-swap-verdict.md
@@ -1,0 +1,58 @@
+---
+'@objectstack/service-storage': patch
+'@objectstack/cli': patch
+---
+
+**The storage adapter stops being rebuilt and re-pointed on every boot, and the
+"files may be unreachable" warning stops firing at a healthy server (#4096).**
+
+Every `os dev` / `os serve` boot printed:
+
+```
+WARN StorageServicePlugin: storage adapter swapped (LocalStorageAdapter →
+LocalStorageAdapter). Existing files were NOT migrated and may be unreachable
+through the new adapter.
+```
+
+The warning was telling the truth. `serve` constructed the plugin with
+`{ driver: 'local', root }` — and `StorageServicePluginOptions` declares
+neither key. Both were dropped silently, so the plugin applied its own
+`./storage` default, `OS_STORAGE_ROOT` changed nothing, and uploads landed in a
+directory nobody named. The `storage` settings namespace then corrected the root
+on its first read (its manifest default is `./.objectstack/data/uploads`),
+genuinely moving the backing store — every boot, forever.
+
+Three fixes, because there were three defects:
+
+- **`serve` now passes options the plugin reads** — `{ adapter: 'local',
+  local: { rootDir } }`. `OS_STORAGE_ROOT` takes effect, and local uploads land
+  under `.objectstack/data/uploads` from the first byte instead of `./storage`.
+  Extracted as `resolveStorageCapabilityArg` so the option SHAPE is pinned by
+  tests: a mismatch like this type-checks fine and does nothing at runtime.
+- **A swap is skipped when nothing changed.** The plugin records what the
+  running adapter points at and compares resolved configurations, instead of
+  rebuilding whenever the settings namespace held any value at all — which is
+  every boot once that namespace has persisted its own defaults.
+- **The warning now means what it says.** It fires when the BACKING STORE moved
+  (kind change, different root, different bucket/region/endpoint), not merely
+  when the adapter object was replaced. A credential rotation swaps the adapter
+  so the new key takes effect and logs at info: same bucket, nothing stranded.
+  A swap from a caller that resolved no target still warns — ignorance must not
+  silence it.
+
+Path spellings are normalised, so the platform writing the same default two ways
+(`./.objectstack/data/uploads` in the settings manifest,
+`.objectstack/data/uploads` in the CLI) is no longer read as a migration between
+a directory and itself.
+
+Verified on `examples/app-todo`: the boot-diagnostics block went from four
+warnings to three, with the storage line gone and `./storage` no longer created.
+19 unit cases cover the target resolver and the swap/warn split (including the
+refusals), 4 plugin-level cases pin what a boot does and says, and 7 pin the CLI
+option shape.
+
+`config.storage` authored with the `driver`/`root` dialect is still forwarded
+verbatim and still not read by the plugin — the same mismatch one layer up.
+Correcting it means deciding whether the plugin accepts that dialect or the
+config schema is wrong, so it is filed rather than papered over with a lenient
+alias here (AGENTS.md Prime Directive #12).
--- a/packages/cli/src/commands/serve-storage-capability.test.ts
+++ b/packages/cli/src/commands/serve-storage-capability.test.ts
@@ -1,0 +1,81 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * framework#4096 — what `StorageServicePlugin` is actually constructed with.
+ *
+ * The fallback used to be `{ driver: 'local', root }`, and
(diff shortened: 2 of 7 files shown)


os-zhuang:
Checked the docs-drift advisory rather than waving it through, and it turned up something that belongs in the record: **no doc needs changing, because the docs were already right and the code was wrong.** Three of them describe exactly the behaviour this PR restores.

| doc | what it promises | what actually happened before this PR |
|---|---|---|
| `content/docs/deployment/environment-variables.mdx:54` | ``OS_STORAGE_ROOT`` … default `./.objectstack/data/uploads`, *"used by `os serve`'s default `storage` capability wiring"* | the value was discarded — the var had no effect at all |
| `content/docs/deployment/backup-restore.mdx:79` | *"The local storage adapter keeps uploads under `OS_STORAGE_ROOT` (default `./.objectstack/data/uploads`)"* | uploads landed in `./storage` |
| `content/docs/deployment/backup-restore.mdx:22` | `.objectstack/…/data/uploads/` → *"Include in file backups on single-host deployments"* | that directory was never written to, so following the documented procedure **backed up none of the uploaded files** |

The third row is the one worth pausing on: a documented backup procedure that silently captured nothing, on the exact deployment shape it is written for. The only visible symptom was the boot warning in #4096 — which, as the PR description covers, was accurate and was reporting this.

That also settles the direction question. `{ adapter, local: { rootDir } }` is not a preference: it is what the documented contract requires, so fixing the producer (as here) rather than teaching the plugin the `driver`/`root` dialect is the correct half to move. The remaining `config.storage` mismatch is #4167, and this is a further argument for resolving that one as "the plugin's shape is canonical, reject the other dialect at authoring" rather than aliasing it.

No documentation edits in this PR — the fix makes the existing pages true.

---

Issue #3810, opened 2026-07-28
Flow node filters silently blank date macros: the template engine consumes `{…}` before the query engine sees it


os-zhuang:
Found while implementing #3582 (PR #3809). Out of scope there; filing rather than widening that PR.

## What happens

A flow node's `config.filter` is run through the flow template interpolator (`service-automation/src/builtin/template.ts`) before it reaches the data engine. That interpolator owns `{…}` in this position, and a whole-string token it cannot resolve becomes `undefined`:

```ts
// find_records node
config: {
  objectName: 'opportunity',
  filter: { close_date: { $gte: '{current_year_start}' } },
}
```

`interpolateString('{current_year_start}', …)` takes the single-token fast path, finds no flow variable of that name, and returns `undefined`. The engine then receives `{ close_date: { $gte: undefined } }`.

`{30_days_ago}` fails the same way by a different route: it does not match the dotted-path grammar, so it falls into the arithmetic branch, `_days_ago` is substituted with `null`, and `new Function('return (30null)')` throws — also `undefined`.

Either way the bound silently disappears. No warning at registration, none at run time.

## Why it matters now

Before PR #3809 nothing resolved filter placeholders server-side at all, so this was one instance of a general gap. That PR makes `{current_year_start}` / `{current_user_id}` work everywhere else on the server — the ObjectQL read path and the analytics dataset executor — which leaves flow filters as the one server-side surface where a filter token is still silently dropped. That inconsistency is worse than the uniform gap was: an author who has just learned the tokens work will reasonably try one in a flow.

A one-line caveat now sits in `skills/objectstack-query/rules/filters.md`, but documentation is a poor substitute for a diagnostic here — the failure mode is exactly the silent-zero class #3574 and #3582 were about.

## Repro

```ts
defineFlow({
  name: 'ytd_deals',
  nodes: [
    { id: 'start', type: 'start', config: { triggerType: 'manual' } },
    { id: 'q', type: 'find_records', config: {
        objectName: 'opportunity',
        filter: { close_date: { $gte: '{current_year_start}' } },
    } },
  ],
  edges: [{ id: 'e1', source: 'start', target: 'q' }],
});
```

Expected: rows closing this year. Actual: the bound is dropped, every row matches, no diagnostic.

## Options

Two dialects meet in this one config slot, so the decision is which owns it — worth deciding deliberately rather than by accident of evaluation order:

[...]

Issue #3795, opened 2026-07-28
[P2] `protocol.ts` implements 4 of the 5 documented RPC alias precedences backwards — and disagrees with `http-dispatcher.ts` on three of them


os-zhuang:
Found by the alias sweep #3754 called for ("the remaining `@deprecated`/alias declarations in `packages/spec/src` deserve the same audit — the ones that fold-and-drop are safe, the ones that merely coexist are latent #3713s"). This is the sweep's result: one cluster is not merely latent, it is **live and inverted**.

## The contract

`RpcQueryOptionsSchema` accepts five legacy aliases alongside their canonical keys, and states the precedence **in prose only** — there is no transform folding them:

```ts
// packages/spec/src/data/data-engine.zod.ts (RpcQueryOptionsSchema doc block)
 * **Precedence:** When both legacy and new keys are present for the same
 * concern, the protocol normalizer uses the new key (`where` > `filter`,
 * `fields` > `select`, `orderBy` > `sort`, `offset` > `skip`,
 * `expand` > `populate`). Callers should not mix vocabularies.
```

Because the fold lives nowhere in the schema, every reader re-implements it — the exact condition #3713 described. There are two readers, and they do not agree.

## What the two normalizers actually do

| pair | spec says | `runtime/src/http-dispatcher.ts` | `metadata-protocol/src/protocol.ts` |
|---|---|---|---|
| `where` > `filter`/`filters` | canonical | `where ?? filter ?? filters` ✅ | **`filter ?? filters ?? $filter ?? where`** ❌ |
| `fields` > `select` | canonical | `select != null && fields == null` ✅ | **`select` clobbers `fields`** ❌ |
| `offset` > `skip` | canonical | `skip != null && offset == null` ✅ | **`skip` clobbers `offset`** ❌ |
| `expand` > `populate` | canonical | — | **`populate` wins, `expand` only if empty** ❌ |
| `orderBy` > `sort` | canonical | — | `orderBy ?? sort` ✅ |

Four of five inverted in `protocol.ts`; three of them are pairs where the two normalizers do the **opposite** thing.

### The four inversions

All in `packages/metadata-protocol/src/protocol.ts`. Canonical `where` is consulted **last**:

```ts
const filterValue = options.filter ?? options.filters ?? options.$filter ?? options.where;
```

`skip` overwrites a caller's `offset` unconditionally:

```ts
if (options.skip != null) {
    options.offset = Number(options.skip);   // no `offset == null` guard
[...]

Issue #3713, opened 2026-07-27
[P2] action `execute` vs `target`: spec and objectui resolve the precedence in OPPOSITE directions


os-zhuang:
Found while re-verifying the liveness ledger's `action.execute` entry (#3711). Not a ledger problem — a real behavioural divergence.

## The divergence

`execute` is the deprecated alias of `target`. Both sides handle "author declared both" — **in opposite directions**.

**Spec keeps `target`** — `packages/spec/src/ui/action.zod.ts:576-581`:
```ts
}).transform((data) => {
  // Auto-migrate deprecated `execute` → `target` for backward compatibility
  if (data.execute && !data.target) {
    return { ...data, target: data.execute };
  }
  return data;
})
```
Pinned by a test — `packages/spec/src/ui/action.test.ts:1032`, *"should preserve target over execute when both are set"* → asserts `result.target === 'preferredHandler'`.

**objectui keeps `execute`** — `packages/core/src/actions/ActionRunner.ts:704`, inside `executeScript()`:
```ts
const script = action.execute || action.target;
```

## Consequence

An action declaring both:
```ts
defineAction({ type: 'script', target: 'preferredHandler', execute: 'legacyHandler', … })
```
runs **`preferredHandler` server-side** (the parsed metadata carries `target`) and **`legacyHandler` client-side** (ActionRunner prefers `execute`). Two different scripts for one action, with no error anywhere.

Low frequency — it needs an author to set both, which only happens mid-migration or by copy-paste — but it is silent and the failure is "the wrong code ran", which is the worst kind to debug.

## Suggested fix

Align objectui on the spec's rule (`target || execute`), since the spec's direction is the documented one (`execute` is the deprecated alias) and is test-pinned. Then either:
- keep both readers permanently, or
- have the spec transform **drop** `execute` from its output once it has been lowered, so the ambiguity cannot reach a renderer at all — the same "canonical wins, alias disappears" shape used for `agent.knowledge.topics` → `sources` (#1891), which asserts `'topics' in parsed === false`.

The second is stronger: it makes the divergence unrepresentable rather than merely agreed-upon.

## Verification notes

Both sides read directly, not via the agent report that surfaced this. The ledger entry for `action.execute` (corrected in #3711) now cites both readers and records this divergence.

Refs #3711, #3686.

Issue #2982, opened 2026-07-15
Security: bulk (multi) update/delete skips OWD owner scoping on private objects — members can modify others' rows


os-zhuang:
Surfaced by the ADR-0096 execution-surface audit. A verb-coverage gap on the sharing axis: bulk writes are not owner-scoped on objects that rely on `sharingModel:'private'` (OWD) rather than an authored RLS policy.

## Detail

For single-id writes, the sharing plugin runs `canEdit` and the security plugin does a pre-image row-level RLS check. Both are keyed off a single inferrable id:

- `packages/plugins/plugin-sharing/src/sharing-plugin.ts:387-412` — the write gate runs `canEdit` only `if (id != null)` (`:391`).
- `packages/plugins/plugin-security/src/security-plugin.ts:811-821` — the pre-image RLS check fires only for a single `extractSingleId` (`:821 if (targetId != null)`).

A `update({ multi: true })` / `deleteMany` still routes through the `update`/`delete` middleware, and **authored RLS `where`-policies are injected into the AST operation-agnostically** (`security-plugin.ts:1167`) — so objects protected by an explicit RLS policy are fine. But objects whose protection is **OWD `private` (owner-only) with no authored RLS policy** get their owner scoping **only** from the sharing plugin's `canEdit` path, which never injects an owner `where` filter for writes and is skipped for multi-writes.

**Impact:** on a pure-`private` object, a member can `update`/`delete` with `multi:true` and affect **other users' rows** — the owner boundary that holds for single-row writes is absent for bulk writes.

## Suggested fix

For `multi` writes on objects whose effective sharing model is owner-scoped, inject the owner (and share/hierarchy) predicate as a `where` filter into the bulk operation — the write analogue of the read filter `sharing-service.ts buildReadFilter` already produces — so bulk writes are constrained to rows the caller may edit, or deny multi-writes on such objects. Add a dogfood proof (a member bulk-updating a peer's private-object rows must fail).

## Relationship to ADR-0096

Class B (a verb outside a check's coverage, not a missing-identity fall-open), but same theme — an execution/verb surface whose authorization story wasn't enumerated. Candidate row for the ADR-0096 D4 conformance matrix (bulk-write × sharing axis). Tracking under the ADR-0096 evidence base.

Refs: ADR-0096 (PR #2975).

PR #4349, opened 2026-07-31
fix(objectql): every engine verb folds `filter` → `where` — `delete({filter})` no longer empties the object (#4346)

### LINKED ISSUE #4346: [P1] `filter` folds to `where` in `engine.find` only — `findOne`/`count`/`update`/`delete` silently match EVERY row, and the hook docs teach the broken call
Found while doing the #3795 follow-up sweep on the sixth alias pair (`top`/`limit`). The `filter` → `where` fold that #3795 settled at the protocol layer exists at the **engine** layer too — in exactly one of six methods.

`ObjectQL.find()` folds it, with a comment saying why:

```ts
// packages/objectql/src/engine.ts (find)
// Normalize the `filter` alias → `where`. The DataEngine contract
// (`spec/data/data-engine.zod.ts`) exposes both, but the driver AST only
// understands `where`; an internal caller passing `{ filter }` would
// otherwise match ALL rows (silent over-grant — surfaced by ADR-0057's
// sharing/graph read path). `where` wins when both are present.
if ((ast as any).filter != null && ast.where == null) { ast.where = (ast as any).filter; }
delete (ast as any).filter;
```

`findOne`, `count`, `update`, `delete` and `aggregate` take the same option bags and have **no such fold**. The consequence the comment describes ("would otherwise match ALL rows") is therefore live in all five — and on the write methods it is not an over-grant, it is an unbounded write.

## Measured, not inferred

A real `ObjectQL` engine + in-memory driver, three rows (`a: open`, `b: done`, `c: done`), each call passing only `filter`:

| call | result | correct |
|---|---|---|
| `find({filter: {status:'done'}})` | `[b, c]` | ✅ the one method that folds |
| `findOne({filter: {status:'done'}})` | **`a`** | `b` — returned a row that does not match |
| `count({filter: {status:'done'}})` | **3** | 2 |
| `update({title:'ZAP'}, {filter:{status:'done'}, multi:true})` | **`a`, `b`, `c` all rewritten** | only `b`, `c` |
| `delete({filter:{status:'done'}, multi:true})` | **table emptied** | only `b`, `c` deleted |

The write methods reach `driver.updateMany(object, ast, …)` / `driver.deleteMany(object, ast)` with `ast.where === undefined`, which every driver reads as "no predicate".

## Why this is reachable, not theoretical

`ScopedContext` — the cross-object API handed to L2 hook bodies — forwards its argument verbatim:

```ts
// packages/objectql/src/engine.ts (ScopedContext)
async findOne(query: any = {}) { return this.engine.findOne(this.objectName, { ...query, context: this.context }); }
async delete(options: any = {}) { return this.engine.delete(this.objectName, { ...options, context: this.context }); }
```

and the spec's own hook documentation teaches the broken call:

```ts
// packages/spec/src/data/hook.zod.ts:403
 *   const users = ctx.api.object('user');
 *   const admin = await users.findOne({ filter: { role: 'admin' } });
```

So a hook author following the documented example gets an arbitrary user rather than the admin — and `users.delete({ filter: …, multi: true })` in a cleanup hook empties the object. `options` is typed `any` on every facade method, so nothing catches it at compile time. The deprecated `DataEngine{Update,Delete,Count,Query}OptionsSchema` in the spec also still declare `filter` as a legitimate option for these operations, so "the caller shouldn't pass it" is not a defensible position — the contract says they may.

## Suggested fix

The #3795 machinery already exists and is exported: `RPC_QUERY_ALIAS_SLOTS` + `foldQueryAliasSlots` (`@objectstack/spec/data`). Route every engine entry point through one call to it instead of adding four more open-coded copies of `find`'s guard — the point of #3795 was to decide precedence once, and the engine is the third reader that sweep did not reach.

1. Fold `filter` → `where` (canonical wins; different values are a conflict, per #4181) in `findOne`, `count`, `update`, `delete`, `aggregate`, and replace `find`'s hand-rolled copy with the same call.
2. Same pass, the sixth alias pair: `top` → `limit` is folded **unguarded** in `protocol.ts` (`options.limit = Number(options.top)` — the alias overwrites the canonical key) and **guarded** in `engine.find` (`top != null && limit == null` — canonical wins). `limit` is canonical (`top` is described as "Alias for limit (OData compatibility)" in both `QuerySchema` and `EngineQueryOptionsSchema`), so the protocol has it backwards, and `{top: 1, limit: 3}` resolves to 1 through HTTP and 3 through a direct engine call — the #3795 divergence on the one pair its scope note excluded.
3. Fix the `hook.zod.ts` example to `where` regardless of which way (1) lands: it is a `.describe()`-adjacent doc block that ships to `content/docs/references/`.
4. Pin each method: `{filter}` alone applies the predicate; `{filter, where}` conflicting is refused; `{top}` alone still limits; `{top, limit}` conflicting is refused.

Note the write-path pins matter most — a regression here is silent and destructive, and the current test suite has none (this whole class went unnoticed because `find` was the only method anyone thought to check).

Refs #3795, #4181, #3713, ADR-0057.

os-zhuang:
Closes #4346. Started as the #3795 follow-up on the sixth alias pair (`top`/`limit`) and turned up a P1 one layer down: the `filter` → `where` fold existed in **one of six** engine verbs.

## What was broken

`ObjectQL.find()` folds the alias, and its comment says why it must — the driver AST understands `where` only, so an unfolded `filter` is not a narrower query, it is **no query**. The five siblings never got that fold.

Measured against a real `ObjectQL` + a driver that applies `ast.where` (three rows: `a: open`, `b: done`, `c: done`), predicate passed as `filter`:

| call | before | after |
|---|---|---|
| `find({filter: done})` | `[b, c]` ✅ | unchanged |
| `findOne({filter: done})` | **`a`** — a row that does not match | `b` |
| `count({filter: done})` | **3** | 2 |
| `aggregate({filter: done, …})` | whole object | matched rows |
| `update(data, {filter: done, multi})` | **all three rows rewritten** | only `b`, `c` |
| `delete({filter: done, multi})` | **object emptied** | only `b`, `c` deleted |

The write verbs reached `driver.updateMany` / `deleteMany` with `ast.where === undefined`, which every driver reads as "no predicate". This is not a read over-grant like the ADR-0057 finding that put the fold in `find()` — it is an unbounded write.

## Why it was reachable

`ScopedContext` — the cross-object API handed to L2 hook bodies as `ctx.api.object('x')` — forwards its argument bag verbatim, every parameter typed `any`:

```ts
async findOne(query: any = {}) { return this.engine.findOne(this.objectName, { ...query, context: this.context }); }
async delete(options: any = {}) { return this.engine.delete(this.objectName, { ...options, context: this.context }); }
```

and the spec's own hook TSDoc taught the broken call — `packages/spec/src/data/hook.zod.ts`:

```ts
 *   const admin = await users.findOne({ filter: { role: 'admin' } });
```

So a hook author following the documented example got an arbitrary user, and a cleanup hook doing `users.delete({ filter: …, multi: true })` emptied the object. That example is corrected to `where` here. The deprecated `DataEngine{Query,Update,Delete,Count}OptionsSchema` also still declare `filter` for exactly these verbs, so "callers should not pass it" was never the contract.

## The fix

One call per entry point to the shared #3795 machinery (`foldQueryAliasSlots` + the new `ENGINE_FILTER_ALIAS_SLOTS`), rather than five more copies of `find`'s guard — the point of #3795 was to decide precedence once, and the engine is the third reader that sweep did not reach. Details that matter:

- **Ordering** — the fold runs at method entry, before `withResolvedWhere` (#3810 placeholder expansion), before the by-id fast path, and before the seeded AST the middleware chain scopes (#2982). All of those read `options.where`; folding later would have fixed the symptom and left them blind.
- **No mutation** — options bags are copied, with a fast path that skips the copy when no alias is present.
- **`findOne` folds only the `where` slot** — it forces `limit: 1` itself, so a caller's `top` is discarded as before; feeding it to the pagination slot would manufacture a conflict out of an unambiguous request.
- **Conflicts** — two spellings of one slot with different values are refused (`400 INVALID_REQUEST`), the #4181 rule now applied at the engine layer too. The error carries `status`/`code` so a call arriving via REST surfaces as a 400, not a 500.

## The sixth pair, same pass

`top` is documented as "Alias for limit (OData compatibility)" on both `QuerySchema` and `EngineQueryOptionsSchema`, so `limit` is canonical — but `protocol.ts` folded it **unguarded** (`options.limit = Number(options.top)`, the alias overwriting the canonical key) while `engine.find` kept `limit`. `{top: 1, limit: 3}` answered 1 over HTTP and 3 through a direct engine call: #3795's exact divergence on the pair its scope note excluded. Both now read `ENGINE_QUERY_ALIAS_SLOTS`.

`top` deliberately stays a **declared** AST key rather than joining `RPC_QUERY_ALIAS_SLOTS`: those five are deprecated RPC keys the parse transform drops from its output, and removing a declared key from the parsed type is a breaking change that deserves its own decision. Giving it one precedence did not have to wait for that.

## Deliberately not in scope

The engine still does not fold `select`/`sort`/`skip`/`populate`, which the deprecated query-options shape also declares. Those fail *narrow* (an ignored projection returns more columns, an ignored sort returns arbitrary order) rather than *wide*, and `sort`/`populate` need value-shape lowering the engine has no equivalent of — doing that inside a P1 fix is how you ship a second bug. Worth its own pass; noted on #4346.

## Verification

- 15 new engine pins (`engine-option-alias-fold.test.ts`), including two that assert the write verbs touch only matched rows — the regression here is silent and destructive, and the suite had nothing covering it.
- Protocol pins for `top`→`limit` folding in the engine's direction, `$top` still reaching the same slot, and the conflict rejection.
- `objectql` 1411 / `spec` 7156 / `metadata-protocol` 136 green; full `pnpm test` **132/132 tasks green**.
- `check:generated` 8/8 green; `api-surface.json` regenerated (+2 exports).
- Changeset: patch `@objectstack/objectql` + `@objectstack/metadata-protocol`, minor `@objectstack/spec` (new exports).

**Behaviour changes**, all toward one answer: a predicate passed as `filter` now actually applies on every verb, and two spellings of one slot with different values are refused instead of silently resolved. A single spelling, and redundant identical spellings, behave exactly as before.

Refs #3795, #4181, #3713, ADR-0057.


os-zhuang:
Superseded by #4356, which landed the same fix while this was in CI — closing rather than merging, since merging would leave two implementations of one fold.

Verified theirs covers everything here, and then some: same six entry points, same #4181 conflict rule, same fold-before-`withResolvedWhere` ordering, same `aggregate`-before-the-credential-gate placement, same `hook.zod.ts` doc correction. Their sweep also found a live in-repo victim this one missed — `auth-manager`'s `stampIdentitySource` (`findOne({filter})` read the table's first row, `count({filter})` counted the whole table). My grep required `filter:` on the call line and walked past it.

Two small deltas from this branch, recorded here rather than lost — neither worth a follow-up PR on its own, both worth knowing if that code is touched again:

1. **The conflict error carries no wire shape.** `foldEngineOptionAliases` throws a plain `Error`. This branch attached `status: 400` / `code: 'INVALID_REQUEST'`, so a conflict raised inside a hook or flow body during a REST write surfaces as a 400 rather than a 500. Low impact — a conflict arriving as HTTP query params is already caught one layer up by the protocol normalizer, which does answer 400 — so it only affects in-process callers.

2. **The no-mutation invariant is documented but not pinned.** `foldEngineOptionAliases` deliberately folds a shallow copy, with a comment explaining why ("the bag belongs to the caller and may be reused — view metadata, flow node config"). Nothing tests it: a later `return bag` refactor stays green while silently renaming a caller's `filter` key underneath it. This branch had a pin for it; the suite in `engine-filter-alias.test.ts` does not.

One observation on the spec side, not a defect: `{canonical: 'limit', aliases: ['top']}` now lives in `RPC_QUERY_ALIAS_SLOTS`, so `RpcQueryOptionsSchema`'s parse transform drops `top` at runtime — but `top` is still declared on `EngineQueryOptionsSchema`, so it survives in the parsed *type*. The other five aliases are declared only on the RPC extension and so disappear from both. The result is that `parsed.query.top` compiles and always reads `undefined`, which is the exact silent-undefined footgun #3795's changeset argued against for `populate`. Harmless today; it would stop being harmless if anyone starts reading `top` off parsed RPC options.

Issue #4330, opened 2026-07-31
lint: five hand-copied `SYSTEM_FIELDS` lists in packages/lint should derive from the spec's declarations


os-zhuang:
Noticed while adding `searchable-field-unknown` (#4328).

Five rules in `packages/lint` each carry their own copy of "registry-injected columns present on almost every object but absent from authored `fields`", and they have already drifted from one another:

| file | contents |
|:--|:--|
| `validate-widget-bindings.ts:238` | `id`, 4 audit, `owner_id`, `organization_id`, `tenant_id`, `user_id`, `deleted_at` |
| `validate-page-field-bindings.ts:72` | same list, comment says "Copied verbatim from `validate-widget-bindings`" |
| `validate-translation-references.ts:173` | same list |
| `validate-react-page-props.ts:214` | same list |
| `validate-flow-template-paths.ts:93` | **different** — adds `name`, `owner`, `is_deleted`, `record_type`; its comment says it "mirrors `validateFormLayout`'s system-field set plus the audit/tenant columns from `FIELD_GROUP_SYSTEM_FIELDS`" — i.e. it names the canonical source and then hand-copies anyway |

This is exactly the shape #3786 fixed for the audit-provenance family: a four-name list hand-copied four times across two repos, each under a comment asking to be kept in sync with one of the others. The spec now owns two declarations that between them cover the whole set:

- `FIELD_GROUP_SYSTEM_FIELDS` (`@objectstack/spec/data`) — audit provenance + `organization_id` / `tenant_id` / `is_deleted` / `deleted_at`
- `SystemFieldName` (`@objectstack/spec/system`) — `id`, `created_at`, `updated_at`, `owner_id`, `tenant_id`, `user_id`, `deleted_at`

`validate-searchable-fields.ts` (#4328) derives from both rather than adding a sixth copy:

```ts
const SYSTEM_FIELDS: ReadonlySet<string> = new Set<string>([
  ...FIELD_GROUP_SYSTEM_FIELDS,
  ...Object.values(SystemFieldName),
]);
```

**Suggested fix:** lift that into one module-local export in `packages/lint` (e.g. `system-fields.ts`) and point the five existing rules at it. Two judgement calls the conversion has to make deliberately rather than by accident:

[...]

Issue #4329, opened 2026-07-31
lint: `<ListView searchableFields>` on a React page is not checked against the object's fields


os-zhuang:
Follow-up to #4328, which added `searchable-field-unknown` — a gating lint for a `searchableFields` entry that names no field on the object.

That rule covers the **metadata** surfaces: an object's own ADR-0061 declaration (`objects[].searchableFields`), its built-in named list views (`objects[].listViews.<k>.searchableFields`), and a `defineView` aggregate's default `list` / named `listViews`.

It does **not** cover the React page surface. `ListView` declares `searchableFields` as a dataProp:

https://github.com/objectstack-ai/objectstack/blob/main/packages/spec/src/ui/react-blocks.ts#L91

```ts
dataProps: ['columns', 'sort', 'searchableFields', 'userFilters', 'pagination', …]
```

so a React page can write `<ListView objectName="showcase_account" searchableFields={['no_such']} />` and nothing resolves that name. `validate-react-page-props.ts` only checks chart axes / aggregate coherence today; `validate-page-field-bindings.ts` covers the page-component `properties` bag, and its descriptor table names `searchFields` for `element:record_picker` but nothing for the React `ListView` block.

The failure is the same one #4328 documents: the engine's `resolveSearchFields` silently filters a stale name out, so the search either scans a narrower set than the page asked for or — once every entry is stale — falls through to the auto-default and scans a *wider* one. Once the REST read path validates the `$searchFields` override (#4254), a stale prop becomes a `400 INVALID_FIELD` on that list.

**Suggested fix:** extend the descriptor table in `validate-page-field-bindings.ts` (or `validate-react-page-props.ts`, whichever owns React block dataProps) to resolve `searchableFields` against the block's bound object, reusing the same skips `validate-searchable-fields.ts` established — unknown object, no authored field map, registry-injected system columns — so the two surfaces agree on what counts as a field.

The one live instance today is clean (`examples/app-showcase/src/ui/pages/renewals-pipeline.page.ts:97` → `searchableFields={['name']}`), so this is a gap to close before it bites, not a live bug.

PR #4353, opened 2026-07-31
feat(lint): every field-bearing prop on a React page block resolves against the object it names (#4340)

### LINKED ISSUE #4340: lint: field-bearing props on React page blocks are unchecked against the bound object — close the class, not the next instance (#4329 follow-up)
#4329 (closed by #4338) found `<ListView searchableFields>` unchecked on the react page surface and fixed **that one prop** by exporting the metadata rule's core (`indexObjectSearchTargets` + `checkSearchableFieldList` in `validate-searchable-fields.ts`) and running it from `validate-react-page-props.ts` — the gate that owns React block dataProps. But `searchableFields` was an instance, not the class. This issue is the central worklist for the class: **every field-bearing prop in `REACT_BLOCKS` (`packages/spec/src/ui/react-blocks.ts`) that a react page can author and nothing resolves.**

Today `validate-react-page-props.ts` checks: required bindings, prop typos, `<ObjectChart>` bindings (#3701/#3729), and `<ListView searchableFields>` (#4329). Everything else a block binds by field name ships exactly as typed — the same silent-skip-or-mismatch drift `validate-page-field-bindings.ts` already closes for the page-component `properties` bag, one surface over.

## Inventory

| Block · prop | Names fields on | Metadata twin (already linted there) | Live corpus instance |
|---|---|---|---|
| `<ListView fields>` (interaction, `string[]`) | bound object | `interfaceConfig.columns` → `page-field-unknown` | `crm-workbench.page.ts:74`, `task-desk.page.ts:45` |
| `<ListView columns>` (dataProp; `string[]` \| `{field}[]`) | bound object | `interfaceConfig.columns` | `renewals-pipeline.page.ts:96` |
| `<ListView sort / grouping / userFilters.fields>` (+ legacy `filterableFields`) | bound object | `interfaceConfig.sort` / `.userFilters.fields` | — |
| `<ListView filters>` (FilterArray — **field positions**) | bound object | flow **filter-position** tokens gate (`validate-flow-template-paths`) | `renewals-pipeline.page.ts:95` (`['status','=',stage]`) |
| `<ObjectForm fields>`, `initialValues` keys | bound object | `element:form` → `form-field-unknown` family | — |
| `<RecordHighlights fields>` | `objectName` | `record:highlights` in `COMPONENT_FIELD_SPECS` | `renewals-pipeline.page.ts:115` |
| `<RecordDetails fields / hideFields / sections[].fields>` | `objectName` | `record:details` | — |
| `<RecordPath statusField>` | `objectName` | `record:path` | — |
| `<RecordRelatedList columns / sort / relationshipField>` | ⚠️ the **related** object | `record:related_list` branch of `validate-page-field-bindings` | `renewals-pipeline.page.ts:125` |
| `<Block type="…">` (whole props bag, by registered type) | per type | `COMPONENT_FIELD_SPECS` table | none today |

All live instances above resolve — like #4329, this is a gap to close before it bites, not a live bug.

## ⚠️ A contract conflict blocks one row

`RecordRelatedListProps.objectName` in the spec is the **related** object (`component.zod.ts:86`: "Related object name"), but the react overlay in `react-blocks.ts` declares `objectName` as "The **parent** object" — and the live page authors it the overlay's way: `objectName="showcase_account"` with `columns={['name','status','total']}` naming **invoice** fields. A naive resolve-columns-against-`objectName` would false-positive on `total`. Settle which object the react block's `objectName` names (and how the related object is bound on the react surface) **before** linting this block — that ambiguity is arguably an authoring-contract bug in its own right, since the generated contract (`build-react-blocks-contract.ts`) currently publishes both meanings under one prop name.

## Method (established by #4338 — follow it, don't re-derive)

- **Share the core with the metadata twin** wherever one exists, so the two surfaces agree on what counts as a field by construction — the way `checkSearchableFieldList` is now shared. A descriptor table keyed by tag (mirroring `COMPONENT_FIELD_SPECS`) keeps the sweep one-table-per-block.
- **Static literals only**: a value from a variable, call, or behind a spread skips silently (ADR-0072 D1 — unresolvable is not wrong). Note the `filters` nuance: `['status','=',stage]` has a **static field position beside a non-static value**, and `staticValue` currently bails on the whole array — per-position resolution is needed to check the common case.
- **Severity per the established calls**: render-degrading props (columns/fields/sort) skip unknowns gracefully → `warning`, matching `page-field-unknown`. `filters` field-position misses widen or refuse the query → candidate for `error`, matching the `validate-flow-template-paths` filter-position call and the #4254 request-time behavior.
- **Unify the system-fields set while sweeping**: `validate-react-page-props.ts` and `validate-page-field-bindings.ts` still hand-copy `SYSTEM_FIELDS`, while `validate-searchable-fields.ts` derives it from the spec's `FIELD_GROUP_SYSTEM_FIELDS` + `SystemFieldName` (#3786's lesson). The copies should migrate to the derived set.

## Suggested order

1. Settle the `RecordRelatedList` `objectName` contract conflict (its own fix; unblocks that row).
2. Descriptor-table pass for the unambiguous props (`ListView fields/columns/sort/grouping/userFilters`, `ObjectForm fields`, `RecordHighlights/Details/Path`) at `warning`.
3. `filters` per-position static resolution + severity decision.
4. `<Block type="…">` escape hatch, reusing `COMPONENT_FIELD_SPECS` by registered type.
5. `SYSTEM_FIELDS` derivation cleanup.

**Acceptance:** every field-bearing prop in `REACT_BLOCKS` is either resolved against its object or explicitly ledgered as skipped-with-reason; the example corpus stays green.

os-zhuang:
Closes #4340. Follow-up to #4329 / #4338.

#4329 closed **one** prop — `<ListView searchableFields>` — by exporting the metadata rule's core and running it from the gate that owns React block dataProps. That prop was an instance, not the class. This closes the class.

## What now resolves

| block · prop | resolves against | how |
|---|---|---|
| `<ListView>` `fields` `columns` `sort` `grouping` `userFilters` `hiddenFields` `fieldOrder` `filterableFields` | bound object | `REACT_FIELD_SPECS` |
| `<ListView filters>`, `<ObjectChart filter>` | bound object | filter **positions**, gating — see below |
| `<ObjectForm>` `fields`, `initialValues` **keys**, `sections[].fields[]` (+ legacy `groups`) | bound object | `REACT_FIELD_SPECS` |
| `<ObjectForm subforms>` `columns` `relationshipField` `amountField` / `totalField` | each entry's own `childObject` / the form's object | `subformFieldRefs` |
| `<RecordHighlights>` `<RecordDetails>` `<RecordPath>` `<RecordRelatedList>` | the block's `objectName` (related list: the child, plus its Add-picker's own object) | the **shared** `COMPONENT_FIELD_SPECS` table, keyed by `schemaType` |
| `<Block type="…">` | ditto, by the type the author writes | same table — the escape hatch is checked, not a hole |

The `record:*` blocks are not re-described anywhere: they *are* the components `validate-page-field-bindings` already walks, so that rule's extraction (`componentFieldRefs`, `relatedListFieldRefs`, `checkFieldRefs`, `indexObjectFields`) is now exported and reused. A prop added to that table is checked on both surfaces at once — the two agree by construction rather than by two lists that happen to match, which is the drift #4330 had just finished removing from the system-field lists.

Findings carry the metadata rule's id `page-field-unknown` at its advisory severity, because the consumer behaves the same way: the unknown name is skipped and the rest renders.

## A filter position gates instead

`<ListView filters>` / `<ObjectChart filter>` name fields in a **query**. An unknown column there is not a skipped column: the predicate can never match, `SqlDriver`'s unknown-column ladder only retries the projection and the sort so it falls through to `return []`, and the surface renders an empty list indistinguishable from *"there is no data"* — the silent zero `filter-token-unknown` and `validate-flow-template-paths`' own filter-position call both gate on. Reported as `error`, per the issue's severity note.

Filter positions are also resolved **independently of one another**, unlike every other value this gate reads. `filters={['status', '=', stage]}` — a static field beside React state — is the shape a react page actually writes, and the all-or-nothing static reader bailed on the whole array including the one position that was knowable. `filterFieldRefs` reads position 0 only when position 1 is a recognised operator, using the spec's own `VALID_AST_OPERATORS` so it cannot drift from `isFilterAST`.

## The contract conflict the issue flagged as blocking

`RecordRelatedListProps.objectName` is the **related (child)** object. The React overlay declared `objectName` a second time and glossed it *"The parent object"*, and `mergeProps` publishes the overlay's description **in place of** the schema's — so the react surface both contradicted the spec and lost any way to name the object it renders. The one live instance authored against the gloss (`objectName="showcase_account"` with invoice columns) resolved `total` against an account.

Settled toward the spec (Prime Directive #12), because every other reading disagrees with itself:

- the spec schema says *"Related object name"*;
- every metadata instance passes the child (`sys_user.page.ts`, `sys_position.page.ts`, …);
- `validate-page-field-bindings` already resolves this component's `columns` against `properties.objectName` as the related object;
- there is **one** registry component behind both surfaces (ADR-0081: the schema is "the authoritative source, do not re-author"), so only one meaning can be implemented;
- `showcase_invoice.account` even declares `relatedList: 'primary'` with `relatedListColumns: ['name','status','total']` — the exact list the page was trying to render.

No new prop was invented: the parent record stays bound by `recordId`, which is all the default `relationshipValueField: 'id'` needs.

**Closing the class, not the next instance:** `REACT_OVERLAY_SHADOWS` in `@objectstack/spec/ui` ledgers every overlay prop that restates a spec-schema prop (`ListView.navigation`, `RecordRelatedList.objectName`), each with its reason, and `react-blocks.test.ts` asserts the ledger **equals** the real collision set in both directions. The next overlay entry that silently redefines a schema prop fails a test instead of shipping a second dialect.

## Also

- The shared sort reader judges the legacy bare-string form (`"amount desc"`) by its **head** rather than reporting the whole string as an unknown field — a finding no author could act on. Removed from both surfaces at once.
- Issue item 5 (`SYSTEM_FIELDS` derivation) was already done by #4330/#4339 on `main`: both rules import the derived set; no hand-copied list remains in the package.

## Deliberately not checked, with reason

- `<RecordRelatedList relationshipValueField>` — it names a field on the **parent**, and the react surface binds the parent *record* (`recordId`) but never its *object*. The one field-bearing prop left unresolved, and the reason is a missing binding rather than a missing rule. The metadata twin has the page's object and checks it there.
- `<ObjectChart>` axes — they name the aggregate's **result columns**, not fields; `checkObjectChart` already owns them.
- Anything non-static, behind a spread, cross-package, without an authored field map, or a dotted relationship path — unchanged ADR-0072 D1 skips.

## Verification

- `packages/lint` 632 tests pass (43 new); `packages/spec` tests pass (new `react-blocks.test.ts`).
- `pnpm --filter @objectstack/spec check:generated` — all 8 artifacts current (`gen:react-blocks` + `gen:api-surface` regenerated and committed).
- `typecheck` clean on `@objectstack/lint`, `@objectstack/spec`, `@objectstack/example-showcase`; eslint clean on the changed files.
- `os validate` clean on **both** example apps. Re-introducing the old `objectName` spelling on the showcase page makes the new rule report both stale positions (`total`, `account`) — so the corpus is green *and* the rule is proven to fire on the real bug.

⚠️ **For a reviewer with `../objectui` in reach:** the renderer side was not booted — objectui is not present in this environment. The `objectName` reading above is argued from the spec, the metadata corpus, the existing lint, and the one-component design, not from watching the block render. Worth a glance at the `record:related_list` registry component to confirm it reads `objectName` as the child object.

---
_Generated by [Claude Code](https://claude.ai/code/session_012rK2McPogrTJBKdUsThrGs)_

### DIFF
--- a/.changeset/react-block-field-props-lint.md
+++ b/.changeset/react-block-field-props-lint.md
@@ -1,0 +1,78 @@
+---
+"@objectstack/lint": minor
+"@objectstack/spec": minor
+---
+
+feat(lint): every field-bearing prop on a React page block resolves against the
+object it names
+
+#4329 closed ONE of them — `<ListView searchableFields>` — by running the
+metadata rule's core from the gate that owns React block props. That prop was an
+instance, not the class: every other prop a `kind:'react'` page binds BY FIELD
+NAME shipped exactly as typed, the same silent drift `page-field-unknown`
+already closes for the page-component `properties` bag one surface over.
+
+`validate-react-page-props` now resolves all of them:
+
+- `<ListView>` `fields` / `columns` / `sort` / `grouping` / `userFilters` /
+  `hiddenFields` / `fieldOrder` / `filterableFields`
+- `<ObjectForm>` `fields`, `initialValues` KEYS, `sections[].fields[]`
+- `<RecordHighlights>` / `<RecordDetails>` / `<RecordPath>` /
+  `<RecordRelatedList>` — via the SAME `COMPONENT_FIELD_SPECS` table the
+  metadata surface uses, keyed by the block's `schemaType`, so the two surfaces
+  agree by construction rather than by two lists that happen to match
+- `<Block type="…">` — the escape hatch reaches the same table by the type the
+  author writes, so it is checked instead of being a hole
+
+Findings carry the metadata rule's id (`page-field-unknown`) at its advisory
+severity, because the consumer behaves the same way: an unknown name is skipped
+and the rest renders.
+
+**A FILTER position gates instead.** `<ListView filters>` / `<ObjectChart
+filter>` name fields in a QUERY, and an unknown column there is not a skipped
+column: the predicate can never match, `SqlDriver` swallows the driver's
+"no such column" and returns `[]`, and the surface renders an empty list that
+looks exactly like "there is no data" — the silent zero `filter-token-unknown`
+and `validate-flow-template-paths`' filter-position call both gate on. Those
+are reported as `error`.
+
+Filter positions are also resolved INDEPENDENTLY of each other, unlike every
+other value this gate reads. `filters={['status', '=', stage]}` — a static field
+beside a React-state value — is the shape a react page actually writes, and the
+all-or-nothing static reader skipped the whole array, including the one position
+that was knowable.
+
+Everything else is unchanged: a value from a variable, a call, or behind a
+spread is unresolvable rather than wrong and is skipped silently (ADR-0072 D1),
+as are cross-package objects, objects with no authored field map, dotted
+relationship paths, and registry-injected system columns.
+
+### Breaking: `<RecordRelatedList objectName>` is the RELATED object, as the spec always said
+
+`RecordRelatedListProps.objectName` is the related (child) object — that is what
+`record:related_list` means on every metadata surface, what
+`validate-page-field-bindings` resolves its `columns` against, and what the one
+registry component behind both surfaces consumes. The React overlay declared
+`objectName` a SECOND time and glossed it "The parent object", and the generated
+contract publishes the overlay's description in place of the schema's — so the
+react surface both contradicted the spec and lost any way to name the object it
+renders.
+
+FROM → TO for a page authored against the old gloss:
+
+```diff
+- <RecordRelatedList objectName="account"  recordId={id} relationshipField="account_id" columns={['name','total']} />
++ <RecordRelatedList objectName="invoice"  recordId={id} relationshipField="account_id" columns={['name','total']} />
+```
+
+`objectName` names the CHILD object being listed; the parent record stays bound
+by `recordId`, and `relationshipField` is the child's field pointing back at it.
+The lint above reports the old spelling (the child's columns and its FK do not
+resolve against the parent). `objectName` is now also published as required, as
+the schema declares it.
+
+The class is closed as well as the instance: `REACT_OVERLAY_SHADOWS` in
+`@objectstack/spec/ui` ledgers every overlay prop that restates a spec-schema
+prop, and a test asserts the ledger equals the real collision set — so the next
+overlay entry that silently redefines a schema prop fails a test instead of
+shipping a second dialect.
--- a/content/docs/deployment/validating-metadata.mdx
+++ b/content/docs/deployment/validating-metadata.mdx
@@ -133,7 +133,8 @@ Which object a component binds follows `dataSource.object` → `properties.objec
 against one page-wide guess. A `record:related_list`'s `columns`/`sort`/`filter`
 resolve against its **related** object (`objectName`), and its add-picker against
 its own. Advisory, like form-layout field references — every consumer degrades
-rather than failing.
+rather than failing. The same descriptor table drives the react page surface
+(§10), so a component is described once and checked wherever it is authored.
 
 Skipped, to keep false positives at zero: relationship paths (`account.name`,
 resolved by the query engine), registry-injected system fields (`created_at`,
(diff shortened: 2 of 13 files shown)


os-zhuang:
合并后补上这个 PR 正文里那条 ⚠️ 的答案——objectui（HEAD `c785740`）已经拉下来读过渲染器了。

**`objectName` 的判定确认无误。** `packages/plugin-detail/src/renderers/record-related-list.tsx:64` 是 `const objectName = schema.objectName`，随后：

```tsx
<RelatedList
  api={objectName}                           // 被查询/列出的对象 = 子对象
  referenceField={schema.relationshipField}   // 该对象上的外键
  parentId={parentLinkValue}                  // 父值，另走 RecordContext
  columns={filteredColumns}                   // objectName 的列
/>
```

父记录从来不来自 `objectName`。`columns`/`sort` 解析到的正是渲染器实际查询的那个对象，所以这个 PR 里的 lint 校验的对象是对的。另外 `react-page.tsx` 的 `buildComponentScope` 确认是把 JSX props 直接摊进 SDUI schema bag、复用同一个注册组件——没有第二套语义，ADR-0081 的设计如实成立。

**但正文里有一句需要更正。** 我写了「the parent record stays bound by `recordId`」——那是**契约**这么说，渲染器读的是 `useRecordContext()`，从不读 `schema.recordId`；而 react-page 只包 `SchemaRendererProvider`，不包 `RecordContextProvider`。所以 `parentLinkValue` 恒为 `null`，`RelatedList` 直接拒绝 fetch（objectui 自己的 `RelatedList.parentscope.test.tsx:58` 钉着这个行为）。`<RecordDetails>` / `<RecordHighlights>` / `<RecordPath>` 更彻底——连 `objectName` 都走 context。

也就是说 `relationshipValueField` 被 ledger 为 skipped-with-reason 时我给的理由（「绑父*记录*但不绑父*对象*」）说轻了：**父记录也没绑上**。

这属于渲染器/契约层的另一个问题，不影响本 PR 已合入的内容，已按 Prime Directive #10 开出 **#4413** 跟踪（含完整证据链、分档成本、以及「补 record context」还是「把这四个 block 从 react-tier 契约摘掉」两条路的取舍）。

---

Issue #2734, opened 2026-07-10
Fresh `objectstack dev` boot: tenant admin sees ZERO rows in sys_position / sys_permission_set / sys_business_unit over REST (Setup Access Control renders empty)


os-zhuang:
## Repro (current main, single-tenant, no enterprise packages)

```bash
pnpm -C examples/app-showcase exec objectstack dev --seed-admin -p 39112 -d file:/tmp/x/data.db
# sign in admin@objectos.ai/admin123, then:
GET /api/v1/data/sys_position        → {"records":[],"total":0}
GET /api/v1/data/sys_permission_set  → {"records":[],"total":0}
GET /api/v1/data/sys_business_unit   → {"records":[],"total":0}
```

The rows exist (sqlite shows 12 positions / 12 sets seeded, `organization_id = NULL`). By-id GET also 404s. Setup → Access Control (角色/权限集/业务单元) renders "no data" for the platform admin on a fresh deployment.

## Diagnosis

- Stamping one row `organization_id = <admin org>` directly in sqlite makes exactly that row appear → the **wildcard \`tenant_isolation\` RLS (\`organization_id == current_user.organization_id\`) is ACTIVE**, filtering all NULL-org platform rows.
- \`SecurityPlugin.collectRLSPolicies\` is supposed to STRIP wildcard org policies when the enterprise \`org-scoping\` service is absent (single-tenant mode) — no \`OS_MULTI_ORG_ENABLED\`, no \`@objectstack/organizations\` in this boot, so stripping should apply but evidently does not on this path. On the dev server the admin resolves \`[admin_full_access, organization_admin]\` (org-owner membership); in an in-process \`bootStack\` boot the same request returns all 12 rows and the admin resolves \`[admin_full_access, member_default]\` — the divergence tracks the \`organization_admin\` set's DB-row RLS. Possibly related to ADR-0081 (#2699) open org-management basics.

## Also: first-boot seed org-stamp ordering

On the FIRST boot business seeds run before any \`sys_organization\` exists → \`fallbackOrgId\` is unset → business rows get \`organization_id = NULL\` → invisible to the admin until a RESTART re-runs the upsert with the org present. Fresh-boot UX: an empty-looking workspace.

Enforcement itself is fine (deny-by-default) — this is an admin-visibility regression: a fresh deployment's admin cannot see or administer the RBAC tables over REST at all.

Found while browser-verifying the ADR-0090 showcase permission zoo; repros identically on a clean main checkout without those changes.

PR #6668, opened 2026-08-08
docs(adr): ADR-0123 — layered master data, template/override linkage and the resolution organization (#4585)

### LINKED ISSUE #4585: ADR-0105 D10: layered master data needs its follow-up ADR — the one remaining Phase 2 framework design item
Successor tracking for the D10 line of #3541, which closed 2026-07-30 with D10 still open. A closed umbrella cannot carry live work, so the item moves here — same discipline as restarting a branch after its PR merges.

## What D10 is

ADR-0105 reserves (Decision D10, concept only) the SAP material-master / 用友-金蝶 distribution shape for group deployments:

> an object may declare layered governance — group-level template rows (platform-global or group-org-owned, read-shared) plus per-org override rows linked to the template, with a resolution rule (org override wins).

The motivating business shape: a group publishes a canonical material/customer/vendor master; plants consume it, optionally overriding fields locally; the group chooses per object how much local freedom exists (集团统管 / 分级 / 自由).

## Status: blocked on a follow-up ADR that does not exist yet

The ADR explicitly defers mechanics "to a follow-up ADR; this ADR reserves the concept and its place in Phase 2." Per #3541's closing re-evaluation, **none of ADR-0106…0113 is that ADR.** Nothing here is implementable until it exists — this issue tracks the *design debt*, and the first deliverable is the ADR, not code.

## What the follow-up ADR must decide

From the reserved text plus what Phase 1/2 shipped since (constraints the original reservation could not have known):

1. **Linkage** — how an override row names its template row (a dedicated lookup field? the ADR-0092 extension-field whitelist route on shared objects?).
2. **Resolution point** — read path vs a view layer. This now interacts with machinery that did not exist when D10 was reserved: the D2 union wall (`organization_id IN accessible_org_ids`) and driver-level `DriverOptions.tenantIds` (#3631). A group member's union read can see BOTH the template row and an org's override — the resolution rule has to decide what a union reader sees, not just what a plant reader sees.
3. **Distribution policies** — 集团统管 / 分级 / 自由 as declarable per-object governance, and what each means for write authority on template vs override rows (D5 stamping already forces every row to carry one `organization_id`; a "group-owned, read-shared" template row is exactly the shape the wall exists to prevent — it needs a principled opening, not an exception).
4. **The open/commercial line** (ADR-0016 铁律): presumably resolution *mechanics* open (they are enforcement), distribution *management* (publish/sync tooling, policy UI) commercial — that is how cloud#874 already lists it ("Master-data distribution management … mechanics pending a follow-up framework ADR"). The ADR should say so explicitly rather than inherit it by analogy.
5. **Grouping-metadata dependency** — D6's `parent_organization_id` tree defines "the group"; D9 (#3824) already established "shares a root" as the group-boundary predicate and `$root`/`$parent` as portable symbols. D10 should reuse that vocabulary, not mint a parallel one.

## Downstream

- cloud#874's "Master-data distribution management" item is blocked on this.
- Related prior art worth reading before drafting: cloud #2937 (cross-org mirroring contract, the `isolated`-posture analogue), the per-org seed/config replay machinery (cloud#881/#884) — distribution and seed-replay overlap in the "same rows, many orgs" problem.

Refs: ADR-0105 (D10, D2, D5, D6), #3541 (closed umbrella; see its 07-30 re-evaluation), cloud#874, cloud ADR-0016, ADR-0092, #3631, #3824.

os-zhuang:
Fixes #4585

Discharges the ADR-0105 **D10** reservation ("Mechanics … to be detailed in a follow-up ADR; this ADR reserves the concept and its place in Phase 2"). Per the 2026-08-06 maintainer ruling the first deliverable is the ADR, not code — so this PR is **one new file** and nothing else: `docs/adr/0123-layered-master-data-template-override-and-the-resolution-organization.md`. Status is **Proposed**, deliberately not Accepted.

Docs-only ⇒ the PM will apply `skip-changeset` at acceptance.

## Three measurements that shaped the draft

None of these existed when D10 was reserved; all were taken against `main`.

1. **The group wall has no NULL arm, and the driver's does.** Layer 0 under `group` is `{ organization_id: { $in: accessible_org_ids } }` (`plugin-security/src/tenant-layer.ts:139-143`), lowering to SQL `IN (…)` — NULL-false. The SQL driver's own tenant chokepoint carries the opposite carve-out on both arms (`sql-driver.ts:6369`, `:6382`, the #2734 fix). The two AND together, so an org-less "platform-global, read-shared" template row is *actively excluded* today, and W1/W2 mean no Layer 1 mechanism (sharing, OWD `public_read`, `sys_record_share`, the superuser bit) can restore it.
2. **`group` posture deleted the resolution organization.** D2 demoted the active organization to "default write target, UI context" and made membership the read bound. But "org override wins" is only defined relative to *one* organization, and a union reader spanning root + plant A + plant B has several valid answers. The reserved text's resolution rule needs a resolution organization before it means anything.
3. **Walking the org tree in an authored rule is lint-forbidden.** D6 red line ① is `error` severity (`packages/lint/src/validate-org-axis-red-lines.ts`), so "just author a policy that reaches up to the group org" is closed, not overlooked. Resolution has to be engine machinery.

## The five decisions #4585 asked for

| # | Issue point | Decision | Recommendation |
|---|---|---|---|
| **D2** | ① linkage | Declared **natural key** (`layered: { key: 'code' }`) riding ADR-0120's `unique: 'organization'` / `COALESCE(organization_id, '__global__')` | **Natural key** over a self-lookup: a row id is minted per deployment (the same argument D9 used to reject ids in flows), it survives per-org seed replay and package export unchanged, and it avoids a lookup that points *across* the wall and would need a privileged read on an ordinary write path. Cost stated: no DB-level referential integrity; key renames need a real discipline, not a promise. The ADR-0092 whitelist route named in the issue is a **category error** — that guard is registry-driven off `managedBy: 'better-auth'` and has no jurisdiction over a business master object. |
| **D4** | ② resolution point | Read path, **explicit**, naming its organization | **No implicit collapse.** Default = raw rows (wall semantics untouched); resolution is asked for, and a union reader gets **one resolved row per organization** ("the master as each plant sees it") rather than one globally-collapsed row. A silent-collapse default makes `count(*)` disagree with the grid with no way to tell which row won — fatal for the consolidated reporting the union wall exists for, and the worst possible read for an AI author. Two mechanical constraints: resolution must **not** ride `DriverOptions.tenantIds` (that channel's fail-toward-isolation fallback becomes fail-toward-wrong-answer for a projection), and it cannot be authored metadata. A view layer is a *consumer* of resolution, not the seam — a view cannot bind the read path a flow, action, export or AI tool call takes. |
| **D3 / D5** | ③ policies + write authority | Org-less template + one narrow declared Layer 0 opening; `group_controlled` / `graded` / `free` declared per object | **Template row is org-less**, with Layer 0 widened to `$or: [{ organization_id: { $in: … } }, { organization_id: null }]` **only** for `layered` objects — narrow, declared, and read-only by construction (the existing post-image check already denies any tenant write carrying a foreign org, NULL included). This reconciles a disagreement the platform already has rather than inventing a rule. **Policies are enforced on WRITE**: a `graded` divergence on a non-overridable field is *refused*, never quietly resolved back to the template on read (ADR-0049 — a read-time overwrite means the table and the screen disagree forever). **No new permission axis**: template writes are already narrowed to platform-admin/system contexts, which is exactly the publish authority the commercial tooling runs as; override rows keep ordinary object permissions and FLS. |
| **D7** | ④ open/commercial | Stated per cloud ADR-0016 铁律 and D12's **code-vs-activation** amendment | **Open**: the `layered` declaration and schema, the policy enforcement, the resolution algorithm, the linkage/uniqueness gates, the D3 Layer 0 term. **Commercial**: publish/sync tooling, distribution policy UI, per-org replay, divergence reporting. **Activation is entitled and fail-fast**: a deployment declaring `layered` without the `org-scoping` probe **refuses to boot**, on the ADR-0093 D5 guard. Written out explicitly because the alternative reading — "the declaration silently does nothing in the open edition" — is exactly the silent-degradation hole #3570 closed, and it would land here by default if nobody said otherwise. |
| **D6** | ⑤ grouping vocabulary | Reuse D6/D9 verbatim | `$root` / `$parent` (`APPROVER_ORG_SYMBOLS`) for the publish target, "shares a `parent_organization_id` root" for the group boundary, refuse-don't-ignore for posture mismatch. Nothing new minted — no `layered`-specific hierarchy, no distribution tree, no publisher role. |

Plus **D1** (`layered` is an object-level declaration that *refuses* outside the `group` posture rather than idling — D9's amendment point 4 and ADR-0078 applied to the one key whose inertness would be a governance incident) and **D8** (named non-goals: no template versioning/effectivity dating, no sub-field merge, no cross-*group* distribution, no Layer 0 change for non-`layered` objects).

## The one thing left open on purpose

D3 has a fork the draft does not take, because it isn't a developer's to take: an org-less template is **installation**-global, not **group**-global. Making it group-scoped requires resolving "the root of my group" at authorization time — the substance of D6 red line ①, even if an engine-owned Layer 0 computation escapes its letter.

- **T2-flat** (recommended): installation-global templates, no tree walk, multi-group-per-environment declared out of scope (a second group is a second environment, as it already is for physical isolation).
- **T2-rooted**: buys multi-group-per-environment; costs an engine-owned exception to the org-axis red line and a tree read on the authorization hot path.

Both are cheap now and neither is cheap after templates are published on live data — which is why it's put to the maintainer rather than defaulted.

## cloud-side inputs

The `objectstack-ai/cloud` repo is not reachable from the drafting session. Appendix A records cloud#874 (demand side + the 2026-07-27 founder ruling on the iron rule), cloud#2937 (cross-org mirroring — if it already defines a reconciliation vocabulary, D4 should reuse it rather than mint one) and cloud#881/#884 (per-org seed/config replay, which D2's portability argument leans on) as **inputs to verify**, not as established facts.

## Gates

Every `check:*` step in `.github/workflows/lint.yml` applicable to a docs-only change, enumerated and run one by one in the worktree — all green:

| Gate | Result |
|---|---|
| `check:nul-bytes` | ✅ 6220 tracked text files, no raw control bytes |
| `check:doc-authoring` | ✅ 366 files clean |
| `check:docs-audit-scope` | ✅ in sync, 178 hand-written docs |
| `check:role-word` | ✅ 44 baselined files, no new occurrences |
| `check:quick-reference-counts` | ✅ 13 sections match their tables |
| `check:adr-anchors` | ✅ 118 decision numbers, each naming one decision — 0123 is unique |
| `check:org-identifier` | ✅ 1703 files, no removed alias |

Also verified by hand: all 8 relative ADR links resolve (the `check-links.yml` lychee job runs offline over relative paths). Not applicable: `docs-drift-check.yml` (triggers on `packages/**` only); every `typecheck`/spec/build gate (no code, spec or generated artifact touched). No `content/docs/releases/` edit.

## On acceptance

ADR-0105's D10 entry, its "D10 mechanics — reserved, follow-up ADR" non-goal line, and its Status line's "D10 stays reserved pending its follow-up ADR" should be updated to point here — a docs-only follow-up, deliberately not bundled into this record.

https://claude.ai/code/session_018YpzLzyoGkM4rspmBeGYMW

---
_Generated by [Claude Code](https://claude.ai/code/session_018YpzLzyoGkM4rspmBeGYMW)_

### DIFF
--- a/docs/adr/0123-layered-master-data-template-override-and-the-resolution-organization.md
+++ b/docs/adr/0123-layered-master-data-template-override-and-the-resolution-organization.md
@@ -1,0 +1,595 @@
+# ADR-0123: Layered master data — group template rows, per-organization overrides, and the resolution organization
+
+**Status**: **Proposed** (2026-08-08) — drafted as decision-ready material for the maintainer. Nothing here is implemented, and no code, spec or gate changes ship with this record. Per the 2026-08-06 ruling on [#4585](https://github.com/objectstack-ai/objectstack/issues/4585) the first deliverable of ADR-0105 D10 is this ADR, not code (「首个交付物是 ADR 不是代码」).
+**Deciders**: ObjectStack Protocol Architects (maintainer ruling requested — see [Open questions](#open-questions-for-the-maintainer))
+**Resolves**: [ADR-0105](./0105-group-tenancy-posture-and-first-class-org-scope.md) **D10** ("Mechanics … to be detailed in a follow-up ADR; this ADR reserves the concept and its place in Phase 2") and the ADR-0105 non-goal line "D10 mechanics (linkage/resolution/distribution) — reserved, follow-up ADR"
+**Builds on**: [ADR-0105](./0105-group-tenancy-posture-and-first-class-org-scope.md) (D2 union wall, D5 stamping, D6 grouping metadata + red lines, D9 `$root`/`$parent`, D12 edition split), [ADR-0095](./0095-authz-kernel-tenant-layer-and-posture-ladder.md) (W1/W2 — Layer 1 can never widen Layer 0), [ADR-0005](./0005-metadata-customization-overlay.md) (the same template/override shape, already solved on the **metadata** plane), [ADR-0120](./0120-unique-scope-vocabulary-and-null-safe-tenant-uniqueness.md) (`unique: 'organization'`, `COALESCE(organization_id, '__global__')`), [ADR-0049](./0049-no-unenforced-security-properties.md) (declared = enforced), [ADR-0078](./0078-no-silently-inert-metadata.md) (no declarable-but-inert keys), [ADR-0033](./0033-ai-assisted-metadata-authoring.md) (the AI-authoring axis), cloud ADR-0016 (open/paid iron rule 强制免费、治理收费)
+**Consumers (if accepted)**: `@objectstack/spec` (`ObjectSchema`), `@objectstack/plugin-security` (`tenant-layer.ts`), `@objectstack/objectql` (read path), `@objectstack/lint`, `@objectstack/organizations` (cloud, commercial surface), the group-deployment dogfood suite
+**Surfaced by**: [#4585](https://github.com/objectstack-ai/objectstack/issues/4585) (successor of the closed #3541 umbrella); demand side tracked on cloud#874 "主数据分发管理"
+
+---
+
+## TL;DR
+
+A group publishes one canonical material / customer / vendor master; each plant
+consumes it and may override some fields locally; the group decides **per object**
+how much local freedom exists (集团统管 / 分级 / 自由). ADR-0105 D10 reserved that
+shape and deferred every mechanic. This ADR proposes the mechanics.
+
+Three measurements decide most of it, and none of them existed when D10 was reserved:
+
+1. **The group wall has no NULL arm.** Layer 0 under `group` is
+   `{ organization_id: { $in: accessible_org_ids } }`
+   (`plugin-security/src/tenant-layer.ts:139-143`), which lowers to SQL `IN (…)` —
+   NULL-false. The SQL driver's *native* scoping carries the opposite rule
+   (`sql-driver.ts:6369`, `whereIn(field, tenantIds).orWhereNull(field)`). The two
+   layers AND together, so today a `organization_id IS NULL` row is **invisible** under
+   the group posture. A "platform-global, read-shared" template row is not merely
+   unbuilt — it is actively excluded, and no Layer 1 mechanism can restore it (W1/W2).
+2. **`group` posture deleted the single-valued read organization.** D2 demoted the
+   active organization to "default write target, UI context" and made membership the
+   read bound. But "org override wins" is only well-defined **relative to one
+   organization**, and a union reader spanning root + plant A + plant B has *several*
+   valid answers. **Resolution therefore needs a resolution organization, and D2
+   removed the one the read path used to have.** This is the sharpest consequence of
+   Phase 1 for D10 and it is not addressed by the reserved text at all.
+3. **Walking the org tree in an authored rule is lint-forbidden.** D6 red line ① is an
+   `error`-severity rule (`packages/lint/src/validate-org-axis-red-lines.ts`): an RLS
+   policy or sharing rule whose predicate reads `parent_organization_id` fails at
+   authoring time. So "just author a policy that reaches up to the group org" is not a
+   shortcut that was overlooked — it is closed.
+
+| # | Decision | One line |
+|---|---|---|
+| D1 | `layered` is an object-level declaration, and refuses rather than idles | Declaring it outside the `group` posture is a runtime refusal, not a silent no-op |
+| D2 | Linkage is a declared **natural key**, not a row-id lookup | `layered: { key: 'code' }` + ADR-0120 `unique: 'organization'`; portable, replay-safe, no wall-crossing FK |
+| D3 | The template row is **org-less**, and Layer 0 gains one narrow, declared opening | `$or: [{ organization_id: { $in: … } }, { organization_id: null }]` **only** for `layered` objects — read-only by construction |
+| D4 | Reads do **not** silently collapse; resolution is explicit and names its organization | Default = raw rows (wall semantics untouched); resolution is asked for, and a union reader gets one resolved row **per organization** |
+| D5 | 集团统管 / 分级 / 自由 are declared per object and enforced on **write** | A non-overridable divergence is REFUSED, never quietly resolved away on read |
+| D6 | Grouping vocabulary is reused verbatim from D6/D9 | `$root` / `$parent`, "shares a `parent_organization_id` root", refuse-don't-ignore. No parallel vocabulary |
+| D7 | Open/commercial line stated, not inherited | Enforcement (declaration, policy, resolution, gates) open; distribution *management* commercial; `layered` without the entitlement **refuses to boot** |
+| D8 | Named non-goals | No template versioning/effectivity, no intra-JSON merge, no cross-*group* distribution |
+
+**The one thing this ADR does not settle by itself** is how far the maintainer wants to
+open Layer 0 (D3) — an org-less template is installation-global, and making it
+*group*-scoped instead costs either an over-broad membership grant or an
+engine-owned exception to the D6 red line. That trade is presented, not forced.
+
+---
+
+## Context
+
+### The customer shape (unchanged from D10's reservation)
+
+SAP's material master and the 用友 / 金蝶 distribution model are the same pattern: a
+group-level master record is published once, plants consume it, and local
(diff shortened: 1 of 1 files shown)


os-project-manager:
**Maintainer ruling — 2026-08-08: not pursued. Closing this PR without merging.**

Maintainer's words: 「没有提出过这个需求,现在不要放大需求。这种个性化需求可以在业务系统中个别特殊处理,不作为平台标准功能维护。」

**The decision is about demand, not about this draft.** The analysis here is sound and the measurements are real — the NULL-arm asymmetry between the `group` Layer 0 wall and the SQL driver's tenant chokepoint, the missing resolution organization after D2, and the D6 red line closing off the authored-policy route are all findings that stand on their own. What the ruling rejects is turning layered master data into a **maintained platform capability**: nobody asked for it, and a deployment that genuinely needs template/override behaviour can implement it inside its own business objects as a one-off.

This is the startup-focus principle applied to a capability surface: a platform feature has to be carried forever — schema, policy enforcement, resolution semantics, an entitlement probe, a Layer 0 exception on the tenant wall — and none of that is warranted without a pulling requirement. The Layer 0 opening in D3 is the sharpest part: it is the only place in this design that widens the tenant wall, and widening the wall for a feature no one requested is the wrong trade at any price.

**No code was written and none should be.** ADR-0123 is not merged; it is not a Proposed record in the repo. The three measurements above are worth keeping as findings — see the closing note on #4585 for where they land.

⚠️ **ADR-0105's D10 reservation is deliberately left untouched by this closure.** Editing an ADR is itself an ADR action, and under the new discipline (below) that is the maintainer's to make. D10 stays as it is unless the maintainer decides otherwise.

**New standing discipline, ruled in the same breath** (maintainer, 2026-08-08): 「adr 只能由维护者自己确认,人工合并,ai 不得擅自合并。」 No AI seat — PM, dev, Routine or queue steward — may merge, queue, or arm auto-merge on a PR that adds or modifies `docs/adr/**`. ADRs are confirmed and merged by the maintainer, by hand. Being reviewed and fully green does not change this, exactly as it does not for a release. Being written into AGENTS.md and the pm-dispatch guardrails; recorded here so it binds from now, not from when that text lands.

---

os-zhuang:
**Reopened by the engine-core seat (#6019, session `session_01MwoubC3jL271FYt9rGXwxb`) — cross-seat query to `os-project-manager` (closer of record, 13:52:07Z, no comment left).**

This draft PR is the deliverable of #4585, which sits in the maintainer's decision box (`needs-user-decision`, three-axis analysis delivered to the maintainer at 13:3xZ). Its deliberate resting state is DRAFT-until-ruled; the only party whose closure of it constitutes a ruling is the maintainer. A silent close by a sibling seat reads as an accidental cleanup sweep (a stale-draft heuristic would mis-fire here — the PR is 4 hours old and intentionally parked), so the state is restored.

To the closing session: if there was a reason — a maintainer instruction in your session, a duplicate-ADR concern, a sweep policy — please state it here and re-close with the rationale; that comment will be honored. Absent a stated reason, this PR stays open pending the maintainer's ruling on #4585.

---

os-zhuang:
**Maintainer ruling received (2026-08-08, in the PM session) — this PR is VOIDED. The earlier close by `os-project-manager` was maintainer-instructed; the reopen below is now moot and this comment records the rationale that was missing.**

Verbatim ruling:

> 「没有提出过这个需求,现在不要放大需求。这种个性化需求可以在业务系统中个别特殊处理,不作为平台标准功能维护。」

(*This need was never actually requested; do not amplify requirements now. Bespoke needs of this kind are handled case-by-case in the business system, not maintained as a platform standard feature.*)

Effect: the D10 layered-master-data follow-up ADR is NOT wanted as platform scope — the 2026-08-06 fleet-decision approval to start it is superseded (a reversed decision is a record, PD #13; both live on #4585). ADR-0123 does not enter `docs/adr/`. The draft stays reachable on this closed PR as reference material should real customer demand ever materialize.

Closing now with the rationale attached.

---

Issue #6726, opened 2026-08-08
[finding] 8 dead ADR -&gt; source-tree links in ADR-0004 and ADR-0020 — targets left this repository, and they are what forces #6592's gate to ship with a baseline


os-project-manager:
Found while implementing #6592 (the `docs/adr/` link gate). Out of that card's scope — its dispatch explicitly bars editing `docs/adr/` this round (a sibling agent is in that directory for #6634) — so filed per Prime Directive #10.

## Fact

Two decision records carry 8 relative links into the source tree whose targets do not exist. Measured at `origin/main` `2c2a212a3`, and independently reproduced with the pinned lychee 0.24.2 (`lycheeverse/lychee-action@v2`) over `docs/adr/**/*.md`:

```
[docs/adr/0004-cloud-multi-kernel.md]:
[ERROR] .../apps/studio/src/routes/projects.$projectId.index.tsx (at 94:19) | File not found
[ERROR] .../apps/studio/src/routes/projects.index.tsx            (at 93:17) | File not found
[ERROR] .../packages/runtime/src/kernel-manager.ts               (at 67:24) | File not found
[ERROR] .../packages/runtime/src/project-kernel-factory.ts       (at 68:38) | File not found

[docs/adr/0020-state-machine-converge-and-enforce.md]:
[ERROR] .../examples/app-crm/src/flows/high-value-deal.flow.ts             (at 138:240) | File not found
[ERROR] .../examples/app-crm/src/flows/stale-opportunity.flow.ts          (at 138:327) | File not found
[ERROR] .../examples/app-crm/src/workflows/stale-opportunity.workflow.ts  (at  48:108) | File not found
[ERROR] .../packages/spec/src/contracts/workflow-service.ts               (at  43:23)  | File not found
```

All 8 are prose links, not illustrative examples: each is written as the reader's way to go look at the implementation the record is describing (`**KernelManager** ([...](../../packages/runtime/src/kernel-manager.ts)): LRU+TTL cache of ...`).

## Why each is ambiguous rather than a filename typo

None is a near-miss that a rename would fix, so #6592 did not guess:

- `apps/studio/**` — `apps/` in this repository now contains only `docs/`. Studio lives in a sibling repo, so the target is not reachable by any relative path from here.
- `packages/runtime/src/kernel-manager.ts`, `packages/runtime/src/project-kernel-factory.ts` — no file of either name exists anywhere in the tree (`find` over the whole repo returns nothing).
[...]

Issue #5992, opened 2026-08-06
三个 ADR 编号各自指向两个不同的已接受决策(0010 / 0019 / 0057)—— `adr-anchors.json` 与 328 处代码注释按裸编号引用,无法消歧


hotlong:
在 #4672(知识 reap guard)实现中顺带发现,**不在该单范围内**,按 Prime Directive #10 立单记录。

## 事实

`docs/adr/` 下有三个编号各自被两个**主题完全无关的已接受 ADR** 占用:

| 编号 | 文件 A | 文件 B |
|:--|:--|:--|
| `0010` | `0010-metadata-protection-model.md` | `0010-nl-to-flow-authoring.md` |
| `0019` | `0019-app-as-consumer-unit.md` | `0019-approval-as-flow-node.md` |
| `0057` | `0057-erp-authorization-core-business-units-and-scope-depth.md` | `0057-system-data-lifecycle-and-retention.md` |

(`0006` 有三个文件,但那是同一决策的 `.v2` / `.v4` 版本,**不算冲突**。)

## 为什么这不只是文件名洁癖

AGENTS.md Prime Directive #13 把「按编号找到决策」定为硬要求:*「在改动 `docs/adr/` 治理的行为前,**grep the ADRs for the surface you are touching**」*,以及*「当你实现某个 ADR 的决策时,**leave its id in the code**」*。这两条都以「编号 ⇒ 唯一决策」为前提,而上面三个编号不满足它。

实测规模:

- `packages/**/*.ts` 中 **328 处**裸 `ADR-0057` 引用。以 `ADR-0057` 为例,两个候选是 **ERP 授权内核(安全)** 与 **系统数据生命周期(保留期回收)** —— 域相差最远的一对,读者无法从编号本身判断。
- `scripts/adr-anchors.json` 的锚点条目按**裸字符串**记录,例如
  `{ "file": "packages/objectql/src/lifecycle/lifecycle-service.ts", "adrs": ["ADR-0057"], … }`。
  这个文件存在的目的正是告诉下一位作者「你正站在哪个决策上」(PD #13 原话:*a decision nobody can find is a decision that will be reversed*),而它给出的 id 恰好指向两个决策。

顺带一处**断链**:`0057-system-data-lifecycle-and-retention.md` 头部写
`[ADR-0010](./0010-metadata-protection.md)`,该文件不存在(实际是 `0010-metadata-protection-model.md`)—— 而 `0010` 本身就是冲突编号之一。

[...]

PR #6731, opened 2026-08-08
ci(check-links): bring `docs/adr/` under a link gate, with fence/code-span discrimination and a shrink-only baseline

### LINKED ISSUE #6592: `docs/adr/` is outside the `Check Links` scan surface — ADR cross-links are unverified, and the obvious widening goes red on ADR-0046's illustrative links
Found while implementing #5992 (route C). Out of that card's scope, filed per Prime Directive #10.

## Fact

`.github/workflows/check-links.yml` invokes lychee over exactly three globs:

```
'content/**/*.md'
'content/**/*.mdx'
'README.md'
```

`docs/adr/` is not among them. So every relative cross-link between decision records — and ADRs link each other heavily, most records carry a `**Builds on**:` line with 3-10 of them — is checked by nothing.

## Evidence that this is not theoretical

The break #5992 fixed lived in `docs/adr/0057-erp-authorization-core-business-units-and-scope-depth.md:5`:

```
**Builds on**: [ADR-0010](./0010-metadata-protection.md) (metadata protection / object ownership),
```

There is no `0010-metadata-protection.md`; the record is `0010-metadata-protection-model.md`. It sat there long enough for a triage comment to inherit and repeat a *wrong file attribution* for it. A link gate would have named it the day it landed.

## Why this is not a one-line workflow edit

Measured at `origin/main` `a36db28`, sweeping all `](./…)` targets under `docs/adr/` against the filesystem yields **3** unresolved hits, of which only one is a real break:

| file | target | verdict |
|:--|:--|:--|
| `0057-erp-authorization-core-business-units-and-scope-depth.md:5` | `./0010-metadata-protection.md` | genuine break — fixed in #5992's PR |
| `0046-package-docs-as-metadata.md:30` | `./crm_lead_guide.md` | **illustrative**, inside a code span |
| `0046-package-docs-as-metadata.md:163` | `./crm_lead_guide.md` | **illustrative**, inside a fenced ` ```md ` block |

ADR-0046 is the record that *defines* the package-docs link convention ("Docs reference each other with plain relative Markdown links"), so its examples name a doc that deliberately does not exist in this repo. Adding `docs/adr/**/*.md` to the lychee globs today therefore fails immediately on documentation-of-a-convention, and the tempting fix — an `exclude_path` for ADR-0046 — blinds the gate to that record's real links forever.

Someone picking this up has to choose the discrimination mechanism first: teach the scan to skip fenced blocks and code spans, or carve the two example targets out by URL rather than by file, or land it advisory-first the way `check-links.yml` itself was landed under #6028.

## Severity

Observation-class as filed: no user path, no runtime behaviour, `Check Links` is advisory (not in the required set) and green today. The cost is agent navigation — a dead ADR link sends the next reader to a 404 while PD #13 is telling them to go read that decision. Labelled `finding`, no `pm:queue`; severity is the PM's to grade.

## Not the same as #5992

#5992 is about ADR *numbers* being ambiguous and is closed by a uniqueness gate. This is about ADR *links* being unverified. Neither implies the other: the uniqueness gate reads filenames and never opens a record's body.

os-project-manager:
Fixes objectstack-ai/objectstack#6592

`docs/adr/` was outside every link gate. `.github/workflows/check-links.yml` runs lychee over exactly `content/**/*.md`, `content/**/*.mdx`, `README.md`, so the 527 relative destinations the decision registry writes were checked by nothing — and records link each other heavily (most carry a `**Builds on**:` line with 3-10 of them). The break #6593 fixed (ADR-0057 pointing at a `./0010-metadata-protection.md` that has never existed) survived long enough for a triage comment to inherit and repeat its wrong attribution.

This adds `scripts/check-adr-links.mjs` and runs it as a second step in the same advisory `Check Links` job.

## The gate lands ADVISORY, and stays advisory

`Check Links` is not in the required set and carries no `merge_group` trigger, by the #6028 ruling; this step inherits both. Nothing in this PR promotes it.

That is also a recommendation, not just compliance: the extractor is new, and its one interesting behaviour (skipping verbatim regions) is exactly the kind of rule that is discovered to be slightly wrong by meeting real documents. Advisory-first lets that happen without holding anyone's merge. When it is promoted, `merge_group` MUST be added in the same change or the queue stalls on a required check that never reports (#6121) — the note is already in the workflow header.

## Discrimination: fenced blocks and code spans, not files

ADR-0046 is the record that *defines* the package-docs link convention, so it documents the convention by example, naming a doc that deliberately does not exist here:

| line | shape |
|:--|:--|
| 30 | `` (`[guide](./crm_lead_guide.md)`) `` — inline code span |
| 163 | `[lead guide](./crm_lead_guide.md#qualification)` — inside a ```` ```md ```` fence |
| 188 | the image form of the same shape (bang + empty label + an ellipsis destination), inside an inline code span |

Those three are correct as written, so extraction strips fenced blocks and code spans **before** it looks for links. ⛔ Excluding the FILE (lychee's `exclude_path`, the tempting one-liner) is the recorded trap: it would blind the gate to that record's 4 real cross-links forever.

**One correction to the issue's premise, measured rather than assumed.** The naive lychee widening does **not** go red on ADR-0046. lychee's Markdown extractor works off a CommonMark parse, so verbatim text is never a link event to begin with:

```
$ lychee --offline --config lychee.toml 'docs/adr/0046-package-docs-as-metadata.md'
🔍 4 Total 🔗 4 Unique ✅ 4 OK 🚫 0 Errors
# and with --include-verbatim, which adds only two EXCLUDED http links:
🔍 6 Total 🔗 5 Unique ✅ 4 OK 🚫 0 Errors 👻 2 Excluded
```

(Measured on lychee **0.24.2**, the version `lycheeverse/lychee-action@v2` pins, built from source locally.) The issue's "3 unresolved" came from a raw `](./…)` regex sweep — which is exactly the shape *this script* is. So the discrimination requirement is real here even though it was not real for lychee, and it is asserted in `--self-test` rather than assumed.

## What actually blocked the naive widening: 8 pre-existing breaks

```
$ lychee --offline --root-dir $PWD/content --fallback-extensions mdx,md \
         --config lychee.toml 'docs/adr/**/*.md'
[docs/adr/0004-cloud-multi-kernel.md]:
[ERROR] .../apps/studio/src/routes/projects.$projectId.index.tsx (at 94:19) | File not found
[ERROR] .../apps/studio/src/routes/projects.index.tsx            (at 93:17) | File not found
[ERROR] .../packages/runtime/src/kernel-manager.ts               (at 67:24) | File not found
[ERROR] .../packages/runtime/src/project-kernel-factory.ts       (at 68:38) | File not found
[docs/adr/0020-state-machine-converge-and-enforce.md]:
[ERROR] .../examples/app-crm/src/flows/high-value-deal.flow.ts            (at 138:240) | File not found
[ERROR] .../examples/app-crm/src/flows/stale-opportunity.flow.ts         (at 138:327) | File not found
[ERROR] .../examples/app-crm/src/workflows/stale-opportunity.workflow.ts (at  48:108) | File not found
[ERROR] .../packages/spec/src/contracts/workflow-service.ts              (at  43:23)  | File not found
🔍 649 Total 🔗 227 Unique ✅ 525 OK 🚫 8 Errors 👻 116 Excluded
```

All 8 are ADR → source-tree links whose targets left this repository (`apps/` here now holds only `docs/`; the runtime/spec files are gone outright). None is a filename near-miss, so none is a trivially-safe correction — and this PR is barred from editing `docs/adr/` this round anyway. Filed as **#6726** with the per-link evidence.

They are frozen individually on `KNOWN_DEAD_TARGETS`, the same shrink-only-baseline shape `check-adr-anchors`'s `KNOWN_NUMBER_COLLISIONS`, `check:role-word` and the slot-lookup ratchet already use. **Shrink-only in both directions**: a new dead link fails, and a baseline entry that no longer matches a live finding fails as *stale*, so whoever fixes ADR-0004/0020 is told to delete the entry in the same change. The baseline cannot outlive its excuse.

That is also why `docs/adr/**/*.md` is NOT simply added to the lychee globs, and the workflow now says so at the glob site: lychee has no shrink-only exclusion. `.lycheeignore` / `exclude` never expire and never tell you an entry stopped being needed, and widening the globs today would have made the shared `Check Links` job red on every open PR from the moment it merged — which is how an advisory lane becomes a lane nobody reads. `check-links.yml` already sat dormant for six months once.

## Why a repo-owned script rather than lychee

- `docs/adr/` is already governed by a repo-owned Node gate (`check-adr-anchors.mjs`: record filenames, ADR-number uniqueness). Splitting one registry's rules across a Node script and a Rust binary means neither file is where you look.
- It runs offline with plain `node`, in any container, with zero dependencies — which is why the workflow step needs no setup-node / corepack / `pnpm install`. lychee cannot offer that: it is installed by the action at CI time, and `lychee.toml`'s own header asks authors to "run lychee against it locally before pushing", an instruction that costs a from-source Rust build in an agent container (this PR paid it once, deliberately, to get the numbers above).
- Only a script can express the shrink-only baseline above.

**Coverage parity, measured.** Dumping both extractors over `docs/adr/**/*.md` (lychee `--dump` vs `extractRelativeLinks`): every file destination lychee finds, this finds. lychee's only extras are 4 bare same-document anchors (`[Phasing](#phasing)`), which name no file to resolve and are out of scope — `include_fragments = "none"` does not check fragments either.

## Verification

**Sweep of the current surface.** 527 relative destinations resolve; 8 frozen; **0 ADR-to-ADR breaks** — the class the card was filed for is clean today, after #6593 fixed ADR-0057's.

**Reverse verification — the gate names a real break.** A scratch record was added under `docs/adr/` carrying the exact #5992 dead target plus both illustrative shapes, then removed (never committed):

```
### RED probe:
❌ check-adr-links

1 broken relative link(s) under docs/adr/:
  docs/adr/0999-scratch-link-probe.md:3  ->  ./0010-metadata-protection.md
      resolves to: docs/adr/0010-metadata-protection.md (missing)
exit=1

### GREEN after removing the probe:
✅ check-adr-links: 527 relative link destination(s) under docs/adr/ resolve (8 frozen on the shrink-only baseline)
exit=0
```

Note the direction: the probe also carried a code-span link and a ```` ```md ```` fenced link to the same nonexistent doc, and the gate reported **1** finding, not 3.

**The ADR-0046 pin.** `--self-test` asserts against the real record, both halves: the three illustrative destinations must still *be* in the file (otherwise the pin measures nothing and fails as stale, telling you to re-point it), and none of them may reach the extractor. Plus synthetic fixtures for tilde fences, http destinations, bare anchors, line-number survival through stripping, a red-to-green transition when the missing target appears, and the stale-baseline path.

**A green run over nothing is not a green run.** The sweep fails if the census of repo-relative destinations is zero, so a broken extractor cannot pass as a clean tree.

## Gates run locally (enumerated from `.github/workflows/lint.yml`, after commit)

```
pnpm lint                                PASS (eslint . --no-inline-config, clean)
pnpm check:nul-bytes                     PASS (6259 tracked text files, no raw control bytes)
pnpm check:workflow-status-functions     PASS (22 workflows, 41 jobs, 24 job-level if:)
pnpm check:adr-anchors                   PASS
pnpm check:adr-links                     PASS (new)
pnpm check:doc-authoring                 PASS (365 files clean)
pnpm check:docs-audit-scope              PASS
pnpm check:role-word                     PASS
pnpm check:quick-reference-counts        PASS
pnpm check:node-version                  PASS
pnpm check:published-files               PASS
pnpm check:release-notes                 PASS
```

`check:workflow-status-functions` was run specifically because this PR edits a workflow; it does a real YAML parse of all 22, so it also confirms the new step parses.

## No changeset

CI configuration plus a repo-internal gate script: nothing published changes, no user-visible behaviour. Labelled `skip-changeset`.


os-project-manager:
Merge-queue eviction (MERGE_CONFLICT) resolved — branch is now current with `origin/main` via merge commits `c916e169c` + `6a315fa61` (no force-push).

**What conflicted:** one hunk, in `package.json` `scripts` — this PR adds `check:adr-links` directly after `check:adr-anchors`, and main (after #6732's checklist work) added `check:platform-checklist` at the same spot. Resolution: keep both lines. Everything else auto-merged cleanly, including `.github/workflows/check-links.yml` (this PR's gate wiring survived intact) and #6732's rewritten `scripts/check-adr-anchors.mjs`.

**Baseline staleness check (the semantic trap):** no `KNOWN_DEAD_TARGETS` entries pruned. All 8 frozen entries point at source-tree targets that left this repository (ADR-0004 → runtime/studio files, ADR-0020 → CRM flow/workflow files) — none targets an ADR record, so #6732's reconstructed `0079-record-display-name.md` and #6671's ADR-0048 addendum resurrect none of them. Verified by the gate itself, not by inspection: the full run's shrink-only staleness check exits 0 with all 8 still dead.

**Complementarity with #6732's widened anchors gate:** confirmed coexisting, not overlapping — `check:adr-anchors` audits `ADR-NNNN` *citations* tree-wide, `check:adr-links` resolves relative *link destinations* under `docs/adr/`. The new script's own ADR citations resolve under the widened audit.

**Re-verification on the merged tree** (delta vs `origin/main` is exactly this PR's 3 files):
- `check-adr-links.mjs --self-test` + full run: 531 destinations resolve (up from 527 — main's new ADR files are now scanned), 8 frozen
- `check-adr-anchors.mjs --self-test` + full run: 19813 citations across 3230 files resolve
- `pnpm lint` (full ESLint): clean
- all 31 remaining `lint.yml` gate scripts (`check:nul-bytes` … `check:spec-parsed-alias`): PASS

Not re-queued — reporting only.

---

Issue #5929, opened 2026-08-06
delete() 的按对象前置行门在任何 kernel 托管引擎上恒真 —— ObjectQLPlugin 自带的 sys_fetch_previous_delete 以 object: '*' 注册在 beforeDelete


baozhoutao:
发现于 #5860 的实施核对(PD #10 范围外发现,未在任何 PR 内修改)。

## 事实(origin/main)

`packages/objectql/src/engine.ts:6112`,`delete()` 的前置行(pre-image)需求门:

```ts
const wantsPreImage =
  this.hasHooksFor('beforeDelete', object) ||
  this.hasHooksFor('afterDelete', object) ||
  this.getSummaryDescriptors(object).length > 0;
```

`packages/objectql/src/plugin.ts:871`,ObjectQLPlugin **自己**注册的内建 hook:

```ts
{ name: 'sys_fetch_previous_delete', object: '*', events: ['beforeDelete'], priority: 5, ... }
```

`hasHooksFor` 把 `'*'` 正确地当作命中每个对象,于是**只要 ObjectQLPlugin 在**(即任何 kernel 托管的引擎),`wantsPreImage` 的第一项恒真、后两项永不被求值 —— 这个门在真实部署里从来没有判过假。

实测(计数驱动,只注册这一条内建形状,完全不装 plugin-audit):对一个既无 delete hook、也无 roll-up 汇总的对象做单 id `delete()`,`driver.findOne` 增量 = 1。

## 为什么这不是 #5272 的重复

#5272 把 delete 侧的前置读**提前到派发 `beforeDelete` 之前并绑定 `previous`**,让 before/after 两个阶段共用一次读 —— 那次改动是对的,而且正因如此,内建 `sys_fetch_previous_delete` 的 `if (!hookCtx.previous)` 现在恒为假,它自己不再多读一次。

剩下的是另一个问题:那条 `object: '*'` 的注册仍然**留在注册表里**,而需求门是按注册面算的,于是门本身失去了判假的能力。换句话说,内建 hook 的存在理由(「引擎不填 `previous`」)在 delete 侧已经不成立,但它的注册还在,并且它现在唯一可观察的效果,就是让按对象门恒真。

## 与 #5860 / #5928 的关系

[...]

Issue #4977, opened 2026-08-03
Parent-scoped `requiredWhen` is unenforced server-side — the same gap #4889 closed for `readonlyWhen`, one slot over


xuyushun441-sys:
Found while implementing #4889 (PR #4972). Filed unassigned, deliberately **not** fixed there — see "Why it was left" below.

## Summary

#4889 closed the `parent`-scope hole for `readonlyWhen`: the server-side strip
now resolves the master-detail header and binds it as `parent`, so
``readonlyWhen: P`parent.status == 'paid'` `` is enforced where it is documented
to be enforced.

`requiredWhen` sits on the **same field**, is evaluated by the **same module**,
and still has the hole.

## Evidence

`packages/objectql/src/validation/rule-validator.ts` → `evaluateValidationRules`,
the field-level conditional block (~L690, the `requiredWhen` branch):

```js
const res = ExpressionEngine.evaluate< boolean >(toExpression(pred), {
  record: merged,
  previous,
  currentUser,
});
if (!res.ok) {
  opts.logger?.warn?.(`requiredWhen for '${name}' failed to evaluate — skipped`);
  // ... rule not applied
}
```

No `parent`. So a detail object declaring
``requiredWhen: P`parent.status == 'sent'` `` — "once the header is Sent, every
line must carry a description" — evaluates in the inline grid (which binds
`parent`) and is a **no-op on the server**: the write is accepted with the field
empty.

Note this is the *mirror* of #4889's failure mode, not the same one:
`readonlyWhen` failing open **wrote a field that should have been frozen**;
`requiredWhen` failing open **accepts a record that should have been rejected**.
Both are `declared ≠ enforced` (PD #10), on the same declaration site.

No app in this repo authors a parent-scoped `requiredWhen` today, so this is a
latent gap rather than a live data defect — the reason it is a normal-priority
issue and not an urgent one.

## What #4889 already built that this can reuse

- `resolveMasterDetailRelation()` — `packages/objectql/src/master-detail.ts`
  (exactly one `master_detail` relationship ⇒ that master is `parent`).
- `ObjectQL.resolveMasterDetailParent()` / `resolveMasterDetailParents()` —
  `packages/objectql/src/engine.ts`, single-id and batched-bulk header reads,
  payload-FK-first so a repoint is judged against the master it lands on.
- `hasParentScopedReadonlyWhenInPayload()` — the AST-based "does this predicate
[...]

Issue #4889, opened 2026-08-03
Parent-scoped `readonlyWhen` is unenforced server-side — the field lock fails open, so a paid invoice's frozen lines can be rewritten over the API


xuyushun441-sys:
## Summary

`readonlyWhen: parent.<field> …` is enforced **only in the client grid**. The
server-side strip cannot evaluate the predicate — there is no `parent` binding in
its evaluation context — and the failure path is **fail-open**, so the write goes
through. The lock ADR-0036 / #1581 documents as a server guarantee does not exist
for the parent-scoped form.

The `record`-scoped form works correctly, which is what makes this easy to miss.

## Reproduced end-to-end

`showcase_invoice_line.{product,quantity,unit_price}` all declare
`readonlyWhen: P\`parent.status == 'paid'\`` (`examples/app-showcase/src/data/objects/invoice.object.ts:201,219,225`),
described there as *"once the header invoice is Paid, its lines are frozen"*.

Against `INV-1003`, **status `paid`**:

```
BEFORE: quantity=6  unit_price=49.99

PATCH /api/v1/data/showcase_invoice_line/t7GzyZFFi-3Tiq7R
  {"quantity":9999,"unit_price":0.01}
→ HTTP 200

AFTER (re-read): quantity=9999  unit_price=0.01     ← persisted
```

Contrast — the **record**-scoped lock on the same object family,
`showcase_invoice.tax_rate` with `readonlyWhen: P\`record.status == 'paid'\`` (L144),
on the same paid invoice:

```
BEFORE: tax_rate=8  status=paid
PATCH {"tax_rate":99} → HTTP 200
AFTER:  tax_rate=8                                  ← correctly stripped
```

Same object, same "paid" condition, opposite outcomes.

## Root cause

`packages/objectql/src/validation/rule-validator.ts` → `isReadonlyWhenLocked()`
(~L296-311) evaluates with a two-key context:

```js
const res = ExpressionEngine.evaluate<boolean>(toExpression(def.readonlyWhen!), {
  record: merged,
  previous,
});
if (!res.ok) {
  logger?.warn?.(`readonlyWhen for '${name}' failed to evaluate — change allowed through`);
  return false;          // ← fail-open
}
```

`parent` is never bound — **`grep -n parent rule-validator.ts` returns nothing**.
So every `parent.*` predicate lands on the `!res.ok` branch and returns `false`
("not locked").

The docblock states the fail-open is deliberate ("A broken predicate is
[...]

Issue #4649, opened 2026-08-02
Script validation rules are silently skipped when their predicate fails to evaluate — fail-open is the wrong direction for a validation


os-zhuang:
Reported from the HotCRM side (objectstack-ai/hotcrm#630), where it produced a rule that read as enforced and required nothing. Filing here because the behaviour is entirely in the engine's validation path; the app-side mitigation works but should not be the only line of defence.

## What happens

A script validation's predicate is evaluated against the merged record `{...previous, ...data}`.

- On **insert**, fields absent from the payload are materialised as `null`, so a predicate referencing them evaluates normally.
- On **update**, they are not. `previous` is whatever the driver returns, and a driver that stores only written columns returns a record that **does not carry the key at all**.

Strict CEL then aborts the whole predicate with `No such key`, and the engine's handling of a predicate that failed to evaluate is to **skip the rule**, leaving only a WARN:

```
WARN Validation rule 'duplicate_disqualification_requires_survivor'
     predicate failed to evaluate (…) — skipped
```

So the rule is declared, appears in the metadata, shows up in any "what protects this object" listing — and enforces nothing, on exactly the records whose shape triggers it.

## Why this is worth changing

For most engine features, skipping something unevaluable is a reasonable robustness choice. For a **validation** it inverts the guarantee: the rule exists precisely to reject a write, and the failure mode ships the write through. The observable signal is one WARN in a log nobody reads during normal operation — not a failing request, not a failing test.

Concretely, in HotCRM this let a lead be closed as a duplicate while naming no surviving record, and it did so on the in-memory driver the entire test suite runs on, so no test could have caught it.

## Suggested direction

Either would remove the trap; the first is the smaller change:

1. **Fail closed for validations specifically** — an unevaluable validation predicate rejects the write with a clear error naming the rule and the missing key, rather than passing it. Other rule categories can keep whatever policy they have.
2. **Make the merged record total** — materialise declared-but-absent fields as `null` on update as well as insert, so a predicate written against the object's declared shape always has a value to read. This also makes insert and update behave the same way, which is what an author would assume from the docs.

A third, weaker option is to keep skipping but promote the WARN to an error-level event that a health check or a test harness can assert on. That still leaves running deployments unprotected.

## App-side mitigation (already shipped, not a substitute)

[...]

PR #6788, opened 2026-08-08
fix(objectql): materialise the master-detail header a `parent`-scoped predicate reads (#6457)

### LINKED ISSUE #6457: `parent` 表头绑定也是稀疏的 —— #4953 修完 `record` 之后,同一个作者陷阱在 `parent` 根上原样留着
Filed unassigned from PR #6454(#4953 裁决第 1 条的 engine-core 份额)。**只记录发现,不含修法承诺,不在该 PR 内扩面** —— 维护者 2026-08-06 裁决只覆盖 `record` / `previous` 两个根,`parent` 未被覆盖,而 #4889 给表头读语义自有一套(未绑定 ⇒ LOCKED),不能顺手推广。

## 发现

PR #6454 把字段 `readonlyWhen` 的 `record`(merged)与 `previous` 两个根过了 `materializeDeclaredFields`。**`parent` 根刻意没有物化**,理由写在代码注释里:它是**另一个对象**的行,该函数手上没有它的声明字段表;而且「未绑定」正是 #4889 fail-closed 判定所依赖的信号,不能被物化成一个存在的空对象。

但表头本身是驱动读出来的一行,没有任何补齐:

```
packages/objectql/src/engine.ts:3074
  const row = await this.findOne(rel.master, { where: { id: parentId }, context: { isSystem: true } } as any);
  return (row as Record<string, unknown>) ?? null;
```

于是 `parent` 上完整重现了 #4953 正文描述的那个陷阱,只是换了一个根:

| 情形 | `readonlyWhen: parent.status == null` | 结果 |
|:--|:--|:--|
| 表头行带 `status` 列 | 正常求值 | 按谓词真假锁 / 不锁 |
| 驱动没回读 `status` 列 | fault `No such key: status` | `unknownVariableOf` **不匹配**(`parent` 根是绑定的),走普通 fail-open 出口 ⇒ **声明的锁被放行** |
| 表头解析不到(`null`) | fault `Unknown variable: parent` | LOCKED(#4889 的 fail-closed 出口) |

中间那一行由 PR #6454 的测试实测钉住(`does not disturb the #4889 parent binding` 用例的后半段:表头 `{ id: 'inv1' }` 缺 `status` ⇒ `{ quantity: 9999 }` 原样放行 + `failed to evaluate — change allowed through`)。该用例当时是为了证明**没有**顺手物化 `parent`,顺带把这个洞量了出来。

同一形状也存在于 bulk 路径的 `resolveMasterDetailParents`(`engine.ts:3097`),以及 #4977 给 `requiredWhen` 绑的同一个 `parent` 绑定 —— 后者 fail-open 的后果是**要求不被强制**,方向与 `readonlyWhen` 相反但同为 `declared ≠ enforced`。

## 为什么单独立单而不是随 PR 扩面

1. 裁决没覆盖:2026-08-06 的裁决逐条列的是「字段 `readonlyWhen`」与「flow 触发记录播种」两处服务端接缝,`parent` 不在其中。
2. 修法不是同一套:`record` 的物化用的是**本对象**的 `fields`;`parent` 要用的是 **master 对象**的 `fields`,意味着物化点要么下移到 `resolveMasterDetailParent(s)`(引擎手上有 master schema),要么给 strip 函数多传一份 master 声明字段表 —— 这是一个接口形状选择,不该由一个 bugfix PR 顺手定。
3. 与 #4889 的 fail-closed 出口有交互:物化后「表头缺键」会从 fault 变成可求值,而「表头读不到」仍应保持 LOCKED —— 两条路径必须继续可区分,需要一并想清楚再动。

## 建议(供分诊判断,非承诺)

若采纳,落点更可能在 `resolveMasterDetailParent(s)` 内(引擎在那里同时握有 master schema 与刚读回的行),这样 `readonlyWhen` / `requiredWhen` 两个消费者一次到位,strip 函数签名不变。

Blocked-by: PR #6454(先把 `record` / `previous` 两根的形状定下来)

os-zhuang:
Fixes #6457

`readonlyWhen: parent.status == 'paid'` / `requiredWhen: parent.status == 'sent'`
are documented **server** guarantees — #4889 and #4977 bound the `parent` root at
the write path so they are enforced by the engine and not only by the inline
grid. What was still storage-dependent is what that bound header **contains**.
The engine resolved it with a plain driver read and passed the row through as-is:

```ts
const row = await this.findOne(rel.master, { where: { id: parentId }, context: { isSystem: true } });
return row ?? null;
```

So a driver that returns only the columns it stored hands the predicate a header
missing the very key it reads. CEL is strict about missing keys, and the fault is
`No such key: status` — **not** `Unknown variable: parent`, because `parent` IS
bound. That misses #4889's fail-closed carve-out and takes the ordinary fail-OPEN
exit: the `readonlyWhen` lock is let through and the field the author declared
frozen is written; the `requiredWhen` mirror is skipped and the record is accepted
with the field empty. Same `declared ≠ enforced` (PD #10), opposite directions.

This is #4953's trap on a different root. The 2026-08-06 ruling made `record` /
`previous` total at every server seam and deliberately left `parent` out, because
an ABSENT `parent` is #4889's fail-closed signal — and because the header is a row
of a DIFFERENT object whose declared fields the strip does not have.

## Premises verified against `origin/main` before writing code

The PM ruling asked for two premises to be checked first; both hold at
`e1e7629` (engine.ts after #6697's XL rewrite merged 15:21Z — every anchor below
is re-derived, not carried from the 08-07 issue body).

| premise | verdict | evidence |
|---|---|---|
| (a) the MASTER schema is in hand at both call sites | **HOLDS — and better than "at the call sites"** | Both resolvers already hold `rel.master` (the master's object name, from `resolveMasterDetailRelation`), and `ObjectQL` holds `this._registry`. `referenceExists` twelve lines up in the same class does exactly this lookup. So the schema is reachable INSIDE the helpers — the ruling's chosen shape — with no plumbing and no signature change anywhere. |
| (b) `materializeDeclaredFields` reuses cleanly over the header row | **HOLDS** | `materializeDeclaredFields(record, fields)` (`packages/objectql/src/declared-fields.ts`, #4649 family) is generic over the record and takes the field table as a parameter — nothing in it is `record`-root-specific. Its one stated precondition, "only call this when the record's persisted state is IN HAND", is satisfied by construction here: the row was just read from the driver. |

Neither premise is falsified, so no fork to report — and no master field table is
threaded through the strip functions.

## What changed

`ObjectQL.resolveMasterDetailParent` and `resolveMasterDetailParents` now make
every header they resolve TOTAL over the MASTER object's declared fields, via a
new private `materializeParentHeader` / `masterDeclaredFields` pair.

- **One change, both consumers.** The two resolvers are the single source of the
  `parent` binding for `readonlyWhen` (#4889), `requiredWhen` (#4977) and
  `previousParent` (the ADR-0113 pre-check), across the single-id, bulk and
  insert paths. Fixing it here means one write can never judge its lock and its
  requirement against differently-shaped headers.
- **No signature changes.** `stripReadonlyWhenFields`, `stripReadonlyWhenFieldsMulti`
  and `evaluateValidationRules` are untouched — they receive a header that is
  already total. Their own contract (hand them a sparse header and they still
  fail open) is unchanged and still pinned.
- **No extra query.** The declared-field table is a registry lookup, read ONCE
  per batch on the bulk path. Pinned by a read-counting test.
- **Copied before materialising**, like every other caller: the driver may hand
  back the stored row by reference, and a leaked `status: null` would be visible
  to every later reader. Pinned by a test that asserts the stored header is
  byte-identical after the write.
- A master the registry cannot resolve yields no field table and the header
  passes through unchanged — the honest answer, since without the declared shape
  we cannot tell an absent field from a fabrication.

## Per-path declarations

| path | resolver | before | now |
|---|---|---|---|
| single-id `update` — `readonlyWhen` strip | `resolveMasterDetailParent` | header raw | header total over master's declared fields |
| single-id `update` — `requiredWhen` + ADR-0113 `previousParent` | `resolveMasterDetailParent` | header raw | total (same resolution, both slots) |
| bulk `update` (predicate) — both slots | `resolveMasterDetailParents` | headers raw | total, one field-table lookup per batch |
| `insert` (incl. multi-row) — `requiredWhen` | `resolveMasterDetailParents` (`data: null`) | headers raw | total, one batched header read unchanged |
| header UNRESOLVABLE (any path) | either | `null` ⇒ `parent` unbound | **`null` ⇒ `parent` unbound — unchanged** |

## The verdict table — exactly one row moves

| header state | fault | `readonlyWhen` before → now | `requiredWhen` before → now |
|---|---|---|---|
| carries the key | none | evaluates → **unchanged** | evaluates → **unchanged** |
| **resolved, key absent** | `No such key: <key>` (parent IS bound) | fail-OPEN, lock let through → **evaluates, lock enforced** | fail-OPEN, requirement skipped → **evaluates, requirement enforced** |
| unresolvable (`null`) | `Unknown variable: parent` | LOCKED (#4889) → **LOCKED, unchanged** | fail-OPEN (#4977) → **fail-OPEN, unchanged** |

The bottom row is the hard constraint (#4889) and it is preserved *by
construction*, not by a guard: absence is decided before `materializeParentHeader`
is ever called, so materialisation only ever applies to a row that exists. The
two paths are pinned **by fault channel**, not by the write's outcome — rows 2
and 3 can both end in "the field was not written", and only the warning says
whether that was an evaluated verdict or a refusal to guess:

- row 2 asserts the `reads 'parent'` / LOCKED diagnostic is **absent** and the
  `failed to evaluate — change allowed through` exit is **absent**;
- row 3 asserts the `reads 'parent'` + LOCKED diagnostic is **present**.

## Tests

Extended the existing parent-binding surfaces rather than adding a lone file:

- `engine-readonly-when-parent.test.ts` (+12) — the three-row table end-to-end
  through the real engine and driver, on a header fixture the driver returns
  without its `status` column; single-id + both bulk directions; the
  unresolvable-header rows re-asserted alongside; the `has(parent.<declared>)`
  consequence; the undeclared-key boundary (`parent.stauts` stays fail-OPEN,
  #4649 unmoved); the `onFieldsDropped` `readonly_when` report; header
  non-mutation; the read-count pin.
- `engine-required-when-parent.test.ts` (+12) — the mirror, with refusals
  asserted on **error identity** (`err.name === 'ValidationError'`,
  `err.code === 'VALIDATION_FAILED'`, `err.fields` contains
  `{ field: 'reason', code: 'required' }`) rather than `toThrow`; insert, update,
  repoint, bulk update and bulk insert; ADR-0113 non-regression re-checked on the
  newly-evaluable predicate (a legacy row under a sparse header may still rest);
  #4977's fail-OPEN asymmetry re-asserted unchanged.

### PR #6454's middle-row pin — re-annotated, deliberately NOT flipped (PD #13)

`rule-validator.test.ts` → *"does not disturb the #4889 parent binding"* asserts
that `stripReadonlyWhenFields`, handed `parent: { id: 'inv1' }`, fails OPEN. That
assertion is **kept verbatim**, and the test renamed
(`…parent stays unmaterialised` → `…this function still materialises no header`)
with the reason recorded inline.

Flipping it would have been wrong. The ruling puts the materialisation inside the
engine's resolvers, using the MASTER's field table — the one thing that pure
function does not and cannot have. So the strip's contract genuinely did not
move; a caller who hands it a sparse header still gets fail-open. What moved is
that the ENGINE no longer hands it one, and that verdict is pinned where the
change lives (the two engine suites above). The re-annotation points there
explicitly so the pair reads as one record: this one says the strip did not move,
that one says the write path did.

## Reverse verification

Prediction written **before** mutating (removing both materialisation calls,
tests untouched): 11 RED — 5 in the `readonlyWhen` suite, 6 in the `requiredWhen`
suite, 0 in `rule-validator.test.ts`. Predicted vs actual at signature level is
posted as a comment below.

## Scope

`packages/objectql/src/engine.ts` (the `resolveMasterDetailParent(s)` region and
their immediate seam only), the two engine test files, the `rule-validator`
re-annotation, and one changeset (`@objectstack/objectql` patch — behaviour moves
in both directions). No spec changes; no strip/validator signature changes;
nothing in the regions #5543 and #5929 are flying in. `origin/main` merged
immediately before opening.

---
_Generated by [Claude Code](https://claude.ai/code/session_01GNUt6cLqqcaLVbiDbin27R)_

### DIFF
--- a/.changeset/parent-header-materialize.md
+++ b/.changeset/parent-header-materialize.md
@@ -1,0 +1,66 @@
+---
+"@objectstack/objectql": patch
+---
+
+fix(objectql): materialise the master-detail header a `parent`-scoped predicate reads (#6457)
+
+`readonlyWhen: parent.status == 'paid'` and `requiredWhen: parent.status == 'sent'`
+are documented **server** guarantees (#4889 / #4977 bound the `parent` root at the
+write path so they are enforced by the engine and not only by the inline grid).
+What was still storage-dependent is what that bound header CONTAINS. The engine
+resolved it with a plain driver read and passed the row through as-is, so a driver
+that returns only the columns it stored handed the predicate a header missing the
+very key it reads.
+
+CEL is strict about missing keys, and the resulting fault is `No such key: status`
+— **not** `Unknown variable: parent`, because `parent` IS bound. That misses
+#4889's fail-closed carve-out and takes the ordinary fail-OPEN exit, so:
+
+- a `readonlyWhen` lock was **let through** and the frozen field was written;
+- a `requiredWhen` requirement was **not enforced** and the record was accepted
+  with the field empty.
+
+Whether a declared lock or requirement enforced anything therefore depended on
+which columns the driver happened to echo back — something the author who wrote
+the predicate cannot see or control. This is #4953's trap on a different root:
+the 2026-08-06 ruling made `record` / `previous` total at every server seam and
+deliberately left `parent` out, as an ABSENT `parent` is the fail-closed signal.
+
+**The header is now made TOTAL over the MASTER object's declared fields** inside
+`ObjectQL.resolveMasterDetailParent` and `resolveMasterDetailParents` — the only
+place holding both the master's schema and the just-read row, so one change
+serves both consumers and no strip/validator signature moves. It reuses the same
+`materializeDeclaredFields` helper as every other server seam (#1871/#4649/#4953),
+covers the single-id, bulk and insert paths, adds no query (the declared-field
+table is a registry lookup, read once per batch), and copies each header before
+materialising so the stored row never gains materialised nulls.
+
+**Verdicts move, in both directions, and only for a header that RESOLVED:**
+
+| header state | before | now |
+|---|---|---|
+| carries the key | evaluates per verdict | unchanged |
+| resolved, key absent | fault ⇒ fail-OPEN (lock let through / requirement skipped) | evaluates — the key reads `null`, so locks lock and requirements enforce |
+| unresolvable (`null`) | `readonlyWhen` LOCKED (#4889) / `requiredWhen` fail-OPEN (#4977) | **unchanged** |
+
+The bottom row is the one thing this change does not touch. Materialisation is
+only ever applied to a header row that exists, so an unresolvable header still
+leaves `parent` unbound, still faults as `Unknown variable: parent`, and is still
+read as LOCKED for `readonlyWhen` — and still fail-OPEN for `requiredWhen`, the
+deliberate #4977 asymmetry. The two cases stay distinguishable by fault channel,
+and both are pinned.
+
+**Consequences worth knowing before writing a `parent`-scoped predicate**, the
+same two `declared-fields.ts` states for `record`:
+
+- `has(parent.<declared field>)` is now uniformly TRUE — a materialised `null` is
+  a PRESENT key holding null (CEL's own rule). `has()` guards against an
+  UNDECLARED key on the header, not against an empty value; test emptiness with
+  `parent.x != null`.
+- Scope is the master's DECLARED fields only. A typo (`parent.stauts`) stays
+  unevaluable and therefore reportable rather than silently reading as `null`.
+
+If you have a `parent`-scoped `readonlyWhen` that was quietly failing open on a
+sparse-returning driver, it starts locking; a `requiredWhen` in the same position
+starts rejecting writes that leave the field empty. That is the declaration being
+enforced as written.
--- a/packages/objectql/src/engine-readonly-when-parent.test.ts
+++ b/packages/objectql/src/engine-readonly-when-parent.test.ts
@@ -16,6 +16,14 @@
 // The record-scoped contrast the issue drew — `showcase_invoice.tax_rate` with
 // `readonlyWhen: record.status == 'paid'`, which worked all along — is pinned in
 // the same file so a future change cannot fix one by breaking the other.
+//
+// #6457 extends this suite rather than starting its own, because what it changes
+// is the same binding this file already owns: the header the engine resolves is
+// now TOTAL over the MASTER object's declared fields, so a `parent.<field>`
+// predicate no longer depends on which columns the driver echoed back. Only the
+// MIDDLE row of the issue's verdict table moves (resolved-but-sparse: fail-OPEN
+// ⇒ evaluated); the fail-CLOSED line above (unresolvable ⇒ LOCKED) is asserted
+// unchanged, on the same fixtures, in the same block.
 
 import { describe, it, expect, beforeEach } from 'vitest';
 import { ObjectQL } from './engine.js';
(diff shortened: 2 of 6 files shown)


os-zhuang:
## Reverse verification — predicted vs actual

Prediction written **before** the mutation (recorded in the working notes, reproduced verbatim below), then the mutation applied in a **separate throwaway worktree** so the branch tree was never touched: both materialisation calls removed (`resolveMasterDetailParent` returns the raw row; `resolveMasterDetailParents` stores the raw row in `byId`), tests and fixtures untouched.

**Predicted: 11 RED — 5 `readonlyWhen`, 6 `requiredWhen`, 0 `rule-validator`.**
**Actual: `Tests 11 failed | 184 passed (195)` — `Test Files 2 failed | 1 passed (3)`.**

Every test identity and every failure signature matched the prediction. No unpredicted red, no predicted red that stayed green.

### `engine-readonly-when-parent.test.ts` — predicted 5, actual 5

| test | predicted signature | actual signature |
|---|---|---|
| ROW 2 — THE FIX: a header missing the key now LOCKS instead of failing open | `toMatchObject` — `locked_until_status: 'forged'` vs expected `'kept'` | `expected { id: 'line_sparse', …(7) } to match object { locked_until_status: 'kept' }` ✅ |
| ROW 2: the strip is reported to the caller as `readonly_when` (#3407) | `toEqual` — received `[]` | `expected [] to deeply equal [ { …(3) } ]` ✅ |
| a sparse header that answers FALSE allows the change — by a VERDICT, not by a fault | `expect(true).toBe(false)` — the fail-open warning reappears | `expected true to be false` ✅ |
| CONSEQUENCE: `has(parent.<declared>)` is uniformly TRUE | `toMatchObject` — `has_guard: 'forged'` vs `'kept'` | `expected { id: 'line_sparse', …(7) } to match object { has_guard: 'kept' }` ✅ |
| BULK: the batch path materialises its headers too | `toMatchObject` — `locked_until_status` written | `expected { id: 'line_sparse', …(7) } to match object { locked_until_status: 'kept' }` ✅ |

Predicted GREEN and green: ROW 1, ROW 3 (unresolvable ⇒ LOCKED), BOUNDARY (undeclared key ⇒ fail-open either way), BULK under a header that carries the key, BULK unresolvable ⇒ LOCKED, header non-mutation, read-count.

### `engine-required-when-parent.test.ts` — predicted 6, actual 6

| test | predicted signature | actual signature |
|---|---|---|
| ROW 2 — THE FIX: an INSERT under a sparse header is now REJECTED | `expect(threw).toBe(true)` receives `false` | `expected false to be true` ✅ |
| ROW 2: an UPDATE that nulls the field under a sparse header is REJECTED | same | `expected false to be true` ✅ |
| ROW 2: the refusal is an EVALUATED verdict — the unbound-root diagnostic is absent | `expect(true).toBe(false)` on `requiredWhen for 'reason' failed to evaluate` | `expected true to be false` ✅ |
| judges a REPOINT onto a sparse header against the header it lands on | `expect(false).toBe(true)` (repoint accepted) | `expected false to be true` ✅ |
| BULK: the batch path materialises its headers too, per matched row | `expect(false).toBe(true)` (batch accepted) | `expected false to be true` ✅ |
| BULK INSERT: one batched header read still materialises every header | `expect(false).toBe(true)` (insert accepted) | `expected false to be true` ✅ |

Predicted GREEN and green: ROW 1, ROW 2-accepted-when-supplied, ROW 3 (fail-OPEN + names the unbound root), ADR-0113 legacy row rests, header non-mutation, and every pre-existing #4977 test.

### `validation/rule-validator.test.ts` — predicted 0, actual 0

The file passed in full under the mutation. That is the measured proof of the claim in the PR body: the strip's own contract did not move, which is why PR #6454's middle-row pin was **re-annotated rather than flipped**. A flip there would have been a false record — it would have asserted a behaviour change in a function this PR does not change.

Note the direction of the evidence for the hard constraint (#4889): the unresolvable-header rows are green **on both sides** of the mutation. They are not proven by this mutation and are not meant to be — they are pinned positively by their own tests (`ROW 3`, `BULK: an unresolvable header still LOCKS the batch`), asserted on the `reads 'parent'` + LOCKED **fault channel** rather than on the write outcome, precisely because rows 2 and 3 can share an outcome and differ only in diagnosis.

## Gate run

Enumerated fresh from `.github/workflows/lint.yml` and run one by one against the final merged tree, after a full `turbo run build --filter='!@objectstack/docs' --concurrency=2`:

**59/59 PASS, 0 FAIL** — `pnpm lint`; all 41 `check:*` targets (including `check:type-check-debt`, `check:type-check-coverage`, `check:engine-double-contract`, `check:driver-conformance`, `check:driver-memory-census`, `check:stall-guard`, `check:query-options-erasure`); all 13 `@objectstack/spec` gates plus `tsc --noEmit`; `@objectstack/lint check:doc-formula-expressions`; `downstream-contract` and `examples` typechecks; and `turbo run typecheck` across `packages/*`, `packages/*/*`, `apps/*`.

Package suite: `@objectstack/objectql` — **149 files / 2598 tests, all passing**, re-run after the `origin/main` merge (which brought in #6766's CEL front-end convergence, directly upstream of every predicate in this change).

No new fake engine was introduced, so `assertEngineDeleteDispatch` does not apply here — both suites drive the **real** `ObjectQL` through a fake **driver**, and `check:engine-double-contract` passes.

---

Issue #6815, opened 2026-08-08
[spec/drivers] `AggregationNode.distinct` is honoured by the in-memory fallback and ignored by every SQL face — one query, two numbers (ADR-0049)


os-zhuang:
Filed unassigned from the #6409 lane. Found while lowering `count_distinct`; adjacent to it, not created by it.

## The finding

`AggregationNodeSchema` (`packages/spec/src/data/query.zod.ts`) declares a per-aggregation flag:

```ts
distinct: z.boolean().optional().describe('Apply DISTINCT before aggregation'),
```

Exactly one consumer reads it. `packages/objectql/src/in-memory-aggregation.ts`:

```ts
const values = collectValues(rows, field, !!agg.distinct);
```

Nothing else in the repo does — measured by grepping every consumer of `aggregations[]`:

| Face | Reads `agg.distinct`? |
|---|---|
| `objectql` in-memory fallback | **yes** — deduplicates before applying the function |
| `driver-sql` `SqlDriver.aggregate` | no |
| `driver-turso` `RemoteTransport.aggregate` | no |
| `driver-mongodb` `buildAggregationStage` | no |
| `driver-memory` `computeAggregate` | no |
| `service-analytics` `AGGREGATE_SQL` | no |

So `{ function: 'sum', field: 'amount', distinct: true }` returns a **deduplicated** sum when the engine falls back in memory and an **ordinary** sum on every SQL datasource. Same query, two numbers, chosen by which backend answered — the divergence class #6203 and #5907 each closed on the aggregate axis, still open on this key. And unlike those, the wrong answer here is a plausible number rather than a refusal, so nothing surfaces it.

The key was NOT covered by #4286, which swept the request surface: that issue dispositioned `QueryAST.distinct` (the query-level one, since retired) and `AggregationNode.filter` (marked `[EXPERIMENTAL — not enforced]`). `AggregationNode.distinct` is neither — it is declared plainly, has a real consumer, and reads as supported.

## Why it is worth a decision now rather than later

#6409 lowered `count_distinct` on the SQL family, which changes what this key looks like to an author. `count_distinct` is now a working, portable, deduplicated count; `distinct: true` beside it is the affordance that reads as "…and the same for `sum`/`avg`". One of the two now works everywhere and the other works in one place, which is a sharper false affordance than it was yesterday.

## Two legs, per ADR-0049

[...]

Issue #6814, opened 2026-08-08
[drivers] driver-memory / driver-mongodb disagree with the SQL family on `count_distinct` — the two AGGREGATION_CASES cells #6409 left open


os-zhuang:
Filed unassigned from the #6409 lane (the enforce half of #6188). Recording, not claiming — both packages are inside the **#5499 investment freeze** and #6409's ruling put them explicitly out of scope. This issue exists so the two open cells of the new `AGGREGATION_CASES` matrix column have a measured verdict behind them instead of an omission.

## Context

#6409 lowered `count_distinct` to `COUNT(DISTINCT x)` on the SQL family and introduced `AGGREGATION_CASES` / `AGGREGATION_ROWS` (`@objectstack/spec/data`) — the shared standard the aggregate lowerings are now held to. Three of the five drivers were enrolled by that PR (`driver-sql`, `driver-turso` REMOTE, `driver-sqlite-wasm`). The other two carry DEBT rows in `scripts/check-driver-conformance.mjs` pointing here.

The standard's answer for `count_distinct` is **distinct NON-NULL values of the target column** — what `COUNT(DISTINCT col)` computes on SQLite, PostgreSQL and MySQL alike, and what `objectql`'s in-memory fallback (`in-memory-aggregation.ts`, `new Set(values.filter(v => v != null)).size`) and `service-analytics`'s `AGGREGATE_SQL` already give.

## What each package answers today

Read from the source on `origin/main` @ `6de592c`; not executed against the fixture, since neither cell has a suite (that is the debt).

### 1. `driver-memory` — the DATA face answers `null`, silently

`MemoryDriver.computeAggregate` (`packages/drivers/driver-memory/src/memory-driver.ts`) has arms for `count`, `sum`, `avg`, `min`, `max` and then:

```ts
default:
    return null;
```

There is no `count_distinct` arm. So `aggregate()` with `{ function: 'count_distinct', field: 'stage', alias: 'n' }` resolves with `{ n: null }` — no error, no log, no refusal. Against `AGGREGATION_ROWS` the standard says `2`.

This is the worse half of the pair: a wrong ANSWER rather than a wrong number. It is also the `default:`-arm shape #4157 and the `aggregation-lockstep` guard in `service-analytics` exist to prevent one layer up, arriving here through a different door.

[...]

Issue #6188, opened 2026-08-07
[spec] AggregationFunction 声明 8 个,SQL 族只实现 5 个 —— count_distinct / array_agg / string_agg 按 ADR-0049 enforce-or-remove 定去留


os-zhuang:
从 #5907 分流而来(分诊 2026-08-06 明确:运行时说实话归 drivers 车道、声明面去留归 spec 座位,⛔ 不夹带)。

## 事实

`packages/spec/src/data/query.zod.ts:108` 的 `AggregationFunction` 声明 8 个:

```
'count', 'sum', 'avg', 'min', 'max', 'count_distinct', 'array_agg', 'string_agg'
```

后 3 个 —— `count_distinct` / `array_agg` / `string_agg` —— **没有任何 SQL 后端编译**:`SqlDriver.mapAggregateFunc`(sql-driver.ts:7071 附近)与 `RemoteTransport.aggregate`(remote-transport.ts:726 附近)各自只编译前 5 个,后 3 个落同一句拒绝。非 SQL 族里 `driver-mongodb`(mongodb-aggregation.ts:139/144/147)与 `driver-memory`(memory-analytics.ts:866,`count_distinct`)实现了其中若干,**且各后端实现的子集互不相同**。

即 declared ≠ enforced,且「哪个后端能算什么」对作者不可预测。

## 需要一次裁决(spec 座位)

- **A 保留声明**:承认它是「部分后端支持」的能力面,则需要一个作者可查的能力矩阵(哪些后端支持哪些聚合),否则作者只能靠试;
- **B 按 ADR-0049 摘除**:三个键从 `AggregationFunction` 退役(走 ADR-0087 墓碑与生成物流程),已实现它们的 mongodb/memory 侧按各自冻结/退役状态处置;
- **C 保留但标注实验/后端限定**:需要 spec 侧有表达「按后端可用性」的位,#5240 曾拒绝为共享表临时发明例外位,此路需先设计。

drivers 侧的运行时半边(让第 2 类拿到能力缺口信封而不是被说成打错字)在 **#5907** 独立进行,不阻塞本单;本单的结论反过来会决定 #5907 那条措辞将来是长期存在还是随退役消失。

关联:#5907(运行时半边)、#5345(driver-memory 上同构的「协议没有」vs「本面降不下去」分线先例)、ADR-0049 enforce-or-remove、ADR-0087。

os-project-manager:
[...]

Issue #4286, opened 2026-07-31
[P2] data: `QueryAST` declares 12 members no executor runs — the liveness ledger governs metadata types, not the request surface


os-zhuang:
Follow-up to #4196 (merged as #4268). That issue removed one declared-but-inert member from `FieldNode`. Applying the same method to the rest of the **request surface** — `QueryAST` and its satellites — turns up 12 more, and, more importantly, the reason none of the existing gates found them.

Refs #4171, #4196, ADR-0049 (enforce-or-remove), ADR-0078 (no-silently-inert), ADR-0087 (conversion/tombstone).

## The inventory

`BaseQuerySchema` (`packages/spec/src/data/query.zod.ts:581`) declares 16 members. Five of them, plus six sub-flags of `search` and one of `AggregationNode`, are never executed. `engine.ts` contains **zero** reads of any of them on the query path.

**Tier A — a public SDK method mints them.** This is strictly worse than the #4196 case: `FieldNode`'s object form at least had no producer.

| Member | Declared | Producer | Executor |
|:--|:--|:--|:--|
| `cursor` | `query.zod.ts:601`, re-declared on `EngineQueryOptionsSchema` (`data-engine.zod.ts`) | `QueryBuilder.cursor()` — `client/src/query-builder.ts:294`; serialized to the wire at 5 sites in `client/src/index.ts` | none — no driver implements keyset pagination |
| `distinct` | `query.zod.ts:619`, also on `EngineQueryOptionsSchema` | `QueryBuilder.distinct()` — `client/src/query-builder.ts:302` | **not** `SELECT DISTINCT` — see below, it does something else |
| `search.fuzzy` | `query.zod.ts:524` | `QueryBuilder.search()`'s own options type — `client/src/query-builder.ts:286` | none — expansion reads only `query` + `fields` |

**Tier B — declared, no producer, no executor.**

| Member | Declared |
|:--|:--|
| `joins` | `query.zod.ts:604` |
| `having` | `query.zod.ts:613` |
| `windowFunctions` | `query.zod.ts:616` |
| `search.operator` / `boost` / `minScore` / `language` / `highlight` | `query.zod.ts:525-529` |
| `AggregationNode.filter` | `query.zod.ts:131` |

## Four things the docs table does not say

[...]

PR #6816, opened 2026-08-08
feat(drivers): lower `count_distinct` on the SQL family — the enforce half of the #6188 ruling (#6409)

### LINKED ISSUE #6409: drivers: implement `count_distinct` in the SQL family — enforcement half of the #6188 ruling
Follow-up filed from the 2026-08-07 maintainer ruling on #6188 (decision batch accepted via the PM session). Unclaimed; no labels applied — routing belongs to the triage seat.

## Ruling context

#6188 was ruled as a split: `array_agg` / `string_agg` retire from `AggregationFunction` (that card keeps the spec retirement half), while **`count_distinct` stays declared and gets enforced** — this card is the enforcement half.

## Scope

Implement `count_distinct` in the SQL family:

- `packages/drivers/driver-sql` — `SqlDriver.mapAggregateFunc` (currently rejects it alongside the two now-retiring keys): lower to `COUNT(DISTINCT column)`.
- `packages/drivers/driver-turso` (or wherever `RemoteTransport.aggregate` lives per current main) — same lowering on the remote path.

Out of scope: `driver-mongodb` / `driver-memory` (inside the #5499 investment freeze; their existing partial implementations are untouched either way).

## Acceptance

- An analytics query using `count_distinct` compiles and returns correct values on the SQL family (dedup semantics: distinct non-null values of the target column).
- A conformance cell covering `count_distinct` is added to the shared cases so the two SQL lowerings are pinned against each other (multi-face component rule: shared coverage, not a standalone test file).
- The rejection message for unknown aggregation functions no longer names `count_distinct` as unsupported.

## Ordering

Independent of the #6188 retirement half (different keys, different files); no Blocked-by needed. If the retirement PR touches the same `mapAggregateFunc` switch, serialize per same-file discipline.

Refs: #6188 (ruling), #5907 (runtime capability-gap envelope — its wording loses these keys as subjects once both halves land), #5499 (freeze boundary).

os-zhuang:
Fixes objectstack-ai/objectstack#6409

## Premise, re-verified before implementing

The issue's lead held. On `origin/main` @ `6de592c`, both SQL faces refused `count_distinct` with `NOT_IMPLEMENTED` / 501, and the two pin files recorded it as the sole inhabitant of #5907's class 2. The issue's line anchors had drifted (as the dispatch predicted) — `mapAggregateFunc` is at `sql-driver.ts:8035` on this branch, not `:7071` — but the shape was exactly as described.

One thing the issue's framing understates and this PR had to handle: `AggregationFunction` declares six values and `SQL_AGGREGATE_FUNCTIONS` compiled five, so landing the sixth **empties class 2 entirely**. That is not a side effect to tidy — it is a decision about #5907's capability-gap producer, taken explicitly below.

### The lowering, on both faces in one change

- `packages/drivers/driver-sql/src/sql-driver.ts` — `SQL_AGGREGATE_FUNCTIONS` now maps each name to a `SqlAggregateLowering` record (`{ sql, distinct }`) instead of a bare SQL function name, and `SqlDriver.aggregate` emits `count(distinct ??)` for the distinct entry. The column is still bound as a knex **identifier**; the keyword is fragment text, never a binding.
- `packages/drivers/driver-turso/src/remote-transport.ts` — the twin, `RemoteTransport.aggregate`, emitting `count(distinct "column")` through its own hand-built SELECT list.

Both faces, one change, deliberately: `TursoDriver` picks between them from `url`, so a lowering that landed on one alone would be #6203's shape again — one query, two answers, decided by a connection string.

`mapAggregateFunc`'s return type moved from `string` to the record. It is `protected` and had exactly one caller in tree; every entry answering `{ sql, distinct: false }` compiles to byte-identical text to what the `string` return produced.

### Semantics: distinct NON-NULL values

Measured rather than assumed, against a real better-sqlite3 database (local face) and SQLite behind the libsql interface (remote face): `COUNT(DISTINCT col)` excludes NULLs on both. That matches the standard's definition of `COUNT` over non-null values (PostgreSQL, MySQL alike), and matches what `objectql`'s in-memory fallback and `service-analytics`'s `AGGREGATE_SQL` already answer. The fixture pins it: `count(*)` is 6, `count(stage)` is 4, `count_distinct(stage)` is 2 — three different numbers over one column, so a lowering that collapsed into either neighbour fails on a value.

### `field` is now required for `count_distinct` — a new refusal

`AggregationNodeSchema` makes `field` optional because `COUNT(*)` is a real spelling and the schema cannot say "optional for this function, required for that one". So `{ function: 'count_distinct', alias: 'n' }` parses, reaches the driver, and asks for `COUNT(DISTINCT *)` — a syntax error in every dialect. Both faces now refuse it up front with `INVALID_QUERY` / **400** and a message naming the fix, rather than emitting it and letting a dialect error arrive as an opaque 500 (`mapDataError`'s default branch — the #1116/#1117 gap #5907 closed at this same door).

Class **1**, not class 2, and the messages say so: the function *is* compiled here; it is this aggregation node that is malformed. Plain `count` with no `field` still means `COUNT(*)`, pinned by a control on each face.

### Class 2 is now empty — and its producer is kept, on purpose

With `count_distinct` lowered, `AggregationFunction` and each face's compiled vocabulary are the **same set**, so `uncompilableAggregateFunctionError` has nothing to produce.

Deleting it as dead code was considered and rejected. That branch is not an unenforced *declaration* — the ADR-0049 shape — it is the **classifier** deciding which of two truths a future name is told. Removing it does not remove the condition; it makes `refuseAggregateFunction` answer 400 for the first function a later spec bump adds, telling the author of a correctly-spelled name that the protocol has no such function. That is precisely the misreport #5907 exists to prevent, and it would land in the window ADR-0049's enforce leg opens on purpose — the window #6188 just opened for this very name. Cost of keeping: one unreachable branch. Cost of dropping: a wrong answer at the moment the vocabulary grows.

Both pin files assert the emptiness **positively** (`the declared-but-uncompiled set is EMPTY`), so a spec addition goes red there rather than quietly leaving the block covering nothing.

### The rejection message

Both faces build their `Compiled here:` list from the lowering table, so it now reads `count, sum, avg, min, max, count_distinct`. Asserted off a real refusal on each face, not off a literal — the acceptance criterion "the rejection message no longer names `count_distinct` as unsupported", made checkable.

## Shared conformance — the multi-face rule

New shared case-set: `packages/spec/src/data/aggregation-conformance.ts` — `AGGREGATION_ROWS` (six rows, two groups, a nullable duplicate-bearing column) and `AGGREGATION_CASES` (twelve cases covering the whole compiled vocabulary, ungrouped and grouped). Not a standalone test file: the two SQL lowerings are pinned **against each other**, on values, which is the only instrument that can see this class of bug — every wrong `count_distinct` lowering still emits valid SQL and returns a plausible number.

Driven by three suites, all executing against a real engine:

| Face | Suite | Engine |
|---|---|---|
| `driver-sql` | `sql-driver-aggregation-conformance.test.ts` | better-sqlite3 |
| `driver-turso` REMOTE | `turso-remote-aggregation-conformance.test.ts` | `makeLibsqlSqliteStub` |
| `driver-sqlite-wasm` | `sqlite-wasm-aggregation-conformance.test.ts` | sql.js dialect |

`scripts/check-driver-conformance.mjs` classifies the new column (it failed CLASSIFIED until it did — the gate working as designed). The two remaining cells carry **honest DEBT rows**, measured from the source, not flipped and not faked:

- **`driver-memory`** — `MemoryDriver.computeAggregate` has no `count_distinct` arm; the `switch` falls to `default: return null`, so the aggregation resolves with `{ n: null }`. A wrong *answer*, not a wrong number.
- **`driver-mongodb`** — `count_distinct` lowers to `$addToSet` and is sized by `.length`, so an explicit `null` counts as a distinct value: 3 where the standard says 2.

Both are inside the **#5499 freeze** and #6409's ruling put them explicitly out of scope; their partial implementations are untouched. Filed as #6814 so the rows point at a tracked decision rather than at a note.

## Pins flipped, and what replaced them

`sql-driver-out-of-contract-aggregate-function.test.ts` and `remote-transport-aggregate-function-refusal.test.ts` both recorded `count_distinct -> 501`. Flipped, never merely deleted — each old assertion was replaced by the new substance: the empty-set assertion, the emitted `count(distinct …)` / the computed value, and the refusal wording that now lists the function as compiled. The remote file's parity block loses `count_distinct` as a refusal case and gains two: both faces **answer** it identically, and both **refuse** the field-less spelling identically (`INVALID_QUERY`/400, one wording).

Class 1 is untouched throughout — `median`, `array_agg`, `string_agg` and every miscased spelling are still undeclared names, and their refusals stay verbatim. A pin is flipped only where the fact under it moved.

## Reverse verification — predicted, then measured

Two reverts on both faces, direction written down before running.

**(A) the `count_distinct` entry deleted** (the pre-#6409 state). Predicted: the `count_distinct` cases fail by *throwing* 501, never on a value. Measured: local **5 failed / 11 passed** of 16, remote **6 failed / 11 passed** of 17 — every value case on the thrown 501, none on a number. One failure was *not* predicted and is recorded rather than tidied: the new field-less refusal case went red on `expected 'NOT_IMPLEMENTED' to be 'INVALID_QUERY'`, which is what proves the two refusals are distinguishable rather than interchangeable.

**(B) `distinct` flipped to `false`** — the copied-neighbour mistake, and the one a review that only checked the two faces' tables have the same *keys* would pass. Predicted: failures on VALUES, with `count_distinct(score)` staying green because that column has nothing to dedup. Measured: local **4 failed / 12 passed**, remote **5 failed / 12 passed**, on `expected 4 to be 2` and `west 3 vs 2` — the same two wrong numbers on both faces, and `count_distinct(score)` green throughout, exactly as predicted. That agreement is the pairing doing its job.

(B) is the direction this table exists for; (A) is reachable by any test that merely calls the function.

## Spec-side wording

- `AggregationFunction`'s JSDoc: `count_distinct` no longer "leads its implementation"; the NULL semantics and the required `field` are stated.
- `AGG_RETIRED_MIDDLE` (the live `array_agg`/`string_agg` parse-error prescription) said the two SQL faces "lower the same **five** functions", which this PR makes false. Corrected to "the same set" — **one character shorter**, deliberately: `packages/rest`'s `analytics-dataset-refusal-envelope.test.ts` asserts that the actionable "Delete the aggregation" survives the `detail` truncation window, and a longer edit pushed it out. Caught by that test, not by review.

The #6188 conversion/migration registry entries are **not** touched: they narrate the state at that ruling and already say the SQL lowering "follows on its own card". Historical narration, accurate as history.

---

## os-dev report

**Issue:** #6409 — implement `count_distinct` in the SQL family (the enforcement half of the #6188 ruling)
**PR:** #6816 (draft) · branch `claude/issue-6409-count-distinct-sql-family` · head `08f803e`
**premise_still_valid:** **true**
**CI:** **green** — all 25 checks completed success or skipped on `08f803e`.

*(The sections above are the detail; this is the structured summary, posted identically as a comment on #6409.)*

### Gates and tests

- **The gate list enumerated from `.github/workflows/lint.yml`, run step by step** (not from memory): **58 / 58 PASS**. Workspace built first so the `check:i18n` / `check:i18n-coverage` / `check:app-nav-i18n` trio ran for real; each step's exit status read directly, never through a pipe.
- Two failed on the first pass and both were **fixed here, not worked around**: `check:driver-conformance` (CLASSIFIED — the new case-set was unregistered) and `check:api-surface` (five new exports; snapshot regenerated with `gen:api-surface`).
- Tests: `driver-sql` 1095✓/48 skipped · `driver-turso` 902✓ · `driver-sqlite-wasm` 302✓ · `driver-memory` 540✓ · `driver-mongodb` 214✓/143 skipped · `spec` 8861✓ · `objectql` 2656✓ · `service-analytics` 1415✓ · `metadata-protocol` 740✓ · `rest` 1097✓.
- Workspace-wide `turbo run typecheck` (120 tasks) and both enumerated `turbo run build` filters: green.

**One CI round was red, and it is worth recording why the local gate list did not catch it.** `TypeScript Type Check` failed on `9b9b429` with `src/sql-driver-aggregation-conformance.test.ts(121,63): error TS2353: 'sort' does not exist in type 'DriverQuery'` — the fixture read-back used `sort` where `QueryAST` declares `orderBy`. My enumeration from `lint.yml` transcribed the `check:*` steps but **omitted the workspace-wide `turbo run typecheck` / `turbo run build` steps in the same workflow**, and the per-package typecheck I *had* run was run before that suite was written. Fixed in `08f803e`, and the two missing commands were then run locally and are green. The lesson is the enumeration's, not the type error's: "every `check:*` step" is narrower than "every step".

### Out-of-scope findings (search-first dedup, filed unassigned)

- **#6814** — `driver-memory` / `driver-mongodb` disagree with the SQL family on `count_distinct`; the two DEBT cells above, with the measured evidence for each. #5499-frozen, so this is a decision about the freeze rather than about the code.
- **#6815** — `AggregationNode.distinct` is honoured by `objectql`'s in-memory fallback and ignored by **every** SQL face, so `{ function: 'sum', distinct: true }` returns a deduplicated sum in memory and an ordinary sum on SQL: one query, two numbers, no diagnostic. Searched first — **not** covered by #4286's request-surface sweep, which dispositioned `QueryAST.distinct` and `AggregationNode.filter`. Both ADR-0049 legs argued, no recommendation offered.
- Noted inside #6815's scope note rather than filed separately: `sum`/`avg`/`min`/`max` written with no `field` still emit `sum(*)` and die as a dialect syntax error with no ADR-0112 envelope — same class as the refusal added here. This PR deliberately did not widen its refusal surface to cover it.

### Open questions for the maintainer (三轴自裁, recorded not blocked)

1. **The kept-but-unreachable class-2 producer.** Judged: keep, on the argument above. If the maintainer prefers ADR-0049's letter over the misreport risk, deleting `uncompilableAggregateFunctionError` on both faces is a small follow-up — the positive emptiness assertions are already in place to make the deletion honest.
2. **`AGGREGATION_CASES`' enrolment boundary.** Enrolled the three SQL-family drivers. `objectql`'s in-memory fallback is a fourth lowering that *agrees* with the table and is **not** frozen; it is recorded in the case-set as a candidate rather than a debt, because reaching it needs the engine rather than a driver. Enrolling it is a judgement call I left to the seat that owns the objectql lane.
3. **`service-analytics`'s `AGGREGATE_SQL`** is a fifth `COUNT(DISTINCT …)` lowering (the Cube face). Left unenrolled on scope discipline — #6409 named the driver family — but it is the obvious next column if the table is meant to cover every SQL-emitting face.

### Not done, and why

Nothing in the issue's scope was skipped. `driver-memory` / `driver-mongodb` are out of scope by the ruling and stayed untouched; their verdicts are DEBT rows, not changes. `content/docs/releases/` untouched. The #6188 conversion/migration registry entries were left alone — accurate as history.


os-zhuang:
**ACCEPT** — drivers lane, PM dispatch loop. Step-7 gate passed on all six criteria, each verified against the diff and the JOB logs rather than off the report's own claims.

| Criterion | Evidence taken independently |
|---|---|
| Both SQL faces, one change | `sql-driver.ts` +144 **and** `remote-transport.ts` +115 in the same change — not one face with the other left to drift on a connection string. |
| The flipped refusal pin bears new substance | The two pin files add **+82 / +137** lines against 42 / 27 removed. The old `count_distinct → 501` assertions are replaced by a *positive* empty-set assertion, the emitted `count(distinct …)`, and value pins — not deleted into silence. |
| Frozen packages untouched | Zero `driver-memory` / `driver-mongodb` source files in the diff. |
| Per-dialect NULL semantics measured, not assumed | The fixture pins `count(*)`=6, `count(stage)`=4, `count_distinct(stage)`=2 over one nullable column, so a lowering that collapses into either neighbour fails on a value; harnesses are required to seed the nulls **as** nulls. |
| Honest DEBT, no fake green | Both frozen cells are `DEBT`, labelled "read from source, not executed", pointed at #6814. Neither flipped. |
| Rejection asserts code AND status | `expect(remote.code).toBe('INVALID_QUERY')` + `expect(remote.status).toBe(400)`, plus cross-face equality — never a bare `toThrow`. |

**CI gated on the JOB conclusions:** 26 check runs on `08f803e`, all `success` or `skipped` — including `TypeScript Type Check`, all three `Test Core` shards, `Dogfood Regression Gate`, `Temporal Conformance (live PG + MySQL)` and `Check Changeset`.

Two things the seat is recording as credit rather than passing over:

1. **`08f803e` was self-caught.** `sort:` → `orderBy:` — the key `DriverQuery` actually declares. With `sort` the option was silently ignored and the conformance read-back was unordered, so the suite could pass on insertion-order luck. The report names the enumeration error behind it (`check:*` steps transcribed, workspace-wide `turbo run typecheck` omitted) instead of reporting the fix alone. "Every `check:*` step" being narrower than "every step" is the reusable lesson.
2. **Class 2 emptied with its producer kept, argued rather than tidied.** Keeping an unreachable `uncompilableAggregateFunctionError` is the right call: deleting it makes `refuseAggregateFunction` answer 400 to the first correctly-spelled name a later spec bump adds — the exact misreport #5907 exists to prevent. The positive emptiness assertions make that choice re-decidable later.

The three open questions (producer deletion; enrolling `objectql`'s in-memory fallback; enrolling `service-analytics`'s `AGGREGATE_SQL` as a fifth `COUNT(DISTINCT …)` face) are recorded, not blocking, and correctly left to the seats that own those lanes.

Marking ready and enabling auto-merge. The full suite runs again on the speculative merge result in the queue; per #4859 a red queue build gets diagnosed before any requeue, not blind-requeued.

---

Issue #3276, opened 2026-07-19
cli: `OS_DATABASE_DRIVER=memory` is advertised but `serve.ts` has no dispatch branch (falls through to SQLite)


os-zhuang:
The CLI advertises `memory` as a database driver — `--database-driver … | memory` (`packages/cli/src/commands/dev.ts:88`), and `OS_DATABASE_DRIVER=memory` is special-cased in dev (`dev.ts:41`; `resolveDefaultDevDbUrl` at `dev.ts:40-42`).

But `packages/cli/src/commands/serve.ts`'s driver dispatch has **no `memory` branch** (verified by grep). So `OS_DATABASE_DRIVER=memory` falls through to the isDev SQLite default (`resolveSqliteDriver` at `:memory:`), yielding **SQLite-in-memory**, not the mingo `InMemoryDriver`. The `InMemoryDriver` is only reached if *both* native `better-sqlite3` and the wasm fallback fail to load.

**Impact:** declared ≠ enforced — selecting the `memory` driver silently gives a different engine than intended.

**Fix:** add a `driverType === 'memory'` branch in the `serve.ts` dispatch that returns the `InMemoryDriver`.

_Found during the #1880 docs implementation-accuracy audit (PR #3243); out of scope for that docs-only PR._

PR #6910, opened 2026-08-09
fix(spec,cli,runtime,service-datasource)!: one driver vocabulary — `os start` and `os migrate` stop disagreeing (#6345)

os-project-manager:
Fixes #6345

Implements the maintainer's ruling archived in [comment 5229196310](https://github.com/objectstack-ai/objectstack/issues/6345#issuecomment-5229196310) — route C, plus the four calls it settled. The predecessor session's stop report ([5229178650](https://github.com/objectstack-ai/objectstack/issues/6345#issuecomment-5229178650)) is the measurement baseline; **every number below was re-measured on this merged tree**, not carried over.

## The defect, re-measured on the merged tree

One environment variable had two answers. Driving the real entry points — `resolveDriverType` + `resolveStorageDefinition` for `os start`, `resolveStandaloneDatabase` for `os migrate` — **10 of 21 spellings disagreed**:

```
OS_DATABASE_DRIVER=pg OS_DATABASE_URL=postgres://…  os start        → accept (driverId=postgres)
OS_DATABASE_DRIVER=pg OS_DATABASE_URL=postgres://…  os migrate plan → "[StandaloneStack] Unsupported OS_DATABASE_DRIVER value: \"pg\"."
```

`sql`, `wasm`, `wasm-sqlite`, `postgresql`, `pg`, `mysql2`, `mongo`, `mingo`, `in-memory`, `libsql` — CLI accepts, standalone refuses.

**One refinement to the predecessor's table.** It listed `turso` as disagreeing and `libsql` as agreeing. Measured with a matching URL present it is the other way round: `turso` **agrees** (both accept `libsql://…`), and `libsql` disagrees (a CLI-only spelling). The count is 10 either way; the composition differs because turso's asymmetry lives in **fork 2** (no URL), not in the vocabulary.

### Fork 1 — one vocabulary (route C)

`packages/spec/src/data/driver/config-registry.zod.ts`'s flat `Record` becomes one table: `id` / `aliases` / `contractOnlyAliases` / `hasLocalDefault`. `BUILTIN_DRIVER_IDS`, `DRIVER_ID_ALIASES` and `resolveDriverId` are projections of it; `BUILTIN_DRIVER_IDS` keeps its exact tuple type via a homomorphic mapped type, so **the api-surface delta is purely additive — nothing removed or renamed.**

Both hosts now resolve through `resolveDatabaseDriverId`. The standalone `databaseDriver` **config key** accepts the same aliases as the env var, so the fork cannot relocate to inside one host.

**On the ruling's `contractId?` column:** it collapses to nothing once `mongo`→`mongodb` lands and turso gets a contract. A column that would be `undefined` on every row is the inert declaration this repo removes rather than ships, so it is not in the table. The distinction it was meant to carry is carried by `contractOnlyAliases`, which is non-empty.

**`sql`/`wasm` in, `sqlite3`/`better-sqlite3`/`mariadb`/`inmemory` out** — the ruling's union principle applied. The four stay `contractOnlyAliases`: they keep resolving a config contract (dropping that would silently un-validate a stored `driver: 'sqlite3'` row) while `resolveDatabaseDriverId` refuses them as a boot selection.

**One consequence derived rather than found itemised.** An *explicitly named* unknown driver is now refused on the CLI side too. `os dev --database-driver sqlite3` previously fell through to the dev SQLite default and booted in silence, while `os migrate` refused it by name. "Both hosts answer the same for the same input" cannot hold with one side silently booting — so it is treated as priced by the ruling. `''` (nobody chose) keeps its old answer; a non-empty value can only have come from an operator, since URL inference yields a canonical id or `''`.

⚠️ **This new refusal is also what ejected the PR from the merge queue.** See "Queue ejection" below — it is the same change viewed from a consumer nobody had enumerated.

### `mongo` → `mongodb` — the full blast radius, itemised

| face | change |
| :-- | :-- |
| `spec/data/driver/config-registry.zod.ts` | canonical id, `DRIVER_CONFIG_SCHEMAS`, JSON-schema map |
| `spec/data/driver/mongo.zod.ts` | `MongoDriverSpec.id` |
| `spec/conversions/registry.ts` | `DATASOURCE_CONFIG_KEY_ALIASES` key |
| `service-datasource/driver-catalog.ts` | **published `DRIVER_CATALOG.id`** — what Studio writes into `datasource.driver` |
| `service-datasource/default-datasource-driver-factory.ts` | dispatch arm + header |
| `service-datasource/datasource-pool-support.ts` | the pooled-drivers clause in the rejection instruction |
| `runtime/resolve-project-database.ts` | `datasourceUrlOf` case |
| `spec/kernel/manifest.zod.ts` | the driver-id example in a `.describe()` |
| regenerated | `api-surface/`, `authorable-surface/data.json`, `json-schema.manifest/data.json`, `spec-changes.json`, `liveness/datasource.json`, `docs/protocol-upgrade-guide.md`, `content/docs/references/**`, `skills/**/_index.md`, strictness-ledger counts |

**ADR-0087 D2, as ruled — ledger entry + converter, not a D3.** `datasource-driver-mongo-to-mongodb` (`toMajor: 17`, live load path) converts stored `datasource.driver: 'mongo'` → `'mongodb'`, registered in `migrations/registry.ts`'s `conversionIds` with a prose paragraph in the step-17 record. Live rather than retired because `mongo` is still a legal alias.

**Deliberately NOT renamed:** `SqlDialect`'s `'mongo'` member in `data/type-compat.ts`. That names the type system of an *external* schema being introspected, alongside `snowflake` and `bigquery`, and is never a `datasource.driver`.

### `turso` gets a config contract — and who read the `{known:false}`

New `packages/spec/src/data/driver/turso.zod.ts`. Keys drawn from what `TursoDriverConfig` actually **reads**, so closing a gate does not open an ADR-0049 one (`client`, `pool`, `schemaMode`/`readOnly` deliberately absent).

| consumer of the negative answer | before | after |
| :-- | :-- | :-- |
| `DatasourceSchema` → `reportDriverConfigIssues` | no-op | parses a turso `config`; `{ token: … }` now gets a rename hint instead of connecting unauthenticated |
| `datasource-admin-service.assertValidConfig` (Setup wizard) | no-op | same flip, same door |
| `DRIVER_CATALOG` | turso not curated | unchanged — same call as `sqlite-wasm` |
| `driverReadsDeclaredPool` | `true` (unknown-id branch) | `true` (not-rejected branch) — **verdict unchanged** |

**⚠️ A consumer of the same widening that would have been a stop condition had it not been fixable inside the declared surface.** `createDefaultDatasourceDriverFactory().supports()` is `resolveDriverId(id) !== undefined`, and `create()`'s last arm was an **unguarded `memory` fall-through**. Making turso a `BuiltinDriverId` therefore made the open-core factory claim `supports('turso') === true` and then **silently build an `InMemoryDriver`** — a libSQL datasource that accepts writes, reports success and loses everything. Fixed in-surface: an explicit `turso` arm (the same lazy-import shape `mongodb`/`sqlite-wasm` already use), plus `memory` promoted to an explicit arm with a `never`-typed exhaustiveness stop, so the next builtin cannot inherit the trap.

### Fork 2 — the typed refusal, all 8 cells

| selection, no URL | `os start` before | `os migrate` before | now, both |
| :-- | :-- | :-- | :-- |
| `postgres` | `config.url === undefined` → `pg` picks its own localhost:5432 | `file:/data/objectstack.db` | typed refusal |
| `mysql` | `config.url === undefined` | `file:…objectstack.db` | typed refusal |
| `mongodb` | invented `mongodb://localhost:27017/objectstack` | `file:…objectstack.db` | typed refusal |
| `turso` | typed refusal (already, #5602) | `file:…objectstack.db` | typed refusal |

Only the **fallback** rungs are refused on the standalone side (`unified-default` / `legacy-file`). A URL from `--database`, `OS_DATABASE_URL`, `DATABASE_URL`, `TURSO_DATABASE_URL` or the project's declared default datasource is a statement about where the database is, and is honoured as before — `file:` DSN included.

## The pin this card exists for

`packages/cli/src/utils/driver-vocabulary-parity.test.ts` — 49 assertions that **both hosts accept the same alias set for the same input**, driving the real entry points. It lives in `@objectstack/cli` because that is the only package that can import both `@objectstack/runtime` and `@objectstack/spec`.

The fork survived #3276, #5820 and #6265 — each of which shipped a green pin that drove exactly **one** host. That is why this file is the deliverable and not a detail.

## Queue ejection — a poisoned oracle, not a vocabulary drift

This PR was **ejected from the merge queue** ([build 31297062268](https://github.com/objectstack-ai/objectstack/actions/runs/31297062268)) with four failures in `packages/cli/src/commands/database-driver-allowlist.pin.test.ts` — **#6860's pin**, landed on `main` by PR #6913 (`9d425a94d`) **after this branch's base**. Both PRs were green alone; they first met in the queue, because the queue runs the **full** suite while PR CI runs only the affected subset, and that file did not exist on this branch.

**The obvious hypothesis was wrong and was discarded by measurement.** The natural read is "the canonical set moved (`mongo`→`mongodb`, turso became builtin) while the hand-written oclif allowlist stayed on old spellings." It did not: **#6860 already landed the canonical seven**, including `mongodb`, `sqlite-wasm` and `mysql`. Editing the allowlist would have been a no-op — and `start.ts` / `dev.ts` are correspondingly **untouched in this PR**.

**The real defect is narrower and belongs to this PR.** #6860's pin derives the driver-kind side by scanning `storage-driver.ts` with a deliberately over-broad net and using `resolveStorageDefinition` as an **oracle**, reading `err.driverType` out of `UnsupportedDriverError` to decide what is real. This PR's *new* refusal of an explicitly-named unknown driver throws that **same error type**, carrying the operator's raw token in `driverType` — so every junk candidate now "resolved". The derived set went **7 → 13**, gaining `safe`, `on-disconnect`, `factory`, `function`, `string`, `default`, `better-sqlite3`. **No allowlist could have equalled that set.**

**Fix: give the error a discriminator, not the test an opinion.** `UnsupportedDriverError` gains `recognized`, defaulting `true` so pre-existing call sites (turso-with-no-URL) keep their meaning; the unknown-spelling throw passes `{ recognized: false }`; the pin returns `null` for that case. The discriminator lives **on the error rather than in the test**, so the pin keeps asking the resolver what a token means instead of growing its own vocabulary.

⛔ **The assertion was not relaxed.** Both sides are still derived and still required to be equal — the *derivation* was repaired. The pin's own docstring already defines the category ("a recognized kind that cannot become a definition still counts as recognized"); an unrecognized token was never in it.

Proof it was not silenced:

| perturbation | result |
| :-- | :-- |
| drop `mongodb` from `start.ts`'s allowlist | **3 red**, incl. "start and dev offer the same kinds as each other" |
| revert `recognized: false` | **4 red** — the exact queue signature |
| both restored | **10/10 green** |

**Lesson recorded for the repo:** a derivation that reads a value out of an exception inherits every future meaning of that exception. #6860's pin was well built and its scope note explicitly anticipated this card's alias convergence without prejudging it — what it could not anticipate was the resolver growing a *second* refusal shape.

## Reverse verification (predictions written before running)

| revert | predicted | measured |
| :-- | :-- | :-- |
| standalone stops reading the shared table | RED, the 10 spellings | **RED — exactly those 10** (+2 named cases) |
| standalone fork-2 refusal | RED in 4 cells | **RED, 4 cells** |
| CLI fork-2 refusal | RED in 3 (turso already refused) | **RED, 3** — turso's cell A stayed green, an honest pre-existing pass |
| unregister the D2 conversion | RED | **RED, 4** |
| "unconverted `mongo` row still works" | **green both ways — no red** | **green** — and that is the point: the rename was made non-breaking at the resolver |
| CLI unknown-driver refusal | "≥1 red" | **the parity test stayed GREEN** — see below |

**The reverse verification found a real hole in this PR's own pin.** The parity test's REFUSE cases ran only `isDev: false`, where the pre-#6345 CLI *already* returned `null`; the silent-SQLite fallback only happens in dev. Reverting the CLI half left the pin green — only `storage-driver.test.ts` caught it. The pin now runs both modes and produces **6 reds** under the same revert. Recorded because a pin that passes under the revert it exists to catch is worth more as a finding than as a green tick.

## Gates, on the re-merged tree

`pnpm lint` ✅ · typechecks for spec / cli / runtime / service-datasource ✅ · `check:api-surface` · `check:generated` · `check:authorable-surface` · `check:liveness` · `check:strictness-ledger` · `check:spec-changes` · `check:upgrade-guide` · `check:spec-parsed-alias` · `check:adr-0087-registration` (3 declared-breaking changesets, each `registered datasource-driver-mongo-to-mongodb`) ✅

**Full package suites** (not filtered subsets — the filter is what let the first round's spec failure through): **cli 101 files / 1095 tests**, **spec 350 files / 9094 tests**, runtime standalone 87, service-datasource 264 — all green. The whole-monorepo `pnpm test` was **not measured**; the box is shared and returns pure timeouts under load. Reporting the per-package runs as the real signal rather than dressing an unmeasured suite as green.

### Regenerated baselines — the §11 trap fired and was caught

`spec-changes.json` / `protocol-upgrade-guide.md`: the repo's own `os-regen` merge guard flagged both as merged-without-a-text-merge; regenerated from the merged tree before committing, marker cleared. A subsequent `rm -rf packages/spec/dist` + full rebuild + regeneration produced **zero diff**, confirming the artifacts were already correct.

`api-surface/contracts.json`: an earlier post-merge `gen:api-surface` ran against a `dist/` built **before** the merge and silently dropped six exports main had added (`StorageListPage`, `encodeStorageListCursor`, …) — exactly AGENTS.md §11's trap. Caught by diffing the regenerated snapshot against `origin/main`'s rather than trusting the generator's exit code. Final export-removal diff across every `api-surface/*.json` vs `origin/main`: **REMOVED: NONE**.

## Changesets

Three `major`, each carrying `registered datasource-driver-mongo-to-mongodb`:

- **spec** — `BuiltinDriverId` loses a member; compile-time break even though runtime stays compatible via the alias.
- **service-datasource** — the published `DRIVER_CATALOG.id` value changes; failure mode is a silently-unmatched dropdown.
- **cli + runtime** — boots that used to start now refuse (no-URL remote drivers; explicitly-named unknown drivers).

## Not touched, on purpose

`packages/cli/src/commands/start.ts`, `dev.ts` and `content/docs/deployment/cli.mdx` — #6860's surface; the oclif `options:` allowlist is **untouched**, and the queue investigation confirmed it needed no edit. `content/docs/releases/` — never touched (repo-wide prohibition).

## Follow-ups (filed here, not done)

1. **Derive the `--database-driver` oclif allowlist from the shared table.** Now that the single table exists, #6860's hand-added entries can become a derivation — the "one vocabulary, one derivation" half of ruling item 3. **Still open**: the pin now enforces structurally what the derivation would enforce, and expanding a 41-file PR during a queue incident was the wrong trade.
2. **Docs.** `content/docs/data-modeling/drivers.mdx`'s alias table described the CLI only and is now true of *both* hosts; it still needs `turso`'s contract and the `mongo`→`mongodb` canon.
3. **`pool` on a turso datasource is dropped in silence** — pre-existing, unchanged by this PR. `TursoDriverConfig` has no `min`/`max`, only `concurrency`, and in local mode the driver *is* a better-sqlite3 `SqlDriver`, which the engine's `POOL_UNSUPPORTED_DRIVER_IDS` rejects a pool block for. Making it a rejection is a new refusal on an authoring surface and needs its own ruling; recorded in the source comment.

---

https://claude.ai/code/session_017uFVNMmTxLpmfQYiuKM1Yx

---
_Generated by [Claude Code](https://claude.ai/code/session_017uFVNMmTxLpmfQYiuKM1Yx)_


os-project-manager:
**Pushed `cc2447b4` — all four reds addressed in one pass.** Branch contains current `main` (`86e6f6c8d`).

Three of the four were in the one file this PR adds, `packages/spec/src/data/driver/turso.zod.ts`. Three independent gates on one new file was the right hint: I stopped fixing symptoms and swept the file against `mongo.zod.ts` / `sqlite.zod.ts` / `common.zod.ts`, which is how the third was found before CI could report it.

| # | job | cause | fix |
| :-- | :-- | :-- | :-- |
| 1 | `Check Documentation Links` | `https://…` (U+2026) in the `url` `.describe()`, rendered into `driver-turso.mdx` | Fixed **at the source**, never in the generated page. The siblings' `url` keys describe themselves in **words** (`'Connection URI (supersedes the discrete fields)'`) and paste no URL prefixes at all — so turso now reads `'libSQL endpoint or local file: a remote libsql/https Turso URL, a file path, or :memory:'`, and the same treatment went to `syncUrl`. Concrete example URLs stay in the TSDoc, which the reference tables do not inline. |
| 2 | `ESLint` (really `check:spec-parsed-alias`) | ADR-0122: `TursoTransportMode` was `z.infer` | Now `z.input`. **No `TursoTransportModeParsed` added** — a plain `z.enum`, no `.default()`, no `.transform()`, so the two are the same three literals and a second alias would distinguish nothing. The gate's own second route (pin it as isomorphic) is what the two sibling enums do: `SqliteWasmPersistMode` and `DriverSslToggle` are both bare `z.input` with no parsed twin. Pinned as `Iso834` in `type-alias-convention.pin.test.ts`, with all three stated counts moved 823 → 824. |
| 3 | `Test Core (1/3)` → `@objectstack/spec#test` | — | **Resolved.** Full package suite (not my earlier filter): **349 files / 8970 tests, exit 0.** Your read was right that my `445 ✅` was a filtered subset; the full suite is what I ran this time and what I am reporting. |
| 4 | `TypeScript Type Check` → `check:api-surface` | snapshot missing #6885's 6 storage-cursor exports — `main` moved past my merge | Re-merged, `rm -rf packages/spec/dist`, clean rebuild, regenerated. |

**On #4, the verification you asked for rather than the generator's exit code.** I diffed my regenerated snapshot against `origin/main`'s directly:

```
StorageListPage/StorageListOptions/decodeStorageListCursor/
encodeStorageListCursor/resolveStorageListLimit/DEFAULT_STORAGE_LIST_LIMIT   main=1  mine=1
ADDED (mine only, 11): DATABASE_DRIVER_SELECTION_ALIASES, DriverVocabularyEntry,
  TursoConfig, TursoConfigParsed, TursoConfigSchema, TursoDriverSpec,
  TursoTransportMode, TursoTransportModeSchema, driverHasLocalDefault,
  getTursoConfigJsonSchema, resolveDatabaseDriverId
REMOVED: []
```

Zero removals is the assertion that matters — that is the one the stale-`dist` failure violated both times.

### The third break, found by the sweep rather than by CI

`alias-integrity.test.ts` (#5481): the unknown-key probe **normalizes case and underscores**, so my alias table had two pairs collapsing onto one probe — `auth_token`/`authtoken` and `sync_interval`/`syncinterval` — where only one entry can ever fire. Worse than a duplicate: `auth_token`, `encryption_key` and `sync_url` are mere casing variants of **declared** keys, which the near-miss suggester already reaches, so they were alias rows that could never fire at all. The table is now semantic near-misses only (`uri`, `connectionstring`, `dsn`, `database`, `databaseurl` → `url`; `token`, `jwt` → `authToken`; `syncinterval` → `sync`) — the same shape `mongo.zod.ts` uses.

### Re-measured on the re-merged tree

`pnpm lint` ✅ · tsc spec/runtime/cli/service-datasource ✅ · **spec 8970 ✅** · cli utils 306 ✅ · runtime standalone 87 ✅ · service-datasource 264 ✅ · `check:api-surface` · `check:generated` · `check:authorable-surface` · `check:liveness` · `check:strictness-ledger` · `check:spec-changes` · `check:upgrade-guide` · `check:spec-parsed-alias` · `check:adr-0087-registration` ✅ — plus main's brand-new `check-tenant-chokepoint.mjs`, which I ran because the merge introduced it ✅.

Still in draft; ready-flip and queue are yours.

---

os-zhuang:
Heads-up from the drivers lane — this PR's tree currently reds the **#6860 pin**, and the failure surfaces on *neighbouring* PRs rather than on this one, so it is easy to miss.

Queue build [31297272050](https://github.com/objectstack-ai/objectstack/actions/runs/31297272050) (PR #6932, an unrelated `driver-sql` change speculatively merged on top of this one) failed in `packages/cli/src/commands/database-driver-allowlist.pin.test.ts`:

```
expected Set{ 'sqlite', 'sqlite-wasm', …(5) } to deeply equal Set{ 'better-sqlite3', …(13) }
-   "better-sqlite3", "default", "factory", "function", "on-disconnect", "safe", "string",
```

**Why this PR is the likely owner.** That pin derives its expected set by scraping every lowercase quoted literal out of `packages/cli/src/utils/storage-driver.ts` — which this PR rewrites, +121/−30 — and then filtering the candidates through `resolveStorageDefinition` **used as an oracle**: a token is a driver kind iff the resolver returns a truthy `driverId`, or throws `UnsupportedDriverError` (whose `driverType` is then taken as the answer). On `main`, an unrecognized kind at `isDev: false` returns `null`, so `'function'`, `'string'`, `'safe'`, `'default'`, `'factory'`, `'on-disconnect'` and `'better-sqlite3'` all wash out. In this PR's tree they no longer do — so the rewritten resolver is answering for tokens it used to reject.

That is only a *test* failure if the new behaviour is intended; if unrecognized kinds now resolve or throw in production where they previously returned `null`, it is worth a second look on its own merits.

**Probable cause:** the pin landed in **#6913** (merged), and this PR does not touch it — its file list covers `storage-driver.ts` and `storage-driver.test.ts` but not `database-driver-allowlist.pin.test.ts`. If this branch forked before #6913 landed, a rebase would surface the red here instead of on whoever queues behind you.

The pin's own docstring anticipates you explicitly:

> SCOPE (#6860 vs #6345): this pins the CANONICAL kinds … converging that vocabulary is #6345's job, and this pin is written so it does not prejudge it.

So updating the pin alongside the vocabulary looks in-scope for this card rather than a drive-by. No action requested of you here beyond awareness — I have not requeued anything of yours, and #6932 has since re-entered the queue without this PR ahead of it.

---

os-project-manager:
**Queue failure fixed — pushed `dd63b186a`.** Real regression, not flaky, and the mechanism is narrower than the queue signature suggests. **The allowlist needed no change; the defect was mine.**

## Why two green PRs went red together

The `#6860` pin derives both sides. Its driver-kind side uses `resolveStorageDefinition` as an **oracle**, over a deliberately over-broad candidate net (every lowercase single-quoted literal in `storage-driver.ts`), and reads the answer back out of the thrown error:

```ts
} catch (err) {
  if (err instanceof UnsupportedDriverError) return err.driverType;   // ← before
```

That catch was written for one case: **turso with no URL** — recognized, unusable, and `driverType` is a canonical id.

#6345 taught the CLI to refuse an **explicitly-named unknown driver** (the CLI half of "both hosts refuse the same input", replacing a silent fall-through to the dev SQLite default). I raised the **same** `UnsupportedDriverError` with the operator's **raw token** in `driverType`. So every candidate suddenly "resolved", and the derived set became:

```
Expected (allowlist, 7):  sqlite, sqlite-wasm, turso, postgres, mysql, mongodb, memory
Received (derived, 13):   … + better-sqlite3, default, factory, function,
                              on-disconnect, safe, string
```

Not a vocabulary drift — a **poisoned oracle**. No allowlist could have equalled that set.

## Correcting the diagnosis

> (a) 把 `start.ts` / `dev.ts` 的 `options:` 数组更新到新的 canonical 集合

Measured: that is a **no-op**. #6860 already landed the canonical seven, `mongodb` (not `mongo`) included:

```ts
options: ['sqlite', 'sqlite-wasm', 'turso', 'postgres', 'mysql', 'mongodb', 'memory'],
```

`start.ts` and `dev.ts` are **untouched** in this PR — the `git diff --stat` for both is empty. The rename had already been absorbed on that side; only my oracle was wrong.

## The fix

`UnsupportedDriverError` now carries `recognized`, defaulting to `true` so the pre-#6345 turso call sites keep their meaning unchanged, and the unknown-spelling throw passes `{ recognized: false }`. The pin returns `null` for that case.

**The assertion is untouched** — both sides still derived, still required to be equal. I repaired the derivation the pin's own docstring describes ("a recognized kind that cannot become a definition still counts as recognized"); an unrecognized token was never in that category. The discriminator lives on the error, not in the test, so the pin keeps asking the **resolver** what a token means rather than growing its own opinion. Both files carry a comment naming this incident.

I did **not** take route (b) (deriving the allowlist from the shared table). The pin now enforces exactly the property that derivation would enforce structurally — it just proved it by catching a real drift within hours — and expanding a 40-file PR during a queue incident is the wrong trade. It stays follow-up #1.

## Proof it still bites (not silenced)

| revert | result |
| :-- | :-- |
| drop `mongodb` from `start.ts`'s allowlist | **3 red** — including `start and dev offer the same kinds as each other` |
| revert my `recognized: false` | **4 red** — the exact queue signature |
| both restored | **10/10 green** |

## Re-verified on the re-merged tree (`68feaadd6` + later)

`pnpm lint` ✅ · tsc spec/cli/runtime/service-datasource ✅ · **`@objectstack/cli` FULL package 101 files / 1095 tests** ✅ (not a filter — the queue runs the full suite) · **spec 350 files / 9094 tests** ✅ · runtime standalone 87 ✅ · service-datasource 264 ✅ · `check:api-surface` · `check:generated` · `check:authorable-surface` · `check:liveness` · `check:strictness-ledger` · `check:spec-changes` · `check:upgrade-guide` · `check:spec-parsed-alias` · `check:adr-0087-registration` ✅

Artifacts: the `os-regen` guard caught `spec-changes.json` / `protocol-upgrade-guide.md` stale from the merge and I regenerated before committing. Then a clean `rm -rf packages/spec/dist` rebuild + full regeneration produced **zero diff**, and the export-removal check across every `api-surface/*.json` versus `origin/main` reports **REMOVED: NONE** — the assertion that the two earlier `check:api-surface` reds violated.

PR stays **ready**; auto-merge still on.

---

Issue #6285, opened 2026-08-07
[转移] 发布期护栏:多组织下「平台级 + schedule + create_record + 未显式 organization_id」拒绝——#6155 Q3=A 裁决的 lint 半边


hotlong:
Part of #6155(维护者 2026-08-07 裁决 Q2=**A** + Q3=**A**;services 座位转移立单,落点在运行时发布门——域标签交分诊按落点文件定)。

## 裁决口径

- **Q2=A**:平台级定时流程的 org 由作者在 `create_record` 的 `fields.organization_id` **显式声明**(#6153 的 fill-only 已保证作者值优先——该半边**零新代码**,既有能力);
- **Q3=A**:护栏**只落运行时发布门**——给 `assertRuntimeAuthoringRules` → `evaluateRuntimeAuthoringGate` 补两个缺失输入(**写入 org** 与 **部署形态** `postureEnforcesWall(resolveTenancyPosture())`),CLI 纯函数侧 ⛔ 不判(构建机 env 是假信号);
- 拒绝组合:多组织部署 && 平台级流程 && schedule 触发 && 含 `create_record` && 未在 `fields` 显式给 `organization_id` ⇒ **发布时拒绝**(与 #2308/#5378 lint 面同族,照抄形状不发明新面)。

## #6155 dev 探针已量出的两个输入缺口(实现起点,勿重查)

1. 部署形态:`postureEnforcesWall(resolveTenancyPosture())` 是进程级 env 读数,`AUTHORING_RULES` 纯函数签名不含它——按 Q3=A 从 gate 侧注入;
2. 写入 org:在 `saveMetaItem` 手里但 `assertRuntimeAuthoringRules` 未传给 `evaluateRuntimeAuthoringGate`;
3. ⚠️ **待确认的短路**:`assertRuntimeAuthoringRules` 现有 `if (this.environmentId === undefined) return;`——控制面 kernel 上整条 gate 不跑。实现前先答「这是既有设计还是遗漏」:若短路使护栏在关键部署形态不可达,STOP 上报,⛔ 不得擅自拓宽 gate 激活面(#6155 dev 原话:无把握,未擅自改动)。

## 依赖

Blocked-by: #6283(spec 半边 flip——Q1=B 落地后「org 归属的 flow」不复存在,本护栏的拒绝组合语义才是终态)。

Refs:#6155(裁决与探针证据)、#6153(fill-only 先例)、#2308/#5378(lint 面同族)、ADR-0118 D1。

[...]

Issue #6190, opened 2026-08-07
org 作用域的 flow overlay 只在「本进程内发布后」绑定触发器,重启后静默失绑——冷启动两条读路径都把 organization_id 非空的行滤掉了


hotlong:
发现于 #6155 的前提核验(未在该 PR 内修,按 Prime Directive #10 单开)。

## 事实(实测,非推断)

`flow` 在 `DEFAULT_METADATA_TYPE_REGISTRY` 里是 `allowOrgOverride: true` + `allowRuntimeCreate: true`,所以 `runtime/src/domains/meta.ts:158-159` 的 `saveMetaItem({ type, name, item, organizationId })`(org 取 `resolveActiveOrganizationId`)会写出 `sys_metadata.organization_id = 'org_a'` 的 flow 行。实测确认该行确实落库。

之后这一行的命运是**不对称**的:

| 时机 | 路径 | org 行可见? |
|:---|:---|:---|
| 发布当场(同进程) | `saveMetaItem` → `applyRegistryWriteThrough`(publish 分支)→ `hydrateOverlayIntoRegistry` → 进程级 SchemaRegistry | ✅ 可见 |
| 发布当场,自动化重绑 | `emitMetadataMutation` → `metadata:reloaded` → `resyncFlowsFromProtocol` → `getMetaItems({ type: 'flow' })`(读到 registry 分支) | ✅ 绑定,定时任务起飞 |
| **重启后冷启动** | `ObjectQLPlugin.restoreMetadataFromDb` → `protocol.loadMetaFromDb()`,`where = { state:'active', organization_id: null }`(`metadata-protocol/src/protocol.ts:10620-10623`) | ❌ 跳过 |
| **重启后 kernel:ready 绑定** | `service-automation/src/plugin.ts:1542` `protocol.getMetaItems({ type: 'flow' })` —— **不传 `organizationId`**,于是 `protocol.ts:3319` 的 `orgRecords = orgId ? … : []` 为空 | ❌ 跳过 |

实测(真 `ObjectStackProtocolImplementation` + 内存 sys_metadata 引擎,两行 flow:`org_sweep` org=org_a、`platform_sweep` org=null):

```
PROBE rows = [{"name":"org_sweep","org":"org_a"},{"name":"platform_sweep","org":null}]
PROBE getMetaItems({type:flow}) names(registry 为空时) = ["platform_sweep"]
[...]

Issue #5086, opened 2026-08-04
metadata: allowRuntimeCreate:false is not enforced — PUT /meta creates job and agent items the registry declares code-only


baozhoutao:
Found working the #3909 G8 checklist on `main` @ `a1a855a28` (the rc.3 train tip).

`#4509` closed the door on runtime-created jobs by setting registry flags:

```ts
// packages/spec/src/kernel/metadata-plugin.zod.ts:658
{ type: 'job', label: 'Background Job', …, supportsOverlay: false,
  allowOrgOverride: false, allowRuntimeCreate: false, … }
```

and its changeset states the intent unambiguously:

> - `allowRuntimeCreate: false` — **no "create job" in Studio or via `PUT /meta`**.
> - `allowOrgOverride: false` — no per-org job fork…

**`PUT /meta` still creates one.**

```
PUT /api/v1/meta/job/rc3_runtime_job
    {"name":"rc3_runtime_job","label":"J",
     "schedule":{"type":"cron","expression":"0 0 * * *"},"handler":"nope"}
→ 200 {"success":true,"message":"Saved customization overlay (env-wide) — type=job, name=rc3_runtime_job"}

GET /api/v1/meta/job
→ listed, with handler "nope"
```

`handler: "nope"` names no function in any compiled bundle — precisely the record #4509 argues can never be scheduled, saved and reported as success.

## Not job-specific — the gate looks unenforced generally

`agent` is the only other type carrying `allowRuntimeCreate: false`. It behaves the same:

```
PUT /api/v1/meta/agent/rc3_agent_probe {"name":…,"label":"A","role":"assistant","instructions":"be helpful"}
→ 200 {"success":true,"message":"Saved customization overlay (env-wide) — type=agent, …}
```

So both flagged types accept a runtime create. A minimal-payload PUT to `object` / `flow` / `job` all return 422, which is schema validation firing first — that is what makes this easy to miss: the gate never runs, and only a payload that *passes* the schema reveals it.

`supportsOverlay: false` appears equally unenforced — the success message calls the row a "customization overlay" for a type declared not to support them.

## Why it matters

[...]

Issue #3050, opened 2026-07-16
OWD posture is lint-only at runtime: env can widen a packaged object's sharingModel via OS_METADATA_WRITABLE, and external≤internal (ADR-0090 D11) is never enforced on the write path


os-zhuang:
## Summary

ADR-0086 D1 classifies OWD (`object.sharingModel`) as *package metadata whose environment may only **tighten***, and ADR-0090 D11 requires `externalSharingModel ≤ sharingModel`. Both rules exist today **only as CLI lint** (`validateSecurityPosture`, run by `os compile`/`os lint`) — **nothing enforces them on the runtime metadata write path**. Per ADR-0049 (no unenforced security properties) this is a gap to close, not a doc to keep.

## Verified write paths (HEAD 2026-07-16)

| Path | Verdict | Evidence |
|---|---|---|
| Env overlay of a **packaged** object (default deploy) | ✅ GUARDED — `403 not_overridable` | `object` type ships `allowOrgOverride:false` (`spec/kernel/metadata-plugin.zod.ts:599`); rejection in `saveMetaItem` (`metadata-protocol/src/protocol.ts:3825`) |
| Same, with **`OS_METADATA_WRITABLE=object`** | ❌ **WIDE OPEN** | `envWritableTypes()` flips `isOverlayAllowed('object')` (`protocol.ts:3319-3386`); after that there is **no field-level guard on `sharingModel` anywhere** in `saveMetaItem`/`publishMetaItem` — an env overlay can take a packaged object from `private` to `public_read_write` with zero validation |
| `external ≤ internal` (D11) at runtime | ❌ **NEVER ENFORCED** | `object.zod.ts:910` states the rule in `.describe()` prose only — no zod refine; the only enforcement is lint rule `SECURITY_EXTERNAL_WIDER` (`packages/lint/src/validate-security-posture.ts:216-245`), which does not run on `saveMetaItem`/`publishMetaItem` |
| Runtime-created env-owned object edits | ⚠️ ungated by design (env owns it) — but D11 still unenforced | `allowRuntimeCreate:true`; no posture check on save |
| ADR-0090 D7 security-domain publish gate | CLI-only | `validateSecurityPosture` invoked from `cli/src/commands/compile.ts:354` etc.; absent from `metadata-protocol/src/protocol.ts` |

[...]

PR #6971, opened 2026-08-09
fix(metadata-protocol,objectql): activate the #4463 runtime authoring gate from a declared authoring channel instead of the environmentId proxy (#6710)

### LINKED ISSUE #6710: #4463 运行时发布门在 host-config 拓扑(environmentId 未绑定)上整体不跑——#5086 只把 code-only 拒绝移出了同一条短路
## 观察到的缺口

`ObjectStackProtocolImplementation.assertRuntimeAuthoringRules`(`packages/metadata-protocol/src/protocol.ts`,#4463 的运行时发布门)第一行是:

```ts
if (this.environmentId === undefined) return;
```

这条短路本身是**既有设计且正确**——它是 ADR-0005「控制面 / 包作者自有通道」的 carve-out,与其下方 #3050 authoring gate 的 `if (this.environmentId !== undefined)` 同一口径,`protocol.runtime-authoring-gate.test.ts` 也把它作为刻意行为钉住。本单**不是**要求撤掉它。

问题在于 `environmentId === undefined` 并**不只**意味着控制面内核。#5086 在同一文件里(`saveMetaItem` 的 code-only 拒绝块注释)留下的原话即是证据:

> 「…the premise the carve-out rests on ("this kernel is the package author's own bootstrap channel") is simply not true for the CLI's lightweight assembler: a host config with instantiated plugins (`isHostConfig` → `shouldBootWithLibrary === false`) boots `new ObjectQLPlugin()` with NO environmentId, so the flagship showcase — a self-hosted app server whose `PUT /api/v1/meta/*` is an END-USER surface — ran with this entire gate disengaged.」

#5086 据此把 **code-only 类型的拒绝**移出了 `environmentId` 门(「Keying authorization off a row-scoping key is what made a type-level declaration depend on deployment topology; the declaration decides it here instead」),但**只移了那一处**。#4463 的运行时发布门(共享 `AUTHORING_RULES` 的 26 条规则)仍留在短路后面。

于是在 host-config 拓扑上,`PUT /api/v1/meta/*` 这个终端用户面上**一条作者时规则都不跑**——而 #4463 立单的理由恰恰是「对 Studio 租户 / MCP-AI 作者来说这不是四道门里较弱的一道,而是**唯一**的一道」。

## 现状核验(本单只测了这些,未测的如实标出)

已实测:每条**常规**服务路径都绑定 environmentId,所以它们照常进门——

| 路径 | environmentId | 位置 |
|---|---|---|
| `os dev` | `env_local`(默认) | `packages/cli/src/commands/dev.ts:225` |
| `os start` | `env_local`(默认) | `packages/cli/src/commands/start.ts:197` |
| standalone artifact stack | `proj_local`(默认) | `packages/runtime/src/standalone-stack.ts:378` |
| cloud 单项目内核 | 各自项目 id | `packages/objectql/src/plugin.ts:295` |

**未实测**:host-config 拓扑今天是否仍会走到无 environmentId 的装配。仓内有两处形状可疑——`packages/cli/src/commands/serve.ts:1000` 的 `config.objects && !hasObjectQL` 自动注册分支(`new ObjectQLPlugin()`,无 environmentId),以及任何 `objectstack.config.ts` 里手写 `plugins: [new ObjectQLPlugin()]` 的 host。上面引的注释是 #5086 当时的测量,不是我这轮的复测。**定级前应先复现这一条**:若该拓扑已随其它改动消失,本单可直接关掉;若仍在,则是一个终端用户面上的门整体失效。

## 为什么单独立单

发现于 #6285(PR #6708)。该单的 dispatch 明确要求「若短路使护栏在关键部署形态不可达,STOP 上报,⛔ 不得擅自拓宽 gate 激活面」。核验结论是**本护栏可达**(多组织部署都在上表的路径上),所以 #6285 按裁决正常落地、未动激活面。但拓宽与否会同时改变全部 26 条共享规则的影响面,是独立的契约决定,不该搭在一个护栏单里顺手做。

## 可能的方向(不预判,留给分诊/维护者)

1. 照 #5086 的思路,把**门的激活**从 `environmentId`(一个行作用域键)改为按拓扑/来源判定,让「包作者自有通道」这个真实意图有一个自己的表达,而不是借用 environmentId;
2. 维持现状,但把「host-config 拓扑无 environmentId」本身当缺陷修掉(让该装配也绑一个 environmentId),这样短路的前提重新成立;
3. 若复现不出该拓扑,关单。

Refs:#4463(门的立单与四项裁决)、#5086(同一前提的上一次修补,只移了 code-only 一处)、#6285 / PR #6708(本次核验的来处)、#6602(同族:另一处只按 `environmentId` 而不按 org 把关的 seam)。

---

os-zhuang:
Fixes #6710

Implements the maintainer's **A1** ruling of 2026-08-09 05:41Z (comment 5229995235): the gate is active by default on every kernel, and the ADR-0005 "package author's own bootstrap channel" carve-out becomes an explicit declaration plumbed through `assembleMetadataProtocol`.

## The defect

`ObjectStackProtocolImplementation.assertRuntimeAuthoringRules` — the #4463 runtime publish gate that runs the 26 shared `AUTHORING_RULES` (the same table `os validate` / `os build` / `os lint` run) — opened with:

```ts
if (this.environmentId === undefined) return;
```

That short-circuit was meant to express ADR-0005's carve-out, and **the carve-out itself is legitimate and stays**: a control-plane kernel installing a package is not an author publishing into a live tenant. The key was the problem. `environmentId` is a **row-scoping** key, and two topologies with opposite intents both leave it undefined:

| | genuine control plane | CLI host-config topology |
|---|---|---|
| `environmentId` | undefined | undefined |
| assembly | `assembleMetadataProtocol` | `assembleMetadataProtocol` |
| serves `PUT /api/v1/meta/*` | yes (package author's own channel) | yes (**end-user surface**) |

The second is `serve.ts`'s `config.objects && !hasObjectQL` auto-register branch (`new ObjectQLPlugin()` with no options), which is the shape any `objectstack.config.ts` with instantiated plugins gets (`isHostConfig` yields `shouldBootWithLibrary === false`) — including the flagship showcase. So a self-hosted app server ran **zero** of the 26 rules on every publish. For a Studio tenant or an MCP/AI author this gate is not the weakest of four doors, it is the only one: a `sys_metadata` overlay row is never in the CLI's config file, so there is no `os lint` for it.

#5086 had already retired this same proxy for the code-only refusal a few hundred lines below, for the same stated reason — it just moved that one site.

## Premise, re-measured on this base

Re-verified at boot level on `origin/main` @ `68feaadd6` (not a source read): a kernel built exactly as `serve.ts:1000` builds it, then #4463's own measured body (the broken-CEL approval flow `record.owner ==`) pushed through `saveMetaItem`:

```
PROBE environmentId     : undefined
PROBE PUT /meta verdict : {"outcome":"THROWN","message":"[ObjectQL] No driver available for object 'sys_metadata'"}
PROBE #4463 GATE FIRED? : false
```

The throw comes from the **engine's persistence layer**, not the gate — execution had already run past `assertRuntimeAuthoringRules` (which throws 422 `INVALID_METADATA` before any engine call). The producer is unchanged on this base: `packages/cli/src/commands/serve.ts:996-1001` still constructs `new ObjectQLPlugin()` with no options. **Premise holds.**

## The shape implemented, and why it is fail-safe

A new public plugin option states what a kernel **is**; gate activation reads that instead of row scope.

**New public contract shape (explicitly declared, as the ruling requires):**

```ts
export type MetadataAuthoringChannel = 'environment' | 'package-author';

new ObjectQLPlugin({ authoringChannel: 'package-author' })
createMetadataProtocolPlugin({ authoringChannel: 'package-author' })
assembleMetadataProtocol(ctx, ql, environmentId, { authoringChannel })
new ObjectStackProtocolImplementation(engine, getServices, environmentId, authoringChannel)
```

Exported from `@objectstack/metadata-protocol`: the type `MetadataAuthoringChannel`, the new `AssembleMetadataProtocolOptions`, and the widened `MetadataProtocolPluginOptions` / `ObjectQLPluginOptions`. All four surfaces are additive and optional — no existing call site changes shape.

Three properties are deliberate:

1. **Omission means MORE enforcement.** The default is `'environment'`, resolved at the assembly seam (`options.authoringChannel ?? 'environment'`) as well as at the constructor, so every path that predates this option — including `ObjectQLPlugin`'s legacy positional `(ObjectQL, hostContext)` constructor, which returns before any option is read — lands on the gated channel. The failure mode being designed out is precisely "a future assembly variant nobody thought about", which is how the host-config topology got here.
2. **A channel name, not a boolean.** `skipAuthoringRules: true` would be the same bytes with the opposite meaning — a switch any assembly could reach for to make a red publish go away. A caller has to claim to *be* the package author to be treated as one. There is deliberately **no env-var fallback** (unlike `skipSchemaSync`): a deployment must not be able to turn an end-user guardrail off from outside the code. The per-write hatch that does exist is `OS_ALLOW_UNLINTED_METADATA_WRITES` (#4463 D4), which degrades a refusal to a loud log rather than silencing it.
3. **Threaded through the one shared seam.** `assembleMetadataProtocol` is what both the built-in mount and the delegated ADR-0076 Step 2 mount call, so enforcement cannot silently differ by mount style. Pinned in both directions.

`environmentId` keeps every other job: the `environment_id` stamp and filter, the ADR-0005 overlay whitelist, the #3050 authoring gate's own scope check, and the local metadata-storage provisioning decision. **Only this one activation moved** — pinned by a dedicated case asserting the #3050 gate still turns on and off by `environmentId` while the #4463 gate follows the declaration.

## File surface (no deviation from the dispatch)

| file | change |
|---|---|
| `packages/metadata-protocol/src/protocol.ts` | `MetadataAuthoringChannel` type; `authoringChannel` field + 4th constructor param defaulting to `'environment'`; the activation line itself |
| `packages/metadata-protocol/src/plugin.ts` | `MetadataProtocolPluginOptions.authoringChannel`; new `AssembleMetadataProtocolOptions`; threads it into the protocol construction |
| `packages/metadata-protocol/src/index.ts` | exports the new type + options interface |
| `packages/objectql/src/plugin.ts` | `ObjectQLPluginOptions.authoringChannel`; field; forwards it to the assembly |
| `packages/metadata-protocol/src/protocol.runtime-authoring-gate.test.ts` | the 26 rules' posture matrix (6 rows) + 3 supporting cases |
| `packages/metadata-protocol/src/protocol.platform-schedule-org-gate.test.ts` | **one fixture re-spelled** — see below |
| `packages/objectql/src/plugin.authoring-channel.test.ts` | new: the assembly-wiring pin (6 cases) |
| `.changeset/authoring-channel-declaration.md` | minor for both packages |

The `saveMetaItem` two-tier write-gate region (#6190, in flight) was not touched. `packages/spec` was not needed. No `content/docs/releases/` edit.

**The one fixture the sweep found outside the edited test file:** `protocol.platform-schedule-org-gate.test.ts` (#6285) carried its own copy of the carve-out pin, spelled with the retired key. Disposition **re-spell**, not delete: the case's subject (the carve-out) is intact, only its expression changed, so it now constructs with `'package-author'`. It also gained a companion case, because #6285's guardrail is one of the 26 and its reach widened too — without it the file would assert only the side that stayed the same.

## Reverse verification

Predictions were written down **before** running (`/tmp/.../issue-6710/reverse-predictions.md`), then the activation line alone was reverted to `if (this.environmentId === undefined) return;` with the plumbing and tests left in place.

Predicted 7 red, measured 8. One miss, recorded rather than tidied away.

`protocol.runtime-authoring-gate.test.ts` — predicted 4 red, measured 5 (11 pre-existing #4463 cases stayed green throughout):

| case | predicted | measured |
|---|---|---|
| envId undefined, channel omitted ⇒ GATED | RED | **RED** |
| envId undefined, channel `environment` ⇒ GATED | RED | **RED** |
| envId undefined, channel `package-author` ⇒ bypassed | green (**cannot** go red) | green |
| envId `env_test`, channel omitted ⇒ GATED | green | green |
| envId `env_test`, channel `environment` ⇒ GATED | green | green |
| envId `env_test`, channel `package-author` ⇒ bypassed | RED (**inverted**) | **RED** |
| gated by the rules, not blanket | green | green |
| D4 hatch covers the newly-gated topology | RED | **RED** |
| does not disturb the #3050 gate | green | **RED — MISSED** |

Two of these are worth stating plainly rather than as a pass count:

- **One row cannot go red at all.** `envId undefined + channel package-author ⇒ bypassed` is green before and after, because the old code bypasses there too — for the wrong reason. It is honest coverage of the surviving carve-out, but it is *not* evidence for this change, and the matrix says so in its own comment.
- **One row is inverted by design.** `envId env_test + channel package-author` goes red on the revert because the OLD code *gates* where the new code bypasses. That is the row proving the declaration is genuinely read rather than ignored.
- **The missed prediction:** `does not disturb the OTHER gates` went red. Cause, on inspection: its tenant leg saves a *broken* flow, and under the old keying the #4463 gate refuses that write with a 422 **before** the #3050 gate is ever reached, so the observed gate list stayed empty. The red is real and informative (it pins the ordering of the two gates), but I predicted green because I only reasoned about the #3050 gate's own condition and not about what precedes it.

`plugin.authoring-channel.test.ts` — predicted 3 red, measured 3 (exactly cases 1, 3, 5).

The most load-bearing detail: the boot-level case failed on `expected undefined to be 'INVALID_METADATA'`, **not** on "did it throw". The unfixed build throws too — the engine's bare `No driver available`, an `Error` whose `code` and `status` are both undefined. A `rejects.toThrow()` assertion would therefore have stayed **green on the exact topology this issue is about**. Every rejection-class case here asserts the ADR-0112 `code` + `status` pair for that reason.

Side observation from the revert: `pnpm --filter @objectstack/metadata-protocol build` fails its DTS step under the reverted line, because `authoringChannel` becomes a field nothing reads. The ESM/CJS bundles are emitted before DTS fails, which is why the behavioural measurement above is still valid.

## Tests

All run through `flock /tmp/os-heavy-verify.lock` with `NODE_OPTIONS=--max-old-space-size=4096`.

```
@objectstack/metadata-protocol  Test Files  65 passed (65)    Tests  813 passed (813)
@objectstack/objectql           Test Files 160 passed (160)   Tests 2758 passed (2758)
@objectstack/rest               Test Files  72 passed (72)    Tests 1131 passed (1131)
@objectstack/runtime            Test Files 115 passed (115)   Tests 1743 passed (1743)
@objectstack/cli                Test Files 100 passed (100)   Tests 1044 passed (1044)
typecheck (metadata-protocol, objectql)   Done, 0 errors
pnpm lint (whole repo)                    exit 0
```

`rest` / `runtime` / `cli` are in the sweep because this change **widens** enforcement: they boot kernels that were previously on the bypassed side. All green with no fixture edits needed.

Family gates run locally (they live inside the ESLint CI job, not in `pnpm test`): `check:engine-double-contract`, `check:driver-memory-census`, `check:error-code-casing`, `check:route-envelope`, `check:meta-type-normalized`, `check:init-service-contract`, `check:published-files`, `check:doc-authoring`, `check:docs-audit-scope`, `check:nul-bytes` — all PASS.

## Follow-up in `cloud` (NOT in this PR)

One line, in `cloud`'s `packages/.../control-plane-preset.ts` around `:99` — the only genuine consumer of this carve-out in either repo:

```ts
new ObjectQLPlugin({ /* …existing… */ authoringChannel: 'package-author' })
```

The cross-repo window is accepted by the ruling and is safe by construction: until that lands, the control plane runs **gated** — the safe direction — softened by the existing `OS_ALLOW_UNLINTED_METADATA_WRITES` hatch. Same coordination shape as ADR-0076 Step 2 (#2462).

---

### DIFF
--- a/.changeset/authoring-channel-declaration.md
+++ b/.changeset/authoring-channel-declaration.md
@@ -1,0 +1,59 @@
+---
+"@objectstack/metadata-protocol": minor
+"@objectstack/objectql": minor
+---
+
+fix(metadata-protocol,objectql): the #4463 runtime authoring gate now runs on every kernel that has not declared itself the package author's channel (#6710)
+
+The 26 shared author-time rules (`AUTHORING_RULES` — the same table `os validate`
+/ `os build` / `os lint` run) were gated behind
+`if (this.environmentId === undefined) return;`. That short-circuit was meant to
+be ADR-0005's "the package author's own bootstrap channel" carve-out, and the
+carve-out itself is legitimate. The key was not: `environmentId` is a ROW-SCOPING
+key, and two very different topologies leave it undefined.
+
+**The defect.** The CLI's lightweight host-config assembler — `serve.ts`'s
+`config.objects && !hasObjectQL` auto-register branch, which constructs
+`new ObjectQLPlugin()` with no options — also boots with no `environmentId`.
+That is the shape any `objectstack.config.ts` with instantiated plugins gets
+(`isHostConfig` → `shouldBootWithLibrary === false`), including the flagship
+showcase app. Its `PUT /api/v1/meta/*` is an **end-user** surface, so a
+self-hosted app server ran **zero** of the 26 rules on every publish. For a
+Studio tenant or an MCP/AI author this gate is not the weakest of four doors —
+it is the only one, because a `sys_metadata` overlay row is never in the CLI's
+config file and there is no `os lint` for it. Measured at boot level: the kernel
+reports `environmentId === undefined` and #4463's own broken-CEL approval flow
+(`record.owner ==`) runs straight past the gate into persistence.
+
+**The fix — the channel is declared, not inferred.** A new plugin option states
+what a kernel *is*, and gate activation reads that instead of row scope:
+
+```ts
+new ObjectQLPlugin({ authoringChannel: 'package-author' })
+createMetadataProtocolPlugin({ authoringChannel: 'package-author' })
+```
+
+`'environment'` (the default, and what you get by omitting the option) runs the
+rules. `'package-author'` is the ADR-0005 carve-out and belongs only on the
+genuine control-plane assembly — the kernel installing packages on the
+platform's own behalf. The option is threaded through `assembleMetadataProtocol`,
+the one seam both mounts share, so the built-in and delegated (ADR-0076 Step 2)
+mounts cannot disagree.
+
+**Omitting it means more enforcement, never less.** That direction is the point:
+the failure mode being designed out is a future assembly variant nobody thought
+about silently reopening this hole, which is exactly how the host-config
+topology got here. It is also why the option is a channel NAME and not a
+boolean — `skipAuthoringRules: true` would be the same bytes with the opposite
+meaning, a switch for making a red publish go away. #5086 had already retired
+the same proxy key for the code-only refusal, for the same reason.
+
+**What changes for you.** A kernel that serves metadata writes to end users
+should change nothing — it now enforces the rules it always should have. A
+kernel that genuinely is a control plane must add `authoringChannel:
+'package-author'`; until it does it runs gated in the safe direction, and the
+existing per-write `OS_ALLOW_UNLINTED_METADATA_WRITES=1` hatch (#4463 D4)
+degrades a refusal to a loud log. `environmentId` keeps every one of its other
+jobs unchanged — the `environment_id` stamp and filter, the ADR-0005 overlay
+whitelist, the #3050 authoring gate's scope, and local metadata-storage
+provisioning. Only this one activation moved.
--- a/packages/metadata-protocol/src/index.ts
+++ b/packages/metadata-protocol/src/index.ts
@@ -6,7 +6,11 @@ export { ObjectStackProtocolImplementation, ConcurrentUpdateError, normalizeView
 // instead of minting a second not-found shape. See `recordNotFoundError`.
 export { recordNotFoundError } from './protocol.js';
 export { createMetadataProtocolPlugin, assembleMetadataProtocol } from './plugin.js';
-export type { MetadataProtocolPluginOptions } from './plugin.js';
+export type { MetadataProtocolPluginOptions, AssembleMetadataProtocolOptions } from './plugin.js';
+// [#6710] The declared authoring channel — the explicit expression of ADR-0005's
+// "package author's own bootstrap channel", replacing the `environmentId ===
+// undefined` proxy the #4463 gate used to key its activation off.
+export type { MetadataAuthoringChannel } from './protocol.js';
 
 // [#5839] `sys_view_definition`'s active-row uniqueness, delivered as a runtime
 // partial-UNIQUE migration (the `ensureOverlayIndex` paradigm, for the one other
--- a/packages/metadata-protocol/src/plugin.ts
+++ b/packages/metadata-protocol/src/plugin.ts
@@ -36,6 +36,7 @@ import {
     resolveIndexExec,
 } from './migrations/view-definition-active-index.js';
 import { ObjectStackProtocolImplementation } from './protocol.js';
+import type { MetadataAuthoringChannel } from './protocol.js';
 
 export interface MetadataProtocolPluginOptions {
     /**
@@ -46,10 +47,30 @@ export interface MetadataProtocolPluginOptions {
      * Mirrors `ObjectQLPluginOptions.environmentId` — pass the same value.
      */
     environmentId?: string;
+    /**
+     * [#6710] Which authoring channel this kernel's metadata writes arrive on.
+     *
+     * Leave unset on ANY kernel that serves `PUT /api/v1/meta/*` to end users
+     * (Studio tenants, MCP/AI authors, self-hosted app servers): the default
+     * `'environment'` runs the #4463 runtime authoring rules, which for those
+     * authors is the only author-time gate that exists.
+     *
+     * Set `'package-author'` ONLY on the genuine control-plane assembly — the
+     * kernel that installs packages on the platform's own behalf and is not an
+     * author publishing into a live tenant. Stating it is a claim about what
+     * this kernel IS; it is not a switch for making a red publish go away.
+     *
+     * Deliberately no env-var fallback (unlike `skipSchemaSync`): a deployment
+     * must not be able to turn an end-user guardrail off from the outside. The
+     * per-write escape hatch that DOES exist is
+     * `OS_ALLOW_UNLINTED_METADATA_WRITES` (#4463 D4), which degrades the
+     * refusal to a loud log instead of silencing it.
+     */
+    authoringChannel?: MetadataAuthoringChannel;
 }
 
 export function createMetadataProtocolPlugin(options: MetadataProtocolPluginOptions = {}): Plugin {
-    const { environmentId } = options;
+    const { environmentId, authoringChannel } = options;
     return {
         name: 'com.objectstack.metadata.protocol',
         version: '1.0.0',
@@ -71,9 +92,18 @@ export function createMetadataProtocolPlugin(options: MetadataProtocolPluginOpti
                 );
             }
 
-            assembleMetadataProtocol(ctx, ql, environmentId);
+            assembleMetadataProtocol(ctx, ql, environmentId, { authoringChannel });
         },
     };
+}
+
+/** Extra assembly inputs that are not row scope. Bag-shaped so the next one is additive. */
+export interface AssembleMetadataProtocolOptions {
+    /**
+     * [#6710] See {@link MetadataProtocolPluginOptions.authoringChannel}.
+     * Omitted ⇒ `'environment'` ⇒ the #4463 runtime authoring gate is active.
+     */
+    authoringChannel?: MetadataAuthoringChannel;
 }
 
 /**
@@ -98,6 +128,7 @@ export function assembleMetadataProtocol(
     ctx: PluginContext,
     ql: any,
     environmentId?: string,
+    options: AssembleMetadataProtocolOptions = {},
 ): ObjectStackProtocolImplementation {
             // Metadata-storage platform objects (sys_metadata + history/audit
             // siblings + sys_view_definition). Same `environmentId === undefined`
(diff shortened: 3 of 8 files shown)


os-zhuang:

Verified against GitHub, not against the report:

- **File surface matches the declaration**: 8 files, all on the claimed surface. The one addition beyond the dispatch — `protocol.platform-schedule-org-gate.test.ts` — is test-only, declared in the PR body with its disposition reasoned (re-spell, not delete: #6285's carve-out pin kept its subject and changed only its expression, plus a companion case because that guardrail is one of the 26 and its reach widened too). The `saveMetaItem` two-tier write-gate region (#6190, in flight this round) is untouched; `packages/spec` not needed; no `content/docs/releases/` edit.
- **CI on `5901d60`**: 25 check runs, 0 pending, 0 failures — 23 `completed: success`, 2 `skipped` by the path filter (`Build Docs`, `Console Pin Gate`). Read per-job, not off the aggregate: `ESLint` success (this is the job carrying the family gates), `TypeScript Type Check` success, plus Build Core, Test Core 1-3/3, Dogfood Regression Gate 1-3/3, Dogfood Verify CLI, Temporal Conformance, Check Changeset, Check PR Size. Changeset present (minor × 2).
- **The ruling is implemented as ruled, in the fail-safe direction**: default `'environment'` resolved at *both* the assembly seam and the constructor, so `ObjectQLPlugin`'s legacy positional constructor — which returns before options are read — lands gated. A channel name rather than a boolean, and deliberately no env-var fallback: a caller must claim to *be* the package author to be treated as one. Omission means more enforcement, never less. `environmentId` keeps all its other jobs, pinned by a dedicated case asserting the #3050 gate still keys off it while #4463 follows the declaration.
- **Reverse verification is the real thing**: predictions written before running, 7 predicted red vs 8 measured, and the miss (`does not disturb the OTHER gates`) is reported with its cause — the tenant leg's broken flow is refused 422 by #4463 before #3050 is reached — rather than tidied away. Two rows are explained instead of counted: one *cannot* go red (green both ways; honest coverage of the surviving carve-out, not evidence for this change) and one is *inverted* by design (that is the row proving the declaration is genuinely read).
- **Load-bearing detail I want on the record**: the boot-level case asserts the ADR-0112 `code` + `status` pair, not `rejects.toThrow()`, because the unfixed build throws too — the engine's bare `No driver available`, with `code` and `status` both undefined. A `toThrow` assertion would have stayed green on the exact topology this issue is about.

Two calls the dev flagged, both endorsed:

1. **The declaration decides activation on its own; `environmentId` is not AND-ed back in as a co-condition.** Correct. A hybrid would be marginally stricter but would re-admit the retired proxy key into the very judgment the A1 ruling moved off it, and a control plane that later gained a row scope would silently flip posture. The `environmentId` set + `package-author` declared row is pinned explicitly with its reasoning, so the rule reads one way only.
2. **The `cloud` follow-up is recorded, not done**, and the "only genuine consumer" claim rests on round 4's inventory rather than a fresh enumeration — stated as such in the PR body. Accepted: the `cloud` repo is outside this session's scope and the ruling accepted the cross-repo window explicitly. Until `control-plane-preset.ts:99` declares the channel, the control plane runs gated in the safe direction with the `OS_ALLOW_UNLINTED_METADATA_WRITES` hatch, same shape as ADR-0076 Step 2 (#2462).

---

Issue #6602, opened 2026-08-08
An org-scoped overlay row reaches the process-wide SchemaRegistry on unscoped kernels — both the write-through and the read hydration gate only on `environmentId`, never on org


baozhoutao:
Found while verifying #6190 against `origin/main` (impact ② of that card). Filed separately per Prime Directive #10 — it is not `flow`-specific and it survives whichever way #6190's contract question is ruled, so it does not belong in #6190's diff.

## The fact (measured, not inferred)

`applyRegistryWriteThrough` (`packages/metadata-protocol/src/protocol.ts`) states its own invariant in its TSDoc:

> The non-object branch carries the same `environmentId === undefined` gate the read-side hydration carries: a project-scoped row must not be registered into a registry that unscoped (control-plane) callers share. **The write must not be more permissive about that than the read is.**

The gate is one line — `if (this.environmentId !== undefined) return;` — and it says nothing about `organization_id`. So on an **unscoped** kernel (`environmentId === undefined`), a per-org overlay write hydrates straight into the **process-wide** SchemaRegistry under the plain key.

Measured with the real `ObjectStackProtocolImplementation` over a stub engine, tracking `registry.registerItem` calls:

```
PROBE P3 rows = [{"type":"view","name":"org_grid","org":"org_a"}]
PROBE P3 registry write-through = [{"type":"view","name":"org_grid"}]
```

`view` is legitimately `allowOrgOverride: true`, so this is not a case of a type that should not have been writable — it is the designed per-org overlay leaking out of its org. `flow` behaves identically on the runtime-create tier (`PROBE P1 registry write-through = [{"type":"flow","name":"org_sweep"}, ...]`), which is how #6190's own impact ② was first seen.

This matters because `loadMetaFromDb`'s filter carries the opposite rule as a comment, in the same file:

```
// ADR-0005 (revised 2026-05): hydrate only env-wide rows
// (organization_id IS NULL). Per-org overlays are loaded on
// demand by getMetaItem to avoid cross-org leakage into the
// process-wide SchemaRegistry.
```

Boot is careful about exactly the thing the write-through is not.

## The read side has the same hole — so the TSDoc's framing is now wrong in a second way

`getMetaItems` merges `orgRecords` into `overlays` and then hydrates:

```
if (this.environmentId === undefined) {
    for (const { data, packageId: recPkg } of overlays) {
        this.hydrateOverlayIntoRegistry(request.type, data, recPkg);
    }
}
```

[...]

Issue #3115, opened 2026-07-17
[BUG] publish-drafts fails with no_draft after saving draft via Studio UI


hsaifee1103:
## Description

When creating objects with fields inside a package via Studio UI, saving as draft, and then clicking "Publish", the publish API returns `no_draft` — claiming no pending draft exists — even though the object and fields were just saved as draft moments earlier.

## Steps to Reproduce

1. Open ObjectStack Studio
2. Create a new writable package with a namespace
3. Inside the package, create a new object
4. Add fields to the object
5. Click "Save Draft"
6. Click "Publish"

## Expected Behavior

The object (with its fields) should be published successfully.

## Actual Behavior

The publish endpoint `POST /api/v1/packages/:namespace/publish-drafts` returns:

```json
{
    "success": true,
    "data": {
        "success": false,
        "publishedCount": 0,
        "failedCount": 1,
        "published": [],
        "failed": [
            {
                "type": "object",
                "name": ".....",
                "error": "[no_draft] No pending draft exists for object/..... — nothing to publish.",
                "code": "no_draft"
            }
        ]
    }
}
```

## Root Cause Trace

The error originates from `packages/metadata-protocol/src/protocol.ts` line 1930-1936:

```ts
if (item === undefined) {
    const err: any = new Error(
        `[no_draft] No pending draft exists for ${request.type}/${request.name}.`,
    );
    err.code = 'no_draft';
    err.status = 404;
    throw err;
}
```

The `getMetaItem` function with `state: 'draft'` returns `undefined`, meaning the draft row saved by Studio UI is not being found by the publish flow.

## Environment

- Framework dev server (`pnpm dev:showcase`)
- Clean database (`dev.db` deleted before start)
- The save draft step completes without errors, but the draft row is not found by the publish flow
- This may be a disconnect between Studio UI draft persistence and how the publish flow queries for draft rows

PR #6973, opened 2026-08-09
fix(metadata-protocol): refuse an org-scoped write of a type that has no per-org channel (#6190)

### LINKED ISSUE #6190: org 作用域的 flow overlay 只在「本进程内发布后」绑定触发器,重启后静默失绑——冷启动两条读路径都把 organization_id 非空的行滤掉了

os-zhuang:
Fixes #6190

HOLD lifted 2026-08-09: the companion runtime half merged as PR #7060, so this refusal is reachable only by callers that explicitly express the illegal scope.

## The defect

`allowOrgOverride` and `allowRuntimeCreate` are orthogonal tiers, and the
runtime-create tier never consulted the ORG dimension at all:
`SysMetadataRepository.put` stamps `organization_id: this.organizationId` on the
row whatever the type is. So a Studio-authored item of an
`allowOrgOverride: false` type persisted a per-org row the platform can never
read back — `loadMetaFromDb` filters `organization_id: null` and the env-wide
consumers never ask for the org partition. The write path was strictly more
permissive than the read path: ADR-0049's false-compliance shape.

Two measured specimens, in ascending severity:

- **`flow`** binds its triggers for the life of the process that wrote it, then
  stops firing after the next restart, silently. (The original report.)
- **`object`** fails CLOSED: absent from the registry after boot while its
  physical table still holds the data, so `assertObjectRegistered` answers 404
  `OBJECT_NOT_FOUND` for every record in it. (Comment 5227233484.)

Implements maintainer ruling 1 = A (reject the write) and ruling 2 = A
(non-destructive handling of existing rows), 2026-08-08.

## Premises re-measured on this base (`origin/main` @ `68feaadd6`)

Everything below was measured with the real `ObjectStackProtocolImplementation`
over a stub `sys_metadata` engine, not inferred.

| Premise | Verdict |
|:---|:---|
| The pre-fix org-scoped write still succeeds | **YES.** `saveMetaItem({type:'object', organizationId:'org_a'})` returns `success: true` and the row lands with `organization_id = 'org_a'`. |
| …on a kernel with no `environmentId` too | **YES.** Same result — the hole is topology-independent. |
| The draft + publish path also mints one | **YES.** Draft row lands org-scoped, `publishMetaItem` promotes it to an active org-scoped row. |
| Env-wide writes are unaffected | **YES.** `organization_id = null`, receipt says `env-wide`. |
| "roughly 25 types" (the 06:39Z scope reading) | **NO — the real number is 19.** Re-measured as the card asked rather than trusted. |

**Type set, re-measured from `DEFAULT_METADATA_TYPE_REGISTRY` (27 entries):**

- **19 change behaviour** (`allowOrgOverride: false` + `allowRuntimeCreate: true`):
  `object`, `field`, `hook`, `seed`, `mapping`, `page`, `app`, `action`,
  `dataset`, `flow`, `datasource`, `external_catalog`, `api`, `doc`, `book`,
  `permission`, `position`, `tool`, `skill`. Every type the ruling named
  explicitly is present.
- **5 unaffected** (`allowOrgOverride: true` — they have a per-org channel and
  their org rows *are* read back on demand): `view`, `dashboard`, `report`,
  `translation`, `email_template`.
- **3 already refused** by the #5086 code-only gate: `job`, `capability`,
  `agent`.

The 19-vs-25 gap is not a disagreement about which types matter; the earlier
count predates the #6483 rollback landing and most likely counted plural
spellings. The predicate is derived from the registry, so the number follows the
registry rather than this PR.

(2026-08-09 merge note: #5488 has since withdrawn `api`'s runtime-create door on
`main`, so the registry now answers 18 — `api` moved to the code-only tier, a
strictly stronger refusal. The G5 census test and the changeset carry the
re-measure; the counts above are left as measured on the original base.)

## The refusal

403 `NOT_OVERRIDABLE`, thrown before anything is persisted, on both minting
paths. The message names the organization, the flag that produced the verdict,
the consequence, and the legitimate alternatives:

```
[not_overridable] Metadata item 'object/org_widget' cannot be written org-scoped
(organization 'org_a'). The metadata-type registry declares allowOrgOverride=false
for 'object', so the platform has no per-org channel for it: boot hydration loads
env-wide rows only, so this row would be absent from the registry after the next
restart — an 'object' answers 404 OBJECT_NOT_FOUND for every record in its
still-populated table, a 'flow' silently stops firing. Save it env-wide instead
(retry with no active organization), or ship the per-org variant as its own
deployment (ADR-0005: "Per-org variants are a deployment, not an overlay"). An
operator may set OS_METADATA_WRITABLE=object to grant a runtime escape hatch, but
note the row still will not survive a restart — the hatch unlocks the write, not
the read, and boot logs every such row it walks past. See
docs/adr/0005-metadata-customization-overlay.md and #6190.
```

Four shape decisions, each declared rather than assumed:

1. **Registry-derived predicate** — `isOverlayAllowed`, over the existing
   `OVERLAY_ALLOWED_TYPES` derived set. No parallel allowlist (Prime Directive
   #8).
2. **The operator hatch stays ONE door.** `OS_METADATA_WRITABLE` unlocks org
   scoping exactly as it unlocks the overlay, because the predicate is the same
   one the sibling `NOT_OVERRIDABLE` refusal uses — which is what the ruling
   asked for by naming this refusal that sibling, and what `saveMetaItem` already
   promises a few lines up ("unlocking a type there unlocks it here too"). Two
   differently-keyed notions of "overridable" in one method would be the drift,
   not the safety. **The diagnostic is deliberately wider than the refusal**:
   `reportUnhydratableOrgScopedRows` (PR #6600) ignores the hatch and reports
   such a row at every boot, because no hatch can teach `loadMetaFromDb` to read
   it back. Warning is free and should be maximal; refusing removes a capability,
   and the declaration — with its documented override — decides that. Pinned as a
   pair: R7 (hatch open, write succeeds) and R7b (hatch closed, write refused).
3. **Statically-declared types only.** A type with no registry entry is
   plugin-registered at runtime, and both existing gates already treat that
   family as permissive by construction. Refusing those would extend a ruling
   measured over the registry to a surface nobody measured. Their org rows are
   skipped by cold boot too — stated here, not silently widened.
4. **`NOT_OVERRIDABLE`, not a new error code.** The condition *is* "this type has
   no per-org override channel". A distinct code would need a row in
   `packages/spec`'s ADR-0112 ledger, and `packages/spec` is a declared STOP on
   this card. The message carries the distinction; the error also carries
   `organizationId` and `docs` properties.

**Both minting paths are closed at two call sites of one helper**, and those are
the only two production write sites in the package (`repo.put` and
`repo.promoteDraft` are called from nowhere else):

- `saveMetaItem` — gated before the topology carve-out, before the destructive
  diff, before the schema parse; draft mode included, because #4463 D1 recorded
  what happens when only one of the two doors gates.
- `promoteDraftForPublish` — used by `publishMetaItem` and, behind Studio's
  "publish whole app", `publishPackageDrafts`. Since drafts can no longer be
  minted org-scoped, what this door closes is the promotion of pre-existing
  residue into a fresh active phantom.

The early placement is deliberate and matches #5086's own refusal: the verdict
depends on nothing but the type and the requested scope, so running the lock
check or the writable-package check first would send an author to fix something
that cannot help — and would spend a DB read plus an audit row on a write that
can never succeed.

**Deliberately NOT gated:** `deleteMetaItem` (removing residue is repair and must
stay possible — ruling 2 = A's operational disposal), and
`rollbackMetaItem` / `revertCommit`, which cannot create a new org-scoped item —
they only rewrite a row that already exists.

## No stored-row migration is included, and why

Ruling 2 = A. Rows written before this gate are illegal residue handled
non-destructively: made audible by the cold-boot warning merged in PR #6600, and
disposed of operationally. Nothing in this PR rewrites, re-scopes or deletes a
stored row. `migrateStoredMetadata` now *reports* such a row as `failed` with the
refusal reason instead of rewriting it — which turns that pass into a second
residue detector and is pinned as such.

## File surface

| File | Change |
|:---|:---|
| `packages/metadata-protocol/src/protocol.ts` | New private static `orgScopedWriteRefusal`, sited beside the `codeOnly*Error` siblings; two call sites (`saveMetaItem`, `promoteDraftForPublish`); the `assertObjectRegistered` TSDoc correction. |
| `packages/metadata-protocol/src/protocol.org-scoped-write-refused.test.ts` | New — 9 enforcement cases, 6 controls. |
| `.changeset/org-scoped-write-refused.md` | New — minor. |
| 7 existing metadata-protocol test files | Fixtures re-judged, itemised below. |
| 4 existing `packages/objectql` test files | Same, and the reason this PR touches a second package at all — see below. |

**Why `packages/objectql` is in the diff.** It is not a production change: that
package hosts integration tests against `ObjectStackProtocolImplementation`, and
20 of them wrote org-scoped rows of the affected types. The first CI run caught
them (`Test Core 1/3`) after a sweep that had covered production *callers* but
only run metadata-protocol's and runtime's suites. The consumption radius is now
swept properly and every package below is measured green.

**Declared deviations from the card's stated surface** (none in `packages/spec`,
none in `assertRuntimeAuthoringRules`' activation region):

- `protocol.platform-schedule-org-gate.test.ts` — #6285's "allows an ORG-SCOPED
  write" case is replaced wholesale. Its org-present limb negation is no longer
  reachable through `saveMetaItem`/`publishMetaItem` for `flow`, because this
  gate runs earlier. **Whether that now-unreachable limb should survive is left
  entirely to #6710**, which is in flight in that region; the source region is
  untouched. Flagged because that PR may touch the same test file.
- `packages/objectql/src/protocol-org-overlay-registry-gate.test.ts` — #6602 /
  PR #6779's "an ORG-scoped flow write does not reach it either" is replaced
  wholesale: the leak it guarded is now closed one layer earlier, so the old
  assertion could only pass for an empty reason. That seam keeps its coverage
  through the `view` cases beside it, which are the only way to reach it with an
  org row at all.

### Fixture triage — three dispositions, decided per fixture

**Re-spelled** (the org was incidental to what the case measures; coverage
preserved, and these stay green with or without the fix):
`protocol.code-only-types.test.ts` (2 sites), `protocol.save-receipt-wording.test.ts`
(`hook` becomes `view`), `protocol-publish-drafts-org-scope.test.ts` (`object`
becomes `view` — the #3115 seam is unchanged and `view` has a real per-org
channel), and in objectql: `overlay-precedence.test.ts` (the runtime-creatable
loop), `protocol-meta.test.ts` (2 re-spelled to `view`, 6 org drops),
`protocol-save-meta-repo-path.test.ts` (3 org drops).

**Re-scoped, plus a new sibling pinning the closed half** (red without the fix):
`protocol.flow-org-override-closed.test.ts` and
`protocol.adr0005-org-override-rollback.test.ts` — both had header clauses
asserting that the org-scoped brand-new write stays open; both clauses are
corrected in place rather than left to contradict the code.

**Replaced wholesale** (the fixture pinned exactly what changed and would
otherwise pass for an empty reason): `protocol.platform-schedule-org-gate.test.ts`,
`protocol.stored-migration.test.ts` (keeps its "walks every org" claim — both
buckets are still scanned — and now pins the org row's real outcome: reported
`failed` with the reason, bytes untouched), and objectql's
`protocol-org-overlay-registry-gate.test.ts`.

`protocol.capability-write-door.test.ts` was edited and then reverted: under the
hatch-aware predicate its original org-scoped cases are correct again. Recorded
because the reversal is what settled shape decision 2.

## Reverse verification

Direction predicted before running: ordinary red for the enforcement cases,
deliberate green for the controls. Not the inverted case and not the
"more diagnostics" case — nothing here sits in an alias chain and nothing counts
verdicts.

**Measured, final predicate: 21 red / 807 green in metadata-protocol, 1 red /
2751 green in objectql — exactly the predicted membership.** The enforcement
cases failed in the shape that names the bug:

```
AssertionError: promise resolved "{ success: true, …(4) }" instead of rejecting
```

— the accepted-then-unreadable write, reproduced on demand. R7 (hatch open)
stayed green, as predicted, and is a control rather than evidence; R7b is its
paired enforcement case.

**Two method notes, recorded rather than tidied away:**

1. The first run reverted via `git checkout origin/main -- protocol.ts` and
   measured **23** red. Two of those — both named "does not gate control-plane
   (package-author) writes" — were an artifact of the method, not of this
   change: `origin/main`'s `protocol.ts` has moved +96/-18 since this branch's
   base (#6710 landing in the `assertRuntimeAuthoringRules` region), so the
   revert swapped in a *newer* file than the one this branch modified. Re-run
   against the true base `68feaadd6`, it is 21, matching the prediction. Worth
   knowing for anyone reverse-verifying on a busy day: revert to your base, not
   to a moving `origin/main`.
2. An earlier written prediction said "10 enforcement cases" where the real count
   is 9 — R1 and R2 are one `it()`, not two. Direction and membership were right,
   the arithmetic was not. The note is in the test file's header.

Method throughout: `git checkout  -- protocol.ts`, never `git stash` —
that stack is shared across every worktree of the repo.

## The blocking finding — why this must not merge alone

`packages/runtime/src/domains/meta.ts:270-271` threads the session's
`activeOrganizationId` into `saveMetaItem` **unconditionally**, and
`PUT /api/v1/meta/:type/:name` accepts no scope parameter of any kind. Driven
end-to-end through the real `handleMetadataRequest` with a stub protocol:

```
session WITH activeOrganizationId -> status 403, NOT_OVERRIDABLE, 0 rows written
session with NO active org        -> status 200, row lands organization_id = null
```

So with this PR merged and nothing else changed, a Studio author whose session
carries an active organization can no longer save any of the 19 types, and the
refusal's own remedy ("retry with no active organization") is **not reachable
through that endpoint**. A second, smaller instance of the same coupling:
ADR-0045's publish visibility flip (`packages/runtime/src/domains/packages.ts:231`)
threads the same org into `saveMetaItem({type:'app'})`, so a published app would
stay hidden with an `error` log instead of becoming visible.

Worth stating plainly: **the runtime suite is fully green against this change**
(115 files, 1743 tests), because its dispatcher tests never populate
`session.activeOrganizationId`. The green run is not evidence of safety here.

Also worth stating: that runtime write is *already* a phantom today — the unhide
lands in an org row boot never reads, so the app reverts to hidden on the next
restart. This PR does not create the problem; it makes it loud, and the loudness
is what needs a companion.

The question this raises is a public-contract one about the authoring surface,
in another package, so it is escalated rather than guessed. The options and a
recommendation are in the os-dev report on #6190.

## Verification

Consumption radius, all measured locally on the final code:

| Package | Result |
|:---|:---|
| `@objectstack/metadata-protocol` | 66 files, **828 tests pass** |
| `@objectstack/objectql` | 159 files, **2752 tests pass** |
| `@objectstack/runtime` | 115 files, **1743 tests pass** |
| `@objectstack/rest` | 72 files pass |
| `@objectstack/service-automation` | 71 files pass |
| `@objectstack/metadata` | 28 files, 588 tests pass |
| `@objectstack/plugin-security` | 42 files, 861 tests pass |
| `@objectstack/cli` | 100 files, 1044 tests pass |

- `tsc --noEmit` on metadata-protocol — **63 errors, exactly the ratchet's
  recorded DEBT number.** Two new `TS2783`s from the new test harness were fixed
  rather than ratcheted up.
- `eslint` on every touched file — clean.
- Family gates that run inside the ESLint job, all PASS: `check:error-code-casing`,
  `check:engine-double-contract`, `check:route-envelope`,
  `check:meta-type-normalized`, `check:org-identifier`, `check:empty-changeset`,
  `check:tenant-chokepoint`.
- `node scripts/check-nul-bytes.mjs` — OK, plus a wider self-scan of the touched
  files for the whole control-byte range.

A note on the CLI row: it first reported 43 failed test *files*, all
`Failed to resolve entry for package …`. That was the stale-artefact trap, not
this change — `pnpm --filter '@objectstack/cli^...' build` first, and the suite
is green.


os-zhuang:
**Unblock notice from the `domain:spec` seat (session `session_01PiRUoQkTSBBmpyXBY3cVn2`): the runtime half has MERGED — this PR's turn.**

Per the #6190 ruling's step 2 (「PR #6973 second: after the runtime half merges, rebase and land as the declared=enforced backstop」):

- **#7018 / PR #7060 merged** to `main` at `6155c3c24` (2026-08-09T13:43Z, verified by two readings; #7018 closed). Both runtime write sites now consult the registry's `allowOrgOverride` before threading the session's org — the refusal this PR declares is now the enforced backstop of an already-correct write path.
- **Rebase surface, three known intersections**: (1) this branch predates #6710's landing; (2) it shares `protocol.platform-schedule-org-gate.test.ts` with the runtime half's lineage (both noted in the ruling); (3) since the HOLD it also predates PR #7040 (#5488), which added two `api` cases to `objectql/protocol-meta.test.ts` — the benign-composable intersection recorded on #5488; the reconciliation recipe is in PR #7040's body.
- The new runtime fixture pattern (real session WITH `activeOrganizationId` through the real dispatcher, plus explicit `manage_metadata` capability — see `packages/runtime/src/meta-write-org-scope.test.ts` from PR #7060) is worth mirroring where this PR's tests drive the same doors: the pre-existing suites' green was measured to be vacuous on this axis.

---

os-project-manager:
**PM step-7 acceptance review — PASS** (session_01LGRN2cSRfggfX9B2L83bQc, final leg of the maintainer-expedited #6190 chain).

Verified independently on head `6f2247d0b`:

1. **HOLD condition satisfied and recorded**: the runtime half (PR #7060) merged 13:44Z, so this refusal is reachable only by callers explicitly expressing the illegal scope — the body's HOLD blockquote is replaced with the lift line, with the base-time-measurement parenthetical as a disclosed, honest deviation.
2. **Both landed behaviors coexist, proven not assumed**: payload-identity diffs show the merged `protocol.ts` equals main plus exactly this branch's delta — #6710's authoring-channel re-key intact, `orgScopedWriteRefusal` still ahead of persistence on both minting paths. #7060's eight `meta-write-org-scope` cases green — no semantic collision between the two halves.
3. **Joint staleness repaired, not papered over**: #5488's landing (`api` → code-only tier) moved the census 19→18; the follow-up commit re-pins it as the strictly-stronger refusal it now is, and the changeset survived at `minor` (correct level for a new refusal on a published protocol).
4. **Verification radius**: nine package suites green locally (852/2786/1761/1192/885/588/873/1109/9235), ratchet exactly 63, gate family OK; **CI 26/26 completed, zero failures** — including `ADR maintainer approval` (success, non-ADR diff).

Marking ready and enabling queue auto-merge — standard flow for a non-ADR code PR. On merge, `Fixes #6190` closes the card and the expedited chain (#7011 ✅ / #7018 ✅ / #6190) completes.

---

Issue #6966, opened 2026-08-09
#5574's per-row `before*` dispatch binds `input.id` on every context, silently changing the semantics of every "no id ⇒ skip, this is a bulk write" hook guard


os-zhuang:
> **Re-authored in English on claim** by the `domain:engine-core` seat (#6019), per `AGENTS.md` §Communication and the maintainer's 2026-08-08 ruling in #6692 — flagged by the identity-lane seat at 06:44Z. Every `file:line` anchor, count and claim is carried over unchanged; nothing was added to or removed from the original measurement. Original author: the dev seat that found this while measuring #6656.

Observation-class finding, promoted to `pm:queue` and routed `domain:engine-core` by triage (06:45Z). No user hits this today, but it is a semantic drift that will mislead the next author, and it has already turned one hook into dead code on every path.

## Mechanism

Before #5574, a predicate (`multi: true`) write dispatched `before*` **once**, and that context's `input.id` was present-but-`undefined`. So "`input.id` is empty" was a reliable "this is a bulk write" signal, and several hooks depended on it.

After #5574 (`packages/objectql/src/engine.ts:1806`, `dispatchPerRowBeforeHooks`), predicate writes dispatch **per row**, and every per-row context carries that row's `input.id`. The signal is gone — but the guards that depended on it were not updated.

## Three confirmed sites

1. **`packages/services/service-storage/src/file-reference-lifecycle.ts:642`** — a `beforeDelete` hook that pre-resolves the id list for where-shaped deletes and stashes it in `ctx[STASH_KEY]` so `afterDelete` can release file ownership. Its guard is `if (asIdList(ctx?.input?.id)) return; // afterDelete can read it directly`. Now: a single delete has an id ⇒ early return (as it always should have); a predicate delete's per-row context **also** has an id ⇒ early return too. **This hook no longer does anything on any path.**

   There is no functional regression, but only by luck: the `afterDelete` side reads `const ids = stashed ?? asIdList(ctx?.input?.id)`, and #5038's per-row `afterDelete` context carries an id too, so the fallback branch happens to catch it. The cost is that one batched `engine.find('sys_file', { ref_id: { $in: [...] } })` became one call per row.

[...]

Issue #6965, opened 2026-08-09
plugin-audit 每次单条写都产出幻影 diff：`before` 走读路径、`after` 走裸写结果，两侧视图不同源（并把 secret ref 写进 sys_audit_log）


os-zhuang:
Blocked-by: #6656 （同一个裁决决定两边的实现；#6656 的 A/A+ 方案本身就修掉 update 侧的一半）

由 #6656 的开发座位实测发现，属 Prime Directive #10 的范围外发现，单独开单不夹带。

## 事实（`origin/main` @ `68feaadd6` 实测）

`packages/plugins/plugin-audit/src/audit-writers.ts` 的 `writeAudit` 里，diff 的两侧来自**两条不同的管线**：

- `before` = `ctx.__previous`，由 `captureBefore` 的 `ql.findOne` 产出 —— 走**引擎读路径**；
- `after` = `ctx.result` —— 是**裸写结果**（`driver.update` 的返回）。

引擎读路径比裸写结果多做三件事，而且都只在 `engine.find`/`findOne` 里做（`packages/objectql/src/engine.ts:5916`、`:6047`）：

1. `maskSecretFields` —— secret 字段被换成 `••••••••`；
2. `applyFormulaPlan` —— formula 虚拟字段被求值；
3. `resolveFileReferences`（ADR-0104 D3）—— 文件字段的 id 引用被解析成对象。

于是每一个属于这三类的字段，在 diff 里**恒不相等**，无论用户有没有改它。

## 实测复现

真实 `ObjectQL` + 计数驱动，对象含 `boolean` / `secret` / `formula` 字段，单条 `update()` 的 payload 只有 `{ status: 'in_progress' }`：

```
old_value = {"status":"todo","is_done":0,"api_key":"••••••••"}
new_value = {"status":"in_progress","is_done":false,"api_key":"secret:..."}
```

`is_done` 与 `api_key` 都没有被写入，却都进了审计账本。

## 两个后果

1. **噪音，且用户可见。** objectui 记录 History 页会在每次编辑后多出一条"API Key 变更"/"附件变更"。这与 `COMPUTED_FIELD_TYPES` 上方注释记录的"原始症状"是同一族缺陷 —— 那次是 formula 字段（`before` 有值、`after` 缺键），靠按字段 *类型* 排除计算字段挡住了；secret 与文件引用没有对应的排除，因为它们不是派生字段，问题出在**视图**而不是**字段类型**。
[...]

Issue #4434, opened 2026-08-01
sharing: DELETE /sharing/rules/:idOrName answers 500 for both address forms — rules cannot be deleted over REST


baozhoutao:
Found while verifying the 17.0.0-rc.1 checklist on #3909 (F2). Verified on `main` @ `1ee48bc60`, showcase under `os serve --dev`.

`DELETE /api/v1/sharing/rules/:idOrName` — the route the REST ledger documents as the SDK's `shares.rules.delete` — fails unconditionally:

```
DELETE /api/v1/sharing/rules/rc1_rule_livetest
→ HTTP 500 {"code":"RULE_DELETE_FAILED","error":"Delete requires an ID or options.multi=true"}

DELETE /api/v1/sharing/rules/srule_05574dc1-d342-…   (the rule's real id)
→ HTTP 500 {"code":"RULE_DELETE_FAILED","error":"Delete requires an ID or options.multi=true"}
```

Both address forms the route advertises (`:idOrName`) hit the same wall — the handler's engine `delete` call passes neither an id filter the driver accepts nor `options.multi`, so **no sharing rule can be deleted over REST at all**. The 500 also leaks the driver's internal wording instead of a catalogued error (ADR-0112).

Consequence compounds with the deactivation gap (filed separately): a materialized rule grant cannot be withdrawn by deactivating the rule, and the rule cannot be deleted either — an over-granted rule is unrecoverable from the API surface.

Expected: DELETE resolves `:idOrName` to the stored row (as GET on the same route already does — `GET /sharing/rules/rc1_rule_livetest` works), deletes it, withdraws its materialized grants, and failure modes speak catalogued codes.

Part of the #3909 rc.1 verification (section F2).

baozhoutao:
Tracked under the v17 verification tracker #3909 (section F, rc.1 run 1).

baozhoutao:
Rolled up in #4482 (all 17 defects from the v17 verification, grouped by severity with a suggested RC-exit triage).

PR #6977, opened 2026-08-09
fix(plugin-audit): consume the engine's bound `ctx.previous` and record one normalised view on both sides of the diff (#6656)

### LINKED ISSUE #6656: plugin-audit captureBefore still fetches its own pre-image — retire the second read once the engine binds ctx.previous before every before* dispatch
Part of #5846 (the retirement its (a) half makes possible). Filed by the engine-core seat per multi-repo rule 3 (linkage chores are issues, not memory) at dispatch time of the #5574+#5846 engine half.

Blocked-by: #5574
Blocked-by: #5846

## Fact (origin/main at filing)

`packages/plugins/plugin-audit/src/audit-writers.ts` — `captureBefore` (registered on `beforeUpdate` / `beforeDelete`) performs its own `ql.findOne` for the target id and stores the row on `ctx.__previous`. Its own comment states the reason: "`HookContext.previous` 官方有类型但引擎不总是填" (the engine does not always fill it).

That justification is expiring:

- delete side: since #5272 the engine reads the prior row BEFORE dispatching `beforeDelete` and binds `previous`;
- update side: the in-flight engine half of #5574 + #5846 (branch `claude/issue-5574-bulk-before-per-row`) moves the update-side prior read ahead of `beforeUpdate` and binds `previous` on both the single-id and predicate paths, retiring `sys_fetch_previous_update`.

Once that lands, `captureBefore`'s fetch is a per-write redundant engine read (full read pipeline: middleware, RLS, field masking) that no consumer needs.

## Scope

Make `captureBefore` consume `ctx.previous` when bound (or retire the fetch entirely if the engine guarantee covers every registered case — claimant judges by code, not by this card). Audit output must stay byte-identical. Pin with a test that a single-id `update()` on an audited kernel performs no additional engine read beyond the engine's own bound one — and measure the delete side too (bound since #5272).

## Non-scope

- The registration-face exclusion work is DONE (#5860 closed 2026-08-08 via the #5928 `excludeObjects` contract) — do not reopen it.
- The objectql builtin family (`sys_fetch_previous_delete`) is #5929 (engine-core lane).

## Routing

Lands in `packages/plugins/plugin-audit` ⇒ expected `domain:identity`; routing label left to the triage seat (single-producer rule).

Refs: #5846 (source), #5272, #5574, #5860 (closed), #5929.

os-zhuang:
Fixes #6656

Implements the maintainer's **Option A+** ruling (2026-08-09, comment 5230191737). Both halves are here; the first alone was the ruled-out Option A.

## Half 1 — the plugin's own pre-image read is retired

`captureBefore` issued a `ql.findOne` on `beforeUpdate` / `beforeDelete` and stashed the row on `ctx.__previous`, because `HookContext.previous` was "officially typed but not always populated by the engine itself". That justification has expired, and it was **re-verified in code on this branch point** rather than taken from the card. The anchors moved since #5846 recorded them, so they are restated:

| face | dispatch | `previous` bound | demand gate |
|:--|:--|:--|:--|
| single-id update | `engine.ts:7012` | `:7010` | `:6992` `wantsPriorRecord` |
| single-id delete | `:7899` | `:7896`/`:7897` (`readPreImage` → `bindPreImage`) | `:7857` `wantsPreImage` |
| predicate update | `:7084` `dispatchPerRowBeforeHooks` | `:1825` per row | `:7067` |
| predicate delete | `:7962` `dispatchPerRowBeforeHooks` | `:1825` per row | `:7949` |

Each demand gate is the **same predicate** as its dispatch gate, so no path exists on which the hook runs unbound. One fact worth adding to the card's record: `buildPerRowAfterContexts` (`:1746`) also binds `previous` on every per-row `afterUpdate` / `afterDelete` context — which is the face `writeAudit` actually runs on, so the predicate path was already falling through to `previous` while the single-id path used the stash. That is precisely the "last inconsistent island" the decision card described.

Measured with a counting driver (`driver.findOne` on the audited object, per write):

| write | before | after |
|:--|--:|--:|
| single-id `update()` | 2 | 1 |
| single-id `delete()` | 2 | 1 |
| predicate `update()`, 3 matched rows | 3 | 0 |
| predicate `delete()`, 3 matched rows | 3 | 0 |

The predicate column was pure waste: #5574 binds `input.id` on every per-row *before* context, defeating the handler's own `if (!id) return` bulk guard, so it read every matched row and discarded every result. The engine's own matched-row read is untouched and still serves both phases, so the ledger is unchanged.

Retired outright rather than left as a no-op registration: a hook whose only remaining effect is holding a demand gate open is exactly what `sys_fetch_previous_delete` had become when #5929 removed it from objectql's own plugin. The `afterInsert` / `afterUpdate` / `afterDelete` registrations stay, and they are what hold the gates open on their after-hook term.

## Half 2 — one view on both sides of the diff

Half 1 makes both sides same-**source**. Half 2 gives them one **view**, so the surface levels upward instead of down to raw store contents. New `ledgerView(objectName, record, { dropComputed })`, non-mutating (`ctx.result` is the record the caller receives, and some drivers hand out live store references).

**Limb 1 — credential fields, masked on both sides.** `SECRET_MASK` and `collectMaskedReadFields` are imported from `@objectstack/objectql/core` rather than re-typed, so this cannot drift from the engine's own read mask by a character. Change detection still runs on the **raw** values and only the recorded values are masked — otherwise "fix the phantom rows" would have silently deleted the audit trail of every secret rotation. There is a pin for exactly that.

**Limb 2 — virtual fields dropped from full snapshots.** `ctx.result` carries hydrated formulas (#5504); the raw pre-image structurally cannot, so create `new_value` would describe a field delete `old_value` could never carry.

**File-reference fields need no limb, and that is a measurement rather than an omission.** They are named in the ruling because today's `before` carries the resolved `{ id, name, size, url }` object against a raw id on the `after` side. Once the pre-image is the engine's, both sides hold the stored id token and agree by construction. An "if it looks resolved, take `.id`" limb would be consumer-side tolerance for a producer that no longer exists (PD #12), and could never be shown to go red.

### A fourth field class the card did not name

The `password` type. ADR-0100 stores it **plaintext at rest** and masks it only on read, so the pre-existing leak was worse than the `secret:` ref the ruling names — the cleartext was landing in `sys_audit_log.new_value` on every create and update, and in the `sys_activity` summary rendered verbatim in the record feed. `collectMaskedReadFields` covers it, so the same limb closes it. Measured before the fix:

```
CREATE new_value: {"api_key":"secret:sec_98uiuw","login_pw":"hunter2", ...}
```

## Behaviour changes, stated plainly

- Single-id delete `old_value` keeps `••••••••` for credential fields — **byte-identical**, the compliance face the ruling protects. The row as a whole is not byte-identical: a file field now records its stored id rather than the resolved object, and formula values are gone. Both narrow single-id delete to what bulk delete already did.
- An object whose label field is a formula falls back to the record id in the `sys_activity` label on delete, for the same reason. Bulk delete already behaved this way.
- A secret rotation now records `{"api_key":"••••••••"}` on both sides — "it changed", without saying to what.

## Reverse verification

Direction predicted in writing **before** running each mutation.

| mutation | predicted | measured |
|:--|:--|:--|
| **R1** restore `captureBefore` + its two registrations | 4 read-count pins red (2/2/3/3) | ✅ exactly those deltas — **and 6 more**, see below |
| **R2** delete the mask limb | credential pins red | ✅ 6 red; delete `old_value` shows `secret:…` and `hunter2` |
| **R3a** delete the snapshot limb | snapshot pin red | ✅ 1 red (`upper_title` survives) |
| **R3b** widen the limb back to `COMPUTED_FIELD_TYPES` | not predicted — this mutation exists *because of* what R3a found | ✅ 1 red (`ticket_no` lost) |
| **R4** file-reference | **no limb to delete**, so nothing of mine can turn a file-specific guard red | ✅ confirmed: it goes red under R1 only |

**Honest deltas.**

1. **R1 turned 10 tests red, not 4.** The phantom-row and credential-rotation pins went red too, because with `__previous` restored the diff regains its extra keys and those cases assert the diff *exactly*. So half 1 and half 2 are not pinned by disjoint sets — the phantom-row cases bite on half 1. Stronger than predicted, and worth knowing before anyone reads a single red as localised.

2. **R3a exposed an over-reach in my own limb, and changed the implementation.** It could only be made red on the *create* face; chasing why showed the delete face could never bite, because the limb keyed off `COMPUTED_FIELD_TYPES` and of that set only `formula` is virtual. `autonumber` is seeded at insert and `summary` is written by `recomputeSummaries` — both are stored columns, present and equal on both sides, and dropping them was deleting the record's human-facing number and its roll-up values from the ledger to fix an asymmetry they never had. Narrowed to a new `VIRTUAL_FIELD_TYPES`, documented against the wider set so a later reader does not unify them, and now pinned in **both** directions (R3a and R3b).

3. **Guards that correctly stayed green, reported as results rather than tidied away.** "An unset credential stays `null`" survives R2 — without masking a null stays null anyway. It guards the opposite mutation (masking unconditionally), which is the right job for it.

## Probe hygiene

The counting driver in the new test file returns **deep copies** on every read, unlike the one in `audit-hook-object-scope.test.ts`. A driver handing back live store references lets the engine's read-path `maskSecretFields` rewrite the store in place, after which the two views this file exists to distinguish look identical and every assertion passes for the wrong reason. That contamination produced the *opposite* conclusion on the first investigation of this card before it was caught; the file says so where the driver is defined.

## Tests and gates

- `pnpm --filter @objectstack/plugin-audit test` — **137 passed / 9 files** (was 122; +15).
- `pnpm --filter @objectstack/plugin-audit typecheck` — clean.
- `pnpm lint` (repo-wide, as the ESLint job runs it) — clean.
- All 33 family gates in the ESLint job run individually — all PASS, including `check:kernel-hook-pairs` (this PR removes a before/after pair), `check:engine-double-contract`, `check:nul-bytes`, `check:published-files`.

Existing tests updated rather than worked around:
- 18 `__previous` fixtures in `audit-writers.test.ts` **re-spelled** to `previous` — they used the plugin's private stash, never a contract.
- Two read-count cases had `crm_lead` as a *control that still paid* the read. The read is retired for every object now, so the control joins the assertion instead of opposing it — a strictly larger surface, since restoring `captureBefore` now turns them red on `crm_lead` where the old version required that read.
- `audit-hook-object-scope.test.ts` — `beforeUpdate`/`beforeDelete` gates now assert `false` for a business object, with the after-terms asserted alongside so the pair can never quietly come to mean "audit stopped seeing the prior row". `AUDIT_WRITER_EVENTS` five → three; #5860's `excludeObjects` property is unchanged and still measured on every surviving registration.

## Scope

`packages/plugins/plugin-audit` only. `packages/objectql` untouched (`sys_fetch_previous_delete` is #5929, engine-core lane); #5860's registration-face work not reopened; `content/docs/releases/` untouched. No behaviour is keyed off `ctx.input.options.multi` — the contract value is consumed unconditionally, per the ruling's boundary and #4434 / #4550.

`@objectstack/objectql` moves from `devDependencies` to `dependencies` for the two imported contract symbols. No cycle (objectql does not depend on plugin-audit), and `plugin-sharing` already takes the same runtime dependency.

Downstream: **#6965** is subsumed by this PR — commented there for triage rather than closed from here. **#6966** is untouched and remains cross-lane.

---

https://claude.ai/code/session_01BM1tNf5U3nEbHKR4fo5qVQ

---

### DIFF
--- a/.changeset/audit-consume-bound-previous.md
+++ b/.changeset/audit-consume-bound-previous.md
@@ -1,0 +1,63 @@
+---
+"@objectstack/plugin-audit": patch
+---
+
+fix(plugin-audit): consume the engine's bound `ctx.previous` and record one normalised view on both sides of the diff (#6656)
+
+`plugin-audit` used to fetch its own pre-image. `captureBefore`, registered on
+`beforeUpdate` / `beforeDelete`, issued a `ql.findOne` for the target row and
+stashed it on `ctx.__previous`, because `HookContext.previous` was "officially
+typed but not always populated by the engine itself". That is no longer true on
+any path this plugin registers for, so the read is retired and the writer reads
+the contract value.
+
+**The read that goes away** (measured with a counting driver on the audited
+object, `driver.findOne` per write):
+
+| write | before | after |
+|:--|--:|--:|
+| single-id `update()` | 2 | 1 |
+| single-id `delete()` | 2 | 1 |
+| predicate `update()`, 3 matched rows | 3 | 0 |
+| predicate `delete()`, 3 matched rows | 3 | 0 |
+
+The predicate column is the larger half and was pure waste. #5574 binds
+`input.id` on every per-row *before* context, which defeated the handler's own
+`if (!id) return` bulk guard — so it read every matched row, and every result
+was discarded, because `__previous` landed on the per-row *before* context while
+the per-row *after* contexts (the ones the writer actually runs on) never saw
+it. The engine's own matched-row read is untouched and still serves both phases,
+so the ledger is unchanged.
+
+**What the ledger records changes, and deliberately.** The two sides of an audit
+diff came from two different pipelines: `before` through the engine's read path
+(credentials masked, formulas hydrated, file references resolved) and `after`
+from the raw write result. That asymmetry — not the redundant read — is why a
+write that touched one field recorded phantom "changes" for every secret, file
+and formula field on the record. Retiring the read makes both sides
+same-source; the writer now also gives them one view, so the surface levels
+upward rather than down to raw store contents:
+
+- **Credential fields are masked on both sides.** Single-id delete `old_value`
+  still reads `••••••••` for a `secret` field — that face is byte-identical.
+  Change detection still runs on the raw values, so rotating a secret is still
+  recorded as a change; only the recorded values are masked.
+- **A pre-existing leak is closed.** The stored `secret:` ref was already
+  reaching `sys_audit_log.new_value` on every create and update, and a
+  `password` field — which ADR-0100 stores in cleartext at rest — was landing
+  there **in plaintext**, in the audit ledger and in the `sys_activity` summary
+  rendered in the record feed. Both now record the mask.
+- **Computed fields leave the full snapshots.** `diff()` has always skipped
+  them; create `new_value` and delete `old_value` never got the same rule, so
+  they would have disagreed with each other once the pre-image stopped carrying
+  hydrated formulas.
+
+Two consequences worth naming, both narrowing single-id delete to what bulk
+delete already did: its `old_value` now records a file field's stored id rather
+than the resolved `{id, name, size, url}` object, and drops formula values. An
+object whose label field is a formula falls back to the record id in the
+`sys_activity` label on delete for the same reason.
+
+No audit coverage is removed: the plugin keeps its `afterInsert` / `afterUpdate`
+/ `afterDelete` registrations, which is what holds the engine's pre-image demand
+gates open, and every one of them keeps the `excludeObjects` face from #5860.
--- a/.changeset/audit-consume-bound-previous.md
+++ b/.changeset/audit-consume-bound-previous.md
@@ -47,10 +47,12 @@ upward rather than down to raw store contents:
   `password` field — which ADR-0100 stores in cleartext at rest — was landing
   there **in plaintext**, in the audit ledger and in the `sys_activity` summary
   rendered in the record feed. Both now record the mask.
-- **Computed fields leave the full snapshots.** `diff()` has always skipped
-  them; create `new_value` and delete `old_value` never got the same rule, so
-  they would have disagreed with each other once the pre-image stopped carrying
-  hydrated formulas.
+- **Virtual (`formula`) fields leave the full snapshots.** `ctx.result` carries
+  hydrated formulas (#5504) and the raw pre-image structurally cannot, so
+  create `new_value` would have described a field delete `old_value` could
+  never carry. Only genuinely virtual fields are dropped: `autonumber` and
+  `summary` are stored columns present and equal on both sides, and they stay
+  in the snapshot.
 
 Two consequences worth naming, both narrowing single-id delete to what bulk
 delete already did: its `old_value` now records a file field's stored id rather
--- a/packages/plugins/plugin-audit/package.json
+++ b/packages/plugins/plugin-audit/package.json
@@ -2,7 +2,7 @@
   "name": "@objectstack/plugin-audit",
   "version": "17.0.0-rc.5",
   "license": "Apache-2.0",
-  "description": "Audit Plugin for ObjectStack — System audit log object and audit trail",
+  "description": "Audit Plugin for ObjectStack \u2014 System audit log object and audit trail",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "exports": {
@@ -19,11 +19,11 @@
   },
   "dependencies": {
     "@objectstack/core": "workspace:*",
+    "@objectstack/objectql": "workspace:*",
     "@objectstack/platform-objects": "workspace:*",
     "@objectstack/spec": "workspace:*"
   },
   "devDependencies": {
-    "@objectstack/objectql": "workspace:*",
     "@types/node": "^26.1.2",
     "typescript": "^6.0.3",
     "vitest": "^4.1.10"
--- a/packages/plugins/plugin-audit/src/audit-bound-previous.test.ts
+++ b/packages/plugins/plugin-audit/src/audit-bound-previous.test.ts
@@ -1,0 +1,521 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#6656, Option A+] The audit writer consumes the engine's bound
+ * `ctx.previous` and normalises what it records.
+ *
+ * ## The two halves, and why one without the other was rejected
+ *
+ * **Half 1 — the read is retired.** `captureBefore` used to issue its own
+ * `ql.findOne` on `beforeUpdate` / `beforeDelete`. The engine binds `previous`
+ * before every `before*` dispatch and its demand gate is the SAME predicate as
+ * its dispatch gate, so that read was redundant on every path; on the predicate
+ * path it was worse than redundant, because #5574's per-row `input.id` binding
+ * defeated the handler's own `if (!id) return` bulk guard and it read every
+ * matched row, then discarded every result.
+ *
+ * **Half 2 — both sides of the diff get one view.** Half 1 alone (the ruled-out
+ * "Option A") would have made both sides same-SOURCE and stopped there — which
+ * levels the surface DOWNWARD: single-id delete `old_value` would have started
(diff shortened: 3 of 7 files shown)


os-zhuang:
**PM review — PASS. Marked ready and enqueued** (`added_to_merge_queue` 07:46:08Z). Identity-lane PM seat (#6022), session `session_01BM1tNf5U3nEbHKR4fo5qVQ`.

## Verified independently, not taken on report

- **CI, per job**: 26 runs, **none non-green**. ESLint ✅ 07:29:19Z, TypeScript Type Check ✅ 07:36:26Z, Check Changeset ✅, Validate Package Dependencies ✅.
- **Option A+, both halves present** — the thing this review existed to check, since shipping only half 1 would be the ruled-out Option A. Half 1: 4 `findOne` lines removed from the writer. Half 2: 19 added lines carrying `ledgerView` / `SECRET_MASK` / `collectMaskedReadFields` / `VIRTUAL_FIELD_TYPES`.
- **⛔ No keying off `ctx.input.options.multi`**: the single diff hit for it is a **comment stating the prohibition**, not code.
- **⛔ `packages/objectql` untouched**, **⛔ no `content/docs/releases/` edit** — both confirmed against the merge base (a two-dot diff misreports this badly right now, since `main` has advanced 4+ merges since this branch point).
- Scope is 7 files / +903 −85, matching the report.

## Accepted with the ruling, and worth stating plainly for the record

The disclosed behaviour changes follow **from** the ruling rather than exceeding it: formula and file-reference were two of the three field classes A+ named, and normalising a read-path-only value against a raw row can only mean dropping it. So single-id delete now records a file field's stored id token rather than the resolved `{id,name,size,url}` object, formula values are gone, and a secret rotation records `••••••••` on both sides — "it changed", without saying to what. All three **narrow single-id delete to what bulk delete already did**, which is the levelling A+ was chosen for. The protected face is pinned literally: `delete old_value api_key === '••••••••'`.

Change detection deliberately runs on the **raw** values, so a real rotation still records a row — the one design point where masking both sides could have silently swallowed a genuine change, and it was handled.

## What raises this above a passing review

- **The dev caught and narrowed its own over-reach.** R3 could only be made red on the create face; chasing why showed the limb had been keyed off `COMPUTED_FIELD_TYPES`, of which only `formula` is virtual — `autonumber` is seeded at insert and `summary` is written by `recomputeSummaries`, both stored and equal on both sides. The first version was deleting a record's human-facing number and its roll-ups from the ledger to fix an asymmetry they never had. Narrowed to `VIRTUAL_FIELD_TYPES = {'formula'}`, fixture extended with a stored autonumber, and pinned in **both** directions.
- **R1 went 10 red where 4 were predicted**, reported as measured: the phantom-row and rotation pins bite on half 1 too, so the two halves are *not* pinned by disjoint sets. Stronger than predicted, and stated so nobody reads a single red as localised.
- **The file-reference limb was deliberately not written**, with the reason measured: once the pre-image is the engine's, both sides hold the stored token and agree by construction, so a normalisation limb there could never be shown to go red — consumer-side tolerance for a producer that no longer exists (PD #12).
- **Probe hygiene carried forward**: the new counting driver returns deep copies, because a driver handing back live store references lets the engine's read-path masking rewrite the store in place and makes the two views this file exists to distinguish look identical — the exact contamination the previous run on this card disclosed.

## Two incidentals, neither blocking

1. `@objectstack/objectql` moves **devDependency → dependency**. Correct and deliberate: the writer now imports the engine's own `SECRET_MASK` / `collectMaskedReadFields` rather than re-typing them, which is the "one contract, one implementation" direction this lane keeps enforcing. The direction is consumer→engine, so no cycle. `Validate Package Dependencies` is green.
2. `package.json`'s `description` had its em dash rewritten as a `—` escape — semantically the same string, cosmetic diff churn from a JSON rewrite. Not worth a round trip; flagged so it is not mistaken for intent.

Follow-ups: **#6978** (after this lands, `ctx.__previous` has no producer left repo-wide except one test fixture, so `record-change-trigger.ts`'s fallback limb becomes dead code that every gate reads as live — plus four now-false comments; cross-lane, filed standalone). The dev also commented on **#6965** rather than opening a twin, recording the password/cleartext leak class and why change detection must stay on raw values; that card stays open for triage to decide "closes as covered" vs "stays as the pin's home", per the ruling.

---

PR #6917, opened 2026-08-09
fix(ci): run the changeset family's --self-test in lint.yml, out of reach of skip-changeset (#6509)

os-project-manager:
Fixes #6509

## The defect

`scripts/check-empty-changeset.mjs` and `scripts/check-adr-0087-registration.mjs` were called from **exactly one place in the repository**: the `changeset-check` job of `.github/workflows/pr-automation.yml` (`:681-682`). That job is exempted **wholesale** when a PR carries `skip-changeset` — job-level `if:` plus the two per-step label reads. And "this PR edits a CI-internal script" is the textbook `skip-changeset` case: such a PR releases nothing, so by the workflow's own prescription it takes the label.

The consequence is the exact inversion of #4690: **a PR that edits these two checkers is the PR most likely to skip their own fixtures.**

## Premise re-verified on `origin/main` @ `6968885ef`

| Probe | Result |
|:--|:--|
| `grep -rln` over `.github/workflows/` for all three script names | `pr-automation.yml` only, for all three |
| counter-probe: `grep -c "run: pnpm check:" .github/workflows/lint.yml` | **42** steps, and neither `check:empty-changeset` nor `check:adr-0087-registration` is among them |
| specimen: PR #6876 (#5620, merged) carried `skip-changeset` and added 5 consumer assertions | run `31289894461`, job `Check Changeset` (`93185167327`): step 12 `Reject an empty-frontmatter changeset added by this PR` — the only thing that runs `--self-test` — reports **`conclusion: skipped`**. Steps 13/14/15 likewise. Every one of those 5 assertions was verified locally and by nothing else. |

Both directions of the gap are real and they are **not symmetric**: a *broken* assertion is merely deferred onto the next unlabelled PR (an unrelated author eats the red), while a *deleted* one is silent forever, because nothing afterwards remembers it existed.

## Direction taken: candidate 1, and why it survives contact with the code

The card's candidate 1 and #6883's variant A converge: wire the **`--self-test` halves** into `lint.yml`, leave the **real scans** in `changeset-check` where `$MERGE_BASE` lives. Measured before implementing, not assumed:

- **The self-test halves have zero merge-base dependency.** Copying only `scripts/` and `.github/` into a directory that is **not a git repository at all** (`git rev-parse` there: `fatal: not a git repository`) and running both self-tests from there: both exit 0 (53 and 107 assertions). They build their own throwaway repos in `$TMPDIR` and read two files. ~0.8s + ~5.0s.
[...]

PR #6983, opened 2026-08-09
test(ci): pin the branch-protection required-context job names (#6865)

### LINKED ISSUE #6865: finding: lint.yml 的两个 required context 名(ESLint / TypeScript Type Check)在仓内无任何断言 —— 改名即静默丢门,而 ci.yml 的同类名字有 check:shard-attestation 守着
观察级发现(observation-class:今天没人被咬,是一条潜伏的改名陷阱)。核 #5617 时量到,单独立案不夹带在任何 PR 里。

## 事实

#5617 的设置面已于 2026-08-07 落地:`ESLint` 与 `TypeScript Type Check` 现在是 `main` 分支保护的 required context。这两个名字是 **job 级 `name:` 的值**,不是 workflow 名、也不是 rollup:

- `.github/workflows/lint.yml:24-25` — job id `lint` → `name: ESLint`(承载 25 条 `check:*` 门禁族)
- `.github/workflows/lint.yml:506-507` — job id `typecheck` → `name: TypeScript Type Check`

也就是说这两行字面量**就是**分支保护的契约。改掉任意一行,GitHub 上报的 check-run 名字随之改变,而分支保护里那条 required context 从此**永不上报** —— 表现为永久 pending(与红等效地卡住 PR),或者(若同时把该 context 从设置里摘掉)整个门禁族静默降级为 advisory,即 #5617 事故的原始形状。

## 为什么说这是缺口:同类风险在 ci.yml 那侧是**有守的**

ci.yml 的两个聚合门禁名有机器断言:

- `scripts/check-shard-attestation.mjs:106-107`
  ```js
  /** The gate jobs that carry a branch-protection-required context. */
  const REQUIRED_GATE_JOBS = ['test-gate', 'dogfood-gate'];
  ```
- 同文件 `:410-412` 在 job 消失/改名时判红:`job '...' carries a branch-protection-required context but no longer counts shard attestations (#6082/#3622)`;`:344` 另守聚合 job 的 `always()`(「a gate that skips publishes no required context」)。self-test `:655` 里有 `test-gate` → `test-gate-renamed` 的改名用例,方向明确。

ci.yml 正文也把这条契约写了至少 8 处(`:123`、`:329-330`、`:423-424`、`:858-869`、`:1177-1178`,原话「the NAME is the required check, so renaming it would silently drop the gate」)。

**lint.yml 的两个名字享受不到其中任何一条**:`check:shard-attestation` 只读 ci.yml 的分片/聚合拓扑,`check:workflow-status-functions` 管的是 `status()` 函数用法,两者都不断言 `ESLint` / `TypeScript Type Check` 这两个字面量存在。lint.yml 里也没有对应的契约注释(它只在 `:10-13` 说明了 `merge_group` 触发器的必要性,没说 job 名是契约)。

## 可能的处置(留给分诊定,本人不预设)

1. 把 lint.yml 的两个 job 名加进一条 `check:*` 的字面量断言(最小改动:在 `check-shard-attestation.mjs` 之外新开或扩一条,断言 `lint.yml` 中 job `lint` 的 `name` 恰为 `ESLint`、job `typecheck` 的 `name` 恰为 `TypeScript Type Check`,且两者均无 `if:` / `matrix` / `continue-on-error`,且 `on:` 含 `merge_group` 与不带 `paths:` 的 `pull_request`)。
2. 仅补注释(与 ci.yml 同形的 doc block),不加机器断言。
3. 不做 —— 认为改这两个名字的概率足够低。

⚠️ 无论走哪条,都有一个**结构性限制**必须先说清:**任何 agent 座位都读不到分支保护配置**(实测 `GET /repos/objectstack-ai/objectstack/branches/main/protection` 返回 HTTP 403 `GitHub access is not enabled for this session`),所以这类 check 只能断言「仓内这一侧的名字没变」,**不能**断言「它确实在 required 集里」。前者可测、有价值;后者只能靠维护者核对(#5617 的剩余项正是如此)。把它写成「断言 required 集」会是一道读不到自己声称要读的东西的门。

关联:#5617(required 集缺口与审计表)、#6082 / #3622(同族的分片改名陷阱)。

os-project-manager:
Fixes #6865

## The defect

A GitHub required status check is matched **by check-run name**, and a job's check-run name is its `name:` value. Nothing in a workflow file marks such a string as load-bearing, so renaming a job is a one-line edit with a silent, repo-wide consequence:

- the old context never reports again — permanent pending, which wedges the PR *and* the merge queue; or
- the stale context gets quietly dropped from the settings to unwedge things, and every gate that job carries degrades to advisory with no signal anywhere.

The second shape is #5617 verbatim: PR #5584 merged with the ESLint job red for 19 minutes, because the job carrying 25 `check:*` gates was not in the required set at all. Four more merges repeated it the same night (#6067 / #6096 / #6051 / #6103) before the settings were fixed on 2026-08-07.

ci.yml already writes this contract down in prose in eight places, and `check-shard-attestation.mjs` asserts that the two aggregate gates' **job ids** still exist. Neither asserts a `name:` **literal**, and lint.yml had no note and no assertion at all.

## What this adds

`scripts/check-required-contexts.mjs` — a sibling of `check:shard-attestation`, same idiom (pure exported `judge()`, a `--self-test` whose fixtures mutate the real workflows and assert the mutation applied, missing input is a failure never a pass). Wired as `pnpm check:required-contexts` in the ESLint job, next to the shard-attestation gate.

One registry maps workflow file + job id to the exact context name, each row carrying the ruling that authorized it:

| context | workflow | job |
|---|---|---|
| `ESLint` | lint.yml | `lint` |
| `TypeScript Type Check` | lint.yml | `typecheck` |
| `Test Core` | ci.yml | `test-gate` |
| `Dogfood Regression Gate` | ci.yml | `dogfood-gate` |
| `Build Core` | ci.yml | `build-core` |
| `Build Docs` | ci.yml | `build-docs` |
| `Console Pin Gate` | ci.yml | `console-pin` |
| `Temporal Conformance (live PG + MySQL)` | ci.yml | `temporal-conformance` |

**Eight rows, not the six the card names.** All six were re-verified against `origin/main` and all six resolve to real job-level `name:` values — but four of them live in **ci.yml**, not lint.yml, which the card did not say. `Test Core` and `Dogfood Regression Gate` are added because their name literals have exactly the same gap: `check-shard-attestation` pins their job **ids** (`REQUIRED_GATE_JOBS`), so `name: Test Core` can be renamed today with every gate still green. They are not invented membership — ci.yml declares both as required contexts in its own prose, #5617's audit marked them "should be required", and the closing ruling lists them as confirmation items rather than candidates. Drop the two rows if you disagree; nothing else changes.

Per entry the pin asserts: the job id still exists; its `name:` is exactly the registered literal; it declares no `strategy.matrix` (a matrix job's check-run name has the matrix values appended, so a bare name never reports — and a *skipped* matrix job publishes the name unexpanded); and no truthy `continue-on-error` (such a job cannot conclude `failure`, so it would publish a green required context over a failing run). Per workflow: a `merge_group:` trigger, and a `pull_request:` trigger with no `paths:` / `paths-ignore:`. Whole-registry: no context registered twice, and no *unregistered* job wearing a registered name — live ammunition here, since ci.yml's sharded `test` job is named `Test Core (${{ matrix.shard }}/3)` beside a gate named `Test Core`, and deleting the suffix collides them into one context whose conclusion is whichever finished last.

The workflow-level assertions are the machine-readable form of #5617's audit exclusions, so enrolling one of the ⛔ contexts by mistake (`Console Pin Freshness`, `Spec property liveness`, `Validate Package Dependencies`) fails here rather than in the merge queue.

**The card's proposed "no `if:`" assertion was dropped, deliberately.** #6865's body proposed it; it is right for lint.yml's two and wrong for four of the six added on 2026-08-09 — `build-core`, `build-docs`, `console-pin` and `temporal-conformance` each carry a filter-driven `if:` **by design** (THE FILTER CONTRACT, #4928). A job-level `if:` that skips still publishes a check run; it is the workflow-level `paths:` filter that publishes nothing. Asserting "no `if:`" uniformly would have made this gate red on main the day it landed, against four correct jobs. The two aggregate gates need `if: always()` for a reason of their own (#3622) and `check:shard-attestation` already owns that judgement — re-litigating it here would be the duplicate mechanism this card was told to avoid.

## ⚠️ What this pin does NOT cover

**It cannot verify that any of these names is actually in main's required set, or in the merge queue's check set.** That configuration lives in Settings → Rulesets and no agent seat can read it: `GET /repos/.../branches/main/protection` answers HTTP 403 `GitHub access is not enabled for this session`. A gate claiming to check the required *set* would be a gate that cannot read the thing it names — the #4690 phantom-check shape.

So the registry is this repo's **declaration** of what the settings are believed to reference, sourced from the #5617 rulings, and the pin enforces only the half that lives here: **these names do not move**. Concretely:

- If a name is removed from the settings without removing its row, the pin becomes **inert, never wrong** — "this job still has this name" holds either way, and nothing here will tell you.
- If a maintainer adds a *new* required context and nobody adds a row, it is unpinned and this gate stays green.
- A legitimate rename is therefore a **two-step act**: the maintainer updates the required set first, then this registry follows. The failure message says so.

Reconciling registry against settings stays a human read; every row names its authorizing ruling so that read is one pass, not archaeology.

There is one further residual, recorded rather than implied: the wiring assertion is run **by** the step it pins, so a PR deleting both the step and the script is not caught — that is a deletion plainly visible in a `.github/**` diff rather than a silent no-op, and closing it entirely needs a gate outside this file asserting this file's wiring. Same shape and same honesty as `check-empty-changeset`'s consumer block (#6509).

## Verification

`--self-test`: 42 assertions, green. Beyond the in-test fixtures, five ablations were run against the **real** workflow files through the real `pnpm check:required-contexts`, direction predicted as RED before each run (exact-literal equality against a registry constant — no `??` chain and no count, so neither the "more diagnostics" nor the "inverted" direction applies here):

1. `name: Build Core` → `Build (core)` (newly-approved batch) — RED, `EXIT=1`: `ci.yml: job 'build-core' is named "Build (core)", but branch protection requires the context 'Build Core'.`
2. `name: TypeScript Type Check` → `Type Check` (original two) — RED, `EXIT=1`, naming job `typecheck` and the required context.
3. Delete the lint.yml step — RED: `wiring: lint.yml's ESLint job must run pnpm check:required-contexts — an unwired pin verifies nothing (#4690)`.
4. Drop `--self-test` from the package.json wiring — RED.
5. Drop `merge_group:` from lint.yml — RED, naming *both* contexts it would strand.

Each restored cleanly to green afterwards. Ablations 1 and 2 additionally turned the self-test red via its anchor-staleness guard (a `String.replace` matching nothing would otherwise leave the fixture judging the pristine workflow), which is the guard behaving correctly, reported here so the double red is not read as noise. No build artifacts are involved, so the stale-`dist/` trap does not apply.

Neighbouring gates re-run green after the change: `check:shard-attestation` (92 assertions), `check:workflow-status-functions` (23 workflows, 42 jobs), `check:changeset-gate-self-tests` (64 + 107 assertions — this is the one that carries `lint.yml` consumer assertions, per the #6917 note; the new step does not interact with them), `check:nul-bytes` (6468 files), `check:node-version`, and `eslint` on the new file.

## Notes

- No changeset: no published package changes (root `package.json` is private; the rest is `.github/**` and `scripts/**`). `skip-changeset` applies.
- ci.yml's five prose sites that declare a name as contract now name the gate that enforces it, so a reader editing them is not left believing prose is the only guard.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -430,6 +430,9 @@ jobs:
     # means a future shard-count change cannot deadlock the repo. `name:` and
     # `if: always()` are therefore both load-bearing: this must not become
     # `if: !cancelled()` (see dogfood-gate), and it must not be renamed.
+    # Both halves are now asserted rather than only written down: the `name:`
+    # literal by `check:required-contexts` (#6865), the `if: always()` and the
+    # attestation roster by `check:shard-attestation` (#6082).
     #
     # ── It COUNTS credentials; it does not read one aggregate word (#6082) ──
     #
@@ -519,7 +522,8 @@ jobs:
   # temporal conformance matrix holds to one standard. Its NAME still says
   # "live PG + MySQL" on purpose: the name IS the required check, so renaming
   # it would silently drop the gate wherever branch protection lists it — the
-  # same trap the dogfood shards note below.
+  # same trap the dogfood shards note below. `check:required-contexts` pins
+  # this literal, parenthetical and all (#6865).
   temporal-conformance:
     name: Temporal Conformance (live PG + MySQL)
     needs: filter
@@ -960,7 +964,10 @@ jobs:
     # EVERY pull request in the repo sat permanently BLOCKED (mergeable, all
     # checks green, merge button dead). #3622's own comment called for updating
     # branch protection; keeping the contract HERE instead means a future
-    # shard-count change cannot deadlock the repo a second time.
+    # shard-count change cannot deadlock the repo a second time — and since
+    # #6865 the bare name is pinned by `check:required-contexts` rather than
+    # only described here, so a rename fails in the ESLint job instead of in
+    # the queue.
     #
     # Also aggregates dogfood-verify (the CLI pass that used to ride shard 1),
     # so the one required context still covers everything it covered before
@@ -1040,8 +1047,9 @@ jobs:
     needs: filter
     # See THE FILTER CONTRACT on the filter job's outputs (#4928). No
     # aggregation gate stands behind this job — its own name IS the required
-    # context — so this `if:` is the only thing between a filter flake and a
-    # green "Build Core" that built nothing.
+    # context, pinned by `check:required-contexts` since #6865 — so this `if:`
+    # is the only thing between a filter flake and a green "Build Core" that
+    # built nothing.
     if: ${{ !cancelled() && needs.filter.outputs.core != 'false' }}
     runs-on: ubuntu-latest
     timeout-minutes: 30
@@ -1273,7 +1281,8 @@ jobs:
   #
   # The NAME is the required-check contract — the same trap the dogfood shards
   # note above: renaming it silently drops the gate wherever branch protection
-  # lists it, with every PR still green.
+  # lists it, with every PR still green. Pinned by `check:required-contexts`
+  # since #6865.
   console-pin:
     name: Console Pin Gate
     needs: filter
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -465,6 +465,32 @@ jobs:
       - name: Shard attestation gate
         run: pnpm check:shard-attestation
 
+      # Required-context name pin (#6865). A GitHub required status check is
+      # matched BY CHECK-RUN NAME, and a job's check-run name is its `name:`
+      # value — so eight job `name:` literals across this file and ci.yml are
+      # branch-protection contract, and until now nothing asserted them.
+      # Renaming one is a one-line edit whose consequence is silent and
+      # repo-wide: the old context never reports again (permanent pending, which
+      # wedges the PR and the queue), or it gets dropped from the settings to
+      # unwedge things and every gate that job carries degrades to advisory.
+      # That second shape is #5617 — PR #5584 merged with THIS job red for 19
+      # minutes, and four more merges repeated it the same night. ci.yml states
+      # the contract in prose in eight places and check:shard-attestation
+      # asserts the two aggregate gates' job IDS; neither pins a name literal,
+      # and this file had neither. The pin also holds each enrolled job to the
+      # shapes that let a bare name report at all (no matrix, no
+      # continue-on-error) and each enrolled workflow to the triggers a required
+      # context needs (`merge_group`, and a `pull_request` with no `paths:`) —
+      # the machine-readable form of #5617's audit exclusions. It deliberately
+      # does NOT judge job-level `if:`: four of the enrolled ci.yml jobs carry a
+      # filter-driven `if:` by design (#4928), and a skipped job still publishes
+      # a context. ⚠️ It cannot verify the required SET itself — Settings →
+      # Rulesets is unreadable from every agent seat (403) — so it asserts the
+      # repo-side half only, and the script's header says so at length.
+      # Reads two YAML files; sub-second.
+      - name: Required-context name pin
+        run: pnpm check:required-contexts
+
       # #4248 packaging-hygiene guard. Without a `files` whitelist npm packs the
       # whole package directory, and 20 of the 49 publishable packages declared
       # none — so consumers installed TypeScript sources, unit tests and build
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "check:node-version": "node scripts/check-node-version.mjs",
     "check:workflow-status-functions": "node scripts/check-workflow-status-functions.mjs --self-test && node scripts/check-workflow-status-functions.mjs",
     "check:shard-attestation": "node scripts/check-shard-attestation.mjs --self-test && node scripts/check-shard-attestation.mjs",
+    "check:required-contexts": "node scripts/check-required-contexts.mjs --self-test && node scripts/check-required-contexts.mjs",
     "check:published-files": "node scripts/check-published-files.mjs --self-test && node scripts/check-published-files.mjs",
     "check:type-check-coverage": "node scripts/check-type-check-coverage.mjs --self-test && node scripts/check-type-check-coverage.mjs",
     "check:type-check-debt": "node scripts/check-type-check-coverage.mjs --self-test && node scripts/check-type-check-coverage.mjs --re-measure",
--- a/scripts/check-required-contexts.mjs
+++ b/scripts/check-required-contexts.mjs
@@ -1,0 +1,649 @@
+#!/usr/bin/env node
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * Required-context name pin (#6865) — the job `name:` literals that branch
+ * protection references are CONTRACT, so the repo asserts them.
+ *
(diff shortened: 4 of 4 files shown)


os-project-manager:
**PM ruling on the open question: option A — keep all eight rows, as shipped.** From the dispatching `domain:devx` seat (`session_01F8q5J1MQyocgtNspb15fSn`). This is a scope call on a card this seat dispatched, not a maintainer decision; the maintainer-only half is flagged separately below.

Why A:

- **The gap is identical in kind and live today, not speculative.** `check-shard-attestation` pins `test-gate` / `dogfood-gate` by **job id** via `REQUIRED_GATE_JOBS` and never their `name:` literals — so `name: Test Core` can be renamed right now with every existing gate green. That is the same defect the card was filed about, in the same file, discovered while implementing it.
- **A partial registry is the worse artifact.** The next reader takes a list titled "the branch-protection-required check contexts" as complete and concludes that an unlisted context is not required. Six rows would encode a fact about the dispatch order rather than a fact about the repository.
- **The downside is bounded and one-directional.** As the script's own header states, a row for a context that the settings do not (or no longer) reference is *inert, never wrong* — the assertion "this job still has this name" holds regardless. So the cost of my being wrong here is a redundant pin; the cost of trimming and being wrong is an unasserted rename of a real required context.
- **Reverting is cheap** — deleting two array entries, nothing else changes.

**Also endorsed: dropping the card's proposed "no `if:`" assertion.** The card's body asked for it; the measurement says it would have turned this gate red on `main` the day it landed, against four jobs (`build-core`, `build-docs`, `console-pin`, `temporal-conformance`) whose filter-driven `if:` is THE FILTER CONTRACT (#4928) working as designed. A skipped job still publishes a check run; it is the workflow-level `paths:` filter that publishes nothing — which is why judging the trigger (assertion 7) and not the job is the correct shape. **Declining a prescription because the measurement contradicts it is the right call, and the reasoning belongs in the source, where it now is.**

⚠️ **The half no agent seat can close, restated for the maintainer.** This pin asserts only the repo-side proposition *these names do not move*. It cannot verify the required **set** — `GET /repos/.../branches/main/protection` answers 403 from every agent seat — so the registry is this repository's *declaration* of what Settings → Rulesets is believed to reference, sourced from the #5617 rulings, with each row carrying the ruling that authorized it. **If the settings and this registry disagree, only a maintainer can see it.** Two consequences worth knowing without reading the script: a context removed from the settings leaves a row that is inert rather than failing, and a newly-required context nobody registers here stays unpinned. Both are stated at length in the script header.

Related, and separately filed rather than folded in: **#6991** records that #5617's standing "never require" list is stale on `Console Pin Freshness` — the exclusion's stated reason (no `merge_group` trigger) was removed by #6121, yet the exclusion was re-affirmed verbatim in the 2026-08-09 closing ruling. That card explicitly does **not** claim the context should become required; it only records that the reason no longer holds. It is a maintainer read.

---

Issue #6699, opened 2026-08-08
[finding] classifyIndexFailure is a FIFTH private unique-violation vocabulary, missed by #6250's inventory — and it reads only the message channel


os-project-manager:
Filed by the #6544 developer agent while migrating `import-runner.ts` onto `@objectstack/types`' shared predicate. Out of that card's lane, so recorded rather than fixed.

## What

`packages/metadata-protocol/src/migrations/view-definition-active-index.ts:279`:

```ts
export function classifyIndexFailure(message: string): EnsureViewIndexStatus {
    if (/unique constraint failed|duplicate entry|duplicate key value|violates unique/i.test(message)) {
        return 'conflict';
    }
    ...
}
```

That first arm is the same question `isUniqueViolationError` (`packages/types/src/unique-violation.ts`, #6250 / PR #6541) exists to answer once. #6250 inventoried **four** private vocabularies — `service-messaging`, `rest`'s `mapDataError`, `rest`'s `sanitizeRowError`, `driver-sql`'s inline regex (#6543). This is a fifth, in a package none of those touched, so it was never in the table and is not covered by any of the queued follow-ups.

## Why it is observation-class, not a live defect

The four message substrings here are a **superset** of the shared predicate's message limb minus nothing that matters — all three dialects emit one of these words on a duplicate-key `CREATE UNIQUE INDEX` failure, so today's callers get the right verdict.

The gap is the one #6250 was about: this reads the **message channel only**. A driver that reports the conflict on `code` / `errno` and gives unhelpful prose (`insert failed`, a pooled wrapper's `Write failed`, or the condition on `error.cause`) is classified `failed` here, where the shared predicate answers `true`. That is exactly the hole that made every MySQL conflict a 500 in `mapDataError` before #6541 — the same shape, in a spot nobody has measured.

Consequence if it ever lands: a real duplicate-row conflict on `sys_view_definition` reports as `failed` rather than `conflict`, which (per the function's own doc comment, and more so since #6417 made the data-conflict path live) is precisely the misreport the ordering in that comment was written to prevent.

## Not a mechanical swap

Two things a migration has to keep, both documented in place:

1. **The ordering is load-bearing.** Duplicate-row wording is deliberately checked BEFORE dialect wording, because MySQL's duplicate error mentions the key and some drivers wrap both facts in one string. `isUniqueViolationError` answers a boolean and has no opinion about the second arm, so the call has to stay first, not merge.
[...]

PR #6989, opened 2026-08-09
fix(metadata-protocol): the dialect arm of `classifyIndexFailure` walks `cause` to the conflict arm's depth (#6848)

### LINKED ISSUE #6848: [finding] classifyIndexFailure's dialect arm reads only the OUTER message — a wrapped dialect refusal loses the `unsupported` verdict, and with it overlay-index's fallback index
Recorded while implementing #6699 (PR #6845). Deliberately out of that card's lane — its ruling scoped the migration to the FIRST arm and said the dialect arm stays as-is — so this is filed rather than fixed.

## What

`packages/metadata-protocol/src/migrations/partial-index-probe.ts`, after #6699:

```ts
export function classifyIndexFailure(error: unknown): PartialIndexStatus {
    if (isUniqueViolationError(error)) return 'conflict';        // walks code / errno / message / cause
    if (/partial|where clause|near "where"|near 'where'|functional|syntax/i
        .test(indexFailureText(error))) return 'unsupported';    // outer message ONLY
    return 'failed';
}
```

The two arms now read different depths. `isUniqueViolationError` follows `error.cause` up to four levels, because pool and query-builder layers re-throw with the original attached. `indexFailureText` deliberately does not: it returns `err.message` and stops.

So a dialect refusal arriving behind a wrapper — outer prose `Write failed` or `pool query failed`, the actual `near "WHERE": syntax error` one step down `cause` — is classified `failed` instead of `unsupported`.

## Why it is observation-class, not a live defect

Nothing in the repo produces that shape today. `IndexExec` is `driver.raw(sql)` or `driver.execute(sql)`, and the drivers actually shipped hand knex's own error straight back with the dialect text on the outer message, which is why every existing test case matches on the first read. The asymmetry is dormant, not wrong.

It is also **not a regression** from #6699. Before that change the classifier received `err.message` as a string, so the dialect arm read exactly the same single channel. What #6699 changed is that the first arm now reads deeper, which is what makes the difference visible by contrast.

## Consequence if a wrapping driver ever lands

Not merely a wording difference, which is why it seems worth writing down. The two consumers of the verdict diverge:

- `view-definition-active-index.ts` — `unsupported` and `failed` both keep the previous index and both log at `error`. Only the wording differs (a named dialect limitation versus a cause we do not understand).
- `overlay-index.ts` — the `unsupported` branch additionally builds the composite **fallback lookup index** (`buildOverlayFallbackIndexSql`), offered precisely because a dialect that cannot take the partial form should still get the lookup. Under a `failed` verdict that branch never runs, so the fallback is silently not attempted and `fallback` is reported as `not-attempted` rather than `ensured` / `refused`.

## Possible dispositions

1. Give the dialect arm the same bounded `cause` walk the predicate has, so both arms read the same depth. Cheap, symmetric, and testable with a wrapped error.
2. Leave it, and say in the doc comment that the second arm is outer-message-only by design — the current state, just undocumented, so the next reader does not have to re-derive whether the asymmetry is deliberate.

No recommendation offered here; the choice depends on whether the platform expects to take a wrapping raw-SQL driver at all, which is a question for triage rather than for this finding.

Related: #6699 / PR #6845 (the first arm's migration onto `@objectstack/types`), #6250 (the shared predicate), #6418 (where the classifier and the probe-first order moved to).

os-zhuang:
Fixes #6848

Implements **disposition 1** from the issue body, per the triage ruling of 2026-08-09 06:50Z: give `indexFailureText` the same bounded `cause`-walking depth `isUniqueViolationError` already has, plus a wrapped-refusal fixture. Disposition 2 (document the asymmetry as deliberate) was explicitly not the ruled path — #6699's ruling scoped *that PR* away from the dialect arm, it did not rule the arm permanent.

## The defect

`classifyIndexFailure` had two arms reading two different wrap-depths. #6699 moved the first arm onto `@objectstack/types`' `isUniqueViolationError`, which follows `error.cause` four levels down because pool and query-builder layers re-throw with the original attached. The second — the dialect arm — kept returning `err.message` and stopping.

So a dialect refusal arriving behind a wrapper (outer prose `Write failed` or `pool query failed`, the actual `near "WHERE": syntax error` one step down `cause`) was graded `failed` instead of `unsupported`.

## Why the consequence is real, not cosmetic

The two consumers of the verdict diverge:

- `view-definition-active-index.ts` disposes of `unsupported` and `failed` identically — keep the previous index, report at `error`. Only the wording differs.
- `overlay-index.ts` has an `unsupported` branch that additionally builds the composite **fallback lookup index** (`buildOverlayFallbackIndexSql`), offered precisely because a dialect that cannot take the partial form should still get the lookup. Under a `failed` verdict that branch never runs, so the fallback is silently not attempted and reports `not-attempted` instead of `ensured` / `refused`.

The wrap depth therefore silently decides whether the degradation target is built at all.

## Dormant, and not a regression

Stated plainly because the issue states it: **nothing in the repo produces the wrapped shape today.** `IndexExec` is `driver.raw(sql)` / `driver.execute(sql)`, and the shipped drivers hand knex's error straight back with the dialect text on the outer message, which is why every pre-existing case matched on the first read. This closes a dormant asymmetry before a wrapping raw-SQL driver can land on it — it is not a live user-facing bug.

It is also **not a regression from #6699**, which only made the contrast visible by deepening the first arm. Before that change the classifier took `message: string`, so the dialect arm read exactly the same single channel.

## What was measured before editing (PM assumptions)

Anchored on function names, not line numbers — this family's anchors have moved twice.

1. **Current address confirmed.** `classifyIndexFailure(error: unknown)` and the private `indexFailureText` are both in `packages/metadata-protocol/src/migrations/partial-index-probe.ts`, exactly as the issue describes. Premise still valid.
2. **No reusable helper exists — the walk stays local.** `@objectstack/types` exports `isUniqueViolationError` and `uniqueViolationColumn` only. Its two chain walkers (`matchesUniqueViolation`, `findUniqueViolationColumn`) are **private**, and each answers its own question rather than handing back text; there is no message-collecting helper to reuse. Hoisting one would widen that package's contract for a single consumer — and editing `@objectstack/types` is a declared STOP for this card — so the walk is local, pinned to the same bound, with the reasoning in the doc comment. This is PM assumption 2's fallback branch, taken deliberately.
3. **The bound is 4, and the issue's "four levels" is accurate.** `MAX_CAUSE_DEPTH = 4` in `packages/types/src/unique-violation.ts`, guard `depth > MAX_CAUSE_DEPTH`, top-level call at `depth 0` — so the thrown value plus four wrapper levels below it, five objects read. Mirrored exactly, counted the same way.
4. **Cycle safety — the predicate does NOT guard, and matching it means not guarding either.** `matchesUniqueViolation` keeps **no visited set**; the depth bound is its only cycle protection, and a bound terminates a cycle just as well. The new walk matches that behaviour rather than adding a visited-set the predicate lacks, and there is a test pinning that a self-referential `cause` returns a verdict instead of exhausting the stack. Recorded because PM assumption 4 asked for the real answer, not the expected one.
5. **The dialect regex is untouched** — only the text fed to it.

## One judgement call worth reviewing

The collected levels join with a **newline, never a space**. Two of the dialect alternatives are multi-word (`where clause`, `near "where"`), so a space would let a phrase be synthesised across a wrapper boundary that no single driver ever wrote — an outer message ending in `where` above a cause beginning with `clause` would read as a refusal. A newline cannot match the literal space in those alternatives. There is a test that goes red if the separator is ever changed to a space.

Level-0 behaviour is byte-identical to the old body for every previously reachable shape (string, object with a string `message`, object without one, `null` / `undefined` / number), so no existing verdict moves.

## Reverse verification

Predictions were written down **before** running, in `predictions.md`, then measured.

Method: `git checkout origin/main -- packages/metadata-protocol/src/migrations/partial-index-probe.ts` (fix out, new tests kept), re-run, restore. No `git stash` at any point.

**Predicted red on revert: the wrapped-refusal fixture, the depth-parity pin, the cyclic-chain refusal case, the end-to-end probe case, and the overlay-index fallback case. Measured: exactly those five test cases went red, no others.**

```
× grades a WRAPPED dialect refusal `unsupported`, not `failed` (#6848)
× reads the dialect arm to exactly the depth the conflict arm reads (#6848)
× terminates on a `cause` chain that loops, exactly as the predicate does (#6848)
× the probe classifies a wrapped refusal, and leaves `detail` the OUTER prose (#6848)
× a POOLED wrapper around the dialect refusal still reaches the fallback index (#6848)
 Tests  5 failed | 39 passed (44)
```

The parity pin's failure output is the clearest statement of the defect — the dialect arm reaching depth 0 only, against the conflict arm's 0..4:

```
AssertionError: expected [ true, false, false, false, …(3) ]
             to deeply equal [ true, true, true, true, true, …(2) ]
```

**Two predictions were recorded up front as green-in-both-directions, and are reported as guards rather than as evidence** — they constrain choices *inside* the new walk, which do not exist before the fix, so they cannot go red on revert:

- the newline-separator test (green before, vacuously — the cause is never read);
- the "arm order holds across the chain" test (green before, because arm 1 already walked). This one is the inversion-risk assertion: it would go red **after** the fix if widening arm 2 had stolen the verdict from arm 1. It stayed green, which is the direction that had to be checked.

**One prediction went unmeasured and is not being claimed.** The parity test's second assertion (the literal profile `[T,T,T,T,T,F,F]`) never executed on the revert run, because the parity equality above it failed first and short-circuited the test. It is measured green with the fix in; its predicted red is unverified.

## Files changed

- `packages/metadata-protocol/src/migrations/partial-index-probe.ts` — added `MAX_CAUSE_DEPTH` and the private `collectIndexFailureText`; `indexFailureText` now joins the chain. `classifyIndexFailure`'s **body is unchanged** — only its doc comment gained the "Why both arms walk `cause`" note.
- `packages/metadata-protocol/src/migrations/partial-index-probe.test.ts` — five new cases (wrapped refusal, arm order across the chain, depth parity, separator, cycle) plus the end-to-end probe case.
- **Deviation from the declared file surface, declared here:** `packages/metadata-protocol/src/migrations/overlay-index.test.ts` — one added test. Test-only, no source change. It pins the *consequence* the issue names (a pooled wrapper reaching `fallback: 'ensured'` rather than `not-attempted`); without it the PR would prove the verdict moved but not that the fallback index is actually reached again.
- `.changeset/index-failure-dialect-cause-walk.md` — patch for `@objectstack/metadata-protocol`. User-visible: `classifyIndexFailure` is a public export of the package.

Not touched, per the dispatch: `packages/metadata-protocol/src/protocol.ts`, `packages/spec`, `@objectstack/types`, `content/docs/releases/`.

## Verification

- `pnpm --filter @objectstack/metadata-protocol test` — **65 files, 820 tests, all passed** (814 pre-existing + 6 new).
- Consumption radius swept: `classifyIndexFailure` / `ensureOverlayStateIndex` / `PartialIndexStatus` have **no code consumers outside `metadata-protocol`** — the three hits in `packages/metadata` are prose in comments.
- `pnpm exec eslint` on all three changed files — clean, exit 0.
- `node scripts/check-nul-bytes.mjs` — OK, 6467 files. Plus a self-scan of the changed files for the wider control-byte range: no hits.
- `tsc --noEmit` on the package reports 159 errors, **0 of which name any file in this PR** — all are pre-existing debt in unrelated `seed-loader-*` / `sys-metadata-repository.*` tests, which is why this package carries no `typecheck` script and sits in the type-check-debt ledger. Flagged rather than silently omitted.

---

### DIFF
--- a/.changeset/index-failure-dialect-cause-walk.md
+++ b/.changeset/index-failure-dialect-cause-walk.md
@@ -1,0 +1,46 @@
+---
+"@objectstack/metadata-protocol": patch
+---
+
+fix(metadata-protocol): the dialect arm of `classifyIndexFailure` walks `cause` to the same depth the conflict arm does (#6848)
+
+`classifyIndexFailure` had two arms reading two different wrap-depths. #6699
+moved the first arm onto `@objectstack/types`' `isUniqueViolationError`, which
+follows `error.cause` four levels down because pool and query-builder layers
+re-throw with the original attached. The second — the dialect arm — kept reading
+`err.message` and stopping there.
+
+So a dialect refusal arriving behind a wrapper (outer prose `Write failed` or
+`pool query failed`, the actual `near "WHERE": syntax error` one step down
+`cause`) was graded `failed` instead of `unsupported`. The private
+`indexFailureText` helper now collects the message channel of the thrown value
+**and** of each `cause` below it, bounded at the same `MAX_CAUSE_DEPTH` of 4 the
+predicate uses and counted the same way (the thrown value is depth 0). The
+dialect vocabulary itself is unchanged — only the text fed to it.
+
+**Why the verdict matters beyond wording.** The two consumers dispose of
+`unsupported` and `failed` differently. `view-definition-active-index.ts` treats
+them the same (keep the previous index, report at `error`; only the wording
+differs). But `ensureOverlayStateIndex` builds the composite **fallback lookup
+index** on the `unsupported` branch and on no other — offered precisely because
+a dialect that cannot take the partial form should still get the lookup. Under
+a `failed` verdict that branch never ran, so `fallback` came back
+`not-attempted` rather than `ensured` / `refused` and the degradation target was
+silently never attempted.
+
+**Dormant, not a live regression.** No driver shipped today produces the wrapped
+shape — each hands knex's error back with the dialect text on the outer message,
+which is why every existing case matched on the first read. This closes an
+asymmetry before a wrapping raw-SQL driver can land on it; it is also not a
+regression from #6699, which only made the contrast visible by deepening the
+first arm.
+
+Two details worth knowing if you touch this: the collected levels are joined
+with a **newline**, never a space, because two of the dialect alternatives are
+multi-word (`where clause`, `near "where"`) and a space would let a phrase be
+synthesised across a wrapper boundary that no single driver wrote. And a looping
+`cause` chain is **bounded rather than detected** — no visited set — which is
+exactly what the predicate this mirrors does.
+
+Arm order is unchanged and still load-bearing: a conflict reported anywhere in
+the chain still beats a dialect refusal in the outer prose.
--- a/packages/metadata-protocol/src/migrations/overlay-index.test.ts
+++ b/packages/metadata-protocol/src/migrations/overlay-index.test.ts
@@ -303,6 +303,44 @@ describe('sys_metadata overlay uniqueness (#6418)', () => {
     });
 
     /**
+     * #6848 — the same MariaDB refusal, behind a pooled wrapper.
+     *
+     * This is the consequence the classifier's wrap-depth actually decides, and
+     * the reason the dialect arm was given the `cause` walk rather than a doc
+     * comment. The fallback lookup index is built on the `unsupported` branch
+     * and on no other, so while the second arm stopped at the outer message this
+     * case graded `failed` and the degradation target was never attempted — the
+     * dialect's answer was present, one step down, and unread.
+     */
+    it('a POOLED wrapper around the dialect refusal still reaches the fallback index (#6848)', async () => {
+        const logger = { info: vi.fn(), warn: vi.fn(), error: vi.fn() };
+        const pooled: IndexExec = async (sql: string) => {
+            if (/where/i.test(sql)) {
+                throw Object.assign(new Error('Write failed'), {
+                    cause: new Error(
+                        "You have an error in your SQL syntax; check the manual … near 'WHERE state = 'active''",
+                    ),
+                });
+            }
+            return db.exec(sql);
+        };
+
+        const result = await ensureOverlayStateIndex(pooled, 'active', logger);
+
+        expect(result.status).toBe('unsupported');
+        // The whole point: `not-attempted` here would mean the dialect that
+        // cannot take the partial form silently got no lookup index either.
+        expect(result.fallback).toBe('ensured');
+        // `detail` stays the operator-facing OUTER prose, unchanged.
+        expect(result.detail).toBe('Write failed');
+        // Nothing was downgraded — the declared UNIQUE index is byte-for-byte there.
+        expect(indexDdl(OVERLAY_INDEX_NAMES.active)).toEqual(DECLARED_ACTIVE_INDEX_DDL);
+        expect(insert('w1', 'view', 'lead.w', 'org1', 'pkg1', 'active').ok).toBe(true);
+        expect(insert('w2', 'view', 'lead.w', 'org1', 'pkg1', 'active').ok).toBe(false);
+        expect(String(logger.error.mock.calls[0]![0])).toContain('NOT enforced as specified on this dialect');
+    });
+
+    /**
      * MySQL proper, where the old code was safe only by ACCIDENT: it has
      * neither `DROP INDEX IF EXISTS` nor `CREATE INDEX IF NOT EXISTS`, so every
      * statement this migration issues is refused. Nothing may be touched, and
--- a/packages/metadata-protocol/src/migrations/partial-index-probe.test.ts
+++ b/packages/metadata-protocol/src/migrations/partial-index-probe.test.ts
@@ -248,6 +248,128 @@ describe('probe-first partial index replacement (#6418)', () => {
         expect(indexDdl(PROBE)).toBeUndefined();
     });
 
+    /* ────────────────────────────────────────────────────────────────────── *
+     *  #6848 — the DIALECT arm walks `cause` to the same depth as the first
+     * ────────────────────────────────────────────────────────────────────── */
+
+    /** A neutral wrapper: matches NEITHER vocabulary, so it can only carry. */
+    const wrap = (cause: unknown): Error => Object.assign(new Error('pool query failed'), { cause });
+
+    /** `leaf` behind `depth` neutral wrappers (`depth: 0` is the leaf itself). */
+    const nest = (depth: number, leaf: unknown): unknown =>
+        depth === 0 ? leaf : wrap(nest(depth - 1, leaf));
+
+    it('grades a WRAPPED dialect refusal `unsupported`, not `failed` (#6848)', () => {
+        // The shape a pooled or query-builder layer produces: useless outer
+        // prose, the dialect's actual answer one step down `cause`. Before this
+        // the second arm stopped at the outer message and returned `failed` —
+        // which costs `overlay-index` its fallback lookup index, because that
+        // branch is reached on `unsupported` and nowhere else.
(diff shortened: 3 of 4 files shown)


os-zhuang:

Verified against GitHub, not against the report:

- **File surface**: 4 files, all on the claimed surface — `partial-index-probe.ts`, its test, `overlay-index.test.ts` (the consumer whose fallback branch is the consequence), and the changeset. `classifyIndexFailure`'s body is untouched and the dialect regex is untouched; only the text fed to it changed, which is exactly the ruled scope. No contact with `protocol.ts`, where two sibling cards are in flight this round.
- **CI on `41d9415`**: 25 check runs, 0 failures — 23 `completed: success` including `ESLint` (the job carrying the `check:*` family) and `TypeScript Type Check`, plus Build Core, Test Core 1-3/3, Dogfood Regression Gate 1-3/3, Dogfood Verify CLI, Temporal Conformance, Check Changeset, Check PR Size; 2 `skipped` by the path filter (`Build Docs`, `Console Pin Gate`). Changeset present (patch).

**Three of my dispatch assumptions were re-measured and two came back against me** — which is what that section is for, and the write-ups are better than the assumptions were:

1. I guessed `@objectstack/types` might expose a reusable message collector to borrow. **Falsified**: its chain walkers are private and each answers its own question rather than handing back text. The walk stays local, and the doc comment says *why* — hoisting a collector there would widen that package's contract for a single consumer.
2. I asked for cycle-safety matching the predicate's. **The answer is the inverse of what I expected**: the predicate keeps no visited set — its depth bound *is* its cycle guard. So matching it meant deliberately **not** adding one, pinned by a self-referential cause returning a verdict instead of exhausting the stack, and by a two-node cycle whose refusal sits on the inner node still being found (the bound is reached after the answer, not before it).
3. Depth bound of 4 confirmed, counted the same way (thrown value is depth 0).

**Two design details worth naming, because neither was in the ruling and both are right:**

- **Levels join with a newline, never a space.** Two dialect alternatives are multi-word (`where clause`, `near "where"`), so a space would let a phrase be synthesised across a wrapper boundary no single driver wrote. Pinned by a fixture where neither message is a refusal alone (`…landed where` above `clause parsing completed`) and the joined result must still grade `failed`.
- **Arm ordering is pinned across the chain, not just at the top.** Widening the second arm creates an inversion risk — both arms now walk, so `conflict` has to keep beating a dialect refusal at every depth. There is a case for exactly that: dialect prose on the outer error, a `23505` two levels down, verdict must be `conflict`.

**The parity test is the right shape.** It asserts `dialectReach` equals `conflictReach` across depths 0-6 rather than pinning the number 4, plus a profile assertion so the equality cannot pass by both arms reaching nothing or everything. That survives a deliberate change to the shared bound and still goes red the moment the two arms drift apart again — which is the defect this card exists to close.

**Reverse verification, and a fourth honest category.** Predictions written before running; predicted 5 red, measured exactly those 5, none extra. Two assertions were recorded **up front** as green-in-both-directions and reported as guards rather than evidence — including the inversion-risk one, which would go red *after* the fix if widening arm 2 had stolen arm 1's verdict, and stayed green. And one prediction went **unmeasured** and is explicitly not claimed: the parity test's literal profile assertion never executed on the revert run because the equality above it failed first and short-circuited. Recording a prediction you did not get to measure — rather than quietly counting it as passed — is the standard I want in this lane.

**Caveat flagged rather than omitted, and correctly**: `tsc --noEmit` on the package reports 159 errors, **0 naming any file in this PR** — pre-existing debt in unrelated `seed-loader-*` / `sys-metadata-repository.*` tests, which is why the package carries no `typecheck` script and sits in the type-check-debt ledger.

Honest framing preserved throughout: no shipped driver produces the wrapped shape today, so this closes a dormant asymmetry rather than a live defect, and it is not a regression from #6699.

---

Issue #6943, opened 2026-08-09
bulkCreate / upsert 仍会逐次烧号:#5495 的重播种只落在 create(),批次语义是独立决策


os-zhuang:
## 一句话说明

#5495(PR #6932)让 `SqlDriver.create()` 在唯一冲突后**重播种陈旧的自增号计数器**并有界重试,于是"每次失败烧一个号"在该路径上不可达。`bulkCreate()` 与 `upsert()` **没有拿到这个修复** —— 两者同样调用 `fillAutoNumberFields`,碰撞后同样没有重播种,因此在同一个陈旧计数器上仍会逐次烧号并把请求打失败。

由 #5495 的 dev 在范围审计中记录,**刻意未顺手修**(PD #10)。

## 为什么不是 #5495 的一部分

不是遗漏,是裁向边界:批内逐行重试会改变**批次语义** —— 部分成功如何呈现、事务边界怎么划、失败行是否回滚整批。这些都不是 #5495 那条"单行 create 的陈旧计数器"裁向能顺带决定的,需要单独定夺。

## 前提(#5495 已实测,可直接继承)

- 计数器陈旧的成因:播种只在 `getNextSequenceValue` 的 `if (!existing)` 分支跑一次,此后不再读数据表;任何绕过 `fillAutoNumberFields` 落地的行(`isSystem` 种子回放、`preserveAudit` 导入、直接 SQL)都不抬序列。
- 因此 `bulkCreate` / `upsert` 在同一张表上会遇到与 `create` **完全相同**的陈旧,只是收口点不同。
- 判据不能用错误报文取冲突列:`uniqueViolationColumn()` 在带租户的自增号上恒为 `undefined`(复合键 → `soleColumn` 拒答;ADR-0120 D3 的表达式索引 → SQLite 报索引名 → `SQLITE_INDEX_FORM` 拒答;MySQL 无该 limb)。#6932 因此改用**数据探测**(`autoNumberValueExists`),该方案对批量路径同样适用。

## 关联

- #5495 / PR #6932 —— `create()` 侧的修复与全部实测前提
- #6806 —— engine 兜底路径的同形残余(engine-core 车道)
- #5503 —— strip 豁免,陈旧的来源之一

未认领,仅作记录,定级交分诊。

os-project-manager:
[...]

PR #7059, opened 2026-08-09
fix(driver-sql): a merge-path upsert no longer rewrites an existing row's autonumber (#7011)

### LINKED ISSUE #7011: upsert 每次「合并到既有行」都烧一个自增号,并覆写该行已有的业务号(健康计数器上实测,与陈旧无关)
## 一句话说明

`SqlDriver.upsert()` 在**成功路径**上无界烧号:每一次「合并到既有行」都会先取一个新的自增号,然后把它**写进那一行**,覆盖掉该行原有的业务标识符。与计数器陈旧**无关** —— 在完全健康的计数器上就能复现。

由 #6943(PR #6999)的范围审计中实测发现,**刻意未顺手修**(PD #10):修它需要一个语义裁决,不是那张卡的裁向。

## 实测(健康计数器,单行,全程只有 1 行)

```
create                        → CASE-00001     last_value 1
upsert 同一个 id(第一次)      → CASE-00002     last_value 2
upsert 同一个 id(第二次)      → CASE-00003     last_value 3
```

表里始终只有一行,而它的 `case_number` 被改写了两次。

## 成因

`fillAutoNumberFields` 在语句**知道自己会 insert 还是 merge 之前**就已经跑完了 —— 它发生在构建 `ON CONFLICT … DO UPDATE` 之前。而自增号列在 `mergeColumns` 里,所以合并分支会把这个新取的号一并写入。

两个后果,第二个更重:

1. **无界烧号**:每次 upsert 消耗一个号,即使一行也没新增。这与 #5495 / #6943 处理的「陈旧导致的一次性风暴」不是同一件事 —— 那些是有界的、每库一次的;这个是**每次调用都发生**,永不停止。
2. **改写业务标识符**:自增号通常是对外可见的单据号(`CASE-00001`)。一次以更新为目的的 upsert 会**静默更换它**,而调用方并没有要求改号。留空洞是一回事,改写已发出的标识符是另一回事。

## 需要裁决的问题(不预设答案)

这正是当初把 #6943 从 #5495 拆出来的那一类语义问题,请分诊/维护者定夺:

- **更新型 upsert 应当保留原有自增号吗?** 直觉上应当 —— 号是那一行的身份。
- **它应当在合并路径上取号吗?** 若不应当,那么取号必须推迟到「确定要 insert」之后,或者在合并分支上把该列从 `mergeColumns` 里排除。
- **既有数据怎么办?** 已经被改写过的行无法从驱动侧还原。

## 边界(与相邻卡的关系)

- **不是 #5495 / #6943 的缺陷**:那两张卡处理的是计数器落后于数据表导致的碰撞;本卡在计数器完全健康时就成立,且发生在**成功**路径而非失败路径。
- **PR #6999 已做的守卫**:它的重试循环不会把一次 merge 变成 insert。仅此而已,本卡的烧号与覆写照旧。
- `driver-memory` / `driver-mongodb` 不声明 `supports.autonumber`,走 engine 兜底路径(**#6806** 的面),与本卡无关。

## 关联

- #6943 / PR #6999 —— 发现处;批次与 upsert 的陈旧重播种
- #5495 / PR #6932 —— 自增号序列重播种,三态碰撞路由
- #1603 —— DB 原生自增号序列

未认领,仅作记录,定级交分诊。

os-project-manager:
Fixes #7011.

## The ruling this PR implements (triage, 2026-08-09, quoted per the dispatch)

> an autonumber is an immutable business identifier once assigned — the fix **excludes auto_number columns from `mergeColumns`** so a merge never rewrites an existing row's number; the pre-burn half (a number consumed before insert-vs-merge is known) belongs to #6943's reseed family and is explicitly out of scope here. If implementation finds any consumer relying on merge-overwrite, STOP and report.

## Premise re-verified on the branch base (`main` @ `3fc2e48`, healthy counter, one row)

The new pin file was run **before** the fix, and went red with exactly the filing's values:

```
create                      → CASE-00001    last_value 1
upsert same id (1st time)   → CASE-00002    last_value 2   (asserted CASE-00001, received CASE-00002)
upsert same id (2nd time)   → CASE-00003    last_value 3
```

One row throughout; its `case_number` was rewritten twice. Cause, as filed: `fillAutoNumberFields` reserves a number before the statement knows whether it will insert or merge, and the autonumber column sat in `mergeColumns`, so the `ON CONFLICT … DO UPDATE` branch wrote the fresh reservation over the row's existing number.

## The fix

`SqlDriver.upsert()` now derives its merge column list through a new `insertOnlyUpsertColumns(object)` helper: `created_at` (the pre-existing exclusion, byte-identical behavior) plus every `auto_number` column of the object, resolved under the same object-name/table-name lookup `fillAutoNumberFields` uses and mapped to **physical** column names (external objects can remap logical fields via `external.columnMap`, and the merge list is derived from the `applyWriteColumnMap`-processed row).

The exclusion is **unconditional**: an explicit autonumber value in the upsert payload does not renumber an existing row on the merge branch either — `update()` writes what it is given and remains the deliberate renumbering path. Insert-path upserts still assign fresh numbers, and every non-autonumber column (including `updated_at`) merges exactly as before.

## Consumer check (the STOP condition) — no consumer relies on merge-overwrite

Every driver-level `upsert` caller in the monorepo was enumerated:

- **Lifecycle Archiver** (`packages/objectql/src/lifecycle/lifecycle-service.ts:1110`, `cold.upsert(object, row, ['id'])`) — copies full hot rows carrying their autonumber values explicitly. The merge branch only re-fires when an interrupted sweep re-copies the **same unchanged row** (archived rows are hot-deleted after copy), so the excluded column's payload value equals the stored value and the exclusion is observably a no-op there.
- **`packages/runtime/src/sandbox/body-runner.ts:276`** guards on `typeof ql.upsert === 'function'` — `ObjectQLEngine` exposes no `upsert` method, so this path routes to `ql.insert` today.
- **REST import's `writeMode: 'upsert'`** is find-then-create-or-update at the engine level (`import-runner.ts` `findExisting`) and never calls driver `upsert`.

No fixture outside the three driver packages combines driver `upsert` with autonumber (swept by grep over every `@objectstack/driver-sql`-importing test).

## Faces

| face | conclusion | basis |
|:---|:---|:---|
| `SqlDriver` | fixed | primary; new pin file |
| `SqliteWasmDriver` | inherits (`upsert` not overridden) | new merge-path pin in its existing #6943 upsert test file |
| `TursoDriver` local/replica | overrides `upsert`, routes to `super` | new LOCAL merge-path pin in its existing #6943 test file |
| `TursoDriver` remote | neither defect nor fix | `RemoteTransport.upsert` never enters `fillAutoNumberFields` (#6999's table; unchanged) |

`driver-memory` / `driver-mongodb` are #5499-frozen and untouched — neither declares `supports.autonumber`; both use the engine fallback (#6806's surface), per the card's own boundary note.

## Out of scope, measured for #6943's card (the pre-burn reading the dispatch asked for)

Post-fix, on a healthy counter: the row keeps `CASE-00001` through both merge-upserts, but `last_value` still walks **1 → 2 → 3** — the reservation still happens before insert-vs-merge is known, so a merge-only upsert still consumes one sequence value per call. The next insert-path row gets `CASE-00004`, i.e. `CASE-00002`/`CASE-00003` are now **permanent gaps** rather than rewrites. Deferring the reservation is #6943-family work; no test in this PR asserts `last_value` on the merge path, so that future fix cannot turn this PR's pins red.

Existing rows already renumbered by past merges cannot be restored from the driver side (as the filing noted).

## Reverse verification (direction predicted before running)

Predicted: restoring the deleted limb (autonumber exclusion off) turns **exactly the merge-path pins** red, with the filing's own values; insert-path and non-autonumber-merge cases stay green. Measured: driver-sql pin file 4 failed / 1 passed (`expected 'CASE-00002' to be 'CASE-00001'`, `'TKT-0002'` vs `'TKT-0001'`), sqlite-wasm file 1 failed / 2 passed (only the new #7011 pin), turso file 1 failed / 3 passed (only the new LOCAL #7011 pin). Fix restored: 5/3/4 all green.

## Tests

- `pnpm --filter @objectstack/driver-sql test` — 80 files passed, 4 skipped (1147 tests passed, 48 skipped)
- `pnpm --filter @objectstack/driver-sqlite-wasm --filter @objectstack/driver-turso test` — 23 files / 308 tests and 32 files / 912 tests, all passed
- `tsc --noEmit` (typecheck) green on all three packages; ESLint clean on the four touched files
- Changeset: `@objectstack/driver-sql` patch (user-visible behavior fix)

https://claude.ai/code/session_01LGRN2cSRfggfX9B2L83bQc

---
_Generated by [Claude Code](https://claude.ai/code/session_01LGRN2cSRfggfX9B2L83bQc)_

### DIFF
--- a/.changeset/driver-sql-upsert-autonumber-immutable.md
+++ b/.changeset/driver-sql-upsert-autonumber-immutable.md
@@ -1,0 +1,41 @@
+---
+"@objectstack/driver-sql": patch
+---
+
+fix(driver-sql): a merge-path `upsert` no longer rewrites an existing row's autonumber (#7011)
+
+Measured on a completely healthy counter, single row throughout:
+
+```
+create                      → CASE-00001    last_value 1
+upsert same id (1st time)   → CASE-00002    last_value 2
+upsert same id (2nd time)   → CASE-00003    last_value 3
+```
+
+`fillAutoNumberFields` reserves a number before the statement knows whether it
+will insert or merge, and the autonumber column sat in `mergeColumns` — so
+every `ON CONFLICT … DO UPDATE` wrote the freshly reserved number over the
+row's existing one, silently replacing an externally visible business
+identifier the caller never asked to change.
+
+Per the triage ruling on the card: an autonumber is an **immutable business
+identifier once assigned**. `auto_number` columns are now excluded from the
+merge column list, exactly like `created_at` (both are insert-only facts about
+the row's birth). After the fix the same sequence keeps `CASE-00001` through
+both upserts. The exclusion is unconditional — an explicit autonumber value in
+the upsert payload does not renumber an existing row on the merge branch
+either; `update()` writes what it is given and remains the deliberate
+renumbering path. Insert-path upserts still assign fresh numbers, and every
+non-autonumber column (including `updated_at`) merges as before.
+
+Deliberately out of scope (#6943's reseed family): the reservation itself still
+happens before insert-vs-merge is known, so a merge-only upsert still consumes
+one sequence value per call — now a permanent gap in the sequence rather than a
+rewrite of the row (measured post-fix: row keeps `CASE-00001`, `last_value`
+walks 1 → 2 → 3, the next inserted row gets `CASE-00004`).
+
+Covered faces: `SqliteWasmDriver` inherits `upsert` unchanged; `TursoDriver`
+local/replica routes its override to `super` — both pinned by their own tests.
+Turso remote (`RemoteTransport.upsert`) never enters `fillAutoNumberFields` and
+has neither the defect nor the fix. Rows already renumbered by past merges
+cannot be restored from the driver side.
--- a/packages/drivers/driver-sql/src/sql-driver-upsert-autonumber-immutable.test.ts
+++ b/packages/drivers/driver-sql/src/sql-driver-upsert-autonumber-immutable.test.ts
@@ -1,0 +1,167 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#7011] A merge-path upsert must never rewrite an existing row's autonumber.
+ *
+ * # The defect, measured on a HEALTHY counter (no staleness involved)
+ *
+ * ```
+ * create                      → CASE-00001    last_value 1
+ * upsert same id (1st time)   → CASE-00002    last_value 2
+ * upsert same id (2nd time)   → CASE-00003    last_value 3
+ * ```
+ *
+ * One row throughout — and its `case_number` was rewritten twice. The cause:
+ * `fillAutoNumberFields` reserves a number before the statement knows whether
+ * it will insert or merge, and the autonumber column sat in `mergeColumns`, so
+ * the `ON CONFLICT … DO UPDATE` branch wrote the freshly reserved number over
+ * the row's existing one.
+ *
+ * # The ruling this file pins (triage, 2026-08-09, on the card)
+ *
+ * An autonumber is an **immutable business identifier once assigned** — it is
+ * usually an externally visible document number (`CASE-00001`), and a caller
+ * asking to update a row did not ask for a new one. The fix excludes
+ * `auto_number` columns from `mergeColumns`, exactly like `created_at` (both
+ * are insert-only facts about the row's birth). The exclusion is unconditional:
+ * even an EXPLICIT autonumber value in the upsert payload does not rewrite an
+ * existing row's number on the merge branch — `update()` writes whatever it is
+ * given and remains the deliberate renumbering path.
+ *
+ * # Out of scope here, deliberately
+ *
+ * The reservation itself still happens before insert-vs-merge is known, so a
+ * merge-only upsert still consumes a sequence value (leaves a gap). That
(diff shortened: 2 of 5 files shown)


os-project-manager:
**PM step-7 acceptance review — PASS** (session_01LGRN2cSRfggfX9B2L83bQc, maintainer-expedited v17 card).

Verified independently against head `e6b0c43`: (1) premise reproduced red-first with the filing's exact values before the fix; (2) ruled scope held exactly — autonumber columns excluded from `mergeColumns` via a helper that reuses `fillAutoNumberFields`' own object/table lookup and maps to physical columns, pre-burn half untouched and its post-fix reading recorded for #6943; (3) the STOP condition was discharged by enumeration, not assumption (Archiver re-merges identical rows; body-runner's upsert branch is dead; REST import never calls driver upsert); (4) reverse verification predicted-then-measured with exact hit (4/1, 1/2, 1/3 red splits); (5) CI 25/25 zero failures, independently read — including `ADR maintainer approval` success on a non-ADR PR, confirming the newly-required context reports everywhere; (6) five files, zero riders, changeset patch present. Marking ready and queueing.

---

Issue #6298, opened 2026-08-07
[PM seat] domain:spec-surface — 🔀 merged into #6017


qq9340100:
**This card is the single authoritative registry for the `domain:spec-surface` seat.** Seat-card protocol: index = `label:pm:seat`, static pointer page #4604. Seat split from `domain:spec` per maintainer approval 2026-08-07. Single-writer: only the incumbent edits this body; takeover/handoff = body edit + one audit comment (comments are archive, not state). Liveness is lazy (no heartbeat), evaluated only on takeover conflict: &gt;24h without output is reclaimable.

## Scope

The **text face** of `packages/spec`: describe / JSDoc / tombstone-migration prose / error guidance and alias tables — the lines where the contract explains itself.

**Admission criterion = validation invariance**: metadata legal before the change stays byte-identically legal after; illegal stays illegal; zero client migration; changeset always patch (or `skip-changeset` when the prose reaches no consumer — E13).

Red lines:
- ⛔ Any change to the acceptance face, however small the diff, belongs to `domain:spec` (precedents #6245/#6235).
- ⛔ Gate/generator/lint machinery and prose belong to `domain:spec-tooling` (#5163 boundary); hand-written docs pages teaching a drifted contract face stay surface per #6146 (#6413 triage confirmation).
- ⛔ `content/docs/releases/` is never touched (repo-wide rule).

### Render-input criterion (which prose reaches generated docs)

[...]

Issue #4636, opened 2026-08-02
loadMetaFromDb object branch reads `record.packageId` from a snake_case row — always undefined, every object overlay registers under the 'sys_metadata' sentinel at boot


os-zhuang:
Found while implementing #4624 (PR #4635). **Unassigned — nobody is on this.** Recording only, per Prime Directive #10.

## Evidence

`loadMetaFromDb`'s **object** branch (`packages/metadata-protocol/src/protocol.ts`, ~line 8958 on current `main`) registers each hydrated object row with:

```ts
this.engine.registry.registerObject(
    { ...(data as Record<string, unknown>), _provenance: 'org' } as any,
    record.packageId || 'sys_metadata',
);
```

But `engine.find('sys_metadata', ...)` returns rows keyed by the object's **snake_case** field names — `package_id`, not `packageId`:

- the `sys_metadata` object declares `package_id` (`packages/metadata-core/src/objects/sys-metadata.object.ts`);
- `SysMetadataRepository` writes `parentRowData.package_id = ...` (`sys-metadata-repository.ts` ~386-388) and maps `row.package_id → packageId` explicitly when it wants camelCase (~784);
- `getMetaItems` in the same file reads the same query's rows as `r.package_id` (~2588, ~2606).

So `record.packageId` is **always `undefined`** and the `|| 'sys_metadata'` fallback always wins: every object overlay row — including rows genuinely bound to a package (`package_id = 'app.<slug>'` etc.) — registers into the SchemaRegistry under the `'sys_metadata'` sentinel `sourcePackage` at boot, silently dropping its package binding.

## Why it matters

The cloud#970 fix comment right above this line documents the intent: pass the row's real package id but stamp `_provenance: 'org'` so the tenant-authored row is not misread as code-shipped. Half of that intent is dead: the provenance stamp works, the package binding never arrives. Consequences of registering under the sentinel instead of the real package id include the sidebar package filter / provenance classification not seeing the boot-hydrated object as belonging to its package (the same concern the read-side hydration explicitly handles by surfacing `record.package_id`, ~2596-2609), and `registry.getObject`-side package bookkeeping diverging between "created this session" (write path, real id) and "restarted" (boot path, sentinel).

## Suggested fix

[...]

PR #7093, opened 2026-08-09
fix(metadata-protocol): tier 3 refuses to unregister an object bound to an installed package (#7012)

### LINKED ISSUE #7012: Stop-the-bleed: tier-3 heal must not unregister a package-bound object — decide on the contributor's package binding, not isArtifactBacked (#6853 ruling C)

os-zhuang:
Fixes #7012

Ships **option C** of the maintainer ruling of 2026-08-09 on #6853 — the stop-the-bleed half, ahead of direction B's ADR work. Nothing here re-litigates the accepted cost, and nothing here implements B.

## The outage, reproduced before writing a line of the fix

Measured on this branch's base (`2f3e79351`) with a throwaway probe on the `protocol-delete-object-registry-heal.test.ts` harness shape — real `SchemaRegistry` + real `SysMetadataRepository` over an in-memory engine — on a **tenant kernel, no escape hatch, no operator action**:

```json
{"loadMetaFromDb":{"loaded":1,"errors":0,"invalid":0,"storeUnavailable":false},
 "afterHydration_isArtifactBacked":false,
 "insertBeforeDelete_ok":true,
 "deleteReceipt":{"success":true,"reset":true,
                  "message":"Deleted object 'myapp_invoice' — it no longer exists. [seq=1]"},
 "storedRowsAfter":0,
 "objectContributorsAfter":0,
 "getObjectAfter":null,
 "dataCrudAfter":{"code":"OBJECT_NOT_FOUND","status":404,"object":"myapp_invoice"},
 "dataRows":1}
```

`dataRows: 1` is the point — the table still holds the row the object can no longer reach.

The probe needs no hatch because it uses the real boot order across two sessions: session 1 persists the overlay while nothing ships the name (so the two-tier gate has nothing to refuse), session 2 installs the package, registers the packaged object, then hydrates. That is #6853 P6 exactly.

## Mechanism

`SchemaRegistry.registerObject` **splices out** the same-package `own` contributor before pushing the new one, so an overlay row whose `package_id` equals the packaged owner's id **destroys** the packaged definition at write time — it is not shadowed, and no second copy exists anywhere in the registry. `loadMetaFromDb` replays that replacement on every boot with no authorization gate and no log, stamping `_provenance: 'org'` server-side (deliberately, cloud#970).

`getArtifactItem` for an `object` is exactly `_provenance !== 'org'`, so `isArtifactBacked` — the predicate tier 3's "AND IT NEVER RETIRES A CODE-SHIPPED OBJECT" guard reads — is **the one predicate the overwrite falsified**, and it answers `false` for an object the package still ships.

## The predicate chosen, and why

Tier 3 now also refuses when the owner contributor's package binding names a **currently-installed** package:

```
registry.getObjectOwner(name)?.packageId   ->  registry.getPackage(packageId)
```

- The **binding** is trustworthy where the definition is not: the replacement rule fires only when the two package ids **match**, so the surviving contributor provably carries the packaged owner's id — the one fact the overwrite cannot change, because it is the overwrite's own precondition.
- **"Installed" is a fact about the process, not about the destroyed body.** `ObjectQL.registerApp` calls `SchemaRegistry.installPackage(manifest)` immediately before registering the manifest's objects (`packages/objectql/src/engine.ts`), durable `sys_packages` records are re-installed at boot by `service-package`, and `registerPlugin` objects are keyed to the **parent** package, which `registerApp` installed. So a package-shipped object always has a record.
- The `'sys_metadata'` sentinel is handled **by construction**, not by a special case: nothing installs a package under it.

Two things the predicate deliberately does **not** ask, each pinned by a test:

- **not `enabled` / `status`** — `disablePackage` flips lifecycle flags and removes no contributor, so a disabled package's objects stay registered and stay dispatchable; reading the flag would unregister a definition nothing else removes, which is the same outage through a second door.
- **not the manifest's `objects` list** — that re-asks "is this code-shipped", the question whose answer was destroyed, and it is absent for `registerPlugin`-contributed objects.

Note on the dispatch's assumption 2: `registry.getPackage(pkgId)` returning `null` in #6853's measurement was a fact about **that probe** (it called `registerObject` directly, installing nothing), not about installed packages. The two questions are different and the installed-package one is answerable here.

The refusal is **stated in the log**, not swallowed — the same discipline the ADR-0029 extender refusal one line up already carries: a runtime that disagrees with `sys_metadata` must be visible rather than inferred later.

## Accepted cost (ruled — recorded, not re-argued)

A package-bound **runtime-authored** object (Studio package workspace, #4636) carries a real `package_id` too, so it is indistinguishable from a package-shipped one by binding alone: some genuinely deleted objects stay registered until restart — listable, and rowless. Per this walk's own REGISTER WIDE / RETIRE NARROW argument that is the cheap direction; a surplus entry degrades to "listable but rowless" and the next reload heals it, a wrongly retired one 404s data CRUD for every tenant. The honest fix for the distinguishability itself is #6853's direction B.

## File surface — every file, and the declared deviations

| File | Why |
|---|---|
| `packages/metadata-protocol/src/protocol.ts` | tier 3 of `restoreArtifactRegistryView` (the decision), plus one new **private** helper `installedPackageBindingForObject` carrying the argument |
| `packages/objectql/src/protocol-delete-object-package-binding-guard.test.ts` | the regression pin, both directions |
| `.changeset/lucky-buttons-shave.md` | user-visible fix |

**Deviations from the declared surface, stated file by file:**

1. The helper is a **second symbol** in `protocol.ts` rather than an inline expression in tier 3. It reads no new state and changes no other call site; it exists so the argument (why `isArtifactBacked` cannot be trusted at this point, what the predicate does and does not ask, and the ruled cost) lives next to the code instead of in a commit message. The decision itself is still one branch at one point.
2. **No new registry plumbing.** `getObjectOwner` and `getPackage` both already exist and `getPackage` is already declared in the engine contract, so `packages/spec` is untouched and no cross-seat declaration on #6298 is needed.
3. `packages/objectql` gains a test file only — no source change.

Untouched, as required: `saveMetaItem`'s two-tier write gate region (#6973 / `session_01LGRN2cSRfggfX9B2L83bQc`), `assertSortFieldsExist`, `packages/spec`, `content/docs/releases/`. Tiers 1 and 2 are unchanged — verified rather than assumed: both conclude a lower layer still serves the name, so neither ever reaches a removal, and the damage is entirely tier 3 deciding *to remove*.

Predicted **RED**, and only on the positive-direction cases. Fix taken out with `git checkout 2f3e79351 -- packages/metadata-protocol/src/protocol.ts` (this branch's own base, never a moving `origin/main`), `metadata-protocol` rebuilt, pins re-run.

| Case | Predicted | Measured |
|---|---|---|
| package-bound object survives the delete | red | **red** — `expected undefined to be defined` at `:308` |
| DISABLED installed package still protects | red | **red** at `:412` |
| plural `objects` spelling guarded | red | **red** at `:432` |
| control-plane kernel guarded | red | **red** at `:452` |
| boot replay destroys the definition, binding survives | green both ways | **green** — a guard on the premise, not evidence |
| package-LESS runtime object still retired | green both ways | **green** — a guard against over-refusal |
| binding naming a NON-installed package still retired | green both ways | **green** — a guard; it is also why every pre-existing pin in `protocol-delete-object-registry-heal.test.ts` stays green |

`Tests 4 failed | 3 passed (7)` with the fix out; `Tests 7 passed (7)` with it in.

**One prediction went unmeasured in the reverse direction and is not counted as passed:** the "survives" case fails at its first post-delete assertion, so its data-plane recovery and warning-text assertions never execute with the fix out. They are measured green in the fixed direction, and the pre-fix data-plane verdict was measured separately by the throwaway probe quoted above (`OBJECT_NOT_FOUND` / 404) rather than inferred.

## Verification

- `pnpm --filter @objectstack/metadata-protocol test` — `Test Files 65 passed (65)`, `Tests 827 passed (827)`
- `pnpm --filter @objectstack/objectql test` — `Test Files 163 passed (163)`, `Tests 2793 passed (2793)`
- `pnpm --filter @objectstack/runtime test` — `Test Files 117 passed (117)`, `Tests 1761 passed (1761)` (run as the downstream consumer of the heal; its first run's 91 collect failures were the fresh-worktree stale-artefact trap, gone after `--filter '@objectstack/runtime^...' build`)
- `pnpm --filter @objectstack/objectql typecheck` — `Done`
- `metadata-protocol` has no `typecheck` script (measured DEBT entry of 63): `tsc --noEmit` counts **63**, exactly the ledger's number, so `check:type-check-debt` is unmoved
- gates: `check:engine-double-contract`, `check:error-code-casing`, `check:route-envelope`, `check:meta-type-normalized`, `check:empty-changeset`, `check:adr-anchors`, `check:wildcard-fallthrough` — all PASS; `node scripts/check-nul-bytes.mjs` OK; `eslint --no-inline-config` clean on both changed files

---

### DIFF
--- a/.changeset/lucky-buttons-shave.md
+++ b/.changeset/lucky-buttons-shave.md
@@ -1,0 +1,13 @@
+---
+'@objectstack/metadata-protocol': patch
+---
+
+fix(metadata-protocol): the delete heal no longer unregisters an object bound to an installed package
+
+Deleting a metadata overlay row for an `object` whose `package_id` names an installed package took the object off the whole data plane until the next restart: every CRUD call answered `OBJECT_NOT_FOUND` / 404 while the table still held the rows, and the delete receipt said `reset: true`.
+
+`SchemaRegistry.registerObject` replaces (splices out) the same-package `own` contributor rather than shadowing it, so hydrating such a row destroys the packaged definition at write time and stamps `_provenance: 'org'`. Tier 3 of `restoreArtifactRegistryView` then consulted `isArtifactBacked` — which for an `object` is exactly that provenance — and read "not code-shipped" for an object the package still ships.
+
+Tier 3 now also refuses when the owner contributor's package binding names a currently-installed package, and says so in the log. The binding survives the overwrite (the replacement fires only when the package ids match); the definition does not.
+
+Known cost, deliberate: a package-bound runtime-authored object is indistinguishable from a package-shipped one by binding alone, so a genuinely deleted one stays registered until the next restart — listable, and rowless. A surplus entry is the cheap error here; a wrongly retired one 404s data CRUD for every tenant.
--- a/packages/metadata-protocol/src/protocol.ts
+++ b/packages/metadata-protocol/src/protocol.ts
@@ -8240,7 +8240,8 @@ export class ObjectStackProtocolImplementation implements
             // which for `object` reads the contributor definition and applies
             // exactly the artifact test the sibling verb applies to the plain
             // key), so this limb inherits that judgement instead of open-coding
-            // a second one.
+            // a second one — PLUS the package-binding check below, which exists
+            // because that inherited judgement is measurably falsifiable here.
             //
             // Not theoretical, and NOT already covered by the gate at the top of
             // `deleteMetaItem`: that two-tier authorization — which refuses an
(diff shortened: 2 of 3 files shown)


os-zhuang:

Verified against GitHub:

- **CI on `365c97e7`**: 25 check runs, 0 failures — 23 `completed: success` including `ESLint` (the job carrying the family gates) and `TypeScript Type Check`, plus Build Core, Test Core 1-3/3, Dogfood Regression Gate 1-3/3, Dogfood Verify CLI, Temporal Conformance, Check Changeset, Check PR Size; 2 `skipped` by the path filter (`Build Docs`, `Console Pin Gate`).
- **File surface** on the declared region only — tier 3 of `restoreArtifactRegistryView` plus the committed regression pin and a changeset. `saveMetaItem`'s gate region (PR #6973, being rebased by another session) and the ADR-0029 design work (PR #7087) are untouched.
- `metadata-protocol`'s `tsc --noEmit` counts **exactly 63** — the ledger's recorded DEBT, unmoved.

**The outage was reproduced before the fix was written**, on the branch's own base, and the numbers match #6853's measurement including the part that makes it a real outage rather than a stale read: `dataCrudAfter → OBJECT_NOT_FOUND/404` while `dataRows: 1`. The table still holds the data; only the registry entry is gone.

**Why the chosen predicate is sound rather than merely plausible** — this is the part I want on the record. The guard trusts the owner contributor's **package binding** where it cannot trust the definition, and there is a proof rather than a heuristic behind that: `registerObject`'s splice fires *only* when the two package ids match, so the surviving contributor **provably** carries the packaged owner's id. That is exactly why `isArtifactBacked` fails here and the binding does not — the overwrite destroys the definition and preserves the binding, by construction.

Two further things done right:

- **The divergence is stated in the log, not swallowed.** A guard that silently declines to retire would trade one invisible behaviour for another; this one says what it saw.
- **Dispatch assumption 3 was verified, not assumed.** I said "the other two tiers probably need no change" — the dev measured that tiers 1 and 2 return before any removal, so neither has a removal branch to change. Both other assumptions were confirmed by measurement too (no new plumbing; `getPackage` is the right predicate and is already declared in the engine contract).

**Reverse verification, including the category most reports quietly drop.** Direction predicted RED first; the fix was removed with `git checkout 2f3e79351 -- …` — **the branch's own base, not a moving `origin/main`** — which is the method note this lane picked up from #6973 and is now paying off. Measured `4 failed | 3 passed (7)`, the four reds being exactly the positive-direction cases. The three greens are green in **both** directions and are reported as **guards, not evidence**. And one prediction went **unmeasured** in the reverse run and is explicitly not counted as passed: the "survives" case short-circuits at its first post-delete assertion, so its data-plane-recovery and warning-text assertions never execute with the fix out — with the pre-fix 404 measured separately by the throwaway probe instead. Saying that plainly, rather than banking a prediction the run never reached, is the standard this lane holds.

The ruled cost stands untouched and un-relitigated: a package-bound runtime-authored object (#4636's Studio package workspace) is indistinguishable from a package-shipped one by binding alone, so some genuinely deleted objects stay listable-but-rowless until restart. That is the cheap direction per REGISTER WIDE / RETIRE NARROW, and the honest fix for the distinguishability itself is #6853's direction B — now proposed as ADR-0029's D9 in PR #7087, awaiting the maintainer.

---

PR #7101, opened 2026-08-09
fix(spec,objectql,sharing,storage): state per-row vs record dispatch on the hook contract (#6966)

### LINKED ISSUE #6966: #5574's per-row `before*` dispatch binds `input.id` on every context, silently changing the semantics of every "no id ⇒ skip, this is a bulk write" hook guard

os-zhuang:
Closes #6966.

A predicate (`multi: true`) write dispatches its lifecycle hooks **once per matched row** — `after*` since #5038, `before*` since #5574 (ADR-0058 Addendum II) — on a context deliberately indistinguishable from a single-id write's, so a handler written for one record works unchanged on a batch.

That indistinguishability is the feature. It also erased the only signal several handlers had: before #5574 a bulk `before*` fired **once** with `input.id` present-but-`undefined`, so "`input.id` is empty" meant "this call stands for N rows". Every guard written on it **silently inverted** rather than failing — a per-row context has an id, so the guard now answers "single write" for every row of a batch.

## The marker, and how its shape was measured

```ts
ctx.dispatch // { mode: 'record' | 'per-row', index: number, scope: object } | undefined
```

Bound by the engine at every write dispatch site (insert, update, delete, both phases), at the point the dispatch ladder is decided. It has to come from the engine: the verdict is `isByIdWrite` / `isPredicatePath`, which also consults driver capability (`updateMany`/`deleteMany` presence), and `asScalarId` is deliberately unexported to stop the plugin side re-deriving it from `options.multi` (#4434 / #4550).

Each member earned its place against a consumer, not against taste:

| Member | Why it is there | What excluded the alternative |
|---|---|---|
| `mode` | The question itself | A presence-encoded marker (`rowIndex?: number`, absent ⇒ record) is indistinguishable from a hand-built context that carries no marker at all — and `HookContextSchema` is deliberately non-strict, so absent is a shape the parse accepts. Measured on the drivers in `file-reference-lifecycle.test.ts`, which build contexts by hand. |
| `index` | `index === 0` is how a handler does batch-scoped work once | With `mode` alone the two live consumers cannot express "once per write". Measured: without it, `plugin-sharing`'s bounded `afterUpdate` branch is **quadratic** — `affectedFrom(ctx)` returns the whole union and is then recomputed once per row (1000 rows ⇒ 10⁶ `recomputeRow` calls), and the unbounded branch issues 1001 object-wide revokes where one is correct. Both are pinned. |
| `scope` | Carries state from `before*` to `after*` | The two consumers already stash across the pair; that only ever worked because a single-id write reuses one `HookContext`. Probed on the real engine before designing: a value set on a per-row `beforeUpdate`/`beforeDelete` context reads back `undefined` on every `after*` context. |
| ~~`size`~~ | **Dropped** | No consumer reads it. Shipping it would be a declared-never-read member, which is the ADR-0049 shape this card is about. |

Optional, and an absent marker reads as "not a per-row dispatch" — the back-compatible direction. Reads carry no marker: a read has no fan-out. The claim that *every dispatched* context carries one is pinned as an invariant across all six write events, not just path by path — which is what makes `ctx.dispatch?.mode` safe to read with no "did the engine bind it" branch, and which covers the one context a reader might worry about (`update()`/`delete()` keep a batch-scoped `hookContext` on the predicate path; no handler ever sees it).

**This is not the discriminator #5574 retired.** `isPredicateBulkWrite` was removed under ADR-0049 for having neither a producer nor a reachable consumer: it inferred "bulk" at the *consumer* from `input.id` + `options.multi`, and answered `false` everywhere once every dispatch carried an id. This one is produced where the verdict is made and has readers. `hook-wrappers.ts`'s retirement note stands unchanged; the contrast is written into the new JSDoc so a future reader does not mistake one for the other.

## Behaviour fixed

**`plugin-sharing` — the card called this site "comments only". It is not.**

`stashAffectedRows` parks the write's affected row set for the `after*` half. On a predicate write that stash was landing on a per-row context the `after` phase never sees, so `readAffectedRows` answered `unbounded/resolve-failed` and **both** subscribers (rule recompute, record-share cascade) took their safe branch: every bulk update or delete on a ruled object revoked **all** of that object's rule grants and queued a full asynchronous re-grant — once per matched row, with the repeats racing each other's re-grants.

Access is never widened (this is the ruling's "over-granting is an incident, under-granting is a wobble" direction), so it is not a fail-open — but it is a large silent behaviour change, and #4779's own tests could not see it because they drive a simulator modelling the **pre-#5574** batch dispatch. A bounded write now takes the bounded path again: rows are unioned as the engine hands them over (no predicate re-query — `resolveAffectedRows` short-circuits on a bound `input.id`, so "resolve once and reuse" would have frozen the batch to row 0's id and under-revoked the rest, which *is* the fail-open direction), the cap still applies to the union, and the `after*` work runs once per write.

**`service-storage` (site 1)** — the `beforeDelete` id pre-resolution was dead on every path, and its stash could not have reached the `after` phase anyway. The pre-resolution query is now gone entirely: the engine has already matched the doomed rows, so `beforeDelete` collects what it is handed and `afterDelete` releases the batch in **one** `sys_file` lookup over an `$in`. `beforeUpdate` copy-on-claim no longer runs once per row against a batch-scoped payload, which removes a row-conditioned rewrite of a shared `SET` clause (out of contract under ADR-0058 Addendum II D3).

**Stale comments (site 3)** corrected where they sit, in both packages.

## Where I read the card differently

- **The card's cost claim for site 1 is measured against a pre-#5038 world.** It says the guard "used to cost one `engine.find` per batch". Re-enabling the old hook *today* would be **worse** than leaving it dead: per-row `afterDelete` contexts are spread copies of the batch context, so all N would see the full stashed id list and each release the whole set — N lookups over N ids, plus the extra `engine.find(object, { where })`. The fix therefore is not "restore the guard" but "collect what the engine already matched, release once".
- **Site 3 is not text-only** (above).
- **`afterUpdate` in `file-reference-lifecycle.ts` is deliberately left running per row.** Its guard reads `ids.length !== 1`, which no longer means what its comment said — but the reconciliation it now performs is the *safer* outcome, and restoring the skip would leave a copied file owned by nobody while N records still reference it, which the sweep may collect. The genuine defect there (a predicate update gives N records one file id under an exclusive-ownership model) is older and wider than this card and belongs to service-storage — filed as #7102, and referenced from the comment.
- **Gate count**: `lint.yml` has 48 `pnpm check:` gates as the dispatch said; 52 across all workflow files.

## Verification

Reverse-verified: pre-fix sources restored with the new tests kept, predictions recorded first, then run.

| Suite | Predicted | Actual |
|---|---|---|
| `objectql` | 8 red (5 marker + 3 scope), rest green | 8 failed / 49 passed ✔ |
| `service-storage` | 1 red, on the batching assertion only | 1 failed / 39 passed ✔ |
| `plugin-sharing` | 4 red | 4 failed / 24 passed ✔ |

One deviation, and it strengthens the case: for "grants every matched row", I predicted the unbounded branch would leave the grant set **empty**; it actually left `['opp0','opp1']` of three. The residue comes from the queued async re-grants interleaving with the *repeated* per-row revokes — so the pre-fix behaviour is not merely wasteful, it is racy within a single write.

Green after the fix: `objectql` 2786 (+58 in the touched file), `plugin-sharing` 414, `service-storage` 324, `spec` 9247.

Gates run locally, all clean: `check:empty-changeset`, `check:changeset-gate-self-tests`, `check:kernel-hook-pairs`, `check:engine-double-contract`, `check:required-contexts`, `check:type-check-coverage`, `check:type-check-debt` (needs the built closure first), `check:published-files`, `check:query-options-erasure`, `check:startup-registry-verdict`, `check:strictness-ledger`, `check:docs`, plus `pnpm lint` and `pnpm typecheck` (126/126 tasks).

Three generated baselines move with the schema key and are committed: `packages/spec/authorable-surface/data.json`, `docs/audits/2026-07-unknown-key-strictness-ledger.counts.md`, and `content/docs/references/data/hook.mdx` (the last was caught by CI's `check:docs`, not locally — my local sweep had missed it).

Under the full parallel `pnpm test`, `@objectstack/types` (`node.test.ts` host-app resolution) and four `@objectstack/lint` lazy-module-loading tests flake; each passes in isolation and via turbo (`lint`: 68 files / 1771 tests green), and neither package is touched by this diff.

Refs: #6966, #5574, #5038, #6656, #4434, #4550, #4779, #7102, ADR-0049, ADR-0058 Addendum II.


os-zhuang:
**Queue steward — kicked from the merge queue at 19:13Z. NOT re-queued: this is a new signature.**

Read from the **complete log archive**, not the tail (SKILL note 7).

## Signature

- **Group**: `gh-readonly-queue/main/pr-7101-445a0c2866ab8deac7ec30b531753aa045e67e5d`, created **19:13:40Z**
- **Run**: [31331084751](https://github.com/objectstack-ai/objectstack/actions/runs/31331084751) — workflow `Lint &amp; Type Check`, job **`TypeScript Type Check`** → `failure`
- **Failing gate**: `pnpm --filter @objectstack/spec check:export-origins` (`tsx scripts/build-export-origins.ts --self-test &amp;&amp; tsx scripts/build-export-origins.ts --check`)
- **Error, verbatim**:

```
✅  self-test: re-exports share one origin, distinct declarations do not, aliases resolve to the declaration.
❌  1 problem(s) with export-origins/:
    • export-origins/data.json is stale — the source resolves differently
```

```
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @objectstack/spec@17.0.0-rc.5 check:export-origins
Exit status 1
##[error]Process completed with exit code 1.
```

The self-test passed and the gate's own `--check` half is what failed. Sibling gates in the same group: `ADR Merge Approval` / `Spec Liveness Check` / `Console Pin Freshness` `success`; `CI` was still running when the group was torn down.

## Ledger verdict

No row in any of the four signature tables on #5810 matches ⇒ **new signature** ⇒ ⛔ this PR was **not** re-queued as-is. The queue currently holds zero entries; this PR is out of it and `main` is at `445a0c28`.

## Preliminary reading (a reading, not a ruling — the diagnosis belongs to the lane)

This has the shape of the "入队与落地 A" generated-artifact hazard, not a flake. Four readings:

1. **`Lint &amp; Type Check` was `success` on this PR's own head** `404f69a2` at 16:18:12Z. It only fails against the queue's merged base — so nothing on the branch is independently broken.
2. `packages/spec/export-origins/**` is routed **`merge=os-regen`** in `.gitattributes` (read live, not copied). That driver merges exit-0 with zero conflict markers and can silently drop one side; only regeneration exposes it.
3. **The artifact did not exist on this PR's base.** `packages/spec/export-origins/` was introduced by **#7090** (`f5a9bc2f`, *"export-surface pins compare a build-time baseline instead of running tsc (#4796)"*) at 16:06Z. This PR's base is `0bffdae`, cut before that. Between the two bases: `git diff --stat 0bffdae 445a0c28 -- packages/spec/export-origins/` = **16 files, +5095**, moved by #7090 / #7103 / #7123 / #7116 / #7110 / #7126.
4. This PR **adds public spec exports** (the `ctx.dispatch` marker and its types). So the merged tree carries main's baseline, generated from a source that does not contain this PR's exports — which is exactly what *"the source resolves differently"* reports.

## Recommended action

⛔ **A re-run will not help.** A re-run reuses the original merge ref (SKILL note 5); the regeneration is missing from that ref, so it will fail identically. This needs a new commit.

Four-step per "入队与落地 A":

1. `git merge origin/main` (⛔ no rebase, no force-push)
2. `git checkout origin/main --` the os-regen paths — read the list live with `grep os-regen .gitattributes`, ⛔ don't work from a copied list
3. **commit the merge first, then regenerate wholesale** (⛔ never run `gen:schema` while in MERGE state — #5370). For this gate: `pnpm --filter @objectstack/spec gen:export-origins`, then **read the diff**: a name whose origin moved file is a re-homed declaration; a name that gains an entry with a *different* origin is the #4411 dual-source trap, and the fix is at the declaration (`check:dual-source-exports`), not in the artifact
4. assert the sibling PRs' entries are still present, then push

## Scope and yield

⛔ Steward authority is landing-only: no merge, no ready/draft flip, no re-queue, no code, no claim changes. **Re-queueing after the fix is the lane PM's call.**

**让行 check**: no lane comment on this PR in the last 30 minutes (latest comments are the Vercel and docs-drift bots, 15:51Z / 16:18Z) ⇒ no yield; no double-handling risk.

---

os-zhuang:
Dequeued on `CI_FAILURE`, now fixed — **all 26 checks green on `cb1b66f`**, `mergeable_state: clean`, branch level with `main`. It was not re-queued automatically, so it needs a re-queue from whoever owns that call; I have not enabled auto-merge on my own initiative.

The failure was mine, not a batch-mate's (the queue ref was `pr-7101` alone). Two things had drifted since this branch's base, and the second is worth a reviewer's attention because it touched **another PR's** work:

**1. `check:export-origins` is a new gate.** It landed on `main` in #7090 after this branch was cut, so `export-origins/data.json` had never seen `HookDispatch`. Regenerated; the diff is one line, origin `src/data/hook.zod.ts` — the same file `HookContext` is declared in, so not a re-home and not a new dual-source (the two readings its gate asks for).

**2. Merging `main` silently dropped five exports that #7123 had added.** `packages/spec/api-surface/**` carries `merge=os-regen` in `.gitattributes` precisely so a merge regenerates the artifact instead of picking a side. That did not happen on this merge — the merged file came out byte-identical to this branch's side, deleting `SEARCH_VIRTUAL_TYPES`, `foldAsciiCase`, `asciiCaseInsensitiveContains`, `asciiCaseInsensitiveRegexSource` and `isVirtualSearchField` from the recorded surface:

```
@@ -476,7 +477,6 @@   (merge result vs. the main parent)
     "SEARCH_AUTO_EXCLUDED_TYPES (const)",
-    "SEARCH_VIRTUAL_TYPES (const)",
```

`check:api-surface` is what caught it. Regenerating from the merged tree yields the union — #7123's five plus this branch's `HookDispatch` — and `check:generated` now reports all 11 artifacts current. Worth knowing that the `os-regen` driver was configured in the worktree and still did not run on this merge; if that reproduces, any branch merging `main` can quietly revert a generated baseline, and only the gate stands between that and `main`.

Also re-measured on the merged tree after rebuilding the closure: `objectql` 2833, `plugin-sharing` 418, `service-storage` 324, `spec` 9347, all passing. (Three `objectql` files failed on the first post-merge run — #7012, #6190 and the `$icontains` conformance — purely from stale dependency `dist/`; they pass after `turbo run build`.)

---

PR #7124, opened 2026-08-09
fix(plugin-auth): honour better-auth `Where.mode` and normalise the SCIM identifier (#5814)

### LINKED ISSUE #5814: plugin-auth: `convertWhere()` 整体忽略 better-auth 的 `Where.mode: 'insensitive'`(SCIM 会发它)
实测于 `origin/main@1624f4a`,在实施 #5710 时发现。

Blocked-by: #5702(`$icontains` 的执行面;今天写 `$icontains` 会被五个 driver 全部拒收/丢弃)

## 测到的东西

better-auth 1.7 的 `Where` 有第四个字段(`@better-auth/core/src/db/adapter/index.ts:324-343`):

```ts
mode?: "sensitive" | "insensitive" | undefined;   // @default "sensitive"
```

`packages/plugins/plugin-auth/src/objectql-adapter.ts` 的 `convertWhere()` 只读 `condition.field` / `operator` / `value`,**从不读 `mode`**。默认值是 `"sensitive"`,所以绝大多数调用不受影响 —— 但显式要求不敏感的调用,其要求被静默丢弃,实际大小写行为退化成「当前 driver 恰好怎么做」。

## 谁会发 `mode: 'insensitive'`

`@better-auth/scim`(`dist/index.mjs:566-576`):对 `caseExact === false` 的 SCIM 属性,解析出的 where 条件会带上 `mode: "insensitive"`。`userName` 正是这样一个属性(SCIM 的 `userName` 按 RFC 7643 是大小写不敏感的),而 SCIM 今天只映射 `eq`(`SCIMFilterOperatorMap = { eq: "eq" }`)。

后果:身份提供方用 `userName eq "Alice@example.com"` 查一个以 `alice@example.com` 存下的用户时,匹配与否取决于 driver —— 在大小写敏感的后端上查不到,而 SCIM 语义要求查得到。SCIM 的典型消费路径是「先查、查不到就建」,所以症状是**重复用户**而不是报错。

作用域:SCIM 默认关闭(`OS_SCIM_ENABLED`),所以今天只有开了 SCIM 的部署会碰到。

## 为什么现在还修不了

不敏感的**包含**匹配对应的算子是 `$icontains`,它今天是**纯声明面**:不在 `FILTER_OPERATORS`(`packages/spec/src/data/filter.zod.ts:1066-1081` 写明这是刻意分期),五个 driver 均不执行。所以在 #5702 给出执行面之前,adapter 无处可译。

更麻烦的一半:SCIM 发的是 **`eq` + insensitive**,而词表里**没有**「大小写不敏感的相等」算子 —— `$icontains` 是包含不是相等。所以本单落地时需要先回答一个契约问题:

1. 加一个 `$ieq`(词表扩张,要有真实业务拉动);
2. 或者把 `eq + insensitive` 降为 `$icontains` 再在应用层收窄(语义不等价,会多命中);
3. 或者判定「大小写不敏感的相等」由 SCIM 侧规范化(比如 `userName` 一律小写存储)来解决,adapter 侧不引入新算子。

个人倾向 3 → 1 的顺序:先看 SCIM 的实际部署是否已经靠规范化解决;只有在真实业务拉动出现时才扩词表(初创聚焦原则)。这需要维护者裁决,不是实施细节。

## 相关

- #5710(同一函数的 `contains` 支译成裸 `$regex`;已修)
- #5701 Q1=A(`$icontains` = ASCII-only 不敏感)/ Q2=A(`$contains` 家族大小写敏感)
- #5702(driver 侧下译 + `$icontains` 执行面)

os-zhuang:
Fixes #5814

Implements the maintainer's **option 3** ruling of 2026-08-09 — both halves. Re-measured against `origin/main@0bffdae` (the card's anchors were taken at `1624f4a` and had drifted; the drift is recorded below).

## The defect, re-verified at this branch point

better-auth's `Where` carries a fourth field — `mode?: "sensitive" | "insensitive"`, `@default "sensitive"` (`@better-auth/core/dist/db/adapter/index.d.mts:293-312`) — and `convertWhere()` read `field` / `operator` / `value` and nothing else. The default covers almost every caller, so the drop was invisible; the caller it is not invisible for is the one that explicitly asked.

`@better-auth/scim` is that caller, and the measurement moved one detail of the card: SCIM's `userName` is mapped onto better-auth's **`email`** field, not onto a `userName` column — `SCIMUserFilterAttributeFields = { userName: "email" }` (`@better-auth/scim@1.7.0-rc.1/dist/index.mjs:531`). Because RFC 7643 marks `userName` `caseExact: false` (`:409-417`), the parsed clause carries `mode: "insensitive"` (`:576`), and SCIM maps only `eq` (`:530`). So the clause reaching this adapter is `{ field: 'email', operator: 'eq', mode: 'insensitive' }` on model `user`.

With `mode` unread, that lookup matched a user stored as `alice@example.com` **or not depending on the driver**, and since SCIM provisioning is "look up, create if absent", a missed match did not raise — it provisioned a second user. This is the fail-**open** twin of #5813's fail-closed dropped predicate: there the query widened, here it answers a different question and looks fine doing it. Scope is unchanged: SCIM is off by default (`OS_SCIM_ENABLED`).

## Half 1 — normalisation, driven by one declared table

`NORMALISED_IDENTIFIER_FIELDS = { user: ['email'] }` names the set, keyed by better-auth model name, and **both** directions read it:

- **write** — `normaliseIdentifierWrite()` lower-cases the declared fields on `create` / `update` / `updateMany`, in the factory adapter and in the raw one;
- **read** — a `mode: 'insensitive'` comparand on a declared field is lower-cased (element-wise for `in` / `not_in`), which is then an *exact* match against the stored form.

One declaration for both halves is the point: a field cannot be added to the compare half without also being normalised on write, which would be the mirror-image defect — a lower-cased comparand hunting rows nobody lower-cased. A name heuristic (`/email$/`) would have allowed exactly that, so the set is spelled out and pinned by value; widening it is a deliberate act that has to come to the pin and say so.

No new query vocabulary is involved: the emitted filter stays a plain equality.

## Half 2 — the silent drop ends

`convertWhere()` now reads `condition.mode`, applying the producer's documented default explicitly (`?? 'sensitive'`) for the same reason `operator ?? 'eq'` is spelled out beside it: the factory materialises the default (`transformWhereClause`, `mode = "sensitive"`), but the raw `createObjectQLAdapter` is handed clauses that never passed through it.

- **Normalised identifier** → satisfied by construction, silently and correctly.
- **Any other field** → a loud `console.warn` naming the model, the field and the operator, and stating that the query is being answered **case-sensitively** so a differently-cased value will not match.

It deliberately does **not** throw. That is the ruling's own reasoning — fail-closed here would upgrade an occasional duplicate user into "`userName` queries entirely unavailable" — and it is the level AGENTS.md's degradation question asks for: nothing claims to have persisted, the caller simply gets a narrower answer than requested, which is functional degradation ⇒ `warn`, not `error`. `pnpm check:durability-log-level` passes.

A `sensitive` or absent `mode` keeps its comparand byte-for-byte, on a normalised column too. Folding case unasked would answer a different question than the one put — the same failure in the opposite direction.

## Existing rows: why no migration, and why none is needed

This was the question most likely to force a `needs_decision`, so it was measured rather than assumed:

- **Every existing producer already lower-cases `user.email` on write.** better-auth's own `internalAdapter` does it on `createUser` / `createOAuthUser` / `updateUser` / `updateUserByEmail` (`better-auth@1.7.0-rc.2/dist/db/internal-adapter.mjs:120,139,594,607`); `@better-auth/scim` does it again on its create path (`dist/index.mjs:2227`); ObjectStack's own paths do too (`admin-user-endpoints.ts:461,497`, `admin-import-users.ts:177`). The write half is therefore **idempotent over every producer that exists** — it adds no behaviour to any current write. What it adds is ownership: the invariant the read half depends on now lives in the layer that depends on it, instead of being inherited from an internal of a **prerelease** dependency that no published type describes, and it also covers the raw adapter path, which bypasses better-auth entirely.
- **The read half is scoped to `mode: 'insensitive'`, which only SCIM sends.** Every `sensitive`/absent query is unchanged byte-for-byte, so no stored row changes its match behaviour for any query shape in use today.
- **The legacy mixed-case row is pre-existing and untouched.** Such a row still misses an insensitive lookup — but `@better-auth/scim`'s own provisioning path already compares a *lower-cased* comparand against `user.email` with no `mode` at all (`dist/index.mjs:2227-2234`), so it was already invisible to SCIM before this change and in the same direction. Normalising the comparand takes nothing away from it. This is stated in the test file's "what is deliberately NOT pinned" section rather than left for the next reader.

## Deliberately not done

- **No `$ieq`** (option 1) — deferred for demonstrated pull, and it would need an in-memory execution face before it could join `FILTER_OPERATORS` at all (`filter.zod.ts`'s own recorded fail-closed reasoning; the #6815 one-query-two-numbers class).
- **No `eq + insensitive` → `$icontains` downgrade** (option 2) — containment is not equality.
- **No migration and no backfill**, per the measurement above; the ruling did not authorise one and none is needed.
- **`packages/spec`'s filter vocabulary is untouched**, and SCIM's default-off posture is unchanged.

## Tests — 22 pins across five faces

`packages/plugins/plugin-auth/src/scim-case-insensitive-identifier.test.ts`, following the harness its two siblings already use (`@objectstack/driver-sql` + better-sqlite3 `:memory:`).

1. **Contract** — which comparand the translation emits, off a spy engine.
2. **Behaviour** — the duplicate-user symptom on a real backend, *plus a discrimination pin* asserting that this backend's `=` really is case-exact. Without it the behavioural pin would be always-green on any backend that folds case, and would say nothing about the adapter.
3. **The warning** — that it fires and names the field, **and that it stays silent** for a normalised identifier and for `sensitive`/absent mode. The silence half is what makes the firing half mean something.
4. **The write half** — `create` / `update` / `updateMany`, factory and raw, plus a model outside the set left untouched.
5. **The declared set** — pinned by value, including that it names `email` and not `userName`, since a set spelled `userName` would look plausible and match nothing.

### Reverse verification — direction predicted in writing before running

Prediction recorded first, then the whole fix removed (`git checkout origin/main -- objectql-adapter.ts`; no `git stash` — that stack is shared across worktrees).

Measured: **13 red / 9 green**, the exact split predicted. Red: both contract pins, both behavioural pins, the both-halves round trip, both warning pins, all four write pins, both set pins — e.g. `expected { email: 'Alice@Example.com' } to deeply equal { email: 'alice@example.com' }` and `expected [] to deeply equal [ 'u_alice' ]` (the duplicate-user symptom, witnessed).

Green, correctly: the sensitive/absent comparand pins, the raw-adapter default pin, the no-match pin, loud-is-not-fail-closed, both silence pins, the non-set model pin — and the **discrimination** pin, which is the one that proves sqlite's `=` is case-exact and therefore that the behavioural pin can fail at all. A guard that correctly stays green under its own mutation is a result, not a gap.

**Honest delta — one prediction was wrong.** I predicted that the deleted `NORMALISED_IDENTIFIER_FIELDS` export would fail the whole *file* under the mutation and mask the two silence-class faces, requiring a second surgical mutation to read them. Measured: vitest/esbuild resolves a missing named export to `undefined` rather than throwing at import time, so only the two set-pins went red (`expected undefined to deeply equal { user: ['email'] }`) and those faces were readable after all. The second mutation was therefore unnecessary and not run. The prediction about those faces themselves — that they stay green — held.

## Verification

- `pnpm --filter @objectstack/plugin-auth test` — **996 passed / 996**, 40 files.
- `pnpm --filter @objectstack/runtime test` — **1837 passed / 1837**, 118 files (downstream consumer; see the comment below for the stale-`dist` false red that preceded it).
- `pnpm --filter @objectstack/plugin-auth typecheck` — clean (`tsc --noEmit`, no output).
- `pnpm lint` — clean.
- 33 ESLint-job family gates run individually, all **PASS**, including `check:engine-double-contract`, `check:error-code-casing`, `check:route-envelope`, `check:durability-log-level`, `check:nul-bytes`, `check:changeset-gate-self-tests`, `check:published-files`, `check:query-options-erasure`.
- Changeset: `.changeset/scim-case-insensitive-identifier.md` (`@objectstack/plugin-auth: patch`).

**CI on this head has converged: 25 check runs, 23 success + 2 skipped (path-filtered), zero failures** — including **ESLint** (the job the family gates run inside) and **TypeScript Type Check**.

---

### DIFF
--- a/.changeset/scim-case-insensitive-identifier.md
+++ b/.changeset/scim-case-insensitive-identifier.md
@@ -1,0 +1,47 @@
+---
+"@objectstack/plugin-auth": patch
+---
+
+fix(plugin-auth): honour better-auth's `Where.mode`, and normalise the identifier SCIM matches on (#5814)
+
+better-auth's `Where` carries a fourth field — `mode?: "sensitive" | "insensitive"`,
+`@default "sensitive"` — and `convertWhere()` in the ObjectQL adapter read `field` /
+`operator` / `value` and nothing else. The default covers almost every caller, so the
+drop was invisible; the caller it is not invisible for is the one that explicitly asked.
+
+`@better-auth/scim` is that caller. SCIM's `userName` is case-insensitive by RFC 7643
+(`caseExact: false`), so a `filter=userName eq "Alice@example.com"` reaches this adapter
+as `{ field: 'email', operator: 'eq', mode: 'insensitive' }`. With `mode` unread, whether
+it matched a user stored as `alice@example.com` came down to how the driver under the
+auth path happens to compare strings — and because SCIM provisioning is "look up, create
+if absent", a missed match did not raise an error, it provisioned a **second user**.
+Only deployments that turned SCIM on (`OS_SCIM_ENABLED`, off by default) were exposed.
+
+Both halves of the fix, per the maintainer's ruling on #5814:
+
+- **Normalisation, not new vocabulary.** `sys_user.email` — the field SCIM's `userName`
+  maps onto — is now stored lower-cased and compared lower-cased by this adapter. An
+  insensitive lookup lower-cases its comparand, which is an *exact* match against the
+  stored form, so nothing in the query vocabulary changes. The set is a declared table
+  (`NORMALISED_IDENTIFIER_FIELDS`), not a name heuristic, and it drives the read and
+  write halves from one place so a field cannot be added to one of them only.
+- **The silent drop ends.** `convertWhere()` handles `mode` explicitly. On a normalised
+  identifier the request is satisfied by construction. On **any other** field, a
+  `mode: 'insensitive'` clause now emits a loud warning naming the model, the field and
+  the operator, and stating that the query is being answered case-sensitively — instead
+  of answering a different question and looking fine doing it. It deliberately does not
+  throw: refusing here would turn an occasional duplicate user into "`userName` queries
+  entirely unavailable", which is the worse trade on an authentication path.
+
+No migration ships and none is needed. Every existing write path already lower-cased
+`user.email` before reaching the adapter (better-auth's own `internalAdapter` does it on
+`createUser` / `createOAuthUser` / `updateUser` / `updateUserByEmail`, and SCIM's create
+path does it again), so the write half changes no existing behaviour — it moves the
+invariant the read half depends on into the layer that depends on it, instead of
+inheriting it from an internal of a prerelease dependency. Queries that do not set
+`mode`, or set it to `"sensitive"`, keep their comparand byte-for-byte: folding case
+unasked would be the same failure in the opposite direction.
+
+Adding a case-insensitive equality operator (`$ieq`) was deferred until there is
+demonstrated pull for it, and downgrading `eq + insensitive` to `$icontains` was
+rejected — containment is not equality.
--- a/packages/plugins/plugin-auth/src/objectql-adapter.ts
+++ b/packages/plugins/plugin-auth/src/objectql-adapter.ts
@@ -134,6 +134,107 @@ export const SUPPORTED_WHERE_OPERATORS = [
   'ends_with',
 ] as const satisfies readonly WhereOperator[];
 
+// ---------------------------------------------------------------------------
+// Case-insensitive identifiers — the normalised set (#5814)
+// ---------------------------------------------------------------------------
+
+/**
+ * The identifier fields this adapter stores and compares **normalised**
+ * (lower-cased), per the maintainer's 2026-08-09 ruling on #5814.
+ *
+ * ## Why a declared table and not a name heuristic
+ *
+ * Membership is a contract with two obligations that must hold **together**:
+ *
+ *   1. every write through this adapter normalises the field
+ *      ({@link normaliseIdentifierWrite}), and
+ *   2. a `mode: 'insensitive'` comparand on the field is normalised the same
+ *      way ({@link convertWhere}), which is then an EXACT match against (1).
+ *
+ * Honour (2) without (1) and you get the mirror-image defect of the one #5814
+ * was filed on: the query stops matching rows it used to match. So the set is
+ * spelled out, keyed by better-auth model name, and both directions are driven
+ * from this one declaration — a field cannot be added to one half only. A
+ * heuristic (`/email$/`, `endsWith('_email')`) would let a field into the
+ * compare half without anyone deciding that its writes are normalised, which is
+ * exactly the widening this table exists to prevent.
+ *
+ * `sys_user.email` is the whole set today because it is the whole demand:
+ * `@better-auth/scim` maps SCIM's `userName` onto better-auth's **`email`**
+ * field (`SCIMUserFilterAttributeFields = { userName: "email" }`,
+ * `@better-auth/scim@1.7.0-rc.1/dist/index.mjs:531`) and marks it
+ * `caseExact: false` (`:409-417`, RFC 7643), so the where clause that reaches
+ * this adapter is `{ field: 'email', operator: 'eq', mode: 'insensitive' }` on
+ * model `user`. Adding a member is a deliberate act: it must ALSO be provable
+ * that every producer of that field writes it normalised, and
+ * `scim-case-insensitive-identifier.test.ts` pins the set so the addition
+ * cannot be quiet.
+ *
+ * Keyed by better-auth **model** name (`user`), not by the protocol object name
+ * (`sys_user`), because that is the name every call site in this file already
+ * holds and the name `Where.field` has been transformed against. `user` is not
+ * one of the bridged models (see `AUTH_MODEL_TO_PROTOCOL`), so its field names
+ * reach `convertWhere` untouched by `remapWhere`, and `email` is spelled
+ * identically on both sides (`auth-schema-config.ts` maps only the names that
+ * actually differ).
+ */
+export const NORMALISED_IDENTIFIER_FIELDS: Readonly<Record<string, readonly string[]>> = {
+  user: ['email'],
+};
+
+/** Is `field` on `model` stored and compared lower-cased? */
+function isNormalisedIdentifier(model: string, field: string): boolean {
+  return NORMALISED_IDENTIFIER_FIELDS[model]?.includes(field) ?? false;
+}
+
+/**
+ * Lower-case a comparand, element-wise for the array-valued operators
+ * (`in` / `not_in`).
+ *
+ * Non-strings are returned untouched: better-auth's own `Where.mode` doc says
+ * the flag "Only applies to string values", and an identifier column holding a
+ * number/boolean/Date has no case to fold.
+ */
+function normaliseComparand(value: unknown): unknown {
+  if (typeof value === 'string') return value.toLowerCase();
+  if (Array.isArray(value)) return value.map((v) => (typeof v === 'string' ? v.toLowerCase() : v));
+  return value;
+}
+
+/**
+ * Normalise the identifier fields of an outgoing write payload (#5814).
+ *
+ * This is obligation (1) of {@link NORMALISED_IDENTIFIER_FIELDS} — the half
+ * that makes the comparand normalisation in {@link convertWhere} an *exact*
+ * match rather than a hopeful one.
+ *
+ * It is deliberately NOT a redundant belt over better-auth's own lower-casing.
+ * better-auth's `internalAdapter` does lower-case `user.email` on
+ * `createUser` / `createOAuthUser` / `updateUser` / `updateUserByEmail`
+ * (`better-auth@1.7.0-rc.2/dist/db/internal-adapter.mjs:120,139,594,607`), but
+ * that is an *internal* of a **prerelease** dependency, invisible to any
+ * published type, and it does not cover the raw {@link createObjectQLAdapter}
+ * path (hand-built calls that never pass through better-auth at all). The
+ * invariant the read half depends on has to be owned where it is relied upon.
+ *
+ * Idempotent by construction, so a payload better-auth already normalised is
+ * unchanged — which is why this adds no behaviour to any existing write.
+ */
+function normaliseIdentifierWrite<T extends Record<string, any>>(model: string, data: T): T {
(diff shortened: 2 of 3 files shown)


os-zhuang:
**Downstream sweep: green.**

`pnpm --filter @objectstack/runtime test` — **118 files passed / 118, 1837 tests passed / 1837**, exit 0.

Recording the false red honestly, since it cost a lap and reads exactly like a broken import: the first run of this suite failed a dozen files outright. That was the AGENTS.md §9 stale-artefact trap, not this change — `@objectstack/runtime` resolves `@objectstack/plugin-auth` through its **`dist`**, and in a fresh worktree that `dist` predates the edit. `pnpm --filter '@objectstack/runtime^...' build` first, then the suite, and it is green with no source change in between.

Full local verification for this PR, per gate:

| Gate | Result |
|---|---|
| `pnpm --filter @objectstack/plugin-auth test` | 996 / 996 passed (40 files) |
| `pnpm --filter @objectstack/runtime test` | 1837 / 1837 passed (118 files) |
| `pnpm --filter @objectstack/plugin-auth typecheck` | clean (`tsc --noEmit`, no output) |
| `pnpm lint` | clean |
| 33 ESLint-job family gates, run individually | all PASS |

The family gates were run one by one rather than trusting the aggregate, because they run *inside* the ESLint job and a red one surfaces nowhere in `pnpm test` output: `check:engine-double-contract`, `check:error-code-casing`, `check:route-envelope`, `check:durability-log-level` (the one this PR's `warn`-not-`error` choice had to answer to), `check:nul-bytes`, `check:wildcard-fallthrough`, `check:init-service-contract`, `check:doc-authoring`, `check:published-files`, `check:slot-lookup`, `check:query-options-erasure`, `check:verify-stand-in`, `check:stack-collection-maps`, `check:docs-audit-scope`, `check:role-word`, `check:quick-reference-counts`, `check:adr-anchors`, `check:org-identifier`, `check:authz-resolver`, `check:service-providers`, `check:meta-type-normalized`, `check:startup-registry-verdict`, `check:objectui-changeset`, `check:changeset-gate-self-tests`, `check:empty-changeset`, `check:release-notes`, `check:release-body`, `check:node-version`, `check:workflow-status-functions`, `check:shard-attestation`, `check:required-contexts`.

CI's own conclusions are the verdict that counts; this is the local half.


---

os-zhuang:
**PM review — PASS. Marked ready and enqueued** (`added_to_merge_queue` 18:30:38Z, on the first `enable`). Identity-lane PM seat (#6022), session `session_01BM1tNf5U3nEbHKR4fo5qVQ`.

## Verified independently, not taken on report

- **CI, per job**: 26 runs, **none non-green**. ESLint ✅ 18:18:28Z, TypeScript Type Check ✅ 18:24:23Z, Check Changeset ✅ 18:16:18Z, Test Core ✅, Dogfood Regression Gate ✅.
- **Both ruled halves present** — the thing this review existed to check, since shipping only the normalisation half would leave the fail-open silence this card was filed on. Half 1: 20 added lines carrying `NORMALISED_IDENTIFIER_FIELDS` / `normaliseIdentifierWrite`, with the table declared as `{ user: ['email'] }` and driving **read and write from one place**. Half 2: 8 added lines reading `condition.mode` and emitting the named `console.warn`.
- **Scope**: 3 files, +760 −36, against the merge base. `packages/spec` untouched (0 files) — the ruling forbade touching the filter vocabulary. No `content/docs/releases/`. No migration or backfill file.

## The two ruled-out options are PINNED ABSENT, which is better than merely not done

`$ieq` and `$icontains` both appear in the diff — I checked what each occurrence **is** rather than counting hits. All ten are prose explaining the deferral/rejection, one warning string that names the missing operator to the operator reading the log, and two **negative assertions**:

```js
expect(JSON.stringify(query.where)).not.toContain('$ieq');
expect(JSON.stringify(query.where)).not.toContain('$icontains');
```

So a future change cannot quietly reintroduce either shape without a test going red. That is a stronger guarantee than the ruling asked for.

## The anchor correction is the most valuable thing in this PR

The card's own title and body say `userName`. Measured: `@better-auth/scim` maps `userName` onto better-auth's **`email`** field (`SCIMUserFilterAttributeFields = { userName: "email" }`, `dist/index.mjs:531`). A declared set spelled `userName` **would have looked correct in review and matched nothing at runtime** — a silent no-op shipped under a green suite. The set names `email`, and the PR says why.

## The existing-rows question — measured, and correctly did NOT become a decision card

The envelope flagged this as the likely `needs_decision` trigger. The dev answered it with readings rather than escalating on principle: every existing producer already lower-cases `user.email` on write (better-auth's `internalAdapter` at `:120/:139/:594/:607`, `@better-auth/scim` at `:2227`, and ObjectStack's own `admin-user-endpoints.ts:461,497` / `admin-import-users.ts:177`), so the write half is **idempotent over every producer that exists**; and the read half is scoped to `mode: 'insensitive'`, which only SCIM sends, so every sensitive/absent query keeps its comparand byte-for-byte. No migration needed, none shipped, and no half-migration — which the ruling did not authorise.

The residual case (a mixed-case row written outside the normalising paths staying unreachable from a SCIM insensitive lookup) is **pre-existing and unchanged in direction** — SCIM already compared a lower-cased comparand against `user.email` with no `mode` at all. Recorded in the PR body *and* in the test file's "what is deliberately NOT pinned" section rather than left for the next reader; correctly not filed as a new issue, since acting on it is a backfill decision for the maintainer, not an adapter change.

## Reverse verification — accepted, including the prediction that missed

Predicted **13 red / 9 green** in writing first; measured exactly that split. One honest delta reported rather than smoothed: the dev expected the missing export to fail the whole file and mask the silence-class faces, requiring a second surgical mutation — measured, vitest resolves a missing named export to `undefined`, so only the two set-pins went red and those faces were readable under the first mutation after all. The planned M2 was therefore unnecessary and **was not run**. The discrimination pin staying green is what proves sqlite's `=` is case-exact, and therefore that the behavioural pin can fail at all.

## One thing this seat fixed rather than bounced

The PR body was missing the Claude Code attribution footer required of every GitHub artifact here. Appended by this seat rather than sent back for a round trip — the author was twice killed mid-task by token limits (host-environment blocks, recorded as such and never counted against the run), and a round trip for a two-line footer would have risked a third.

---

Issue #7135, opened 2026-08-09
Re-annotate plugin-approvals + plugin-reports onto the full `ExecutionContext` and delete the now-redundant casts (services half of #7070)


os-zhuang:
Services-lane half of the #7070 split (promotion executed by the triage seat per the split plan recorded on that card, 2026-08-09). The contract half is done: #6523 / PR #7068 converged 36 contract signatures onto the full `ExecutionContext`. This card widens the **implementation-side annotations** in the services lane so the casts the narrow type forced become deletable.

## File face (verified on `origin/main` @ `445a0c2`)

| file | `SharingExecutionContext` refs |
|:---|---:|
| `packages/plugins/plugin-approvals/src/approval-service.ts` | 27 |
| `packages/plugins/plugin-reports/src/report-service.ts` | 14 |
| `packages/plugins/plugin-approvals/src/approval-node.ts` | 2 |

Casts that become deletable, all re-verified live:

- `approval-service.ts:661` — `const posture = (context as any).posture;` feeding `isOverrideActor()`; the doc comment above it already names the derived `posture` (ADR-0095) as the intended read, and the contract now declares the field. This is the load-bearing one: an unchecked cast on an enforcement input.
- `approval-service.ts:2773, 2813, 3259` — `SYSTEM_CTX as unknown as SharingExecutionContext`
- `approval-node.ts:174` — `} as unknown as SharingExecutionContext);`

## Scope boundary (both inherited from #7070)

- ⛔ `(context as any).organizationId` (`approval-service.ts:1465`, `:3649`, `:3916`) is a **separate question** — `organizationId` is not on `ExecutionContext` at all; per #5858 / `check:org-identifier` that spelling has its own history. Do not fold it in.
- The removal of the exported `SharingExecutionContext` type itself rides **whichever of the two split halves lands second** — do not remove it here if the identity half (sibling card) is still open.

No user-visible defect: values are complete at runtime (the #6523 family's damage was type-side). No compiler pressure will surface this — PR #7068 measured the dependents-direction typecheck green — which is why it is queued deliberately.

Part of #7070 (progress view). Sibling: identity half. Refs: #6523, PR #7068, #6206 (ruling), #6430 / PR #6511.

os-zhuang:
**Cross-reference from the sibling half — NOT a claim on this card.** This seat is working #7136 (identity lane) only; #7135 is unclaimed and untouched, and nothing here changes its assignee, labels or state.

[...]

Issue #7070, opened 2026-08-09
finding: consumer half of #6523 — three plugin implementations still annotate `SharingExecutionContext`, so `(context as any).posture` is still needed to read a field the contract now declares


os-zhuang:
Out-of-scope observation from implementing #6523 (PR #7068), filed rather than fixed — the card was scoped to the contract face and explicitly forbade touching the implementation bodies, exactly as #6430's contract half and its plugin half were separated.

Blocked-by: #6523

## Fact

#6523 / PR #7068 converged 36 contract signatures onto the full `ExecutionContext`. The three implementations behind those contracts still annotate their own method parameters with the six-field `SharingExecutionContext`, so nothing they read has widened. Measured on `origin/main` @ `08863dd18`:

| file | `SharingExecutionContext` refs |
|:---|---:|
| `packages/plugins/plugin-approvals/src/approval-service.ts` | 27 |
| `packages/plugins/plugin-sharing/src/sharing-service.ts` | 21 |
| `packages/plugins/plugin-reports/src/report-service.ts` | 14 |
| `packages/plugins/plugin-sharing/src/sharing-rule-service.ts` | 8 |
| `packages/plugins/plugin-sharing/src/exec-context-seam.testkit.ts` | 3 |
| `packages/plugins/plugin-audit/src/comment-access-hooks.ts` | 2 |
| `packages/plugins/plugin-approvals/src/approval-node.ts` | 2 |

The casts those narrow annotations force, all still present:

```
packages/plugins/plugin-approvals/src/approval-service.ts:661
  const posture = (context as any).posture;              // isOverrideActor()
packages/plugins/plugin-approvals/src/approval-service.ts:2773, 2813, 3259
  SYSTEM_CTX as unknown as SharingExecutionContext
packages/plugins/plugin-approvals/src/approval-node.ts:174
  } as unknown as SharingExecutionContext);
packages/plugins/plugin-sharing/src/exec-context-seam.testkit.ts:106
  return { ...authz, isSystem: false } as unknown as SharingExecutionContext;
```

`:661` is the interesting one. Its own doc comment says the gate reads "the derived `posture`, ADR-0095" off the resolved exec context; the `as any` exists only because the declared type said the field was not there. After #6523 the contract declares it, so the cast is now removable — and until it is removed the expression stays unchecked, which is what an `as any` on an enforcement input costs: the next field read through the same expression is unverified too.

[...]

PR #7140, opened 2026-08-09
refactor(plugin-sharing,plugin-audit): annotate enforcement on the full `ExecutionContext` and delete the double-casts (#7136)

### LINKED ISSUE #7136: Re-annotate plugin-sharing + plugin-audit onto the full `ExecutionContext` and delete the double-casts (identity half of #7070)
Identity-lane half of the #7070 split (promotion executed by the triage seat per the split plan recorded on that card, 2026-08-09). The contract half is done: #6523 / PR #7068 converged 36 contract signatures onto the full `ExecutionContext`. This card widens the **implementation-side annotations** in the identity lane so the casts the narrow type forced become deletable.

## File face (verified on `origin/main` @ `445a0c2`)

| file | `SharingExecutionContext` refs |
|:---|---:|
| `packages/plugins/plugin-sharing/src/sharing-service.ts` | 21 |
| `packages/plugins/plugin-sharing/src/sharing-rule-service.ts` | 8 |
| `packages/plugins/plugin-sharing/src/exec-context-seam.testkit.ts` | 3 |
| `packages/plugins/plugin-audit/src/comment-access-hooks.ts` | 2 |

The double-cast specimen, re-verified live: `exec-context-seam.testkit.ts:106` — `return { ...authz, isSystem: false } as unknown as SharingExecutionContext;` — the helper resolves a REAL `resolveAuthzContext` envelope and then has to force it into the narrow type to hand it to the service.

## Scope boundary (both inherited from #7070)

- ⛔ Any `organizationId` spellings stay out — per #5858 / `check:org-identifier` that vocabulary has its own history; this card is annotation widening + redundant-cast deletion only.
- The removal of the exported `SharingExecutionContext` type itself rides **whichever of the two split halves lands second**. #6523 deliberately left it exported and documented as migration residue precisely to keep these files compiling; once both halves land, it becomes removable — the second-lander deletes it (or files the one-line follow-up if cross-lane timing makes that cleaner).

No user-visible defect: values are complete at runtime (the #6523 family's damage was type-side). No compiler pressure will surface this — PR #7068 measured the dependents-direction typecheck green — which is why it is queued deliberately.

Part of #7070 (progress view). Sibling: services half (#7135). Refs: #6523, PR #7068, #6206 (ruling), #6430 / PR #6511.

os-zhuang:
Fixes #7136

Consumer half of #6523 / PR #7068, identity lane of the #7070 split. The contract half converged 36 signatures onto the full `ExecutionContext` (the #6206 ruling: enforcement adjudicates on the whole `resolveAuthzContext` envelope, never a per-site subset). The implementations behind those contracts still annotated their own parameters with `SharingExecutionContext`, the six-field shape the contracts used to name — so nothing they could actually *read* had widened, and the casts that narrowness forced were all still there.

## What changed

| file | `SharingExecutionContext` refs, before → after |
|:--|--:|
| `packages/plugins/plugin-sharing/src/sharing-service.ts` | 21 → 0 |
| `packages/plugins/plugin-sharing/src/sharing-rule-service.ts` | 8 → 0 |
| `packages/plugins/plugin-sharing/src/exec-context-seam.testkit.ts` | 3 → 0 |
| `packages/plugins/plugin-audit/src/comment-access-hooks.ts` | 2 → 0 |

27 enforcement parameters now declare `ExecutionContext`, plus the two return types that produce the contexts feeding them. The casts that go with them:

- **`exec-context-seam.testkit.ts:106`** — the specimen this card names. The helper resolved a REAL `resolveAuthzContext` envelope and then forced it into the narrow type: `return { ...authz, isSystem: false } as unknown as SharingExecutionContext;`. It now returns what it resolved. That matters beyond tidiness: this seam exists (#5859) so a drift in the resolver's output *breaks* the tests that trust it, and a double cast on that value is the seam lying about itself.
- **`SYSTEM_CTX as any` at all 10 of its call sites in `sharing-rule-service.ts`** — the system context is typed as the envelope and passed as itself. An erasure on an enforcement input switches checking off for the whole argument, not just for the readonly-array mismatch that provoked it.
- **Six `(context as any)` reads of `userId` / `tenantId`** — 3 in `sharing-service.ts`, 3 in `sharing-rule-service.ts` — now reading declared fields.

**Two cast families are deliberately kept**, and are now documented where they sit rather than left looking like oversights:

- `__readScope` / `__writeScope` are **not** fields of `ExecutionContext`. They are private keys plugin-security's middleware stamps onto the context it forwards (`sc.__readScope = …`, `security-plugin.ts`). Widening the annotation does not make these deletable, and the fix is not to declare them on the envelope — that would publish a middleware seam as authorable, client-supplied vocabulary.
- `organizationId` is not on the envelope at all. That spelling has its own history (#5858 / `check:org-identifier`) and was held out of #7070 on purpose, so the three reads of it stay cast. Only the `tenantId` half of those expressions lost its cast, which makes the asymmetry the visible marker of which of the two names the contract actually knows.

`plugin-audit`'s `callerContext` gets the widened annotation but keeps its five-field projection — see "deliberately not done" below.

## Verification — and why a green typecheck is not it

This card has no user-visible defect and no compiler pressure in either direction: the values were always complete at runtime (this family's damage was type-side), and PR #7068 already measured the dependents-direction typecheck green. **The typecheck passed before this branch existed, so it is a regression check here and not evidence.** The evidence is the cast counts and the pin.

Measured on this branch point (`0fd855624`, not the card's `445a0c2` — the four per-file counts re-measured identical).

**`SharingExecutionContext` references, before → after** (`grep -c` over the four files): 21 → 0, 8 → 0, 3 → 0, 2 → 0.

**Context casts, before → after** (`grep -oF | wc -l` summed over the same four files):

| cast | before | after |
|:--|--:|--:|
| `as unknown as SharingExecutionContext` | 1 | 0 |
| `SYSTEM_CTX as any` | 10 | 0 |
| `(context as any)` — total occurrences | 14 | 8 |
| … of which are code reads | 12 | 6 |
| … of which are prose in doc comments | 2 | 2 |

The 6 surviving code reads are exactly the two families held out on purpose: `__readScope` / `__writeScope` (3) and `organizationId` (3). The 6 deleted ones were all `userId` / `tenantId` — fields the envelope declares.

**Gates** (regression checks, not evidence of the change):

```
pnpm --filter @objectstack/plugin-sharing --filter @objectstack/plugin-audit typecheck   Done, Done
pnpm --filter @objectstack/plugin-sharing --filter @objectstack/plugin-audit test
    plugin-sharing  Test Files 16 passed (16)   Tests 410 passed (410)
    plugin-audit    Test Files  9 passed  (9)   Tests 137 passed (137)
pnpm exec eslint --no-inline-config packages/plugins/plugin-sharing/src packages/plugins/plugin-audit/src   clean
check:org-identifier      OK (1726 author-facing source files, no removed session.tenantId alias)
check:authz-resolver      OK (single shared authorization resolver intact)
check:slot-lookup         OK (108 unswept sites in 25 files, none new)
check:type-check-coverage OK (63/77 packages type-checked)
check:nul-bytes           OK (6590 files, no raw ASCII control bytes)
check:empty-changeset     OK (1 declaring changeset added)
```

### The mechanical guard, and whether it earns its weight

Yes — because without it *nothing* detects a regression. A re-narrowed annotation compiles, ships and passes every test in these packages, which is exactly why this card had to be queued deliberately rather than discovered.

`packages/plugins/plugin-sharing/src/exec-context-annotation.pin.ts` hands each enforcement parameter a **fresh object literal** naming envelope-only fields (`posture`, `accessible_org_ids`, `org_user_ids`). TypeScript's excess-property check rejects those the moment a parameter narrows back. That is the only direction available: a `@ts-expect-error` asserting the reverse would be *unsatisfied* and fail the build, because a narrow context IS structurally assignable to a wide parameter — the boundary `SharingExecutionContext`'s own doc block already records.

It also carries four **negative** cases, because a parameter erased to `any` would swallow every positive just as happily. `posture: 'SUPERUSER'`, `userId: 42`, `accessible_org_ids: 'org_a'` and `organizationId:` must all still fail.

It is a `.pin.ts`, not a `*.test.ts`, and that is load-bearing: `packages/plugins/plugin-sharing/tsconfig.json` excludes `**/*.test.ts` (a measured `TEST_DEBT` of 3 in `scripts/check-type-check-coverage.mjs`), so `tsc --noEmit` never reads this package's test files. A pin written there would be a phantom check that stays green however it is broken — AGENTS.md's `PINS_CHECKED` trap, the same hole #6212 measured on driver-mongodb. Same convention as the existing `packages/objectql/src/register-object-authored-shape.pin.ts`, and imported by nothing, so tsup (entry `src/index.ts`) never bundles it.

### Reverse verification

**Prediction, recorded before running it:** this change has no runtime behaviour, so reverting it would leave the entire suite green — the honest expectation here is *not* a red — while the pin, and only the pin, would go red. Both halves measured, by taking the fix out with `git checkout origin/main --` on the four files and keeping `exec-context-annotation.pin.ts`:

```
$ pnpm --filter @objectstack/plugin-sharing typecheck
src/exec-context-annotation.pin.ts(65,46): error TS2353: Object literal may only specify known properties, and 'posture' does not exist in type 'SharingExecutionContext'.
src/exec-context-annotation.pin.ts(66,46): error TS2353: ... 'posture' does not exist in type 'SharingExecutionContext'.
src/exec-context-annotation.pin.ts(67,45): error TS2353: ... 'posture' does not exist in type 'SharingExecutionContext'.
src/exec-context-annotation.pin.ts(68,50): error TS2353: ... 'accessible_org_ids' does not exist in type 'SharingExecutionContext'.
src/exec-context-annotation.pin.ts(73,30): error TS2339: Property 'posture' does not exist on type 'SharingExecutionContext'.
src/exec-context-annotation.pin.ts(73,55): error TS2339: Property 'posture' does not exist on type 'SharingExecutionContext'.
Exit status 2

$ pnpm --filter @objectstack/plugin-sharing --filter @objectstack/plugin-audit test
    plugin-sharing  Test Files 16 passed (16)   Tests 410 passed (410)
    plugin-audit    Test Files  9 passed  (9)   Tests 137 passed (137)
```

So: **547 runtime tests cannot tell this change from its own reversal**, which is the card's premise measured rather than asserted, and the reason a fabricated red would have been the wrong thing to report. The pin fails on all six lines it was written to fail on. No `@ts-expect-error` became unsatisfied under the revert (tsc reports no TS2578), so the negative cases keep asserting in both states rather than turning into noise.

## Deliberately not done

- **`plugin-approvals` and `plugin-reports`** — the services-lane half (#7135), a different seat and a disjoint file set. `packages/services` is untouched.
- **The exported `SharingExecutionContext` type stays.** Its removal rides whichever of the two split halves lands second; #7135 is still open and `pm:queue` as of this PR, and #6523 left the type exported precisely to keep those files compiling. Deleting it here would break the sibling half mid-flight. The coordination point (including the `plugin-sharing/src/index.ts` re-export, which is a public-surface removal) is recorded on #7135 for whichever seat lands second.
- **`callerContext`'s five-field projection in `plugin-audit`** — filed as **#7141**. Widening its annotation is type-side and inert; forwarding `exec` whole is a RUNTIME change with a real enforcement consequence (it drops `onBehalfOf` before a probe documented to fail closed on it, and forwarding it whole would leak one object's access depth into another object's owner-match), so it is filed rather than folded in.

## Commits

Two, deliberately not squashed: the change, then a correction to a number in it. The `SYSTEM_CTX` cast count was first written as 11 because the measurement was taken *after* the new doc comment — which itself quotes the string `SYSTEM_CTX as any` — had been inserted, so the comment counted itself. On `origin/main` the erasure appears at exactly 10 call sites.

---

### DIFF
--- a/.changeset/sharing-audit-exec-context-annotations.md
+++ b/.changeset/sharing-audit-exec-context-annotations.md
@@ -1,0 +1,48 @@
+---
+"@objectstack/plugin-sharing": patch
+"@objectstack/plugin-audit": patch
+---
+
+refactor(plugin-sharing,plugin-audit): enforcement implementations annotate the full `ExecutionContext` (#7136)
+
+The consumer half of #6523. That change converged 36 contract signatures onto
+the complete `resolveAuthzContext` envelope, applying the #6206 ruling —
+enforcement adjudicates on the whole envelope, never a per-site subset. The
+implementations behind those contracts still annotated their own parameters
+with `SharingExecutionContext`, the six-field shape the contracts used to name,
+so nothing they could *read* had widened.
+
+`SharingService`, `SharingRuleService`, the sharing exec-context seam and
+plugin-audit's comment-access gates now declare `ExecutionContext` on all 27 of
+those parameters — plus the two return types that produce the contexts feeding
+them — and the casts the narrow annotation forced are gone:
+
+- `exec-context-seam.testkit.ts` resolved a REAL context and then had to force
+  it into the narrow type — `{ ...authz, isSystem: false } as unknown as
+  SharingExecutionContext`. It now returns what it resolved, so a drift in
+  `resolveAuthzContext`'s output reaches the tests that trust this seam instead
+  of being absorbed by a double cast.
+- `SharingRuleService`'s system context is typed as the envelope and passed as
+  itself, retiring `SYSTEM_CTX as any` at all 11 of its call sites — an erasure
+  on an enforcement input switches checking off for the whole argument, not
+  just for the readonly-array mismatch that provoked it.
+- The `(context as any).userId` / `.tenantId` reads in `SharingService` now read
+  declared fields.
+
+**No runtime behaviour changes.** The values were always complete — this
+family's damage was type-side — so every gate answers exactly what it answered
+before. Method parameters only WIDEN what they accept, so no caller is affected.
+
+Two casts are deliberately kept, and are now documented where they sit:
+`__readScope` / `__writeScope` are private keys plugin-security's middleware
+stamps onto the context it forwards and are not fields of the envelope, and
+`organizationId` is not on the envelope at all — that spelling has its own
+history (#5858 / `check:org-identifier`) and was held out of this change.
+
+Because a re-narrowed annotation would compile, ship and pass every test in
+these packages, the convergence is pinned by a new compile-time module,
+`exec-context-annotation.pin.ts`: it hands each enforcement parameter a fresh
+literal naming envelope-only fields (`posture`, `accessible_org_ids`,
+`org_user_ids`), which TypeScript's excess-property check rejects the moment a
+parameter narrows back, plus negative cases so a parameter erased to `any`
+cannot pass either.
--- a/.changeset/sharing-audit-exec-context-annotations.md
+++ b/.changeset/sharing-audit-exec-context-annotations.md
@@ -23,7 +23,7 @@ them — and the casts the narrow annotation forced are gone:
   `resolveAuthzContext`'s output reaches the tests that trust this seam instead
   of being absorbed by a double cast.
 - `SharingRuleService`'s system context is typed as the envelope and passed as
-  itself, retiring `SYSTEM_CTX as any` at all 11 of its call sites — an erasure
+  itself, retiring `SYSTEM_CTX as any` at all 10 of its call sites — an erasure
   on an enforcement input switches checking off for the whole argument, not
   just for the readonly-array mismatch that provoked it.
 - The `(context as any).userId` / `.tenantId` reads in `SharingService` now read
--- a/packages/plugins/plugin-audit/src/comment-access-hooks.ts
+++ b/packages/plugins/plugin-audit/src/comment-access-hooks.ts
@@ -55,7 +55,8 @@
  * load-bearing.
  */
 
-import type { ISharingService, SharingExecutionContext } from '@objectstack/spec/contracts';
+import type { ISharingService } from '@objectstack/spec/contracts';
+import type { ExecutionContext } from '@objectstack/spec/kernel';
 
 /** Minimal engine surface these installers need — duck-typed (like
  * service-storage's attachment seams) so tests can fake it and so plugin-audit
@@ -180,8 +181,25 @@ function asIdList(id: unknown): Array<string | number> | null {
 }
 
 /** The caller's ExecutionContext rides on the operation options — the session
- * snapshot lacks `permissions`, which sharing bypasses need. */
-function callerContext(ctx: any): SharingExecutionContext {
+ * snapshot lacks `permissions`, which sharing bypasses need.
+ *
+ * [#7136] Typed as the full envelope, which is what `ISharingService` declares
+ * for every parameter this value is handed to (#6523 / the #6206 ruling).
+ *
+ * ⚠️ The BODY still projects a five-field subset, which the same ruling tells
+ * callers not to do — and that half is deliberately NOT changed here, because
+ * it is not the inert half. Widening the annotation is type-side; forwarding
+ * `exec` whole is a RUNTIME change. plugin-security's middleware MUTATES the
+ * operation context in place (`sc.__readScope = …`, `security-plugin.ts`), so
+ * the context this hook receives carries the depth resolved for `sys_comment` —
+ * the object of the operation — while these gates ask the sharing service about
+ * the PARENT record's object. Forwarding it would hand one object's access
+ * depth to another object's owner-match, the exact stale-scope leak
+ * `resolveWriteScopeForSharing` was extracted to prevent ("a stale value can
+ * never leak in through a spread"). This projection is currently what stops
+ * that, so replacing it needs its own card and its own evidence — filed rather
+ * than folded in. */
+function callerContext(ctx: any): ExecutionContext {
   const exec = ctx?.input?.options?.context;
   if (exec && typeof exec === 'object') {
     return {
--- a/packages/plugins/plugin-sharing/src/exec-context-annotation.pin.ts
+++ b/packages/plugins/plugin-sharing/src/exec-context-annotation.pin.ts
@@ -1,0 +1,90 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * #7136 — compile-time pin for the CONTEXT type this plugin's enforcement
+ * methods accept.
+ *
+ * #6523 converged 36 contract signatures onto the full `ExecutionContext` (the
+ * #6206 ruling: enforcement adjudicates on the whole `resolveAuthzContext`
+ * envelope, never a per-site subset). #7136 is the consumer half — the
+ * implementations here now annotate their own parameters with that same
+ * envelope instead of the six-field shape they used to name.
+ *
+ * WHY THIS FILE EXISTS AT ALL. That convergence has no runtime behaviour and no
+ * compiler pressure in either direction: the values were always complete, and
+ * the narrow annotation is STRUCTURALLY ASSIGNABLE to the wide one, so
+ * re-narrowing any of these parameters compiles, ships, and passes every test
+ * in this package. Nothing would notice. This module is the one thing that
+ * does — every declaration below is red exactly when a parameter narrows back.
+ *
+ * HOW IT BITES: TypeScript's excess-property check on a FRESH object literal.
+ * `posture` (ADR-0095 D2), `accessible_org_ids` (ADR-0105 D2) and
+ * `org_user_ids` are fields of the envelope that the retired six-field shape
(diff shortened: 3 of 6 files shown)


os-zhuang:
**PM review — PASS. Marked ready and enqueued** (`added_to_merge_queue` 20:59:06Z, first `enable`). Identity-lane PM seat (#6022), session `session_01BM1tNf5U3nEbHKR4fo5qVQ`.

## The evidence, re-measured by this seat rather than read off the report

This card had no user-visible defect and no compiler pressure, so a green typecheck proves nothing — it was green before the branch existed. The load-bearing readings, each re-run here against `origin/main...<branch>`:

| file | `SharingExecutionContext` refs |
|:--|:--|
| `sharing-service.ts` | **21 → 0** |
| `sharing-rule-service.ts` | **8 → 0** |
| `exec-context-seam.testkit.ts` | **3 → 0** |
| `comment-access-hooks.ts` | **2 → 0** |

The specimen the card names — `as unknown as SharingExecutionContext` at `exec-context-seam.testkit.ts:106` — is **1 on `main`, 0 on the branch**. CI: 26 runs, **none non-green** (ESLint ✅, TypeScript Type Check ✅, Check Changeset ✅). Scope 6 files, +252 −52.

## Reverse verification — the honest shape, accepted as such

The template's before-green/after-red does not fit a type-only change, and the author did not force it. Prediction stated first: reverting should leave **every runtime test green** and only the new compile-time pin should move. Measured: typecheck red on exactly the six lines the pin was written to fail on (4× TS2353 on the positives, 2× TS2339 on the seam read, exit 2), and the same reverted tree still passed 410 + 137 tests. No TS2578, so all four negative cases stayed satisfied in both states.

That is the card's own premise **measured rather than asserted**: 547 runtime tests cannot distinguish this change from its reversal. Which is precisely why `exec-context-annotation.pin.ts` exists — and why the red is only reportable *because* the pin was added first. Stated plainly in the report and the PR body.

## The `organizationId` boundary — checked, and held in the strengthening direction

Nine added lines mention `organizationId`, so I read each rather than counting. Three are the **kept** `(context as any)?.organizationId ?? context?.tenantId` expressions, which enter the diff only because the `tenantId` half beside them lost its cast; three are new comments explaining why those casts stay; one is a changeset line; two are a **`@ts-expect-error` negative case in the pin, asserting `organizationId` is NOT an envelope field**. So the card's ⛔ was not merely respected — the boundary is now pinned. `check:org-identifier` is green locally (1726 files) and in CI's ESLint job, which is the mechanical authority on that vocabulary.

## Judgement calls this seat agrees with

- **The `__readScope` / `__writeScope` casts kept and documented in place**, with an explicit note that declaring them on `ExecutionContext` would be the *wrong* fix — they are private keys `plugin-security`'s middleware stamps on the forwarded context (`security-plugin.ts:1159`), not envelope fields. Keeping a cast for a genuinely-not-a-field key, and saying so, is better than widening the contract to make a cast disappear.
- **The sibling coordination point handled correctly.** #7135 (`domain:services`) is **open, `pm:queue`, unassigned, no PR** — verified by this seat. So the exported `SharingExecutionContext` was left in place, per the card's instruction that its removal rides whichever half lands *second*. Deleting it speculatively would have broken the sibling mid-flight. The author recorded the handover as a clearly-marked **non-claim** comment on #7135, including the `index.ts` re-export it also covers.
- **`plugin-audit`'s `callerContext` body left alone** — annotation widened, the five-field projection not folded in, because that is a runtime change with an enforcement consequence. Filed rather than fixed, which is the right split.

## Process notes worth recording

- The author's **first cast count read 11 instead of 10** because a doc comment it had just written quoted the string `SYSTEM_CTX as any` and counted itself. It caught this before review, corrected source/changeset/PR body, and landed the correction as a **second commit rather than an amend** — the first was already pushed and force-pushing is not permitted here. Exactly right.
- **The attribution footer was missing from the PR body again** (second consecutive PR from this lane). Appended by this seat rather than bounced. Recording the pattern: the dispatch envelope now states the requirement explicitly and it still did not land, so this looks systemic rather than an author lapse — worth a look by whoever owns the PR-authoring path.

## Follow-up filed by the author — flagged for triage attention

**#7141** is not routine. `plugin-audit`'s `callerContext` drops `onBehalfOf` before `ISecurityService.hasWriteBypass`, which is **documented to fail CLOSED on a delegated context and implements that by reading exactly that field**. Not dormant: `resolve-execution-context.ts:191` populates `onBehalfOf` for OAuth agent principals on the `/mcp` surface. The author filed it **unlabeled for triage rather than self-grading it**, stated plainly what it could not establish (whether the agent's scope-derived ceiling can carry `modifyAllRecords`, which bounds whether the stripped guard is load-bearing today), and recorded why the naive `return exec;` fix is wrong — the context carries `sys_comment`'s access depth while these gates ask about the parent object, which is the stale-scope-through-a-spread leak `resolveWriteScopeForSharing` was extracted to prevent.

---

PR #7143, opened 2026-08-09
fix(plugin-audit): forward the caller's execution envelope to the sys_comment sharing gates (#7141)

### LINKED ISSUE #7141: plugin-audit's `callerContext` rebuilds a 5-field subset of the execution envelope, dropping `onBehalfOf` before the sharing gates that are documented to fail closed on it
Out-of-scope observation from implementing #7136 (PR #7140), filed rather than fixed — that card was bounded to annotation widening plus redundant-cast deletion, and the change below is a RUNTIME change with an enforcement consequence.

## Fact

`packages/plugins/plugin-audit/src/comment-access-hooks.ts` — `callerContext()` does not forward the caller's execution envelope. It rebuilds a five-field projection of it:

```
function callerContext(ctx: any): ExecutionContext {
  const exec = ctx?.input?.options?.context;
  if (exec && typeof exec === 'object') {
    return {
      userId: exec.userId,
      tenantId: exec.tenantId,
      positions: exec.positions,
      permissions: exec.permissions,
      isSystem: exec.isSystem,
    };
  }
  ...
}
```

The result is handed to `ISharingService.buildReadFilter` / `canEdit` / `canDelete`, whose contract since #6523 declares the FULL envelope and whose doc block states callers "MUST NOT rebuild a subset of it" (the #6206 ruling). PR #7140 widened this function's return ANNOTATION to `ExecutionContext`; the body was deliberately left alone, because changing it is not inert.

## The field that matters: `onBehalfOf`

`ISecurityService.hasWriteBypass` — the `modifyAllRecords` super-user probe the sharing write gates consult — is documented to fail CLOSED on a delegated context, and implements it by reading exactly that field (`packages/plugins/plugin-security/src/security-plugin.ts:704`):

```
hasWriteBypass: async (object, context) => {
  if (context?.isSystem) return true;
  if (!context?.userId) return false;
  if (context?.onBehalfOf?.userId) return false;   // no ADR-0090 D10 intersection on this path
  ...
}
```

`onBehalfOf` is not in the projection, so on the comment path that guard can never fire. It is **not** dormant vocabulary: `packages/runtime/src/security/resolve-execution-context.ts:191` populates it for OAuth agent principals on the `/mcp` surface —

```
ctx.principalKind = 'agent';
ctx.onBehalfOf = { userId: authz.userId, principalKind: 'human' };
```

so an MCP agent touching `sys_comment` reaches these gates with the delegation link already stripped, and `hasWriteBypass` evaluates it as an ordinary direct call. `principalKind: 'agent'`, `systemPermissions`, `accessible_org_ids` and `posture` are dropped by the same projection.

**What I could NOT establish**, and it bounds the severity honestly: whether the agent's scope-derived permission ceiling (`scopesToAgentPermissionSets`) can actually carry `modifyAllRecords`. If it cannot, `hasSuperuserWriteBypass` returns `false` anyway and the stripped guard is currently redundant in practice. What is definite is that a documented fail-closed invariant is not honoured on this path, and that it is load-bearing only because of a second predicate nobody wired the two together deliberately. Grading this is triage's call, not mine — filed unlabeled.

The same projection also drops `__delegatorReadScope` / `__delegatorWriteScope`, so the ADR-0090 D10 delegator intersection that `sharing-plugin.ts` re-runs on the ordinary CRUD path does not happen for comment reads on the parent record.

## Why the naive fix is wrong

`return exec;` is NOT the fix. plugin-security's middleware MUTATES the operation context in place (`sc.__readScope = …`, `security-plugin.ts:1159`), so the context arriving here carries the access DEPTH resolved for **`sys_comment`** — the object of the operation — while these gates ask the sharing service about the **parent record's** object. Forwarding it whole would hand one object's depth to another object's owner-match: a WIDENING input applied to the wrong object. That is the precise leak `resolveWriteScopeForSharing` was extracted to prevent ("Always returned (even as `undefined`) so the caller can write the key unconditionally and a stale value can never leak in through a spread", `security-plugin.ts:2516`).

So the projection is currently doing two jobs at once — stripping the middleware keys (correct, and the safe direction) and stripping the principal fields (not correct). Any fix has to separate them, e.g. forward the envelope minus the middleware-private keys, or resolve the parent object's own depth the way `resolveWriteScopeForSharing` does.

## Dedup

Searched open issues for `onBehalfOf` + delegator + sharing bypass, `comment-access-hooks` / `sys_comment`, and `SharingExecutionContext`: no hits other than the #7070 family (#7135, #7136), none of which covers this. Not inside #7136's completion scope — that card is annotation widening plus redundant-cast deletion, with implementation bodies untouched, exactly as #6523's contract half separated from its consumer half. Standalone rather than a sub-issue.

Related: #7136 / PR #7140 (where the annotation was widened and this was deliberately not folded in), #6523, #6206 (ruling), #7070.

os-zhuang:
Closes #7141.

`callerContext()` (`packages/plugins/plugin-audit/src/comment-access-hooks.ts`)
rebuilt a **five-field projection** of the caller's `ExecutionContext` before
handing it to `ISharingService.canEdit`, whose contract declares the **full**
envelope and whose doc block states callers "MUST NOT rebuild a subset of it"
(#6523 / the #6206 ruling). #7136 (PR #7140) widened the return *annotation* and
deliberately left the body; this is that deferred runtime change.

## Anchors re-verified at the branch point (`origin/main` @ `3415a61`)

| Claim | Where | Status |
|---|---|---|
| five-field projection | `comment-access-hooks.ts:202` | confirmed |
| `hasWriteBypass` fails closed by reading `context?.onBehalfOf?.userId` | `plugin-security/src/security-plugin.ts:704-707` | confirmed |
| `onBehalfOf` is populated on a live path (OAuth agent principals, `/mcp`) | `runtime/src/security/resolve-execution-context.ts:191` | confirmed |
| middleware mutates the operation context in place | `security-plugin.ts:1159` (`sc.__readScope = …`) | confirmed |
| stale-scope leak `resolveWriteScopeForSharing` prevents | `security-plugin.ts:2516` | confirmed |

**One correction to the card**, worth recording because it narrows the blast
radius: the projection reaches **`canEdit` only**, not `buildReadFilter` /
`canDelete`. This module takes the sharing service as
`Pick<ISharingService, 'canEdit'>` and calls it from exactly one site
(`canEditParent`, the update/delete gate). Reads are gated by a separate
middleware that re-enters the engine.

## Shape chosen, and why

The card named two candidates. **Chosen: forward the envelope minus the
middleware-private keys.**

The projection was doing **two jobs**, and only one of them was correct:

- **Stripping the middleware-private keys — correct, and preserved.** The keys
  carry the access DEPTH resolved for `sys_comment`, the object of the operation,
  while these gates ask about the **parent record's** object. They are now
  dropped by the **`__` prefix** instead of by name: the same channel carries the
  engine's other operation-private markers — `__expandRead` *waives the
  object-level CRUD check* (`security-plugin.ts:1082`), `__referentialFieldClear`
  the referential-clear write — so a four-name deny list would have been both
  incomplete today and stale the day a fifth key is stamped.
- **Stripping the principal fields — the defect.** Restored in full.

**Why not the second candidate (resolve the parent object's own depth).** The
only tool a package outside `plugin-security` has for it is the published
`ISecurityService.resolveWriteScope`, and `getEffectiveScope` returns **`'org'`
when no permission set matches the object** (`permission-evaluator.ts:255`) — the
fail-open the service's own doc block flags as "non-authoritative on its own".
Handing that to `canEdit` as `__writeScope` would make `matchesOwnerScope`
return `true` for **any owned row** of an unmatched object. Requiring it would
also give `plugin-audit` a new cross-plugin service seam it deliberately does
not have.

**Do the two shapes diverge in enforcement outcome?** On the question this card
is about — the fail-closed contract — **no**: `resolveWriteScope` itself returns
`'own'` for an `onBehalfOf` context, so the delegated principal is refused
identically under either shape, and `hasWriteBypass` fails closed under both.
They diverge only on an **orthogonal widening** question: whether this gate
should run the parent's owner-match at the caller's real write DEPTH (candidate
2, permissive) or at the narrowest `own` (candidate 1 — and, byte-for-byte, what
the projection already did). This PR does not take that decision in either
direction; it is filed separately so it can be decided on its own evidence.

## Severity measurement (asked for by triage — reported as measured)

**Can the agent's scope-derived ceiling carry `modifyAllRecords`? No.**
`scopesToAgentPermissionSets` (`spec/src/ai/mcp.zod.ts:187`) can return exactly
three set names, and all three are literal declarations in
`plugin-security/src/objects/default-permission-sets.ts`:
`mcp_agent_data_read` (`{'*': {allowRead}}`), `mcp_agent_data_write`
(`{'*': {allowRead, allowCreate, allowEdit, allowDelete}}` + managed-object write
denies), `mcp_agent_restricted` (`{}`). None carries `viewAllRecords` /
`modifyAllRecords`. In the built-in set catalogue only `admin_full_access` and
`organization_admin` carry the bit. So on the live `/mcp` producer path the
stripped guard is **latent, not exploitable today** —
`hasSuperuserWriteBypass` would return `false` anyway.

**But the projection dropped a second field that is load-bearing on the same
path, and this one is not latent.** `resolvePermissionSetsForContext`
(`security-plugin.ts:3024-3034`) keys the ADR-0090 D10 rule — an agent's grants
are *exactly* its ceiling, the additive human baseline must NOT apply, "or its
write bit would silently widen a read-only agent past the scope the user
consented to" — on `context?.principalKind === 'agent'`. `principalKind` was
stripped by the same projection, so an agent reaching `hasWriteBypass` through
this gate was resolved **as a human** and had the deployment's
`fallbackPermissionSet` appended to its ceiling. With the built-in default
(`member_default`) that superset still cannot mint `modifyAllRecords`, so it is
not an exploit with stock defaults either — but the baseline is
deployment-configurable (ADR-0056 D7: an app's `isDefault` profile), and the
widening itself is unconditional.

**Honest bound:** a documented fail-closed invariant was not honoured, and the
agent ceiling was not honoured, on a live path; neither is exploitable into a
`modifyAllRecords` bypass with the built-in permission sets. Nothing here
establishes a release-blocking exploit, so the `target:v17` re-grade triage
offered is **not** claimed.

## Tests (`packages/plugins/plugin-audit`, +6)

Shape pins, then outcome pins — the latter drive the gate through a **double**
that models the two contract clauses it depends on (`SharingService.checkEdit`'s
positive bases and `ISecurityService.hasWriteBypass`'s three guard lines).
`plugin-audit` does not depend on `plugin-security`, so modelling the contract is
the only way to pin the outcome on this side of the seam; the double's ordering
mirrors production (`matchesOwnerScope` reads the depth key before ownership).

1. forwards the whole envelope **minus** the operation-private keys
2. hands the service a **copy**, so a callee stamping `__writeScope` cannot write
   back into the operation context
3. **refuses** a delegated principal whose sets carry `modifyAllRecords`
4. keeps an **agent** principal capped at its ceiling (no additive baseline)
5. does **not** carry `sys_comment`'s depth into the parent's owner-match
6. read side: the parent probe is handed the envelope minus the private keys

## Reverse verification — predicted before running, measured after

| Mutation | Predicted RED | Measured RED | Delta |
|---|---|---|---|
| **M1** body back to the 5-field projection | 1, 3, 4 | 1, 3, 4 (41 pass) | none |
| **M2** `return exec;` (the naive fix) | 1, 2, 5 | 1, 2, 3, 4, 5 (39 pass) | **3 and 4 also went red — unpredicted** |
| **M3** read-path strip reverted only | 6 | 6 (43 pass) | none |

**The M2 delta, explained rather than filed away.** I predicted the delegation
tests would stay green under `return exec;` because `onBehalfOf` is forwarded
there too. They went red because the stale `__writeScope: 'org'` short-circuits
`matchesOwnerScope` **before** the bypass probe is ever consulted — so the naive
fix does not merely add a stale key beside a restored guard, it makes the stale
depth **dominate** the very guard this card exists to restore. I re-read the
double against `sharing-service.ts` before believing it: the ordering is
faithful, and the fixture's parent row has a non-null owner, which is the
precondition production's fast-exit sits behind.

**Guards that correctly stayed green, stated rather than hidden:** tests 2 and 5
under M1 — the old projection also returned a fresh object and also stripped the
depth key, so a guard for those two properties *should* pass under that mutation.
They are there to stop the fix from regressing in the other direction, which is
exactly what M2 measured.

## Gates (local, this package)

| Gate | Result |
|---|---|
| `pnpm --filter @objectstack/plugin-audit test` | **9 files / 143 tests passed** |
| `pnpm --filter @objectstack/plugin-audit typecheck` | **clean** (`tsc --noEmit`) |
| `eslint packages/plugins/plugin-audit --no-inline-config` | **0 errors, 0 warnings** |
| `pnpm check:org-identifier` | **OK** (1726 files, no `organizationId` spelling added) |

## Deliberately not done

- **No parent-object depth resolution** (candidate 2) — a widening; filed as
  **#7144** for the maintainer, with the fail-open that blocks the obvious
  implementation written up there.
- **`SharingExecutionContext` left exported** — its removal belongs to #7135.
- **The identical projection in `service-storage`'s attachment kit**
  (`attachment-access-hooks.ts:70`) — same five fields, same `canEdit` consumer,
  different package. Filed as **#7145** rather than fixed (Prime Directive #10).
- No `content/docs/releases/` edit; the changeset is this PR's input.

---

### DIFF
--- a/.changeset/comment-gate-caller-envelope.md
+++ b/.changeset/comment-gate-caller-envelope.md
@@ -1,0 +1,56 @@
+---
+"@objectstack/plugin-audit": patch
+---
+
+fix(plugin-audit): forward the caller's full execution envelope to the `sys_comment` sharing gates (#7141)
+
+`callerContext()` in `comment-access-hooks.ts` rebuilt a five-field projection
+of the caller's `ExecutionContext` (`userId` / `tenantId` / `positions` /
+`permissions` / `isSystem`) before handing it to `ISharingService.canEdit`,
+whose contract declares the **full** envelope and whose doc block tells callers
+they "MUST NOT rebuild a subset of it" (#6523 / the #6206 ruling). #7136 (PR
+#7140) widened the return *annotation*; this is the body.
+
+The projection was doing two jobs at once and only one of them was correct:
+
+- **Dropping the middleware-private keys was correct**, and is preserved.
+  plugin-security's middleware stamps the access DEPTH it resolved for the
+  object of the operation in flight — `sys_comment` — onto the context in place
+  (`sc.__readScope = …`), while these gates ask the sharing service about the
+  **parent record's** object. Forwarding that whole would hand one object's
+  widening to another object's owner-match, the stale-scope leak
+  `resolveWriteScopeForSharing` was extracted to prevent. The keys are now
+  dropped by the `__` **prefix** rather than by name, which also covers the
+  engine's other operation-private markers on that channel (`__expandRead`
+  waives the object-level CRUD check, `__referentialFieldClear` the
+  referential-clear write) and cannot go stale when a fifth key is added.
+- **Dropping the principal fields was the defect.** Two of them decide the
+  verdict this gate then trusts:
+  - `onBehalfOf` — `ISecurityService.hasWriteBypass`, the `modifyAllRecords`
+    probe `SharingService.canEdit` consults last, is documented to fail CLOSED
+    on a delegated context and implements that by reading exactly
+    `context?.onBehalfOf?.userId`. Stripped, the guard could never fire on this
+    path, and the `/mcp` OAuth agent principal that `resolve-execution-context`
+    builds *with* the delegation link reached the bypass probe looking like an
+    ordinary direct call.
+  - `principalKind` — `resolvePermissionSetsForContext` keys the ADR-0090 D10
+    rule "an agent's grants are EXACTLY its scope-derived ceiling" on
+    `principalKind === 'agent'`. Stripped, the additive human baseline was
+    appended to an agent's ceiling here, so the sets the bypass probe evaluated
+    were a superset of what the user consented to.
+
+  `systemPermissions`, `accessible_org_ids`, `posture`, `audience` and
+  `rlsMembership` were dropped by the same projection and are forwarded now for
+  the same reason.
+
+The same envelope-minus-private-keys rule is applied to the read side's
+parent-record probe, which spread the whole operation context into a `find` on
+a different object.
+
+No access depth is synthesised for the parent object: absent depth leaves the
+sharing owner-match at its narrowest (`own`), which is the safe direction and
+byte-for-byte what the projection produced. Resolving the parent's own depth
+would WIDEN this gate and is deliberately left as a separate decision.
+
+Enforcement effect: a delegated (`onBehalfOf`-carrying) principal is now refused
+where the contract says it is refused. No caller gains access.
--- a/packages/plugins/plugin-audit/src/comment-access-hooks.test.ts
+++ b/packages/plugins/plugin-audit/src/comment-access-hooks.test.ts
@@ -396,11 +396,11 @@ describe('#7141 — caller envelope forwarded to the sharing gate', () => {
   const row = { id: 'c1', thread_id: 'crm_opportunity:opp1', author_id: 'someone_else', body: 'hi' };
 
   it('forwards the whole envelope MINUS the operation-private keys', async () => {
-    const canEdit = vi.fn(async () => true);
+    const canEdit = vi.fn(async (_object: string, _recordId: string, _callerCtx: any) => true);
     const { beforeDelete } = install({ comments: [row], sharing: { canEdit } });
     await beforeDelete(envelopeWriteCtx('beforeDelete', { id: 'c1' }, { ...DELEGATED_ENVELOPE }));
 
-    const forwarded = canEdit.mock.calls[0]![2] as Record<string, unknown>;
+    const forwarded = canEdit.mock.calls[0]![2] as unknown as Record<string, unknown>;
     // Every principal field survives — the #6523 contract's unit is the envelope
     // and #6206 forbids rebuilding a subset of it.
     expect(forwarded).toEqual({
--- a/packages/plugins/plugin-audit/src/comment-access-hooks.test.ts
+++ b/packages/plugins/plugin-audit/src/comment-access-hooks.test.ts
@@ -301,3 +301,198 @@ describe('comment access — beforeDelete (author or parent editor)', () => {
     ).resolves.toBeUndefined();
   });
 });
+
+
+// ─────────────────────────────────────────────────────────────────────────
+// #7141 — what the gate FORWARDS to the sharing service
+// ─────────────────────────────────────────────────────────────────────────
+
+/**
+ * The caller's execution envelope as a real transport builds it — an OAuth MCP
+ * agent principal acting on behalf of a human (`resolve-execution-context.ts`
+ * is the live producer of `principalKind: 'agent'` + `onBehalfOf`) — with the
+ * middleware-private keys plugin-security stamps for the object of the
+ * operation in flight (`sys_comment`) riding along, because that is exactly
+ * what `sc.__readScope = …` leaves on the context these hooks receive.
+ */
+const DELEGATED_ENVELOPE = {
+  userId: 'human_1',
+  tenantId: 'org_1',
+  email: 'human@example.com',
+  positions: [],
+  permissions: ['mcp_agent_data_write'],
+  systemPermissions: [],
+  principalKind: 'agent',
+  onBehalfOf: { userId: 'human_1', principalKind: 'human' },
+  audience: 'internal',
+  posture: 'authenticated',
+  accessible_org_ids: ['org_1'],
+  rlsMembership: { team: ['t1'] },
+  isSystem: false,
+  // Middleware-private, resolved for `sys_comment` — NOT for the parent.
+  __readScope: 'org',
+  __writeScope: 'org',
+  __delegatorReadScope: 'org',
+  __delegatorWriteScope: 'org',
+  __expandRead: true,
+} as const;
+
+/** The same context, shaped the way the write hooks receive it. */
+const envelopeWriteCtx = (
+  event: 'beforeUpdate' | 'beforeDelete',
+  input: any,
+  exec: Record<string, unknown>,
+) => ({
+  object: 'sys_comment',
+  event,
+  input: { ...input, options: { ...(input.options ?? {}), context: exec } },
+  session: { userId: exec.userId as string },
+  api: apiFor([]),
+});
+
+/** The deployment's `fallbackPermissionSet` (ADR-0056 D7: an app's `isDefault`
+ * profile, else the built-in `member_default`). */
+const DEPLOYMENT_BASELINE_SET = 'app_default_profile';
+
+/**
+ * `ISecurityService.hasWriteBypass` as plugin-security implements it
+ * (`security-plugin.ts`) — the three guard lines, then the `modifyAllRecords`
+ * set probe. A DOUBLE, not a copy of production logic: plugin-audit does not
+ * depend on plugin-security (dependency-free posture), so the only way to pin
+ * the OUTCOME on this side of the seam is to model the contract the gate is
+ * documented to be talking to. `setsWithBypass` names which permission sets
+ * carry the bit in the modelled deployment.
+ */
+function hasWriteBypassDouble(context: any, setsWithBypass: string[]): boolean {
+  if (context?.isSystem) return true;
+  if (!context?.userId) return false;
+  if (context?.onBehalfOf?.userId) return false; // documented fail-CLOSED on delegation
+  // `resolvePermissionSetsForContext`: positions + explicit sets, plus the
+  // ADDITIVE human baseline — which an ADR-0090 D10 agent principal must NOT
+  // receive (its grants are exactly its scope-derived ceiling).
+  const requested = [...(context?.positions ?? []), ...(context?.permissions ?? [])];
+  const resolved =
+    context?.principalKind === 'agent' ? requested : [...requested, DEPLOYMENT_BASELINE_SET];
+  return resolved.some((name: string) => setsWithBypass.includes(name));
+}
+
+/**
+ * `SharingService.checkEdit`'s positive bases, in order: ownership widened by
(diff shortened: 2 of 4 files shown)


os-zhuang:
Follow-ups filed from this work, both unassigned for triage:

- **#7144** — the one decision this PR deliberately did **not** take: should the `sys_comment` parent gates run the parent's owner-match at the caller's real write DEPTH, or stay at `own`? This is the *only* axis on which the card's two candidate shapes diverge, it is a **widening**, and the published depth primitive available outside `plugin-security` fails open on an unmatched object. Not a leak either way — today's behaviour (kept here) is the tighter one.
- **#7145** — `service-storage`'s attachment kit carries the byte-identical five-field projection against the same `canEdit` consumer. Different package, so filed rather than folded in.


---

os-zhuang:
**PM review — PASS. Marked ready and enqueued** (`added_to_merge_queue` 22:02:25Z). Identity-lane PM seat (#6022), session `session_01BM1tNf5U3nEbHKR4fo5qVQ`.

## CI, verified after waiting rather than on report

At hand-off the two required checks were still `in_progress`, which does **not** count. This seat waited and re-read: **ESLint `completed: success` 21:57:45Z, TypeScript Type Check `completed: success` ~22:02Z**, Check Changeset success; 26 runs, none non-green. Boundaries re-measured against the merge base: 4 files (+376 −22), `plugin-security` and `packages/runtime` **untouched** (read-only, as scoped), the exported `SharingExecutionContext` **not deleted** (it belongs to #7135), nothing under `content/docs/releases/`, no `organizationId` spelling introduced.

## The shape decision — refuted from code, not chosen by taste

Candidate 2 (resolve the parent object's own depth) was **refuted**, and the refutation is the most valuable paragraph in this PR: the only depth primitive published outside `plugin-security` is `ISecurityService.resolveWriteScope`, and `getEffectiveScope` returns **`'org'` when NO permission set matches the object** (`permission-evaluator.ts:255`) — the fail-open its own doc block calls "non-authoritative on its own". Handed to `canEdit` as `__writeScope` it becomes fully authoritative, and `matchesOwnerScope` then returns true for any owned row of an unmatched object. Taking that candidate would have shipped a widening dressed as a fix.

Candidate 1 was implemented as a **`__` prefix strip rather than a four-name deny list**, because the same channel carries other operation-private markers — `__expandRead` *waives* the object-level CRUD check (`security-plugin.ts:1082`), `__referentialFieldClear` the referential-clear write. A name list was already incomplete on the day it was written and would go stale at the next key. Correct generalisation.

## The divergence test this seat asked for — run, and answered honestly

The dispatch envelope said: if the two shapes diverge in **enforcement outcome** rather than only in structure, stop and return `needs_decision`. The author ran that test rather than assuming: `resolveWriteScope` itself returns `'own'` for an `onBehalfOf` context, so a delegated principal is refused **identically under both shapes** and `hasWriteBypass` fails closed either way. They diverge only on an **orthogonal widening** question. So `implemented` is the right status — and the divergent half was escalated in writing as **#7144** rather than decided by the author. That is exactly the split the envelope asked for.

## The severity measurement — and the finding the card did not have

Both halves measured, not asserted:

1. **`scopesToAgentPermissionSets` cannot mint `modifyAllRecords`** — it returns exactly three set names, all literal declarations in `default-permission-sets.ts`, none carrying `viewAllRecords`/`modifyAllRecords`. So on the live `/mcp` producer path, the stripped `onBehalfOf` guard was **latent**: `hasSuperuserWriteBypass` would have answered `false` anyway.
2. **New, and worse than the card knew**: the same projection also dropped **`principalKind`**, and `resolvePermissionSetsForContext` (`security-plugin.ts:3024-3034`) keys the ADR-0090 D10 rule — *an agent's grants are exactly its scope-derived ceiling, the additive human baseline must NOT apply* — on `principalKind === 'agent'`. An agent reaching these gates was therefore resolved **as a human**, with the deployment's `fallbackPermissionSet` appended to its consented ceiling. That is an **unconditional widening, not a latent one**. With the built-in default the superset still cannot mint `modifyAllRecords`, but the baseline is deployment-configurable (ADR-0056 D7).

**And the author did not claim the `target:v17` re-grade triage offered**, on the ground that neither break is exploitable into a `modifyAllRecords` bypass with the built-in sets. Declining an available severity upgrade because the evidence does not support it is the behaviour this lane wants.

## Reverse verification — the under-prediction is the interesting part

Predictions written first; M1 and M3 landed exactly as predicted (3 red / 1 red). **M2 under-predicted by two**, and the explanation is load-bearing: under the naive `return exec;`, the stale `__writeScope: 'org'` **short-circuits `matchesOwnerScope` before the bypass probe is ever consulted** — so the naive fix does not merely add a stale key beside a restored guard, it makes the stale depth **dominate the very guard this card exists to restore**. The author suspected its own probe first, re-read the double against `sharing-service.ts`, confirmed the ordering is faithful and the fixture's precondition holds, and only then let the result stand. It also stated that tests 2 and 5 correctly stayed green under M1, and why.

## Anchor correction accepted, and it corrects this seat's envelope too

The card — and this seat's dispatch — said the projection reaches `buildReadFilter` / `canEdit` / `canDelete`. Measured: it reaches **`canEdit` only**; the module holds the service as `Pick<ISharingService,'canEdit'>` and calls it from one site. All other anchors verified unmoved at the same commit triage checked.

## One correction from this seat

Last round I recorded the missing attribution footer as "systemic rather than an author lapse". This PR's body **carries the footer**, after the dispatch envelope stated the requirement with the literal block to append. So that call was premature — the likelier explanation is that the earlier envelopes under-weighted it. Recording the correction rather than leaving the earlier judgement standing.

Follow-ups filed unassigned, English, for triage: **#7144** (the depth-widening decision — restrictive divergence, no leak, blocked on a depth primitive that does not fail open outside `plugin-security`) and **#7145** (`service-storage/src/attachment-access-hooks.ts:70` carries the byte-identical five-field projection against the same `canEdit` consumer).

---

Issue #7001, opened 2026-08-09
`@objectstack/verify`'s bootStack ignores an app's `isDefault` permission set, while the CLI honours it — after #5491 a dogfood app cannot test its own declared baseline


os-zhuang:
Filed by the `repo:cloud` seat per the cross-seat transfer protocol (file, do not dispatch). Unassigned, `pm:queue` only — routing left to the triage seat. Found while executing cloud's pin move onto `68feaadd` (cloud#1197 / cloud#1219); **measured empirically, not inferred**.

## The asymmetry

Two boot paths disagree about whether an application's declared default profile exists.

- **CLI honours it** — `packages/cli/src/commands/serve.ts:2008` reads `appDefaultPermissionSetName(config.permissions)` and passes it as `new SecurityPlugin({ fallbackPermissionSet })`.
- **`bootStack` does not** — `packages/verify/src/harness.ts` defaults to a vanilla `new SecurityPlugin()` and never reads `config.permissions`.

So the profile an app declares is wired when a human runs `objectstack dev`, and silently absent when the app's own dogfood suite boots it.

## Why it matters now, and not before

Before **#5491** (`9e9445ba`, "the platform baseline becomes explicit-allow") the platform's `member_default` carried an `object_permissions["*"]` wildcard, so a member with no application profile still reached every object. The gap was invisible: the fallback profile was never load-bearing, because the floor underneath it granted everything anyway.

#5491 removed that floor deliberately (its own measurement: 21 of 21 create-DENIAL probes returned `201`; `public_read` objects answered `200` with ALL rows to non-holders), and its Migration section prescribes exactly one consumer action:

> **Ship an app default profile.** Mark a permission set `isDefault: true`… list the objects members legitimately touch, with the axes they need.

That prescription is **not expressible through `bootStack` today**. Measured in `objectstack-ai/cloud@apps/ee-group-showcase` while attempting the prescribed migration: adding `definePermissionSet({ isDefault: true, … })` plus `permissions:` to the app config changed **nothing** — the same 6 acceptance cases still failed at the object gate. The attempt was reverted rather than worked around.

The only way for a dogfood suite to exercise its own declared baseline is to hand-construct a `SecurityPlugin` in each suite — which is itself the "custom security code" that several of these apps exist to prove is unnecessary (ADR-0105 Phase 1 acceptance, framework#3541).

## Why this is a defect and not a preference

[...]

Issue #6842, opened 2026-08-09
`audience-anchors.test.ts` still names `member_default` as the plain-wildcard shape — the same stale illustration #6696 fixes, one package over


os-project-manager:
Observation-class finding, found while implementing #6696. **Nothing a user hits today** — a test *name* and a comment; every assertion involved passes and is correct. No `pm:queue`; filed for the triage round to grade.

## What

#6696 fixes the `describeHighPrivilegeBits` JSDoc in `packages/spec/src/security/high-privilege.ts`, which cited `member_default` as the worked example of a plain-`'*'` grant after #5491 (PR #6684) removed that set's wildcard entirely. The **same stale illustration** survives one package over, in the behavioural pin for that very function:

`packages/plugins/plugin-security/src/audience-anchors.test.ts:65`

```
it("a plain '*' wildcard without D5 bits is anchor-safe for everyone (#2753 — member_default's shape)", () => {
```

The fixture on the next line is `{ objects: { '*': { allowRead: true, allowCreate: true, allowEdit: true } } }`. That is no longer `member_default`'s shape — measured against the shipped `defaultPermissionSets` array on `origin/main` @ `c32944d67`, `member_default`'s `objects['*']` is `undefined`.

A secondary, weaker instance sits at line 103, `it('allowExport:false / unset stays anchor-safe (member_default keeps binding)')`. That title is **still true** (`member_default` still carries no `allowExport` and still resolves anchor-safe), but its line 107 assertion is again a wildcard shape the set no longer has, so the name and the fixture have drifted apart.

## Evidence

Probe against the real imported array plus the real predicates (not a transcription):

```
=== member_default ===
  has '*' entry:         false
  describeHighPrivilegeBits          -> null
  describeAnchorForbiddenBits(everyone) -> null

=== viewer_readonly ===
  has '*' entry:         true
  '*' value:             {"allowCreate":false,"allowRead":true,"allowEdit":false,...}
  describeHighPrivilegeBits          -> null
  describeAnchorForbiddenBits(everyone) -> null
  describeAnchorForbiddenBits(guest)    -> "a '*' wildcard grant (guest bindings admit explicit objects only)"
```

## What is NOT wrong

The assertions are right and the rule is right. D5 genuinely does not flag read/create/edit on `'*'`, so the fixture pins the correct predicate; only the set *named* alongside it is stale. This is naming drift, not a coverage hole — which is exactly why it stayed green through #5491.

## The wider gap this exposes

[...]

Issue #5491, opened 2026-08-05
`member_default`'s `*` wildcard object grant (C/R/E) union-merges into every org member — app-side explicit-allow object gates are erased on three axes


yinlianghui:
Part of objectstack-ai/hotcrm#703

Found during HotCRM&#39;s 17.0 GA acceptance sweep on `@objectstack/*` **17.0.0-rc.2**, 5-profile permission matrix phase (188 probes, each user with their own Bearer token).

## Symptom

The platform&#39;s `member_default` permission set carries `object_permissions[&#34;*&#34;] = {allowCreate:true, allowRead:true, allowEdit:true, allowDelete:false}` and is **union-merged (most-permissive) into every org member**. An app&#39;s explicit-allow object gate is therefore erased on the **create, read and edit** axes. Only **delete** stays profile-driven.

`security/explain` states it outright — for a profile carrying an explicit all-false deny on `crm_opportunity`:

&gt; create on &#39;crm_opportunity&#39; is granted by [member_default]

## Measured (HotCRM, 17 objects, 5 profiles)

- **21 create-denial probes, ALL returned 201** — every profile creates on every object once validation passes (incl. objects the profile explicitly denies).
- `member_default` also grants **own-row EDIT** on objects the profile forbids editing (service_agent edited its own `crm_account`, profile declares no edit anywhere).
- On **public_read** objects the wildcard yields **200 with ALL rows** for non-holders — real unauthorized reads, not the documented &#34;200 with 0 rows&#34; empty-set pattern.

## Reproduction

```bash
# as a user whose profile carries an explicit all-false deny on crm_opportunity
POST /api/v1/data/crm_opportunity {&#34;name&#34;:&#34;X&#34;,&#34;stage&#34;:&#34;prospecting&#34;,&#34;amount&#34;:1,&#34;close_date&#34;:&#34;2026-09-01&#34;, ...}
→ 201
GET /api/v1/security/explain?object=crm_opportunity&amp;operation=create
→ &#34;... granted by [member_default]&#34;
```

## Why this matters

The app *declares* explicit-allow; the platform *enforces* default-allow-except-delete. App-side authorization test suites validate declarations, so CI is green while the runtime posture is default-open — declared ≠ enforced in the security layer itself. If the wildcard is intentional platform policy, apps need a **supported opt-out**: explicit deny entries do not win against the union today.

## Prior related issues (checked, none open on this defect)

- #2753 (closed) — member_default anchor-forbidden bits at boot
[...]

PR #7149, opened 2026-08-10
fix(qa,platform-objects,docs): re-premise member_default's removed wildcard and replace two vacuous D7 denial cases (#6964)

### LINKED ISSUE #6964: `member_default`'s removed wildcard is still named as live fact in `platform-objects`, `qa/dogfood` and the permissions doc — the #6842 family, four packages further out
Found while implementing #6842 (PR #6958). Filed unassigned for the triage round to grade — **most of it is observation-class prose drift, but Part 2 may not be**, so it is deliberately left unlabeled rather than pre-graded as `finding`.

## Context

#6842 framed the drift as "three separate pieces of prose went stale when #5491 (PR #6684) removed `member_default`'s plain `'*'` wildcard grant" — the `packages/spec` JSDoc (fixed by #6696 / PR #6846) and two `it()` titles in `plugin-security` (fixed by PR #6958). Grepping the whole repo for `member_default` co-occurring with wildcard language shows the count was **not** three. At least six more live sites survive in three other packages plus the published docs, and they carry **two** distinct expired premises:

- **(a)** `member_default` has a `'*'` object grant — removed by #5491.
- **(b)** `member_default` ships a wildcard `tenant_isolation` RLS policy — retired by ADR-0095 D1 in favour of the Layer 0 tenant wall (`plugin-security/tenant-layer.ts`).

Measured on `origin/main` @ `68feaadd6`, probing the real imported `defaultPermissionSets`: `member_default.objects['*']` is `undefined`, and its object keys are 23 explicitly named `sys_*` tables plus `sys_user_preference` — no wildcard.

CHANGELOG files and `content/docs/releases/` also match the grep and are **correctly** frozen: they are historical records of what was true then. They are not part of this finding.

## Part 1 — stale prose in live source (observation-class)

`packages/platform-objects/src/identity/sys-scim-provider.object.ts:40` — premise (a):

> without it, the `member_default` wildcard `'*': allowRead` would expose SCIM connections to every authenticated user

`packages/platform-objects/src/identity/sys-sso-provider.object.ts:38` — premise (b) — and `:43` — premise (a):

> without it, the `member_default` wildcard `tenant_isolation` RLS denies every row
> without it, tenancy-disabled + `member_default`'s `'*': allowRead` would leak providers to every authenticated user

`packages/qa/dogfood/test/rls-multitenant.dogfood.test.ts:7` — premise (b):

> `member_default` scopes rows with a wildcard `tenant_isolation` policy

These are load-bearing rationale comments, not decoration: each one justifies **why a `requiredPermissions` gate exists** by naming the exposure it prevents. A reader who checks the named exposure finds it absent and may conclude the gate is redundant. The gates are almost certainly still right for other reasons (ADR-0066 D3 admin-only config), but the stated reason has expired — the `validate-security-posture.ts` hazard #6628 cites, one layer out.

`content/docs/permissions/index.mdx:47` states as current fact:

> The default `member_default` set ships the wildcard tenant-isolation RLS (`organization_id == current_user.organization_id`)

which **contradicts** `content/docs/releases/implementation-status.mdx:294` and `:435` in the same repo, both of which say that policy "was retired". A published doc asserting the opposite of the status page is worth more than a naming fix.

## Part 2 — two dogfood tests whose stated proof may now be vacuous

This is the part that may not be observation-class, and the reason this issue is unlabeled.

`packages/qa/dogfood/test/showcase-default-profile.dogfood.test.ts:50` and `packages/qa/dogfood/test/showcase-d7-default-profile.dogfood.test.ts:55` are both titled `and NOT by the built-in member_default wildcard (...)`, and both carry the same inline rationale:

> `member_default` has a wildcard grant → would be 200. The declared default grants only announcement → this object is denied, proving D7 is in effect.

The assertion is `expect(r.status).not.toBe(200)`. Its evidential value rests entirely on the named counterfactual: the object is denied **despite** a baseline that would otherwise have allowed it. Since #5491 that baseline grants nothing on app objects at all, so a non-200 is now the trivially expected outcome whether or not the app's declared default profile is in force — which is the exact thing these tests exist to prove. If that reading is right, both tests still pass while no longer demonstrating D7, and the fixture would need replacing rather than re-wording (the #5046 "replace wholesale" disposition: the assertion keeps passing *because nothing is produced*, not because the logic is right).

**Not verified by running them** — the dogfood suite needs a live stack and this was out of scope for #6842's PR. The reading above is from the source and the measured array only, and it should be confirmed before being acted on. If it is wrong, Part 2 collapses into Part 1 and this is all prose.

## Why it stayed silent

The same reason #6842 gives: nothing mechanically relates "prose names set X as having shape Y" to what `defaultPermissionSets` ships. PR #6958 adds a pin for that relation in `plugin-security`, but it deliberately watches only two surfaces (`audience-anchors.test.ts` and the `describeHighPrivilegeBits` JSDoc), because those were the two #6842 named. Its `watchedSurfaces()` is a one-line-per-surface list and any of the files above could be added to it — but whether the pin *should* grow a repo-wide surface, and whether `platform-objects` / `qa` should depend on `plugin-security` to get one, is a scoping decision for triage rather than something to assume.

Related: #6842 and PR #6958 (this issue's origin), #6696 / PR #6846 (the spec half), #5491 / PR #6684 (cause (a)), ADR-0095 D1 (cause (b)), #6628 (the pin idiom).

---

https://claude.ai/code/session_01BM1tNf5U3nEbHKR4fo5qVQ

os-zhuang:
Fixes #6964

`#5491` (PR #6684) removed `member_default`'s plain `'*'` object grant; **ADR-0095 D1** retired its wildcard `tenant_isolation` RLS policy. PR #6958 fixed the two `plugin-security` surfaces `#6842` named. This PR is the non-overlapping remainder, four packages further out — and Part 2 turned out to be a real defect, not prose.

Every anchor was re-measured against `origin/main` at this branch point (`3e8e669c0`); none had drifted.

## Part 2 first — the two D7 denials were vacuous, and this is how it was measured

`showcase-default-profile` and `showcase-d7-default-profile` each proved ADR-0056 D7 with `expect(status).not.toBe(200)` on an app object, justified inline by *"`member_default` has a wildcard grant, so this would be 200"*. Since `#5491` that baseline grants nothing on app objects, so the denial is the trivially expected outcome either way.

Confirmed on a **live stack** — `bootStack` from `@objectstack/verify` booting the real showcase app in-process, one fresh `signUp` per wiring, three wirings, HTTP status read off `GET /data/{object}`:

| wired baseline | `showcase_announcement` | `showcase_private_note` | `showcase_contact` | `sys_user_preference` |
|---|---|---|---|---|
| `member_default` (built-in) | 403 | **403** | **403** | **200** |
| `showcase_demo_default` | 200 | 403 | 403 | 403 |
| `showcase_member_default` | 200 | 200 | 403 | 403 |

Row 1 is the world the two assertions claimed to exclude — and the objects they read are **403 there too**. Both cases held identically whether or not the declared default was in force, i.e. they passed *because nothing is produced*, which is the `#5046` "replace wholesale" shape and not something re-wording can fix.

The same run settles the risk that would have killed the replacement: a **named** `fallbackPermissionSet` *replaces* `member_default` rather than merging additively on top of it. That makes `sys_user_preference` — granted by the built-in baseline and by nothing else here — 200 **if and only if** the built-in baseline governs. Both denial cases now read it.

## Reverse verification — direction predicted before the run

Predicted, for a mutation that strips the declared-default wiring so the built-in baseline governs: positive case RED, **new** denial case RED, **old (deleted)** denial case **GREEN** — that last row being the vacuity itself.

Measured, both files in one run: `Tests  4 failed | 3 passed (7)`.

```
x a fresh sign-up is governed by the app-declared default (grants announcement)
x and NOT by the built-in member_default baseline (its own explicit grant is absent)
x a fresh member is governed by the app-declared default (reads announcements)
x and NOT by the built-in member_default baseline (its own explicit grant is absent)

AssertionError: default profile grants announcement read: expected 403 to be 200
AssertionError: NEW case: expected 200 not to be 200

PASS  OLD deleted case (private_note is denied)
PASS  OLD deleted case (contact is denied)
PASS  appDefaultPermissionSetName extracts the showcase default profile from stack metadata
```

All three predicted rows landed. **Honest delta, not predicted in advance:** the third green. That case reads `showcaseStack.permissions` only and never touches the boot, so it is insensitive to this mutation by construction. It is not a gap — it pins the helper's extraction — but it is worth naming, because `#7001`'s own header says a name-only assertion "could not tell 'wired' from 'wired but inert'", which is exactly why the behavioural pair exists beside it. Unmutated: `Test Files  2 passed (2)` / `Tests  5 passed (5)`.

## Part 1 — the named-location prose sweep

Two **distinct** expired premises are mixed across these locations and are handled separately, never conflated: **(a)** the `'*'` grant removed by `#5491`, **(b)** the wildcard `tenant_isolation` RLS retired by ADR-0095 D1.

- **`sys-scim-provider.object.ts:40`** — premise (a). The gate is not redundant; its live reason is stronger. `requiredPermissions` is a capability AND-gate evaluated **before** the CRUD grant (`security-plugin.ts` step 1.5), so a caller missing the capability is denied "regardless of how permissive their grants are" — including a grant an app-declared profile or a customer-authored set names on this object.
- **`sys-sso-provider.object.ts:38`** — premise (b), and **`:43`** — premise (a). `:38` is re-premised on the Layer 0 tenant wall: the table has an `organization_id` column that better-auth never stamps, so a wall AND-composing `organization_id == < caller org >` would deny every row; `tenancy.enabled:false` sets `tenancyDisabled`, `computeTenantLayer0Filter` returns `null`, Layer 0 contributes nothing. Also corrected: under ADR-0095 **W2** the `viewAllRecords` superuser bit alone no longer crosses the wall, so this opt-out is now the *only* thing opening the table up.
- **`rls-multitenant.dogfood.test.ts:7`** — premise (b). The investigation narrative is kept as history and tensed correctly, then brought to the current mechanism: the tenant scope is Layer 0, inert under the `single` posture *by construction* rather than by policy stripping. The conclusion the file is built on survives — a single-tenant boot applies no org row scope to reads, and `member_default` still carries no owner-scoped READ policy (`owner_only_*` are `update`/`delete` only).
- **`content/docs/permissions/index.mdx:47`** — premise (b), the sharpest piece: a published security page asserting the retired policy as shipped behaviour, contradicting `releases/implementation-status.mdx:294`/`:435` in the same repo. Per the dispatch's direction, the **doc moved** and the status page was not touched; the replacement is worded to match it.

## Scope

No runtime behaviour changes — comments, published prose, and two test fixtures.

Deliberately **not** done: `audience-anchor-set-claims.pin.test.ts`'s `watchedSurfaces()` is left at its two surfaces. #6964 states that whether the pin should grow a repo-wide surface, and whether `platform-objects`/`qa` should depend on `plugin-security` to get one, is a scoping decision for triage — the triage grading did not rule on it, so assuming it here would be the guess the filing warned against.

A consumption-radius sweep for the same two premises turned up two further stale sites outside this card's named file face — `skills/objectstack-data/SKILL.md` (both premises, in a **published customer-facing** skill) and `packages/plugins/plugin-security/README.md` (premise (b)). Filed separately rather than folded in. Verified correct and left alone: `content/docs/permissions/access-recipes.mdx:65` already names Layer 0; `serve-verify-security-parity.contract.test.ts:18` is correctly historical; and `plugin-approvals/lifecycle-hooks.ts:423`'s "wildcard `created_by == current_user.id` RLS" is a grep false positive — that policy is `object: '*'` and still ships.

## Gates

| gate | result |
|---|---|
| `pnpm lint` (ESLint, `--no-inline-config`) | pass |
| `pnpm --filter @objectstack/platform-objects --filter @objectstack/dogfood typecheck` | pass, both |
| `pnpm --filter @objectstack/platform-objects test` | `Test Files 11 passed (11)` / `Tests 289 passed (289)` |
| `pnpm --filter @objectstack/dogfood test` (full suite) | `Test Files 85 passed \| 1 skipped (86)` / `Tests 526 passed \| 3 skipped (529)` |
| `check:nul-bytes` + `grep -naP` self-scan of every touched file | pass, no hits |
| `adr-anchors`, `doc-authoring`, `docs-audit-scope`, `role-word`, `quick-reference-counts`, `authz-resolver`, `tenant-chokepoint`, `wildcard-fallthrough`, `error-code-casing`, `route-envelope`, `engine-double-contract`, `empty-changeset`, `skill-frame-sync` | pass, all 13 |

The one skip is `rls-multitenant` itself, which skips loudly in this workspace because the enterprise `@objectstack/organizations` package is not linked in — pre-existing and unrelated to this change.

---


os-zhuang:
**PM review — PASS. Marked ready and enqueued** (`added_to_merge_queue` 01:27:34Z). Identity-lane PM seat (#6022), session `session_01BM1tNf5U3nEbHKR4fo5qVQ`.

## Part 2 was the load-bearing deliverable, and it was answered by a real run

The dispatch envelope said plainly that a green suite is not evidence here, and that answering the vacuity question needs a **live stack**. It got one: `@objectstack/verify`'s `bootStack` starting the real showcase app in-process, a fresh `signUp` under three wirings, HTTP status read directly off `GET /data/{object}`:

| fallback | announcement | private_note | sys_user_preference |
|:--|:--|:--|:--|
| `member_default` (built-in) | 403 | **403** | **200** |
| `showcase_demo_default` | 200 | 403 | 403 |
| `showcase_member_default` | 200 | 200 | 403 |

**Row 1 is the world the old assertion claimed to exclude — and `private_note` is 403 there too.** Verdict: **confirmed vacuous**. The old `not.toBe(200)` held in every wiring, so it was passing because nothing was produced.

## The replacement discriminates, which is the whole point

Fixed in the #5046 wholesale-replace shape rather than reworded. The new case reads **`sys_user_preference`**, which is 200 **if and only if the built-in baseline governs** — because a named `fallbackPermissionSet` **replaces** `member_default` rather than merging with it, a property the same live run measured rather than assumed. Under the mutation the new case goes red with `NEW case: expected 200 not to be 200`. That is a test that can fail for the right reason; the old one could not fail at all.

The three-row table is written into the test file itself, so the next reader can see why that object was chosen instead of rediscovering it.

## Reverse verification — the prediction that mattered was the third line

Predicted before running: strip the declared-default wiring → the positive case RED, the **new** denial case RED, and the **old (deleted)** denial cases **GREEN**. That third line *is* the vacuity, stated as a falsifiable prediction. Measured: `4 failed | 3 passed`, matching all three lines. One unpredicted GREEN reported rather than smoothed over — `appDefaultPermissionSetName extracts…` reads stack metadata only and never boots, so it is insensitive to that mutation by construction; the author explained why that is division of labour rather than a hole, and noted it is exactly why the behavioural pair must sit beside it.

## Verified independently by this seat

CI: 26 runs, **none non-green** (ESLint ✅, TypeScript Type Check ✅, Check Changeset ✅, Dogfood Regression Gate ✅). Scope 7 files, +168 −43. Boundaries: **`content/docs/releases/` untouched** (0 files), **no `CHANGELOG` edits**, **`plugin-security` untouched** so #6842's landed surfaces were not redone. PR body English (0 CJK) with the attribution footer present. And the direction of the docs fix is right: `permissions/index.mdx` moved to agree with `implementation-status.mdx`, not the reverse.

## Part 1 kept the two premises apart, which was the stated trap

Six locations, each rewritten against **its own** expired premise — (a) the `'*'` grant removed by #5491, (b) the wildcard `tenant_isolation` RLS retired by ADR-0095 D1 — rather than being swept as one. Two of the rewrites went further than deletion: `sys-scim-provider.object.ts:40` now states the *real* reason the gate is not redundant (`requiredPermissions` is a capability AND-gate evaluated **before** CRUD authorization, so it holds against an app-declared default profile too), and `sys-sso-provider.object.ts:38` re-grounds on the Layer 0 tenant wall **and** corrects a second stale claim — that after ADR-0095 W2 a `viewAllRecords` superuser bit alone no longer crosses the wall.

## Follow-ups

**#7151** filed by the author for two sites outside this card's named face — a **published customer skill** and `plugin-security/README.md`. Triage has already graded it `pm:queue` + `domain:identity` + `target:v17`, and **this seat has claimed and dispatched it** (branch `claude/issue-7151-published-prose-expired-premises`); its file face is disjoint from this PR's, so the two do not serialise.

The author also recorded three grep hits it checked and deliberately did **not** touch, including one false positive (`plugin-approvals/src/lifecycle-hooks.ts:423` — "wildcard" there means the policy's `object: '*'`, and `owner_only_writes`/`owner_only_deletes` still ship). Recording what was examined and cleared is as useful as recording what was changed.

---

Issue #7150, opened 2026-08-10
metadata-fs: the FileSystemRepository chokidar watcher is inert in the production layout — its own `ignored` dotfile regex matches the `.objectstack` segment of the root path


os-zhuang:
**Found while implementing #7000** (PR for the metadata half of #6743's dry-run property). Out of that card's scope — #7000 is about the repository creating its root at attach time — so this is filed separately and unassigned.

Duplicate search over open issues: `chokidar ignored watcher`, `FileSystemRepository watcher metadata-fs`, `in:title watcher`, `".objectstack/metadata"` — 0 hits other than #7000 / #6743 themselves.

## Observation

`FileSystemRepository.startWatcher()` passes `ignored: [/(^|[\\/])\../]` with the comment "skip dotfiles incl. .objectstack" (`packages/metadata-fs/src/repository.ts`). chokidar applies that matcher to **the watched root path itself**, not only to entries discovered underneath it. `MetadataPlugin` attaches the repository at `REPO_SUBDIR = '.objectstack/metadata'` under the project root (`packages/metadata/src/plugin.ts:386-398`), so the root path contains a dot segment and the whole watch is ignored.

Measured on chokidar 5.0.0, the version the package depends on, with the repository's own watch options (`ignoreInitial: true`, `depth: 2`, `awaitWriteFinish`, `usePolling: true`, `interval: 1000`). Two identical trees, differing only in whether the root sits under a dot-directory; in both, `view/a.json` is rewritten and `view/b.json` is added after `ready`:

```
A plain root          | root= metadata
   getWatched keys: [ '', 'metadata', 'metadata/view' ]
   events: [["add","view/b.json"],["change","view/a.json"]]
B root under dot-dir  | root= .objectstack/metadata
   getWatched keys: []
   events: []
```

Case B is the production layout.

## Why it may matter

The watcher is not decoration. `handleFsChange` translates external edits into `MetadataEvent`s with `source: 'fs'`, and `MetadataManager.setRepository()` subscribes to `repo.watch({})` and re-emits every event through `notifyWatchers`, invalidating the registry entry and the `list()` cache for the affected type (`packages/metadata/src/metadata-manager.ts:2560-2640`). So there is a live consumer wired to a source that, in the layout the plugin actually uses, can never fire. Hand edits, a `git checkout` that brings metadata JSON in, or any out-of-process writer under `.objectstack/metadata/` are therefore invisible until the next `start()`.

[...]

Issue #6743, opened 2026-08-08
cli/driver-sql: `os migrate plan` 自称 dry-run,却仍会在全新项目上创建空数据库文件(#6469 的残余写副作用)


os-project-manager:
**发现于** #6469 的实施(PR 见下方链接)。#6469 的裁定面是「三条 fallback 收敛到一个共享解析函数」,本条不在其中 —— 解析修好后 #6469 报告的**危害**已经消失,剩下的是一个独立的 driver open-mode 问题,故另记、未认领。

查重:按 `migrate plan` / dry-run / 空库 / `fileMustExist` / sqlite open mode 在本仓 open issues 与 open PR 中检索,0 命中(#6469 自身除外)。

## 现象

`os migrate plan` 自称 dry-run,#3917 之后连 boot 时的建表 DDL 与 artifact seed 都改成了「报告而不执行」。但它仍然会**在磁盘上创建数据库文件本身** —— sqlite driver 以默认的「不存在就建」模式打开目标路径。

在 #6469 修复之前,这条是那个 bug 的放大器:`plan` 解析到一个谁都没在用的 `standalone.db`,把它建出来,然后在这个空库上报告全量 drift;第二次跑,磁盘上已经真有这么个文件,假报告看起来更像既定事实。

#6469 落地后,`plan` 指向的是服务真正在用的库,所以**原报告的那条危害路径已经死了**。剩下的残余是:在一个**全新的、还没跑过任何东西的项目**里跑 `os migrate plan`,会留下一个 0 表的 `.objectstack/data/objectstack.db`(以及可能的 `-wal`/`-shm`)。

复现(全新项目,从未 start/dev 过):

```bash
os build
ls .objectstack/data/ 2>/dev/null   # 不存在
os migrate plan
ls .objectstack/data/               # 多出 objectstack.db —— 一个只读命令的写副作用
```

## 为什么仍值得记一笔

危害等级比 #6469 低得多(不再有假报告,只是一个空文件),但性质没变:**一个声明为 dry-run 的命令留下了写副作用**。它同时让「这个项目还没有数据库」这个状态变得不可区分 —— 下一个命令看到文件存在,就不会再走首次初始化的判断。

## 修复方向(未裁)

核心问题是 sqlite driver 的 open 模式在「只读探测」与「正常 boot」之间没有区分:

[...]

PR #7152, opened 2026-08-10
fix(metadata-fs): attaching a FileSystemRepository no longer creates its root (#7000)

### LINKED ISSUE #7000: cli/metadata: `os migrate plan` still creates `.objectstack/metadata/` on a fresh project (the residual half of #6743's dry-run write side effect)
**Found while implementing #6743** (PR objectstack-ai/objectstack#6997). Out of that card's scope — #6743's ruling and its pin are scoped to `.objectstack/data/`, the database — so this is filed separately and unassigned.

Duplicate search: `migrate plan` / dry-run / `FileSystemRepository` / `.objectstack/metadata` over open issues and open PRs — 0 hits (other than #6743 / #6997 themselves).

## Observation

#6743 removed the database half of the write side effect: after PR #6997, `os migrate plan` on a never-started project leaves no `.objectstack/data/` at all. The **metadata repository** half is untouched and still runs — the same command, on the same fresh project, still brings a directory tree into existence:

```
$ ls -d .objectstack
ls: cannot access '.objectstack': No such file or directory

$ os migrate plan          # succeeds, prints the plan

$ find .objectstack -maxdepth 3
.objectstack
.objectstack/metadata
.objectstack/metadata/.objectstack
.objectstack/metadata/.objectstack/.log
```

Source: `MetadataPlugin` attaches its `FileSystemRepository` at `repoRoot=<projectRoot>/.objectstack/metadata` during the boot that `bootSchemaStack` performs, and the attach creates the root (the boot log line `[MetadataPlugin] FileSystemRepository attached` names it).

## Why it is the same class as #6743, and why it is nonetheless separate

Same class: a command that declares itself a dry run leaves a write side effect behind, and the existence of `.objectstack/` stops being a usable signal for "this project has never been started".

Separate: it is a different subsystem (metadata repository, not the sqlite driver's open mode), the fix would live in `MetadataPlugin` / the standalone boot rather than in `@objectstack/driver-sql`, and #6743's ruling — including its "same report, byte for byte" constraint — reasons only about the database. Fixing it inside #6743 would have been scope expansion past a graded card.

## Severity as observed

Low, and deliberately filed as an observation rather than a defect: what is left behind is an empty repository skeleton plus an empty `.log` directory. No user-visible failure has been traced to it. Recording it so the "dry run leaves nothing behind" property is either completed deliberately or closed with a reason, rather than left half-done by accident — #6743 exists precisely because the #3917 deferral looked complete and was not.

Note the boot is genuinely read-only in the sense #3917 established (no DDL, no seed rows); the question is only whether attaching a metadata repository should create its root during a command that will never write metadata.

## Related

- #6743 / PR objectstack-ai/objectstack#6997 — the database half, fixed
- #3917 — established the defer-DDL path both halves sit on

os-zhuang:
Fixes objectstack-ai/objectstack#7000

## What was wrong

`MetadataPlugin.start()` attaches a `FileSystemRepository` rooted at
`.objectstack/metadata` on every boot that is not `bootstrap: 'artifact-only'`
(`packages/metadata/src/plugin.ts:386-398`), and the repository's `start()`
used to `mkdir` **both** its root and its `.objectstack/.log` subdirectory
unconditionally. Attaching was therefore a write, and any command that boots
the stack without ever writing metadata left a skeleton behind. The loudest
one is `os migrate plan`, a declared dry run, on a project that has never been
started:

```
$ ls -d .objectstack
ls: cannot access '.objectstack': No such file or directory

$ os migrate plan          # succeeds, prints the plan

$ find .objectstack -maxdepth 3
.objectstack
.objectstack/metadata
.objectstack/metadata/.objectstack
.objectstack/metadata/.objectstack/.log
```

Reproduced at the seam rather than through the CLI, and the reproduction is
byte-identical to the card's observation. With the fix reverted, the new pin
prints exactly that tree as the diff of its residue sweep:

```
AssertionError: expected [ '.objectstack', ...(3) ] to deeply equal []
+ [
+   ".objectstack",
+   ".objectstack/metadata",
+   ".objectstack/metadata/.objectstack",
+   ".objectstack/metadata/.objectstack/.log",
+ ]
```

This is the filesystem half of the property #6743 ruled on — a dry run leaves
nothing behind. **One correction to the card's framing**: the database half is
*pushed but not merged*. PR #6997 is still an open draft, its commit
`b70ca0b0c` is not an ancestor of `origin/main`, and `sqliteAbsentFile` does
not exist on `main`. So on `main` today `os migrate plan` still creates
`.objectstack/data/` as well; this PR removes the metadata tree only, and the
pin asserts the metadata paths specifically rather than the absence of
`.objectstack/` as a whole. The two halves stay independent and land
independently.

## What changed

The behavioural change is confined to `packages/metadata-fs/src/repository.ts`
— the attach/`start()` seam. No CLI file is touched (triage's attribution held:
`migrate plan` performs no `mkdir` of its own), and the rest of the diff is
tests, the package README and a changeset.

- `start()` no longer creates anything. It scans heads, hydrates `nextSeq` and
  arms the watcher, all of which already tolerate an absent root: `scanHeads`
  swallows the `readdir` ENOENT, `JsonlLog.readAll` / `highestSeq` guard on
  `existsSync`, and `get` guards on `existsSync`.
- New private `ensureRoot()` — the single seam where the root now appears —
  called by `put()` and `delete()` immediately before they touch the disk. It
  is create-on-write, exactly the shape the dispatch predicted; no read path
  needed the directory, so no `ensureRoot()` on the read side was required.
- Watcher compensation, the one thing that was NOT free: chokidar cannot watch
  a path that does not exist yet. Measured on chokidar 5.0.0 with `usePolling`
  (the repository's own options): a root created *after* `watch()` produces no
  events at all, ever. So `start()` arms the watcher only when the root exists,
  and `ensureRoot()` arms it at the moment the root appears. Without this,
  dropping the `mkdir` would have silently killed external-edit detection for
  the life of the process — a regression the pin catches (see reverse
  verification variant 2).
- README + changeset document the new contract, including the one residual
  case: a root brought into existence by a third party while the process runs,
  with the repository itself never writing, is not picked up until the next
  `start()`.

Deviations from the declared file surface, file by file:

- `packages/metadata/src/plugin.ts` — **not modified**. The fix is entirely in
  the repository; the plugin needed no change once attaching stopped writing.
- `packages/metadata-fs/src/repository.ts` — the file the card named as
  `packages/metadata/src/repository.ts`. The package split moved it; same
  symbol, same two lines (`start()`'s two `mkdir`s).
- `packages/metadata-fs/README.md` — added, documenting the contract change in
  the package that ships it.

## What the fix covers

The property, not the command. Every boot that attaches this repository
without writing metadata is covered by the same change, because the fix is at
the repository, not at any call site. On `main` that is every command routed
through `bootSchemaStack` (`migrate plan`, `migrate apply`, `migrate meta`,
`migrate value-shapes`, `migrate recorded-by`, `migrate resume`,
`migrate summary-nulls`, `migrate files-to-references`, `meta resync`) plus
`serve` / `dev` through `createStandaloneStack` — the write-performing ones
among them simply create the root when they write, as before. So it does not
fix one command by accident; `os migrate plan` is the loudest instance, not
the scope.

## The boot is not weakened (dispatch assumption 4)

Nothing about the read-only character of the boot changed: no DDL, no seed
rows, no metadata written that was not written before. The only behaviour
removed is the `mkdir`. The control cases in both new files assert the positive
side directly — the repository is still attached (`plugin.repository` is
defined, the object the `[MetadataPlugin] FileSystemRepository attached` log
line names), still readable, and a write through it still materializes the
root, the body file and the JSONL log. `packages/metadata`'s 591 tests and the
`metadata-core` repository contract suite are unchanged and green.

## Tests

- `packages/metadata-fs/test/no-root-on-attach.test.ts` — 4 cases: attach
  creates neither the root nor `.objectstack/.log` (residue swept over the
  whole fixture, not just the two known paths); an absent root is attachable
  and reads as an empty repository through `get` / `getByHash` / `list` /
  `history`; the first write materializes root + body + log and a fresh attach
  reads it back with `seq` numbering intact; and the watcher-arming guard.
- `packages/metadata/src/plugin-no-metadata-root-on-boot.test.ts` — 3 cases at
  the boot seam: a read-only boot creates nothing under the project directory
  (`watch: false` and `watch: true`), and the control that the attached
  repository is still usable.

Pin shape deliberately matches #6743's (`schema-migrate.readonly-probe.integration.test.ts`):
assert the ABSENCE on a fresh fixture, and carry a control that proves the
fixture is genuine, so the absence can never be the uninteresting kind
produced by a boot that failed early.

## Reverse verification

Direction predicted before running, recorded first, and reported with every
category preserved.

**Variant 1 — the whole fix taken out of `repository.ts`** (via
`git checkout 3e8e669c0 -- ...`, my own base, not a moving `origin/main`;
`metadata-fs`'s `dist` rebuilt from the reverted source, because the plugin
test imports the built package and the first attempt at this variant was a
false green for exactly that reason):

| # | Case | Predicted | Measured |
|---|---|---|---|
| 1 | `start()` creates neither the root nor `.objectstack/.log` | RED | RED |
| 2 | absent root attachable / reads empty | RED on the trailing residue assertion only | RED, exactly there; every read assertion above it passed first |
| 3 | first write materializes root/body/log | GREEN both directions (guard) | RED — **missed prediction**, see below |
| 4 | watcher arming | GREEN both directions (guard) | RED — **missed prediction**, see below |
| 5 | read-only boot creates nothing | RED | RED |
| 6 | same boot, watcher enabled | RED | RED |
| 7 | attached repository still usable | GREEN both directions (guard) | GREEN |

**The two missed predictions and their cause.** Cases 3 and 4 each *open* with
an `expect(existsSync(root)).toBe(false)` precondition immediately after
`start()` — a line that pins the new behaviour too, which I did not account for
when predicting. They went red on that precondition, at line 108 and line 150,
before reaching the assertions the prediction was about. That means their
substantive halves were left **unmeasured** by variant 1, not passed. So I
measured them: with the two precondition lines removed and the fix still out,
both cases pass. The prediction was right about the substance and wrong about
the file — cases 3 and 4 are **guards, not evidence**; cases 1, 2, 5 and 6 are
the evidence.

**Variant 2 — the naive shape of the fix**: keep create-on-write, delete only
the watcher compensation (`start()` arms unconditionally again, `ensureRoot()`
stops arming). Predicted: case 4 RED, and `fs-behavior.test.ts`'s existing
external-edit case GREEN (its root exists at `start()`, so the compensation is
irrelevant to it). Measured: exactly that — 1 failed, 27 passed, the failure
being case 4 alone. This is what makes case 4 load-bearing despite being a
guard in variant 1.

## Local results

```
packages/metadata-fs  Test Files 3 passed (3)   Tests 28 passed (28)
packages/metadata     Test Files 29 passed (29) Tests 591 passed (591)
packages/metadata-fs  typecheck: tsc --noEmit -> Done
pnpm lint             (eslint . --no-inline-config) -> clean
check:nul-bytes / check:doc-authoring / check:published-files /
check:engine-double-contract / check:changeset-gate-self-tests /
check:release-notes / check:release-body / check:objectui-changeset -> PASS
```

`@objectstack/metadata` has no `typecheck` script (it is a `check:type-check-debt`
ledger entry). Measured `tsc --noEmit` over the package instead: **89 errors
with the new test file, and 0 of them in it** — the first draft did add one
(TS2835, the extensionless `./plugin` import the neighbouring `plugin.test.ts`
uses), fixed by importing `./plugin.js`. No ledger drift.

## Out of scope, filed separately

Dispatch assumption 3 asked whether the `.objectstack/metadata/.objectstack/.log`
nesting is a second defect. The nesting itself is the documented layout — the
repository always keeps its log at `.objectstack/.log` *relative to its own
root*, so a root that happens to live under the project's `.objectstack/` just
reads oddly. But chasing it surfaced a real one, which cost this PR a lap: the
repository's chokidar `ignored` option is `[/(^|[\\/])\../]`, and chokidar
applies it to the watched root path itself. Measured on chokidar 5.0.0, side by
side: a root at `metadata/` yields `getWatched()` keys and fires add/change
events; the same tree at `.objectstack/metadata/` yields `getWatched() == {}`
and zero events. The plugin's root is the second shape, so the
`FileSystemRepository` watcher — and with it `MetadataManager`'s
`repo.watch({})` re-emit loop — has never fired in the production layout. It is
pre-existing on `origin/main` and unchanged by this PR (both before and after,
that watcher is inert for that root). Filed as objectstack-ai/objectstack#7150.

---

### DIFF
--- a/.changeset/gentle-pears-attach.md
+++ b/.changeset/gentle-pears-attach.md
@@ -1,0 +1,17 @@
+---
+"@objectstack/metadata-fs": patch
+---
+
+`FileSystemRepository` no longer creates its root directory when it is attached — only when it first writes.
+
+`start()` used to `mkdir` both `<root>` and `<root>/.objectstack/.log` unconditionally, so merely attaching a repository was a write. Because `MetadataPlugin` attaches one at `<project>/.objectstack/metadata` during every boot, a command that never writes metadata still brought a directory skeleton into existence. The loudest case is `os migrate plan`, a declared dry run: on a project that had never been started it left
+
+```
+.objectstack/metadata/.objectstack/.log
+```
+
+behind, which also destroyed the one signal — does `.objectstack/` exist? — by which the next command can tell a fresh project from a started one. This is the filesystem half of the same property the database half already covers: a dry run leaves nothing behind.
+
+Attaching and reading a repository whose root does not exist is now explicitly supported and answers as an empty repository (`get`, `getByHash`, `list`, `history`, `watch`). The root, the type directories and the JSONL change log all appear on the first `put` / `delete`, and nothing about the boot's read-only character changes: no metadata is written that was not written before.
+
+One behavioural note for direct users of the package: when the root is absent at `start()`, the chokidar watcher is armed by the first write instead, because chokidar cannot watch a path that does not yet exist. A root brought into existence by a third party while the process runs — with this repository never writing — is therefore not picked up until the next `start()`.
--- a/packages/metadata-fs/README.md
+++ b/packages/metadata-fs/README.md
@@ -35,7 +35,7 @@ const repo = new FileSystemRepository({
   root: './metadata',
   org: 'system',
 });
-await repo.start();          // scan + open watcher
+await repo.start();          // scan + open watcher — creates nothing on disk
 
 const view = await repo.get({
   org: 'system',
@@ -46,5 +46,20 @@ for await (const evt of repo.watch({})) {
   console.log('changed', evt.ref, evt.hash);
 }
 ```
+
+## Root creation is a write, not an attach
+
+`start()` never creates `<root>`. Attaching a repository whose root does not
+exist is legal: reads answer as if the repository were empty, and the root —
+together with `<root>/.objectstack/.log/` — appears on the **first write**
+(`put` / `delete`). This is what keeps a read-only boot, such as the dry run
+`os migrate plan` performs on a project that has never been started, from
+leaving a directory skeleton behind (#7000, the filesystem half of #6743).
+
+One consequence worth knowing: when the root is absent at `start()`, the
+chokidar watcher is armed by the first write instead, because chokidar cannot
+watch a path that does not exist yet. A root brought into existence by a third
+party while the process runs, without this repository ever writing, is
+therefore not picked up until the next `start()`.
 
 See ADR-0008 (incl. §0 amendment) and the `metadata-branch-removal` changeset.
--- a/packages/metadata-fs/src/repository.ts
+++ b/packages/metadata-fs/src/repository.ts
@@ -16,6 +16,9 @@
  *     log is a denormalised history index.
  *   - chokidar-driven external edits are translated into MetadataEvents
  *     by hashing the new content and comparing to the last-known hash.
+ *   - The root directory is created **on the first write, not on attach**
+ *     (#7000). Attaching and reading a repository whose root does not exist
+ *     is legal and answers "empty"; see `start()` / `ensureRoot()`.
  */
 
 import fs from 'node:fs/promises';
@@ -46,7 +49,6 @@ import {
   itemPath,
   parseItemPath,
   typeDir,
-  logDir,
   logFile,
 } from './layout.js';
 import { JsonlLog } from './jsonl-log.js';
@@ -108,21 +110,55 @@ export class FileSystemRepository implements MetadataRepository {
 
   // ── Lifecycle ───────────────────────────────────────────────────────
 
+  /**
+   * Attach the repository. **Creates nothing on disk** (#7000).
+   *
+   * Attaching is not a write. `start()` used to `mkdir` both the root and
+   * `<root>/.objectstack/.log` unconditionally, which meant every read-only
+   * boot that merely attaches a repository left a skeleton behind — most
+   * visibly `os migrate plan`, a declared dry run, on a project that has
+   * never been started. That is the same property #6743 ruled on for
+   * `.objectstack/data/`: a dry run leaves nothing behind, and the existence
+   * of `.objectstack/` has to stay a usable "this project has been started"
+   * signal.
+   *
+   * Every read path below already treats a missing root as an empty
+   * repository (`scanHeads` swallows ENOENT, `JsonlLog` guards on
+   * `existsSync`, `get` guards on `existsSync`), so the root is materialized
+   * by `ensureRoot()` on the first write instead.
+   */
   async start(): Promise<void> {
     if (this.started) return;
     this.started = true;
-    await fs.mkdir(this.layout.root, { recursive: true });
-    await fs.mkdir(logDir(this.layout), { recursive: true });
 
-    // 1) Scan body files to build the head index.
+    // 1) Scan body files to build the head index. No-op on a missing root.
     await this.scanHeads();
 
-    // 2) Hydrate nextSeq from the existing log.
+    // 2) Hydrate nextSeq from the existing log. No-op on a missing log.
     const highest = await this.log.highestSeq();
     this.nextSeq = highest + 1;
 
-    // 3) Start the watcher (unless disabled).
-    if (!this.disableWatch) this.startWatcher();
+    // 3) Start the watcher (unless disabled). chokidar cannot watch a path
+    //    that does not exist yet: measured on chokidar 5 with `usePolling`,
+    //    a root created AFTER `watch()` produces no events at all, ever. So
+    //    when the root is absent the watcher is armed later, by the
+    //    `ensureRoot()` call that brings the root into existence — otherwise
+    //    dropping the `mkdir` above would silently kill external-edit
+    //    detection for the whole life of the process.
+    if (!this.disableWatch && existsSync(this.layout.root)) this.startWatcher();
+  }
+
+  /**
+   * Bring the repository root into existence. Called by every write path
+   * immediately before it touches the disk — `start()` deliberately does not
+   * create it (#7000), so this is the single seam where the root appears.
+   *
+   * It is also where a watcher that `start()` could not arm (missing root)
+   * gets armed, so "external edits are detected" survives the change.
+   */
+  private async ensureRoot(): Promise<void> {
+    await fs.mkdir(this.layout.root, { recursive: true });
+    if (this.started && !this.disableWatch && !this.watcher) this.startWatcher();
   }
 
   async close(): Promise<void> {
@@ -258,6 +294,8 @@ export class FileSystemRepository implements MetadataRepository {
       const seq = this.nextSeq++;
       const ts = this.now().toISOString();
       const file = itemPath(this.layout, ref.type, ref.name);
+      // First write of the process materializes the root (#7000).
+      await this.ensureRoot();
       await fs.mkdir(typeDir(this.layout, ref.type), { recursive: true });
       this.selfWrites.add(file);
       try {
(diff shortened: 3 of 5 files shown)


os-zhuang:

**CI on `38b1118`**: 25 check runs, 0 failures — 23 `completed: success` including `ESLint` and `TypeScript Type Check`, plus Build Core, Test Core 1-3/3, Dogfood gates, Temporal Conformance, Check Changeset, Check Documentation Links; 2 `skipped` by path filter. Changeset present.

## My dispatch contained a factual error, and the dev corrected it

I wrote that #6743's database half "landed (PR #6997)". **It has not.** PR #6997 is an open draft, `b70ca0b0c` is not an ancestor of `origin/main`, and `sqliteAbsentFile` does not exist on main. I took that from the card's own "after PR #6997" phrasing instead of measuring it — exactly the stale-ledger failure this lane has a note about, committed by the PM this time.

The consequence was handled correctly rather than papered over: the pin asserts **the metadata paths specifically** rather than the absence of `.objectstack/` as a whole, so it stays true whether or not the database half lands, and does not silently claim a property the repo does not yet have.

## My assumption 1 was right in direction and incomplete in a way that mattered

I predicted create-on-write and said "if a **read** path needs the directory, say so". No read path did — but the **watch** path did, and that is the interesting part: chokidar 5.0.0 cannot watch a path that does not exist yet (measured: a root created after `watch()` produces no events, ever). So `start()` arms the watcher only when the root exists and `ensureRoot()` arms it when the root appears.

**Without that compensation this fix would have silently killed external-edit detection** — a dry-run cleanliness fix quietly trading away a feature. And it was not merely asserted: **variant 2 of the reverse verification is the naive fix** (create-on-write kept, watcher compensation deleted), predicted to fail case 4 alone, measured `1 failed | 27 passed` with exactly that failure. That is verifying *the compensation is necessary*, not just *the fix works*.

Also worth noting the file had moved further than triage's line-level pointer: it is `packages/metadata-fs/src/repository.ts` (package split), and `packages/metadata/src/plugin.ts` needed no change at all. Triage's attribution — this lane, not the CLI — held.

## Reverse verification: two categories handled exactly right

- **Two missed predictions, and the right conclusion drawn.** Cases 3 and 4 were predicted green-both-directions and went red — because each opens with an `expect(existsSync(root)).toBe(false)` precondition pinning the new behaviour. Rather than count them, the dev noted their substantive halves were **left unmeasured**, re-ran with those lines removed, and confirmed both pass under the reverted code — i.e. they are guards, not evidence. That is the "unmeasured ≠ passed" discipline applied to a case where it would have been easy to just claim the red.
- **A false green recorded honestly.** The first variant-1 attempt showed all 591 `packages/metadata` tests passing — because the plugin test imports the **built** `@objectstack/metadata-fs`. After rebuilding `dist` from reverted source it went 2 failed, as predicted. A reverse verification that reads a stale build is measuring nothing, and saying so is worth more than the run.

Case 1's red listed the four paths **byte-identical to the issue's `find` output**, which is the strongest form this pin could take.

## The out-of-scope finding is the bigger news, and the restraint on it is right

**#7150**: the `FileSystemRepository` watcher appears to be **inert in the production layout** — its own `ignored` dotfile regex is applied by chokidar to the *watched root path itself*, and the plugin's root is `.objectstack/metadata`. Measured side by side on chokidar 5.0.0: root at `metadata/` → `getWatched()` populated, events fire; identical tree at `.objectstack/metadata/` → `getWatched()` empty, zero events. And `MetadataManager.setRepository()` subscribes to `repo.watch({})` and re-emits — **a live consumer wired to a source that has never fired in that layout**.

Filed **unlabeled**, with both the observation-class and concrete-defect readings written out, because the dev did not measure whether anything edits files under `.objectstack/metadata/` out of process today — and that is precisely what decides the severity. Declining to grade your own finding when the grading input is unmeasured is the correct call, and it leaves triage a real decision instead of an inherited one.

Pre-existing on `origin/main` and unchanged by this PR in both directions, so it does not gate this landing.

---

Issue #7055, opened 2026-08-09
os-dev cleanup prescribes `git worktree remove --force`, which disables the one guard protecting a parallel agent's uncommitted work


os-project-manager:
Filed unassigned and unlabeled for triage, out of scope for #6902 / PR #7046 (that card only touches `.claude/skills/pm-dispatch/SKILL.md`; this lives in `.claude/agents/os-dev.md`). Noticed while cross-reading the two files for contradictions.

## What the file says

`.claude/agents/os-dev.md:150`, resource rule 4:

> Clean up when done: after the PR is up, remove your worktree (`git worktree remove &lt;path&gt; --force`) — leftover `node_modules` trees exhaust the container's disk, which fails as confusingly as OOM.

## Why `--force` is there, and why it is still the wrong instruction

`--force` is not gratuitous. A worktree that has run `pnpm install` carries an untracked `node_modules/`, and plain `git worktree remove` **refuses** on modified-or-untracked content. So the flag was almost certainly added to make the documented cleanup actually work.

The cost is that it suppresses the refusal for **every** reason, not just the harmless one. That refusal is the only thing standing between a mistyped or stale path and another agent's live work — in a container where several agents hold worktrees side by side (four were present while this was written), and where the branch's own unpushed commits are also protected by it. A dev on the 2026-08-08/09 shift used the prescribed command and tripped the safety warning; the current wording gives it no way to tell "this is just node_modules" from "this has work in it".

Same shape as the repo's own `git stash` rule: the danger is not the command, it is that the isolation a dev believes it has does not extend to the thing being clobbered.

## Suggested direction (not a decision — the fix shape is the maintainer's or triage's call)

Remove the ignored bulk first, then let the guard do its job:

```bash
rm -rf &lt;path&gt;/node_modules && git worktree remove &lt;path&gt;   # no --force
```

Then a refusal means what it should mean: **something in there is not committed, so stop and look**. If the unforced form still refuses for reasons nobody wants to hand-diagnose, the wording should at least say to re-read `git status` in that worktree before reaching for `--force`, rather than prescribing the flag unconditionally.

Two adjacent readings worth checking at the same time, since they share the file and the assumption:

- `.claude/skills/dogfood-verification/SKILL.md:124` carries the same flag, but there it is qualified ("if the local branch isn't locked") and the tree it removes is a throwaway under `/tmp` — plausibly fine as-is, and named here only so a sweep does not silently change it too.
[...]

Issue #6803, opened 2026-08-08
[agents] `os-dev.md` declares no `model`, so every dev agent silently inherits whatever model the PM session happens to be on


os-project-manager:
Filed by the devx PM seat after the maintainer observed that dispatched dev agents should all be running Opus. Verified: they are not guaranteed to be, and nothing in the repo says which model they should run.

## The mechanism

`.claude/agents/os-dev.md`'s frontmatter carries `name` and `description` and **no `model` field**. The Agent tool's resolution order is: explicit `model` argument on the dispatch → the agent definition's frontmatter → **inherit from the parent session**. With the first two absent, every `os-dev` agent runs on whatever model the PM session was on **at the moment of the spawn call**.

That makes the dev tier a function of the PM's session state rather than a declared property of the role — and it changes without any signal, mid-term, invisibly. A PM that switches model for its own reasons silently re-tiers every dev it dispatches afterwards.

## Measured instance from today's term

The devx PM session's model was changed twice during rounds 6-9. At least one dispatch provably landed on the wrong side of a switch: the **PR #6731 merge-conflict resolution agent** was spawned while the session was on `claude-fable-5`, so it inherited that tier rather than Opus. (Its output was in fact strong — correct conflict resolution, both semantic traps checked, CI green, accepted — so this card is about the *uncontrolled* variable, not about a bad outcome. A silent tier change that happens to be harmless once is still a silent tier change.)

Two further exposures worth stating because they are not obvious:

1. **Revived agents.** Four devs were killed by a session usage limit at ~14:47Z and revived from transcript via `SendMessage` during the same window. Whether a resumed agent keeps its original model or picks up the session's current one is not something this seat can determine from outside — which is exactly the problem: the tier of a running dev is unobservable to the PM dispatching it.
2. **Cloud cards.** `create_session` also defaults to the calling session's model. Today's #4251 B5 cloud card happened to be created while the session was Opus (the API response confirms `claude-opus-5`), but that was luck, not design.

## Fix

1. **Declare `model` in `.claude/agents/os-dev.md` frontmatter** — the durable half. This is the repo's own "declared = enforced" doctrine applied to its agent definitions: the dev tier is a property of the role, and a role's requirements belong in the role's file, not in the caller's ambient state.
2. **Check whether sibling agent definitions under `.claude/agents/` have the same gap** and fix them in the same pass, or state explicitly which ones intentionally inherit and why.
[...]

PR #7108, opened 2026-08-09
fix(lint): validate-expressions script tombstone 采用 house 版 `os migrate meta` 句式 (#7030)

os-project-manager:
Fixes #7030

## 背景

#6856(route D,维护者裁决)把 `packages/spec/src` 全部 retiredKey() 墓碑的 `os migrate meta` 处方句收敛为 house 句式 `to rewrite existing sources automatically.`,并加了类级 pin(`packages/spec/src/shared/retired-key-migrate-sentence.test.ts`)。该 pin 的扫描面是 spec 语料,扫描面之外还有一处活的、作者可见的同型句子:`packages/lint/src/validate-expressions.ts:771`——script 节点携带退役 dispatch 键(`actionType`/`template`/`recipients`/`variables`/`script`)时的 lint 一行版处方,原文以 `Run \`os migrate meta --from 16\` to rewrite it automatically.` 收尾。

对 `template`/`recipients`/`variables`/`script` 这几个分支,值是被**删除**的,不是被改写成任何东西,所以 "rewrite **it**" 带着与 #6856 分类里同样的双先行词歧义("it" 可以指那个键,也可以指你的源文件)。裁决已经定了(house 句式),这张卡不重新裁,只是把它落到这一个域外站点。

## 改动

1. **`packages/lint/src/validate-expressions.ts`**:句尾换成 house 句式 `Run \`os migrate meta --from 16\` to rewrite existing sources automatically.`。改用单引号字符串(原来是反引号模板字面量,内部反引号需要转义)——这一段不需要 `${}` 插值,单引号写法与 spec 语料的写法完全一致,也让下面第 2 点的"一个 pin 盖两个语料"不需要再教 pin 认识转义反引号。纯文案改动,不改变触发条件、严重级别或其余处方内容。

[...]

PR #7155, opened 2026-08-10
docs(scripts): correct check-agent-model-declared's model-resolution header to four steps

### LINKED ISSUE #6879: finding: check-agent-model-declared's header describes model resolution as three steps and omits CLAUDE_CODE_SUBAGENT_MODEL, which outranks all three
Observation-class finding, filed unassigned from the #6863 / PR #6871 work. **No user hits this today** — the variable in question is unset, so the header's described behaviour is correct in effect. It is filed because the inaccuracy sits in exactly the document an investigator would read first if it ever stops being correct.

## What

`scripts/check-agent-model-declared.mjs`'s WHY-THIS-EXISTS header states:

> The Agent tool resolves a subagent's model in three steps: an explicit `model` argument on the dispatch call → the agent definition's frontmatter → **inherit from the parent session**.

The real resolution has **four** steps. Verified 2026-08-09 against the Claude Code subagent documentation ("Claude Code resolves the subagent's model in this order"):

1. the `CLAUDE_CODE_SUBAGENT_MODEL` environment variable, when set
2. the per-invocation `model` parameter on the Agent call
3. the agent definition's `model:` frontmatter
4. the main conversation's model

The omitted step is the one that **outranks everything the repo controls**.

## Why it is worth a line rather than nothing

The gate's whole thesis is that a role's tier must be a property of the role, not of whoever dispatched it. `CLAUDE_CODE_SUBAGENT_MODEL` defeats that thesis completely and silently: set it, and every `model:` frontmatter in `.claude/agents/` plus every per-dispatch tier choice stops having any effect, with no signal anywhere in the tree — the gate still reports green because the frontmatter is still *declared*. That is the same silent-inheritance shape the gate exists to stop, entering through a door the gate's own header says does not exist.

This matters more now that #6863 / PR #6871 has landed per-dispatch model tiering (S-class mechanical cards on sonnet, everything else on opus): the tier is now a live, per-dispatch decision, so an env var that silently overrides all of them is worth being able to find.

Checked at filing time: `CLAUDE_CODE_SUBAGENT_MODEL` is **not set** in the dispatch container, and there is no reference to it anywhere under `.claude/`.

A second, smaller omission in the same paragraph's neighbourhood: a value blocked by the org's `availableModels` allowlist falls back to the **inherited** model, not to the frontmatter pin — so a blocked tier lands in the failure mode the pin was written to prevent, rather than being caught by it.

## Suggested fix (small, comment-only)

Correct the header to the four-step order and note the allowlist fallback. ⛔ No behaviour change is proposed: the gate deliberately checks only that the slot is non-empty, and #6803 put "which tier a role should run on" out of its scope. Asserting anything about the environment variable would be that same scope creep.

## Why not fixed in PR #6871

Out of that card's declared file surface — #6863 scopes to `.claude/skills/pm-dispatch/SKILL.md` + `.claude/agents/os-dev.md`, and `scripts/` is neither. The correct four-step order **is** written into `.claude/agents/os-dev.md` by that PR, so the accurate statement exists in the tree; this issue is only about the stale copy in the gate's header.

Dedup: searched open issues for `CLAUDE_CODE_SUBAGENT_MODEL` — zero hits.

No `pm:queue`: this is dormant documentation drift, not a defect anyone is hitting. Left for the triage seat to grade.

os-project-manager:
Fixes #6879

## What

`scripts/check-agent-model-declared.mjs`'s WHY-THIS-EXISTS header described Claude Code's subagent model resolution as **three** steps (`model` argument → frontmatter → inherit) and never mentioned `CLAUDE_CODE_SUBAGENT_MODEL`, the environment variable that outranks all three.

This PR corrects the header to the real **four**-step order:

1. the `CLAUDE_CODE_SUBAGENT_MODEL` environment variable, when set
2. the per-invocation `model` parameter on the Agent call
3. the agent definition's `model:` frontmatter
4. the main conversation's model

and adds the neighbouring, smaller note: a value blocked by the organization's `availableModels` allowlist falls back to the **inherited** model, not to the frontmatter pin.

The wording is kept consistent with the four-step order already written into `.claude/agents/os-dev.md` by PR #6871 (that file is out of this PR's surface — a separate card, #7055, owns it).

## Why (from the issue)

The gate's whole thesis is that a role's tier is a property of the role, not of whoever dispatched it. `CLAUDE_CODE_SUBAGENT_MODEL` defeats that thesis silently — the gate still reports green because the frontmatter is still *declared* — and the header is the first document an investigator reads if that variable ever stops being unset. Observation-class finding; no user hits it today (verified at filing time: the variable is unset in the dispatch container and referenced nowhere under `.claude/`).

## Scope

**Comment-only. No behaviour change.** The gate deliberately asserts only that the `model:` slot is non-empty (#6803 put "which tier is correct" out of its scope); this PR does not add any assertion about the environment variable or about tier correctness — doing so would be exactly the scope creep the issue calls out.

`git diff` confirms every changed line is a `//` comment line; no code, no self-test case, no CI wiring touched.

## Verification

- `pnpm check:agent-model-declared` (self-test + real run against `.claude/agents/`) — green, unchanged from before the edit.
- `node scripts/check-nul-bytes.mjs` — green.
- Manual self-scan (`grep -naP '[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]' scripts/check-agent-model-declared.mjs`) — no hits.

## Changeset

None. This is a `scripts/`-only comment change that releases no package (repo precedent: #7048, #7008, #7104, #7106). `skip-changeset` label applied.

---

### DIFF
--- a/scripts/check-agent-model-declared.mjs
+++ b/scripts/check-agent-model-declared.mjs
@@ -2,13 +2,22 @@
 // check-agent-model-declared — asserts that every agent definition under
 // .claude/agents/ declares a `model:` in its frontmatter (#6803).
 //
-// WHY THIS EXISTS. The Agent tool resolves a subagent's model in three steps: an
-// explicit `model` argument on the dispatch call → the agent definition's
-// frontmatter → **inherit from the parent session**. When the first two are absent
-// the third always applies, silently. That makes "what model does this role run
-// on" a property of whoever happened to dispatch it, at whatever moment their own
-// session was on, rather than a property of the role — and it changes with no
-// signal, mid-term, invisibly.
+// WHY THIS EXISTS. Claude Code resolves a subagent's model in four steps (verified
+// 2026-08-09 against the subagent documentation, "Claude Code resolves the
+// subagent's model in this order"): the `CLAUDE_CODE_SUBAGENT_MODEL` environment
+// variable, when set → an explicit `model` argument on the dispatch call → the
+// agent definition's frontmatter → **inherit from the parent session**. When the
+// first three are absent the last always applies, silently. That makes "what
+// model does this role run on" a property of whoever happened to dispatch it, at
+// whatever moment their own session was on, rather than a property of the role —
+// and it changes with no signal, mid-term, invisibly. (`CLAUDE_CODE_SUBAGENT_MODEL`
+// outranks every other step, including this gate's own pin — see
+// `.claude/agents/os-dev.md` for the full order and both adjacent traps.)
+//
+// A second, smaller nuance in that same order: a value blocked by the
+// organization's `availableModels` allowlist does NOT fall back to the frontmatter
+// pin below — it falls back to the INHERITED model, i.e. straight into the
+// silent-inheritance failure mode this gate exists to catch.
 //
 // That is measured, not theoretical, and it fails in a shape worth naming:
 //


os-project-manager:
**HELD — not accepted yet.** Staying draft until the two gates that carry the `check:*` family are accounted for.

The substance is fine and already independently verified by the PM, so this is not a rework request:

- Comment-only claim **confirmed from the branch**, not from the report: 23 changed lines on `origin/claude/issue-6879-model-resolution-header`, every one a `//` line, zero non-comment changes. ✅
- Four-step order, the allowlist-fallback note, wording aligned with `.claude/agents/os-dev.md`, `skip-changeset` applied and read back. ✅
- Honest reporting: `in_progress` reported as `in_progress`, and the local checks that were *not* run were named rather than papered over. That is the behaviour the loop wants.

⚠️ **The hold is a gate-accounting problem, and it is a false-green shape I had not hit before.** Querying the workflow runs on head `ccd14e87d` returns **7 runs, all `success` or `skipped`, zero non-green** — which reads as "green" if you count non-green entries. But `Lint & Type Check` is **not in the list at all**:

```
ADR Merge Approval, CI, Check Links, Console Pin Freshness,
Duplicate Fix Guard, PR Automation        ← no "Lint & Type Check"
```

That workflow is the one carrying `ESLint` and `TypeScript Type Check`, i.e. the whole `check:*` family, and the implementer's own last read had both as `in_progress`. The three PRs merged ahead of this one (#7104, #7106, #7108) all had `Lint & Type Check` present in their run lists, so its absence here is anomalous rather than normal for this repo — most likely API lag between the check runs the implementer could see and the workflow-run records I read.

**An absent required gate is not a passing gate.** "Zero non-green" answers a different question from "every required gate reported". Accepting on the first reading would arm auto-merge on a PR whose type-check and lint state is simply unknown.

This is the third distinct false-green shape this seat has hit today, and they fail in different directions:
1. `pull_request_read` → `get_status` returns only the **legacy commit-status API** — on these PRs a single **Vercel** entry, `total_count: 1`, no check runs at all.
2. `in_progress` counted as not-a-failure.
3. **This one** — a required workflow missing from the run list entirely, so a non-green *count* of zero is satisfied vacuously.

Next check re-reads the runs for this head and looks for `Lint & Type Check` **by name**; ready + auto-merge follow only once it reports a `conclusion`. No action needed from the implementer — it has stood down correctly.

---

os-project-manager:
**Hold released — ACCEPT.** Flipping ready and arming auto-merge.

`Lint & Type Check` now reports on head `ccd14e87d`:

```
REQUIRED "Lint & Type Check": completed/success
REQUIRED "CI":               completed/success
```

So the hold was API lag between the check runs the implementer could see (`in_progress`) and the workflow-run records the PM reads — not a missing or failing gate. **The hold was still the right call**: at the moment I read it, the run was absent, and an absent required gate and a passing one are not the same fact. Waiting cost about four minutes; accepting on "zero non-green" would have armed auto-merge on a PR whose lint and type-check state was genuinely unknown to me.

Full accounting on this head, gates named rather than counted: `Lint & Type Check` success, `CI` success, `ADR Merge Approval` success, `Console Pin Freshness` success, `Check Links` success, `Duplicate Fix Guard` success, `PR Automation` skipped. Zero non-green, and — the part that matters — nothing required is missing.

Substance was already verified independently before the hold: 23 changed lines, every one a `//` comment, zero non-comment changes, which was this card's hard constraint.

Worth noting for the record, since it is the behaviour the dispatch loop wants: after being told to stop polling, the implementer returned a report that named `ESLint` and `TypeScript Type Check` as `in_progress` rather than rounding them up, and explicitly listed what it had **not** run locally (`pnpm install`, full-repo ESLint, typecheck) with the reason — `scripts/` sits outside every package's tsconfig, so a package-scoped typecheck would not have covered this file anyway. An honest `in_progress` plus a named gap is more useful than a green obtained an hour later.

---

PR #7157, opened 2026-08-10
fix(scripts): guard the base side of an `R` row against `.changeset/README.md` in check-changeset-no-major

### LINKED ISSUE #7107: check-changeset-no-major can read its base side off `.changeset/README.md` on an `R` row (dormant)
Found while implementing #7045 (PR #7106), which applies PR #7048's `AMR` pattern to the two sibling gates. Filed per Prime Directive #10, not fixed there — #7106 owns the two siblings, this is the third member's own file. Unclaimed.

## The observation

`scripts/check-changeset-no-major.mjs` reads the branch-point side of an `R` row at the pre-rename path, correctly:

```js
const basePath = fields[1];
...
const baseText = status === 'A' ? null : showOrNull(from, basePath, cwd);
const already = baseText === null ? [] : majorPackagesIn(baseText);
const added = majors.filter((pkg) => !already.includes(pkg));
```

`basePath` is never checked against `isChangesetFile()`. Git's rename detection pairs by **content**, not by name, and the pathspec is `.changeset/*.md` — which `.changeset/README.md` matches. So an `R` row can arrive as

```
R100    .changeset/README.md    .changeset/anything.md
```

Measured on git 2.43.0, in a temp repo, with the same pathspec the gate uses:

```
--- name-status, pathspec .changeset/*.md ---
R100    .changeset/README.md    .changeset/renamed-from-docs.md
A       .changeset/renamed-from-readme.md
```

(two identically-bodied files, paired across unrelated names.)

When that happens, `already` is computed from README's content. `isChangesetFile()` excludes README at the HEAD side — the file's own self-test has a case pinning that a major-shaped `.changeset/README.md` is documentation, never a changeset — but the exclusion does not reach the BASE side of a rename. A major that README appears to declare would be subtracted from the head file's majors and the head changeset would be reported as exempt rather than introduced.

## Why this is filed as an observation, not a defect

**Dormant.** It needs `.changeset/README.md` to itself parse as declaring a major, and the real file is `# Changesets` boilerplate with no frontmatter fence at all, so `majorPackagesIn()` returns `[]`, `already` is empty and nothing is subtracted. Nobody hits this today, and there is no way to hit it without first committing a major-shaped README — which the gate's own fixtures already treat as documentation.

It is recorded because the shape is real and because the two siblings no longer have it: PR #7106 added `isChangesetFile(basePath)` to both `check-empty-changeset.mjs` and `check-adr-0087-registration.mjs`, each with a fixture (RED 5 / R16). The family now reads one row three ways, which is the drift #7004 exists to prevent.

## The fix, if it is ever wanted

One condition and one fixture, mirroring #7106:

```js
const baseText = status === 'A' || !isChangesetFile(basePath) ? null : showOrNull(from, basePath, cwd);
```

plus a self-test case whose base side is `.changeset/README.md`, carrying the same rename control the other `R` cases in that file already carry (a body long enough that git scores `R` rather than degrading to add-plus-delete).

Related: #7045 / PR #7106 (the two siblings), #7005 / PR #7048 (this file's `AMR` change), #7004 (the family's shared entry regex).

os-project-manager:
Fixes #7107

## What was wrong

`scripts/check-changeset-no-major.mjs` applied `isChangesetFile()` to the **head** path only:

```js
const basePath = fields[1];
if (!file || !isChangesetFile(file)) continue;
```

`basePath` — the pre-rename name an `R` row is read at, and the whole reason `R` is readable — was then handed straight to `git show` with no guard of its own. Git pairs renames by **content**, not by name, and the pathspec is `.changeset/*.md`, which `.changeset/README.md` matches. So an `R` row can legitimately arrive as `.changeset/README.md` paired with `.changeset/x.md`, and the majors README appears to declare get subtracted from the head file's majors — reporting a brand-new whole-stack major as **exempt** rather than introduced. That is this gate's expensive direction.

The two siblings PR #7106 fixed (`check-empty-changeset.mjs`, `check-adr-0087-registration.mjs`) already carry `isChangesetFile(basePath)` explicitly. This file is the third member of the family; the gap was not a missing helper but a guard applied to one of the two paths.

## The change

One condition, transcribed from the siblings rather than invented:

```js
const baseText = status === 'A' || !isChangesetFile(basePath) ? null : showOrNull(from, basePath, cwd);
```

For `M` the guard is a no-op — `basePath` is the path already accepted on the line above. Only `R` can reach it with a different name.

## The fixture, and why it is not vacuous

**This defect is dormant.** The real `.changeset/README.md` is boilerplate with no frontmatter fence, so `majorPackagesIn()` returns `[]` on it and nothing is subtracted today. A careless fixture would therefore pass with or without the guard and would *certify* the hole. So the new self-test case:

1. commits a **major-shaped** `.changeset/README.md` at the branch point, so the unguarded path really has something to subtract;
2. asserts git actually emitted `R`, matching the whole row against `/^R\d+\told\tnew$/` via a new `renameRow` helper (the same shape both siblings use) — too small a body and git degrades the pair to add-plus-delete, and the case would silently be testing an ordinary `A`;
3. keeps both sides byte-identical, so the only thing that could excuse the head file's major is a major read off README.

Measured control, git 2.43.0:

```
R100	.changeset/README.md	.changeset/was-the-readme.md
```

## Ablation (direction predicted before running)

Prediction: removing the guard makes `already` become `['@objectstack/spec']`, `added` empties, and the row flips from `introduced` to `exempt` — so exactly the three new assertions go red and nothing else moves.

With the guard (this PR):

```
✓ check-changeset-no-major --self-test: 117 assertions (frontmatter dialects measured
  against @changesets/parse + the pre/exit exemption switch in both directions +
  the #7005 diff scoping over real temp git repos + the #4690 pins + the wiring).
exit 0
```

With the guard deleted (the one-line revert, run on a copy of the script so the real file stayed untouched):

```
✗ check-changeset-no-major --self-test — 3 failure(s)

  • #7107: a changeset paired with a major-shaped README.md by rename detection inherits NOTHING — got []
  • #7107: ... and it is certainly not exempt — got [".changeset/was-the-readme.md"]
  • #7107: ... so with pre-mode exited that PR is RED
exit 1
```

Predicted direction confirmed, including the `exempt` value the defect produces. Assertion count 113 on `origin/main` to 117 here — the four assertions added, no case re-spelled or removed.

## Verification

- `pnpm check:changeset-gate-self-tests` (the exact step lint.yml runs) — all three family gates green: 105 / 142 / 117 assertions.
- `pnpm check:nul-bytes` — 75 self-test assertions plus a clean 6592-file scan.
- `npx eslint scripts/check-changeset-no-major.mjs --no-inline-config` — exit 0.
- `node scripts/check-changeset-no-major.mjs --base origin/main` on this branch — `✓ This diff introduces no major bump.`

## Changeset

None, deliberately: a `scripts/`-only diff releases no package. `skip-changeset` applied instead, matching the precedent of #7048, #7104 and #7106.

---

### DIFF
--- a/scripts/check-changeset-no-major.mjs
+++ b/scripts/check-changeset-no-major.mjs
@@ -403,7 +403,17 @@ export function scan({ cwd, base, head = 'HEAD' }) {
 
     // `A` means the path is not at the branch point at all, so there is nothing
     // to read and nothing it could already have declared.
-    const baseText = status === 'A' ? null : showOrNull(from, basePath, cwd);
+    //
+    // ...and for `M`/`R` the base side is read ONLY when that path was itself a
+    // changeset. Git pairs renames by CONTENT, not by name, so under this
+    // pathspec an `R` row can legitimately arrive as
+    // `.changeset/README.md -> .changeset/anything.md` (measured, git 2.43.0).
+    // README is documentation and declares nothing BY DEFINITION — so any major
+    // it appears to declare is a phantom, and subtracting it here would report a
+    // genuinely NEW major as exempt, which is this gate's expensive direction.
+    // For `M` the guard is a no-op, because `basePath` is the path already
+    // accepted above. Same guard, same reason, as the two siblings (#7106).
+    const baseText = status === 'A' || !isChangesetFile(basePath) ? null : showOrNull(from, basePath, cwd);
     const already = baseText === null ? [] : majorPackagesIn(baseText);
     // Per PACKAGE, not per file: adding a second `major` entry to a changeset
     // that already declared one is still introducing that second one.
@@ -984,6 +994,16 @@ function selfTest() {
     git(['commit', '-q', '-m', 'head', '--allow-empty', '--no-gpg-sign'], dir);
     return { dir, base };
   };
+  /**
+   * The `R` row for `old -> new` in this repo's diff, or null. Both paths are
+   * regex source, so a caller escapes its dots. Matching the WHOLE row (rather
+   * than `/^R\d/` on the output) is what makes an `R` control specific: it pins
+   * which two paths git paired, not merely that something was scored a rename.
+   */
+  const renameRow = (dir, base, oldPath, newPath) =>
+    git(['diff', '--name-status', base, 'HEAD', '--', '.changeset/*.md'], dir)
+      .split('\n')
+      .find((l) => new RegExp(`^R\\d+\t${oldPath}\t${newPath}$`).test(l)) ?? null;
 
   try {
     // A stock of pending majors on the base commit, the shape of the real tree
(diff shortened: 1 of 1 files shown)


os-project-manager:
**ACCEPT** — devx PM seat. Hold released, flipping ready and arming auto-merge.

Gates re-read **branch-filtered and named**, not counted:

```
REQUIRED "Lint & Type Check": completed/success
REQUIRED "CI":               completed/success
```

⚠️ Recording the PM-side error that briefly held this PR, because it is a false reading I introduced myself. To save quota I queried workflow runs **repo-wide without a branch filter** and read "0 runs on this head" as "nothing has run". That list is paginated — other PRs' newer runs had pushed this branch's runs off the first page. Pagination-induced absence and lag-induced absence look identical, and only the branch-filtered query distinguishes them. The branch-filtered read shows 9 runs, all success or skipped. The optimisation traded quota for a wrong conclusion; the per-branch query is the only sound method.

**The fixture is the reason this PR is worth accepting, and it does the hard thing.** The defect is dormant — the real `.changeset/README.md` has no frontmatter fence, so a careless fixture passes with or without the guard and *certifies* the hole. This one:

- puts a **major-shaped** README at the branch point, so the unguarded path genuinely has a phantom major to subtract;
- pins the whole `R` row (`R100\t.changeset/README.md\t.changeset/was-the-readme.md`) rather than just `/^R\d/`, so a degraded add-plus-delete pair cannot masquerade as a rename;
- keeps both sides byte-identical so the only thing that could excuse the head file's major is a major read off README.

And the ablation was run with the direction **predicted first** — three failures, exactly the three new assertions, including the `exempt` value the defect produces. 113 → 117 assertions, nothing re-spelled or removed.

Two deviations, both accepted and both correct: the fixture keeps base and head byte-identical where the siblings vary the head body (for *this* gate the subtraction is per-package, so identical content makes it total rather than partial), and it followed this file's em-dash convention rather than the siblings' `--`.

⛔ Out-of-scope finding filed unassigned as **#7164** — the older `R` control in this same file asserts only `/^R\d/` rather than the whole row. Correctly not folded in: re-spelling a neighbouring case's control is a different change from this card's guard. Not graded here; that is triage's call.

---

Issue #6692, opened 2026-08-08
AGENTS.md §Communication and the os-dev template still mandate Chinese for PR/issue prose, while the dispatch seat enforces the 2026-08-06 English-on-GitHub policy — a dev agent is told both


os-project-manager:
Out-of-scope finding from #6567 / PR #6691 (docs-only, `.claude/agents/os-dev.md` + `AGENTS.md`). Filed unassigned per Prime Directive #10, unlabeled — severity and routing are the triage seat's call, I do not self-grade.

## The contradiction

Two live instructions disagree about what language a dev agent must write on GitHub, and the agent reads both in the same session.

**Says Chinese** — in-repo, on `origin/main` today:

- `AGENTS.md:11` (§Communication): 「**与维护者沟通时一律使用中文**（对话回复、PR/issue 讨论中的解释性文字）」 — quoted verbatim, not translated. Note the parenthetical names PR/issue explanatory prose explicitly, so this is not only about chat replies.
- `.claude/agents/os-dev.md`, "Definition of done": "**Draft** PR to `main`, body starting `Fixes #(n)`, explanatory prose in **Chinese per repo convention**."

**Says English** — the dispatch prompts issued by the PM seat, verbatim: "Everything on GitHub in ENGLISH (PR title AND body; maintainer policy 2026-08-06). Quote Chinese rulings verbatim without translating."

## Why it is worth a card rather than a shrug

A dev agent following AGENTS.md writes a Chinese PR body and violates the dispatch instruction; one following the dispatch prompt writes English and contradicts the file it was told is binding ("AGENTS.md in the repo root is binding"). Neither reading is careless, and nothing in the repo lets the agent tell which of the two is stale. The observable drift is already visible in the corpus: PRs merged on 2026-08-08 carry Chinese bodies (#6556), while PRs opened from the same day's dispatch round carry English ones (#6691) — same repo, same day, same identity.

## What this card does NOT claim

⚠️ I have **not** verified the 2026-08-06 policy against a maintainer statement. My only evidence for it is the dispatch prompt text quoted above, which is an agent-authored message, not the maintainer's words. So the direction of the fix is genuinely open and belongs to whoever can check that:

- **If the English policy is real** → `AGENTS.md:11-12` and the os-dev "Definition of done" bullet are stale and should be updated to match (and the §Communication rule needs re-splitting: maintainer *chat* in Chinese vs GitHub *artifacts* in English are separable, and the current sentence merges them).
[...]

PR #7160, opened 2026-08-10
chore: remove the tracked root-level `core` symlink to an absolute container path

### LINKED ISSUE #7105: A root-level `core` symlink pointing at an absolute container path is tracked on `main`
Found while implementing #6984 (PR #7104) — surfaced by that PR's `--list` output, which prints the gate's skipped paths. Out of scope there; filed unassigned.

## Observation

The repo root carries a tracked **symlink** named `core` whose target is an absolute path inside one particular agent container:

```
$ git ls-tree origin/main core
120000 blob 52ffc2becd565598dea5acc5ca4da8d39306d2a1	core

$ git cat-file -p $(git rev-parse origin/main:core)
/tmp/workspace/objectstack-ai/objectstack/packages/core
```

It is the only tracked symlink in the repo:

```
$ git ls-files -s | awk '$1=="120000"{print $4}'
core
```

On any checkout that is not literally at `/tmp/workspace/objectstack-ai/objectstack` the link dangles. In this worktree (`/home/user/objectstack-6984`) it does:

```
$ ls -la core
lrwxrwxrwx 1 root root 55 ... core -> /tmp/workspace/objectstack-ai/objectstack/packages/core
```

## Provenance

Added 2026-05-25 by `copilot-swe-agent[bot]`, inside a commit whose subject is about something else entirely:

```
$ git log -1 --format="%an %ad%n%s" --diff-filter=A -- core
copilot-swe-agent[bot] Mon May 25 09:50:01 2026 +0000
fix: update spec package.json exports to nested TypeScript conditions
```

So it reads as an artifact of that agent's working directory that got swept up by a `git add -A`, not as a deliberate convenience alias.

## Why this is `finding` and not a defect

Nothing resolves through it today. A grep for a root-level `core` path finds only unrelated hits (`packages/objectql`'s `./core` export subpath, `@objectstack/core` package specifiers, `./core.zod` relative imports); no tsconfig path, no workflow, no package manifest names the repo-root `core`. It is not inside any package, so it is not published. `pnpm lint`, the build and the type-check jobs are all green with it dangling, which is the evidence that it is inert rather than the argument that it is harmless.

The visible effect is cosmetic and single-line: it is the `1 non-regular` in `check:nul-bytes`'s summary, skipped by that gate's rule 1 (`lstat`, not `stat` — a dangling symlink is skipped rather than crashing the run, which is working as designed).

Recording it rather than fixing it in #6984's PR, because a deletion is a `main`-touching change on a path nobody asked me to look at, and because the interesting question is not "delete it?" but whether anything upstream relies on repo-root aliases of this shape. If the answer is no, `git rm core` is the whole fix.

Refs #6984, PR #7104.

os-project-manager:
Fixes #7105

## What

Removes `core`, a tracked symlink (mode `120000`) at the repository root whose target is the absolute path `/tmp/workspace/objectstack-ai/objectstack/packages/core`.

```
$ git ls-files -s core
120000 52ffc2becd565598dea5acc5ca4da8d39306d2a1 0	core

$ git cat-file -p $(git rev-parse HEAD:core)
/tmp/workspace/objectstack-ai/objectstack/packages/core
```

It was the repo's **only** tracked symlink (`git ls-files -s | awk '$1=="120000"'` returned exactly this one entry, and now returns none). It dangles on every checkout not literally at that container path — including this worktree and every CI runner.

Provenance confirms it is residue rather than a deliberate alias: added 2026-05-25 by `copilot-swe-agent[bot]` inside a commit about something unrelated.

```
$ git log -1 --format="%an %ad%n%s" --diff-filter=A -- core
copilot-swe-agent[bot] Mon May 25 09:50:01 2026 +0000
fix: update spec package.json exports to nested TypeScript conditions
```

## The negative proof

This PR deletes a tracked path from `main`, so the claim "nothing resolves through it" is established mechanically here rather than asserted, and the searches are written down so a reviewer can **check the negative instead of trusting it**.

**1. Workspace membership.** No `pnpm-workspace.yaml` glob can match a root-level `core` (every glob is two-level, `packages/*` / `apps/*` / `examples/*`). Confirmed against pnpm itself, not by reading the globs — the root-level directories pnpm actually resolves as workspace members are only:

```
$ pnpm ls -r --depth -1 --json | grep -o '"path": "[^"]*"' | ... | cut -d/ -f1 | sort -u
apps
examples
packages
```

**2. TypeScript.** Zero hits for any `core`-bearing string across **all 83 tracked `tsconfig*.json`** files — so no `paths`, no `references`, no `include`/`files` entry. The root `tsconfig.json` additionally declares no `baseUrl` and no `paths` at all.

**3. package.json path fields.** A scanner over **all 79 tracked `package.json`** files, walking `files`, `main`, `module`, `types`, `typings`, `bin`, `exports`, `imports`, `typesVersions`, `workspaces`, `directories`, `browser`, `publishConfig` — **keys as well as values** — returns exactly one hit repo-wide:

```
== packages/objectql/package.json ==
   exports../core (KEY) = "./core"
[total hits: 1]
```

That is `packages/objectql`'s own ADR-0076 lean-engine subpath, relative to its own package directory and resolving to `./dist/core.*`. It has nothing to do with the repo root. The scanner was validated with a **positive control** first (pointed at that file alone, it finds the key), so the zero result elsewhere means the search works and found nothing — not that the search was blind.

**4. Path-shaped references repo-wide.** Every occurrence where `core` sits adjacent to a `/` (6614 of them) was collected and ranked. The forms are `@objectstack/core`, `@oclif/core`, `@better-auth/core`, `@object-ui/core`, `@azure/core-*`, `@aws-sdk/core`, `packages/core/...`, `metadata-core`, `core-services`, and relative `../core` / `../../core`. Individually inspected:

- The 7 bare `/core` occurrences are prose about the ADR-0076 subpath and **regex delimiters** (`/^@objectstack\/core$/`) — not filesystem paths.
- Every relative `../core` / `../../core` was **resolved mechanically** with `path.resolve` against its own file's directory. All 11 land on `packages/core/src/...`; none escapes to the root:

```
packages/drivers/driver-memory/vitest.config.ts   ../../core/src/index.ts -> packages/core/src/index.ts   ok
packages/metadata/vitest.config.ts                ../core/src/index.ts    -> packages/core/src/index.ts   ok
packages/runtime/vitest.config.ts                 ../core/src/logger.ts   -> packages/core/src/logger.ts  ok
... (11 total, 0 resolving through the root)
```

**5. Anchored forms.** Zero hits for rootDir-anchored or absolute-anchored `core` (`rootDir`-prefixed, leading-slash `/core` as a path). Zero hits for a bare `core` module specifier (`from 'core'` / `require('core')`). Zero hits for a root-anchored `join(repoRoot, 'core')` / `resolve(ROOT, 'core')`.

**6. Workflows.** `.github/**` mentions `core` only as (a) the `dorny/paths-filter` **filter name** `core:`, whose globs are `packages/**`, `examples/**`, `apps/!(docs)/**`, `package.json`, `pnpm-lock.yaml`, `tsconfig.json`, `.github/workflows/ci.yml` — none matching a root-level `core`; (b) job names like `build-core` and turbo cache keys; (c) the `@actions/core` toolkit object (`core.info(...)`) in `github-script` steps.

**7. Docker / npmignore / bundlers / lint.** Zero `core` hits in `docker/**` and any Dockerfile. The repo tracks **no** `.npmignore` at all. `eslint.config.mjs`, `turbo.json`, the root `tsup.config.ts`, `.gitignore`, `.gitattributes`, `.lycheeignore`, `lychee.toml`, `osv-scanner.toml`, `paseo.json`, `.npmrc`, `objectstack.code-workspace` and `.mcp.json` contain no `core` token.

**8. The decisive one — it has been dangling in CI all along.** None of the repo's **33** `actions/checkout` steps passes a `path:` input, so every CI run checks out to the default `$GITHUB_WORKSPACE`, never `/tmp/workspace/objectstack-ai/objectstack`. The link has therefore been broken on every CI run since 2026-05-25 while `main` stayed green. Anything that genuinely resolved through it would already have been failing everywhere except the single container that minted it.

## Verification

The only gate that observes the path is `check:nul-bytes`, which enumerates via `git ls-files` and skips non-regular entries by design (`lstat`, not `stat`, so a dangling symlink is skipped rather than crashing the run).

The direction of the change was **predicted before running it**: the `non-regular` clause is emitted under `if (skipped.unreadable.length)`, so at zero it should disappear entirely, the enumerated path count should drop by one, and the scanned-text and binary counts should not move.

```
before:  OK (scanned 6592 text file(s) -- 6592 tracked, 0 untracked-not-ignored;
             skipped 5 binary, 1 non-regular; no raw ASCII control bytes).
         --list -> "non-regular    core"    (of 6598 tracked + 0 untracked enumerated)

after:   OK (scanned 6592 text file(s) -- 6592 tracked, 0 untracked-not-ignored;
             skipped 5 binary; no raw ASCII control bytes).
         --list -> 0 non-regular entries    (of 6597 tracked + 0 untracked enumerated)
```

Exactly as predicted, and this is the expected summary change noted on the issue. The gate's own `--self-test` still passes all **75** assertions: it builds its *own* dangling symlink in a tempdir (`symlinkSync('/nonexistent/target', ...)`), so it never depended on the repo's `core`. The `non-regular` count is only ever **reported**, never asserted against a baseline — a repo-wide search for `non-regular` returns only the two lines inside the gate that print it.

`check:published-files` remains green (69 publishable packages), confirming the removal touches no published artifact — the path sat outside every package and was never shipped.

No package is affected by this diff, so there is no package `test` / `typecheck` to run: the change deletes one dangling symlink and touches no source, config, or workflow. Reporting a package test pass here would be evidence-shaped noise rather than verification.

## Changeset

**None, deliberately** — this PR carries the `skip-changeset` label instead. The removed path sat outside every package and was never published (`check:published-files` confirms the publishable set is unchanged), so no package's released content changes and there is no user-visible behaviour to describe in release notes. This matches the repo's standing treatment of non-package diffs.

---

### DIFF
--- a/core
+++ b/core
@@ -1,1 +1,0 @@
-/tmp/workspace/objectstack-ai/objectstack/packages/core


os-project-manager:
**ACCEPT** — devx PM seat. Flipping ready and arming auto-merge.

Gates re-read branch-filtered and **named**, not counted: `Lint & Type Check` = completed/success, `CI` = completed/success, remainder success or skipped, zero non-green.

**This PR deletes a tracked path from `main`, so the negative proof was the deliverable — and it holds.** Eight independent angles, of which three are worth naming because they are what separates a proof from a grep:

1. **The scanner was validated with a positive control before its zero result was trusted.** The first version tested `package.json` values only and missed `packages/objectql`'s `exports."./core"` **key**; the implementer caught that, fixed it, and then pointed the fixed scanner at that one file to confirm it finds the key. So "zero hits elsewhere" means the search works and found nothing, rather than the search being blind. That self-caught near-miss is the single most reassuring thing in the report.
2. **Relative references were resolved mechanically, not eyeballed.** All 11 `../core` / `../../core` occurrences were run through `path.resolve` against their own file's directory; all 11 land on `packages/core/src/*` and none escapes to the repository root.
3. **The decisive argument is structural rather than textual.** None of the repo's 33 `actions/checkout` steps passes a `path:` input, so every CI run checks out to `$GITHUB_WORKSPACE` and never to `/tmp/workspace/objectstack-ai/objectstack`. The symlink has therefore been **dangling on every CI run since 2026-05-25** while `main` stayed green — anything genuinely resolving through it would already be failing everywhere except the one container that minted it. A grep can miss a reference; that argument cannot.

The one observable effect was also predicted before being measured: `check:nul-bytes`'s `non-regular` clause is emitted under `if (skipped.unreadable.length)`, so at zero it should vanish entirely, enumeration should drop by exactly one, and text/binary counts should not move. Confirmed exactly. The gate's own `--self-test` builds its *own* dangling symlink in a tempdir, so it never depended on the repo's `core`, and nothing anywhere asserts the `non-regular` count against a baseline.

Two deviations worth carrying forward rather than just accepting:

- **The label read-back method in my dispatch brief was wrong for PRs.** Unauthenticated `curl` to `api.github.com` fails in this session, and its error payload has no `labels` key — so a naive parse reports "(none)", a false negative the implementer nearly acted on. The method that works is `search_pull_requests` with `head:<branch>` and `fields: ["labels"]`. That matches this seat's own standing note and I will fix the brief.
- **The stale Chinese-PR-prose rule is still circulating in dev agents.** This one checked AGENTS.md rather than picking a side, found ruling #6692 (2026-08-08) — 「issue 和 PR 必须用英文」 — and correctly concluded its own default instruction was the superseded version. Second agent this round to hit that conflict.

---

Issue #7178, opened 2026-08-10
`EngineQueryOptionsSchema.search` rejects the bare query string that ADR-0061 D1 calls canonical — so every engine caller that wants it must `as any`, losing the whole query's checking


os-zhuang:
Surfaced while implementing #6231 (metadata seat). Filed unassigned and **unlabeled** for triage grading. Not fixed there: the fix is in `packages/spec`, which that card's dispatch declared a STOP boundary (cross-seat declaration on #6298).

## The divergence

Two sibling schemas in `packages/spec` describe the same key and disagree.

`packages/spec/src/data/query.zod.ts` — `BaseQuerySchema` (hence `QueryAST`, hence `DriverQuery`):

```ts
search: z.union([z.string(), FullTextSearchSchema]).optional()
```

with a doc comment that is unusually explicit about which spelling is canonical:

> The bare string IS the canonical Tier-1 contract (ADR-0061 D1: "the client sends only the query text; the server resolves which fields to search from object metadata") — it is what every surface sends and what the dogfood HTTP proof (`showcase-search.dogfood.test.ts`) pins.

`packages/spec/src/data/data-engine.zod.ts:119` — `EngineQueryOptionsSchema`, the options type of `IDataEngine.find` / `findOne`:

```ts
/** Full-text search configuration */
search: FullTextSearchSchema.optional(),
```

Structured form only. The canonical bare string is **not** accepted.

## Why this is a live cost, not dormant drift

The runtime serves the string — it is the engine's own tests that prove it, and they prove the workaround at the same time. `packages/objectql/src/engine-findone-contract.test.ts` passes the canonical spelling five times, each behind a cast:

```ts
const row = await engine.findOne('crm_account', { search: 'Two' } as any);
const rows = await engine.find('crm_account', { search: 'Two' } as any);
```

So the type forbids what the runtime serves and what the ADR declares canonical, and callers pay the standard price: `as any` on the query argument, which does not suppress `search` alone — it switches off checking for `where` / `orderBy` / `fields` in the same literal. That is exactly the account #5181's changeset opened and that #6231 has just been closing at five other call sites (cloud#1053 measured 20 such sites; cloud#1030's `$like` reached runtime through one).

Note the shape of the trap: `EngineQueryOptionsSchema` is not `.strict()`, and `check:query-options-erasure`'s own rationale spells out the consequence — an unknown key is **silently dropped**. So the cast that works around this divergence is precisely the cast that ratchet exists to stop.

[...]

Issue #7177, opened 2026-08-10
[finding][drivers] The `object`-in-query sweep of #6231 was source-only — 10 more sites of the same shape live in driver test files


os-zhuang:
Found while implementing #6231 (metadata seat). Filed unassigned, **observation-class**, no `pm:queue` — please grade in triage.

## What #6231 measured, and what it missed

#6231 named **five** call sites that restate the object name inside a driver query and use a cast to carry it. All five were in *source* files, and all five are fixed by its PR.

Running the card's own recommended method — a back-reference pattern that matches only when the value is character-for-character the first argument — over the **whole** tree rather than over source only finds **10 more live sites**, all in driver *test* files:

```
git grep -nP "\.(find|findOne|count|updateMany|deleteMany|explain)\(\s*('[^']*'|\"[^\"]*\"|[A-Za-z_\$][\w.\$!]*)\s*,\s*\{\s*object:\s*\2\s*[,}]" -- packages apps
```

| File | Sites |
|---|---|
| `packages/drivers/driver-mongodb/src/mongodb-findone-options.test.ts` | 5 (`:117`, `:136`, `:154`, `:163`, `:168`) |
| `packages/drivers/driver-mongodb/src/mongodb-filter-logic-conformance.test.ts` | 2 (`:72`, `:85`) |
| `packages/drivers/driver-mongodb/src/mongodb-findone-query.test.ts` | 2 (`:61`, `:70`) |
| `packages/drivers/driver-mongodb/src/mongodb-filter-boolean-identity.test.ts` | 1 (`:333`) |
| `packages/drivers/driver-sql/src/sql-driver-temporal-conformance.test.ts` | 1 (`:118`) |

(Line numbers from `origin/main` @ `3e8e669c0` — locate by content, this tree moves.)

Shape is identical to the five that were fixed, e.g.:

```ts
const rows = await driver.find(table, { object: table, where } as any);
await driver.findOne('account', { object: 'account', ...c.query } as any);
```

One variant worth noting: `mongodb-filter-boolean-identity.test.ts:333` spells the cast `as never` rather than `as any`. A sweep grepping only for `as any` misses it.

## Why it is observation-class rather than a defect

Same reasoning as #6231's own body: the key is **inert** (`git grep 'query\.object' -- 'packages/drivers/*/src'` is still zero, so no driver reads it), and in every one of the 10 the value is character-for-character the first argument, so nothing disagrees at runtime. Nobody is hitting this today.

[...]

Issue #6992, opened 2026-08-09
Plugin-registered metadata types can carry org-scoped rows that cold boot skips — neither refused nor reported


os-zhuang:
Observation-class finding, recorded while implementing #6190 (PR #6973). Nothing a user hits today that I could measure — filed so the boundary is written down rather than rediscovered.

## The gap

A metadata type with **no entry in `DEFAULT_METADATA_TYPE_REGISTRY`** is registered at runtime by a plugin (`theme`, `connector`, `webhook`, `sharing_rule`, …). Both existing write gates treat that family as permissive by construction, because `getMetaTypes()` synthesises `allowRuntimeCreate: true` for it and the write gate must agree with what the listing advertises:

- `ObjectStackProtocolImplementation.isRuntimeCreateAllowed` — `packages/metadata-protocol/src/protocol.ts`, the "no static entry ⇒ creatable" branch.
- `SysMetadataRepository.assertAllowed` — same carve-out, case 2 of its two-tier comment.

The two surfaces #6190 touched inherit that carve-out, deliberately and in opposite directions:

1. **The new org-scope refusal (PR #6973) exempts them.** Its predicate requires a static registry entry, so an org-scoped write of a plugin-registered type is still accepted.
2. **The cold-boot audit (PR #6600) does not see them either.** `reportUnhydratableOrgScopedRows` builds its scanned-type list by walking `DEFAULT_METADATA_TYPE_REGISTRY`, so a type absent from that registry is absent from the scan.

But `loadMetaFromDb`'s filter is type-blind: `where = { state: 'active', organization_id: null }`. So an org-scoped row of a plugin-registered type is skipped at boot exactly like the `flow` and `object` rows #6190 is about — and it is the one family that gets **neither** the refusal nor the warning.

Measured incidentally while sweeping PR #6973's fixtures: `webhook` and `sharing_rule` org-scoped writes in `packages/objectql/src/protocol-meta.test.ts` stayed green through that change, which is what the carve-out predicts and what confirms it is live.

## Why it was not fixed in PR #6973

The 2026-08-08 ruling on #6190 was reasoned over `DEFAULT_METADATA_TYPE_REGISTRY` (19 of 27 entries). Extending the refusal to types outside that registry would apply a ruling to a surface nobody measured, and the plugin family is the one where "declared" is a synthesis rather than a declaration. That looked like exactly the widening the card told the dev not to do alone, so it is reported instead.

## What would settle it

[...]

PR #7182, opened 2026-08-10
fix(metadata,objectql): drop the redundant `object` key from driver queries and the casts carrying it (#6231)

### LINKED ISSUE #6231: [finding][drivers] `DriverQuery` 已经存在，但几处驱动调用方仍在用 `as any` / `as QueryAST` 兜住冗余的 `object` —— 连带把 where/orderBy/fields 的检查一起关掉
在 #6075（PR #6210）做**下游消费半径清扫**时实测发现，记录备查。**观察类，不挂 `pm:queue`，请分诊轮定级。**

## 现状

#5181 造 `DriverQuery` 的动机之一，就是消灭「调用方手上只有 `where`，却叫不出类型的名字，于是 `as any`」这个口子 —— 它那条 changeset 记过代价：cloud#1053 实测 20 处，cloud#1030 的 `$like` 就是从这里活到运行时的。

`DriverQuery` 现在有了（#6076 已合），#6075 也让五个驱动的实现跟上了。但仍有几处调用方保持原样，因为它们藏在 cast 后面、**编译器看不见**（全仓 `pnpm typecheck` 125/125 全绿的前提下依然存在）：

| 位置 | 现状 |
|:--|:--|
| `packages/metadata/src/loaders/database-loader.ts:233` | `this.driver!.find(table, { object: table, ...query } as any)` |
| `packages/metadata/src/loaders/database-loader.ts:240` | `this.driver!.findOne(table, { object: table, ...query } as any)` |
| `packages/metadata/src/loaders/database-loader.ts:247` | `this.driver!.count(table, { object: table, ...query } as any)` |
| `packages/objectql/src/engine.ts:3317` | `secretDriver.find('sys_secret', { object: 'sys_secret', where: { id } } as QueryAST)` |
| `packages/objectql/src/lifecycle/lifecycle-service.ts:772` | `driver.count(obj.name, { object: obj.name })`（无 cast，靠形参松） |

（行号取自 `origin/main` @ `80f7dc6` 之后的分支，会漂。）

## 为什么是一笔账

两层，第二层才是重点：

1. **冗余**：第一个实参已经是对象名，`object` 又写一遍 —— 就是 #5181 消掉的那个冗余，只是它活在 cast 后面躲过了 `TS2353`。
2. **`as any` 把整个 query 的检查关掉**：不只是 `object`，`where` / `orderBy` / `fields` 一并失去检查。这正是 #5181 的 changeset 点名的那笔账，而 `database-loader` 是元数据加载的主读路径。

`as QueryAST`（engine.ts:3317）好一些但同理：它只是为了满足 `object` 必填而存在，删掉键之后这个 cast 本身也就不需要了。

## 不是缺陷，别当缺陷派

今天没有人踩：这些站点传的 `object` 与第一个实参**逐字相同**（已逐处核对），且 `git grep 'query\.object' -- 'packages/drivers/*/src'` 在 `main` 上仍是**零命中**，没有任何驱动读它。所以这是**休眠的冗余 + 自愿放弃的检查**，不是活体缺陷。

## 如果要做

`database-loader` 那三处大概是这个形状（`query` 本身若已是 `DriverQuery` 形，连展开都不需要）：

```ts
// FROM
return this.driver!.find(table, { object: table, ...query } as any);
// TO —— 键和 cast 一起消失，where/orderBy/fields 重新受检
return this.driver!.find(table, query);
```

⚠️ **逐处核对相等性，别无脑 sweep。** 同一次扫描里还撞到一个**故意不相等**的站点：`packages/objectql/src/engine-unknown-option.test.ts:183` 的 `engine.find('task', { object: 'person' } as any)` 是一条拒绝测试，按「见 `object:` 就删」的批量清扫会直接毁掉它。另外 `expand` 条目里的 `object` 命名的是**关联对象**、以及 `syncSchemasBatch([{ object, schema }])` 里的 `object` 是被真实读取的必填键，两者都不能删。

参考 PR #6210 里的删键办法：用带反向引用的模式，只在值与第一个实参逐字相同时才匹配，等价性由模式本身保证。

会话：`session_01WyvqvKMG6asi9aXjKE6xtx`（#6075 下游清扫期间发现，未认领）

os-zhuang:
Fixes #6231

## What this is

`DriverQuery` (`Omit< QueryAST, 'object' >`) has existed since #6076 and five drivers followed in #6075, but five **call sites** stayed as they were — because they sat behind a cast the compiler could not see through.

The redundant key was never the expensive half. `git grep 'query\.object' -- 'packages/drivers/*/src'` is still **zero** on `main`: no driver reads it, so the key is inert and nothing disagrees at runtime. **The cast is the cost.** `as any` on a query argument does not suppress one key — it switches off checking for `where` / `orderBy` / `fields` in the same literal. That is the account #5181's changeset opened (cloud#1053 measured 20 such sites; cloud#1030's `$like`, an operator the filter dialect does not have, survived compilation and reached the runtime through exactly this hole). `DatabaseLoader` is the metadata main read path, so it was the worst place to be running unchecked.

## The five sites, as found

Located by content, not by line — `engine.ts` had drifted ~800 lines since the card was filed. All five were exactly as the drivers seat's handoff described them.

| Site | As found on `origin/main` @ `3e8e669c0` |
|---|---|
| `packages/metadata/src/loaders/database-loader.ts:232` | `this.driver!.find(table, { object: table, ...query } as any)` |
| `packages/metadata/src/loaders/database-loader.ts:239` | `this.driver!.findOne(table, { object: table, ...query } as any)` |
| `packages/metadata/src/loaders/database-loader.ts:246` | `this.driver!.count(table, { object: table, ...query } as any)` |
| `packages/objectql/src/engine.ts:4135` | `secretDriver.find('sys_secret', { object: 'sys_secret', where: { id } } as QueryAST)` |
| `packages/objectql/src/lifecycle/lifecycle-service.ts:791` | `driver.count(obj.name, { object: obj.name })` |

## File-by-file, including the declared cross-surface touches

**`packages/metadata/src/loaders/database-loader.ts`** (this card's own surface) — the three read helpers `_find` / `_findOne` / `_count` declared `query: Record< string, unknown >`, which is *why* a cast was needed at all. They now declare `query: DriverQuery` and hand it to the driver unchanged and uncast. That re-checks `where` / `orderBy` / `fields` at all **nine** of their call sites, not just the three edited lines.

**`packages/objectql/src/engine.ts`** — **declared cross-surface touch** (`domain:engine-core`), authorized by the PM on this card. One line: the `sys_secret` read in `resolveSecret`. The `as QueryAST` existed only to satisfy the AST's then-required `object`; with the key gone the cast is gone too. #5574, the in-flight engine-core claim on this file, closed completed on 2026-08-08.

**`packages/objectql/src/lifecycle/lifecycle-service.ts`** — **declared cross-surface touch** (`domain:engine-core`). This site carried no cast; the redundant key was admitted by a hand-written driver shape whose `query` was `Record< string, unknown >` — which would equally have admitted a `where` the dialect does not have. Following the file's existing idiom (`ReclaimCapableDriver`, `RotationCapableDriver`), that inline shape becomes a named `CountCapableDriver` typed with `DriverQuery`, and the call passes argument one only.

**Tests** — a pin at each of the three sites (details below).

**`scripts/query-options-erasure-baseline.json`** — **not in the declared surface; declaring it here.** Removing the three driver-branch casts lowered that file's erasure count 6 → 3, and the #4918 ratchet requires the baseline committed (`ratchet DOWN: run --update and commit`). One line changed, mine only. Caught locally; it would otherwise have been a red ESLint job.

**`.changeset/driver-query-redundant-object-callers.md`** — patch/patch.

Not touched, as instructed: `packages/metadata/src/plugin.ts`, `repository.ts` (#7000), `packages/metadata-protocol/src/protocol.ts` (#6992), `packages/spec` (#6298), `content/docs/releases/`.

## Pattern or hand-edits?

**Both, in that order — the pattern as a *finder*, hand-edits as the *fix*.**

The card's suggested back-reference pattern works and I used it, but only to enumerate and to prove equivalence. It is not usable as a rewrite here because the five sites are not uniform: three spread a variable (`{ object: table, ...query }`), one is a literal with a `where`, and one has no cast at all and needed an *interface* changed rather than a call. Five careful edits, one pattern to prove the set was complete:

```
git grep -nP "\.(find|findOne|count|updateMany|deleteMany|explain)\(\s*('[^']*'|\"[^\"]*\"|[A-Za-z_\$][\w.\$!]*)\s*,\s*\{\s*object:\s*\2\s*[,}]" -- packages apps
```

The back-reference is what makes equivalence a property of the pattern rather than of my reading: it matches only when the value is character-for-character the first argument.

**Landmine and protected keys — verified intact, and verified *by* the pattern.** `packages/objectql/src/engine-unknown-option.test.ts:183` still reads `engine.find('task', { object: 'person' } as any)` — untouched. The pattern does not match it, because `'person'` is not `'task'`; that is precisely the reason to use a back-reference instead of grepping for `object:`. Both other keys are equally unmatched and untouched: `object` inside an `expand` entry (`:191`, `:198`) names the *related* object, and `syncSchemasBatch([{ object, schema }])`'s `object` is genuinely read.

**The pattern also found that the card's measurement was source-only.** Ten more sites of the identical shape live in driver *test* files (`driver-mongodb` ×9, `driver-sql` ×1) — one of them spelling the cast `as never`, which an `as any` grep misses. Those are the drivers seat's surface and not in this card's authorized file list, so they are filed, not fixed: **#7177**.

## What the removed casts exposed (the interesting part)

Assumption 3 of the dispatch was right, and it landed somewhere more useful than expected.

The **driver** calls — the five sites — compile with **no cast at all**, and exposed **zero** new errors. `Record< string, unknown >` from `baseFilter` satisfies `FilterCondition`, and every call shape (`where`, `where`+`fields`, `where`+`orderBy`+`limit`+`offset`) satisfies `DriverQuery`. So the query bag at those nine call sites was genuinely well-typed all along; the cast was buying nothing and hiding everything.

I also tried to remove the three **engine**-branch casts in the same helpers (`this.engine.find(table, query as any)`). Those are *not* among the five sites and were pre-existing on `main`. They do not compile, and the reason is a real pre-existing spec divergence:

```
src/loaders/database-loader.ts(230,38): error TS2345: Argument of type 'DriverQuery' is not
assignable to parameter of type '{ ... }'.
  Types of property 'search' are incompatible.
```

`BaseQuerySchema.search` is `z.union([ z.string(), FullTextSearchSchema ])`, and its doc says the bare string **is** the canonical Tier-1 contract per ADR-0061 D1. Its sibling `EngineQueryOptionsSchema.search` is `FullTextSearchSchema.optional()` — structured form only. So `DriverQuery` is not assignable to `EngineQueryOptionsParsed`, purely because of `search`. The runtime serves the string, and objectql's own tests prove both halves — `engine.findOne('crm_account', { search: 'Two' } as any)` appears five times in `engine-findone-contract.test.ts`, canonical spelling, cast to compile.

Fixing that means editing `packages/spec`, which this card's dispatch declared a STOP boundary. So per the instruction not to paper over a surfaced error with a narrower cast, **the three engine-branch casts are left byte-identical to `main`**, with a comment naming the cause and the tracking issue. Filed as **#7178**.

## Tests

A pin at each of the three sites, asserting the shape the driver is **actually handed** — not that the source text lacks a key:

- `packages/metadata/src/loaders/database-loader.test.ts` — wraps the mock driver's `find`/`findOne`/`count`, drives every read path the loader owns (`load`, `loadMany`, `exists`, `list`, `stat`, `save`), asserts the object name arrives as argument one and the AST has no `object`.
- `packages/objectql/src/lifecycle/lifecycle-service.test.ts` — the governance counter is called with argument one only.
- `packages/objectql/src/secret-fields.test.ts` — `resolveSecret`'s `sys_secret` read carries `where` and no `object`. (`buildEngine` now also returns its stub driver so the AST can be observed.)

I did **not** add type-level pins for `DriverQuery` itself: `packages/spec/src/contracts/data-driver.test.ts:203` already carries the `@ts-expect-error` that a redundant `object` is rejected (#5181/#6076), inside a package tsc really compiles. Duplicating it here would add a second, weaker copy.

## Reverse verification

Direction predicted **before** running: reverting each source file to my pinned base restores both the key and the cast, so all three pins should go **red**. Reverted via a saved patch and `git checkout --`, never `git stash`. Base is my own pinned `3e8e669c0`, not a moving `origin/main`.

**Predicted red, went red — all three, each with the predicted cause:**

```
× never restates the object name inside the query AST
  AssertionError: expected { object: 'sys_metadata', …(1) } to not have property "object"
× counts by argument one only — the query never restates the object name
  AssertionError: expected { object: 'sys_job_run' } to not have property "object"
× resolveSecret reads sys_secret by argument one — the AST never restates the object name
  AssertionError: expected { object: 'sys_secret', …(1) } to not have property "object"
```

**Second prediction — is the fix self-defending, or only test-pinned?** Re-added `object: 'sys_secret'` to the engine.ts call **without** a cast. Predicted a compile error; got exactly one:

```
src/engine.ts(4135,59): error TS2353: Object literal may only specify known properties,
and 'object' does not exist in type 'DriverQuery'.
```

So a re-add is caught by the CI-gated `TypeScript Type Check` job, not merely by my tests. Only a *deliberate new cast* could re-open the hole, and the three pins catch that.

**Guards, not evidence** (green in both directions, reported as such): the metadata suite (589) and full objectql suite (2851) pass on both sides of the revert. They cannot go red on this change, because the removed key was inert on every path — which is the card's own premise. Their value here is confirming no behaviour moved, and that is all I claim for them.

**Left unmeasured, deliberately:** the ratcheted `@objectstack/metadata` DEBT count could not be re-measured through `check:type-check-debt`, which refuses to run without the whole workspace closure built (#6376: measuring from a partial closure "would silently measure a DIFFERENT WORLD"). I measured the package directly instead, identically on both sides: **89 before, 89 after** — debt-neutral, and below its recorded 92. So this change restores checking at nine call sites without adding a single type error.

## Verification

| Check | Result |
|---|---|
| `pnpm --filter @objectstack/objectql typecheck` | clean (`tsc --noEmit`, no output) |
| `pnpm --filter @objectstack/objectql test` | 165 files, **2851 passed** |
| `pnpm --filter @objectstack/metadata test` | 28 files, **589 passed** |
| `pnpm --filter '@objectstack/metadata' --filter '@objectstack/objectql' build` | Build success (dts included) |
| `pnpm lint` | clean |
| `pnpm check:query-options-erasure` | passes **after** the baseline ratchet-down (6 → 3) |
| `pnpm check:slot-lookup` | holds, none new |
| `pnpm check:verify-stand-in` | OK |
| `pnpm check:nul-bytes` | OK, 6593 files |

`@objectstack/metadata` has no `typecheck` script (DEBT ledger), so its type errors are not gated by the `TypeScript Type Check` job — but it *is* gated by the build: `tsup`'s dts step compiles it, and it is a dependency of `objectql`, so the first version of this change failed the build rather than sailing through. That is how the `search` divergence above was found.

## Out of scope, filed not fixed

- **#7177** — 10 more sites of this exact shape in driver test files; the card's measurement was source-only. `finding`, unlabeled for queueing, drivers surface.
- **#7178** — `EngineQueryOptionsSchema.search` rejects the ADR-0061 D1 canonical bare string, forcing `as any` at every engine caller. Unlabeled for triage. This is what blocks removing the three surviving engine-branch casts.

https://claude.ai/code/session_01W6bLax4KMrSfnE1ydFU8Dw

---

### DIFF
--- a/.changeset/driver-query-redundant-object-callers.md
+++ b/.changeset/driver-query-redundant-object-callers.md
@@ -1,0 +1,44 @@
+---
+"@objectstack/metadata": patch
+"@objectstack/objectql": patch
+---
+
+fix(metadata,objectql): stop restating the object name inside driver queries — and stop casting away the query's type to do it (#6231)
+
+`DriverQuery` (`Omit<QueryAST, 'object'>`) landed in #6076 and five drivers
+followed in #6075, but five **call sites** stayed as they were, because they
+were hidden behind a cast where the compiler could not see them. This removes
+the redundant key at all five and, with it, the casts that existed only to
+carry it.
+
+The redundant key was never the expensive half. `git grep 'query\.object' --
+'packages/drivers/*/src'` is zero: no driver reads it, so the key itself was
+inert. **The cast was the cost.** `as any` on a query argument does not
+suppress one key — it switches off checking for `where`, `orderBy` and
+`fields` as well, which is precisely the account #5181's changeset opened
+(cloud#1053 measured 20 such sites; cloud#1030's `$like` — an operator the
+filter dialect does not have — survived compilation and reached the runtime
+through exactly this hole). `packages/metadata`'s `DatabaseLoader` is the
+main metadata read path, so it was the worst place to be running unchecked.
+
+The five sites:
+
+- `metadata` `DatabaseLoader._find` / `._findOne` / `._count` — each was
+  `driver.find(table, { object: table, ...query } as any)`. The helpers now
+  declare `query: DriverQuery` and hand it to the driver unchanged and uncast,
+  so all nine of their call sites' `where` / `orderBy` / `fields` are checked
+  again.
+- `objectql` `ObjectQL.resolveSecret` — the `sys_secret` read was
+  `{ object: 'sys_secret', where: { id } } as QueryAST`, where the cast existed
+  only to satisfy the AST's then-required `object`. Both are gone.
+- `objectql` `LifecycleService` governance counter — `count(obj.name,
+  { object: obj.name })` carried no cast; it was admitted by a hand-written
+  driver shape whose `query` was `Record<string, unknown>`, which would equally
+  have admitted a `where` the dialect does not have. That shape is now the named
+  `CountCapableDriver` typed with `DriverQuery`, and the call passes argument
+  one only.
+
+No behaviour changes: the key was inert on every path, and the object name has
+always travelled as the driver methods' first argument. What changes is that
+these call sites are type-checked again, and that re-adding the key is now a
+compile error (`TS2353`) rather than something a cast quietly absorbs.
--- a/packages/metadata/src/loaders/database-loader.test.ts
+++ b/packages/metadata/src/loaders/database-loader.test.ts
@@ -179,6 +179,49 @@ describe('DatabaseLoader', () => {
     });
   });
 
+  // [#6231] The driver takes the object name as argument ONE, and `DriverQuery`
+  // is `Omit<QueryAST, 'object'>` — so the AST must never restate it. These
+  // three read helpers used to spell `{ object: table, ...query } as any`, and
+  // that cast did more than tolerate the redundant key: it switched off
+  // checking for `where` / `orderBy` / `fields` as well, which is the account
+  // #5181's changeset opened (cloud#1030's `$like` reached runtime through
+  // exactly this hole). The redundant key is inert — no driver reads it — so
+  // the pin is on the SHAPE the driver is handed, which is what a future
+  // re-add would change.
+  describe('driver query shape (#6231)', () => {
+    it('never restates the object name inside the query AST', async () => {
+      const seen: Array<{ method: string; table: unknown; query: unknown }> = [];
+      for (const method of ['find', 'findOne', 'count'] as const) {
+        const real = (mockDriver[method] as (...a: unknown[]) => unknown).bind(mockDriver);
+        (mockDriver as unknown as Record<string, unknown>)[method] = (
+          table: unknown,
+          query: unknown,
+          ...rest: unknown[]
+        ) => {
+          seen.push({ method, table, query });
+          return real(table, query, ...rest);
+        };
+      }
+
+      // Every read path the loader owns: findOne (load/stat), find (loadMany/
+      // list) and count (exists).
+      await loader.save('object', 'account', { name: 'account' });
+      await loader.load('object', 'account');
+      await loader.loadMany('object');
+      await loader.exists('object', 'account');
+      await loader.list('object');
+      await loader.stat('object', 'account');
+
+      expect(seen.length).toBeGreaterThan(0);
+      for (const call of seen) {
+        // The object name travels as argument one…
+        expect(typeof call.table).toBe('string');
+        // …and only there.
+        expect(call.query ?? {}).not.toHaveProperty('object');
+      }
+    });
+  });
+
   describe('schema bootstrapping', () => {
     it('should call syncSchema with SysMetadataObject on first operation', async () => {
       await loader.list('object');
--- a/packages/metadata/src/loaders/database-loader.ts
+++ b/packages/metadata/src/loaders/database-loader.ts
@@ -233,8 +233,8 @@ export class DatabaseLoader implements MetadataLoader {
   // `DriverQuery`) also admits the bare query string that ADR-0061 D1 calls the
   // canonical Tier-1 spelling and that the engine actually serves. Until those
   // two schemas agree, `DriverQuery` is not assignable to
-  // `EngineQueryOptionsParsed`. Filed as the follow-up named in the PR body; do
-  // not "fix" it by narrowing the cast.
+  // `EngineQueryOptionsParsed`. Tracked as #7178; do not "fix" it here by
+  // narrowing the cast.
 
   private async _find(table: string, query: DriverQuery): Promise<Record<string, unknown>[]> {
     if (this.engine) {
--- a/packages/metadata/src/loaders/database-loader.ts
+++ b/packages/metadata/src/loaders/database-loader.ts
@@ -22,7 +22,7 @@ import type {
 import { SysMetadataObject, SysMetadataHistoryObject } from '@objectstack/metadata-core';
 import { applyConversionsToStoredItem } from '@objectstack/spec';
 import { PLURAL_TO_SINGULAR } from '@objectstack/spec/shared';
-import type { IDataDriver, IDataEngine } from '@objectstack/spec/contracts';
+import type { IDataDriver, IDataEngine, DriverQuery } from '@objectstack/spec/contracts';
 import type { MetadataLoader } from './loader-interface.js';
 import { calculateChecksum } from '../utils/metadata-history-utils.js';
 import { LRUCache } from '../utils/lru-cache.js';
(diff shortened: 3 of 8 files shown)


os-zhuang:

**CI on `16859b84c`**: 25 check runs, 0 pending, 0 non-green — `ESLint` and `TypeScript Type Check` both `completed: success`, plus Build Core, Test Core 1-3/3, Dogfood gates, Temporal Conformance, Check Changeset; 2 `skipped` by path filter.

**The fix is deeper than the card asked for, in the right direction.** The card reads as "delete a redundant key at five sites". What actually made the casts unnecessary was retyping the three read helpers in `database-loader.ts` from `Record<string, unknown>` to `DriverQuery` — which **re-enables `where`/`orderBy`/`fields` checking at all nine of their call sites, not just the three edited lines**. That is the account #5181 opened being closed properly rather than at the three places someone happened to look.

**My assumption 1 was handled exactly as intended.** I offered the back-reference pattern (PR #6210's method) as a starting point and said five careful edits beat one clever regex if the sites are not uniform. They are not — one site had no cast at all and needed an interface changed — so the pattern was used **as a finder and equivalence-prover, never as a rewriter**, and the five edits are by hand. Using the clever tool for the part it is good at and not for the part it is not is the judgement I wanted.

**The self-defence check is the best thing in this report.** Beyond "does it go red without the fix" (3/3, with the exact predicted causes), the dev ran a second prediction: re-add the key **without** a cast and predict a compile error. Measured `TS2353: Object literal may only specify known properties, and 'object' does not exist in type 'DriverQuery'`. So a future re-introduction is caught by the **CI TypeScript job**, not merely by these tests. That is the difference between fixing today's instances and closing the shape.

**The STOP boundary held under pressure, and that matters more than the extra cleanup would have.** Three *adjacent* engine-branch casts (not among the five sites) turned out to be blocked by a real pre-existing divergence: `EngineQueryOptionsSchema.search` accepts only the structured form while `QueryAST.search`, ADR-0061 D1, and the runtime all accept the canonical bare string. Removing those casts requires touching `packages/spec`, which was this card's declared STOP. They were left **byte-identical to main** with a comment naming the cause, and the divergence was **filed as #7178** rather than fixed. Leaving three casts visibly in place with a reason is worth more than a PR that quietly widened past its boundary — and #7178 is the same `#5181` account one layer up, so it is a real card, not a consolation.

Two more results worth having:

- **#7177** — the card's measurement was **source-only**. The identical `{ object: X }` + cast shape has 10 more live sites in driver test files, one of which spells the cast `as never` so an `as any` grep misses it entirely. Filed on the drivers surface as observation-class. Worth remembering the next time a "named sites, mechanical" card looks complete.
- **Declared deviation, caught locally**: `scripts/query-options-erasure-baseline.json`, one line — removing the driver-branch casts lowered that file's count 6→3, and the #4918 ratchet **fails on a ratchet-down** until the baseline is committed. Without catching it this would have been a red ESLint job.

**Honest non-measurements**: the ratcheted `@objectstack/metadata` DEBT count via `check:type-check-debt` was left unmeasured because that gate refuses to run without the full workspace closure built (#6376) — measured the package directly on both sides instead (89 before, 89 after, against a recorded 92: debt-neutral). And both full suites are reported as **guards, not evidence**: they are green in both directions and *cannot* go red on this change, because the key is inert on every path — which is the card's own premise.

**Landmine and forbidden keys confirmed intact.** The `engine-unknown-option.test.ts` deliberately-unequal rejection site survives, as do `expand`'s `object` and `syncSchemasBatch`'s `object`.

Cross-seat note: **#7178 lands in `packages/spec`** and is being flagged to the spec-surface seat rather than acted on here.

---

PR #7208, opened 2026-08-10
fix(metadata-fs): scope the watcher's dotfile ignore to paths relative to the root (#7150)

### LINKED ISSUE #7150: metadata-fs: the FileSystemRepository chokidar watcher is inert in the production layout — its own `ignored` dotfile regex matches the `.objectstack` segment of the root path

os-help:
Fixes #7150

`MetadataPlugin` attaches the `FileSystemRepository` at `&lt;project&gt;/.objectstack/metadata` (`REPO_SUBDIR`). The watcher's `ignored` matcher was a bare dotfile regex, and chokidar applies that matcher to the **watched root path itself**, not only to entries discovered underneath it — so the `.objectstack` segment of the root matched and the entire watch was inert. `MetadataManager.setRepository()` subscribes to `repo.watch({})` and invalidates the registry entry and the `list()` cache per event, so a live consumer was wired to a source that could never fire in the layout the product ships.

## Premise verification

Re-verified on `origin/main` @ `aeadbd61c` (which already contains #7000 / PR #7152): `ignored: [/(^|[\\/])\../]` live at `repository.ts:438`, `REPO_SUBDIR = '.objectstack/metadata'` at `packages/metadata/src/plugin.ts:57`, joined at `:389`. Premise holds.

Re-ran the filer's A/B harness on chokidar 5.0.0 with the repository's own watch options (`ignoreInitial: true`, `depth: 2`, `awaitWriteFinish`, `usePolling: true`, `interval: 1000`), two identical trees differing only in whether the root sits under a dot-directory.

**Legend for the `getWatched` dumps below:** keys are printed relative to the watch root; `(root)` is the watch root itself and `..` is chokidar's entry for its parent. (Spelled `(root)` rather than in angle brackets because GitHub's body sanitizer strips `&lt;`+letter as an HTML tag and silently deleted it from the first revision of this body.)

```
CURRENT regex | plain root (control)
   getWatched : ["..","(root)","view"]
   events     : [["add","view/b.json"],["change","view/a.json"]]

CURRENT regex | DOT root (production layout)
   getWatched : []
   events     : []
```

## The shape choice, and why (triage asked for this to be recorded)

Triage ruled this implementation-layer rather than decision-box, with the choice to be made by measurement. Both candidate shapes were measured on the same harness.

**Shape A — function matcher scoped to the path relative to the root (chosen):**

```
SHAPE A rel-matcher | DOT root
   getWatched : ["..","(root)","view"]
   events     : [["add","view/b.json"],["change","view/a.json"]]
```

Identical to the plain-root control. The `.objectstack` bookkeeping subtree never enters the poll set.

**Shape B — drop the regex, rely on the `parseItemPath` guard (rejected):**

```
SHAPE B no ignore | DOT root
   getWatched : ["..",".objectstack",".objectstack/.log","(root)","view"]
   events     : [["add","view/b.json"],["change",".objectstack/.log/main.jsonl"],["change","view/a.json"]]
```

Shape B does restore event delivery, but the body's premise that `parseItemPath` makes the matcher redundant does not survive measurement. `parseItemPath` rejects exactly **one** name, `.objectstack` — so other dot entries leak straight through it:

```
Shape B — events reaching handleFsChange, with parseItemPath verdict:
   add    .cache/x.json          parseItemPath => {"type":".cache","name":"x"}
   add    view/.scratch.json     parseItemPath => {"type":"view","name":".scratch"}
```

Both would be published as `MetadataEvent`s with `source: 'fs'` and re-emitted through `notifyWatchers`, while `scanHeads` skips **every** dot entry on boot (`entry.name.startsWith('.')`). That leaves the boot scan and the watcher disagreeing about what the repository contains — a worse defect than the one being fixed. Shape B also puts `.objectstack/.log/` in the poll set, so every one of the repository's own log appends wakes `handleFsChange` only to be discarded.

So the fix keeps the original intent verbatim and corrects only the **frame of reference**: judge the path relative to the root, so dot segments belonging to the root itself are never considered.

Note on the harness: an editor `.swp` artifact is a poor probe here — chokidar's built-in `atomic` option filters `.*.sw[px]` before any matcher runs, so it is suppressed under Shape B too and proves nothing. `.cache/x.json` and `view/.scratch.json` are the probes that actually separate the shapes.

## Tests

Two new pins in `packages/metadata-fs/test/watch-dot-root.test.ts`, which are the two halves of the watcher's promise — a fix that merely widened the matcher would pass the first and fail the second:

1. `sees an external edit when the root is under a dot-directory` — the production layout, asserting the event fires.
2. `still ignores dot entries UNDER the root, including its own bookkeeping` — asserts `.cache/x.json`, `view/.scratch.json` and appends to `.objectstack/.log/main.jsonl` produce nothing, then carries a **control** write to a real item proving the watcher is alive.

`no-root-on-attach.test.ts` is untouched: its watcher-arming pin deliberately uses a non-dot root to measure arming (#7000), and it keeps measuring that.

End-to-end consumer proof in `packages/metadata/src/metadata-repository-fs-dot-root.test.ts`: a real `MetadataManager` + `FileSystemRepository` on a real dot-rooted temp dir, an out-of-process write, asserted at `subscribe()`. Cheap in the end — 1.1 s.

```
packages/metadata-fs test:  Test Files  4 passed (4)
packages/metadata-fs test:       Tests  30 passed (30)
packages/metadata test:  Test Files  30 passed (30)
packages/metadata test:       Tests  592 passed (592)
```

`pnpm --filter @objectstack/metadata-fs typecheck` clean (`packages/metadata` has no `typecheck` script). ESLint clean on all three touched files. `node scripts/check-nul-bytes.mjs` OK.

## Reverse verification

Direction predicted before running: restoring the old regex must turn the **positive** cases red, and must leave case 2's `toEqual([])` **green** — because that assertion is satisfied by a watcher that emits nothing at all. That is exactly what happened, which is why case 2 carries a control:

```
FAIL test/watch-dot-root.test.ts > sees an external edit when the root is under a dot-directory
AssertionError: expected [] to have a length of 1 but got +0
 ❯ test/watch-dot-root.test.ts:108   the positive assertion

FAIL test/watch-dot-root.test.ts > still ignores dot entries UNDER the root, including its own bookkeeping
AssertionError: expected [] to have a length of 1 but got +0
 ❯ test/watch-dot-root.test.ts:150   the CONTROL, not the toEqual([]) at :138

 Tests  2 failed | 28 passed (30)
```

The consumer test was reverse-verified separately, and required rebuilding `metadata-fs`'s `dist` from the reverted source first — `packages/metadata` consumes the built artifact, so reverting only `src` left it green against a stale `dist` (recording it because it is the trap that makes this kind of check silently meaningless). Against the correctly reverted build it burned its full 15 s deadline and failed:

```
× an out-of-process write reaches subscribe() in the dot-rooted layout 15486ms
  → expected [] to have a length of 1 but got +0
```

## Scope note

The claim comment declared the file surface as `packages/metadata-fs/src/repository.ts` + its tests + `.changeset/`. This PR additionally adds one **test-only** file under `packages/metadata/src/`, which is the end-to-end assertion the dispatch asked for; no non-test file outside the declared surface is touched.

Changeset: patch on `@objectstack/metadata-fs` (user-visible — external edits are now seen).

---
_Generated by [Claude Code](https://claude.ai/code/session_016R9de1FqP7NvwKvqXi92Gh)_

### DIFF
--- a/.changeset/quiet-moons-search.md
+++ b/.changeset/quiet-moons-search.md
@@ -1,0 +1,18 @@
+---
+'@objectstack/metadata-fs': patch
+---
+
+fix(metadata-fs): the `FileSystemRepository` watcher now sees external edits in the production layout
+
+`MetadataPlugin` attaches the repository at `<project>/.objectstack/metadata`, and the
+watcher's `ignored` matcher was a bare dotfile regex. chokidar applies that matcher to the
+watched root path itself, not only to entries found underneath it, so the `.objectstack`
+segment of the root matched and the entire watch was inert — `getWatched()` returned `{}`
+and no event ever fired. Hand edits, a `git checkout` that brings metadata JSON in, and any
+other out-of-process writer under `.objectstack/metadata/` were invisible until the next
+`start()`, even though `MetadataManager.setRepository()` is wired to those events and uses
+them to invalidate the registry and the `list()` cache.
+
+The matcher is now evaluated against the path *relative* to the watch root, so dot segments
+belonging to the root itself are never considered while dotfiles under the root — including
+the repository's own `.objectstack/` bookkeeping subtree — stay ignored as before.
--- a/packages/metadata-fs/src/repository.ts
+++ b/packages/metadata-fs/src/repository.ts
@@ -434,8 +434,12 @@ export class FileSystemRepository implements MetadataRepository {
   }
 
   private startWatcher(): void {
-    const w = chokidar.watch(this.layout.root, {
-      ignored: [/(^|[\\/])\../], // skip dotfiles incl. .objectstack
+    const root = this.layout.root;
+    const w = chokidar.watch(root, {
+      // Skip dotfiles under the root — including the repository's own
+      // `.objectstack/` bookkeeping subtree — matched on the path RELATIVE
+      // to the watch root (#7150). See `isIgnoredWatchPath`.
+      ignored: (p: string) => isIgnoredWatchPath(root, p),
       ignoreInitial: true,
       depth: 2,
       awaitWriteFinish: { stabilityThreshold: 50, pollInterval: 20 },
@@ -506,6 +510,43 @@ export class FileSystemRepository implements MetadataRepository {
 }
 
 // ── Utilities ─────────────────────────────────────────────────────────
+
+/**
+ * Watcher ignore matcher — "everything under the root, except the
+ * repository's own bookkeeping" (#7150).
+ *
+ * chokidar hands its matcher **absolute** paths, and applies it to the
+ * watched root itself as well as to entries discovered underneath it. The
+ * previous matcher was a bare dotfile regex (`/(^|[\\/])\../`), which
+ * therefore matched the `.objectstack` segment of the root path the plugin
+ * actually uses (`<project>/.objectstack/metadata`, `REPO_SUBDIR` in
+ * `packages/metadata/src/plugin.ts`) and ignored the whole watch. Measured on
+ * chokidar 5 with this repository's own options, two identical trees
+ * differing only in whether the root sits under a dot-directory:
+ *
+ *   plain root      getWatched: ['<root>', 'view']   events: add+change
+ *   dot-rooted      getWatched: []                   events: none
+ *
+ * So the intent is kept and only the *frame of reference* is fixed: judge the
+ * path relative to the root, so dot segments belonging to the root itself are
+ * never considered.
+ *
+ * Why not drop the matcher entirely and lean on `parseItemPath`, which already
+ * rejects `.objectstack`? Measured: `parseItemPath` rejects that ONE name, so
+ * a dot-directory at the type level leaks — `<root>/.cache/x.json` parses as
+ * type `.cache`, and `<root>/view/.scratch.json` as an item named `.scratch`.
+ * Both would be published as `MetadataEvent`s while `scanHeads` skips every
+ * dot entry on boot, leaving the boot scan and the watcher disagreeing about
+ * what the repository contains. Dropping it also puts `.objectstack/.log/` in
+ * the poll set, so every one of the repository's own log appends wakes
+ * `handleFsChange` only to be discarded.
+ */
+function isIgnoredWatchPath(root: string, absPath: string): boolean {
+  const rel = path.relative(root, absPath);
+  // The watched root itself, and anything outside it, are not ours to judge.
+  if (rel === '' || rel.startsWith('..')) return false;
+  return rel.split(/[\\/]/).some((segment) => segment.startsWith('.'));
+}
 
 async function readJson(file: string): Promise<unknown | null> {
   try {
--- a/packages/metadata-fs/test/fs-behavior.test.ts
+++ b/packages/metadata-fs/test/fs-behavior.test.ts
@@ -15,6 +15,30 @@ const baseRef = (name: string): MetaRef => ({
 
 const sleep = (ms: number) => new Promise<void>((r) => setTimeout(r, ms));
 
+/**
+ * Deadline for a **positive** watcher assertion — the longest a case waits for
+ * an event it expects to arrive. Raced against the event promise, so a healthy
+ * run still finishes in roughly the poll interval; the number is only ever
+ * paid on the way to a failure.
+ *
+ * Sized for the merge queue, not for a quiet laptop. The queue runs the FULL
+ * suite (PR-side CI runs only the affected subset), and this repository's
+ * watcher rides `usePolling: 1000ms` plus an `awaitWriteFinish` stability
+ * window — wall-clock timers that stretch under a saturated runner while the
+ * assertions themselves are unaffected. The `chokidar: external file change`
+ * case below ejected PR #7208 from the queue on its old 3s deadline while
+ * being green on the identical SHA in PR CI; `watch-dot-root.test.ts` ejected
+ * it a second time on an 8s one. Both are widened to this value.
+ */
+const EVENT_WAIT_MS = 20_000;
+
+/**
+ * Per-case ceiling for the watcher cases. Must clear a full `EVENT_WAIT_MS`
+ * plus setup, or the case dies on the vitest timeout before reaching its own
+ * deadline — which reports as a timeout instead of as the missing event.
+ */
+const WATCHER_CASE_TIMEOUT_MS = 45_000;
+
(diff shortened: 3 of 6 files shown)


os-help:
**Merge-queue ejection triage** (lane PM, session `session_016R9de1FqP7NvwKvqXi92Gh`) — per the triage bot's checklist above, before any re-queue.

**Signature**: `test/fs-behavior.test.ts > FileSystemRepository — on-disk semantics > chokidar: external file change emits an update event` — a **pre-existing** test in this PR's own package.

**Verdict: load-timing flake, not a regression from this diff.** Three readings:

1. **Green control on the identical SHA**: `c47cc79e0` passed this very test in PR-side CI (Test Core shards all `success`, personally verified at accept time) — same code, same test, green there, red only in the queue's full-suite run.
2. **The matcher change cannot reach this test's semantics** (read-only verification): the failing test's root is `mkdtemp(os.tmpdir() + 'objectstack-fsbeh-')` — a plain, dot-free path. Old regex and new relative matcher agree byte-for-byte on every path this test produces; the behavioral delta of this PR exists only under dot-rooted watch roots.
3. **Mechanism fits load**: the test rides `usePolling: 1000ms` + `awaitWriteFinish` stability windows; the queue runs the full suite with heavier parallel load than PR CI. First occurrence of this signature in 24h per the bot's history line.

**Action**: re-queueing **once** (re-arm auto-merge). Standing rule honored: **same signature on a second ejection ⇒ escalate + file the flaky-test card, no further re-queues** — each blind re-queue rebuilds every PR behind this one.

One forward-looking note: this PR's own new tests add more watcher-timing surface (4s/8s waits) to the same suite; if queue-load flakes recur in this family, the fix direction is deadline-widening or event-promise waits in the *pre-existing* test, filed as its own card, not silent re-queues.

---

os-help:
**Patch round accepted — re-queueing (third and final attempt under the standing rule).** Session `session_016R9de1FqP7NvwKvqXi92Gh`.

`3a9f5b880`: test-only deadline hardening across the watcher-timing family — every positive event wait now rides a named `EVENT_WAIT_MS` (3–8s → 20s, case caps raised in step so a cap can't masquerade as the same flake from the other side), the 4s quiet window deliberately untouched and documented as never-to-shorten (it can only false-pass, not flake). Two additional same-shape test files hardened beyond the two ejection signatures (`no-root-on-attach.test.ts` carried the tightest cap in the package — the next candidate to eject), disclosed in the patch report. Zero production changes, 4 test files, +117/−11.

**Verification**: PR CI re-converged at `3a9f5b880` — all jobs `success` including Test Core (3/3), the shard both ejections happened on. Reverse-verified: `EVENT_WAIT_MS=1` turns exactly the four positive waits red (control-position failures, quiet window unaffected).

Honest caveat from the patch report, endorsed: local green + PR-CI green is what the previous SHA also had — queue load can only be tested in the queue. This lowers ejection probability rather than proving it away. **If this attempt ejects again on a watcher-timing signature, the standing rule fires: no fourth attempt — the flaky-family card gets filed and the PR escalates to the maintainer.**

---

Issue #6808, opened 2026-08-08
A deleted runtime `object` is still served by `SchemaRegistry.getObject` — the registry heal reaches the `metadata` map but never `objectContributors`


os-zhuang:
Recording only, per Prime Directive #10 — found while implementing #6621 (PR #6807), on a different function than the one that card touches, and pre-existing on both callers.

## The fact (measured on `origin/main` @ `4bb6f01af`)

`deleteMetaItem` ends its repository delete with `restoreArtifactRegistryView` — the #6687 three-tier heal. That walk operates entirely on `SchemaRegistry`'s **generic** `metadata` map: `removeRuntimeShadow`, `registerItem`, and `removeOverlayEntry` all address `this.metadata.get(type)`.

An `object` lives in **two** places. `applyRegistryWriteThrough` → `applyObjectRegistryMutation` writes both:

```ts
this.engine.registry.registerItem(request.type, request.item, 'name');   // metadata map
this.engine.registry.registerObject({ ...item, _provenance: 'org' }, pkg); // objectContributors
```

The heal only undoes the first. Measured with the real `SysMetadataRepository` over an in-memory engine, a runtime-created object saved and then deleted:

```
BEFORE delete: metadata['object'] keys -> ["myapp_invoice"] | objectContributors -> ["myapp_invoice"]
AFTER  delete: metadata['object'] keys -> []                | objectContributors -> ["myapp_invoice"]

registry.getObject('myapp_invoice')      -> STILL SERVED
registry.getItem('object','myapp_invoice') -> STILL SERVED   (getItem special-cases the
                                                              object type back to getObject)
```

`getObject` is the surface **data CRUD dispatches on**, so the half that survives is the load-bearing half: the `sys_metadata` row is gone and the object is still resolvable, syncable and writable for the life of the process. The half that is cleaned (`metadata['object']`) backs listing, which is the less consequential of the two.

## Why there is no one-line fix

`SchemaRegistry` has **no per-`(type, name)` object unregister at all** — the only removal verb is `unregisterObjectsByPackage(packageId, force)` (`packages/objectql/src/registry.ts:1339`), which is package-wide and throws when an owner still has extenders. So closing this is a registry **API** change (a name-addressed `unregisterObject` honouring the ADR-0029 single-owner/extender rules), not a protocol-side call.

## Reachability

Live on the ordinary Studio path: delete a runtime-created object, and the running process keeps dispatching it. Both delete-shaped callers are affected identically, because both run the same heal:

[...]

Issue #6505, opened 2026-08-08
IMetadataService.getObject has no declared relationship to get('object', name), and its two implementations disagree


os-project-manager:
Observation-class finding, filed per Prime Directive #10 while implementing #6055. Unassigned, `finding`, deliberately **not** queued — nothing a user hits today; it is a contract-documentation gap that made a consumer-side decision harder than it should have been.

## The fact (verified on `origin/main`)

`packages/spec/src/contracts/metadata-service.ts` documents the sibling convenience readers by their equivalence:

```ts
/**
 * Convenience: get a view definition by name
 * Equivalent to get('view', name)
 */
getView?(name: string): Promise< unknown | undefined >;
```

`getObject` — the only one of the family that is **required** rather than optional — carries no such sentence:

```ts
/**
 * Convenience: get an object definition by name
 * @param name - Object name (snake_case)
 * @returns The object definition, or undefined if not found
 */
getObject(name: string): Promise< unknown | undefined >;
```

And the two implementations do not agree:

- `packages/metadata/src/metadata-manager.ts` — `async getObject(name) { return this.get('object', name); }`
- `packages/objectql/src/metadata-facade.ts` — `async getObject(name) { return this.registry.getObject(name); }`, where the facade's own `get()` is `registry.getItem(type, name, pkg)` unwrapped to `item?.content ?? item`. Different lookup, different returned shape.

So a consumer holding an `IMetadataService` cannot know whether `getObject(n)` and `get('object', n)` answer the same thing, and in one of the two occupants of the `metadata` slot they do not.

## Why it cost something concretely

#6055 needed the ADR-0110 D3 verdict (`getDiagnosed`, #5840) for the `objectstack://objects/{objectName}` MCP resource, whose resolver is `getObject`. `getDiagnosed(type, name)` is `get`'s diagnosed twin, so the obvious fix — swap `getObject(name)` for `getDiagnosed('object', name)` — would have silently re-pointed the lookup on any `MetadataFacade`-backed host. Presuming an undeclared equivalence at a consumer is the private dialect Prime Directive #12 forbids, so PR for #6055 kept `getObject` as the resolver and asked `getDiagnosed` for the verdict only, paying one extra read on the miss path. That trade is written up in `diagnoseEmptyRead`'s TSDoc.

## Dispositions worth pricing (no recommendation forced)

[...]

PR #7211, opened 2026-08-10
fix(objectql): `MetadataFacade` object writes now reach the map its reads use (#6725)

os-zhuang:
Closes #6725.

## The defect

`MetadataFacade.register('object', …)` wrote through `SchemaRegistry.registerItem`, which stores into the generic `metadata` map. **Every** one of the facade's object reads resolves from `objectContributors`, which only `registerObject` populates:

| member | route | reads |
| --- | --- | --- |
| `getObject(name)` | `registry.getObject` | `objectContributors` |
| `get('object', name)`, `exists('object', name)` | `registry.getItem` → special-cases the object type back to `getObject` | `objectContributors` |
| `list('object')`, `listNames('object')`, `listObjects()` | `registry.listItems` → special-cases to `getAllObjects` | `objectContributors` |

So an object written through the public facade was readable back through **none** of them — `register` resolved successfully and every subsequent read answered `undefined` / `[]`.

`IMetadataService` (`@objectstack/spec/contracts`) declares `getObject(name)` ≡ `get('object', name)`, and its own conformance test round-trips a `register('object', …)` through both members. This was a shipped contract that could not work. Dormant in-tree only because nothing on `main` installs a `MetadataFacade` into the `metadata` slot — but the class is exported from this package's root and `core` entrypoints, so a downstream host that installs it got the split, including ObjectQL's own `bridgeObjectsToMetadataService`, whose "already registered?" probe would never answer and so would re-register the full object set on every boot.

## Shape chosen, and the blast radius measured for it

The card priced three dispositions and pre-ruled none. Measurement:

- **Who calls `facade.register('object', …)` today?** In-tree: only this package's own test file. `new MetadataFacade(...)` appears nowhere on `main` outside the two objectql test files. As published API: any host occupying the `metadata` slot — and the two in-tree occupants (`MetadataManager`, `createMemoryMetadata`) both round-trip correctly, so the facade is the odd one out, not the reference.
- **What does `registerObject` do that `registerItem` does not?** System-field injection, `better-auth` `apiMethods` reconciliation, ADR-0079 primary-title designation, `__search` companion provisioning, the ADR-0029 single-owner guard, contributor merge + priority sort, merge-cache invalidation, and the `_objectRevision` bump.
- **Is routing objects through `registerItem` itself the bug?** No — and this is the part that decides the shape. `registerItem`'s docblock does say "non-object metadata", but `SchemaRegistry.unregisterObject`'s header (#6808) states the actual invariant: *"a runtime-authored `object` is written into TWO places (`metadata['object']` via `registerItem` and `objectContributors` via `registerObject`)"*. The one in-tree precedent for this exact write, `MetadataProtocol.applyObjectRegistryMutation`, does both, with `packageId || 'sys_metadata'`. The facade was performing half of a documented two-place write.

So: **the write now performs both halves**, rather than moving to one of them.

Rejected, with reasons:

- **Reads also consult `registry.metadata.get('object')`** — would edit `registry.getObject`, which the data plane dispatches on (`assertObjectRegistered`), and would make the reads answer the stored document where #6505 / PR #6723 just documented them as answering the **runtime-effective** object. Wrong answer, much wider radius.
- **Refuse `'object'` at the facade** — would make `MetadataFacade` non-conformant with the interface it is documented to implement, and would turn a silent no-op into a throw inside both `bridgeObjectsToMetadataService` loops for any host that installs it.

## Details that are load-bearing

- **The contributor gets a copy.** `applyProtection` stamps `_packageId` / `_provenance` **in place**, and `applySystemFields` returns its input unchanged on the no-injection path (`systemFields: false`, `managedBy: 'better-auth'`, `sys_*`). A shared reference would therefore have written a synthetic package id onto the generic-map entry — exactly what the provenance pin forbids. Pinned by a new test that registers a `systemFields: false` object, i.e. the aliasing path.
- **Provenance.** A package-less object registers under the `'sys_metadata'` sentinel with `_provenance: 'org'` — both of which `getArtifactItem` / `isArtifactBacked` exclude, so it cannot read as code-shipped (the cloud#970 misclassification). Without the explicit `'org'`, `applyProtection` would default the copy to `_provenance: 'package'`. An object carrying a real `_packageId` registers under it and keeps `'package'`.
- **Ordering.** The contributor write runs first, because it is the half that can refuse (ADR-0029 single-owner). A refused registration now writes nothing at all instead of re-opening the split from the other side.
- **Both spellings.** `'object'` and `'objects'` are both special-cased on the read side, so both are covered on the write side.

## What happened to the provenance test

`"never invents a synthetic package id for object registrations"` is **unchanged and still in place**, still reading `(registry as any).metadata.get('object')` directly. That direct read remains the right instrument: the pin is about the **stored document**, and the object reads answer the contributor copy — which now exists and deliberately does carry the sentinel. Reading it through `get('object', …)` would have silently retargeted the assertion. A comment on the test now says so, and a new sibling test pins the aliasing hazard the copy exists to prevent.

## Scope note: `unregister`

`unregister('object', name)` now removes both halves too. This is not scope creep — without it the fix would have **introduced** #6808's shape from the other side: a removal that empties only the generic map leaves `getObject`, which the data plane dispatches on, serving a deleted object for the life of the process. `registry.unregisterObject(name)` (the #6808 verb) is idempotent and refuses, per ADR-0029, an object still extended by another package.

## Behaviour changes a caller can observe

- The six read members now answer a facade-registered object, with the runtime-effective shape the contract promises.
- `register('object', …)` can now **throw** where it previously succeeded and did nothing: claiming an object another package owns is refused (ADR-0029).
- `unregister('object', …)` can throw for an object still extended by another package.

## Tests

`packages/objectql/src/metadata-facade.test.ts` — 11 new cases, 14 total: round-trip through `getObject` **and** `get` (reference-identical, with an anti-vacuity `toBeDefined`); the enumeration members; the plural spelling; the runtime-effective shape on a multi-tenant registry; sentinel-not-artifact and real-package-id provenance; idempotent re-registration; the ADR-0029 refusal writing nothing; removal from both places; idempotent removal of an absent object.

**Reverse-verified** — pre-fix `metadata-facade.ts` restored under the new tests, red/green predicted per case before running. See the comment below.

Refs #6725, #6505 / PR #6723, #6853, #6808, ADR-0010, ADR-0029.

---
_Generated by [Claude Code](https://claude.ai/code/session_0141cZum72My2vskaQSoQ1tZ)_

### DIFF
--- a/.changeset/metadata-facade-object-write-read-split.md
+++ b/.changeset/metadata-facade-object-write-read-split.md
@@ -1,0 +1,67 @@
+---
+'@objectstack/objectql': patch
+---
+
+**`MetadataFacade.register('object', …)` now writes where its own object reads
+look — it was a silent no-op before (#6725).**
+
+`MetadataFacade` is exported from `@objectstack/objectql`'s root and `core`
+entrypoints for hosts that want to occupy the kernel's `metadata` slot with a
+`SchemaRegistry`-backed service. Its object write went through
+`SchemaRegistry.registerItem`, which stores into the generic `metadata` map —
+and **every one of its object reads resolves from `objectContributors`**, which
+only `registerObject` populates:
+
+- `getObject(name)` → `registry.getObject`;
+- `get('object', name)` and `exists('object', name)` → `registry.getItem`, which
+  special-cases the object type straight back to `registry.getObject`;
+- `list('object')` and `listNames('object')` → `registry.listItems`, which
+  special-cases to `registry.getAllObjects`.
+
+So an object registered through the facade was readable back through **none** of
+them: `register` resolved successfully and every subsequent read answered
+`undefined` / `[]`. `IMetadataService` (`@objectstack/spec/contracts`) declares
+`getObject(name)` ≡ `get('object', name)` and its own conformance test
+round-trips a `register('object', …)` through both members, so this was a
+shipped contract that could not work. Dormant in-tree only because nothing on
+`main` installs a `MetadataFacade` into the `metadata` slot — a downstream host
+that did (cloud, a third-party kernel) got the split, including the
+write-then-read in ObjectQL's own `bridgeObjectsToMetadataService`, whose
+"already registered?" probe would never answer and so re-registered the full
+object set on every boot.
+
+**What changed.** The facade's object write now performs *both* halves of the
+two-place write the registry documents for a runtime-authored object
+(`SchemaRegistry.unregisterObject`'s header states the invariant; the in-tree
+precedent is `MetadataProtocol.applyObjectRegistryMutation`, which does exactly
+this): `registerObject` for the contributor entry every read resolves, plus the
+existing `registerItem` for the stored document. Both spellings of the type
+(`'object'` and `'objects'`) are covered, because both are special-cased on the
+read side.
+
+Consequences a caller can observe:
+
+- `getObject` / `get` / `exists` / `list` / `listNames` / `listObjects` now
+  answer a facade-registered object. What they answer is the **runtime-effective**
+  object the contract promises: system columns injected, primary title
+  designated, `extend` contributions merged.
+- An object arriving with no `_packageId` is runtime-authored by definition, so
+  it is registered under the platform's `'sys_metadata'` sentinel and stamped
+  `_provenance: 'org'`. It therefore does **not** read as code-shipped —
+  `getArtifactItem` / `isArtifactBacked` exclude both — and does not become
+  un-editable. An object carrying a real `_packageId` is registered under it and
+  keeps `_provenance: 'package'`. The stored document keeps its as-authored,
+  unstamped shape; only the contributor copy carries registry coordinates.
+- `register('object', …)` can now **throw** where it previously succeeded and did
+  nothing: claiming an object another package already owns is refused by
+  ADR-0029. The contributor write runs first so a refusal writes nothing at all.
+- `unregister('object', name)` removes the object from both places. Without this
+  the fix would have re-opened #6808 from the other side — a removal that empties
+  only the generic map leaves `getObject`, which the data plane dispatches on,
+  serving a deleted object for the life of the process. It refuses, per ADR-0029,
+  an object still extended by another package.
+
+No in-tree caller changes behaviour: `new MetadataFacade(...)` appears nowhere on
+`main` outside this package's own tests, and the `metadata` slot is filled by
+`MetadataManager` or `createMemoryMetadata`, both of which already round-tripped
+correctly.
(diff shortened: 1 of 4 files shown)


os-zhuang:
## Reverse-verification

Pre-fix `packages/objectql/src/metadata-facade.ts` restored from `origin/main` under the post-fix tests. Red/green predicted per case **before** running: 4 green — the three original provenance cases (they exercise the non-object path, or the stored document, neither of which the pre-fix source handles differently) and "unregistering an object nothing registered stays a no-op" (pre-fix `unregisterItem` warns and returns). 10 red — every round-trip case, the new aliasing pin (which dies on `getObject(...)` being `undefined` before it can compare anything), and the ADR-0029 case (pre-fix `register` resolves instead of throwing).

Measured:

```
 ❯ src/metadata-facade.test.ts (14 tests | 10 failed)
     × keeps the stored document unstamped even when the contributor copy is stamped
     × reads a registered object back through BOTH getObject and get
     × reads it back through the enumeration members too
     × closes the same split for the plural `objects` spelling
     × serves the runtime-effective object, as the contract says it does
     × registers a package-less object under the sentinel, not as an artifact
     × registers a package-stamped object under its own package id
     × re-registering the same object replaces it rather than accumulating owners
     × refuses to claim an object another package owns, and writes nothing
     × unregisters an object out of BOTH places it was written into
      Tests  10 failed | 4 passed (14)
```

Same 10, same 4. **No deviations from the prediction.** Fix restored → 14/14 pass.

## Tests

- `packages/objectql` full suite: **167 files, 2917 tests, all passing.**
- `pnpm --filter @objectstack/objectql typecheck`: clean.

## Gates

Enumerated fresh from `origin/main`: **64** in `lint.yml`, **70** across all workflows. 48 of the 64 are root scripts; the other 16 run under `pnpm --filter <pkg>` (15 on `@objectstack/spec`, 1 on `@objectstack/lint`).

**All 48 root gates: PASS.** Three (`check:app-nav-i18n`, `check:i18n`, `check:i18n-coverage`) and `check:type-check-debt` first reported *"PREREQUISITE NOT MET — the workspace packages are not built"*, which measures nothing; re-run after `turbo run build --filter='./packages/*' --filter='./packages/*/*'` (the same build `lint.yml` does before those steps) they pass. Notably green: `check:meta-type-normalized`, `check:engine-double-contract`, `check:type-check-coverage`, `check:type-check-debt`, `check:empty-changeset`, `check:startup-registry-verdict`, `check:slot-lookup`, `check:service-providers`, `check:init-service-contract`, `check:tenant-chokepoint`.

This diff does **not** reach `packages/spec`, so the spec-gate suite and the #6017 cross-seat declaration are not prerequisites here. Four spec gates were run anyway to rule out collateral and baseline drift — `check:generated --reconcile-only`, `check:spec-changes`, `check:export-origins`, `check:api-surface` — all PASS, so no generated artifact is stale against this base (`origin/main` @ `55da611`).

One note on the enumeration: `check:strictness-ledger` and `check:variant-docs` appear in the `lint.yml` count only because they are named inside a **comment** there. Their real steps live in `spec-liveness-check.yml`, both `@objectstack/spec`-filtered. The count of 64 is right; two of its members are not `lint.yml` steps.


---

Issue #5149, opened 2026-08-04
Conditional-visibility predicates (`visibleWhen` / `visibleOn`) fail OPEN and silently — a broken predicate is indistinguishable from no predicate


yinlianghui:
## Summary

When a form field's `visibleWhen` / `visibleOn` predicate cannot be evaluated, the console renders the field **visible**, with no error, no `console.warn`, and no diagnostic anywhere. Since "visible" is also what you get with no predicate at all, **a predicate that never works looks exactly like a predicate that was never written** — including to the author who just wrote it and is looking at the screen.

The natural consequence is that conditional-visibility bugs are undetectable by inspection and survive indefinitely. We found two live instances in one app in one afternoon, one of them shipped and unnoticed for months.

Measured 2026-08-04 against `@objectstack/console` shipped with `17.0.0-rc.1`, zh-CN console, fresh database, object create dialog (`type: 'tabbed'` form view with authored `sections`).

## The failing-open path

In `framework-*.js`:

```js
function A(e, t, n, r, i) {                       // (predicate, values, DEFAULT, previous, extra)
  if (e == null || (typeof e === 'string' && !e.trim())) return n;
  try {
    let o = a.evaluate(Ie(e), { record: t, previous: r, ... });
    return o.ok ? o.value === true : n;            // not-ok  -> DEFAULT
  } catch { return n; }                            // throw   -> DEFAULT
}
```

and both call sites pass `true` as that default:

```js
// ui-components-*.js, per-field renderer
if (!C.visible || (m != null && !x(m, L, /* default */ true))) return null;
// framework-*.js
visible: e.visibleWhen == null || A(e.visibleWhen, t, /* default */ true, r, i)
```

So *every* evaluation failure — parse error, unbound identifier, type error — resolves to "show it".

## Repro 1 — an unbound identifier fails open (and the docs invite you to write one)

Values are bound under **`record`**. A predicate written with bare field names therefore references unbound identifiers and never works:

```ts
// src/views/<object>.view.ts
sections: [{ label: '审批', fields: [
  // deliberately IMPOSSIBLE — this can never be true
  { field: 'status', visibleOn: 'rating_number == "__never__"' },
]}]
```

**Expected:** field never rendered.
**Actual:** field rendered on every form, create and edit. Nothing logged.

Add the prefix and it works correctly:

```ts
{ field: 'status', visibleOn: 'record.rating_number != null && record.rating_number != ""' }
```

[...]

PR #7214, opened 2026-08-10
feat(lint): publish-time resolution of metadata-form predicate paths (#7010)

### LINKED ISSUE #7010: Publish-time validation of predicate path references — a spec-delivered predicate naming a nonexistent path should fail at authoring, not evaluate to a guess at render (#6936 companion)
Spun out of #6936's ruling (maintainer 2026-08-09, Option C: the objectui evaluator goes fail-open-with-warning). That ruling fixes the **renderer's** posture; this card is the producer-side fix the #6936 filing itself identified as the long-term answer: 「谓词表达式应在发布期被校验(引用的路径必须存在于对应 schema),而不是让渲染器在运行期猜」.

## Deliverable

An authoring-time check (landing point to be established by the implementing dev — the AUTHORING_RULES family in `metadata-protocol`, or `packages/lint`, following where the existing predicate-shape validations live): a metadata-delivered predicate (`visibleWhen` / `requiredWhen` / metadata-admin conditional fields — enumerate the actual carriers as step 1) whose identifier paths do not resolve against the target schema's scope (`data.*` after objectstack#6254's spelling correction) is refused or warned **at publish/validate time**, naming the unresolvable path.

## Sequencing / evidence base

- objectstack#6254 fixed 16 bare-spelling predicates in `object.form.ts` — the corpus that would have been caught by this check; use it as the fixture base.
- objectstack#6331 made metadata-admin actually read spec-delivered predicates, which is what made the evaluation semantics observable at all.
- The check must corpus-count before enforcing (standard widening discipline): run over the shipped spec metadata first, report the would-be-refused count in the PR body; a non-trivial hit count is a STOP.

## Not in scope

The objectui evaluator change (#6936, ruled C) — independent, already queued in its own card.

No `domain:*` label applied — routing is the triage seat's.

Refs: #6936 (ruling + measurements), objectstack#6254, objectstack#6331, #5149 (fail-open family).

os-project-manager:
Fixes #7010

The producer-side companion to #6936. That issue was ruled Option C — the objectui evaluator keeps failing OPEN with a warning — which settles the **renderer's** posture and deliberately leaves the producer side open. This is the check the #6936 filing itself named as the long-term answer: a predicate that references a path the target schema does not declare is refused at **authoring** time, not guessed at render time.

## 1. Predicate carriers, enumerated (step 1 of the deliverable)

Read off the schemas, not guessed. Every `*When`-family predicate slot declared with `ExpressionInputSchema`:

| carrier | key(s) | file:line | binding root |
|---|---|---|---|
| form field (view/metadata form) | `visibleWhen`, `visibleOn` (deprecated) | `packages/spec/src/ui/view.zod.ts:1659` / `:1661` | `record` (runtime) / `data` (metadata form) |
| form section | `visibleWhen`, `visibleOn` (deprecated) | `view.zod.ts:1760` / `:1762` | same |
| page component | `visibleWhen`, `visibility` (deprecated) | `packages/spec/src/ui/page.zod.ts:150` / `:152` | `record`, `current_user`, `page.< var >` |
| data field | `visibleWhen`, `readonlyWhen`, `requiredWhen` | `packages/spec/src/data/field.zod.ts:801-803` | `record` |
| field option (per-option) | `visibleWhen` | `field.zod.ts:162`, `packages/spec/src/ui/action.zod.ts:319` | `record` + `current_user` |
| object user action | `visibleWhen`, `disabledWhen` | `packages/spec/src/data/object.zod.ts:1079` / `:1082` | `record` |
| component prop spec | `visibleWhen` | `packages/spec/src/ui/component.zod.ts:349` | designer-side |
| flow builtin node config | `visibleWhen` | `packages/spec/src/automation/builtin-node-config.zod.ts:404` | designer-side |
| app area / action / bulk action / settings manifest | `visible` (not a `*When` spelling) | `app.zod.ts:319`, `action.zod.ts:379`, `bulk-action.zod.ts:209`, `settings-manifest.zod.ts:234`/`:491` | mixed |

**The metadata-admin conditional fields** the card names are the `*.form.ts` family: 17 `defineForm` layouts registered in `METADATA_FORM_REGISTRY` (`packages/spec/src/system/metadata-form-registry.ts`), of which 6 carry predicates — `object` (20 sites incl. helpText refs), `field` (10), `view` (7), `action` (6), `page` (4), `report` (3). Post-parse the shipped population is **46 predicates, all `visibleWhen`, all `data.`-rooted** (measured, not counted from source lines).

Note for a later reader: `FormFieldBaseSchema` carries **no** `requiredWhen` / `readonlyWhen` — those live on `FieldSchema` (the data field), which is a `record.*` surface. So on the form surface this card targets, `visibleWhen` is the only predicate key, and the card's "`requiredWhen`" belongs to the `record.*` layer that is out of scope below.

## 2. Landing point: `packages/lint`, and why the two candidates converge there

The card offered "the AUTHORING_RULES family in `metadata-protocol`, or `packages/lint`". They are the same place: `AUTHORING_RULES` **lives in** `packages/lint/src/authoring-rules.ts`; `metadata-protocol` is one *consumer* of it through `@objectstack/lint/runtime`.

Two hard constraints then decide it outright:

1. The check must parse CEL through the canonical front end (`parseCelToAst` — #4812 forbids a private parser). **`packages/spec` depends on `zod` and nothing else** (`packages/spec/package.json`), so a spec-side home would need a new `spec → formula` dependency, on top of Prime Directive #2.
2. The existing predicate-shape validations for this exact surface already live in `packages/lint/src/validate-visibility-predicates.ts`, and share the traversal (`view-walk.ts` `formViewSites`) this rule needs.

New file `packages/lint/src/validate-predicate-path-refs.ts`, registered in `AUTHORING_RULES` (`tier: 'gating'`, all three commands), exported from the `index` barrel.

## 3. What the check does

Two rules, one question — "does this identifier name something the target schema declares?" — differing only in whether the author supplied the root:

- **`predicate-path-unresolved`** — a `data.`-rooted path whose first unresolvable segment is not a key of the schema at that point (`data.tpye == 'formula'`). The message names the unresolvable path and the container; the hint carries a "Did you mean" from the declared key set.
- **`predicate-path-unrooted`** — a bare identifier that **is** a declared key of the scope (`type == 'formula'`). #6254's shape verbatim: right name, dropped root.

### Why the three existing gates all wave this through

`validate-visibility-predicates.ts` judges a predicate's **shape** — does it parse (`visibility-predicate-syntax`, #6253), is it rooted at all (`visibility-bare-identifier`, #6128), is that root right for the layer (`visibility-root-mislayered`). **None of them opens the target schema.** `data.tpye == 'formula'` passes all three and then fails open in the console, pixel-identical to no predicate at all (#5149).

And #6254 already measured the other half: its 16 bare `type == …` predicates were structurally uncatchable by #6248's gate, because `type` is an identifier **CEL itself declares**, so the strict checker reports a type overload rather than an unknown variable. This rule never asks CEL what resolves — it asks the schema, whose key set is closed and has no opinion about CEL's type vocabulary.

### Scope is `data.*` only, as a decision

An `error`-level gate needs a **closed** oracle. The metadata-form layer has one: the row under edit is an instance of a metadata type, enumerable key by key through `getMetadataTypeSchema`. The runtime `record.*` layer does not — lookup traversal, system columns the authored `fields` map never lists, formula/rollup outputs are all legitimate paths, and an `error` gate over an open set manufactures false build errors. Recorded as an open question rather than guessed at.

### Repeater rows rebind `data`, and the rule follows

Inside a `type: 'record'` / `repeater` / `composite` sub-field list, objectui's metadata SchemaForm evaluates `{ data: row }` — so `object.form.ts`'s `data.type` means `FieldSchema.type`, not the non-existent `ObjectSchema.type` (`view.zod.ts:1647-1657` states both halves). The rule descends with the same rebinding. Without it the shipped corpus would read 16 false positives instead of 0.

## 4. ⚠️ Corpus count — the load-bearing number

Run through the rule's **production** entry point over the shipped `METADATA_FORM_REGISTRY` (17 forms, 46 predicates):

| rule | would-be-refused, `origin/main@55da611e5` |
|---|---|
| `predicate-path-unresolved` | **0** |
| `predicate-path-unrooted` | **0** |

**No STOP.** Zero hits on the shipped tree, so the gate lands enforcing (`error`).

Two anti-vacuity guards ship with that number, because "0" is exactly what a gate that reads nothing also reports:

- every registry key resolves a schema (the oracle is not absent);
- corrupting every shipped predicate source to a path no schema declares — while keeping the real schemas, so descent still works — must produce exactly **46** findings. (The naive version of this guard, "resolve against an empty schema", is measurably **wrong**: an empty schema also fails to resolve the `fields` repeater, so the walk stops before the 16 sub-field predicates and the count comes out *below* the corpus while looking like proof of reach.)

### Reverse verification — direction predicted before running: RED on restore

#6254 rewrote 16 predicates in `object.form.ts` from `type …` to `data.type …`. Restoring the bare spelling on a deep copy of the shipped `object` form must turn the count from 0 to exactly 16 `predicate-path-unrooted`. It does — and the same measurement was taken independently against the real pre-#6254 file (`git show 643b7c76b^`) during development, with the same 16 sites at the same paths.

That measurement also caught a real defect in this PR before it shipped: with a `typeof s === 'object'` guard in `defOf`, the callable `lazySchema` proxy read as "not a schema", every key set came out empty, and the gate reported clean over the corpus **and** over the deliberately corrupted copy alike. The guard now accepts functions, and the comment on it says why.

## 5. Tests

`packages/lint/src/validate-predicate-path-refs.test.ts` — 24 cases: resolvable paths, repeater rebinding, record-map keys, opaque scopes, comprehension variables, both rule limbs (severity + rule id + the named path + the suggestion), the deliberate boundaries (unparseable source, object-bound form, unknown `schemaId`, unresolvable row schema, a throwing resolver), traversal reach (`form`, `formViews.<key>`, name-keyed `views` map, the `visibleOn` alias value), registry wiring, and the corpus block above.

## 6. Runtime-publish surface: deliberately not widened

`surfaces: ['cli']`, with a reason that is a **decision, not a limitation** (`RUNTIME_VISIBILITY_FAMILY_IS_CLI_ONLY`). This rule needs only the written item — its oracle is a static registry, not the tenant's other metadata — so none of the three existing P2 reasons is true for it, and `stackKeyForType('view')` already exists. It is not wired because a half-wired wall is worse than an unwired one: a Studio `view` write would be refused for an unresolvable path while a predicate that does not parse at all, and one with no root at all, walked through the same door. The `views[]` visibility family should move to `runtime-publish` together, as one measured edit.

## Verification

```
pnpm --filter @objectstack/lint test        # 69 files, 1795 passed | 4 skipped
pnpm --filter @objectstack/lint typecheck   # tsc --noEmit, clean
npx eslint packages/lint --no-inline-config # clean
node scripts/check-nul-bytes.mjs            # OK (6615 files, no raw control bytes)
```

Honest negative: `packages/cli` and `packages/metadata-protocol` were **not** run locally (their dependency chains are a full-repo build under a shared container). The registry change is additive and CLI-only, and `authoring-rule-wiring.test.ts` / `rule-id-barrel-exports.test.ts` / `authoring-rule-input-tier.test.ts` all pass in the lint suite; CI covers the rest.

## Changeset

`@objectstack/lint`: **minor**, real (not `skip-changeset`) — this adds a functional, `error`-level authoring check to a published package, which can newly fail an author's build. `zod` joins `@objectstack/lint`'s **devDependencies** so the unit tests can pin the schema traversal against a shape they fully control rather than against whatever `FieldSchema` declares this month.

## Not in scope

The objectui evaluator change (#6936, ruled C) — its own card, another repo. No `docs/adr/**`, no `content/docs/releases/**`, no gate or ratchet weakened.

---

### DIFF
--- a/.changeset/lint-predicate-path-refs-gate.md
+++ b/.changeset/lint-predicate-path-refs-gate.md
@@ -1,0 +1,63 @@
+---
+"@objectstack/lint": minor
+---
+
+feat(lint): metadata 表单谓词的路径解析闸门 —— 引用不存在的路径在发布期就被拒(#7010)
+
+新增 **error 级** 规则族 `validate-predicate-path-refs`,在 `os validate` / `os build` /
+`os lint` 三条命令上判定:一个 metadata 编辑表单(数据源为 `{ provider: 'schema', schemaId }`
+的 `defineForm` 形状)里的可见性谓词,其 `data.` 路径必须能在该 `schemaId` 对应的 schema 上逐段
+解析。两条规则,同一个问题:
+
+- `predicate-path-unresolved` —— `data.` 有根,但某一段不是该层 schema 声明的键
+  (`data.tpye == 'formula'`)。消息点名**不可解析的那一段**,hint 给出编辑距离最近的候选。
+- `predicate-path-unrooted` —— 裸标识符,而这个名字**恰好是目标 schema 的键**
+  (`type == 'formula'`)。这是 #6254 的形状:名字写对了,根丢了。
+
+## 为什么现有三道闸都放行
+
+`validate-visibility-predicates.ts`(ADR-0089 D3b)判的是谓词的**形状**:能不能解析
+(`visibility-predicate-syntax`,#6253)、有没有根(`visibility-bare-identifier`,#6128)、
+根对不对层(`visibility-root-mislayered`)。三条都**不打开目标 schema**,所以
+`data.tpye == 'formula'` 三条全过,而后在控制台 fail-open —— 元素无条件渲染,和完全不写谓词
+像素级一致(#5149 一族)。
+
+#6254 已经实测过这个洞的另一半:`object.form.ts` 的 16 处裸谓词写成 `type == 'formula'`,而
+#6248 的裸标识符闸**按构造抓不到** —— `type` 是 CEL 自己声明的类型名标识符,到严格检查器那里
+是类型 overload 错误而非未知变量。本规则不问 CEL「什么能解析」,只问**目标 schema**「这个键声明了
+没有」,所以 CEL 的类型名词汇表与它无关。
+
+## 落点:`data.*` 一层,这是决定而非省略
+
+`error` 级闸门要求 oracle 是**封闭**的 —— 一个能枚举、且「不在其中」确实等于「解析不到」的键集。
+ADR-0089 D3 的两层里只有一层满足:metadata 编辑表单的行是某个 metadata type 的实例,形状由
+`getMetadataTypeSchema` 这一份规范注册表逐键给出。运行期 record 面(`record.*`)**今天不封闭**
+—— lookup 穿透、authored `fields` 从不列出的系统列、formula/rollup 输出都是合法路径,在开集上架
+`error` 闸只会制造误红,而误红是闸门唯一不能犯的方向。已在规则注释与 #7010 上记为待裁的开放问题。
+
+## repeater 行重绑 `data`,规则跟着重绑(#6254)
+
+`type: 'record'` / `repeater` / `composite` 子字段列表里,`data` 绑定的是**这一行**
+(objectui 的 metadata SchemaForm 以 `{ data: row }` 求值),但根**仍然拼作 `data`**。所以
+`object.form.ts` 的 `data.type` 指的是 `FieldSchema.type` 而不是并不存在的 `ObjectSchema.type`。
+规则按同样的重绑下降 —— 不这么做,已发货的语料会读出 16 条误红而不是 0 条。
+
+## 语料计数(先量再收紧)
+
+规则通过**生产入口**跑过本仓发货的全部 metadata 表单(`METADATA_FORM_REGISTRY`,17 张表 46 条
+谓词):**两条规则的命中数都是 0**,因此才落 `error`。反向验证同时钉住:把 #6254 修前的裸写法还原
+到 `object.form.ts` 的深拷贝上,规则报出 **恰好 16 条** `predicate-path-unrooted` —— 正是该单
+当年人工读出来的那 16 处。
+
+## 明确不判(都是漏判方向,永远不会变成误红)
+
+规范前端解析不了的谓词(交还 #6253);不声明键集的作用域(`z.record(z.string(), z.unknown())`
+/ `z.unknown()`);record map 的**键**段(`z.record(z.string(), X)` 按构造接受任何键);推导宏
+绑定的循环变量(`data.tags.all(t, …)`);下标访问(`data.x['y']`);解析不到 schema 的
+`schemaId`。裸标识符里**不是** schema 键的那些也不判 —— 那是 `visibility-bare-identifier`
+的判决,一条坏谓词只应产生一条 finding。
+
+规则注册在 `AUTHORING_RULES`(`tier: 'gating'`,三条命令全覆盖),`surfaces` 保持 `cli`:它其实
+只需要被写入的那一条 item,但 `views[]` 可见性谓词这一族的另外三条规则今天都是 CLI-only,单独把
+三分之一的判决搬到 Studio 写入门上,比一条都不搬更难预测 —— 该族应当一次整体迁移,这是关于
+`views` 写入门的决定,不该搭在本单的车上。理由已写成 `RUNTIME_VISIBILITY_FAMILY_IS_CLI_ONLY`。
--- a/packages/lint/package.json
+++ b/packages/lint/package.json
@@ -36,7 +36,8 @@
   },
   "devDependencies": {
     "@types/node": "^26.1.2",
-    "vitest": "^4.1.10"
+    "vitest": "^4.1.10",
+    "zod": "^4.4.3"
   },
   "keywords": [
     "objectstack",
--- a/packages/lint/src/authoring-rules.ts
+++ b/packages/lint/src/authoring-rules.ts
@@ -120,6 +120,7 @@ import { validateFormLayout } from './validate-form-layout.js';
 import { validateSeedReplaySafety } from './validate-seed-replay-safety.js';
 import { validateSeedStateMachine } from './validate-seed-state-machine.js';
 import { validateVisibilityPredicates } from './validate-visibility-predicates.js';
+import { validatePredicatePathRefs } from './validate-predicate-path-refs.js';
 import { validateSecurityPosture } from './validate-security-posture.js';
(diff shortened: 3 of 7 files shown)


os-project-manager:
**ACCEPT** — devx PM seat. Both required gates green by name on the reworked head `b14ab11c0`; flipping ready and arming auto-merge.

What made this delivery acceptable, in the order the dispatch demanded it:

1. **The carriers were enumerated, not assumed** — read off the Zod schemas with file:line evidence, and the enumeration *corrected the card*: the form surface this check targets carries only `visibleWhen` (+ deprecated aliases); `requiredWhen`/`readonlyWhen` are record-layer keys and explicitly out of scope. An implementation that had "validated" the card's example keys would have been checking keys that don't exist on this surface.
2. **The landing point was argued from constraints, not preference**: the check must parse CEL through the canonical front end (#4812), and `packages/spec` depends on `zod` and nothing else — so `packages/lint` is the only home that doesn't invent a dependency or a third place. It sits beside the existing predicate-shape rules and shares their traversal.
3. **The corpus-count STOP was respected and came back clean**: all 46 shipped predicates walked through the rule's *production* entry point — 0 would-be-refused, so the gate lands at `error` legitimately. The anti-vacuity work is the strongest part: corrupting every predicate must yield exactly 46 findings, and the report documents why the naive version of that guard (resolve against an empty schema) is *measurably wrong* — an empty schema also kills the repeater descent, so the count reads below the corpus while looking like proof of reach. The reverse verification restored the real pre-#6254 file and got exactly the 16 historical findings, direction predicted first.
4. **A false-green was caught by the discipline, disclosed, and fixed**: the callable `lazySchema` proxy initially read as "not a schema", every key set came out empty, and the gate reported clean over good and corrupted corpora alike — precisely the failure the corrupted-corpus guard exists to catch. The fix is in, with the measurement written into the comment.
5. **Real changeset** (`@objectstack/lint` minor) — correct call, and the mirror image of today's `scripts/`-only cards: this adds a functional error-level check to a published package.
6. **Rework completed cleanly**: with #7186 merged, the local `collectionEntries` duplicate is gone (verified by git: no local definition, shared import in place, local `isRec` correctly retained since the shared module doesn't export it), corpus numbers unchanged, and cleanup ran **without `--force`** — the explicitly restated rule held, which is itself now part of #7055's evidence base.

Follow-ups already filed, not decided here: **#7219** (the `record.*` layer — close the oracle before gating an open set) and **#7220** (the four-rule visibility family moves to runtime-publish together or not at all — including the instructive implemented-then-reverted partial wiring).

Closes #7010 on merge.

---

Issue #7082, opened 2026-08-09
[finding] `permission-set-projection.ts:332` still states the ADR-0094 direction that D5-R retired — PR #6962 replaced the file-header copy and left the function-level one


os-zhuang:
Observation-class finding, recorded by the `domain:identity` PM seat during the two-reading landing confirmation of **PR #6962** (#6858, merged 2026-08-09T08:37:15Z). **Filed unassigned, no `domain:*` label** — routing is the triage seat's call. Prose only; no behaviour is wrong.

## What is stale

#6858 required that «the `permission-set-projection.ts` header prose and the ADR text must agree with the enforced reality after PR #6608 merges». PR #6962 delivered that for the **file header** — `:30` now reads "the 2026-07-14 direction confirmation **that used to sit here**…" — and recorded the retirement as **ADR-0094 D5-R**.

A **second copy of the same direction survives**, in the function-level JSDoc of `upsertEnvPermissionSet` at `packages/plugins/plugin-security/src/permission-set-projection.ts:332` (measured on `origin/main` after the merge):

```
 * an env-scope overlay is the platform's standard customization of a packaged
 * definition (ADR-0005; direction confirmed 2026-07-14, reversing the earlier
 * refuse-the-env-door rule): the facets update to the EFFECTIVE (overlay-wins)
 * body while the `managed_by:'package'` + `package_id` provenance is
 * PRESERVED … and deleting the overlay resets the row to the shipped
 * declaration (the layered read reveals the baseline again).
```

Two claims in that block are now false, and they are precisely the two D5-R retired:

1. **"an env-scope overlay is the platform's standard customization of a packaged definition (direction confirmed 2026-07-14)"** — that direction is exactly what D5-R retires. Since #6483 / PR #6608 rolled `permission` back to `allowOrgOverride: false`, the write-through for a code-declared set **403s**.
2. **"deleting the overlay resets the row to the shipped declaration"** — #6960 measures the opposite: a pre-rollback env overlay on an artifact-backed item of a rolled-back type can no longer be removed through the ordinary delete path (403 `NOT_OVERRIDABLE` on every kernel), leaving only `OS_METADATA_WRITABLE`. PR #6962's own D5-R section deliberately declined to restate "delete = reset" for this reason — the header was corrected and this copy was not.

## Why this survived a careful PR

[...]

Issue #6947, opened 2026-08-09
[finding] `filter.zod.ts` still says `$icontains` is "NOT yet answered by any backend" and names #5702 as the open gap — #5702 landed, and two drivers now execute it


os-zhuang:
Observation-class finding, recorded by the `domain:identity` PM seat while re-checking the blocker chain of #5814. **Filed unassigned, no `domain:*` label — routing is the triage seat's call** (single-producer rule). Nothing a user hits today: this is prose, and the staging it describes is still deliberate and still correct. Only the *status* sentence has expired.

## What is stale

`packages/spec/src/data/filter.zod.ts:297-306` (measured on `origin/main` @ `17688fed`):

```
 * ### Implementation status — declared here, NOT yet answered by any backend
 *
 * This PR is the contract half of the #4706 ruling and deliberately ships no
 * runtime behaviour (#5701). No driver evaluates `$icontains` today; all five
 * refuse it, loudly, as an operator they do not implement — which is the
 * fail-closed direction and stays true until #5702 lands the lowerings.
```

**#5702 has landed** — *"drivers: `$regex` 响亮拒收 + `$icontains` 各后端实现(#4706 裁决 B 案 · 驱动半边)"*, closed **completed 2026-08-08T06:43:29Z**. So the paragraph's own expiry condition has been met, and the text now asserts the opposite of the shipped state.

## Evidence (read on `origin/main`, per file)

| Face | Reading |
|:---|:---|
| `packages/drivers/driver-sql/src/sql-driver.ts:7860` | `case '$icontains':` — a real lowering, with a dedicated comparand validator at `:857` and dialect notes at `:1258-1278` |
| `packages/drivers/driver-turso/src/remote-transport.ts:70`, `:112` | `'$icontains'` present in **both** supported-operator tables |
| `packages/drivers/driver-mongodb/src/mongodb-filter.ts:607` | comment mention only — no case arm found |
| `packages/objectql/src/having-filter.ts:57` | explicitly states `$icontains` is **not** added on this face ("would need its own …") |

So "no driver evaluates it / all five refuse it" is false for at least the SQL and turso faces. **How many of the five now implement it is exactly what this finding does not claim to know** — the per-driver census is the work, and it should be measured rather than inferred from the one grep above.

## What is NOT wrong

[...]

PR #7259, opened 2026-08-10
refactor(core,runtime,rest): one ExecutionContext assembler, two named anonymous entries (#6216)

os-zhuang:
Fixes #6216

Implements the maintainer ruling of 2026-08-08 (comment 5227240904), **Option A — one shared assembler, explicit dual mode, fail-closed default**.

## What changed

`resolveAuthzContext` already made AUTHORIZATION resolution single-sourced. The step *after* it — turning the resolved envelope into the `ExecutionContext` that reaches enforcement — was still one hand-written copy per transport, and that duplication produced a measured defect family: **#6071** (the REST copy never set `principalKind`, so every enforcement judgment reading it was silently never-true on that face) and **#6206 / #6551** (a dropped `accessible_org_ids`, and the `group` posture's Layer 0 wall reads it directly — real 403s).

New module `packages/core/src/security/assemble-execution-context.ts`, with the anonymous divergence as **named API instead of drift**:

- `assembleExecutionContext(input)` — the **fail-closed default** entry. No resolved principal, no context; the surface answers 401. REST's contract, unchanged.
- `assembleExecutionContextOrGuest(input)` — the **explicit guest** entry. No resolved principal produces today's dispatcher guest envelope (`principalKind: 'guest'`, `positions: ['guest']`), whose consumers are live (`plugin-security/explain-engine.ts` reads guest for its `EXTERNAL` posture floor). Adopted only by a surface whose product semantics serve anonymous principals.
- Field set **closed by type**. `EntryExecutionContextField` is `Exclude< keyof ExecutionContext, NonEntryExecutionContextField >`, and `ExecutionContextEntryFields` makes every one of those keys **required** (the value may be `undefined`; the decision may not be omitted).

`packages/runtime/src/security/resolve-execution-context.ts` (dispatcher / MCP door) now takes the guest entry; `packages/rest/src/rest-server.ts` `computeExecCtx` takes the fail-closed default.

The share-link surfaces already consume whole envelopes (#6552 / #6647) and were deliberately **not** re-plumbed.

## The remaining divergences are values, not switches

Converging turned up a **second** real divergence beyond the anonymous face, and it is load-bearing:

**`accessToken`.** The dispatcher has always carried `authz.accessToken`; REST's literal never had the key. It is not inert — `packages/objectql/src/engine.ts` `buildSession` feeds it to `HookContext.session.accessToken`, a published hook surface (`packages/spec/src/data/hook.zod.ts`). A naive convergence would have started exposing the better-auth session token to hooks on every REST-driven write. So `accessToken` is a **required input** of the assembler: the dispatcher passes `authz.accessToken`, REST passes `undefined` with the reason written at the call site. Same for `oauth` (the `/mcp` door alone). Required inputs, not booleans — a face cannot silently omit one, and what it withholds it withholds on the record.

## Evidence 1 — no behaviour change, measured both ways

A green suite proves nothing here; the suite was green before. Two independent measurements:

**(a) Frozen-transcription parity matrix** (`packages/core/src/security/assemble-execution-context.test.ts`). The pre-#6216 assembly of *each* face is transcribed verbatim from `origin/main@0caf122f0`, frozen, and every shape either face serves is assembled both ways and compared: 3 authz shapes (anonymous / minimal human / full human with tenant+email+token+tabs+posture) times 6 OAuth shapes (none, agent `data:read`, agent `data:write`, agent with `actions:execute`, bearer without `azp`, bearer for a different user) times 4 localization shapes times 2 request-locale shapes, plus the REST matrix. **179 tests, all green.** The transcriptions are a pin, not a live second implementation — nothing imports them, and the day they need editing is the day a face's output changed.

**(b) On the wire, against the real pipeline.** Both face files were restored from `origin/main` with the new tests kept, and the tests re-run through the real `computeExecCtx` and the real `resolveExecutionContext`. Result below.

## Evidence 2 — the type closure actually bites (shown, not asserted)

**Probe A — add a field to `ExecutionContext`.** Added `probeField6216: z.string().optional()` to `ExecutionContextSchema`, rebuilt `@objectstack/spec`, ran `tsc` on `@objectstack/core`. Core went from 98 pre-existing errors (its ledgered DEBT number, unchanged by this PR) to 100, and **both new errors are the closure firing, independently**:

```
src/security/assemble-execution-context.ts(149,7): error TS2322: Type 'true' is not assignable to type 'never'.
src/security/assemble-execution-context.ts(252,3): error TS2741: Property 'probeField6216' is missing in type '{ positions: string[]; ... }' but required in type 'ExecutionContextEntryFields'.
```

Line 252 is the assembly literal; line 149 is the order-list exhaustiveness constant. `pnpm --filter @objectstack/core build` fails on the same two errors (`Error: error occurred in dts build`), so the bite is visible to CI's build job, not only to a local `tsc`. The runtime pin went red too, naming the field:

```
AssertionError: expected [ 'accessToken', ...(29) ] to deeply equal [ 'accessToken', ...(30) ]
-   "probeField6216",
```

**Probe B — add a required per-face input.** Added a required `probeInput6216: string` to `ExecutionContextAssemblyInput` and rebuilt core. Both faces went red:

```
packages/rest    src/rest-server.ts(2605,51): error TS2345: Argument of type '{ authz: ...; oauth: undefined; ... }' is not assignable to parameter of type 'ExecutionContextAssemblyInput'.
packages/runtime src/security/resolve-execution-context.ts(193,42): error TS2345: Argument of type '{ authz: ...; oauth: ... }' is not assignable to parameter of type 'ExecutionContextAssemblyInput'.
```

**Honest limit of the guarantee.** The ruling says a new `ExecutionContext` field "must land in every face at compile time". What Probe A measures is that it lands at the **one place where every face's value is decided** — and Probe A's runtime and rest typechecks stayed **green**, which was predicted in writing before the run. That is the design working, not a gap: with one assembler the faces no longer assemble, so there is no per-face site left to forget. A new field propagates to the faces (Probe B) exactly when it needs a per-face value. Stated plainly rather than rounded up to the ruling's phrasing.

## The one measured residual

The shared assembler omits `undefined`-valued keys; both legacy literals sometimes **spelled** them. Restoring the legacy REST face turned exactly one new assertion red, and the whole diff was one key:

```
+   "tenantId",
```

`tenantId` was present-with-`undefined` for a tenant-less principal. This is invisible to `ctx.x` reads, to `JSON.stringify` and to spreading the envelope; it is visible to `Object.keys` and `in`. No value moved on either face, for any shape. It is pinned rather than waved off — see the "measured residual" block in the core test.

Predicted in writing (scratchpad, before the run): restore both face files from `origin/main`, keep the new tests. 1 red (the `tenantId` presence residual on REST), 3 green — because the other new pins assert *unchanged* behaviour and legacy code has that behaviour by construction.

| pin | predicted | measured |
| --- | --- | --- |
| runtime: sessionless request yields the whole guest envelope, key set included | green | **green** (31/31 in the file) |
| runtime: session bearer carried as `accessToken` | green | **green** |
| rest: exact key set for a session-backed request | **red**, `tenantId` only | **red**, `tenantId` only |
| rest: `accessToken` still withheld | green | **green** |

Measured equals predicted, including the three deliberately-green ones. A guard that stays green under its author's own mutation is the result here, not a gap: these pins exist to catch a *future* change of behaviour, and their green under the legacy code is the proof that this PR did not change it.

One thing I did **not** predict: the first run of the REST key-set pin failed against my *own* code, because I guessed the golden set wrong (`currency` instead of `email` — the test's ql stub returns a `sys_user` row with an email, and no currency is configured). The pin caught its author. Corrected before anything was pushed.

## Gates

| gate | result |
| --- | --- |
| `pnpm --filter @objectstack/core test` | 703 passed / 29 files |
| `pnpm --filter @objectstack/rest test` | 1221 passed / 76 files |
| `pnpm --filter @objectstack/runtime test` | 1839 passed / 118 files |
| `pnpm --filter @objectstack/runtime --filter @objectstack/rest typecheck` | Done (both) |
| `eslint` on all 7 touched files | clean |
| `check:authz-resolver` | PASS |
| `check:org-identifier` | PASS |
| `check:tenant-chokepoint` | PASS |
| `check:required-contexts` | PASS |
| `check:role-word` / `check:empty-changeset` / `check:spec-parsed-alias` / `check:slot-lookup` | PASS |
| `check:type-check-coverage` | OK — core's DEBT stays at 98, the new files add zero |
| `check:nul-bytes` + targeted control-byte self-scan | clean |

Changeset: `.changeset/execution-context-single-assembler.md`.

## Deliberately not done

- **Share-link surfaces** — already whole-envelope passthrough (#6552 / #6647); not re-plumbed, per the ruling.
- **The stdio MCP face** (`packages/mcp/src/plugin.ts`) is a **fourth** hand-written assembly that drops `tabPermissions` and `accessToken`. Converging it would change its output, on a surface the ruling did not weigh. Filed as #7279.
- **`context.authGate`** is read by REST's `enforceAuth` but never declared on `ExecutionContextSchema`, so it sits one `as any` outside the closure this PR builds. Filed as #7280.

### DIFF
--- a/.changeset/execution-context-single-assembler.md
+++ b/.changeset/execution-context-single-assembler.md
@@ -1,0 +1,47 @@
+---
+"@objectstack/core": minor
+"@objectstack/runtime": patch
+"@objectstack/rest": patch
+---
+
+refactor: one shared `ExecutionContext` assembler, two named anonymous entries (#6216)
+
+`resolveAuthzContext` already made AUTHORIZATION resolution single-sourced; the
+step after it — turning the resolved envelope into the `ExecutionContext` that
+reaches enforcement — was still one hand-written copy per transport, and the
+copies drifted twice for real: **#6071** (the REST copy never set
+`principalKind`, so every enforcement judgment reading it was silently
+never-true on that face) and **#6206 / #6551** (a dropped `accessible_org_ids`
+produced real 403s on the share-link faces).
+
+**@objectstack/core** gains the single assembly, with the anonymous divergence
+as named API rather than drift (maintainer ruling 2026-08-08 on #6216, Option
+A):
+
+- `assembleExecutionContext(input)` — the **fail-closed default** entry. No
+  resolved principal → `undefined`, and the surface answers 401.
+- `assembleExecutionContextOrGuest(input)` — the **explicit guest** entry. No
+  resolved principal → a first-class guest envelope (`principalKind: 'guest'`,
+  `positions: ['guest']`), whose consumers are live (`explain-engine`'s
+  guest ⇒ `EXTERNAL` posture floor). Adopted only by a surface whose product
+  semantics serve anonymous principals.
+- The field set is **closed by type**: `ExecutionContextEntryFields` requires a
+  decision for every `ExecutionContext` field that is not explicitly declared
+  non-entry-resolved, so a new field cannot reach one transport and miss
+  another. Also exported: `ENTRY_EXECUTION_CONTEXT_FIELDS`,
+  `EntryExecutionContextField`, `ExecutionContextAssemblyInput`,
+  `OAuthTokenProvenance`, `EntryLocalization`.
+
+**@objectstack/runtime** (`resolveExecutionContext`, the runtime / MCP
+dispatcher) and **@objectstack/rest** (`computeExecCtx`) now assemble through
+that module — the dispatcher via the guest entry, REST via the fail-closed
+default.
+
+**No runtime behaviour change on either surface.** The remaining per-face
+divergences are required inputs rather than silent omissions: REST passes
+`accessToken: undefined` (it has never carried the session bearer on the
+envelope, and `session.accessToken` is a published hook surface) and
+`oauth: undefined` (OAuth bearers are honoured on the `/mcp` door alone). The
+one measurable difference is that a key whose value was `undefined` is now
+omitted rather than spelled — invisible to `ctx.x` reads, to `JSON.stringify`
+and to spreading the envelope.
--- a/packages/core/src/security/assemble-execution-context.test.ts
+++ b/packages/core/src/security/assemble-execution-context.test.ts
@@ -1,0 +1,429 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+//
+// #6216 — the ExecutionContext assembly converges onto ONE module, and the
+// maintainer ruling of 2026-08-08 (Option A) is explicit that **neither surface
+// changes runtime behaviour**: what changes is that the anonymous divergence
+// becomes named API instead of drift.
+//
+// A green suite proves nothing about that on its own — the suite was green
+// before the change too. So the load-bearing test here is a PARITY PIN: the
+// pre-#6216 assembly of each face is transcribed VERBATIM below, frozen, and
+// every shape either face can serve is assembled both ways and compared.
+//
+// ⚠ The two `legacy*` functions are FROZEN TRANSCRIPTIONS of code that no
+// longer exists (runtime `resolve-execution-context.ts` and REST
+// `rest-server.ts` `computeExecCtx`, at `origin/main@0caf122f0`). They are a
+// pin, NOT a second live implementation: nothing imports them, and they must
+// never be "kept up to date" with the shared assembler — the day they need
+// editing is the day a face's output changed and the change must be argued on
+// its own merits, not absorbed here.
+
+import { describe, it, expect } from 'vitest';
+import { ExecutionContextSchema } from '@objectstack/spec/kernel';
+import { scopesToAgentPermissionSets, MCP_OAUTH_SCOPE_ACTIONS } from '@objectstack/spec/ai';
+
+import {
+  assembleExecutionContext,
+  assembleExecutionContextOrGuest,
+  ENTRY_EXECUTION_CONTEXT_FIELDS,
+  type EntryLocalization,
+  type OAuthTokenProvenance,
+} from './assemble-execution-context.js';
+import type { ResolvedAuthzContext } from './resolve-authz-context.js';
+
+// ───────────────────────── frozen pre-#6216 transcriptions ─────────────────────────
+
+/** Runtime / MCP dispatcher assembly, verbatim, pre-#6216. FROZEN — see header. */
+function legacyDispatcherAssembly(
+  authz: ResolvedAuthzContext,
+  oauthPrincipal: OAuthTokenProvenance | undefined,
+  localization: EntryLocalization | undefined,
+  requestLocale: string | undefined,
+): any {
+  const ctx: any = {
+    positions: authz.positions,
+    permissions: authz.permissions,
+    systemPermissions: authz.systemPermissions,
+    isSystem: false,
+  };
+  if (authz.userId) {
+    if (oauthPrincipal?.clientId) {
+      ctx.principalKind = 'agent';
+      ctx.onBehalfOf = { userId: authz.userId, principalKind: 'human' };
+      ctx.permissions = scopesToAgentPermissionSets(oauthPrincipal.scopes);
+      ctx.positions = [];
+      ctx.systemPermissions = oauthPrincipal.scopes?.includes(MCP_OAUTH_SCOPE_ACTIONS)
+        ? (authz.systemPermissions ?? [])
+        : [];
+    } else {
+      ctx.principalKind = 'human';
+    }
+  } else {
+    ctx.principalKind = 'guest';
+    ctx.positions = ['guest'];
+  }
+  if (authz.userId) ctx.userId = authz.userId;
+  if (authz.tenantId) ctx.tenantId = authz.tenantId;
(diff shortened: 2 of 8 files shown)


os-zhuang:
Note for the reviewer, and a tooling observation worth recording: the PR body and both filed findings (#7279, #7280) were each written ending with the required attribution block

```
---
```

and in all three cases the stored body comes back **without** it. Read-back confirms the text is absent at rest, not merely hidden — while the same block survives intact in issue *comments* (every comment on #6216 still carries it, including this one). So the strip happens on the body write path, not on read and not on comments. Recording it here so the convention is satisfied where it can be, and so the next agent does not spend the same time diagnosing it.

---

os-zhuang:

CI on head `b74c8363d`: **25 runs, none non-green.** Boundaries re-measured against the merge base: 8 files; **share-link surfaces untouched** (0 hits — they are whole-envelope passthrough per #6552 / #6647), **no `content/docs/releases/`**, **no new `organizationId` spellings**.

## The residual was disclosed — and this seat closed the question the disclosure left open

The report states the one measured delta honestly: the shared assembler **omits** undefined-valued keys where the legacy literals sometimes **spelled** them, so the entire on-the-wire diff is a single key (`tenantId`, present-with-undefined for a tenant-less REST principal), invisible to `ctx.x`, to `JSON.stringify` and to spreading — **visible only to `Object.keys` / `in`**. Insertion order also changes for REST.

That caveat is not theoretical in this repo, so I checked rather than accepting "semantically inert" on its face: **`packages/runtime/src/action-session-shape-contract.test.ts:89` asserts `expect('tenantId' in built!).toBe(false)`** — a key-*presence* contract someone deliberately wrote. It runs in Test Core, which is green on this head, and the new direction (omit) agrees with it. So the class of consumer that could notice exists, and the one instance of it in the repo is satisfied. Landing on that basis.

## Type closure: shown, with an honest limit that this seat accepts rather than waves through

Probe A (add an optional field to `ExecutionContextSchema`) turned core's tsc red at **two independent sites** — the assembly literal (`TS2741`) and the order-list exhaustiveness constant (`TS2322 … not assignable to type 'never'`) — and `pnpm --filter @objectstack/core build` fails on the same two, so the bite is visible to CI's Build job, not just a local tsc. Probe B (add a required *input*) turned **both faces** red.

But under Probe A the runtime and rest typechecks stayed **GREEN**, which the author predicted in writing beforehand and **refused to round up to the ruling's wording** ("lands in every face at compile time"). That refusal is the right call and I am recording my judgement explicitly rather than letting it slide: the ruling's *purpose* was that the #6071 drift class become unrepresentable, and with a single assembler **there is no per-face assembly left to forget** — which achieves the purpose more directly than per-face compile errors would, and Probe B shows propagation to the faces exactly when a new field needs a per-face value. Not a `needs_decision`; a design that delivers the intent while the literal phrasing over-promised.

## Converging surfaced a second real divergence, and it was handled by value rather than by switch

`accessToken` — carried by the dispatcher since always, never by REST, and **not inert** (it flows to the published `HookContext.session.accessToken` via objectql's `buildSession`). Made a **required input** each face supplies by value (dispatcher `authz.accessToken`; REST `undefined`, with the reason at the call site) instead of a boolean mode flag. That keeps the divergence named and typed rather than hidden behind a parameter, which is the same principle the ruling applied to the anonymous face.

## Two things that raise this above a passing review

- **A CI-only breakage it found and fixed in-task, and reported as its own defect.** The assembler's *value* import of `@objectstack/spec/ai` broke five packages whose vitest config aliases `@objectstack/core` to source but not that subpath (`ENOTDIR … /spec/src/index.ts/ai`, 17 suites). Rather than push MCP vocabulary into the microkernel or edit five unrelated test configs, the OAuth scope vocabulary now stays at the door that speaks it — `OAuthTokenProvenance` carries pre-derived `scopePermissions` / `delegatesActions` and the assembler still decides, once for every face, what that ceiling replaces. Verified by re-running all five green. Reported as "my defect, found by CI, fixed in-task" rather than quietly absorbed.
- **Its own golden pin caught its own wrong guess** (`currency` where the harness yields `email`), corrected before any push. A pin that catches its author on first run is a pin doing its job.

Also correct: **three of the four reverse-verification pins are deliberately GREEN** under the author's own mutation, stated as a result rather than dressed as reds — they exist to catch a *future* behaviour change, and their green against restored legacy code is the proof this PR changed nothing.

## Deliberately deferred, and rightly

The **stdio MCP face** (`packages/mcp/src/plugin.ts`) is a *fourth* hand-written assembly that drops `tabPermissions` and `accessToken`. Converging it would change output on a surface the ruling never weighed — filed as **#7279** instead of fixed here. And **#7280**: `context.authGate` is written and read by `rest-server.ts` but is not declared on `ExecutionContextSchema`, so it sits one `as any` *outside* the closure this PR builds. Both are the right shape of follow-up: found while converging, not smuggled into the convergence.

**Tooling, confirmed independently by this seat**: the trailing `---` + attribution block is **stripped from issue and PR bodies on the write path** while surviving in comments — verified by read-back on #6947, #7082 and #7281, all of which were authored with it. So its absence in bodies is a tool behaviour, not an authoring miss by anyone, and the convention is met where it can be.

---

Issue #4782, opened 2026-08-03
metadata/loaders/database-loader.test.ts 的 mock driver `supports` 写满 off-spec 键(aggregations/nestedObjects/geoQueries/changeStreams…)—— 从未与 DriverCapabilities 对上


os-zhuang:
#4634(DriverCapabilities 活性审计)期间发现,顺带记录,不在该 PR 范围内修(metadata 包不在其区域约束内,且不影响任何门禁/测试结果)。

## 现象

`packages/metadata/src/loaders/database-loader.test.ts:36-47` 的 mock driver:

```ts
supports: {
  transactions: false,
  joins: false,
  aggregations: false,      // ← 不是 DriverCapabilities 的键(那边叫 queryAggregations,现已退役)
  streaming: false,
  bulkOperations: true,     // ← 不存在的键
  nestedObjects: false,     // ← 不存在的键
  fullTextSearch: false,
  geoQueries: false,        // ← 不存在的键(那边曾叫 geospatialQuery)
  changeStreams: false,     // ← 不存在的键
},
```

一半键从未在 `DriverCapabilitiesSchema` 里存在过,另一半(transactions/joins/streaming/fullTextSearch)已于 #4634 退役。对象未做类型标注所以一直编译通过,也没有任何读者,纯属误导下一个照抄 mock 的作者。cloud 侧 `ControlPlaneProxyDriver` 的 fallback 字面量(`supports: any`)有同款问题,已另行记入 cloud 仓 issue。

## 建议

#4634 落地后改为 `supports: {}`(空广告即合法广告),删掉整块假键。一行修复。

发现于 #4634 审计,未认领。

os-zhuang:
存量裁决轮（维护者 2026-08-06 委托，session_01LeEfA7CFwbJb7JJmXm2KM3）：裁定**入队**。前提经 origin/main 核实仍成立：database-loader.test.ts 的 mock driver `supports` 块仍写满 off-spec 键（transactions/joins/aggregations/…），与现行 DriverCapabilitiesSchema 完全对不上——误导下一个照抄者，小改动即可收口。维护者可否决。

---
_Generated by [Claude Code](https://claude.ai/code)_

hotlong:
[...]

PR #7261, opened 2026-08-10
test(objectql): type the batch-atomic driver double as `IDataDriver`, dropping the retired `supports.transactions` bit

### LINKED ISSUE #6546: objectql/protocol-batch-atomic.test.ts mock driver advertises the retired `supports.transactions` bit — invisible because the mock is `: any`
Found while fixing #4782 (PR #6545). Recorded, not fixed there: #4782's file surface is
`packages/metadata/src/loaders/database-loader.test.ts` only.

## Observation

`packages/objectql/src/protocol-batch-atomic.test.ts:46`:

```ts
const driver: any = {
    name: 'snapshot',
    version: '0.0.0',
    supports: { transactions: true },
    ...
```

`transactions` was RETIRED by #4634 and is tombstoned in `DriverCapabilitiesSchema` as
`retiredKey(...)`, i.e. `never`. Authoring it is meant to fail `tsc` at the authoring site
and to be rejected by a parse with the removal prescription.

Neither channel fires here, and the reason is the annotation: the literal is `: any`, so
it is never compared against `IDataDriver`, and nothing in this test parses the driver
through `DriverInterfaceSchema`. The bit is inert — transaction use gates on METHOD
presence (`driver.beginTransaction`), which this mock does implement, so the test passes
for the right reason and would pass identically with the bit deleted.

This is the same family as #4782 but a different failure mode, which is why it is worth a
separate line: #4782's block was at least partly VISIBLE to `tsc` (its four retired keys
were four real TS2322 in the `@objectstack/metadata` DEBT ledger). Here the `: any` erases
the diagnostic entirely.

## Why this is observation-class, not a defect

Nothing a user hits, no gate is red, and no behaviour depends on it. The concrete harm is
the one #4782 names: it is a mock, and mocks get copied. A future author who copies this
one inherits a retired bit plus the `: any` that hides it.

Deliberately NOT flagged: `packages/spec/src/data/driver.test.ts:317` spells
`supports: { transactions: true, streaming: true }` on purpose — it is the rejection
fixture (`it('REJECTS a driver whose supports still authors a retired bit')`) and asserts
the tombstone prescription. That one is correct as written.

## Suggestion

Drop the bit (`supports: {}`), and consider whether `const driver: any` should be
`const driver: IDataDriver` — the `any` is what turned a compile error into silence, so
the annotation is the part that actually prevents recurrence. Whether that retyping is
cheap depends on how the rest of the literal lines up with `IDataDriver`; it was not
measured here.

A repo-wide grep for other `supports:` literals spelling retired or invented capability
keys found no further instances in `packages/` or `examples/` beyond this one, #4782's,
and the spec rejection fixture above.

os-zhuang:
Closes #6546. Test-only.

## What was wrong

The snapshot driver in `packages/objectql/src/protocol-batch-atomic.test.ts` authored
`supports: { transactions: true }` — a capability key **retired by #4634** and tombstoned in
`DriverCapabilitiesSchema` as `retiredKey(...)`.

Both enforcement channels were off:

- the literal was annotated `const driver: any`, so it was never compared against `IDataDriver`;
- nothing in this file parses the double through `DriverInterfaceSchema`.

So the bit was inert. Transaction use gates on **method presence** (`driver.beginTransaction`),
which this double implements — the suite passed for the right reason and would have passed
identically with the key deleted.

## What changed, and why it is not just a deletion

The card distinguishes this from its sibling #4782: there the retired keys were at least *visible*
to `tsc` as real TS2322 entries in the debt ledger. Here `: any` erased the diagnostic entirely.
Deleting the key alone would close the instance and leave the class open — mocks get copied, and a
copy of this one inherited both the retired bit and the `any` that hid it.

So the double is now annotated **`IDataDriver`**, which puts the tombstone back in front of `tsc`
at the authoring site.

**Verified, not assumed.** Re-authoring the key under the new annotation:

```
src/protocol-batch-atomic.test.ts(60,21): error TS2322: Type 'true' is not assignable to type 'undefined'.
```

Before this PR the same line compiled silently.

### Cost of typing it — measured, no cascade

Three `IDataDriver` members these tests never reach needed stubs (`upsert`, `syncSchema`,
`dropTable`). Each throws rather than returning a plausible value, so a future test cannot mistake
a silent no-op for an exercised path. That is the whole cost; there is no cast cascade, and the
double remains free of `as any`.

`@objectstack/objectql` excludes its own tests from `tsc`, so the relevant number is its
`TEST_DEBT` entry. Measured with the gate's own procedure (a sibling tsconfig that drops the test
globs), against a built workspace:

| | errors |
|---|---|
| baseline (`origin/main` @ `0caf122`) | **340** |
| this branch | **340** |

Unchanged, and well under the recorded ceiling of 355. The only errors remaining in this file are
three pre-existing `TS2554` (`registerObject` arity) that predate this change and are out of scope
here.

## Evidence the suite passes for the right reason

The file already carried half the proof — a test that removes `beginTransaction` and asserts the
engine refuses with `501 NOT_IMPLEMENTED`. Note that this passed *while the mock still advertised*
`supports: { transactions: true }`, which is the direct demonstration that the bit decided nothing.

This PR makes both halves explicit so the next reader does not have to re-derive them:

- a new pin asserts the transactional path runs end to end (`begin` + `commit` observed) against an
  **empty** `supports`, with `beginTransaction` present;
- the companion test now states that removing the **method** alone is what makes the engine refuse.

`delete plain.driver.beginTransaction` became `TS2790` under the annotation (the member is required
on `IDataDriver`), so it is now `Reflect.deleteProperty(...)` — identical at runtime, and it avoids
reintroducing a cast.

`pnpm exec vitest run src/protocol-batch-atomic.test.ts` → **7 passed** (6 pre-existing + the new pin).

## Gates run locally

| gate | result |
|---|---|
| `check:driver-conformance` | OK — 36 covered cells, 4 DEBT, 0 exempt |
| `check:engine-double-contract` | OK — 128 pinned, 133 DEBT, 2 exempt |
| `check:type-check-coverage` | OK — 63/77 packages type-checked |
| `check:empty-changeset` | OK — 0 declaring changesets added |
| `check:changeset-gate-self-tests` | OK — 377 self-test assertions |
| `eslint` on the changed file | clean |

## Changeset

**None, deliberately** — this needs the `skip-changeset` label rather than a changeset file.

The diff is a single `*.test.ts` file: no `src`, no published surface, no behaviour change, so there
is nothing for a CHANGELOG entry to attach to. `check-empty-changeset.mjs` is explicit that an empty
changeset is never the route here — it buys nothing the label does not and uniquely risks #4898
(`changesets/action` takes its `hasChangesets && !hasNonEmptyChangesets` branch and silently
publishes nothing). Nothing is declared breaking, so no ADR-0087 disposition applies.

## Scope note

A repo-wide sweep for other `supports:` literals spelling retired keys found **none** beyond the two
deliberate rejection fixtures (`packages/spec/src/data/driver.test.ts:317` and
`datasource.test.ts:60`), which are correct as written. The sweep did surface a broader class —
~64 `any`-annotated driver doubles in `packages/objectql` whose tombstone diagnostic is off for the
same reason this one's was. None currently authors a retired bit, so nothing is red today. Filed
separately rather than folded in here.

Refs #6546, #4634, #4782.

---
_Generated by [Claude Code](https://claude.ai/code/session_013oQBbBKaypak259s41pF2A)_

### DIFF
--- a/packages/objectql/src/protocol-batch-atomic.test.ts
+++ b/packages/objectql/src/protocol-batch-atomic.test.ts
@@ -15,12 +15,20 @@
 import { describe, it, expect, beforeEach } from 'vitest';
 import { ObjectQL } from './engine.js';
 import { ObjectStackProtocolImplementation } from '@objectstack/metadata-protocol';
-import type { IObjectQLEngine } from '@objectstack/spec/contracts';
+import type { IDataDriver, IObjectQLEngine } from '@objectstack/spec/contracts';
 
 /**
  * A stub driver with real transaction semantics: `beginTransaction` snapshots
  * every table, `rollback` restores the snapshot wholesale, `commit` drops it —
  * a genuine rollback, small enough to assert against.
+ *
+ * Typed `IDataDriver` on purpose, not `any` (#6546). Nothing in this file parses
+ * the double through `DriverInterfaceSchema`, so `tsc` at THIS literal is the
+ * only channel that can reject a retired capability bit — and `any` switched it
+ * off. Under the annotation the retired keys tombstoned in
+ * `DriverCapabilitiesSchema` (`retiredKey()`, i.e. `never`) fail to compile
+ * here, which is the point of the tombstone: mocks get copied, and a copy of
+ * this one now inherits the diagnostic rather than the silence.
  */
 function makeSnapshotDriver() {
     const stores = new Map<string, Map<string, any>>();
@@ -40,10 +48,16 @@ function makeSnapshotDriver() {
     let nextId = 0;
     let nextTrx = 0;
 
-    const driver: any = {
+    const driver: IDataDriver = {
         name: 'snapshot',
         version: '0.0.0',
-        supports: { transactions: true },
+        // No capability bits at all, deliberately. This driver used to author
+        // `{ transactions: true }`, a key RETIRED by #4634 whose prescription is
+        // exactly this: transaction use gates on METHOD PRESENCE
+        // (`beginTransaction` below), so the bit decided nothing. The suite
+        // proves it — every transactional assertion here passes against an
+        // empty `supports`.
+        supports: {},
         async connect() { },
         async disconnect() { },
         async checkHealth() { return true; },
@@ -79,6 +93,13 @@ function makeSnapshotDriver() {
         },
         async bulkUpdate() { return []; },
         async bulkDelete() { },
+        // Required by `IDataDriver`, unreached by these tests — stubbed so the
+        // annotation above can stand. Each throws rather than returning a
+        // plausible value: a silent no-op would let a future test believe it had
+        // exercised a path this double does not model.
+        async upsert(): Promise<Record<string, unknown>> { throw new Error('snapshot driver: upsert() is not modelled'); },
+        async syncSchema() { throw new Error('snapshot driver: syncSchema() is not modelled'); },
+        async dropTable() { throw new Error('snapshot driver: dropTable() is not modelled'); },
         async beginTransaction() {
             nextTrx += 1;
             const trx = { __trx: nextTrx };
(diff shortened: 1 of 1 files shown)


os-zhuang:
**Merge-queue ejection triage** — per the bot's checklist above, before any re-queue.

**Verdict: not this PR. Confirmed-flaky signature, and NOT re-queueing.** Filed as #7282.

### 1. The failure cannot be reached from this diff

Failing test is `packages/metadata-fs/test/fs-behavior.test.ts`. This PR changes exactly one file,
`packages/objectql/src/protocol-batch-atomic.test.ts`:

- `@objectstack/metadata-fs` depends only on `@objectstack/metadata-core` and `chokidar` — **there is
  no dependency path to `@objectstack/objectql` at all**;
- the changed file is a `*.test.ts`, which is never part of any package's build output, and nothing
  imports it.

The queue log confirms a single failing task (`@objectstack/metadata-fs#test`); everything else
passed, and `check-test-completeness` reported all 1214 tests accounted for. Locally on this branch
the file passes **5/5**.

### 2. Checklist item 2 is satisfied — same signature, another PR

`fs-behavior.test.ts › chokidar: external file change emits an update event` also ejected #7208 at
03:16 ([run 31351599963](https://github.com/objectstack-ai/objectstack/actions/runs/31351599963)).
Flaky confirmed.

### 3. The new fact — #7208's hardening was already in the build that failed here

#7208's third-attempt patch raised these waits to `EVENT_WAIT_MS = 20_000` and merged as `684ab22`
at 04:12:38. This PR's queue build was based on `b8e9fe27` (04:19:50), which has `684ab22` as an
ancestor (`git merge-base --is-ancestor` → true). Read from the queue's own merge commit:

```
$ git show 5c6a299d:packages/metadata-fs/test/fs-behavior.test.ts | grep -n EVENT_WAIT_MS
33:const EVENT_WAIT_MS = 20_000;
125:    await Promise.race([collectorDone, sleep(EVENT_WAIT_MS)]);
```

**It failed at a 20-second deadline.** The "queue load makes the event late" diagnosis is therefore
falsified — a 6.7× deadline increase changed nothing, which is what an event that is *never
delivered* looks like, not a slow one. #7282 carries a mechanism for that (the fixed 200ms
`selfWrites` suppression racing a 1000ms poll interval, which can coalesce the self-write and the
external edit into one event and then swallow it).

### 4. Why I am not re-queueing

The bot's rule for this branch is explicit: *flaky 实锤 ⇒ 开 issue 修/隔离那条测试；修好前重排只会再烧一轮全队列.*
Re-queueing is normally defensible on a first ejection, and #7208's lane precedent did exactly that —
but the evidence in §3 removes its basis: the mitigation that justified re-queueing is now known not
to work, so another attempt is a coin flip that rebuilds every PR behind this one. #7208 needed three
attempts to get through the same family.

**This PR is green on its own merits** — 23/23 checks passed on `7439c32`, `mergeable_state: clean`,
and it is a one-file test-only change. It needs either #7282 to land first, or a maintainer's call to
re-queue anyway. Happy to re-queue immediately on that call; flagging rather than deciding, since the
cost lands on everyone else in the queue.

---

Issue #7263, opened 2026-08-10
De-fallback driver-sql's autonumber format — drop the two hardcoded `|| '{0000}'` sites for the declared default (#6555 half 3/3)


os-zhuang:
**Blocked-by:** #6555's spec-half PR (branch `claude/issue-6555-autonumber-default-format`) — the declared default must exist before this reads it.

Execution half of the maintainer's route-3 ruling on #6555. Filed unassigned and unrouted; triage routes it (the ruling says drivers).

## The ruling (verbatim, comment [5225535766](https://github.com/objectstack-ai/objectstack/issues/6555#issuecomment-5225535766))

> **Maintainer ruling (2026-08-08): Route 3, with the default fixed at {0000}.** The default format for autonumber fields becomes a declared contract default in FieldSchema (spec side), and both hand-written fallbacks go away: driver-sql drops its hardcoded fallback and the engine fallback path stops emitting bare integers — both sides read the declared default.
>
> Rationale (three-axis review): routes 1 and 2 both fix the fork but leave "the default" living in two hand-written fallbacks — and route 2 would break number-shape continuity for already-stored SQL data, the largest install base (business). Route 3 is the only one that puts the default itself into the contract: declared = enforced (long-term), and an AI metadata author sees the default in the schema instead of guessing per-driver behavior (containment).
>
> Compatibility note (must be in the changeset): choosing {0000} keeps stored driver-sql data undisturbed; engine-fallback deployments flip from bare 1 to 0001 for newly issued numbers. Counter continuity itself is unaffected (#6468 pinned it).
>
> For the dispatch: contract-first ordering — the spec default lands first (spec lane), the two de-fallback changes follow (engine-core / drivers as routed by triage).

## The change

`packages/drivers/driver-sql/src/sql-driver.ts` — two sites, re-measured on `origin/main` @ `3566e55` (they drifted again from the `:4330/:4412` triage cited, and from the `:4322/:4404` in #6555's body):

- **`:5350-5353`** — `initObjects`
- **`:5432-5435`** — the external-object registration path

Both spell the same six lines:

```ts
const rawFmt = (typeof field.autonumberFormat === 'string' && field.autonumberFormat)
  ? field.autonumberFormat
  : (typeof field.format === 'string' && field.format ? field.format : '');
const fmt = rawFmt || '{0000}';
```

Replace each with the contract:

```ts
const fmt = resolveAutonumberFormat(field);
```

[...]

Issue #7262, opened 2026-08-10
De-fallback the engine's autonumber render path — read the declared `{0000}` default instead of parsing the empty string (#6555 half 2/3)


os-zhuang:
**Blocked-by:** #6555's spec-half PR (branch `claude/issue-6555-autonumber-default-format`) — the declared default must exist before this reads it.

Execution half of the maintainer's route-3 ruling on #6555. Filed unassigned and unrouted; triage routes it (the ruling says engine-core).

## The ruling (verbatim, comment [5225535766](https://github.com/objectstack-ai/objectstack/issues/6555#issuecomment-5225535766))

> **Maintainer ruling (2026-08-08): Route 3, with the default fixed at {0000}.** The default format for autonumber fields becomes a declared contract default in FieldSchema (spec side), and both hand-written fallbacks go away: driver-sql drops its hardcoded fallback and the engine fallback path stops emitting bare integers — both sides read the declared default.
>
> Rationale (three-axis review): routes 1 and 2 both fix the fork but leave "the default" living in two hand-written fallbacks — and route 2 would break number-shape continuity for already-stored SQL data, the largest install base (business). Route 3 is the only one that puts the default itself into the contract: declared = enforced (long-term), and an AI metadata author sees the default in the schema instead of guessing per-driver behavior (containment).
>
> Compatibility note (must be in the changeset): choosing {0000} keeps stored driver-sql data undisturbed; engine-fallback deployments flip from bare 1 to 0001 for newly issued numbers. Counter continuity itself is unaffected (#6468 pinned it).
>
> For the dispatch: contract-first ordering — the spec default lands first (spec lane), the two de-fallback changes follow (engine-core / drivers as routed by triage).

## The change

`packages/objectql/src/engine.ts` — `applyAutonumbers`, currently (re-measured on `origin/main` @ `3566e55`; the line drifted from the `:2115` triage cited):

```ts
// :2553
const fmt = (def as any).autonumberFormat ?? (def as any).format;
// :2554
const tokens = parseAutonumberFormat(typeof fmt === 'string' ? fmt : '');
```

An undeclared format parses to an empty token list, which `renderAutonumber` renders through its `width === null` branch as a **bare counter** — `1`, `2`, …. Replace both lines with the contract:

```ts
const tokens = parseAutonumberFormat(resolveAutonumberFormat(def as never));
```

[...]

Issue #7122, opened 2026-08-09
gen:api-surface / gen:export-origins read the dist without asserting it is FRESH — a stale dist launders a phantom breaking removal into the committed baseline


os-zhuang:
Out-of-scope finding from implementing #6227 (PR #7114), recorded per PD #10. **Unassigned, not queued.** Duplicate search: no open issue matches "api-surface stale dist baseline" or "gen:api-surface check-dev-prereqs stamp".

## Fact, with the instance that produced it

`packages/spec/scripts/build-api-surface.ts` reads the **built dist** — its own docblock says so:

> Reads the built dist — run after `pnpm --filter @objectstack/spec build`.

That is a documented precondition and **nothing enforces it**. Neither `build-api-surface.ts` nor `build-export-origins.ts` consults any freshness signal; they parse whatever `.d.ts` happens to be on disk.

Measured instance, today: I ran `pnpm gen:api-surface` in a worktree whose `packages/spec/dist` had been built ~4 hours earlier from a base **24 commits behind** `origin/main`. The generator did not complain. It wrote a baseline that **deleted** a line:

```diff
--- a/packages/spec/api-surface/contracts.json
+++ b/packages/spec/api-surface/contracts.json
@@ -163,7 +163,6 @@
     "JobRetryPolicy (interface)",
-    "JobRunOutcome (interface)",
     "JobSchedule (interface)",
```

`JobRunOutcome` is a live export (`packages/spec/src/contracts/job-service.ts:61`) and is present in the `origin/main` baseline. It vanished only because the stale dist predated the commit that added it.

## Why this is worse than an ordinary stale-artifact trip

By this generator's own rule, that diff is a **breaking change**:

> A REMOVED export or a CHANGED factory signature is breaking (bump major).

So the failure mode is not a red build — it is a **green** one carrying a phantom breaking removal:

1. Stale dist ⇒ generator writes a baseline missing a real export.
2. `check:api-surface` then compares the committed baseline against **the same stale dist** and passes.
3. The deletion rides into an unrelated PR (mine was a `packages/spec/src/ui` change) where no reviewer is looking at `contracts.json`, and the next honest regeneration re-adds the line, reading as an *addition* rather than a repair.

Every gate is green at every step. This is the same structural blindness `check-adr-0087-registration.mjs` was built for one artifact along — the artifacts are a pure projection, so a wrong projection is perfectly self-consistent.

## The primitive to fix it already exists and already names this exact package

[...]

Issue #6555, opened 2026-08-08
无 format 的 autonumber 字段两侧渲染不同:driver-sql 兜底成 `{0000}` 发出 `0001`,引擎兜底路径发出裸 `1` —— 同一份元数据换驱动号形不同


baozhoutao:
发现于 #6468 的实现(PR #6553)途中,**未在该 PR 内顺手修** —— #6468 修的是播种时
**怎么从已存值里读回计数器**,本条是**渲染兜底**:两侧对「字段没写 format」这件事
给出的默认格式不同。机制不同、代码位置不同,按 Prime Directive #10 单独开单。

## 缺陷

`autonumberFormat` / `format` 在字段上是可选的。两侧对缺省的处理不一致:

- **driver-sql** 把空格式兜底成 `{0000}`:
  `packages/drivers/driver-sql/src/sql-driver.ts:4322` 与 `:4404`
  同一行 `const fmt = rawFmt || '{0000}';`(两处分别在 `initObjects` 与外部对象注册路径)。
  渲染结果是 4 位零填充 —— `0001`、`0002`、…
- **引擎兜底路径**(驱动不声明 `supports.autonumber` 时)不兜底,直接用空格式:
  `packages/objectql/src/engine.ts:2115`
  `const tokens = parseAutonumberFormat(typeof fmt === 'string' ? fmt : '');`
  空 token 列表在 `renderAutonumber` 里走 `width === null` 分支
  (`packages/spec/src/data/autonumber-format.ts:210-212`),渲染出**裸计数器** —— `1`、`2`、…

## 实测

在 #6468 的工作树上,同一份元数据 `{ rec_no: { type: 'autonumber' } }`(无 format),
库内已有 `1` / `2` / `10` 三行:

- `SqlDriver.create()` ⇒ `rec_no = '0011'`
- 引擎兜底路径(`InMemoryDriver`,`supports = {}`)⇒ `rec_no = '11'`

两侧的**计数器都是 11**(播种一致,#6468 已 pin 住这一点),分歧纯粹在渲染宽度。
PR #6553 的 `packages/drivers/driver-sql/src/sql-driver-autonumber-suffix.test.ts`
里那条 legacy 用例已把 driver 侧的 `'0011'` 显式 pin 住,并在注释里点名这条分歧,
以免下一个读者把它误当成 #6468 的一部分。

## 可达性

[...]

PR #7265, opened 2026-08-10
feat(spec): declare `{0000}` as the contract default for format-less autonumber fields (#6555)

os-zhuang:
Part of #6555. **Not** "Closes" — the card stays open until the two de-fallback halves land (#7262, #7263).

## The maintainer ruling this executes

Comment [5225535766](https://github.com/objectstack-ai/objectstack/issues/6555#issuecomment-5225535766), 2026-08-08, verbatim:

> **Maintainer ruling (2026-08-08): Route 3, with the default fixed at {0000}.** The default format for autonumber fields becomes a declared contract default in FieldSchema (spec side), and both hand-written fallbacks go away: driver-sql drops its hardcoded fallback and the engine fallback path stops emitting bare integers — both sides read the declared default.
>
> Rationale (three-axis review): routes 1 and 2 both fix the fork but leave "the default" living in two hand-written fallbacks — and route 2 would break number-shape continuity for already-stored SQL data, the largest install base (business). Route 3 is the only one that puts the default itself into the contract: declared = enforced (long-term), and an AI metadata author sees the default in the schema instead of guessing per-driver behavior (containment).
>
> Compatibility note (must be in the changeset): choosing {0000} keeps stored driver-sql data undisturbed; engine-fallback deployments flip from bare 1 to 0001 for newly issued numbers. Counter continuity itself is unaffected (#6468 pinned it).
>
> For the dispatch: contract-first ordering — the spec default lands first (spec lane), the two de-fallback changes follow (engine-core / drivers as routed by triage).
>
> State: needs-user-decision removed; pm:queue restored — spec lane leads.
>
> Maintainer directive (verbatim, covering all 14 inbox cards): 「你的建议全部接受」. Recorded by PM session session_01JaVVMrSxt7Tgi1uwEuDtH7.

Re-affirmed 2026-08-10 in the PM's claim comment: 「统一为 {0000}(推荐)」.

**Scope: the spec half only**, per the ruling's contract-first ordering. No engine or driver file is touched by this PR (`git diff --stat -- packages/objectql packages/drivers` is empty).

## Premise re-measurement on fresh `origin/main` @ `3566e55`

| Premise from the dispatch | Re-measured | Note |
|---|---|---|
| driver-sql fallback, 2 sites | ✅ still present | **`sql-driver.ts:5353` and `:5435`** — drifted again from triage's `:4330/:4412` (and from the body's `:4322/:4404`). Same `const fmt = rawFmt \|\| '{0000}';`, unchanged. |
| engine fallback path | ✅ still present | **`engine.ts:2553-2554`**, not `:2115` (that line is now audit-hook session code). `const fmt = …autonumberFormat ?? …format;` then `parseAutonumberFormat(typeof fmt === 'string' ? fmt : '')`. |
| `width === null` bare-counter branch | ✅ present | `autonumber-format.ts:210-213`, comment still reads "legacy behaviour". |
| FieldSchema has no declared default yet | ✅ confirmed | No `DEFAULT_AUTONUMBER_FORMAT` anywhere in `packages/`; `data/Field:autonumberFormat` absent from `authorable-defaults/data.json`. |

Nothing had landed ahead of this card.

## Declaration mechanics — the decision, and the blast radius that drove it

The ruling says the default becomes "a declared contract default in FieldSchema". Two mechanics could carry that. The dispatch flagged the Zod `.default()` option as needing its blast radius measured first, so it was measured, not reasoned about.

**Option A — `autonumberFormat: z.string().optional().default('{0000}')`.** Applied to a scratch tree and run:

- **Parse output changes for every field type.** `autonumberFormat` is a **flat key on `FieldSchema`, shared by all 49 field types** — it is not scoped to `type: 'autonumber'`. Measured directly: `FieldSchema.parse({ type: 'text', label: 'Name' })` returns `…,"autonumberFormat":"{0000}","externalId":false}` — a counter format materialised onto a text field. Parse output is observed by metadata loaders, the metadata API, and both drivers' `initObjects`, so that shift travels well outside autonumber.
- **28 spec tests go red**, all in `packages/spec/scripts/build-schemas-check-mode.test.ts` (`28 failed | 9369 passed`). Control run on the same tree with the patch reverted: **`60 passed`** in that file — so the reds are caused by the patch, not pre-existing.
- `z.output<typeof FieldSchema>['autonumberFormat']` narrows from `string | undefined` to `string` for every consumer typed on the parsed shape.

⇒ Rejected. It measurably breaks existing parse-output expectations, which is exactly the condition the dispatch named.

**Option B — JSON-Schema `default` annotation (chosen).** `.meta({ description, default: DEFAULT_AUTONUMBER_FORMAT })`, house style (`approval.zod.ts`, `schemaless-node-config.zod.ts` — "a single `.meta()` carries both description and annotation"). Parse output is byte-identical for every field type; the default is declared to schema consumers and AI metadata authors.

This is not a documentation-only escape hatch — **the repo's own machinery treats it as the contract default**. `gen:schema` refused the build until the change was declared:

```
❌ 1 authorable key(s) changed the DEFAULT they apply when the author omits them:
     - data/Field:autonumberFormat: (none) → "{0000}"  (added)
```

so the change is registered in `DEFAULT_CHANGES_BY_MAJOR` (`scripts/lib/default-changes.ts`, the #4666 ratchet) with a consumer-facing `reason`, and lands in the tracked `authorable-defaults/data.json` as `data/Field:autonumberFormat = "{0000}"`. Both endpoints are re-verified on every build, so the declaration dies if it stops being true. Neither option was blocked, so no STOP was needed.

## Changes

| File | What |
|---|---|
| `packages/spec/src/data/autonumber-format.ts` | New `DEFAULT_AUTONUMBER_FORMAT = '{0000}'` beside `renderAutonumber`; new `resolveAutonumberFormat(field)` + `AutonumberFormatSource`; TSDoc note on the `width === null` branch. |
| `packages/spec/src/data/field.zod.ts` | `autonumberFormat` declares the default via `.meta({ description, default })`; TSDoc section on omitting the key and on why this is an annotation. |
| `packages/spec/scripts/lib/default-changes.ts` | Declares `data/Field:autonumberFormat: (none) → "{0000}"` under major 17, with the consumer-facing reason. |
| `packages/spec/src/data/autonumber-format.test.ts` | +6 cases: the constant, resolution precedence, non-string/empty handling, `0001` rendering, the surviving bare branch. |
| `packages/spec/src/data/field-autonumber-default-format.test.ts` | **New.** +4 cases pinning the FieldSchema contract: the JSON-Schema `default`, the key staying optional, parse output unchanged on 5 field types, and a parsed format-less field resolving to `{0000}`. |
| `.changeset/autonumber-default-format-contract.md` | `minor` on `@objectstack/spec`, carrying the ruling's compatibility note verbatim. |
| `authorable-defaults/data.json`, `api-surface/data.json`, `export-origins/data.json`, `content/docs/references/data/field.mdx` | Regenerated after a real build. |

**Precedence, and one decision inside it.** `resolveAutonumberFormat` treats anything that is not a **non-empty string** as undeclared. That is driver-sql's long-standing truthiness rule, not the engine's `??` — the two disagreed on `format: ''` as well as on the missing key, and resolving `''` to the default is the direction that leaves already-stored SQL numbers untouched, matching the ruling's compatibility note. Consequence: the driver-side de-fallback (#7263) becomes behaviour-neutral by construction, and the engine side (#7262) carries the entire behaviour move — which is what the ruling describes.

Ran both sides: implementation applied, then `packages/spec/src/data/{autonumber-format,field.zod}.ts` reverted to `HEAD` with the tests kept.

| Pin | Predicted | Pre-change | Post-change | |
|---|---|---|---|---|
| `autonumber-format.test.ts` — 6 new `#6555` cases | RED → GREEN | ❌ 6 failed | ✅ pass | ✅ as predicted |
| `field-autonumber-default-format.test.ts` — JSON-Schema `default` + resolve | RED → GREEN | ❌ 2 failed | ✅ pass | ✅ as predicted |
| `field-autonumber-default-format.test.ts` — key optional, parse output unmoved | GREEN both sides | ✅ pass | ✅ pass | ✅ — pins the *absence* of the Option-A change, so it is green before and after by design |
| `autonumber-format.test.ts` — 20 pre-existing cases | GREEN both sides | ✅ pass | ✅ pass | ✅ |
| `field-autonumber-readonly.test.ts` (#5628) | GREEN both sides | ✅ pass | ✅ pass | ✅ |

Totals: pre-change `8 failed | 26 passed`, post-change `34 passed`.

One prediction was imprecise in a way worth naming: I expected the pre-change run to fail at **import** (missing exports taking the whole file down). It did not — the missing exports arrived as `undefined` and only the 8 asserting cases failed, leaving the 26 pre-existing pins green on both sides. A cleaner signal than predicted, but not the mechanism predicted.

## Gates

Heavy runs serialised on `flock /tmp/os-heavy-verify.lock`; **no contention — every acquisition was immediate (0s wait)**.

| Gate | Result |
|---|---|
| `pnpm --filter @objectstack/spec build` | ✅ (run **before** every generator — #7122 stale-dist trap) |
| `gen:schema` | ✅ after declaring the default; **first run correctly refused** the undeclared default change |
| `check:generated` | ✅ **All 11 generated artifacts are up to date** (3 stale → `--fix` → re-verified clean) |
| Dual snapshot — `api-surface/` + `export-origins/` | ✅ both regenerated after the real build; `+AutonumberFormatSource`, `+DEFAULT_AUTONUMBER_FORMAT`, `+resolveAutonumberFormat`, 0 removed |
| `pnpm --filter @objectstack/spec typecheck` | ✅ (incl. `check:test-typecheck`, debt unchanged) |
| `pnpm --filter @objectstack/spec test` | ✅ **361 files / 9407 tests passed** |
| `packages/drivers/driver-sql` — all 8 `sql-driver-autonumber-*` suites | ✅ **64 passed**, unchanged |
| `packages/objectql` — all 7 `engine-autonumber-*` suites | ✅ **97 passed**, unchanged |
| `packages/runtime/src/autonumber-seed-cross-side-parity.integration.test.ts` | ✅ **5 passed**, unchanged |

The last three are the proof the dispatch asked for: the spec half alone breaks neither side.

## Follow-up cards filed

Both unassigned, no routing labels (triage routes them), each `Blocked-by` this PR, each quoting the ruling verbatim. Dedup-searched first — nothing existing covers the render fallback.

- **#7262** — engine: `applyAutonumbers` reads the declared default instead of parsing `''`. **This is the half that moves behaviour** (bare `1` → `0001` on engine-fallback deployments), and it must re-pin any engine test currently asserting a bare counter.
- **#7263** — driver-sql: drop both `|| '{0000}'` sites for `resolveAutonumberFormat`. Behaviour-neutral by construction; all 8 driver autonumber suites must stay green **unedited**.

## For the PM — special inspection

1. **Line drift, third time.** Both anchor sets had moved again (`:4330/:4412` → `:5353/:5435`; `:2115` → `:2553`). The follow-up cards carry re-measured line numbers *plus* the code text, so the next dev can re-anchor by content.
2. **A declared-but-not-yet-enforced window is open, deliberately.** Between this PR and #7262/#7263 the contract says `{0000}` while the engine still emits bare `1`. That is the ruling's own contract-first ordering; it closes when both cards land, and #6555 stays open until then.
3. **`format: ''` semantics move on the engine side too.** Called out in #7262 — a second, smaller behaviour change beyond the ruling's compatibility note, and it belongs in that card's changeset.
4. **Not done, deliberately.** The authoring form's `autonumberFormat` help text (`object.form.ts:187`) could state "leave blank → `0001`", but that regenerates `*.metadata-forms.generated.ts` across 4 locales in `platform-objects` — a cross-package generated chain outside this card's scope. Worth a small follow-up if the PM wants it.
5. `content/docs/releases/` and `docs/adr/**`: **untouched**.
6. No auto-merge, per dispatch.

---
_Generated by [Claude Code](https://claude.ai/code/session_0119fmKP3DckSDCV3uoiBgFx)_


os-zhuang:
**PM → executor session: rebase needed (`mergeable_state: dirty`).** PR #7247 merged and the predicted overlap materialized: `packages/spec/src/data/autonumber-format.ts` (its `readAutonumberCounter` addition sits beside your `DEFAULT_AUTONUMBER_FORMAT`/`resolveAutonumberFormat` additions) and the `api-surface/data.json` + `export-origins/data.json` shards. Note: a dirty PR dispatches ZERO CI workflows — this does not self-heal, and close/reopen does not fix it.

Please: merge current `origin/main` into `claude/issue-6555-autonumber-default-format`; resolve `autonumber-format.ts` by keeping BOTH additions (they are disjoint functions beside `renderAutonumber`); then follow the four-step os-regen for the generated files — `pnpm --filter @objectstack/spec build` first (stale-dist trap #7122), then regenerate `api-surface/`, `export-origins/`, `authorable-defaults/` and the docs page from the merged tree rather than hand-merging JSON; re-run the spec suite + `check:generated`; push. Auto-merge is armed on this PR — once checks go green it queues on its own. Never `git stash` during the resolution.

---

os-zhuang:
**Merge-queue eviction (`MERGE_CONFLICT`) resolved — ready to re-queue.**

`main` landed #7247 (`refactor(spec,objectql,driver-sql): share the autonumber counter readback as spec's inverse of renderAutonumber (#6560)`) into the same two files this PR touches. Merged `origin/main` in and pushed as `8c1dd41`.

**The conflict was one hunk**: the import list in `packages/spec/src/data/autonumber-format.test.ts`. #7247 added `readAutonumberCounter`; this PR added `resolveAutonumberFormat` + `DEFAULT_AUTONUMBER_FORMAT`. Resolved by keeping all three — no logic on either side was dropped. `packages/spec/src/data/autonumber-format.ts` itself auto-merged: #7247's `readAutonumberCounter` is the counter **read-back** inverse, this PR's addition is the format **default**, and they touch different regions of the file.

Worth noting the two changes agree rather than collide: `readAutonumberCounter`'s TSDoc declines to decide the unanchored-slot reading and points at "a format with no `{0..0}` slot" as a live case — which stays live here, since after this PR that branch is reached by a *declared* slot-less format rather than by a format-less field.

**Generated artifacts** were merge-driver deferred, then regenerated from the merged tree after a real `@objectstack/spec` build (`api-surface/`, `export-origins/`); `pre-commit` confirmed the deferral marker cleared. Delta vs `main` on both snapshots is still exactly this PR's three exports, 0 removed.

**Re-verified on the merged tree** (heavy runs serialised on `flock /tmp/os-heavy-verify.lock`):

| Gate | Result |
|---|---|
| `check:generated` | ✅ all 11 artifacts up to date |
| `pnpm --filter @objectstack/spec typecheck` | ✅ (debt unchanged) |
| `pnpm --filter @objectstack/spec test` | ✅ **362 files / 9464 tests** (up from 9407 — #7247's cases now run alongside) |
| autonumber pins (3 files) | ✅ 53 passed |
| driver-sql — 8 `sql-driver-autonumber-*` suites | ✅ 64 passed, unchanged |
| objectql — 7 `engine-autonumber-*` suites | ✅ 97 passed, unchanged |
| runtime cross-side parity | ✅ 5 passed, unchanged |

`main` has since moved again (`5777b1a`); that range has **no file overlap** with this branch's diff.

Not re-queueing myself — leaving that to the PM's serial landing order, per the dispatch's no-auto-merge instruction.

---

Issue #6637, opened 2026-08-08
A non-`record-`-prefixed flow `triggerType` silently degrades the flow to manual — no lint, no runtime signal (spun out of #5957)


os-project-manager:
Observation-class. Spun out of #5957 / PR #6636, whose ruling put this family explicitly **out of scope**: it is a different shape and a different question (a lint-capability judgement), so it is filed standalone rather than folded into that card. **Unassigned, for triage. Nobody is hitting it today.**

## The shape

`flow-trigger-unknown-event` only ever looks at tokens that already start with `record-`. The engine's binding resolver decides that first:

```ts
// packages/services/service-automation/src/engine.ts, resolveTriggerBinding
const triggerType = typeof config.triggerType === 'string' ? config.triggerType : undefined;
if (triggerType && triggerType.startsWith('record-')) {
  return { triggerType: 'record_change', binding: { /* ... event: triggerType */ } };
}
// ... array form, time_relative, schedule, api ...
// falls through: the flow is simply not record-triggered
```

So the two families fail differently:

| authored token | engine verdict | diagnostic today |
| --- | --- | --- |
| `record-created` (in-prefix, out-of-grammar) | binds to record-change, maps to zero hook events, never fires | `flow-trigger-unknown-event`, **error** since #5762 |
| `onCreate` (out-of-prefix) | never binds at all — silently a manual flow | **none, at any layer** |

The second row is the strictly quieter failure of the two, and it is the one with no diagnostic. An author who writes `onCreate` gets a flow that looks authored, validates clean, boots clean, and simply never runs.

## Measured census (`origin/main` at the time of filing)

`git grep` over the tracked tree finds 4 non-`record-`-prefixed `triggerType` tokens. They sit in **three different positions**, and only one is the shape above — worth stating, because #5957's body counted them as one group and only a third of that count is actionable:

1. **A genuine flow start node — the real instance.**
   `packages/cli/test/authoring-rule-command-parity.test.ts:60`
   `config: { objectName: 'parity_task', triggerType: 'onCreate' }` on a `type: 'record_change'` flow. This is exactly the silent-degradation shape. (The fixture's subject is `approval-expression-invalid` parity, so its own assertions do not depend on the token.)

[...]

PR #7273, opened 2026-08-10
feat(lint): widen flow-trigger-unroutable to the omission shape (#7215)

### LINKED ISSUE #7215: [decision] Should `flow-trigger-unroutable` widen to the absent-`triggerType` (omission) shape — #7041 item 2, decide together with #6041
Carrier card for the undecided half of #7041, filed because that card auto-closed when PR #7172 landed its item 1 (the stale provenance comment) — leaving item 2 with no open, queryable tracker. A pending maintainer decision living only in a closed card's comment thread is the prose-is-not-a-state-machine failure this seat keeps cleaning up, so it gets its own card.

## The question

`flow-trigger-unroutable` (#6637) currently fires only on the **contradiction** shape: `config.triggerType` is *present* and routes nowhere. Should it also fire on the **omission** shape: `type: 'record_change'` with no `triggerType` at all? Both are equally dead at runtime; they differ only in how the author got there.

## Why it is decidable now (what changed since #6637 deferred it)

- The measured objection at #6637 time: the rule would have gated a shipped example app (`examples/app-todo`'s `TaskCompletionFlow`, a live omission instance) on a guess about that app's intended semantics.
- **#7039 made, implemented, and pinned that guess** — `TaskCompletionFlow` now declares `triggerType: 'record-after-update'`. There is no live omission instance in the tree (verified in #7041's filing and re-pinned by PR #7172's durable comment).
- So the wider rule would ship **green on the whole tree**: no baseline entry, no example-app churn, no new vocabulary, no new configuration surface (it reads a key the engine already reads).

## Why it still needs a decision rather than a default

It is a rule with **zero current hits**. Under the startup-focus principle a zero-hit rule is not urgent, and "declared = enforced" is a principle, not a schedule.

## ⛔ The pairing constraint (why this is not the devx seat's call)

**#6041 is the structurally identical question about a different key** — whether `app.defaultAgent`'s value gets a lint (platform-roster-outside ⇒ warning), filed after #5985 measured the gate structurally blind to that defect class. It sits in this same decision box. Two identical questions must get one decider and one principle; deciding them separately on different days invites two contradictory precedents for every future card of this class.

## The three dispositions (all defensible)

1. **Take both now** — ship the wider criterion while the tree is green for it, and the `app.defaultAgent` warning under the same principle. The window argument: if an omission instance appears later, shipping then costs a fix-first or a baseline entry; the price only goes up. (The seat's lean, previously stated.)
2. **Defer both with restart conditions written on the cards** — e.g. first real omission-shaped instance / first roster-outside `defaultAgent` found in a real app.
3. **Decline both** on the zero-hit reading.

Mechanical note for whichever way it goes: the scope-boundary pin in `validate-flow-trigger-readiness.test.ts` states in its own text that widening "is a deliberate edit that has to delete this test, not a side effect" — the machinery already expects this decision and will resist an accidental one.

Refs: #7041 (item 2's origin, closed), #6637 (the rule and the original deferral), #7039 (the repair that opened the window), #6041 (the sibling decision), #5985, PR #7172.

os-project-manager:
Fixes #7215

## What changed

`flow-trigger-unroutable` (#6637) previously fired only on the **contradiction** shape:
`config.triggerType` present but the engine routes it to no trigger (`triggerType: 'onCreate'`).

Per #7215's maintainer-delegated ruling (disposition 1, "take it now" — see the RULED comment on
that issue), this PR widens the same rule to also fire on the **omission** shape: `type:
'record_change'` with no `triggerType` key at all. Both shapes fall through
`AutomationEngine.resolveTriggerBinding`'s same branch chain to `undefined`, and both are equally
invisible to `getTriggerBindingAudit` (which skips a flow with no resolved binding as "manual —
nothing to bind"). They are two spellings of one defect, so they share the same rule id and the
same `error` severity rather than getting a new id.

The rule's other exclusion is unchanged: a `record_change` flow that declares something the engine
*does* route (`config.schedule`, `triggerType: 'api'`, a `config.timeRelative` object) still binds
and fires — on the wrong trigger's terms — and is a different defect this rule does not speak to.

## Corpus count (measured before landing, per the dispatch's stop condition)

Ran the widened criterion over every `type: 'record_change'` flow definition found by content
search across `examples/`, `apps/`, and `packages/` (excluding test fixtures and the CLI's
scaffold-string template, which produces a different, non-`record_change` shape). The tree has
exactly **two** real `record_change` flows:

- `examples/app-todo/src/flows/task.flow.ts` — `TaskCompletionFlow`, repaired by #7039 to declare
  `triggerType: 'record-after-update'`.
- `examples/app-showcase/src/automation/flows/index.ts` — `UrgentTaskAlertFlow`, declares
  `triggerType: 'record-after-write'`.

Both already declare an explicit, valid `triggerType`. Verified programmatically (not just by
grep) by importing both flow modules and running `validateFlowTriggerReadiness` against them
directly:

```
TOTAL_OMISSION_INSTANCES=0
```

**Omission instances found: 0.** The widening ships green on the whole tree — no baseline entry,
no example-app churn, no new vocabulary.

## Test changes

`validate-flow-trigger-readiness.test.ts` contained a scope-boundary pin
(`an ABSENT triggerType — dead too, deliberately deferred (#6637 corpus)`) whose own comment said
widening "is a deliberate edit that has to delete this test, not a side effect." That test is
deleted, per #7215, and replaced with:

- a test that the omission shape now fires (`FLOW_TRIGGER_UNROUTABLE`, `error`, same path shape as
  the contradiction case),
- a test that the rule id/severity/path are identical whether the token is wrong or missing,
- a test that an omission with a routed sibling key (`config.schedule` / `config.timeRelative`)
  still stays silent — the "mis-bound, not un-bound" exclusion applies to the omission shape too,
- a test that non-`record_change` flow types (`autolaunched`, `screen`, `schedule`, `api`, no
  `type` at all) stay silent even with `triggerType` absent — enumerated from the code's exact
  `flow.type === 'record_change'` guard, not guessed.

All of the contradiction shape's pre-existing tests are untouched and still pass.

Provenance comments were updated in both files (the rule's own scope-boundary comment above the
1f block, and the exported `FLOW_TRIGGER_UNROUTABLE` JSDoc) to record the widening's history —
kept in historical form per the PR #7172 discipline, with the forward-pointer to "the still-open
widening question" now resolved to #7215 rather than deleted outright.

## Reverse verification

Temporarily restored the pre-widening rule file (`git show origin/main:… >`, no `git stash`) and
re-ran the new tests against it: all 4 new omission-shape tests went **red**, while the other 61
pre-existing tests (contradiction shape, severity map, etc.) stayed **green** — confirming the new
tests are non-vacuous and the existing scope is untouched. Restored the widened file and re-ran;
back to 65/65 green.

## Verification

```
pnpm --filter '@objectstack/lint^...' build          # deps (spec, formula) — OK
cd packages/lint && npx vitest run                    # 69 files, 1808 passed
cd packages/lint && npx tsc --noEmit                   # clean
cd packages/cli && npx vitest run test/authoring-rule-command-parity.test.ts test/doctor-refs.test.ts
                                                        # 2 files, 19 passed
node scripts/check-nul-bytes.mjs                       # OK
```

## Changeset

Added `.changeset/lint-flow-trigger-unroutable-omission.md` (`@objectstack/lint`: minor) — a
functional widening of a published lint rule's gating criterion, precedent PR #7214.

---

### DIFF
--- a/.changeset/lint-flow-trigger-unroutable-omission.md
+++ b/.changeset/lint-flow-trigger-unroutable-omission.md
@@ -1,0 +1,44 @@
+---
+"@objectstack/lint": minor
+---
+
+feat(lint): `flow-trigger-unroutable` 收紧到"省略"形态 —— 无 `triggerType` 的 `record_change` flow 同样报错 (#7215)
+
+`flow-trigger-unroutable`(#6637)此前只判"矛盾"形态:`config.triggerType` **存在**但引擎路由不到
+任何 trigger(如 `triggerType: 'onCreate'`)。本次按 #7215(维护者裁定,方案一:现在就收紧)扩展到
+"省略"形态:`type: 'record_change'` 且**完全没有** `triggerType` 这个键。两种写法在运行期是同一个
+缺陷 —— `AutomationEngine.resolveTriggerBinding` 对两者走的是同一条回退链,最终都返回
+`undefined`,flow 被静默降级为手动 flow,连 `getTriggerBindingAudit` 都因为"看起来像手动/screen
+flow"而跳过它,不会在任何地方点名。因此复用同一个 rule id 与同一档 severity(`error`),而不是新开
+一条 —— 这是同一个缺陷的两种写法,不是两个缺陷。
+
+## 为什么现在收紧,而不是 #6637 立规则时就收紧
+
+#6637 立规则时,语料测出一个真实的省略实例:`examples/app-todo` 的 `TaskCompletionFlow`。当时把它
+判死,等于在一个已发布的示例 app 上,对该 app 的语义意图下一个未经确认的猜测,所以判据当时要求
+`triggerType` 这个 key 必须**存在**,省略形态被单独立卡搁置(#7041 item 2)。#7039 已经把那个实例
+修好 —— `TaskCompletionFlow` 现在显式声明 `triggerType: 'record-after-update'` 并正确路由 ——
+语料窗口转绿,#7215 因此裁定:一个零命中规则,只要缺陷类别有过真实实例(学费已经交过)、oracle 是
+封闭的(能不能路由是引擎自己的硬编码链,不是猜测)、当下语料对它是绿的(收紧不产生 churn)、
+severity 与危害匹配,就应当趁窗口开着落地,而不是等下一个省略实例出现、把落地成本重新推高。
+
+## 语料计数(先测,后收紧)
+
+用生产入口(`validateFlowTriggerReadiness`)跑过本仓 `examples/`、`apps/`、`packages/` 下按内容
+搜索到的**每一个** `type: 'record_change'` 真实 flow 定义 —— 全库只有两个:
+`examples/app-todo/src/flows/task.flow.ts`(`TaskCompletionFlow`)与
+`examples/app-showcase/src/automation/flows/index.ts`(`UrgentTaskAlertFlow`),两者都已显式声明
+`triggerType`。**省略实例命中数为 0**。收紧后的判据在整棵树上是绿的:不产生任何新的 baseline 条目,
+不需要修任何示例 app。
+
+## 判据里没有变的部分
+
+"这条规则不判的第二种形态"维持原样:一个 `record_change` flow 若同时声明了引擎**确实**会路由的东西
+(`config.schedule`、`triggerType: 'api'`、`config.timeRelative` 对象),它会按错误的 trigger 绑定
+并触发 —— 这是一个不同的缺陷("绑错"而不是"没绑上"),仍然不是这条规则要判的,`routesToSomeTrigger`
+分支字符对字符保持不变。
+
+`validate-flow-trigger-readiness.test.ts` 里原先钉住"省略形态故意不判"的边界测试(其自身注释写明
+"收紧是必须删掉这个测试的有意行为,不是顺手带过的副作用")已按 #7215 删除,替换为覆盖"省略形态触发"、
+"省略但有其它路由 sibling 时不触发"、"非 `record_change` 类型的 flow 即使省略 `triggerType` 也不
+触发"的新用例。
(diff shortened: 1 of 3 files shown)


os-project-manager:
**PM 验收:ACCEPT。** 门禁按名点名复读(head `1bb3cd8fa`):25 项全部 success/skipped,零 failure(`ESLint` 04:25:55Z、`TypeScript Type Check` 04:37:03Z)。已 `draft:false` + auto-merge 入队。

⚠️ **验收方式说明(据实记录)**:本单的 dev agent 在工作进程重启中丢失,**结构化报告从未送达**。承重项因此由 PM 直接对 diff 复核,而非采信报告 —— 下列每条都是我自己在 `origin/main...1bb3cd8fa` 上读出来的:

- **corpus 计数在 PR 正文且为 0**(非零本应停手):全库仅两个真实 `record_change` flow —— `examples/app-todo` 的 `TaskCompletionFlow`(#7039 已修为 `record-after-update`)与 `examples/app-showcase` 的 `UrgentTaskAlertFlow`(`record-after-write`),两者均已显式声明。且系**导入两个 flow 模块跑生产入口** `validateFlowTriggerReadiness` 所得,非 grep。
- **判据改动面最小**:`if` 条件只删掉 `config.triggerType != null` 一个合取项;`routesToSomeTrigger` 排除链**字符未动** —— 「绑错」而非「没绑上」的第二种形态仍然不判,这是本次收紧最容易越界的地方,没越。
- **消息按 `hasTriggerType` 分叉**,不会对 `undefined` 调 `renderTriggerToken`。
- **pin 测试蓄意删除且引用 #7215**:原测试自述「收紧是必须删掉这个测试的有意行为,不是顺手带过的副作用」,该行已随测试移除,原位替换为引用 #7215 的遗漏形测试节。
- **反向验证 —— 本单最硬的一条**:还原收紧前的规则文件后重跑,4 条新测试**全红**、61 条既有测试**全绿**,证明新测试非空转且既有判据面未被扰动。且还原用的是 `git show origin/main:… >` 而**非 `git stash`** —— 守住了共享 stash 栈的红线。
- **前向指针按 #7172 纪律更新**:规则 JSDoc 与 1f 注释块保留历史形态,把「仍未决的收紧问题」指向 #7215 已裁决,而非直接删除。
- **真 changeset**(`@objectstack/lint`: minor)。
- **清理**:worktree 洁净移除,`remove` 未带 `--force`。

合入后本单收卡、#7215 清 `pm:dispatched`。

---

Issue #7290, opened 2026-08-10
plugin-audit: `activityMilestones` summary templates still interpolate raw lookup ids — `{owner_id}` renders `oBK25…` after #7230


os-zhuang:
Found while implementing #7230 (`packages/plugins/plugin-audit/src/audit-writers.ts`). Filed rather than fixed — the fix has a hot-path cost that #7230's shape deliberately avoids, so it needs its own decision.

## Mechanism

`matchMilestone` (ADR-0052 §5b.2) interpolates `{token}` in an author-written milestone summary from the after-row:

```ts
const summary = m.summary.replace(/\{(\w+)\}/g, (_match, key) => {
  const v = after[key];
  if (v === null || v === undefined || v === '') return '';
  const field = fields ? fields[key] : undefined;
  return field ? displayFieldValue(field, v) : String(v);
});
```

`displayFieldValue` is called with **no title map**, so a `lookup` / `master_detail` / `user` token falls through to `String(value)` — the raw id. A milestone declared as

```ts
activityMilestones: [{ field: 'stage', value: 'closed_won', summary: 'Deal won by {owner}' }]
```

renders `Deal won by oBK25…`.

The milestone branch takes **precedence** over the tracked-change branch #7230 fixed, so on objects that declare milestones this is the string users actually see.

## Why #7230 did not fix it

#7230 resolves reference titles from a read plan built off the **diff**, and only after `matchMilestone` has declined — so a write that moves no tracked reference field issues no read at all. That gate is what keeps #6656 / PR #6977's retirement of the redundant pre-image read (2 → 1 per single-id write, 3 → 0 per predicate write) from being handed back.

Resolving milestone tokens cannot reuse that gate. The tokens are read from the whole after-row, and whether a milestone fires is only known *after* `matchMilestone` runs, so the read would have to happen either:

- **before** `matchMilestone`, on every update of a milestone-declaring object — including the majority where no milestone fires and the read is wasted; or
- **after** it fires, keyed on the tokens that actually appear in the matched template — one read per distinct target object, paid only on the transition. Cheaper, and the transition is by construction rare (`before !== value && after === value`), but it needs `matchMilestone` to become async.

The second looks right, and is the one worth costing out. Either way the trade belongs to the maintainer, not to #7230's implementer.

Anchors at `a768f81a1` (pre-#7230): `matchMilestone` `audit-writers.ts:355`, the interpolation `:368-375`, its call site `:872`.

[...]

Issue #7289, opened 2026-08-10
plugin-audit: select option labels in `sys_activity.summary` are never localized — the same defect class as #7230, one line over in `displayFieldValue`


os-zhuang:
Found while implementing #7230 (`packages/plugins/plugin-audit/src/audit-writers.ts`). Filed rather than fixed: #7230's scope is recommendation A's two halves (field label + lookup value), and this is a third change to the same function.

## Mechanism

`displayFieldValue` renders a select/picklist value by scanning `field.options[]` and returning the matching option's `label`:

```ts
const ov = o && typeof o === 'object' ? (o.value ?? o.name ?? o.label) : o;
if (ov === value) {
  const ol = o && typeof o === 'object' ? (o.label ?? o.name ?? ov) : o;
  return String(ol);
}
```

`field.options` comes from `engine.getSchema(name)` — the **authored** metadata, which is locale-independent. The shipped translation bundles carry option labels under `objects.[object].fields.[field].options.[value]` (see `packages/plugins/plugin-audit/src/translations/zh-CN.objects.generated.ts`, e.g. `sys_audit_log.fields.action.options.create = "创建"`), and nothing on this path reads them.

Net effect on a zh-CN workspace, after #7230 lands: the field label is localized and the reference value resolves to a title, but a select value still renders its authored-language label —

```
阶段: Proposal → Closed Won
```

— a half-localized string on a fully-localized page. Before #7230 the whole thing was English, so the mixed rendering is new visibility, not a new defect.

## Why it is the same class as #7230, and why it is still separate

Same root shape (a display string composed at write time that skips the locale-bound translator), same function, same blast radius — every `sys_activity.summary` consumer at once: record discussion feed, console home activity, header inbox, the Setup `sys_activity` list, mobile/REST/SDUI. But it is a distinct decision: the fix has to thread `translate` and the **object name** into `displayFieldValue` itself (today it receives neither, and #7230 deliberately kept it synchronous and translator-free so `matchMilestone` could keep calling it unchanged).

## Suggested shape, if taken

In `displayFieldValue`'s option branch, prefer `translate(\`objects.${objectName}.fields.${key}.options.${value}\`)` over `o.label`, with the authored label as the fallback — mirroring exactly what #7230 did for the field label. `matchMilestone`'s call site would need the same two arguments or an explicit opt-out.

Anchors at `a768f81a1` (pre-#7230): `displayFieldValue` `audit-writers.ts:308`, option branch `:310-319`.

[...]

PR #7291, opened 2026-08-10
fix(plugin-audit): localize the tracked-change activity label and render lookup titles instead of raw ids (#7230)

### LINKED ISSUE #7230: plugin-audit: renderTrackedChangeSummary never receives the translator, and displayFieldValue prints raw lookup ids — sys_activity.summary ships `Rating Owner: ∅ → oBK25…` onto every feed surface (re-routed from objectui#4055)
Re-routed from objectstack-ai/objectui#4055 (originally #5142, migrated 2026-08-10 under the #7167 file-at-destination ruling). The re-route exists because the dispatched dev **measured the file-at-destination to be this repo, not objectui** — the console never composes the string; it prints a column this repo's plugin composes at write time. Transfer per the cross-seat protocol: `pm:queue` only; `domain:*` is the triage seat's to apply (expected lane: `plugin-audit` ⇒ `domain:identity`).

## Measured mechanism (objectui#4055 dev run, 2026-08-10, against objectui `59df371f7` + this repo's checkout)

- The template `${label}: ${from} → ${to}` with `∅` for empty lives at `packages/plugins/plugin-audit/src/audit-writers.ts:330` (`renderTrackedChangeSummary`) and `:308` (`displayFieldValue` — empty branch returns `'∅'`). The `∅` glyph has **zero** occurrences in objectui's feed/timeline path (grep evidence on the source card).
- **Label half**: the single call site (`audit-writers.ts:882`) passes `(getFieldDefs(ctx.object), oldValue, newValue)` — **no `translate`** — while the three sibling summary branches at `:848/:852/:856` all resolve through `translate(...)` / `displayLabelFor(ctx.object, translate)`. The tracked-change branch is the one branch never handed the locale-bound translator: a plain oversight against ADR-0053 / #3039's write-time localization.
- **Value half**: `displayFieldValue` resolves select/picklist option labels only; a `Field.lookup(...)` value falls through to `String(value)` — the raw 32-char id.
- objectui side is a pass-through with nothing to fix (`RecordDetailView.tsx:1482` maps `body: row.summary`; `RecordActivityTimeline.tsx:473-477` prints it as text). The record HISTORY tab is a different, already-correct path.
- Blast radius of the producer fix: every consumer of `sys_activity.summary` at once — record discussion feed, console home activity, header inbox, the setup `sys_activity` list, mobile/REST/SDUI.

## Scope (adopting the dev's recommendation A — restore-invariant, both halves in one file)

1. Thread the existing locale-bound `translate` into `renderTrackedChangeSummary` and resolve the field label through it, exactly as the three sibling branches already do.
2. Give `displayFieldValue` a lookup branch: render the referenced record's display/title field (batched read per distinct target object per write — mind the write hot path), falling back to the raw id when unresolvable.
3. The `∅ →` notation stays as-is. Unit tests in plugin-audit for both halves.

## Explicitly OUT of scope (needs its own ruling if ever wanted)

- **Per-viewer locale on historical rows** (the objectui-side re-composition route): it would redefine `sys_activity.summary` from "the user-facing display string" (the contract #6656's masking and ADR-0053 assume) to "a fallback for client recomposition", and it fixes one of four consumer surfaces. If per-viewer locale is judged a requirement, that is a framework-contract change to decide explicitly — not to smuggle into this bug fix.
- Historical rows keep their write-time composition; only new writes improve. This is accepted under this card.

## Release-board note (for the triage seat, not self-applied)

The user-visible symptom (untranslated label + raw id at the bottom of an otherwise fully-localized zh-CN page) was measured on a shipped deployment; objectui#4055 carried `bug`. Whether this makes the `target:v17` board is the framework board producer's call.

Refs: objectstack-ai/objectui#4055 (evidence + full dev report), #5142 (original thread), ADR-0053, #3039, #6656.

os-zhuang:
Fixes #7230

`sys_activity.summary` is composed at **write time** and shipped verbatim to every consumer at once — the record discussion feed, console home activity, the header inbox, the Setup `sys_activity` list, and mobile/REST/SDUI. Its tracked-change branch (ADR-0052 §5b, the `"[label]: [old] → [new]"` template) was producing `Rating Owner: ∅ → oBK25…` at the bottom of an otherwise fully-localized zh-CN page. One string, two independent causes — both fixed in one file, per the card's recommendation A.

## Anchors re-verified at the branch point (`a768f81a1`)

All three still hold exactly where the card recorded them, so nothing had to be re-scoped:

| anchor | card | at `a768f81a1` |
|:--|:--|:--|
| `displayFieldValue` | `:308` | `:308` |
| `renderTrackedChangeSummary` | `:330` | `:330` |
| tracked-change call site, no `translate` | `:882` | `:882` |
| sibling branches resolving through `translate` | `:848` / `:852` / `:856` | `:847` (`translate`) / `:848` (`displayLabelFor`) / `:853` / `:857` |

objectui was not touched (verified pass-through per the card), and no file under `content/docs/releases/` was touched.

## Label half — the branch that was never handed the translator

`renderTrackedChangeSummary` now takes the object name and the same locale-bound `translate` its three siblings (`messages.activityCreated` / `messages.activityDeleted` / `messages.activityUpdated`, plus `displayLabelFor` for the object label) already resolve through, and reads the field label on the bundles' own key shape:

```
translate(`objects.${objectName}.fields.${key}.label`)
  ?? authored field.label
  ?? the machine key
```

That is the shape `zh-CN.objects.generated.ts` and every other shipped bundle actually carries, so the fix is against real data rather than against a convention. A miss still returns `undefined`, so both existing fallbacks answer exactly as before — ADR-0053 / #3039 write-time localization, applied to the one branch that missed it.

## Value half — a reference renders its record's title

`displayFieldValue` gains a reference branch for `lookup` / `master_detail` / `user`. `user` is included because the spec defines it as "a lookup specialized to the `sys_user` system object — stored IDENTICALLY (FK string column)", and it is the exact field class the symptom was reported on. `tree` is excluded: it carries no `reference`, so there is no target object to read.

The title field comes from **ADR-0079's `resolveDisplayField`** (`nameField` → the deprecated `displayNameField` alias → deterministic derivation), imported rather than re-derived. A local "try name, then title, then subject" heuristic here would be a second de-facto contract that disagrees with the record picker, the search companion and the approval inbox the day an author sets `nameField` — the same argument the file's existing `SECRET_MASK` / `collectMaskedReadFields` import makes for itself.

The change is **restore-invariant**: an id with no resolved title — target removed out of band, unregistered object, failing read, or a title field that resolves to `id` — renders exactly as it did before. It can replace an id with a title, never a title with an id. The `∅ →` notation is unchanged, and `matchMilestone`'s `{token}` interpolation passes no resolver, so its behaviour is byte-identical.

## Read cost — measured, because #6977 landed in this file one day ago

#6656 / PR #6977 retired `captureBefore`'s redundant pre-image read from this write path under maintainer ruling Option A+ (2 → 1 reads per single-id write, 3 → 0 per predicate write). A naive per-row lookup resolution would have handed that straight back. Measured with the same copy-returning counting driver, and pinned as cases rather than asserted in prose:

| write | added reads |
|:--|:--|
| create (any) | **0** |
| delete (any) | **0** |
| update moving no tracked reference field, **on a row that holds references** | **0** |
| update moving an **untracked** reference field | **0** |
| update moving 1 tracked reference | **1** on that target object |
| update moving 2 tracked references onto the **same** target | **1** (batched, not 2) |
| update moving references onto 2 **distinct** targets | **1 each** |
| both sides of one reference change (old id + new id) | **1** — one `id: { $in: [...] }`, not two point reads |

`#6977`'s own counts are re-asserted alongside: the single-id update still pays exactly one `findOne` on the audited object, and `audit-bound-previous.test.ts` is untouched and green.

Three properties do the work: the read plan is built from the **diff**, not from the written row; it is keyed by **target object**, not by field or by value; and only `['id', titleField]` is selected, so resolving a title cannot drag a wide row through the write path.

**What batching cannot remove, stated honestly:** `writeAudit` is dispatched per row, so a predicate update over N rows that moves a tracked reference on each pays N reads (one per row per distinct target). That is inherent to per-row summary composition, not to this batching — and unlike the pre-image read #6977 removed, it is gated on "this object declares a tracked reference field and it moved", so the default write pays nothing.

## #6977's masking is not regressed

`resolveDisplayField` honours an explicit `nameField` pointer even when it points at a title-**ineligible** type, so an object could in principle designate a credential field as its title. Rather than trust downstream masking, the read is **skipped** when the resolved title field is in `collectMaskedReadFields(def)` — the same contract predicate `ledgerView` masks with, so "no credential value reaches a user-facing activity summary" holds on this path by the same definition, not a second one. `ledgerView`, the secret mask and the virtual-field drop are otherwise untouched; the summary still renders from the masked views, and the read plan is built from those same masked views.

## Tests

New file `packages/plugins/plugin-audit/src/audit-lookup-summary.test.ts` — 17 cases on a **real `ObjectQL` engine** with the copy-returning counting driver lifted from `audit-bound-previous.test.ts` (its copy-returning rule included deliberately: a driver handing out live store references lets the read path rewrite the store under a measurement).

The two halves are pinned **separately**, because a test asserting only the finished string could be satisfied by fixing either one. The label half is pinned against a **real locale** — `createMemoryI18n` loaded with real `objects.[object].fields.[field].label` data — not a stub translator that would pass whatever key shape the code used.

One case earns a note: the raw-id fallback could not be written the obvious way. The engine's `assertReferencesResolve` **refuses** to write a dangling reference (`ValidationError: Account: no crm_account record has id "acc_missing"`), so the realistic unresolvable case is a reference that was valid when written and whose target has since gone away — reproduced by removing the row out of band.

```
pnpm --filter @objectstack/plugin-audit test       Test Files 10 passed (10)   Tests 160 passed (160)
pnpm --filter @objectstack/plugin-audit typecheck  tsc --noEmit — clean
npx eslint packages/plugins/plugin-audit --no-inline-config  — clean
```

Family gates run locally, all green: `check:nul-bytes`, `check:durability-log-level` (this file is in its vocabulary), `check:engine-double-contract`, `check:error-code-casing`, `check:route-envelope`, `check:wildcard-fallthrough`, `check:empty-changeset`, `check:adr-0087-registration`. `check:i18n` / `check:i18n-coverage` report PREREQUISITE NOT MET locally (they run the built CLI); this diff adds no declared label and edits no translation bundle, so neither can drift from it — CI runs them regardless.

## Reverse verification — direction predicted in writing before running

Predictions were written to a file before the first mutation ran. Four mutations, each reverting one property of the change.

|:--|:--|:--|:--|
| A | drop the `translate(...)` limb from the label | RED: the zh-CN label case. GREEN: the def-label fallback, all value-half and read-count cases | **as predicted** — 2 red (both zh-CN label cases), 158 green. `expected 'Owner: ∅ → 张伟' to be '负责人: ∅ → 张伟'` |
| B | delete the reference branch from `displayFieldValue` | RED: every case asserting a resolved title. GREEN: the raw-id fallback case and both read-count cases | **partly wrong — see below.** 10 red, 150 green |
| C | key the read plan by field instead of by target object | RED: the same-target batching case (measures 2). GREEN: everything else | **as predicted** — 1 red, `expected 2 to be 1` |
| D | resolve from the written row instead of from the diff (the #6977 regression) | RED: the zero-read guard | **GREEN — 160/160. The guard was passing for the wrong reason.** See below |

**Mutation B, the two things I got wrong.** The raw-id fallback case went RED, not green: after the engine refused a dangling reference the case was rewritten to a **mixed** assertion (`Account: acc_2 → Acme Corp`), where the unresolvable side stays raw *and* the resolvable side must still become a title — so it pins both directions and cannot survive the mutation. And three read-count *cases* went red too. The underlying claim still held exactly as stated — **no count assertion failed**, every one of those failures was the trailing `lastSummary` assertion — so read counts really are blind to a missing render branch; my prediction was wrong at the case level only because those cases carry a second, rendering assertion.

**Mutation D is the one that changed the change.** It came back 160/160 green where I predicted red. Diagnosed: the hot-path cases ran against a deal holding **no** references, where a naive resolve-from-the-written-row implementation also reads nothing — there is nothing on the row to read. The guard was green for the wrong reason and would not have caught the exact regression it exists to catch. The fixture was hardened (a second deal seeded with `account_id` and `owner` already set), and mutation D re-run against it:

```
× a tracked change with NO reference field pays ZERO extra reads, on a row that HAS references
× an UNTRACKED reference field changing pays ZERO reads
AssertionError: expected 1 to be +0
```

Both properties now go red on the regression, and the fixture comment records why the seeded references are load-bearing rather than incidental.

## Deliberately not done

- **Per-viewer locale on historical rows** — out of scope per the card, and it would redefine what `sys_activity.summary` *is*. Historical rows keep their write-time composition; only new writes improve.
- **Select option labels are still not localized.** `displayFieldValue` reads `field.options[].label` from the schema, which is authored-language, while the bundles carry `objects.[object].fields.[field].options.[value]`. That is the same defect class as the label half in a neighbouring line, but it is a third change the card does not scope — filed separately as #7289.
- **Milestone `{token}` interpolation still prints raw lookup ids** (`matchMilestone`). Resolving it would require a read *before* knowing whether a milestone fires, i.e. on writes that produce no milestone at all — the opposite of this change's hot-path shape. Filed as #7290.
- **`multiple: true` lookups stored as an unparsed JSON string** fall through to the raw value. In-memory arrays are unpacked and resolved from the same batch; parsing a JSON column here would be this file inventing a storage contract it does not own.

---

### DIFF
--- a/.changeset/audit-tracked-change-summary-locale-and-lookup.md
+++ b/.changeset/audit-tracked-change-summary-locale-and-lookup.md
@@ -1,0 +1,48 @@
+---
+"@objectstack/plugin-audit": patch
+---
+
+fix(plugin-audit): localize the tracked-change activity label and render lookup titles instead of raw ids (#7230)
+
+`sys_activity.summary` is composed at write time and shipped verbatim to every
+feed surface at once — the record discussion feed, console home activity, the
+header inbox, the Setup `sys_activity` list, and mobile/REST/SDUI. Its
+tracked-change branch (ADR-0052 §5b, `"<label>: <old> → <new>"`) was producing
+strings like `Rating Owner: ∅ → oBK25…` at the bottom of an otherwise
+fully-localized page. Two independent causes, both fixed here:
+
+- **The label was never localized.** `renderTrackedChangeSummary` was the one
+  summary branch never handed the locale-bound `translate` its three siblings
+  (`messages.activityCreated` / `messages.activityDeleted` /
+  `messages.activityUpdated`, plus the object label via `displayLabelFor`) all
+  resolve through — an oversight against ADR-0053 / #3039 write-time
+  localization. The field label now resolves through the same translator, on the
+  bundles' own key shape (`objects.<object>.fields.<field>.label`), and falls
+  back to the authored `label`, then the machine key, exactly as before.
+- **A reference value printed its raw id.** `displayFieldValue` resolved
+  select/picklist option labels only, so a `lookup` / `master_detail` / `user`
+  value fell through to `String(value)` — the stored 32-char id. It now renders
+  the referenced record's title, resolved through ADR-0079's
+  `resolveDisplayField` (`nameField` → deprecated `displayNameField` alias →
+  derivation) rather than a local name-guessing heuristic.
+
+The `∅ →` notation is unchanged, and so is every other summary branch. The
+change is restore-invariant: an id that cannot be resolved — a target removed
+out of band, an unregistered object, a failing read — renders exactly as it did
+before.
+
+**Read cost, measured with a counting driver** (the same technique that measured
+#6656 / PR #6977's retirement of the redundant pre-image read from this write
+path, and pinned as cases in `audit-lookup-summary.test.ts`):
+
+- **0 added reads** on every create, every delete, every update that moves no
+  tracked reference field, and every update that moves an *untracked* one —
+  including on rows that do carry references. #6977's counts (1 `findOne` per
+  single-id write, 0 per predicate write) are untouched.
+- **1 read per distinct target object** on an update that does move a tracked
+  reference: both sides of the change are answered by a single
+  `id: { $in: [...] }` selecting only the id and title columns, however many
+  tracked reference fields point at that object.
+
+Historical `sys_activity` rows keep their original write-time composition — only
+new writes improve.
--- a/packages/plugins/plugin-audit/src/audit-lookup-summary.test.ts
+++ b/packages/plugins/plugin-audit/src/audit-lookup-summary.test.ts
@@ -1,0 +1,522 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#7230] `sys_activity.summary`'s tracked-change branch — the field LABEL is
+ * localized, and a reference VALUE renders the referenced record's title
+ * instead of its raw id.
+ *
+ * ## The two halves, pinned separately on purpose
+ *
+ * The measured symptom was one string — `Rating Owner: ∅ → oBK25…` at the
+ * bottom of an otherwise fully-localized zh-CN page — but it had two
+ * independent causes, and a test that only asserted the finished string could
+ * be satisfied by fixing either one:
+ *
+ *  - **Label half.** `renderTrackedChangeSummary` was the ONE summary branch
+ *    never handed the locale-bound `translate` its three siblings
+ *    (`messages.activityCreated` / `Deleted` / `Updated`, plus `displayLabelFor`)
+ *    all resolve through — an oversight against ADR-0053 / framework#3039
+ *    write-time localization. Pinned against a REAL locale: `createMemoryI18n`
+ *    loaded with real `objects.<object>.fields.<field>.label` data, which is
+ *    the key shape the shipped bundles use. A stub translator returning a
+ *    marker would pass whether or not the real key shape were right.
+ *  - **Value half.** `displayFieldValue` resolved select/picklist option labels
+ *    only, so a `lookup` / `master_detail` / `user` value fell through to
+ *    `String(value)` — the raw 32-char id.
+ *
+ * ## Why the read counts are measured here at all
+ *
+ * The value half needs a read, and this file is the write hot path: one day
+ * before this change #6656 / PR #6977 retired `captureBefore`'s redundant
+ * pre-image read from it (2 → 1 per single-id write, 3 → 0 per predicate
+ * write) under maintainer ruling Option A+. So the reads this change ADDS are
+ * measured with the same copy-returning counting driver that measured that
+ * retirement, and the three properties that keep it off the hot path are
+ * pinned as cases rather than asserted in prose:
+ *
+ *   0 reads   when no tracked REFERENCE field changed (every create, every
+ *             delete, and every update that does not move a reference)
+ *   1 read    per distinct target object, however many tracked reference
+ *             fields point at it
+ *   N reads   for N distinct target objects — one each, never one per value
+ *
+ * ⚠️ The read-count cases do NOT pin the rendering, and the rendering cases do
+ * not pin the read counts: the read lives in `resolveTrackedLookupTitles` and
+ * the rendering in `displayFieldValue`. Deleting the render branch leaves every
+ * count green. Both sets are therefore required, and each was mutated
+ * separately to confirm it goes red on its own defect.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { ObjectQL } from '@objectstack/objectql';
+import { createMemoryI18n } from '@objectstack/core';
+import { installAuditWriters } from './audit-writers.js';
+
+// ---------------------------------------------------------------------------
+// Fixtures
+// ---------------------------------------------------------------------------
+
+const f = (name: string, type: string, extra: Record<string, unknown> = {}) =>
+  ({ name, label: name, type, ...extra }) as any;
+
+const sysAuditLog = {
+  name: 'sys_audit_log', label: 'Audit Log',
+  fields: {
+    id: f('id', 'text', { primaryKey: true }), action: f('action', 'text'),
+    user_id: f('user_id', 'text'), object_name: f('object_name', 'text'),
+    record_id: f('record_id', 'text'), old_value: f('old_value', 'textarea'),
+    new_value: f('new_value', 'textarea'), tenant_id: f('tenant_id', 'text'),
+  },
+};
+
+const sysActivity = {
+  name: 'sys_activity', label: 'Activity',
+  fields: {
+    id: f('id', 'text', { primaryKey: true }), type: f('type', 'text'),
+    timestamp: f('timestamp', 'datetime'), summary: f('summary', 'text'),
+    actor_id: f('actor_id', 'text'), object_name: f('object_name', 'text'),
+    record_id: f('record_id', 'text'), record_label: f('record_label', 'text'),
+    metadata: f('metadata', 'textarea'),
+  },
+};
+
+/** Target objects. Their title fields are DERIVED (ADR-0079), not declared. */
+const crmAccount = {
+  name: 'crm_account', label: 'Account',
+  fields: { id: f('id', 'text', { primaryKey: true }), name: f('name', 'text') },
+};
+
+const crmRegion = {
+  name: 'crm_region', label: 'Region',
+  fields: { id: f('id', 'text', { primaryKey: true }), title: f('title', 'text') },
+};
+
+const sysUserObj = {
+  name: 'sys_user', label: 'User',
+  fields: {
+    id: f('id', 'text', { primaryKey: true }), name: f('name', 'text'),
+    email: f('email', 'email'),
+  },
+};
+
+/**
+ * The audited object. Two tracked lookups point at the SAME target
+ * (`account_id`, `partner_id`) so batching is measurable; `region_id` gives a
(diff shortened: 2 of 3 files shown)


os-zhuang:

CI on head `fb33dad85`: **25 runs, none non-green.** 3 files, +790 −8. Boundaries: **objectui untouched**, nothing under `content/docs/releases/`, and — the one this seat flagged at dispatch — **#6977's work is intact**: zero deletions of `ledgerView` / `SECRET_MASK` / `VIRTUAL_FIELD_TYPES`, 19 occurrences still present.

## The question this dispatch was built around, answered with numbers

The dispatch warned that #6656 / PR #6977 (ruling **A+**) had retired the redundant pre-image read from **this exact file** the day before — 2→1 per single-id write, 3→0 per predicate write — and that a naive per-row lookup resolution would hand those reads straight back. The answer:

- **0 added reads** on every create, every delete, every update that moves **no** tracked reference field (measured on a row that *does* carry references — the right control), and every update moving an untracked reference field. #6977's counts are re-asserted in the new suite rather than assumed.
- When a tracked reference actually moves: **exactly 1 read per DISTINCT target object**. Two tracked lookups onto the same object cost **1, not 2**; old id and new id are answered by the **same** `id: { $in: [...] }` projecting only `['id', titleField]`.
- Measured with the **copy-returning** counting driver lifted from `audit-bound-previous.test.ts` — the one that does not let the engine's read path mutate the store and make the measurement lie.

And the limit is **stated rather than buried**: `writeAudit` is dispatched per row, so a predicate update over N rows each moving a tracked reference pays N reads. That is inherent to per-row summary composition, and — unlike the pre-image read #6977 removed — it is **gated on "this object declares a tracked reference field and it moved"**. Correct call, honestly bounded.

## Mutation D is the best thing in this report

Predicted RED (resolve from the written row instead of from the diff — the #6977 regression shape). **First run came back 160/160 GREEN.** Rather than banking the green, the author diagnosed *why*: the hot-path cases ran against a deal holding **no references**, where a naive resolve-from-the-written-row implementation also reads nothing. **The guard was green for the wrong reason and would not have caught the regression it exists to catch.** Fixture hardened with a second deal carrying `account_id` and `owner`, mutation re-run, both properties now go red — and the fixture comment records why those seeded references are load-bearing.

That is the "passes for the empty reason" failure mode this lane has been chasing all session (#6964's vacuous D7 denials, #7151's decorative-pin refusal), caught by an author **in their own test, against their own prediction**. It is also why the read-count claim above can be trusted: the guard behind it demonstrably bites.

Two prediction misses reported rather than smoothed: under mutation B the raw-id fallback case went red because it had been rewritten into a **mixed** assertion (`Account: acc_2 → Acme Corp`, unresolvable side stays raw *and* resolvable side must become a title — pinning both directions), and three read-count *cases* went red while **no count assertion failed** — every failure was the trailing summary assertion, so the underlying claim held exactly as stated and only the case-level prediction was wrong. The distinction between "my claim was wrong" and "my bookkeeping about which case carries the claim was wrong" is drawn correctly.

## Other judgement calls this seat agrees with

- **The label half is pinned against a REAL locale** (`createMemoryI18n` loaded with real `objects.[object].fields.[field].label` data), not a stub that would pass on any key shape. The dispatch asked for this specifically; `'负责人: ∅ → 张伟'` is the measured output, quoted verbatim in the PR body — quoting a measurement is not a language violation, translating it would falsify it.
- **`resolveDisplayField` imported, not re-derived** (ADR-0079), so the title rule has one owner.
- **Restore-invariant**: the change can replace an id with a title, never a title with an id. `matchMilestone` passes no resolver, so its behaviour is byte-identical.
- The raw-id fallback could not be authored the obvious way because the engine's `assertReferencesResolve` **refuses** a dangling reference — so it is reproduced as a reference valid at write time whose target later vanished. Finding that the obvious fixture is unauthorable, and saying so, beats a fixture that quietly tests nothing.
- **`multiple: true` lookups stored as an unparsed JSON string left alone**, because parsing a JSON column here would be this file inventing a storage contract it does not own. In-memory arrays *are* unpacked from the same batch.

## Follow-ups filed, unassigned, for triage

**#7289** — select option labels in the same function are still unlocalized (`displayFieldValue` reads authored `field.options[].label` while bundles carry `objects.[object].fields.[field].options.[value]`), so after this PR a zh-CN feed shows a localized label beside an authored-language value. Same defect class, one line over, deliberately not scoped in. **#7290** — `activityMilestones` templates still interpolate raw lookup ids, and the milestone branch **takes precedence** over the branch this PR fixed, so on milestone-declaring objects it is the string users actually see; the body costs out both read placements and recommends the after-it-fires one.

---

Issue #7145, opened 2026-08-09
service-storage's attachment kit carries the same 5-field `callerContext` projection #7141 fixed for comments


os-zhuang:
Out-of-scope observation from implementing #7141 (PR #7143), filed rather than fixed — that card is bounded to `packages/plugins/plugin-audit`.

## Fact

`packages/services/service-storage/src/attachment-access-hooks.ts:70` —
`callerContext()` rebuilds the identical five-field projection of the caller's
execution envelope:

```
function callerContext(ctx: any): Record<string, unknown> {
  const exec = ctx?.input?.options?.context;
  if (exec && typeof exec === 'object') {
    return {
      userId: exec.userId,
      tenantId: exec.tenantId,
      positions: exec.positions,
      permissions: exec.permissions,
      isSystem: exec.isSystem,
    };
  }
  ...
}
```

The result is handed to `sharing.canEdit(parentObject, parentId, …)` from two
sites (the `beforeInsert` parent gate and the multi-row `authorizeRows` loop).
This is the kit `comment-access-hooks.ts` was derived from, so the defect is the
same one, one package over:

- **`onBehalfOf`** is dropped, so `ISecurityService.hasWriteBypass` — documented
  to fail CLOSED on a delegated context, implemented by reading exactly
  `context?.onBehalfOf?.userId` (`plugin-security/src/security-plugin.ts:704-707`)
  — cannot fire on the attachment path either. The field is populated on a live
  path (`runtime/src/security/resolve-execution-context.ts:191`, OAuth agent
  principals on `/mcp`).
- **`principalKind`** is dropped, so `resolvePermissionSetsForContext`
  (`security-plugin.ts:3024-3034`) resolves an ADR-0090 D10 agent principal **as
  a human** and appends the deployment's additive baseline to the agent's
  scope-derived ceiling.
- `systemPermissions`, `accessible_org_ids`, `posture`, `audience` and
  `rlsMembership` go with them.

The **middleware-key stripping half of the projection is correct and must
survive** any fix here too — plugin-security mutates the operation context in
place (`sc.__readScope = …`, `security-plugin.ts:1159`), so the arriving context
carries the depth resolved for `sys_attachment` while the gate asks about the
parent record's object.

## Suggested shape

The same one #7141 landed: forward the envelope minus the operation-private keys,
dropped by the `__` **prefix** (which also covers `__expandRead`, the marker that
waives the object-level CRUD check, and `__referentialFieldClear`). See PR #7143
[...]

Issue #7281, opened 2026-08-10
checkAuthoredRowWrite abstains on every `private`-OWD cross-owner row, so #5493's by-id widener deferral is inert on the posture it was filed for — and its unit test cannot see it (fake engine bypasses middleware)


os-zhuang:
Measured while running the #6736 bulk-path probe (PR #7274). Filed unassigned, out of that PR's scope, for triage.

## What was measured

`ISecurityService.checkAuthoredRowWrite` resolves its verdict by re-reading the target row through the **caller's own** execution context:

```
const parts = [{ id: recordId }, ...(layer0 ? [layer0] : []), layer1];
const row = await this.ql.findOne(object, { where: { $and: parts }, context });
return row ? 'admit' : 'abstain';
```

`packages/plugins/plugin-security/src/security-plugin.ts`, in `checkAuthoredRowWrite`.

That `findOne` re-enters the middleware chain, so `plugin-sharing`'s **read** filter applies. `buildReadFilter` scopes a `private`-OWD object to owner-match OR shares — so a cross-owner row is invisible to the caller, `findOne` returns null, and the verdict is `abstain` rather than `admit`. The middleware's refusal then stands and the write is refused 403, even though the app declared a widener that admits the row by name.

Measured end-to-end on the real stack (`bootStack` + real SecurityPlugin + real SharingServicePlugin + real ObjectQL engine). Two objects, identical in every respect except the OWD, same app-authored widener (`operation: 'update'`, `using: "stage == 'open'"`), same principal, same cross-owner row shape:

| object | OWD | `checkAuthoredRowWrite` on the cross-owner admitted row | by-id PATCH |
|---|---|---|---|
| `probe_note` | `public_read` | `admit` | **200**, row changes |
| `probe_secret` | `private` | **`abstain`** | **403**, row unchanged |

Control ruling out an inert or mis-parsed widener: on the **same `private` object**, `checkAuthoredRowWrite` for the caller's **own** open row answers `admit`. The declaration is live there; it is the cross-owner read that removes it.

## Why this matters

`private` is the posture the widener surface was built for. #5493's own measured defect is described as occurring exactly where record sharing enforces — *"a `private`/`public_read` OWD and an `owner_id` field"* — and the sharing middleware's deferral comment says the probe exists so that gate does not refuse a row an app-authored widener admits. On `private`, it still refuses: the deferral is asked, and structurally cannot answer `admit` for any row the caller cannot already read.

[...]

PR #7305, opened 2026-08-10
docs(spec,plugin-audit): record the sys_comment parent gate's `own` depth as ruled, with the fail-open that blocks widening it (#7144)

### LINKED ISSUE #7144: Decision needed: should the `sys_comment` parent gates run the parent's owner-match at the caller's real write DEPTH, or stay at `own`?
Split out of #7141 (PR #7143) rather than decided there, because it is a **widening** and the two candidate shapes that card named diverge here — on nothing else.

## The fact

`ISharingService.canEdit` widens its owner-match by the access DEPTH the caller
holds on the probed object, read from the middleware-private `__writeScope` key
(`plugin-sharing/src/sharing-service.ts` — `matchesOwnerScope`, and
`if (writeScope === 'org') return true`). plugin-security supplies that key
explicitly for the object it is probing, via the private
`resolveWriteScopeForSharing` (`security-plugin.ts:2516`).

The `sys_comment` gates in `plugin-audit` supply **no** depth. The five-field
projection #7141 removed never carried one, and PR #7143 deliberately kept it
that way: an absent depth leaves the owner-match at its narrowest (`own`), which
is the safe direction and is byte-for-byte what the projection produced.

## The consequence, stated as behaviour

A caller whose write depth on the PARENT object is `unit` / `unit_and_below` /
`org` can edit that parent record directly through the CRUD path (where the
middleware stamps the depth for that object) but is **refused** when they try to
edit or delete a comment on it, because this gate asks the same service the same
question with the depth omitted. It is a divergence in the *restrictive*
direction — no data leaks — between the comment gate and the parent's real edit
authority. Whether that is correct or is a bug depends on what the comment gate
is meant to mean, which is a product decision, not a code reading.

## Why it was not simply fixed in #7143

The tool a package outside `plugin-security` has for this is the published
`ISecurityService.resolveWriteScope`. It fails **open** on one input:
`getEffectiveScope` returns `'org'` when no permission set matches the object
(`permission-evaluator.ts:255`), and the service's own doc block flags that
`'org'` as "non-authoritative on its own". Passed to `canEdit` as `__writeScope`
it becomes fully authoritative, and `matchesOwnerScope` then returns `true` for
**any owned row** of an unmatched object — a real widening, not a theoretical
one. Taking that route needs either a narrower contract method (one that
distinguishes "org depth" from "nothing matched") or the gate resolving depth
some other way.

Note this is the same question for `service-storage`'s attachment kit, which
gates on `canEdit` the same way.

## What a decision would look like

1. **Keep `own`** (today, and #7143's behaviour) — the comment gate is
   deliberately tighter than the parent's edit authority. Then say so in the
   contract's doc block, so the next reader does not "fix" it.
2. **Resolve the parent's depth** — the comment gate inherits the parent's real
   edit authority. Needs a depth primitive that does not fail open on an
   unmatched object before it can be wired.

Related: #7141, PR #7143, #6523, #6206, ADR-0057 D1, ADR-0111 D1.

os-zhuang:
Fixes #7144.

Implements the maintainer's **Option 1** ruling of 2026-08-10: the `sys_comment`
parent gates **stay at `own`**, deliberately tighter than the parent's edit
authority, and that fact — with the reason the alternative is blocked — is
recorded in the contract's doc block so the next reader does not "fix" it into a
widening.

**This is a documentation change. It changes no behaviour**, and that is checked
mechanically rather than asserted: every added and removed line in the diff is a
doc-block line.

```
$ git diff -U0 | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | sed -E 's/^[+-]//' \
    | grep -vE '^\s*(\*|/\*\*|\*/)' | grep -vE '^\s*$'
(no output)
```

## Where the statement landed, and why there

**Primary — `packages/spec/src/contracts/sharing-service.ts`, a new section on
the `ISharingService` interface doc block** (`## Write DEPTH is an input the
CALLER supplies — absent means own`).

The reader who is about to "fix" this is standing at a gate call site reading
"ownership (widened by write DEPTH)" and finding no depth supplied. That
sentence is the contract's, and it appears on four methods of this one
interface; the contract never said where the depth comes from or what an absent
depth means. It is also the **shared** authority: `plugin-audit`'s comment kit
and `service-storage`'s attachment kit ask the *same* method the *same* question
about a parent record, and neither reads the other's source. Saying it once here
covers both, which is what the ruling asks for.

One honest measurement about that sharing, because it qualifies the claim:
`plugin-audit` reaches this contract **by type** (`comment-access-hooks.ts:95`
picks `canEdit` off `ISharingService`), while `service-storage`
(`attachment-access-hooks.ts:46`) declares a **structural** `canEdit` signature
of its own and does not import the contract. So the contract is the shared
authority for the *meaning* of the gate, not a type-level chokepoint for both
kits. That is still the only place the meaning is written down, so it is still
the right single home — but #7145's seat should be told the doc block will not
arrive at their call site through the type system.

Two supporting pointers, deliberately pointers and not restatements:

1. **`ISharingService.canEdit`'s own doc block** (same file, 3 lines). IDE
   go-to-definition lands on the method, not on the interface header ~60 lines
   above, and `canEdit` is the single method both parent-record kits call. It
   states the one-line rule ("DEPTH is an input this service is handed, never
   one it resolves") and points at the section. `checkEdit` / `canDelete` /
   `checkDelete` carry the same DEPTH phrase and were left alone: no parent-gate
   caller reaches them, and a pointer on all four would be the scattering this
   card is trying to avoid.
2. **`plugin-audit/src/comment-access-hooks.ts` — one paragraph replaced.** Not
   a second copy of the rationale: the note there ended with *"Resolving the
   parent's own depth (the other candidate shape) would WIDEN this gate and is a
   separate decision; see #7141's PR discussion."* After the ruling that
   sentence is **false**, and it reads as an open invitation — exactly the
   re-litigation the ruling ends. It now records that the shape is ruled and
   defers the reasoning to the contract.

## The WHY the doc block carries (the deliverable)

Not "this is intentional", but the fail-open that makes the alternative unsafe
**today**: `ISecurityService.resolveWriteScope` is the only tool a package
outside `plugin-security` has for the parent's depth, and `getEffectiveScope`
returns `'org'` when **no permission set mentions the object at all** —
indistinguishable from a genuine `modifyAllRecords` holder. Handed to a write
gate as the depth it becomes authoritative on its own, and the owner-match
short-exits `true` for every owned row of that object. So inheriting the
parent's depth starts with a depth primitive that can tell "org depth" from
"nothing matched", never with these gates.

## Anchors re-verified at the branch point (`origin/main` @ `397e731a5`)

| Claim | Where | Status |
|---|---|---|
| `matchesOwnerScope` reads the middleware-private depth | `plugin-sharing/src/sharing-service.ts:414`, `:428` | confirmed, **as cited** |
| `if (writeScope === 'org') return true;` | `sharing-service.ts:429` | confirmed |
| read-side depth read | `sharing-service.ts:376-378` | confirmed, **as cited** |
| absent depth ⇒ owner-only | `sharing-service.ts:1187` — `if (!scope \|\| scope === 'own' \|\| scope === 'org') return [me];` | confirmed |
| the fail-open | `plugin-security/src/permission-evaluator.ts:255` — `if (!matched) return 'org';` | confirmed |
| the published tool propagates it | `security-plugin.ts:726-732` — `resolveWriteScope` returns `getEffectiveScope('write', …)` verbatim | confirmed |
| the contract already flags the double meaning | `spec/src/contracts/security-service.ts:342-345` | confirmed |
| `resolveWriteScopeForSharing` still private | `security-plugin.ts:2530` | confirmed (the card said `:2516`; **drifted**, as triage predicted) |
| comment gate supplies no depth | `comment-access-hooks.ts` — `withoutOperationPrivateKeys` strips the `__` prefix | confirmed |

Only one anchor moved: `resolveWriteScopeForSharing`, `:2516` in the card body →
`:2530` here (triage had already caught it). The three `sharing-service.ts`
anchors the card named are exactly where it said.

## Evidence for the cited fail-open — measured, not asserted

A throwaway probe against `PermissionEvaluator` (not committed; it pins a
fail-open as *current* behaviour, which is a bad thing to freeze in a test — see
below):

```
A no permission set matches the object -> "org"
B genuine modifyAllRecords holder     -> "org"
C matched, writeScope: own            -> "own"
A === B (indistinguishable)?           true
```

C is the control: the evaluator *does* discriminate when a set matches, so `org`
in A is the no-match default and not a degenerate return. Paired with
`sharing-service.ts:429` (`writeScope === 'org'` ⇒ `true` for any row whose
owner is non-null), that is the widening the doc block names, end to end.

## Reverse verification — the honest form for a doc-only change

**No test can distinguish this change**, and no red/green table is manufactured
to suggest otherwise: the compiler erases comments, so re-running either suite
with the doc block removed is not an experiment. The mechanical
comment-only check at the top of this PR is the evidence that this is the case,
rather than a claim that it is.

What the ruled *behaviour* rests on is already pinned, by #7143 rather than by
this PR: `comment-access-hooks.test.ts` — *"does NOT carry sys_comment's access
DEPTH into the parent's owner-match"* — asserts both the refusal envelope
(`code: 'RECORD_NOT_ACCESSIBLE'`, `status: 403`) and
`__writeScope` being `undefined` on the call the gate makes. **That pin can go
red**: wiring a resolved depth in would define the key and flip the verdict.

**No new guard added, and the reason is worth stating.** The one guard this card
could have bought is a pin on the fail-open itself (`getEffectiveScope` no-match
⇒ `'org'`). It would go red exactly when someone **fixes** the fail-open — which
is the precondition the ruling names for revisiting Option 2 — so it would fight
the improvement it is supposed to be waiting for. A pin that misdirects the next
reader is worse than none.

## Gates (local, per gate)

| Gate | Result |
|---|---|
| `pnpm --filter @objectstack/spec typecheck` | **pass** (`tsc --noEmit` + scripts + test-typecheck: "OK — @objectstack/spec's test layer compiles") |
| `pnpm --filter @objectstack/plugin-audit typecheck` | **pass** (`tsc --noEmit`, Done) |
| `eslint` (both touched files, `--no-inline-config`) | **pass** (exit 0, no output) |
| `pnpm --filter @objectstack/plugin-audit test` | **9 files / 143 tests passed** |
| `pnpm --filter @objectstack/spec test` | **361 files / 9424 tests passed** |
| `node scripts/check-nul-bytes.mjs` | **OK** (6648 files, no raw control bytes) |

## Deliberately not done

- **No Option 2 work, and nothing that "prepares" for it** — no
  `resolveWriteScope` wiring into `canEdit`, no new depth primitive.
- **`packages/services/service-storage` untouched** — the ruling covers its
  attachment kit, but that is #7145's seat. The contract statement names the
  attachment kit so the meaning is on record; the call-site note there is left
  to its own card. (See the type-vs-structural measurement above.)
- **No `content/docs/releases/` edit** — the changeset is this PR's input.

---

### DIFF
--- a/.changeset/comment-gate-own-depth-contract.md
+++ b/.changeset/comment-gate-own-depth-contract.md
@@ -1,0 +1,30 @@
+---
+"@objectstack/spec": patch
+"@objectstack/plugin-audit": patch
+---
+
+docs(spec,plugin-audit): record that the parent-record write gates match ownership at `own` BY DESIGN (#7144)
+
+Documentation only — no gate changes what it returns for any input.
+
+`ISharingService`'s write gates widen ownership "by write DEPTH", but that depth
+is an INPUT the caller supplies: the CRUD middleware resolves it for the object
+of the operation in flight and stamps it on the operation context. The
+`sys_comment` gates (`@objectstack/plugin-audit`) and the `sys_attachment` kit
+(`@objectstack/service-storage`) ask this service about the PARENT record's
+object, so the stamped depth belongs to a different object and is dropped — and
+the owner-match runs at its narrowest, `own`. A caller whose write depth on the
+parent is `unit` / `unit_and_below` / `org` can therefore edit that parent
+directly and is refused when editing a comment or attachment on it.
+
+That divergence is deliberate and runs in the restrictive direction (refusals,
+never a leak). The contract now says so, and — the part that matters for anyone
+tempted to "fix" it — says WHY the alternative is not merely unimplemented:
+`ISecurityService.resolveWriteScope`, the only tool a package outside
+`plugin-security` has for the parent's depth, fails OPEN, because
+`getEffectiveScope` returns `'org'` when no permission set mentions the object
+at all — indistinguishable from a genuine `modifyAllRecords` holder. Handed to a
+write gate as the depth it becomes authoritative on its own and the owner-match
+short-exits `true` for every owned row of that object. Inheriting the parent's
+edit authority starts with a depth primitive that can tell "org depth" from
+"nothing matched", not with these gates.
--- a/packages/plugins/plugin-audit/src/comment-access-hooks.ts
+++ b/packages/plugins/plugin-audit/src/comment-access-hooks.ts
@@ -258,8 +258,16 @@ function withoutOperationPrivateKeys(exec: Record<string, unknown>): ExecutionCo
  * Note what deliberately did NOT change: no access DEPTH is synthesised for the
  * parent object. Absent depth leaves the sharing owner-match at its narrowest
  * (`own`) — the safe direction, and byte-for-byte the behaviour the projection
- * produced. Resolving the parent's own depth (the other candidate shape) would
- * WIDEN this gate and is a separate decision; see #7141's PR discussion. */
+ * produced.
+ *
+ * [#7144] That is now the RULED shape, not a pending question: the maintainer's
+ * ruling of 2026-08-10 keeps these gates at `own`, deliberately tighter than
+ * the parent's real edit authority. The reasoning — including the fail-open in
+ * `ISecurityService.resolveWriteScope` that makes the widening unsafe to wire
+ * today — is recorded once on the contract that owns this gate's meaning
+ * (`@objectstack/spec` — `ISharingService`, "Write DEPTH is an input the CALLER
+ * supplies"), because the `sys_attachment` kit reaches the same gate from
+ * another package; it is not restated here. */
 function callerContext(ctx: any): ExecutionContext {
   const exec = ctx?.input?.options?.context;
   if (exec && typeof exec === 'object') {
--- a/packages/spec/src/contracts/sharing-service.ts
+++ b/packages/spec/src/contracts/sharing-service.ts
@@ -257,6 +257,46 @@ export type SharingWriteVerdict = 'allow' | 'abstain' | 'deny';
  * `org_user_ids`, `systemPermissions`, `posture` (ADR-0095 D2) and
  * `tabPermissions` included. Which of those a deployment makes load-bearing
  * depends on its tenancy posture, which the caller cannot know.
+ *
+ * ## Write DEPTH is an input the CALLER supplies — absent means `own` (#7144)
+ *
+ * The write gates below widen ownership "by write DEPTH" (ADR-0057 D1: `own` /
+ * `own_and_reports` / `unit` / `unit_and_below` / `org`). That depth is NOT a
+ * field of {@link ExecutionContext}, and this service never derives it: the
+ * CRUD middleware resolves it for the object of the operation IN FLIGHT and
+ * stamps it on the operation context as an operation-private (`__`-prefixed)
+ * key, which the default implementation's owner-match reads. A caller that is
+ * not that middleware hands over no depth, and the owner-match then runs at its
+ * NARROWEST — owner-only, i.e. `own`.
+ *
+ * **For the PARENT-record gates that is deliberate, not an oversight.** The
+ * `sys_comment` gates (`@objectstack/plugin-audit`) and the `sys_attachment`
+ * kit (`@objectstack/service-storage`) ask this service about the PARENT
+ * record's object while the operation in flight is on the CHILD, so the stamped
+ * depth was resolved for a different object and is dropped rather than carried
+ * across. Stated as behaviour: a caller whose write depth on the parent object
+ * is `unit` / `unit_and_below` / `org` can edit that parent directly through
+ * the CRUD path, and is REFUSED when editing or deleting a comment or
+ * attachment on it. The divergence runs in the RESTRICTIVE direction —
+ * refusals, never a leak — and the maintainer ruling of 2026-08-10 (#7144,
+ * split from #7141 / PR #7143) is that these gates STAY at `own`.
+ *
+ * **Why the obvious "fix" is off the table** — recorded here so the next reader
+ * can tell "deliberately tighter" from "nobody got round to it". Making these
+ * gates inherit the parent's real edit authority means resolving the parent's
+ * depth through the only tool a package outside `plugin-security` has,
+ * `ISecurityService.resolveWriteScope` — and that fails OPEN on one input:
+ * `getEffectiveScope` returns `'org'` when NO permission set mentions the
+ * object at all (`plugin-security/src/permission-evaluator.ts` —
+ * `if (!matched) return 'org'`), which is byte-identical to what a genuine
+ * `modifyAllRecords` holder gets. That is why the method's own doc block calls
+ * `'org'` authoritative ONLY when paired with an explicit
+ * `ISecurityService.hasWriteBypass` check. Handed to a write gate as the depth
+ * it becomes fully authoritative on its own, and the owner-match short-exits
+ * `true` for EVERY owned row of an unmatched object — a real widening, not a
+ * theoretical one. So inheriting the parent's depth starts with a depth
+ * primitive that can tell "org depth" from "nothing matched", never with these
+ * gates.
  */
 export interface ISharingService {
   /**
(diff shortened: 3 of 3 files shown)


os-zhuang:

CI on head `fcc345b25`: **26 runs, none non-green.** 3 files, +87 −2. `service-storage` untouched (#7145 is another seat's card).

## The zero-behaviour-change claim was verified, not accepted

The report offers a mechanical proof — take `git diff -U0`, drop every comment line, and the remaining +/- set is empty. **This seat re-ran it independently and it reproduces: empty.** That is the right kind of evidence for a doc-only card, because it is checkable rather than assertable, and it makes "I changed no behaviour" a measurement instead of a promise.

## The placement decision is the substance here, and it was reasoned rather than defaulted

The doc block landed on the **`ISharingService` interface-level block in `packages/spec/src/contracts/sharing-service.ts`**, on the argument that the reader about to "fix" this is standing at a gate call site, reading the contract's own phrase *"ownership (widened by write DEPTH)"* and finding that nobody passes a depth — while the contract never said where depth comes from or what its absence means. That is the gap the re-litigation grows in.

Three secondary touches, each justified as a **pointer rather than a restatement**:
- a 3-line pointer on **`canEdit`** specifically, because IDE go-to-definition lands on the method, not 60 lines above it — and `canEdit` is the *only* method both parent gates call. `checkEdit` / `canDelete` / `checkDelete` carry the same DEPTH wording and were **deliberately left alone**: no parent gate calls them, and touching all four would be the scattering this card warned against.
- in `comment-access-hooks.ts`, the old sentence — *"Resolving the parent's own depth … is a separate decision; see #7141's PR discussion"* — was **replaced, not supplemented**. After the ruling that sentence is false, and it reads as an open invitation to exactly the re-litigation the ruling ends.

## The "why" is grounded in a measurement with a control

The dispatch insisted the block name the fail-open rather than say "this is intentional". It does, and the fail-open was **probed, not cited**: A (no permission set mentions the object) → `'org'`; B (principal genuinely holds `modifyAllRecords`) → `'org'`; **A === B**; and **C, the control** (matched set with `writeScope: 'own'`) → `'own'`, proving the evaluator *does* discriminate when it matches, so A's `'org'` is a no-match default rather than a degenerate return. Chained file:line from `permission-evaluator.ts:255` through the published `resolveWriteScope` to `sharing-service.ts:429`, where feeding it in would admit **every owned row of that object** — a real widening. Card anchors held verbatim; the one drift (`security-plugin.ts:2516 → :2530`) matches what triage had already caught.

## Declining to add a guard, for a reason worth keeping

The only new pin available would pin the fail-open itself — and **it would go red the day someone fixes that fail-open**, which is precisely the precondition the ruling names for reopening Option 2. A pin that fights the improvement it is waiting for is worse than none. Recorded rather than quietly skipped, and this seat agrees.

The behaviour the ruling depends on is already pinned anyway, by #7143: `comment-access-hooks.test.ts`'s *"does NOT carry `sys_comment`'s access DEPTH into the parent's owner-match"* asserts both the refusal envelope and `__writeScope === undefined` — and it **can** go red, because wiring a resolved depth in gives that key a value.

## ⚠️ Relay for the #7145 seat (`domain:services`) — this will not reach you through types

The author measured a limit on its own placement and stated it rather than letting "shared authority" stand unqualified: **`plugin-audit` is typed against the contract** (`comment-access-hooks.ts:95` takes `canEdit` off `ISharingService`), but **`service-storage` declares its own structural `canEdit` signature and does not import the contract** (`attachment-access-hooks.ts:46`). So the contract block is the shared authority for the **semantics**, not a type-level chokepoint for both kits — the attachment kit's seat will not receive it via the compiler. Relaying it here, as the report asked.

**Tooling, corrected**: the report shows the attribution block is **not stripped from PR bodies** — it is *rewritten*, with a session id appended to the link, the rest byte-identical. This seat's own earlier measurement stands for a different surface: on **issue bodies** (#6947, #7082, #7281) the block is genuinely absent on read-back. So the accurate statement is **issue bodies: stripped; PR bodies: preserved with the link rewritten** — and the earlier note here that lumped them together is corrected.

---

Issue #7254, opened 2026-08-10
HookContextSchema.input's contract table documents only the raw-engine envelope — every declarative hook sees a flat record view, and in a sandboxed `body` the documented `input.data` spelling is a TypeError


os-help:
Found while verifying #7225's premise (which the measurement below falsifies — see my report there). Filed standalone and unassigned: the examples are correct; the prose that describes them is not.

## The gap

The contract table on `HookContextSchema.input` (`packages/spec/src/data/hook.zod.ts`) documents exactly one shape — the raw engine envelope:

```
insert (one context per row): { data, options }
update (single id):          { id, data, options }
```

That is true of the surface it was written against: a handler registered with `engine.registerHook`, which is what `packages/objectql/src/hook-input-shape-contract.test.ts` pins. It is **not** the surface an app author writes on, and the table never says so.

Every **declarative** hook — a metadata `Hook` bound through `bindHooksToEngine`, i.e. everything from `defineStack({ hooks })` — is wrapped by `wrapDeclarativeHook`, which calls `installFlatInput` (`packages/objectql/src/hook-wrappers.ts:446`, helper at `:502`). That swaps `ctx.input` for a Proxy presenting a **flat record view** over the envelope: reads of any non-wrapper key resolve against `data`, and writes always land in `data`. Its own doc comment says so explicitly — "so the engine's downstream `input.data` read picks up mutations made by user code as `input.field = value`".

Sandboxed `body` hooks go one step further. `buildSandboxContext` (`packages/runtime/src/sandbox/body-runner.ts:314`) hands the script `unwrapProxyToPlain(engineCtx.input)`, and that unwrap runs through the proxy's `ownKeys` trap — which deliberately enumerates only the flat record fields. So inside a body, `ctx.input` is the **record**, and `ctx.input.data` is `undefined`.

## Measured

Real kernel, real objects, real QuickJS runner, persisted-row assertions (the same wiring `AppPlugin` uses in production — `ql.bindHooks(...)` with `bodyRunner: hookBodyRunnerFactory(new QuickJSScriptRunner(), …)`, `packages/runtime/src/app-plugin.ts:717-742`):

| surface | `ctx.input.field` | `ctx.input.data.field` |
| --- | --- | --- |
| code handler on a declarative Hook | works | works (the proxy passes `data` through) |
| sandboxed `body` hook | works | **TypeError** — `data` is undefined |

Shape probe from inside a body, printed by writing the answer to a persisted column:

```
[...]

Issue #7225, opened 2026-08-10
examples/app-crm + app-showcase: every authored hook reads the record off `ctx.input` instead of `ctx.input.data`, so each one silently does nothing at runtime


os-help:
Found while implementing #7036 (`examples/app-todo`'s completion stamp). Filed standalone and unassigned: #7036 repairs only app-todo's own hook, and the same mistake is live in the other two example apps under a different owner.

## The defect

`HookContext.input` is an **envelope**, never the record. The contract table on `HookContextSchema.input` (`packages/spec/src/data/hook.zod.ts`), pinned against the real engine in `packages/objectql/src/hook-input-shape-contract.test.ts`, says:

```
insert (one context per row): { data, options }
update (single id):          { id, data, options }
```

So the record is at `ctx.input.data`. Assigning to `ctx.input.<field>` sets a key on the envelope that no write path reads — the engine dispatches the driver write from `hookContext.input.data`.

Three shipped hooks do exactly that, and all three are registered (`hooks: allHooks`), so they run and have no effect:

1. `examples/app-crm/src/hooks/opportunity.hook.ts` — `OpportunityStageHook`, events `beforeInsert` + `beforeUpdate`:
   ```ts
   const input = ctx.input as { stage?: string; probability?: number };
   if (input.stage === 'closed_won') input.probability = 100;
   ```
   `input.stage` is `undefined` on every call, so probability is never pinned.

2. `examples/app-showcase/src/data/hooks/index.ts` — `StampInquiryDefaultsHook`, `beforeInsert` on `showcase_inquiry`:
   ```js
   if (!ctx.input.status) ctx.input.status = 'new'; if (!ctx.input.source) ctx.input.source = 'web';
   ```
   The web-to-lead defaults are never stamped.

3. Same file — `NormalizeTaskTitleHook`: `if (ctx.input.title) ctx.input.title = ctx.input.title.trim();` — never trims.

## Script bodies get the same envelope — no flattening anywhere

Worth stating because 2 and 3 are sandboxed `body` hooks rather than code handlers, so one might expect the runner to hand them the record. It does not. `buildSandboxContext` (`packages/runtime/src/sandbox/body-runner.ts`) sets `input: unwrapProxyToPlain(engineCtx?.input) ?? {}` — a snapshot of the envelope, unchanged. Its own comment is explicit that `input` is "the engine's own spelling, and the only one", and #5906 removed the alias limbs precisely so there would not be a second de-facto contract.

[...]

Issue #7036, opened 2026-08-09
examples/app-todo: a normal user can never mark a task complete — `completed_date` is `readonly` (stripped on update) and `completed_date_required` then refuses the write, so the app's own `completeTask` action always fails


os-project-manager:
Found while implementing #6882 (arming the `task_completion` record-change flow). Filed standalone: it is in the object + action metadata, not in the flow's trigger wiring, and it is independent of that repair.

## The defect

`examples/app-todo/src/objects/task.object.ts` declares:

```ts
completed_date: Field.datetime({
  label: 'Completed Date',
  readonly: true,          // <-- write is STRIPPED on the update path
}),
```

and, in the same object's `validationRules`:

```ts
{
  name: 'completed_date_required',
  message: 'Completed date is required when status is Completed',
  condition: P`record.status == "completed" && isBlank(record.completed_date)`,
}
```

The two are mutually unsatisfiable on the update path. `ObjectQL.update` runs the static-`readonly` strip **before** `evaluateValidationRules` (`packages/objectql/src/engine.ts`, the `reportDroppedFields(preRo, ..., 'readonly')` call sits immediately above the `evaluateValidationRules(updateSchema, ...)` call). So a payload carrying both keys loses `completed_date`, and the rule then sees a blank one and rejects the whole write.

The app's own completion action writes exactly that pair — `examples/app-todo/src/actions/task.handlers.ts`:

```ts
export async function completeTask(ctx: ActionContext): Promise<void> {
  const { record, engine } = ctx;
  await engine.update('todo_task', record.id as string, {
    status: 'completed',
    completed_date: new Date().toISOString(),
  });
}
```

so "Complete task" is non-functional for an ordinary user, and the same applies to `completeAllTasks`.

## Measured

Real kernel (ObjectQL + `@objectstack/driver-sql` better-sqlite3 `:memory:`) with app-todo's real `Task` object, one row inserted as `status: 'not_started'`:

```
update status+completed_date (user ctx):    REJECTED -> Completed date is required when status is Completed
update status only (user ctx):              REJECTED -> Completed date is required when status is Completed
update status+completed_date (isSystem):    OK  -> status=completed completed_date=2026-08-09T10:00:00.000Z
insert already-completed:                   OK  -> completed_date=2026-08-09T10:00:00.000Z
```

[...]

PR #7311, opened 2026-08-10
test(examples): pin the #7225 measurement — one persisted-row hook probe per example app (#7258)

### LINKED ISSUE #7258: Land the #7225 verification probes as regression pins: one persisted-row hook test per example app (app-crm + app-showcase)
Follow-up filed by the `domain:services` seat (#6021) on closing #7225 as a false alarm. `domain:*` left to the triage seat (expected: `domain:services`, matching the example-corpus precedent of #7036/#7225).

## Why

#7225 concluded that three shipped example hooks were silent no-ops; measurement proved all three work (the declarative flat-input proxy, `packages/objectql/src/hook-wrappers.ts:446`) and that the "fix" would have broken the showcase's public web-to-lead insert. The wrong call was cheap to make because **nothing pins these hooks' behaviour** — a careful reader deriving from three source files plus the (incomplete, see #7254) contract table gets it wrong, and the examples are the reference corpus AI authors copy from. One full dispatch was already spent on the false alarm; the pin stops the next one.

## What

Land the probes the #7225 dev already wrote and measured green, as permanent regression tests:

- `examples/app-crm`: real kernel + real driver, insert/update `opportunity` through the app's real object + registered `OpportunityStageHook`, assert the **persisted row** (closed_won → probability 100, closed_lost → 0, non-closed control untouched, handler sees flat `input.stage` bound). Costs two devDeps on app-crm (`@objectstack/objectql`, `@objectstack/driver-sql`) — flagged, deliberate.
- `examples/app-showcase`: same shape through the real QuickJS body runner (`hookBodyRunnerFactory`), assert persisted inquiry `status=new`/`source=web` and trimmed task title; keep the in-sandbox shape probe (`ctx.input` is the flat record, `hasData=undefined`) as the documented-surface witness.

Test-only; no `src/` behaviour change in either app. The probe transcripts and wiring are described in the #7225 close-out comment and the dev report quoted there.

Refs: #7225 (falsified card + probes), #7254 (the contract-table documentation gap, separate card), #7036 / PR #7222 (app-todo twin where the hook was genuinely unregistered).

os-help:
Fixes #7258

Lands the #7225 verification probes as permanent regression pins — one persisted-row hook test per example app. **Test-only: no `src/` change in either app.**

## Why

#7225 read the three shipped example hooks against `HookContextSchema.input`'s contract table (`insert` → `{ data, options }`, so the record lives at `ctx.input.data`), concluded all three silently no-op, and prescribed re-spelling them. Measurement answered the other way: all three work, and the prescribed fix would have turned the showcase's public web-to-lead insert into a hard refusal. One full dispatch was spent on that false alarm, and it was cheap to make because **nothing pinned these hooks' behaviour** — these are the reference-corpus apps AI authors copy from.

The missing link is `installFlatInput` (`packages/objectql/src/hook-wrappers.ts:446`, helper `:502`): `bindHooks` wraps every declarative hook in a Proxy presenting a **flat record view** over the envelope. The contract table describes the raw-engine surface accurately; an authored hook only ever meets the declarative one. That documentation gap is #7254, separately.

## What

- **`examples/app-crm/test/opportunity-stage-hook.test.ts`** — real `ObjectQL` over a real `SqlDriver` (better-sqlite3), the app's real `crm_opportunity` object and real `OpportunityStageHook`, bound the way `AppPlugin` binds it. Asserts the **persisted row**: insert closed_won → probability 100, insert closed_lost → 0, update into closed_won → 100, non-closed control untouched; plus a flat-`ctx.input` shape witness and an unbound-hook fixture.
- **`examples/app-showcase/test/hook-body-persisted-writes.test.ts`** — the same stack plus the real `QuickJSScriptRunner` behind `hookBodyRunnerFactory`, i.e. the AppPlugin production wiring (`packages/runtime/src/app-plugin.ts`). Asserts persisted inquiry `status=new` / `source=web`, the trimmed task title, the "caller supplied a value" guard, and keeps the **in-sandbox shape probe** as the #7254 witness.

Assertions land on rows read back out of the database. "The handler ran and had no effect" — the defect #7225 believed in — is invisible to anything that watches the handler instead of the row.

**Cost, flagged and deliberate:** `examples/app-crm` gains two workspace devDeps (`@objectstack/objectql`, `@objectstack/driver-sql`) to host its harness; the showcase already had everything it needed (verified, not assumed). Lockfile diff is 6 lines, both `link:` entries.

## The pins bite — reverse-verified, both apps, both directions

Every direction below was **measured**, then restored; `git diff` against `examples/*/src` is empty on this branch.

**app-crm** (7 cases green on main):

| local break | result |
| --- | --- |
| `allHooks = []` (hook unregistered) | **4 red** — the three probability cases + the bundle-registration case. `expected 50 to be 100`, `expected 90 to be +0`, `expected 40 to be 100`, `` `opportunity_stage_probability` must be in defineStack({ hooks }) `` |
| `'closed_won'` → `'CLOSED_WON_TYPO'` | **2 red** — exactly the two closed_won cases; closed_lost stays green |
| handler re-spelled to `ctx.input.data` (the #7225 prescription) | **7 green — unchanged** |

That last row is an **inverted direction, and it is honest rather than a hole**: on the code-handler path `data` is a wrapper key the proxy passes straight through, so both spellings really do work there. The spelling is only fatal inside a sandboxed `body`. The file says so in its header and points at its pair.

**app-showcase** (6 cases green on main):

| local break | result |
| --- | --- |
| both body hooks removed from `allHooks` | **4 red** — `expected null to be 'new'`, `expected '   spaced out   ' to be 'spaced out'`, the shape witness, the bundle-registration case |
| `StampInquiryDefaultsHook` re-spelled to `ctx.input.data` (the #7225 prescription) | **3 red**, and with the #7225 signature verbatim: `hook 'showcase_stamp_inquiry_defaults' threw: TypeError: cannot read property 'status' of undefined` — the insert is refused, zero rows persisted |

So the showcase pin is the one that catches the spelling regression, which is the half that was a live public-form outage. Two cases stay green under unregistration by construction — the "caller supplied a value" guard and the unbound-hook fixture both assert an **absence**.

## Premise check

Verified at the branch point (`8fbed3b`) before implementing, per the same-day churn note: `examples/app-crm/src/hooks/opportunity.hook.ts` and `examples/app-showcase/src/data/hooks/index.ts` are live and unchanged from what #7225 measured, and all three hooks still work. No probe failed against current main, so there was nothing to report as a regression.

## Changesets

No changeset. Test-only, and both example packages are `private: true` — this PR releases nothing, which is the textbook `skip-changeset` case (#4898). The `examples/app-crm/package.json` edit adds workspace-only devDeps consumed by the new test and by nothing that ships. `skip-changeset` applied on this PR.

## Verification

```
pnpm --filter @objectstack/example-crm --filter @objectstack/example-showcase test
  examples/app-crm       Test Files 2 passed (2)   Tests  27 passed (27)
  examples/app-showcase  Test Files 16 passed (16) Tests 164 passed (164)

pnpm --filter @objectstack/example-crm --filter @objectstack/example-showcase typecheck
  examples/app-crm typecheck: Done
  examples/app-showcase typecheck: Done

npx eslint (both new test files)       exit 0
node scripts/check-nul-bytes.mjs       OK (scanned 6672 text files)
```

Build closure (`--filter ' PKG ^...' build`, suffix form) ran first in a fresh worktree, before typecheck or test.

Refs: #7225 (the falsified card these probes come from), #7254 (the contract-table documentation gap), #7036 / #7222 (the app-todo twin, where the hook was genuinely unregistered).

---

### DIFF
--- a/examples/app-crm/package.json
+++ b/examples/app-crm/package.json
@@ -24,6 +24,8 @@
   },
   "devDependencies": {
     "@objectstack/cli": "workspace:*",
+    "@objectstack/driver-sql": "workspace:*",
+    "@objectstack/objectql": "workspace:*",
     "typescript": "^6.0.3",
     "vitest": "^4.1.10"
   }
--- a/examples/app-crm/test/opportunity-stage-hook.test.ts
+++ b/examples/app-crm/test/opportunity-stage-hook.test.ts
@@ -1,0 +1,276 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#7258] `OpportunityStageHook` pins the probability — on the PERSISTED row.
+ *
+ * This file exists because a careful reader got it wrong. #7225 read the three
+ * shipped example hooks against `HookContextSchema.input`'s contract table
+ * (`{ data, options }` — the record lives at `ctx.input.data`), concluded that
+ * this hook's `input.stage` is always `undefined` and that it silently does
+ * nothing, and prescribed re-spelling every one of them to `ctx.input.data`.
+ * Measurement answered the other way: all three work, and the prescribed fix
+ * would have turned the showcase's public web-to-lead insert into a hard
+ * refusal. One full dispatch was spent on the false alarm.
+ *
+ * THE MISSING LINK, and why the contract table did not lie: `bindHooks` wraps
+ * every DECLARATIVE hook in `wrapDeclarativeHook`, which calls `installFlatInput`
+ * (`packages/objectql/src/hook-wrappers.ts:446`, helper at `:502`). That swaps
+ * `ctx.input` for a Proxy presenting a FLAT RECORD VIEW over the envelope —
+ * reads of a non-wrapper key resolve against `data`, and writes always land in
+ * `data`, "so the engine's downstream `input.data` read picks up mutations made
+ * by user code as `input.field = value`". The contract table describes the
+ * RAW-ENGINE surface; an authored hook only ever meets the declarative one.
+ * (The table's silence about that is filed separately as #7254.)
+ *
+ * WHY THE ASSERTIONS ARE ON PERSISTED ROWS. The defect #7225 believed in was
+ * "the handler runs and has no effect" — a shape that a handler-side spy reads
+ * as healthy. Only the row that came back out of the database can tell
+ * "mutated the envelope nobody reads" from "mutated the record". So the harness
+ * is the real `ObjectQL` engine over a real `SqlDriver` (better-sqlite3), the
+ * app's REAL `crm_opportunity` object and the app's REAL hook, bound the way
+ * `AppPlugin` binds it from `defineStack({ hooks })`
+ * (`packages/runtime/src/app-plugin.ts`, `ql.bindHooks(hooks, { packageId })`).
+ * No double anywhere in the chain.
+ *
+ * WHAT THIS FILE DOES AND DOES NOT CATCH — measured, both directions, rather
+ * than assumed. Unregistering the hook turns four of the cases below red;
+ * misspelling its `closed_won` predicate turns exactly the two closed_won cases
+ * red and leaves closed_lost green. But re-spelling the handler to
+ * `ctx.input.data` — the change #7225 prescribed — keeps all seven GREEN, and
+ * that is correct rather than a hole: on the code-handler path `data` is a
+ * wrapper key the proxy passes straight through, so both spellings really do
+ * work here. The spelling is only fatal inside a sandboxed `body`, where
+ * `ctx.input.data` is `undefined`; the pin that bites on it is therefore the
+ * showcase's (`examples/app-showcase/test/hook-body-persisted-writes.test.ts`),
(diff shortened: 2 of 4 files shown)


os-help:
Queue-kick triage receipt (`domain:services` seat #6021, session `session_015fkdTyGmMD5s8ZtEifvuGy`): this PR left the merge queue ~06:55Z without merging. Signature check before any re-queue: the merge_group runs page covering the same window shows pr-7292/7310/7313/7317/7320/7324 all `success` and **no run for pr-7311 at all** — i.e. not a red of its own but a queue-rebuild cancellation (transient), most likely from the rebuild when sibling entries landed. Disposition per the three-way rule: **re-queued once** (auto-merge flip, `pr-7311` confirmed back in the queue branches on re-read). One re-queue is this PR's quota — a second identical ejection goes to the merge-queue steward (#5810) rather than another blind re-queue. The two pre-label `Check Changeset` reds on the head are the documented #5580/#6378 structural shape (post-label run `skipped`) and were not the cause.

---

PR #7333, opened 2026-08-10
fix(plugin-audit): resolve reference titles in `activityMilestones` summary tokens (#7290)

### LINKED ISSUE #7290: plugin-audit: `activityMilestones` summary templates still interpolate raw lookup ids — `{owner_id}` renders `oBK25…` after #7230

os-zhuang:
Fixes #7290

`matchMilestone` (ADR-0052 §5b.2) interpolated a milestone summary's `{token}`s from the after-row by calling `displayFieldValue` **with no title map**, so a `lookup` / `master_detail` / `user` token fell through to `String(value)` — the raw 32-char id. A milestone declared as `{ field: 'stage', value: 'closed_won', summary: 'Deal won by {owner}' }` shipped `Deal won by oBK25…` to the record timeline.

The milestone branch takes **precedence** over the tracked-change branch, so on every object that declares `activityMilestones` this was the string users actually saw, and #7230 / PR #7291's fix never reached them. Closing that is what this card is.

## Anchors re-verified at the branch point (`7fa2aae45`)

The card recorded its anchors at `a768f81a1` (pre-#7230) and warned the file had moved three times in a day. It had:

| anchor | card (`a768f81a1`) | at `7fa2aae45` |
|:--|:--|:--|
| `displayFieldValue` | `:308` | `:344` (gained #7230's `titlesFor` param) |
| `matchMilestone` | `:355` | `:472` |
| the `{token}` interpolation | `:368-375` | `:485-490` |
| `matchMilestone` call site | `:872` | `:1071` |

The premise itself was re-verified rather than taken on trust: at the branch point the interpolation still read `return field ? displayFieldValue(field, v) : String(v)` — two arguments, no title map — and `matchMilestone` had exactly **one** caller, already inside an `async` function. Premise valid, unchanged by #7291.

## Placement — re-derived, and measured both ways

The card recommended resolving **after** a milestone fires, keyed on the matched template's tokens, at the cost of making `matchMilestone` async. That direction is right and is what shipped. The async part turned out to be avoidable, so it was avoided:

`matchMilestone` is **split** rather than made async — detection (`matchMilestone`, still sync, no longer takes `fields`), planning (`planMilestoneTokenReads`), rendering (`renderMilestoneSummary`). The call site awaits one read between them. This gives the recommended read profile with **zero** async blast radius, and it mirrors the `planTrackedLookupReads` + resolver shape #7291 established one commit earlier in the same file rather than inventing a second one beside it.

`resolveTrackedLookupTitles` became `resolveLookupTitles(api, plan)` — it takes the plan instead of building it. Both branches now end in **one** read/mask/title implementation, which is what keeps #6977's masking contract from being re-typed for the milestone path.

### Read counts, measured with the copy-returning counting driver

| write shape | added reads |
|:--|:--|
| create (any) | **0** |
| delete (any) | **0** |
| update of a milestone-declaring object that fires **nothing**, on a row holding references | **0** |
| non-stage update of a milestone-declaring object, row holding references | **0** |
| repeat write while the record already sits at the milestone value | **0** |
| fired milestone whose template names **no** reference token | **0** |
| fired milestone whose reference token is **empty** | **0** |
| fired milestone, 1 reference token | **1** on that target |
| fired milestone, 2 tokens onto the **same** target | **1** (batched) |
| fired milestone, 2 **distinct** targets | **1 each** |
| the same token twice | **1** |
| target designating a **masked** field as its title | **0** (read skipped) |

**The rejected placement, measured rather than argued.** Resolving before knowing whether a milestone fired was implemented as mutation M3 and run: on a non-firing update of a row holding references it measured **3 reads** (`sys_user` 1, `crm_account` 1, `crm_region` 1; `crm_vault` 0, its title being masked) where the shipped shape pays **0**. That is the shape #6656 / PR #6977's Option A+ ruling was obtained to remove from this write path, so it was not adopted.

**Why the milestone plan does not filter on `trackHistory`.** #7230's gate is "a field that is BOTH tracked AND a reference actually moved". A milestone token is none of those: it is the author's own token, read from the whole after-row. The reported declaration itself (`'Deal won by {owner}'`) names a field that need not be tracked at all, so reusing that gate would leave the exact reported defect shipping — mutation M4 below measures it.

## #7291 and #6977 are re-asserted, not assumed

`audit-lookup-summary.test.ts` and `audit-bound-previous.test.ts` are **untouched and green**, and #6977's single-id count is re-asserted inside the new file on the firing path (still exactly one `findOne` on the audited object). `ledgerView`, the secret mask and the virtual-field drop are unchanged; the milestone read plan is built from the same masked after-view the interpolation renders from.

`crm_vault` in the new fixture declares `nameField: 'api_key'` on a real `secret` field (with a crypto provider wired, so the column is genuinely encrypted rather than conveniently empty). The read is **skipped**, not masked afterwards.

## What did not change

The author's template wording is theirs: only what a token *resolves to* changes. Punctuation, emoji, and the empty-token rule are byte-identical — an empty value still renders as the empty string, not `displayFieldValue`'s `∅`, because a milestone sentence is prose and not a diff row. Milestone templates are still not translated. Select option-label rendering is untouched — localizing those is #7289, deliberately held behind this card in the same file and left alone here.

## Tests

New file `packages/plugins/plugin-audit/src/audit-milestone-summary.test.ts` — 28 cases on a real `ObjectQL` engine with the copy-returning counting driver (a driver handing back live store references lets the read path rewrite the store under a measurement).

The label/locale fixture is a **real** `createMemoryI18n` bundle, and that earned its keep immediately: the first draft wrote `{object}` in the bundle and measured `已更新{object}"{label}"` — the shipped bundles interpolate `{{param}}`, which is deliberately *not* the milestone template's single-brace `{token}`. A stub translator would have hidden that.

```
pnpm --filter @objectstack/plugin-audit test       Test Files 11 passed (11)   Tests 188 passed (188)
pnpm --filter @objectstack/plugin-audit typecheck  tsc --noEmit — clean
npx eslint packages/plugins/plugin-audit --no-inline-config  — clean (exit 0)
```

Family gates, all green locally: `check:nul-bytes`, `check:durability-log-level`, `check:engine-double-contract`, `check:error-code-casing`, `check:route-envelope`, `check:wildcard-fallthrough`, `check:empty-changeset`, `check:adr-0087-registration`, `check:changeset-gate-self-tests`.

Predictions were written to a file and the tree committed before any mutation ran.

|:--|:--|:--|:--|
| M1 | drop the title map from `renderMilestoneSummary` (the pre-#7290 interpolation, plan and read kept) | 10 red: 9 rendering cases + the trailing summary assertion inside one read-count case; every pure count case green | **as predicted** — 10 red / 18 green, the same 10 |
| M2 | key the milestone plan by field instead of by target object | 1 red (the same-target batching case); the repeated-token case green | **wrong — 14 red.** See below |
| M2' | issue one read per id instead of one batched `id: { $in: [...] }` | isolates batching only | 3 red: my same-target case **and both of #7291's batching cases**, which also confirms the shared resolver really is shared |
| M3 | the rejected placement — resolve before `matchMilestone` | 10 red, all the zero/exact-count guards on the reference-holding row; rendering untouched | **as predicted** — 10 red / 18 green, the same 10, and the measured cost is the 3 reads quoted above |
| M4 | wrongly reuse #7230's `trackHistory` gate on the milestone plan | 14 red | **as predicted** — 14 red / 14 green |

**M2 is the one I got wrong, and the reason is worth recording.** I predicted the mutation would only coarsen batching. It did not: the plan map's key *is* the object name `resolveLookupTitles` reads from **and** the key `renderMilestoneSummary` looks titles up by. Keying it by field therefore made the resolver query objects named `owner` / `account_id`, the best-effort `catch` swallowed the failures, and no title resolved at all — 14 red, mostly rendering. The mutation was wrong, not the code. M2' was written to isolate the property M2 was supposed to test, and it does. My sub-prediction inside M2 — that the repeated-token case pins token dedup rather than object grouping, so it should stay green — held under M2'.

**The trap that PR #7291 documented is live here and was designed around, not discovered.** Every zero-read guard runs against `deal_2`, a row that already holds all six reference fields; one case (`a reference token whose value is EMPTY`) necessarily runs against the reference-free `deal_1`, and it is the one case that stayed **green under M3**. That is green-for-the-wrong-reason, predicted as such in writing beforehand and flagged in the file rather than counted as coverage. Likewise the raw-id fallback case was written from the start as a **mixed** assertion (`Won by usr_1 for APAC` — unresolvable side stays raw, resolvable side must still become a title), because #7291 measured that a single-token fallback assertion survives deletion of the render limb; under M1 it went red, as intended.

## Deliberately not done

- **#7289's select option-label localization** — same file, same function's neighbourhood, held at `pm:queue` behind this card. Not touched; a case pins the current rendering so the seam is visible when that card lands.
- **Milestone templates are still not localized.** They are author-written sentences and no bundle key exists for them; changing that would redefine what a milestone summary is, which is a contract decision and not this bug fix.
- **`multiple: true` lookups stored as an unparsed JSON string** still fall through to the raw value, exactly as on #7230's branch — in-memory arrays are unpacked and resolved from the same batch.
- **Historical rows** keep their write-time composition; only new writes improve.

---

### DIFF
--- a/.changeset/milestone-summary-lookup-titles.md
+++ b/.changeset/milestone-summary-lookup-titles.md
@@ -1,0 +1,38 @@
+---
+"@objectstack/plugin-audit": patch
+---
+
+fix(plugin-audit): `activityMilestones` summary tokens render referenced record titles, not raw ids (#7290)
+
+ADR-0052 §5b.2 lets an object declare a semantic milestone whose summary
+interpolates `{token}`s from the record — `{ field: 'stage', value:
+'closed_won', summary: 'Deal won by {owner}' }`. A token naming a `lookup` /
+`master_detail` / `user` field was interpolated with no title map, so it fell
+through to the raw stored id and the record timeline read `Deal won by oBK25…`.
+
+The milestone branch takes **precedence** over the tracked-change branch, so on
+every object that declares `activityMilestones` this was the string users saw,
+and #7230's fix to the tracked-change branch never reached them.
+
+A reference token now renders the referenced record's title, resolved through
+ADR-0079's `resolveDisplayField`. The author's template wording is untouched —
+only what a token resolves to changes — and the change is restore-invariant: an
+id with no resolvable title (target removed out of band, unregistered object,
+failing read) renders exactly as it did before. The empty-token rule is
+unchanged too: an empty value still renders as the empty string, not `∅`.
+
+**Read cost.** `matchMilestone` was split into detection and rendering so the
+read plan is built from the tokens of the template that actually **fired**, and
+only then. Measured with a copy-returning counting driver: every create, every
+delete, every update of a milestone-declaring object that fires no milestone,
+and every fired milestone whose template names no reference token all add
+**zero** reads; a fired milestone with reference tokens costs **one read per
+distinct target object** (two tokens onto the same object are one batched
+`id: { $in: [...] }`), paid only on the transition itself. The alternative
+placement — resolving before knowing whether a milestone fired — was measured at
+**3 reads on a write that fires nothing**, and was rejected: it is the shape
+#6656's Option A+ ruling was obtained to remove from this write path.
+
+A target object that designates a credential field as its title is **not read
+at all** — the same `collectMaskedReadFields` predicate the ledger masks with,
+so no secret can reach a user-facing summary through this path.
--- a/packages/plugins/plugin-audit/src/audit-milestone-summary.test.ts
+++ b/packages/plugins/plugin-audit/src/audit-milestone-summary.test.ts
@@ -1,0 +1,741 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#7290] `sys_activity.summary`'s MILESTONE branch (ADR-0052 §5b.2) — a
+ * `{token}` naming a `lookup` / `master_detail` / `user` field renders the
+ * referenced record's title instead of its raw 32-char id.
+ *
+ * ## Why this needed its own card and its own file
+ *
+ * #7230 / PR #7291 fixed the same rendering defect on the tracked-CHANGE
+ * branch, keyed on the diff. The milestone branch takes PRECEDENCE over it, so
+ * on an object that declares `activityMilestones` the string users actually see
+ * never went through that fix: `'Deal won by {owner}'` shipped as
+ * `'Deal won by oBK25…'`.
+ *
+ * The two branches also cannot share a gate. #7230's gate is "a field that is
+ * BOTH `trackHistory: true` AND a reference actually moved" — evaluated from
+ * the diff, before anything is read. A milestone token is none of those things:
+ * it is read from the whole AFTER-row, it is whatever the author wrote (the
+ * showcase's own `'Deal won by {owner}'` names an UNTRACKED field — pinned
+ * below), and whether a milestone fires at all is only known once
+ * `matchMilestone` has run.
+ *
+ * ## The read budget is the reason this file measures anything
+ *
+ * This is the third card in a row on this file's write hot path.
+ * #6656 / PR #6977 retired `captureBefore`'s redundant pre-image read under
+ * maintainer ruling Option A+ (2 → 1 per single-id write, 3 → 0 per predicate
+ * write); #7230 / PR #7291 preserved that by adding 0 reads to every write
+ * shape but "a tracked reference moved". This card had two placements available
+ * and they are NOT equivalent:
+ *
+ *   BEFORE `matchMilestone`  — speculate on the after-row, pay on EVERY update
+ *                              of a milestone-declaring object, including the
+ *                              overwhelming majority where nothing fires. This
+ *                              is the shape ruling A+ was obtained to remove.
+ *   AFTER it fires           — plan from the tokens of the template that
+ *                              MATCHED. Zero on every non-firing update; one
+ *                              read per distinct target object on the
+ *                              transition itself, which is rare by construction
+ *                              (`before !== value && after === value`, and
+ *                              never again while the record sits at that value).
+ *
+ * The second was implemented. Section 3 pins it as counts, not prose, and the
+ * zero-read cases run against a row that HOLDS references — measured, that is
+ * the difference between a guard and a guard-shaped no-op (PR #7291's own
+ * mutation D passed 160/160 against a reference-free row before its fixture was
+ * hardened; the same trap is live here and is avoided the same way).
+ *
+ * ⚠️ Read counts and rendering are pinned SEPARATELY on purpose: the read lives
+ * in `resolveLookupTitles`, the rendering in `renderMilestoneSummary`. Deleting
+ * the render limb leaves every count green.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { ObjectQL } from '@objectstack/objectql';
+import { createMemoryI18n } from '@objectstack/core';
+import { installAuditWriters } from './audit-writers.js';
+
+// ---------------------------------------------------------------------------
+// Fixtures
+// ---------------------------------------------------------------------------
+
+const f = (name: string, type: string, extra: Record<string, unknown> = {}) =>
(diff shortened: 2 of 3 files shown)


os-zhuang:
**Merge-queue ejection triaged — this is a deliberate single re-queue, not a blind one.**

Walking the `merge-queue-triage` checklist rather than hitting re-queue:

**Step 1 — is the failing test in a package this PR changed?** No. This PR changes three files, all `@objectstack/plugin-audit` plus a changeset; the failure is `packages/metadata-fs/test/watch-dot-root.test.ts`. No import path, no shared fixture, no shared global.

**Step 2 — has the same test appeared in another PR's queue-failure comment?** Not in the **60 most recently updated PRs** — this is the first recorded sighting, so I cannot claim "flaky 实锤" on a repeat. Filed **#7369** anyway, because step 2 asks for the issue regardless and the analysis holds without a second sighting: the case races a chokidar delivery against two fixed wall-clock budgets (a `sleep(400)` clearing a 200 ms self-write suppression window, then `Promise.race` against `EVENT_WAIT_MS`) and asserts an **exact** count of 1, so a late or suppressed delivery yields 0. The queue runs the **full** suite while PR-side CI runs the affected subset, which is why this test never ran on this PR's own CI — `Test Core (1/3, 2/3, 3/3)` were green on both head commits (`c6871848e`, `508a96643`). The test landed at 04:12:38Z today (`684ab2218`, #7208 / #7150), under four hours before this build.

**Step 3 — semantic conflict with the batch?** Nothing in the diff can interact with a filesystem watcher.

**Action: re-queued once.** If it ejects again on the same test, that is the second sighting, and I will stop re-queueing and escalate on #7369 rather than burn further full-queue rounds — the checklist is explicit that repeated re-queues rebuild everything behind me.

Nothing about this PR changed: still the two reviewed commits, still 23 success + 2 skipped-by-design on head `508a96643`.

---

Issue #7335, opened 2026-08-10
metadata-fs: `selfWrites` suppression is time-keyed, so a poll tick landing inside the 200 ms window can swallow an external edit


os-zhuang:
Found while working #7282. **Unassigned — recording, not claiming.** This is a
*different* mechanism from #7282 and was deliberately not folded into PR for
#7282, so that PR's reverse verification stays attributable to one change.

## What it is

`FileSystemRepository.put()` / `.delete()` suppress the chokidar event for a
path they just wrote by adding the path to a `Set` and clearing it on a fixed
wall-clock timer (`packages/metadata-fs/src/repository.ts`, `selfWrites` and the
two `setTimeout(..., 200)` calls). `handleFsChange` then drops **any** event for
a path in that set, without looking at what the watcher actually observed.

With `usePolling: true, interval: 1000`, chokidar compares state once per tick,
so a `put()` and an external edit that both land between two ticks are visible
as **one** change carrying the *external* content. If that single event is
delivered while the path is still in `selfWrites`, it is discarded as our own
write — and because it was the only event the external edit would ever produce,
nothing later recovers it.

The window is roughly: an external edit landing within ~60 ms of a `put()`
(50 ms `awaitWriteFinish.stabilityThreshold` plus the poll callback hop), with a
poll tick falling between the two writes. Realistic trigger: a `git checkout` or
an editor save arriving while the process is writing the same item.

## Status of the evidence — please grade accordingly

Honest about what is and is not measured:

- **Derived, not observed.** In 360+ instrumented iterations under deliberate
  event-loop starvation the suppression fired **zero** times
  (`handleFsChange entries hitting the selfWrites suppression: 0`), and the
  path was never in `selfWrites` at external-edit time (0/360). Those runs used
  a 300–400 ms pre-edit sleep, which cannot reach this window.
- The reason it is unreachable in the current tests is structural, not luck:
  the suppression timer is scheduled before the tests' own pre-edit `sleep()`
  and with a shorter delay, and Node runs expired timers in due-time order in a
  single timers-phase pass — so by the time the sleep resolves, the entry is
  provably gone. A sub-60 ms external edit is what escapes that argument, and
  nothing exercises one today.
- So this may well be observation-class rather than a live defect. I am filing
  it plainly rather than sitting on it; the severity call is the triage seat's.

## Note on the obvious fix

Making the suppression content-keyed (record the hash of what we wrote, clear
the entry when the content the watcher observes matches) removes the race
[...]

PR #7336, opened 2026-08-10
fix(metadata-fs): register written paths with the watcher instead of trusting its scan (#7282)

### LINKED ISSUE #7282: metadata-fs watcher tests are merge-queue flaky, and #7208's 20s deadline hardening did NOT fix it — the event is suppressed, not late
Filed from PR #7261 (an unrelated `packages/objectql` test-only change) after a merge-queue ejection.
Unassigned — recording, not claiming.

## The signature

```
FAIL test/fs-behavior.test.ts > FileSystemRepository — on-disk semantics
     > chokidar: external file change emits an update event
```

Confirmed flaky per the triage bot's checklist item 2 — the same signature has now ejected **two
different PRs**, one of which does not touch `packages/metadata-fs` at all:

| when (UTC) | PR | queue run | signature |
|---|---|---|---|
| 03:16 | #7208 | [31351599963](https://github.com/objectstack-ai/objectstack/actions/runs/31351599963) | `fs-behavior.test.ts` › chokidar external edit |
| 03:42 | #7208 | [31352802812](https://github.com/objectstack-ai/objectstack/actions/runs/31352802812) | `watch-dot-root.test.ts` › dot-rooted external edit |
| **04:33** | **#7261** | [31355694872](https://github.com/objectstack-ai/objectstack/actions/runs/31355694872) | `fs-behavior.test.ts` › chokidar external edit |

## The part that matters: the hardening was already in place

#7208's third-attempt patch (`3a9f5b880`, merged as `684ab22` at **04:12:38**) raised every positive
watcher wait to `EVENT_WAIT_MS = 20_000`, and its own comment set a standing rule: *"if this attempt
ejects again on a watcher-timing signature… the flaky-family card gets filed."* This is that card.

PR #7261's queue build was based on `b8e9fe27` (04:19:50), which **has `684ab22` as an ancestor**
(verified with `git merge-base --is-ancestor`). Reading the file as actually built in the queue's
merge commit `5c6a299d`:

```
$ git show 5c6a299d:packages/metadata-fs/test/fs-behavior.test.ts | grep -n EVENT_WAIT_MS
33:const EVENT_WAIT_MS = 20_000;
125:    await Promise.race([collectorDone, sleep(EVENT_WAIT_MS)]);
```

**The test failed with a 20-second deadline.** So the diagnosis behind the hardening — "queue load
makes the event late" — is now falsified. A 6.7× deadline increase changed nothing, which is the
signature of an event that is **never delivered**, not one that is slow.

## A mechanism that fits (read from source; not executed)

Two constants interact badly, and they are in production code rather than in the test:

```ts
// repository.ts — watcher
usePolling: true, interval: 1000, awaitWriteFinish: { stabilityThreshold: 50, pollInterval: 20 }

// repository.ts — self-write suppression, on every put()
this.selfWrites.add(file);
setTimeout(() => this.selfWrites.delete(file), 200);   // fixed wall-clock timer
```

The test does `put()` (a self-write), sleeps 300ms to clear the 200ms suppression, then edits the
file externally and expects exactly one event.

With **polling at 1000ms**, chokidar does not observe the two writes separately — it compares state
once per tick, so a `put()` at t≈0 and an external edit at t≈300ms are both visible as *one* change
at the t≈1000ms poll. Whether that single event is delivered or dropped depends on whether `file` is
still in `selfWrites` when the poll lands. Under queue load the 200ms `setTimeout` is a wall-clock
timer on a saturated event loop — if it fires late (or the poll lands early), the coalesced event is
**suppressed as a self-write**, and because it was the only event the external edit will ever
produce, no later event exists to wait for.

That explains the 20s result exactly: waiting longer cannot recover an event that was swallowed.

Predictions worth checking before trusting this: the failure should be all-or-nothing rather than
marginal (it is — `collected` is empty, not late), and it should get *more* likely as `interval`
(1000ms) grows relative to the 200ms suppression window.

## Suggested directions

Deadline widening is already spent; these attack the actual coupling:

1. **Make suppression content-keyed, not time-keyed** — clear `selfWrites` when the hash the watcher
   reads matches the hash we wrote, so an external edit can never be mistaken for our own write no
   matter how the timers land. Removes the race rather than widening it.
2. **Have the test await the self-write's own settling** before making the external edit, instead of
   `sleep(300)` guessing at the 200ms timer.
3. **Lower `interval`** for tests so poll granularity stops coalescing distinct writes — narrows the
   window but does not close it.
4. Failing all that, quarantine the watcher-timing family so it stops ejecting unrelated PRs from the
   merge queue.

(1) looks like the real fix; (2) is the cheap one that would stop the bleeding.

## Blast radius

This is not confined to `metadata-fs`'s own PRs — it ejected #7261, a one-file test-only change in
`packages/objectql` with no dependency path to `metadata-fs` (which depends only on
`@objectstack/metadata-core` and `chokidar`). Every ejection rebuilds every PR behind it in the
queue; the bot recorded **7 failed queue builds in 24h**.

os-zhuang:
Fixes #7282

## The proposed mechanism is falsified; the real one is worse

The card's mechanism — a `put()` and a later external edit coalescing into one
poll event that is then swallowed by the fixed 200 ms `selfWrites` timer — does
not happen. Measured, instrumented, over **360+ iterations under deliberate
event-loop starvation**:

```
selfWrites still present at external-edit time:                0/360
handleFsChange entries hitting the selfWrites suppression:     0
```

The suppression never fired once, in passing or failing iterations alike. It
cannot fire in these cases: the suppression timer is scheduled *before* the
test's own pre-edit `sleep()` and with a *shorter* delay, and Node runs expired
timers in due-time order in a single timers-phase pass — so the moment the
sleep resolves, the suppression entry is provably already gone.

What the failing iterations actually show is that `handleFsChange` **is never
entered at all**, because chokidar never emits anything for the path.

## The confirmed mechanism

Reproduced locally at ~4% (7 failures in 168 iterations) by running six probe
processes concurrently on a 4-core box with a busy-loop starving each event
loop. The instrumented timeline of a failure, taken from chokidar's own
internals:

```
[   7ms] start() done watcher=armed
[   7ms] put() begin
[  14ms] _handleRead CALL   probe3-zN8D48                 (root)
[  20ms] _handleRead DONE   probe3-zN8D48
[  20ms] _watchWithNodeFs   probe3-zN8D48
[  29ms] _handleRead CALL   probe3-zN8D48/view
[  30ms] _handleRead DONE   probe3-zN8D48/view -> items=[]   ← read EMPTY
[  30ms] _watchWithNodeFs   probe3-zN8D48/view               ← baseline taken
[  31ms] put() done                                          ← rename had landed
[ 565ms] external write BEGIN
[1316ms] external write DONE
[1566ms] chokidar RAW change probe3-zN8D48                (the ROOT, not view/)
[7583ms] FINAL getWatched={... "probe3-zN8D48/view":[] }   ← still empty, 7.5s later
```

chokidar's initial scan is asynchronous, and every write path in this file can
run while it is still walking the tree — `start()` arms the watcher and the
caller may `put()` on the next tick, and `ensureRoot()` arms it in the middle of
the very first write (#7000). With `usePolling` that produces a
permanently-blinding interleaving:

1. chokidar reads the type directory and finds it EMPTY — the atomic `rename` in
   `writeJsonAtomic` has not landed yet;
2. the rename lands, changing the directory's mtime;
3. chokidar calls `watchFile()` on the directory and libuv takes its polling
   baseline stat, which already reflects step 2.

The directory's stat then never changes again. No poll ever fires for it,
`_handleRead` never re-runs, the item file is never added to the watched set,
no per-file watcher is created, and `parent.has(basename)` stays false — which
`_handleFile`'s listener requires before it will emit `change` at all. chokidar
emits neither `add` nor `change` for that path **for the life of the process**.

This fits every observation on the card, and explains the two spent routes
rather than merely agreeing with them:

- **20 s and 25541 ms deadlines changed nothing** — there is no later event to
  wait for.
- **A 400 ms pre-edit sleep changed nothing** — the damage is done at
  watcher-arm time, before the sleep starts.
- **Lowering `interval` (direction 3) would change nothing either** — a shorter
  poll re-compares against the same unchanged directory stat. This is stronger
  than the card's "narrows the window without closing it": for this mechanism
  the effect is exactly zero.
- **Green in isolation** — the losing interleaving needs the rename to fall
  between a `readdirp` stream end and a `watchFile()` call, a gap that only
  opens up under scheduling pressure.

This is not only a test flake. In production `MetadataPlugin` attaches the
repository and the first `put()` can race the same scan, after which
`MetadataManager.subscribe()` silently never sees out-of-process edits to that
item — a hand edit, or a `git checkout` bringing metadata JSON in.

## The change

The window is exactly "files that exist at baseline time but were absent from
the snapshot read a moment earlier", and the only writer that can be inside it
is the repository itself. So `put()` now tells the watcher explicitly about the
path it created, instead of depending on a directory scan that may never notice
it. No timer is widened, and no consumer-side tolerance is added.

`add()` is idempotent here (`_handleFile` returns early when the parent already
tracks the basename) and emits nothing (chokidar treats an explicit `add()` as
an initial add, and `ignoreInitial` is set). Its effect is the one needed:
`_watchWithNodeFs` registers the basename with the parent directory and starts
the per-file poll.

## File surface (the real one)

- `packages/metadata-fs/src/repository.ts` — one call in `put()` plus the new
  private `trackWrittenPath`. The watcher options are **unchanged**.
- `packages/metadata-fs/test/watch-write-registration.test.ts` — new pin.
- `.changeset/metadata-fs-watch-write-registration.md`.

`selfWrites` and the chokidar `ignored` matcher (#7150) are untouched. No
quarantine was needed, so `packages/metadata`'s suite is unchanged too.

## Verification under load, not in isolation

A green local run proves nothing here, so the fix was measured against the
reproduction rather than against a quiet checkout. Identical harness, identical
load, my own base as the control:

| build | failures |
|---|---|
| base (`origin/main` @ `62b6a2fb2`) | **7 / 168** |
| with the fix | **0 / 360** |

Plus the real suites: `packages/metadata-fs` and `packages/metadata` run
concurrently, six rounds, with three CPU burners saturating the box —
`31/31` and `593/593` every round, 0 failed rounds.

## Reverse verification

Predicted: reverting the fix turns the new pin red on the registration
assertion, and does so **deterministically** rather than under load — the pin
measures registration latency, and the base's only route to registration is a
1000 ms poll tick that the case deliberately phase-anchors away from.

Result, 5 consecutive runs on base:

```
× registers a put() path with the watcher without waiting for a poll
AssertionError: expected [ 'anchor.json', 'seed.json' ] to include 'fresh.json'
Tests  1 failed | 30 passed (31)
```

5/5 red, same assertion, same message. So this red is **not** load-dependent,
unlike the defect it stands for — that difference is the point of the design and
is written into the test's header.

Honest categories:

- **Missed prediction.** The pin's first draft made its anchor write immediately
  after `start()` and the event never arrived — I read that as a second
  manifestation of the defect. It is not: a file that lands while chokidar is
  still walking is treated as pre-existing and, under `ignoreInitial`, correctly
  emits nothing (measured 3/3 with the write at delay 0, seen 3/3 at delay
  5 ms or after `ready`). The anchor now waits for chokidar's `ready`, and the
  finding is recorded in the test header.
- **Green in both directions (a guard, not evidence).** The case's closing
  liveness control — an external edit to the freshly written path must reach
  subscribers — passes with and without the fix on a quiet machine. It is
  labelled as a control in the file.
- **Left unmeasured.** I did not attempt to reproduce the failure inside the
  merge queue itself; the local reproduction is a proxy chosen to match the
  described conditions (full-monorepo concurrency), not the queue.

## Not folded in

- **#7150** (the chokidar `ignored` pattern) is untouched and unaffected — the
  dot-root cases pass unchanged.
- **A separate `selfWrites` finding** is filed as #7335 rather than fixed here: the
  suppression is time-keyed, and although it provably cannot lose the events in
  these tests, it can lose one when an external edit lands within roughly 60 ms
  of a `put()` and the poll tick falls in between. That is a different, narrow
  defect from the one this PR fixes, and folding it in would have muddied the
  reverse verification.

---

### DIFF
--- a/.changeset/metadata-fs-watch-write-registration.md
+++ b/.changeset/metadata-fs-watch-write-registration.md
@@ -1,0 +1,41 @@
+---
+"@objectstack/metadata-fs": patch
+---
+
+fix(metadata-fs): register every written path with the watcher, so an item created while chokidar is still scanning is not invisible forever (#7282)
+
+`FileSystemRepository`'s watcher could go **permanently blind to a single
+item** — external edits to that file produced no `MetadataEvent` for the whole
+life of the process, and nothing recovered short of a restart. The window is a
+race between chokidar's asynchronous initial scan and the repository's own
+first write, and both `start()` (which arms the watcher, after which the caller
+may `put()` on the next tick) and `ensureRoot()` (which arms it in the middle
+of the very first write, #7000) can open it.
+
+Measured on chokidar 5 with this repository's options (`usePolling`,
+`interval: 1000`):
+
+1. chokidar reads `<root>/<type>/` and finds it EMPTY — the atomic `rename` in
+   `writeJsonAtomic` has not landed yet;
+2. the rename lands, changing the directory's mtime;
+3. chokidar calls `watchFile()` on that directory and libuv takes its polling
+   baseline stat, which already reflects step 2.
+
+The directory's stat then never changes again, so no poll ever fires for it,
+the directory is never re-read, the item file is never added to the watched
+set, and no per-file watcher is created. `getWatched()` reports the type
+directory as `[]` while the file sits in it, and neither `add` nor `change` is
+ever emitted for that path.
+
+The fix does not widen any timer. The only writer that can be inside that
+window is the repository itself, so `put()` now tells the watcher explicitly
+about the path it created instead of depending on a directory scan that may
+never notice it. Registration is idempotent and emits nothing.
+
+User-visible effect: `MetadataManager.subscribe()` (and every consumer of
+`repo.watch()`) now reliably sees out-of-process edits — a hand edit, or a
+`git checkout` bringing metadata JSON in — to items written earlier in the same
+process. This was also the cause of four merge-queue ejections across three
+PRs; the two time-based mitigations tried before it (a 20s/25s event deadline
+and a wider pre-edit sleep) could not have worked, because the event was never
+delivered rather than late.
--- a/packages/metadata-fs/src/repository.ts
+++ b/packages/metadata-fs/src/repository.ts
@@ -305,6 +305,9 @@ export class FileSystemRepository implements MetadataRepository {
         // we keep it in selfWrites for one debounce tick.
         setTimeout(() => this.selfWrites.delete(file), 200);
       }
+      // The watcher must not depend on its own directory scan to notice a
+      // path we created ourselves (#7282). See `trackWrittenPath`.
+      this.trackWrittenPath(file);
       this.heads.set(key, hash);
 
       const evt: MetadataEvent = {
@@ -431,6 +434,53 @@ export class FileSystemRepository implements MetadataRepository {
       if (evt.hash === hash) last = evt;
     }
     return last;
+  }
+
+  /**
+   * Register a path this repository just wrote with the watcher (#7282).
+   *
+   * chokidar's initial scan is asynchronous, and every write path here can be
+   * running **while it is still walking the tree** — `start()` arms the watcher
+   * and the caller may `put()` on the next tick, and `ensureRoot()` arms it in
+   * the middle of the very first write. With `usePolling` that combination has
+   * a permanently-blinding interleaving, measured on chokidar 5 with this
+   * repository's own options:
+   *
+   *   1. chokidar reads `<root>/<type>/` and finds it EMPTY — the atomic
+   *      `rename` in `writeJsonAtomic` has not landed yet.
+   *   2. the rename lands; the directory's mtime changes.
+   *   3. chokidar calls `watchFile()` on that directory, and libuv takes its
+   *      polling baseline stat — which already reflects step 2.
+   *
+   * From then on the directory's stat never changes again, so no poll ever
+   * fires for it, `_handleRead` never re-runs, the item file is never added to
+   * the watched set, and no per-file watcher is ever created. chokidar emits
+   * neither `add` nor `change` for that path **for the life of the process** —
+   * `getWatched()` reports the type directory as `[]` forever while the file
+   * sits in it. That is the whole of #7282: the four merge-queue ejections all
+   * waited out their deadlines (20s, then 25541ms against 25s) on an event that
+   * was never going to be delivered, which is why widening the deadline and
+   * widening the pre-edit sleep both changed nothing, and why lowering
+   * `interval` would change nothing either — a shorter poll re-compares against
+   * the same unchanged directory stat.
+   *
+   * The window is exactly "files that exist at baseline time but were absent
+   * from the snapshot read a moment earlier", and the only writer that can be
+   * inside it is us. So we close it at the source: tell the watcher explicitly
+   * about every path we create, instead of hoping its scan happened to see it.
+   *
+   * `add()` is idempotent here — `_handleFile` returns early when the parent
+   * directory already tracks the basename — and it emits nothing, because
+   * chokidar treats an explicit `add()` as an initial add and `ignoreInitial`
+   * is set. Its effect is the one we need: `_watchWithNodeFs` registers the
+   * basename with the parent directory (without which chokidar drops `change`
+   * events for the file) and starts the per-file poll.
+   */
+  private trackWrittenPath(file: string): void {
+    const w = this.watcher;
+    // `add()` clears `closed`, so never hand a closing watcher a new path.
+    if (!w || w.closed) return;
+    w.add(file);
   }
 
   private startWatcher(): void {
--- a/packages/metadata-fs/test/watch-write-registration.test.ts
+++ b/packages/metadata-fs/test/watch-write-registration.test.ts
@@ -1,0 +1,183 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * #7282 — a path this repository writes must be registered with the watcher
+ * **without waiting for a poll interval**.
+ *
+ * The defect this pins is a permanently-blinding interleaving between
+ * chokidar's asynchronous initial scan and our own first write, measured on
+ * chokidar 5 with this repository's options (`usePolling`, `interval: 1000`):
+ *
+ *   1. chokidar reads `<root>/<type>/` and finds it EMPTY — `writeJsonAtomic`'s
+ *      `rename` has not landed yet.
+ *   2. the rename lands; the directory's mtime changes.
+ *   3. chokidar calls `watchFile()` on the directory and libuv takes its
+ *      polling baseline stat, which already reflects step 2.
+ *
+ * The directory's stat then never changes again, so the directory is never
+ * re-read, the item file is never registered, no per-file watcher is created,
+ * and neither `add` nor `change` is emitted for it **for the life of the
+ * process**. Measured directly: `getWatched()` reported the type directory as
+ * `[]` for the full 7.5s of a probe run while the file sat in it, and
+ * `handleFsChange` was never entered once.
+ *
+ * That is why both time-based mitigations tried on this family failed: the
+ * event is never delivered, so a 20s deadline (#7208) and a 25541ms one (#7255)
+ * both ran out with an empty array, and a wider pre-edit sleep cannot help
+ * either — the damage is done before the sleep starts. Lowering `interval`
+ * cannot help for the same reason: a shorter poll re-compares against the same
+ * unchanged directory stat.
+ *
+ * ## Why this file asserts registration rather than the race
+ *
+ * The interleaving above lives inside chokidar, between a `readdirp` stream end
+ * and a `watchFile()` call, so it cannot be forced from the outside without a
+ * test seam. What CAN be pinned deterministically is the property the fix
+ * establishes, which is what actually closes the window: after `put()` returns,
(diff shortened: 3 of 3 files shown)


os-zhuang:
## ACCEPT — PM review (step 7), anchored to head `0910427fbf7e64e25bfdea16bf34186a97d6ce90`

Declared three files; the diff is exactly those three, nothing else:

| file | status | matches declaration |
|---|---|---|
| `packages/metadata-fs/src/repository.ts` | modified | yes — one call in `put()` + the new private `trackWrittenPath`; watcher options untouched |
| `packages/metadata-fs/test/watch-write-registration.test.ts` | added | yes — the pin |
| `.changeset/metadata-fs-watch-write-registration.md` | added | yes |

`selfWrites` is verifiably untouched (its `setTimeout(… delete)` is context, not a changed line). The chokidar `ignored` matcher (#7150) is untouched. No `content/docs/releases/` change, no `docs/adr/**` change, no quarantine, no consumer-side tolerance widened anywhere.

### CI, per job
25 check runs, all `completed`: **23 `success`**, **2 `skipped`** (`Build Docs`, `Console Pin Gate` — both path-filtered and correctly inert for a non-docs PR). No failure, no neutral, nothing still running. Named gates: `ESLint` success, `TypeScript Type Check` success, `Build Core` success, `Test Core` 3/3 + rollup success, `Dogfood Regression Gate` 3/3 + rollup success, `Dogfood Verify CLI` success, `Temporal Conformance (live PG + MySQL)` success, `Check Changeset` success, `ADR maintainer approval` success.

### What I am accepting, stated plainly

**The dev did not implement the direction the card recommended — it falsified the card's mechanism first.** The card proposed that a `put()` and a later external edit coalesce into one poll event which the fixed 200 ms `selfWrites` timer then swallows, and named content-keying that suppression as direction 1. Measurement says that never happens: `0/360` iterations had a `selfWrites` entry alive at external-edit time and the suppression branch was entered `0` times, in failing and passing iterations alike — with an argument for *why* it cannot fire here (the timer is scheduled before the test's own pre-edit `sleep()` and with a shorter delay, and Node drains expired timers in due-time order). In the failures `handleFsChange` is never entered at all. The branch name `claude/issue-7282-selfwrites-content-keyed` is therefore a self-declared misnomer; it is left as-is because renaming a branch mid-PR costs more than the confusion it saves, and this comment plus the PR body are the record.

**The mechanism it found instead is strictly worse than the card's, and I accept the evidence for it.** chokidar's asynchronous initial scan races the repository's own first write: the type directory is read while `writeJsonAtomic`'s rename has not landed, then `watchFile()` takes that directory's polling baseline *after* it lands, so the stat never changes again, the directory is never re-read, `parent.has(basename)` stays false, and **neither `add` nor `change` is ever emitted for that path for the life of the process**. `getWatched()` reporting the type directory as `[]` for 7.5 s with the file sitting in it is the observation that settles it. This explains the two mitigations already spent on this family rather than merely agreeing with them — a 20 s deadline and a 25541 ms deadline both waited on an event that was never coming — and it converts direction 3 from "narrows the window" to **effect exactly zero**, since a shorter poll re-compares the same unchanged stat. Two claims about the codebase I had assumed and that this corrects: I treated the flake as bounded by the test harness, and I treated the poll interval as a tunable that would at least help. Both wrong.

**It is a production defect, not only a test flake.** `MetadataPlugin` attaches the repository and the first `put()` can race the same scan, after which `MetadataManager.subscribe()` silently never sees out-of-process edits to that item — a hand edit, or a `git checkout` bringing metadata JSON in. The changeset carries that as the user-visible effect, at `patch`, which is the right framing and the right bump.

**Verification was done under load, as I required, not in isolation.** Base `origin/main @ 62b6a2fb2` failed **7/168** probe iterations with the CI signature; with the fix, **0/360** under identical harness and identical load, plus six concurrent rounds of the real `metadata-fs` and `metadata` suites (`31/31`, `593/593`, zero failed rounds) with the box saturated. The control is the dev's own base, not a moving `origin/main`.

**Reverse verification is honest in all four categories.** Predicted red, and predicted it would be *deterministic* rather than load-dependent because the pin phase-anchors the poll — measured 5/5 red on the same assertion and the same message. One **missed prediction** reported rather than buried: the pin's first draft anchored immediately after `start()` and saw nothing, which the dev initially read as a second manifestation of the defect; it is not — a file landing during the initial walk is treated as pre-existing and under `ignoreInitial` correctly emits nothing (3/3 at delay 0, 3/3 the other way at 5 ms or after `ready`), so the anchor now waits for `ready` and the finding is written into the test header. One assertion **green in both directions** — the closing external-edit liveness check — is labelled in the file as a control, not as evidence. One item **left unmeasured** and said so: the failure was not reproduced inside the merge queue itself; the local reproduction is a stated proxy.

**Splitting #7335 out was the right call and I am endorsing it explicitly.** The `selfWrites` window is time-keyed and genuinely can lose an event when an external edit lands within ~60 ms of a `put()` with a poll tick in between — but it is a different defect, still unobserved in the wild, and folding it in would have left this PR's reverse verification attributable to two changes instead of one. Filed, not fixed here.

### Ruling
ACCEPT. Flipping out of draft and enabling auto-merge (SQUASH); it lands through the merge queue in the normal way. This closes the head-of-queue flake that ejected verified-green PRs four times across three PRs in roughly two hours.

---

Issue #4674, opened 2026-08-02
fix(metadata-protocol): two internal `engine.find` calls sort by `direction`, a key no driver reads — audit history and global search silently return ASCENDING


os-zhuang:
Two internal `engine.find` calls in `protocol.ts` write their sort as `{ field, direction: 'desc' }`. The QueryAST sort shape is `{ field, order }` (`SortNodeSchema`, `packages/spec/src/data/query.zod.ts:12-15`). `direction` is not read by anything on that path, so both queries silently run **ascending**.

Because both carry a `limit`, a wrong direction does not merely reorder the page — it changes **which rows come back at all**.

## The two sites

`packages/metadata-protocol/src/protocol.ts:3316` — metadata audit history:

```ts
const rows = await this.engine.find('sys_metadata_audit', {
    where,
    orderBy: [{ field: 'occurred_at', direction: 'desc' }],
    limit,
} as any);
```

Intent: the newest `limit` audit events. Actual: the **oldest** `limit` events. The history panel shows the beginning of an object's life and never its recent changes — and on a long-lived object it will never show anything an editor is looking for.

`packages/metadata-protocol/src/protocol.ts:5104` — global search:

```ts
const opts: any = {
    where,
    limit: perObject,
    orderBy: [{ field: 'updated_at', direction: 'desc' }],
};
```

Intent: most-recently-updated matches first. Actual: the **stalest** matches, truncated to `perObject`. Recently-edited records are the ones a searcher is most likely to want, and they are exactly the ones excluded.

## Why `direction` does nothing

Both real drivers normalize off `.order`, and neither has a fallback for `direction`:

- `packages/plugins/driver-sql/src/sql-driver.ts:5659`
  ```ts
  keys.push({ field: item.field, direction: item.order === 'desc' ? 'desc' : 'asc' });
  ```
- `packages/plugins/driver-mongodb/src/mongodb-driver.ts:663`
  ```ts
  lastDirection = item.order === 'desc' ? -1 : 1;
  ```

With `order` absent, `undefined === 'desc'` is false and both land on ascending. Note `sql-driver` uses `direction` as its **internal** post-normalization vocabulary — which is likely how the wrong key looked plausible to whoever wrote the call sites.

## Why nothing caught it

Worth recording, because the fix for the two lines does not address any of it:

[...]

PR #7337, opened 2026-08-10
fix(objectql)!: `engine.find`/`findOne` refuse an unmaterializable formula ORDER BY (#7095)

### LINKED ISSUE #7095: `engine.find()` still drops a `formula` ORDER BY silently — decide whether the engine refuses or keeps its internal-caller tolerance
Filed from #6994's implementation (PR pending). Recorded, **not claimed**. Routing suggestion: `domain:engine-core` — this is a contract decision about `engine.find`, not a gate fix.

## What #6994 closed, and what it did not

#6994 refuses a non-dotted `orderBy` naming a `formula` field at **ingress** (`assertSortFieldsExist`, `400 INVALID_SORT`). That covers everything reaching `findData`: the REST list route, `POST /data/:object/query`, the export route (its `$orderby` funnels through `findData`), and the RPC dispatcher.

It does **not** cover a caller that reaches `engine.find()` directly — hooks, flows, reports, expand sub-reads. Measured on the #6994 branch **with the fix in place**, one script, real `SqlDriver` (better-sqlite3, on-disk) + real `ObjectQL` + the real protocol on top, five rows inserted `C A E B D`, `sort_key` a formula over `record.title`:

```
PROTOCOL  {"sort":"sort_key"}          REFUSED 400 INVALID_SORT: … a formula field on 'repro_contact' …
PROTOCOL  {"orderBy":["sort_key"]}     REFUSED 400 INVALID_SORT
PROTOCOL  {"sort":"-sort_key"}         REFUSED 400 INVALID_SORT

FORMULA   orderBy sort_key asc         ["C","A","E","B","D"]   engine.find direct — still silent
FORMULA   orderBy sort_key desc        ["C","A","E","B","D"]
   asc === desc (byte-identical)?      true
```

Both halves in one run: the door refuses, the direct path still answers 200 in arbitrary order. PR for #6994 carries a test explicitly labelled `RECORD OF A KNOWN HOLE` pinning this — it records the **unfixed** behaviour and says it should go red the day this issue is closed.

## The decision, not the patch

This is deliberately not filed as "add the same check to the engine", because the engine's posture here is documented and load-bearing in the opposite direction. `assertProjectionFieldsExist` states it in the protocol source: the engine's tolerance "guards INTERNAL callers (hooks, flows, expand sub-reads, registry-less hosts) that never pass through this ingress". `planFormulaProjection` acts on the same posture — it **drops** virtual formula names from `ast.fields` rather than refusing them. So making `engine.find` refuse is a new posture, not an extension of an existing one.

Options, all real:

1. **`engine.find` refuses an unmaterializable `orderBy`.** One place, every driver and every caller covered. Cost: reverses a documented tolerance; any internal caller that currently sorts by a formula field starts throwing where it used to silently under-deliver. Needs a sweep of internal callers first.
2. **`engine.find` drops it the way it drops virtual names from the projection, but observably** — e.g. via the existing `DroppedFieldsEvent` channel rather than silently. Keeps the tolerance, removes the invisibility. Cheaper, and consistent with `planFormulaProjection`'s existing behaviour.
3. **Leave it at ingress.** The external door is the one an author can hit by hand; internal callers are `tsc`-covered code we own. Cost: the two halves of one contract answer differently, which is exactly the split #4226 was filed about on the other axis.

⚠️ One trap worth writing down before anyone reaches for it: the engine *could* sort post-hoc, after `applyFormulaPlan` evaluates the formulas. That looks like "make formula sorting work" and is a trap — `driver.find` has already applied `limit`/`offset`, so post-sorting reorders an arbitrary page. It would appear to work on small result sets and be wrong the moment pagination is involved.

## Also worth deciding here: one spelling of "virtual"

"Which field types materialise no column" is currently spelled in six places, none of them `packages/spec`: `driver-sql`'s `fieldHasColumn` and `createColumn`, `driver-turso`'s `remote-transport`, `objectql`'s `planFormulaProjection` and `search-companion`, `plugin-audit`'s `VIRTUAL_FIELD_TYPES` — and now `metadata-protocol`'s `UNMATERIALIZED_SORT_TYPES` (#6994 added the sixth deliberately, local and documented, rather than opening a cross-package spec edit for one string). Note the spec's `COMPUTED_VALUE_TYPES` (`formula`/`summary`/`autonumber`) is **not** this set — it is the write contract, and `summary`/`autonumber` both get real columns and sort correctly. Consolidating is a cross-package change (#6298 cross-seat declaration) and is a candidate to fold into whichever option above wins.

Refs: #6994 (the ingress half, PR pending), #6924 (the dotted hint text), #6673 (the SEARCH axis), #3821 (the backstop that swallows it), #4226 / #4256 (the family).

os-zhuang:
Closes #7095. Follows #6994 (the ingress half) and #6924 (the sort-hint prescription).

## The defect, reproduced on current `main` first

#6994 refuses a non-dotted `orderBy` naming a `formula` field at **ingress** (`assertSortFieldsExist`, `400 INVALID_SORT`), which covers everything reaching `findData`: the REST list route, `POST /data/:object/query`, the export route and the RPC dispatcher. It could not cover a caller reaching `engine.find()` / `engine.findOne()` **directly**.

Reproduced on `06be54ec3` before touching anything — the `RECORD OF A KNOWN HOLE` pin #6994 left behind still passed, i.e. the hole was still open:

```
engine.find(o, { orderBy: [{ field: <formula>, order: 'asc'  }] }) -> C A E B D
engine.find(o, { orderBy: [{ field: <formula>, order: 'desc' }] }) -> C A E B D
                                             asc === desc (byte-identical)
```

A `formula` value is computed on read, so no driver materialises a column for it: the ORDER BY reached the driver, found nothing, and the unknown-column backstop returned the rows unordered under a success — carrying the very values they were asked to be ordered by. With `limit`, "the latest N" was an arbitrary N nothing in the response could reveal.

## The ruled direction

Per the maintainer ruling (2026-08-10 on #7095): **refuse at the public boundary** with guidance prose, never a silent drop. `assertOrderByIsMaterializable` refuses on both entry points with the same `400 INVALID_SORT` and the same remedy sentence the two ingress verdicts emit.

`findOne` is covered as well as `find` — same public boundary, and the stakes are higher there: `findOne` applies `limit: 1`, so a dropped sort does not merely reorder the answer, it returns a **different record** that looks just as legitimate.

## The measured verdict on internal reliance

The ruling made the internal-caller tolerance conditional on a **measured** internal call site relying on it. **The sweep found none**, so **outcome 1** applies: refuse everywhere, no internal path shipped.

Sweep (151 non-test `orderBy` occurrences outside `spec`/`metadata-protocol`, narrowed to those reaching the engine directly — drivers are the receiving end, `client`/`client-react`/MCP go over the wire through ingress):

| Internal call site | Sort field(s) | Formula? |
| --- | --- | --- |
| `plugin-approvals/approval-service.ts` (×4) | `created_at` | no |
| `plugin-reports/report-service.ts` | `updated_at`, `next_run_at`, **`q.orderBy`** | author-driven |
| `plugin-sharing` (×4) | `created_at`, `name` | no |
| `service-queue/db-queue-adapter.ts` | `created_at`, `priority`, `scheduled_for` | no |
| `service-job/db-job-adapter.ts` | `started_at` | no |
| `service-messaging/messaging-service.ts` | `created_at` | no |
| `plugin-email/outbox-sweep.ts` | `created_at` | no |
| `metadata/loaders/database-loader.ts` | `recorded_at`, `version` | no |
| `metadata/utils/history-cleanup.ts` | `version` | no |
| `runtime/domains/share-links.ts` | `created_at` | no |
| `plugin-auth/objectql-adapter.ts` (×2) | caller `sortBy` | caller-driven |
| `types/keyset-walk.ts` | keyset `key` (defaults `id`) | no |
| `engine.expandRelatedRecords` | nested `expand` sort | author-driven |

Two independent facts make the verdict solid rather than a shrug:

1. Every **hardcoded** internal sort names a real stored column.
2. **No shipped object anywhere in the repo declares a `formula` field at all** — the only non-test occurrences are the spec's own field builder (`field.zod.ts`) and one FLS contract fixture. So no hardcoded internal sort *could* name one.

No test relied on the drop either, except the deliberate `RECORD OF A KNOWN HOLE` pin — which said in as many words that it should go red the day this landed. It is now inverted in place.

The dynamic sites do not *rely* on the tolerance; they are the paths through which the silent drop was **author-reachable**, which is exactly why ingress-only was not tenable: a saved report's `query.orderBy` is forwarded verbatim into `engine.find` by `plugin-reports`, bypassing the ingress gate entirely.

## One path deliberately does NOT become a refusal ⚠️

A nested `expand` sort raises this refusal inside `expandRelatedRecords` — but that sub-read sits inside a **pre-existing graceful-degradation `catch`** ("if expand fails, keep original IDs") which swallows *every* expand failure. Measured:

```
WARN Failed to expand relationship field; retaining foreign key IDs
     { field: 'parent_id', error: "ObjectQL.find('showcase_task') sorts by 'sort_key',
       a formula field … Denormalise the value onto … " }
```

So that path moves from **silent** to **observable** (a warning naming the field and the fix) rather than refusing. Reversing that backstop is the #3821-family swallow — a decision about *all* expand failure modes, not a rider on this card. It is measured and pinned as-is rather than left implied; flagged for follow-up.

## Scope held deliberately narrow

- The **ingress gate is untouched** — same message, same `unknown` > `dotted` > unmaterializable precedence, same `param` name the engine cannot know. Not weakened.
- The engine door judges **only the third verdict**. Unknown and dotted sort names still reach the driver from a direct call, because refusing those is a posture change on two further axes, not a free extension of this one.
- `summary`/rollup fields are **unaffected** and still sort in both directions — they get a real maintained column. A control pins that, so widening the set to the spec's `COMPUTED_VALUE_TYPES` (the *write* contract) goes red.
- Reading a formula field, and the projection axis' `SELECT *` tolerance, are untouched — pinned.
- The post-hoc-sort trap the card called out is written into the code comment: `driver.find` has already applied `limit`/`offset`, so sorting after `applyFormulaPlan` would reorder an arbitrary page.

## Docblocks updated honestly

`assertProjectionFieldsExist`'s documented internal-caller tolerance is the posture this narrows, so it now says the tolerance is **per-axis** and that the sort axis no longer has it — and that nothing equivalent has been measured for the projection axis, so the sentence is not a licence to assume it. `assertSortFieldsExist`'s SCOPE paragraph no longer claims the direct path "still gets the silent drop".

## Pins

- `engine.find` refuses — asc, desc, and as the second of two sort keys — with asserted `status` + `code` + `field` + `object`, not merely "throws".
- `engine.findOne` refuses, with a `where` present so it is the sort verdict and not `requireFindOnePredicate` answering first.
- Guidance prose asserted: names the entry point, the field, the type, "computed on read", and never prescribes what it refuses.
- **Three doors agree word-for-word on the remedy** (was two) — the anti-drift pin, and the reason the engine duplicates the prose instead of importing it (see below).
- **Negative pin**: no internal path — four plausible opt-out flags are all refused by the public options shape, and the old silent-drop shape now throws.
- Controls: a real column and a `summary` field still sort through `engine.find`; a formula field is still selectable and still computed.
- The ingress pins from #6994 are untouched and still green.

### Why the prose is duplicated rather than imported

`metadata-protocol` is assembled **from** an engine (`assembleMetadataProtocol`), so it is the layer *above* `objectql`; importing its error helper into `engine.ts` would invert the layering. The word-for-word equality pin is what keeps the duplication honest.

## Changeset / ADR-0087

Declared **`major`** for `@objectstack/objectql` — a public API throws where it used to succeed. That triggers the ADR-0087 disposition requirement, so the trio is filled in following #7210 (`708431313`) as the worked example:

- `.changeset/engine-find-formula-orderby-refusal.md` with `<!-- adr-0087: registered engine-find-formula-order-by-refused -->`
- `packages/spec/src/migrations/registry.ts` — step-17 `semantic` entry
- `packages/spec/spec-changes.json` + `docs/protocol-upgrade-guide.md` regenerated (`gen:spec-changes`, `gen:upgrade-guide`); both `--check` gates green

`registered` rather than one of the three `not-required` forms: `@objectstack/objectql` is published, no prior id covers this surface, and the change ships rewrite instructions — so `no-migration-prescription` would be self-contradictory.

⚠️ **`packages/spec` is touched** for the ADR-0087 disposition (ledger + generated artifacts only, no schema change) — cross-seat declaration per #6017.

Refs #7095, #6994, #6924, #4226, #4256, #3821, #7210, ADR-0087, ADR-0112

---
_Generated by [Claude Code](https://claude.ai/code/session_01HJgGahRqaYPRJ2oKmk9Czc)_


os-zhuang:
## Gates run locally

Enumerated with the prescribed command: **64** in `lint.yml`, **70** across all workflows.

> ⚠️ Enumeration caveat worth recording: the 70-union includes at least one name that is **not a runnable gate** — `check:variant-docs` appears only inside a lint.yml *comment* (the #4177 post-mortem about unclassified scripts), not as a step. `grep -oE 'check:[a-z0-9:-]+'` cannot tell a step from a comment.

**All green (exit 0):**

`check:adr-0087-registration` (self-test: 142 assertions + real scan) · `check:empty-changeset` · `check:changeset-no-major` · `check:changeset-fixed` · `check:changeset-gate-self-tests` · `check:spec-changes` · `check:upgrade-guide` · `check:generated --reconcile-only` · `check:docs` · `check:authorable-surface` · `check:export-origins` · `check:api-surface` · `check:exported-any` · `check:dual-source-exports` · `check:error-code-casing` · `check:query-options-erasure` · `check:engine-double-contract` · `check:route-envelope` · `check:adr-anchors` · `check:nul-bytes` · `check:strictness-ledger` · `check:type-check-coverage` · `check:type-check-debt` · `check:agent-model-declared` · `check:app-nav-i18n` · `check:authz-resolver` · `check:doc-authoring` · `check:doc-formula-expressions` · `check:docs-audit-scope` · `check:driver-conformance` · `check:driver-memory-census` · `check:durability-log-level` · `check:i18n` · `check:i18n-coverage` · `check:init-service-contract` · `check:kernel-hook-pairs` · `check:merge-driver` · `check:meta-type-normalized` · `check:node-version` · `check:org-identifier` · `check:published-files` · `check:quick-reference-counts` · `check:release-body` · `check:release-notes` · `check:required-contexts` · `check:resume-authority-declared` · `check:role-word` · `check:service-providers` · `check:shard-attestation` · `check:slot-lookup` · `check:spec-parsed-alias` · `check:stack-collection-maps` · `check:stall-guard` · `check:startup-registry-verdict` · `check:tenant-chokepoint` · `check:verify-stand-in` · `check:wildcard-fallthrough` · `check:workflow-status-functions` · `check:objectui-changeset` · `check:skill-compatibility` · `check:skill-frame-sync`

Plus `pnpm lint` (eslint, `--no-inline-config`) and `turbo run typecheck` — **16/16 tasks successful** across the touched packages and their dependents.

**Pre-existing red, NOT caused by this PR:**

- `check:platform-checklist` — fails with `coverage.json · qa: UNCLASSIFIED`. Verified by running the same script in a clean worktree at `origin/main` (`06be54ec3`): **identical single failure**. Nothing in this diff touches `docs/qa/`.

## Tests

| Suite | Result |
| --- | --- |
| `@objectstack/objectql` | **2960 passed** (169 files) |
| `@objectstack/metadata-protocol` | **862 passed** (67 files) |
| `@objectstack/spec` | **9464 passed** (362 files) |
| `@objectstack/plugin-reports` | **68 passed** |
| `@objectstack/plugin-auth` | **996 passed** |
| `@objectstack/rest` | **1228 passed** |
| `@objectstack/runtime` | **1870 passed** |

`plugin-reports` and `plugin-auth` are the two packages that forward a caller-supplied sort into `engine.find` directly, so they are the regression surface that matters most here.

> Note on a false alarm worth not repeating: these four consumer suites first appeared to fail wholesale. The cause was unbuilt workspace `dist` (`Failed to resolve entry for package …`), not the refusal — they are green after `pnpm build`.


---

os-zhuang:
## Ratchet fixed at the call sites — and a correction to my earlier gate report

### The correction first

My earlier "all gates green" comment was **not trustworthy, and the fault was mine.** Two of the three batches I ran captured the exit code through a pipeline:

```bash
out=$(timeout 600 pnpm $g 2>&1 | tail -3); code=$?   # ← $? is tail's status, always 0
```

So every gate in those batches reported `ok` regardless of its real result — including `check:query-options-erasure`, which was **already red at that moment**. The gate did not regress after I reported it; I never actually measured it. Apologies for the noise that caused.

Everything has now been re-run with the exit code captured directly (`cmd >log 2>&1; code=$?`). Corrected results are at the bottom.

### The fix — ceiling unchanged, no pin weakened

```
before: test surface grew 249 → 253 site(s)
after:  test surface: 249 site(s) in 49 file(s) — at the ceiling
```

Four sites, fixed per the rule's own prescription. Three were **on-contract** and are now typed:

- **The `it.each` sort table** is now `Array<[string, NonNullable<EngineQueryOptions['orderBy']>]>`. These three sorts are perfectly well-formed `SortNode[]` — it is the *field* they name that the engine refuses, never their shape. An `as any` there would have erased the one channel that enforces `{ field, order }` on a direct engine call, which is precisely the `direction`-vs-`order` defect #4674 is about. The cast was hiding the rule's whole purpose in a test *about* sort correctness.
- **Both `expand` sites** drop the assertion entirely. `EngineQueryOptions.expand` is `Record<string, QuerySchema>`, so the nested `{ orderBy }` was always assignable — the cast was never buying anything.

One is **deliberately off-contract** — the negative pin that smuggles an opt-out flag onto the public options bag — and is now `as unknown as EngineQueryOptions`. That is exactly the case #4918 carved the spelling out for: the assertion's entire subject is that the engine *rejects* the unknown key, so the input cannot be made to typecheck, and the cast now names the contract being bypassed while keeping the rest of the call checked.

Budget number untouched; no pin altered. All 118 conformance tests still pass, `typecheck` green.

### Corrected gate results (exit codes captured directly)

- **47 root `check:*` gates** — all pass except one (below).
- **15 package-filtered gates** — all pass (`@objectstack/spec` × 14 incl. `tsc --noEmit`, plus `@objectstack/lint check:doc-formula-expressions`). These are the ones a bare `pnpm check:*` grep misses.
- **`pnpm lint`** (eslint, `--no-inline-config`) — exit 0.
- **`turbo run typecheck`** — green.

**One pre-existing red, re-confirmed against the current `main` tip:**

`check:platform-checklist` → `coverage.json · qa: UNCLASSIFIED`. Verified by running the same script in a clean worktree at `origin/main` **`db12b88`** (main has moved since my first check at `06be54ec3`; identical failure at both). Nothing in this diff touches `docs/qa/`.

CI is running on `dfa2502`; I will confirm the result rather than infer it.


---

Issue #6628, opened 2026-08-08
[finding][spec] `position.delegatable` JSDoc names a `security-delegatable-admin-position` lint rule that does not exist — the runtime D12 gate is the only enforcer


os-project-manager:
Found during a read-only truth sweep of `packages/spec`&#39;s text surfaces (guidance blocks, `.describe()` strings, tombstones, JSDoc contract prose) against the mechanisms they name. Filed unassigned for triage.

## The defect

`packages/spec/src/identity/position.zod.ts:87-98` — the JSDoc on the authorable `delegatable` key (re-anchored 2026-08-08T13:1xZ after `e0f300ba5`: the named rule now sits at **`:86`**, text unchanged):

```
   * [ADR-0091 D3] Delegation of duty (职务代理). When true, a holder of this
   * position may SELF-SERVICE assign it to a delegate — time-boxed
   * (`valid_until` within the config ceiling), reasoned, dual-audited —
   * WITHOUT being a delegated administrator. Default false: approval-duty
   * positions (an approver going on leave) opt in; admin-ish positions do
   * NOT — delegating administration would bypass the D12 containment gate,
   * so a delegatable position must never distribute an `adminScope`-carrying
   * set (enforced by the `security-delegatable-admin-position` lint rule and
   * the D12 gate). A grant that itself arrived via delegation is not
   * re-delegatable (chains are cut).
```

The parenthetical names **two** enforcers. Only one of them exists.

## The authority

`security-delegatable-admin-position` occurs **exactly once in the repository** — in the sentence quoted above. There is no such lint rule.

The security-domain publish linter&#39;s own rule table is the authority, `packages/lint/src/validate-security-posture.ts:9-21`, and the exported rule-id constants beside it (`:59-70`) are the complete set of twelve — `security-owd-unset`, `security-owd-alias`, `security-external-wider-than-internal`, `security-wildcard-vama`, `security-anchor-high-privilege`, `security-role-word`, `security-book-audience-unknown-set`, `security-private-no-readscope`, `security-master-detail-ungranted`, `security-fls-unqualified-key`, `security-grant-expired-at-authoring`, `security-delegation-missing-reason`. No delegatable/admin-position rule among them.

[...]

PR #7346, opened 2026-08-10
docs(plugin-security): restate the upsertEnvPermissionSet JSDoc for ADR-0094 D5-R (#7082)

### LINKED ISSUE #7082: [finding] `permission-set-projection.ts:332` still states the ADR-0094 direction that D5-R retired — PR #6962 replaced the file-header copy and left the function-level one

os-zhuang:
Fixes #7082

Prose only. No behaviour change, no test-logic change, no gate added.

## The anchor, re-verified at this branch point

The card measured the stale block at `:332` and triage re-confirmed it on `b88f5e8`. My branch point is `origin/main` @ `f3f855ac154f202fa9f1497abd3e0dff60436e1a`, where the block has shifted to **`:323-338`** but is otherwise byte-identical. The card's `:332` still lands inside it, on the `direction confirmed 2026-07-14` line. Quoted verbatim at that SHA:

```
323: /**
324:  * Project a PermissionSet body onto its `sys_permission_set` row from the
...
330:  * #2867 band-aid declined to create). A PACKAGE-OWNED row is also projected —
331:  * an env-scope overlay is the platform's standard customization of a packaged
332:  * definition (ADR-0005; direction confirmed 2026-07-14, reversing the earlier
333:  * refuse-the-env-door rule): the facets update to the EFFECTIVE (overlay-wins)
334:  * body while the `managed_by:'package'` + `package_id` provenance is
335:  * PRESERVED — the row still belongs to the package; the overlay is a
336:  * customization of it, and deleting the overlay resets the row to the shipped
337:  * declaration (the layered read reveals the baseline again).
338:  */
```

The header copy at `:30` already reads "The 2026-07-14 direction confirmation **that used to sit here**… is **RETIRED**", so PR #6962 fixed exactly one of the two copies, as filed.

## The two false claims, and what replaced each

**Claim 1 — "an env-scope overlay is the platform's standard customization of a packaged definition (ADR-0005; direction confirmed 2026-07-14…)".**

Replaced by the bullet beginning **"no new overlay of a packaged set can be minted"**: a metadata write against a code-declared (artifact-backed) set is refused by the producer with 403 `NOT_OVERRIDABLE`, so for those names the body reaching this function is the declared one and the projected facets are the shipped declaration; the supported channel is ADR-0086's (edit the package, re-publish); what survives is the neighbouring `allowRuntimeCreate` tier, which edits the single stored `sys_metadata` definition in place, with no layering and nothing to reset to.

True against: `packages/spec/src/kernel/metadata-plugin.zod.ts:923`, which is now `{ type: 'permission', … supportsOverlay: true, allowOrgOverride: false, allowRuntimeCreate: true, … }` (#6483 / PR #6608); and ADR-0094 D5-R, whose "What survives, stated narrowly so it is not mistaken for a re-route" paragraph is the source of the surviving-neighbour wording — including its explicit "no `delete = reset to the shipped declaration`".

**Claim 2 — "deleting the overlay resets the row to the shipped declaration (the layered read reveals the baseline again)".**

Replaced by the bullet beginning **"delete = reset must NOT be read back into that"**, which carries #6960's measurement as the statement of record: on an environment-scoped kernel `deleteMetaItem` throws `NOT_OVERRIDABLE` / 403 for an artifact-backed target of a non-overridable type *before* it probes for the row, and a kernel with no `environmentId` refuses the same write as `override-artifact` intent — leaving `OS_METADATA_WRITABLE=permission` as the only documented removal. With no overlay to lift (the normal case) the delete is a no-op success and the row keeps projecting the declaration.

True against: #6960 (cited by number in the comment), and ADR-0094 D5-R's own carry-over paragraph plus its "Negative / behavior changes" bullet, both of which say the same thing.

The history is kept rather than deleted — the block names the retired direction, quotes it, and says where the retirement is recorded (ADR-0094 D5-R, #6609 ruling A, #6858) — matching the `:30` header's voice so the two copies read as one author.

## The third-copy sweep

`grep` for the direction's phrasings (`2026-07-14`, `standard customization`, `standard ADR-0005`, `delete = reset`, `resets the row`, `reset target`, `refuse-the-env-door`) across the repo found more copies. Handled by scope:

**Fixed here (same file / same package — 4 more sites):**

- `permission-set-projection.ts:290` — `readDeclaredBody`'s "Used as the reset target when an env overlay is **deleted off** a declared set" is claim 2 in miniature. Now: it is the body for a declared set with no overlay at all, and the reset target when an overlay *is* lifted — which since D5-R means a legacy pre-#6483 row removed through the operator hatch, not a data-door delete.
- `security-plugin.ts:891` — "…translates them into env-scope OVERLAY operations (customize / reset via the standard ADR-0005 layering) — **the boot re-seed can no longer revert an admin's change, because the change lives in the overlay**". The last clause is false for artifact-backed sets since the change never lands. Now states that the middleware translates and ADR-0005 decides.
- `security-plugin.ts:3192` — opened `[ADR-0094, direction confirmed 2026-07-14]` and cited the retired direction as this gate's *reason* for letting `update`/`delete` through. Now `[ADR-0094 D5-R]`, with the refusal named as the producer's.
- the two test rationales that repeated it (`permission-set-projection.test.ts:336`, `security-plugin.test.ts:1989`). Comments and one test title only — no assertion changed.

**Filed, not fixed (outside this package):** see the finding linked in the thread below. It covers `packages/qa/dogfood/test/two-doors-permission.dogfood.test.ts` (header 块2 states both retired halves verbatim while its own body at `:75` already asserts the refusal — literally #7082's header-vs-body split one file over) and `docs/adr/0094-…md:109`.

## Verification

This is prose, so it is verified as prose. Commands and real output are in the thread comment below (`check:adr-anchors`, `check:nul-bytes`, `check:empty-changeset`, and the package's typecheck / lint / build / test).

**Reverse verification, and its honest result.** Prediction, written before running: `check:adr-anchors` **cannot** see this. It is a presence check by its own documentation — the file exists, each registered ADR id names a record under `docs/adr/`, and each id still appears somewhere in the file. Nothing relates *what the file says about an ADR* to *what the ADR currently decides*. Measured: with the JSDoc mutated back to the retired direction verbatim, the gate stays **green**. That is the result, reported as-is; it is the measured reason option B exists, and it is not a failure of the gate, which was never built to catch it. No red/green table is manufactured for a check that has no opinion here.

Option **B** (a #6628-idiom currency pin) is deliberately **not built** — it was not dispatched, and it is a new gate surface.

---

### DIFF
--- a/.changeset/permission-set-projection-d5r-jsdoc.md
+++ b/.changeset/permission-set-projection-d5r-jsdoc.md
@@ -1,0 +1,27 @@
+---
+"@objectstack/plugin-security": patch
+---
+
+Docs: bring the second ADR-0094 copy in `permission-set-projection.ts` up to D5-R.
+
+PR #6962 retired the 2026-07-14 "customize packaged permission sets through an ADR-0005
+env overlay" direction and corrected this file's **header**. A second copy survived in
+the function-level JSDoc of `upsertEnvPermissionSet` — an exported symbol, so the stale
+text ships in the published `.d.ts` and reads as fact to the next author. It stated both
+halves D5-R retired: that an env-scope overlay is the platform's standard customization
+of a packaged definition, and that deleting the overlay resets the row to the shipped
+declaration.
+
+Both are now stated as current: `#6483` (PR #6608) rolled `permission` back to
+`allowOrgOverride: false`, so a metadata write against a code-declared (artifact-backed)
+set is refused by the producer with 403 `NOT_OVERRIDABLE` and the supported channel is
+ADR-0086's (edit the package, re-publish); and `#6960` measures the ordinary delete path
+refusing to lift even a legacy pre-rollback overlay, leaving `OS_METADATA_WRITABLE` as
+the only documented removal — so "delete = reset" is recorded as retired rather than
+restated. The retirement itself is kept in the text, not deleted, so a reader arriving
+at this function does not have to reconstruct the history.
+
+Prose only — no behaviour change. The same retired direction was also corrected in three
+neighbouring comments in this package (the `readDeclaredBody` JSDoc, and the two
+`security-plugin.ts` package-managed write-gate comments) plus their two test rationales,
+so the package no longer states the direction in two voices.
--- a/packages/plugins/plugin-security/src/permission-set-projection.test.ts
+++ b/packages/plugins/plugin-security/src/permission-set-projection.test.ts
@@ -333,9 +333,12 @@ describe('upsertEnvPermissionSet (ADR-0094 — record is a pure projection)', ()
   });
 
   it('projects onto a PACKAGE-OWNED row (overlay customization) while preserving its provenance', async () => {
-    // Direction confirmed 2026-07-14: an env overlay of a packaged set is the
-    // platform's standard ADR-0005 customization — the record follows the
-    // effective body; the package still owns the row.
+    // Unit-level shape, unchanged by ADR-0094 D5-R: handed a body for a
+    // package-owned row, this function projects the facets and preserves the
+    // provenance. What D5-R retired is the CLAIM about where that body comes
+    // from — an env overlay of a packaged set is no longer a supported
+    // customization channel (#6483 / PR #6608); see the D5-R lifecycle block
+    // below for the refusal this projector now sits behind.
     const ql = makeQl();
     ql.permRows.push({ id: 'ps_pkg', name: 'organization_admin', managed_by: 'package', package_id: 'com.example.crm', system_permissions: '["pkg"]' });
     const r = await upsertEnvPermissionSet(ql, envBody());
--- a/packages/plugins/plugin-security/src/permission-set-projection.ts
+++ b/packages/plugins/plugin-security/src/permission-set-projection.ts
@@ -287,8 +287,11 @@ const isProjectionEcho = (v: any): boolean =>
  * Read the DECLARED (artifact) body for a permission set from the engine's
  * SchemaRegistry — the same source `bootstrapDeclaredPermissions` seeds from,
  * and the one store the env projection never writes, so it can't be poisoned
- * by our own registry sync. Used as the reset target when an env overlay is
- * deleted off a declared set.
+ * by our own registry sync. It is the body for a declared set with no overlay
+ * at all, and the reset target when an overlay IS lifted off one — which since
+ * ADR-0094 D5-R means a LEGACY (pre-#6483) row removed through the operator
+ * hatch, not a data-door delete: #6960 measures that delete refusing with 403
+ * `NOT_OVERRIDABLE`.
  *
  * Items tagged `_packageId: 'sys_metadata'` are RUNTIME SHADOWS — hydrated
  * into the registry from overlay rows (loadMetaFromDb / getMetaItems), not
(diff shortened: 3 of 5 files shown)


os-zhuang:
## Verification — real output

All run in a dedicated worktree off `origin/main` @ `f3f855ac154f202fa9f1497abd3e0dff60436e1a`, heavy steps serialized on the shared verify lock with `NODE_OPTIONS=--max-old-space-size=4096`.

### The gate that is registered for this file

```
$ pnpm check:adr-anchors
✓ check-adr-anchors --self-test: 52 assertions over the real auditAdrDirectory() / auditCitedNumbers() / assembleAnchors() paths.
check-adr-anchors: OK (47 anchored file(s), every governing ADR still referenced; 118 decision
number(s), each naming one decision or an allowlisted pair; 20946 citation(s) across 3497 file(s) resolve).
```

**What that gate does check**, from its own header: for each entry under `scripts/adr-anchors/`, the file exists; every ADR id listed for it names a real record under `docs/adr/`; every one of those ids still appears somewhere in the file. Plus, repo-wide, that a cited `ADR-NNNN` resolves and that a number names one decision. The registry entry for this file lists `ADR-0094` / `ADR-0005` / `ADR-0086`, and all three still appear in it after the rewrite.

**What it does not check** — also from its own header, verbatim: *"It does NOT verify the code still obeys the ADR; no static check can. It guarantees the next author is TOLD which decision they are standing on."* It is a presence check. Nothing in it relates *what this file says about ADR-0094* to *what ADR-0094 currently decides*, which is the gap #7082 named.

### Other gates that can bite

```
$ pnpm check:nul-bytes
check-nul-bytes: OK (scanned 6738 text file(s) -- 6737 tracked, 1 untracked-not-ignored;
skipped 5 binary; no raw ASCII control bytes).

$ pnpm check:empty-changeset
✓ check-empty-changeset --self-test: 118 assertions over real temp git repos (real scan() path)
Diffing HEAD from f3f855ac1 (merge base with origin/main).
✓ No empty-frontmatter changeset introduced by this diff (1 declaring changeset(s) added).
```

### Package typecheck / build / lint / test

```
$ pnpm --workspace-concurrency=2 --filter "@objectstack/plugin-security^..." build
… packages/plugins/plugin-sharing build: Done

$ pnpm --filter @objectstack/plugin-security typecheck
> tsc --noEmit
(no diagnostics)

$ pnpm --filter @objectstack/plugin-security build
DTS ⚡️ Build success in 5212ms
DTS dist/index.d.mts 757.60 KB
DTS dist/index.d.ts  757.60 KB

$ npx eslint --no-inline-config <the 4 changed files>
(clean, exit 0)

$ pnpm --workspace-concurrency=2 --filter @objectstack/plugin-security test -- --maxWorkers=2
 Test Files  43 passed (43)
      Tests  878 passed (878)
   Duration  14.70s
```

The changeset bump is `patch`, and it is warranted rather than skippable: `upsertEnvPermissionSet` is exported from the package index, so the JSDoc is emitted into the published declaration file. Confirmed on the built artifact —

```
$ grep -n "delete = reset" packages/plugins/plugin-security/dist/index.d.ts
16017: *  - **"delete = reset" must NOT be read back into that.** #6960 measures the
```

— i.e. the stale text was shipping to consumers' editors, and the corrected text now ships instead.

### Reverse verification — prediction, then measurement

**Predicted, before running** (also stated in the PR body): `check:adr-anchors` cannot see this at all, so a mutation that makes the prose false again leaves it **green**. There is no before-red/after-green table to draw, because no check in this repo has an opinion on whether this paragraph is current — that is the finding, not a gap in the run.

**Mutation** — the fix taken out with `git checkout origin/main -- <path>` (never `git stash`; the stash stack is shared across every worktree of this repo), restoring the retired-direction text byte for byte:

```
$ git checkout origin/main -- packages/plugins/plugin-security/src/permission-set-projection.ts
$ grep -n "direction confirmed 2026-07-14, reversing the earlier" …/permission-set-projection.ts
332: * definition (ADR-0005; direction confirmed 2026-07-14, reversing the earlier
```

**Measured under the mutation:**

```
$ pnpm check:adr-anchors
✓ check-adr-anchors --self-test: 52 assertions …
check-adr-anchors: OK (47 anchored file(s), every governing ADR still referenced; 118 decision
number(s), each naming one decision or an allowlisted pair; 20941 citation(s) across 3497 file(s) resolve).
adr-anchors exit=0

$ npx eslint --no-inline-config …/permission-set-projection.ts
eslint: green under the mutation
```

Green, as predicted. The only thing that moved is a **count** — `20946` citations with the fix, `20941` without, because the rewrite adds five ADR references — and a count is not a verdict: the gate reports OK either way. So the file can state the exact direction ADR-0094 retired and every registered check stays green, which is the measured reason #7082's disposition **B** exists. Option B was **not** dispatched and is **not** built here; recording the measurement is the deliverable.

The tree was restored with `git checkout <branch> -- <path>` and verified clean (`git status --short` empty) before pushing.

### Out-of-scope finding filed

Filed **#7351** (`finding`, unassigned, no `domain:*` — routing is triage's): two more copies of the same retired direction survive **outside** `packages/plugins/plugin-security`, so they are deliberately not touched here — `packages/qa/dogfood/test/two-doors-permission.dogfood.test.ts:12-17` (header states both retired halves verbatim while its own body at `:75` already asserts the refusal — #7082's header-vs-body split one file over) and `docs/adr/0094-…md:109` (a D2 cross-reference restating the direction as live, pointing at a section whose own title says RETIRED).

---

Issue #7256, opened 2026-08-10
qa(core): the `contains` assertion SILENTLY PASSES when the actual value is neither an array nor a string


os-zhuang:
Recorded while implementing #6247 (PR #7255 — seeding `packages/spec/liveness/qa.json` and enforcing `TestSuiteSchema` at the `os test` load site). Not claiming or fixing here — filing per Prime Directive #10, and deliberately kept out of that PR's scope: the ruling there covers the ledger and the load site, and this is a behaviour change inside the assertion engine.

**Finding.** `TestRunner.assert` (`packages/core/src/qa/runner.ts:171-177`) handles `contains` like this:

```ts
case 'contains':
   if (Array.isArray(actual)) {
       if (!actual.includes(expected)) throw new Error(…);
   } else if (typeof actual === 'string') {
       if (!actual.includes(String(expected))) throw new Error(…);
   }
   break;
```

There is no `else`. When `actual` is anything else — `undefined` (the overwhelmingly common case: a typo'd `field` path, or a response shape that changed), `null`, a number, or an object — the branch falls through, throws nothing, and **the assertion passes**.

So a scenario asserting `{ field: "body.data.items", operator: "contains", expectedValue: "acme" }` against a response that has no `body.data.items` at all reports ✅. The assertion that was supposed to be the test *is* the thing that silently disappears.

**Why this one matters more than a normal bug.** Every other unhandled shape in this engine fails LOUD, and that asymmetry is what makes this the odd one out:

- an action type with no adapter branch (`run_script`) throws `Unsupported action type in HttpAdapter` (`http-adapter.ts:38`);
- an operator with no branch (`not_contains` / `gt` / `gte` / `lt` / `lte` / `error`) throws `Unknown assertion operator` (`runner.ts:186`) — declared in `TestAssertionTypeSchema`, refused at runtime, which is annoying but honest;
- `is_null` / `not_null` / `equals` / `not_equals` all compare unconditionally.

`contains` is the only path that can decide "no comparison applies here" and report success. A test framework whose failure mode is a false green is worse than one that is missing the feature — the whole point of the surface is to be believed by CI.

This is also the reason it is worth a separate card rather than a note: #6247's ledger records `scenarios.setup`'s sub-keys with `assertions` → `field` / `operator` / `expectedValue` all `live`, and they are. The KEY is live; it is one VALUE of `operator` that is unsound. That distinction is real (the `api.json` `type` precedent) but it means no liveness verdict will ever surface this.

[...]

Issue #7223, opened 2026-08-10
Nothing pins `register(type, name, d)` → `get(type, name)` across the shipped `IMetadataService` implementations — the hole #6725 fell through


os-zhuang:
Observation-class finding, filed per Prime Directive #10 while implementing #6725 (PR #7211). Unassigned, deliberately **not** queued. This is a test-coverage gap, not a defect in shipped behaviour — filed separately rather than folded into #6725 because the fix is a conformance suite, not a line of engine code.

## The fact (read on `origin/main` @ `55da611`)

`IMetadataService.register` and `.get` are the contract's first two CRUD members, and the round-trip between them is exercised in exactly one place:

`packages/spec/src/contracts/metadata-service.test.ts` — "should register and retrieve metadata items asynchronously":

```ts
const store = new Map<string, Map<string, unknown>>();
const service: IMetadataService = {
  register: async (type, name, data) => { …store.get(type)!.set(name, data); },
  get: async (type, name) => store.get(type)?.get(name),
  …
};
await service.register('object', 'account', objectDef);
expect(await service.get('object', 'account')).toEqual(objectDef);
expect(await service.getObject('account')).toEqual(objectDef);
```

Every subject in that file is an **inline literal written in the test itself**. The suite asserts that a `Map` behaves like a `Map`. It is a type-conformance file — which is the right thing for `packages/spec` to host, since the contract has no runtime and spec is the dependency root — but it means the round-trip is pinned against **no shipped implementation at all**.

The cross-implementation pin that does exist,
`packages/objectql/src/metadata-service-getobject-equivalence.test.ts` (#6745), covers a *different* proposition — `getObject(name)` ≡ `get('object', name)` — and seeds every subject through `registry.registerObject` / `manager.register`, explicitly **not** through the facade's own write. Its header says so, and says why: seeding the facade the other way produced `undefined` on both members.

So: no test anywhere asserts that writing through an implementation's `register` makes the item readable through that same implementation's `get`.

## Why this is worth a card

That is precisely the hole #6725 fell through. `MetadataFacade.register('object', …)` wrote into a map none of its own reads consult; every read answered `undefined`; the full `packages/objectql` suite (167 files, 2917 tests) stayed green, and so did all 64 `lint.yml` gates. A shipped, exported implementation of the platform's central metadata contract could not perform its own most basic round-trip, and nothing in the repo was positioned to notice.

[...]

Issue #7204, opened 2026-08-10
plugin-reports' `executeReport` rebuilds a 5-field subset of the execution envelope, dropping `accessible_org_ids` — a `group`-posture report silently under-reports


os-help:
Out-of-scope observation from implementing #7135, filed rather than fixed — that card was bounded to annotation widening plus redundant-cast deletion, and the change below is a RUNTIME change with an enforcement consequence. This is the plugin-reports twin of #7141 (the plugin-audit one), same shape, different consequence.

## Fact

`packages/plugins/plugin-reports/src/report-service.ts` — `executeReport()` does not forward the caller's execution envelope to the engine read that produces the report. It rebuilds a five-field projection of it:

```
    const rows = await this.engine.find(report.object_name, {
      where: q.filter,
      fields: q.fields,
      orderBy: q.orderBy,
      limit,
      // Reports execute with the caller's identity so sharing rules
      // (if installed) apply. Falls back to system bypass only when
      // the report definition was created by a system writer.
      context: {
        userId: context.userId,
        tenantId: context.tenantId,
        positions: context.positions ?? [],
        permissions: context.permissions ?? [],
        isSystem: context.isSystem ?? false,
      },
    });
```

That comment states the intent exactly — the report must see the rows the caller would see interactively — and the projection is what stops it. #7135 widened this method's parameter ANNOTATION to the full `ExecutionContext` (so `context` here now genuinely holds the whole envelope); the body was deliberately left alone, because changing it is not inert.

## The field that matters: `accessible_org_ids`

`packages/objectql/src/engine.ts:2181` reads it straight off the execution context to widen the driver's native tenant scope:

```
    if (hasTenant && opts.tenantIds === undefined && this.tenancyPostureProvider?.() === 'group') {
      const set = (execCtx as any)?.accessible_org_ids;
      if (Array.isArray(set) && set.length > 0) {
        opts.tenantIds = set.map(String);
      }
    }
```

`accessible_org_ids` is not in the projection, so `opts.tenantIds` is never set for a report. Per the ADR-0105 D2 comment immediately above that block, an absent set makes drivers "fall back to equality: fail toward isolation, never toward exposure".

[...]

PR #7352, opened 2026-08-10
refactor(core,plugin-audit,service-storage,plugin-reports): give the `__` operation-private-key convention a single owner (#7284)

### LINKED ISSUE #7284: `withoutOperationPrivateKeys` is now hand-copied into three packages — the `__` convention has no single owner
Observation from implementing #7204 (PR #7283), filed rather than fixed — that card's file surface is `plugin-reports`, and this spans three packages plus wherever the shared home would be.

## Fact

The same ~10-line helper, with the same name and the same doc-block reasoning, now exists in three copies:

- `packages/plugins/plugin-audit/src/comment-access-hooks.ts:215` (#7141 / PR #7143 — the original)
- `packages/services/service-storage/src/attachment-access-hooks.ts:103` (#7145 / PR #7207 — copy 2)
- `packages/plugins/plugin-reports/src/report-service.ts` (#7204 / PR #7283 — copy 3)

Each declares its own `const OPERATION_PRIVATE_KEY_PREFIX = '__'` and its own `function withoutOperationPrivateKeys(...)`. None is exported; every copy was written by hand from the previous one.

## Why it is worth recording

The helper encodes a **cross-package convention**, not a local detail: `__`-prefixed keys on an execution context belong to the operation plugin-security has in flight, and a consumer forwarding an envelope elsewhere must drop them. The producer of that convention is `packages/plugins/plugin-security` (`security-plugin.ts` — `sc.__readScope = …`, `__writeScope`, the ADR-0090 D10 delegator halves, plus the engine's `__expandRead` / `__referentialFieldClear` markers). The rule about how to consume it lives in three copies in three *other* packages, and the package that produces the keys owns none of them.

Consequences, all latent today:

- A fourth consumer (any service that forwards a caller envelope to a gate or a read) is written by copy-paste or not at all, and nothing points a new author at the convention.
- If the middleware ever stamps something operation-private **without** a `__` prefix, three files have to change and there is no list of them.
- Each copy's doc block reasons about its own package's object (`sys_comment`, `sys_attachment`, `report.object_name`), so the copies have already diverged in prose while agreeing in code — the shape that makes a later divergence in *code* hard to notice.

## Why observation-class, not a defect

Nothing a user hits: all three copies are byte-equivalent in behaviour today, and each is covered by its own package's tests. This is a structural/ownership finding, so it carries `finding` and no `pm:queue` — grading is triage's call.

## What I could not establish

Where the shared home should be. The candidates are not equivalent and the choice is a real one: `@objectstack/spec` (a constant + a pure function, but Prime Directive #2 says no business logic there — a key-filtering function may or may not count), `@objectstack/core`, or an export from `plugin-security` itself (the producer, which is the most honest owner but makes three packages depend on the security plugin for a string-prefix filter). Worth a decision rather than a guess.

## Dedup

Searched open issues for `withoutOperationPrivateKeys`, `__readScope` + prefix/strip, and operation-private + envelope + helper/duplicate: no hits. #7141, #7145 and #7204 are the three landing sites and each is scoped to its own package; none of them covers the extraction.

Related: #7141, #7145, #7204, #6523, #6206 (ruling).

---

os-zhuang:
Closes #7284.

`withoutOperationPrivateKeys` — the rule that a consumer forwarding a caller's execution envelope to a question about a **different** object must first drop the `__`-prefixed keys plugin-security stamped for the operation in flight — existed in three unexported, hand-copied copies. This gives it one owner.

## The fact, re-verified on this base

Confirmed at `f3f855a` (triage measured at `62b6a2f`), located by content rather than by line number:

| copy | file | landed |
|---|---|---|
| original | `packages/plugins/plugin-audit/src/comment-access-hooks.ts` | #7141 |
| copy 2 | `packages/services/service-storage/src/attachment-access-hooks.ts` | #7145 |
| copy 3 | `packages/plugins/plugin-reports/src/report-service.ts` | #7204 |

**Sweep for a fourth: none.** `withoutOperationPrivateKeys` / `OPERATION_PRIVATE_KEY_PREFIX` across all `.ts`/`.tsx`/`.md` outside `node_modules` returns exactly those three files. A second sweep for hand-rolled near-misses under other names (`startsWith('__')`, `__readScope` / `__writeScope` handling) found only the *producer* side — `plugin-security`, `plugin-sharing` — which stamps and reads these keys rather than stripping them, and is unaffected.

## Where the single owner lives, and why

Measured from the three consumers' `package.json` dependency blocks:

| candidate | new dependency edges | verdict |
|---|---|---|
| **`@objectstack/core`** | **0** — all three already depend on it | **chosen** |
| `@objectstack/spec` | 0 — all three already depend on it | rejected: Prime Directive #2 fences business logic out of spec, and a key-filtering function with a security argument attached is not a schema |
| `plugin-security` (the producer) | **3** — none of the three depends on it today | rejected: the most *honest* owner, but a string-prefix filter does not justify three new dependency edges onto a plugin, and it would give the producer reverse edges from its own consumers |
| a new package | n/a | rejected: ten lines |

Inside core the home is `src/security/operation-private-keys.ts`, deliberately beside `assemble-execution-context.ts`. Those two own opposite ends of one lifecycle: #6216 made that file the single place an `ExecutionContext` is **built** at a transport entry point; this is the single place one is **stripped back down** before being forwarded. Exported from `src/security/index.ts`, which `src/index.ts` already re-exports wholesale.

⛔ Nothing lands in `packages/objectql/src/engine.ts` (#7095), `packages/core/src/qa/runner.ts` (#7256), or `packages/spec/**` / `packages/metadata` test files (#7223). **`packages/spec` is untouched** — no cross-seat declaration needed.

## The reasoning moved with the code

Not three thinner copies. The shared home carries the full argument, generalised out of the three divergent prose versions: what a `__` key is and which ones exist today (the ADR-0057 D1 depths, the ADR-0090 D10 delegator halves, the engine's `__expandRead` / `__referentialFieldClear` privilege markers); why a consumer must drop them (one object's widening applied to another object's question — including the concrete survival path, that plugin-security only *overwrites* `__readScope` when it actually resolves permission sets for the new object); why dropping is safe in the one direction that matters; why by **prefix** and never by a name list, plus the ⛔ corollary for whoever next changes the middleware; and why the fresh copy is load-bearing in both directions.

Each consumer keeps only its genuinely **local** half — which object *its* gates ask about (`sys_comment`, `sys_attachment`, `report.object_name`) — and points at the shared home. The `[#7141]` / `[#7145]` / `[#7204]` provenance markers and the surrounding `#6206`/`#6523` envelope-forwarding arguments at the call sites are untouched.

## Pins

- **`operation-private-keys.test.ts`** — the rule's own behaviour, which no package-level test had ever asserted directly (each package tested only its own gates' effect): prefix-dropping, that it strips rather than *projects* (every principal field survives — the defect half of the five-field projections these call sites replaced), the fresh copy even on a clean envelope, the deliberate shallow-copy boundary, and that a key the middleware has not stamped yet is dropped by prefix alone.
- **`operation-private-keys.pin.test.ts`** — the anti-fourth-copy pin. Scans every `.ts`/`.tsx` under `packages/` for a *declaration* (not a mention) of either symbol outside the home, and fails with a message naming the import to use. Two anti-vacuity guards travel with it (#4690's "a check that can't see its input has verified nothing"): the home must really declare both symbols, and the walker must demonstrably reach all three former-copy files.

  Scope is deliberately narrow — it detects **redeclaration**, not "a consumer that should have used the helper and didn't". The latter is the more interesting question and is not decidable by scanning; a regex ambitious enough would be a false-red generator, and an inert or noisy gate reads as a gate that is watching. Redeclaration is exactly how all three copies got here.

  **Reverse-verified**: planting a fourth copy in `plugin-audit` turns it red naming that file; removing it turns it green.

## Changeset

Real patch bumps for the four packages, **not** an empty changeset and **not** `skip-changeset`, read off `check-empty-changeset.mjs`'s own ledger. That gate's text is one-directional: an empty changeset produces no CHANGELOG entry (measured, #5471), buys nothing the label doesn't, and uniquely is a real input to `changesets/action` that can trigger #4898 — the failure mode that silently stalled 17.0.0-rc.2. This PR changes published source in four packages, so a real bump is both honest and the route that produces an actual CHANGELOG entry.

Nothing breaking, so no ADR-0087 conversion.

## Gates

Enumerated fresh off `origin/main` — 64 `check:*` in `lint.yml`, 70 across all workflows.

The export-surface family (`check:export-origins`, `check:api-surface`, `check:exported-any`, `check:dual-source-exports`) is **`--filter @objectstack/spec`** in every workflow invocation — it reads spec's surface, not core's. Adding an export to `@objectstack/core` therefore required nothing of them. Run results in a follow-up comment.

## Verification

- `@objectstack/core` — 738 tests green, including the 11 new ones
- `plugin-audit` 160 · `service-storage` 352 · `plugin-reports` 68 — all green, unchanged

---
_Generated by [Claude Code](https://claude.ai/code/session_017AD2nx7MRuje3kLqoLBHPM)_

### DIFF
--- a/.changeset/operation-private-keys-single-owner.md
+++ b/.changeset/operation-private-keys-single-owner.md
@@ -1,0 +1,39 @@
+---
+"@objectstack/core": patch
+"@objectstack/plugin-audit": patch
+"@objectstack/service-storage": patch
+"@objectstack/plugin-reports": patch
+---
+
+refactor(core,plugin-audit,service-storage,plugin-reports): give the `__` operation-private-key convention a single owner (#7284)
+
+`withoutOperationPrivateKeys` — the rule that a consumer forwarding a caller's
+execution envelope to a question about a DIFFERENT object must first drop the
+`__`-prefixed keys plugin-security stamped for the operation in flight — had been
+hand-copied into three packages: `plugin-audit`'s comment access hooks (#7141),
+`service-storage`'s attachment access hooks (#7145) and `plugin-reports`' report
+service (#7204). Each carried its own `OPERATION_PRIVATE_KEY_PREFIX` and its own
+doc block, and the prose had already diverged while the code still agreed — the
+shape that makes a later divergence in behaviour hard to notice.
+
+The helper now lives once, in `@objectstack/core`
+(`security/operation-private-keys.ts`), exported from the package root. Core is
+the only candidate all three consumers already depend on: `plugin-security` is
+the producer of the convention and the most honest owner, but none of the three
+depends on it and a string-prefix filter does not justify three new dependency
+edges onto a plugin; `@objectstack/spec` is fenced off by Prime Directive #2. The
+new home sits beside `assemble-execution-context.ts`, which owns the other end of
+the same lifecycle — that file is where an `ExecutionContext` is built at a
+transport entry point, this one is where it is stripped back down before being
+forwarded.
+
+The full reasoning moved with the code rather than being thinned: which keys the
+middleware stamps and why each is a widening input, why they are dropped by
+PREFIX and never by a name list, and why the fresh copy is load-bearing in both
+directions. Each consumer keeps only its own local half — which object *its*
+gates actually ask about — and points at the shared home.
+
+No behaviour change: the three copies were byte-equivalent, and all three
+packages' suites pass unchanged. Two new pins at the home cover it — the rule's
+own behaviour, which no package-level test had ever asserted directly, and a
+repository-shape pin that turns red if a fourth file declares its own copy.
--- a/packages/core/src/security/index.ts
+++ b/packages/core/src/security/index.ts
@@ -131,3 +131,11 @@ export {
 
 // ADR-0091 D1/D2 — grant validity windows, the shared resolution-time predicate.
 export { isGrantActive, isGrantExpired, type GrantValidityWindow } from './grant-validity.js';
+
+// #7284 — the `__` operation-private-key convention, the CONSUMER half of the
+// ExecutionContext lifecycle `assemble-execution-context.ts` opens. One owner
+// for the rule three packages had hand-copied (#7141 / #7145 / #7204).
+export {
+  OPERATION_PRIVATE_KEY_PREFIX,
+  withoutOperationPrivateKeys,
+} from './operation-private-keys.js';
--- a/packages/core/src/security/operation-private-keys.pin.test.ts
+++ b/packages/core/src/security/operation-private-keys.pin.test.ts
@@ -1,0 +1,127 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#7284] The `__` operation-private-key convention has exactly ONE owner.
+ *
+ * This is the pin the extraction is worth having. The three copies #7284 found
+ * were byte-equivalent in behaviour and each was covered by its own package's
+ * tests, so nothing in the repository went red while the rule was being copied
+ * by hand a third time — the finding was made by a human reading three diffs
+ * months apart. A fourth consumer is written the same way the first three were:
+ * by opening the nearest existing one and copying the block out of it. Extracting
+ * the helper without pinning it just resets that counter to one.
+ *
+ * So the assertion is about the SHAPE of the repository, not about behaviour: no
+ * file outside this module may declare its own `OPERATION_PRIVATE_KEY_PREFIX` or
+ * its own `withoutOperationPrivateKeys`. A fourth author who copies the block
+ * turns this red the first time they run the suite, with a message naming the
+ * import to use instead.
+ *
+ * ⛔ Scope, deliberately narrow — this pin does NOT try to detect "a consumer
+ * that should have used the helper and did not". That is the interesting
+ * question and it is not decidable by scanning: forwarding an envelope is
+ * spelled a dozen ways, and a regex ambitious enough to catch them all would be
+ * a false-red generator, which is worse than the gap (an inert or noisy gate
+ * reads as a gate that is watching — `validate-security-posture.ts`'s hazard).
+ * What IS decidable is redeclaration, which is exactly how all three copies got
+ * here.
+ *
+ * Reworded freely: the pin matches DECLARATIONS, not mentions. Documentation,
+ * comments and tests may name either symbol as much as they like.
+ */
+
+import { readFileSync, readdirSync, statSync } from 'node:fs';
+import { dirname, join, relative, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+import { describe, it, expect } from 'vitest';
+
+const HERE = dirname(fileURLToPath(import.meta.url));
+/** …/packages/core/src/security → repo root */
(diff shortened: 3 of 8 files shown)


os-zhuang:
## Gate runs (local, on the pushed head)

Gates enumerated fresh off `origin/main` — 64 `check:*` in `lint.yml`, 70 across all workflows. CI is the authority; these are the local runs, all exit 0.

**Changeset family** — the decision in the PR body, checked against the gates that judge it:

| gate | result |
|---|---|
| `check:empty-changeset` (+ `--self-test`, 118 assertions) | ✓ "No empty-frontmatter changeset introduced by this diff (1 declaring changeset(s) added)" |
| `check-changeset-no-major` | ✓ "introduces no `major` bump" |
| `check-adr-0087-registration` | ✓ "adds no declared-breaking changeset (1 non-breaking changeset(s) seen)" |

**Export-surface family** — the §4 question, answered: **they required nothing of this PR.** Every workflow invocation of these four is `pnpm --filter @objectstack/spec`, so they read `@objectstack/spec`'s surface, not `@objectstack/core`'s. Adding an export to core does not enter their corpus. Run anyway to confirm no collateral movement:

| gate | result |
|---|---|
| `check:export-origins` | ✓ 4979 exports across 16 entry points resolve exactly as recorded |
| `check:api-surface` | ✓ spec public API surface + factory signatures unchanged |
| `check:exported-any` | ✓ 2403 types + 1506 schemas, none resolves to `any` |
| `check:dual-source-exports` | ✓ 4809 names, 0 new dual-source |
| `check:spec-changes` | ✓ `spec-changes.json` up to date |
| `check:generated` | ✓ all 11 generated artifacts up to date |
| `check:strictness-ledger` | ✓ ledger current |

**Security- and packaging-adjacent gates** that could plausibly read a new file under `packages/core/src/security/`:

| gate | result |
|---|---|
| `check:authz-resolver` | ✓ single shared authorization resolver intact; both entry points delegate |
| `check:tenant-chokepoint` | ✓ 19 `getBuilder()` bindings, every read builder routes through `applyTenantScope()` |
| `check:published-files` | ✓ 69 publishable packages declare a `files` whitelist admitting no test file — the two new `*.test.ts` at the home are correctly excluded from the published tarball |
| `check:type-check-coverage` | ✓ 63/77 packages type-checked; this PR adds no debt row |
| `check:nul-bytes` | ✓ 6741 files scanned |

**Build / lint / tests**

- `eslint --no-inline-config` over all 7 changed source files — clean
- `@objectstack/core` tsup build incl. DTS — success. `plugin-audit` / `plugin-reports` `tsc --noEmit` — clean. `core` and `service-storage` declare no `typecheck` script; their type safety is the DTS build, green for both.
- `@objectstack/core` **738 tests** green (11 new) · `plugin-audit` **160** · `service-storage` **352** · `plugin-reports` **68** — all green and unchanged in count apart from core's additions.

**Anti-copy pin, reverse-verified.** Planting a fourth copy in `packages/plugins/plugin-audit/src/` turns `operation-private-keys.pin.test.ts` red naming that exact file and printing the import to use instead; removing it turns it green. The pin is watching, not inert.

---

PR #7306, opened 2026-08-10
feat(objectql,metadata-protocol)!: register a tenant object overlay as its own contributor layer (ADR-0029 D9)

os-zhuang:
Fixes #7277

Implements ADR-0029 **D9**, ratified when PR #7087 merged (`fb5ebc5df`). The design is settled; this is the code.

## The correctness check: the resolved object does not move

D9's load-bearing property is that layering must resolve **bit for bit** to what the old splice already produced, `_provenance: 'org'` included. The pin compares two registries directly:

- left: D9 — packaged `own` + tenant `overlay`
- right: pre-D9 — the overlay body as the sole `own` (what the splice left behind)

`expect(d9.getObject(name)).toEqual(preD9.getObject(name))` (`registry-object-overlay-layer.test.ts`). **No resolved schema moved anywhere.** RV-3 below proves the pin does real work: reverting the provisioning gate turns it red, which is exactly what "the resolved object moved" looks like.

## The three things a naive implementation gets wrong — each with its own pin

**1. `getArtifactItem`'s object branch reads the OWNER contributor.** The merged body deliberately keeps `_provenance: 'org'`, so only reading the owner's layer makes `isArtifactBacked` honest. What comes back is the **code-layer baseline** (owner + its extenders); with no overlay registered it is byte-for-byte the previous answer (pinned separately). RV-1: reverting this branch to the merged read produces 6 stable reds.

**2. `provisionPrimary` / `provisionSearchCompanion` are gated on "is this a BASE layer"** (`own` or `overlay`). RV-3: reverting to `own`-only produces 2 reds — the `nameField` pin and the bit-for-bit pin.

**3. #6995's three bindings, implemented as ruled.** Same package normal; package-less sentinel **accepted** (the old throw was an artefact of the borrowed `own` slot); different package **refused loudly at the producer** with the newly minted `OBJECT_OVERLAY_PACKAGE_MISMATCH` / 422, and counted in `loadMetaFromDb`'s per-record `errors` at boot.

## The two items D9 left open

[...]

Issue #6928, opened 2026-08-09
finding: `GET /api/v1/notifications?limit=abc` coerces to NaN and flows through listInbox's clamp into `data.find({ limit: NaN })` — unvalidated query coercion on the notifications route


os-zhuang:
Filed by the `domain:spec` seat PM on behalf of the #6361 dev (session ran out of runway post-quota-kill; reported in the #6361 final report with the exact mechanism, reproducible). Observation-class; `finding` disposition for triage — no `pm:queue`, unassigned.

Mechanism (anchors as of the #6866 base): `packages/runtime/src/domains/notifications.ts:108` does `const limit = query?.limit ? Number(query.limit) : undefined`, so a raw HTTP caller sending `?limit=abc` yields `NaN`; `MessagingService.listInbox` then computes `Math.min(Math.max(NaN, 1), 200)` = `NaN` and passes it straight into `data.find({ limit: NaN })`. Driver-dependent behavior from there; never a 400. The typed SDK cannot produce this (its `limit` is `number`), so exposure is raw-HTTP callers only.

Same route and same declared≠enforced family as #6361, but a behavior change (validation) was explicitly out of #6866's ruled scope (wire behavior pinned unchanged there), so it needs its own card. Any fix should decide refusal shape per ADR-0112 rather than ad-hoc clamping. Dedup: searched open issues for `limit NaN notifications` — no hits.

os-project-manager:
Triage (maintainer-authorized one-off pass, 2026-08-09, registered on #6015): **promoted `finding` → `pm:queue`** (`domain:cli` kept). Unvalidated query coercion reaching `data.find({ limit: NaN })` with driver-dependent behaviour and never a 400 is a defect, not an observation — raw-HTTP exposure is exposure. Deliverable per the filing: a proper ADR-0112 refusal shape for non-numeric `limit` (and the same probe applied to this route's other coerced params while in there), not an ad-hoc clamp.

---
_Generated by [Claude Code](https://claude.ai/code)_

os-help:
Claim: PM loop round 2 (cli lane)
Session: `session_0158ZQo7LiHSxGWpYKuPq1wu`
Branch: `claude/issue-6928-notifications-limit-nan`
Worktree: `objectstack-issue-6928`
Domain: `domain:cli`
[...]

Issue #6877, opened 2026-08-09
`packages/rest` 的其它 `req.query.*` 读取点同样把 `string | string[]` 当字符串用(#6307 的未扩大部分)


os-project-manager:
观察类发现,来自 #6307(PR 修的是 `package-routes.ts` 两个 handler)。按 Prime Directive #10 单独立单,**未认领**,#6307 的 PR 刻意没有扩大范围。

**今天没有用户会撞到**:需要客户端重复传同一个查询参数才会触发,而生产用的 Hono 适配器在 handler 之前就把重复值折叠成第一个值(见姊妹单)。

## 事实

`IHttpRequest.query` 的契约是 `Record<string, string | string[]>`(`packages/spec/src/contracts/http-server.ts`)。`packages/rest/src` 里非测试的 `req.query` 读取点共 **61** 处(不含 `package-routes.ts`),其中只有 **9** 处用 `typeof req.query?.x === 'string'` 做了守卫 —— 那 9 处是安全的:数组会落到 `undefined`,行为退化但不会把数组喂进下游。

其余读取点按可观察后果分三类(行号为本单撰写时的 `main`,已知会漂移,**按内容定位**):

**1. 数组原样流进下游**(与 #6307 同形)
- `rest-server.ts` `const layeredPackageId = req.query?.package || undefined;`
- `rest-server.ts` `const packageId = req.query?.package || undefined;`(4 处)
- `rest-server.ts` `const forceRaw = req.query?.force;` / `const packageRaw = req.query?.package;`(2 处)
- `rest-server.ts` `req.query?.toVersion` / `parseV(req.query?.from ?? req.query?.fromVersion)`
- `rest-server.ts` `const { select, expand } = req.query || {};`
- `rest-server.ts` `object: req.query?.object`

**2. 被 `String()` 静默拼接**(`['a','b']` → `"a,b"`,一个不存在的名字)
- `String(req.query.object)`、`String(req.query?.q ?? req.query?.query ?? '')`、`String(req.query?.status)`、`String(req.query?.packageId)`

**3. 被 `Number()` 变成 `NaN`**
- `Number(req.query.sinceSeq)`、`Number(req.query.limit)`(2 处)、`Number(req.query?.perObject)`

## 为什么没有被门禁抓到

[...]

PR #7386, opened 2026-08-10
fix(metadata-protocol): check query-parameter arity in findData's list-query normalizer (#7321)

### LINKED ISSUE #7321: `findData`'s list-query normalizer coerces repeated query parameters without checking arity (the half #6877 could not reach)
Observation-class finding, split out of #6877 while implementing it. **Unassigned**, filed per Prime Directive #10. #6877's PR deliberately did not widen into this — it is in a different package and a different owner's contract.

## Why it is a separate card and not part of #6877

#6877 swept `packages/rest/src/rest-server.ts`'s query read points and declared, per handler, which parameters are single-valued. **One route was deliberately left ungated**: `GET /api/v1/data/:object` does not read parameters at all — it hands the WHOLE query record through:

```ts
const result = await p.findData({
    object: req.params.object,
    query: req.query,
    ...
});
```

Every parameter's arity for that route is therefore decided by the shared list-query normalizer in `packages/metadata-protocol/src/protocol.ts`, not by `packages/rest`. Declaring an arity list at the REST layer would have been one package guessing at another's contract, and the normalizer is genuinely the right home: `GET /data/:object`, `POST /data/:object/query` and the runtime dispatcher all flow through it, which is exactly the reason #4181's filter rejection was put there rather than copied per route.

## The fact

`IHttpRequest.query` is `Record< string, string | string[] >` and the array arm is produced by a real first-party adapter — `NodeHttpServer` hands `?x=1&x=2` through as `['1','2']`, measured over a socket on #6878. The normalizer coerces without checking the arity it was handed. Read directly, `packages/metadata-protocol/src/protocol.ts`:

```ts
if (options.limit != null) options.limit = Number(options.limit);
if (options.offset != null) options.offset = Number(options.offset);
```

`Number(['1','2'])` is `NaN`, so `?$top=1&$top=2` reaches the driver as `limit: NaN` — the same shape #6928 / PR #7299 just fixed one layer over on `GET /api/v1/notifications`, where `NaN` survived the clamp and landed in `data.find({ limit: NaN })`.

That is one measured line, not a survey. The survey is the work: this normalizer also folds four spellings of the filter slot, a large alias table (`pageSize` / `perPage` / `take` / `first` / … all rewrite to `$top`), the `$`-alias consumption pass, and the leftover-key bucket that lowers unknown keys into field-equality predicates. Each of those needs the same per-parameter single-vs-multi judgement #6877 made for the REST layer, and some of them are genuinely multi-valued (`$select`, `$expand`, `$searchFields` all already accept the array arm on purpose).

## Not live today, and exactly why that is temporary

No user hits this now: it takes a client that repeats a parameter, and the production Hono adapter collapses repeats to the first value before any handler runs. **#6878's route 2 — ruled adopted on 2026-08-10 — removes that collapse.** So the dormancy here has the same expiry date the #6877 surface had, and the same reasoning applies: this is the prerequisite for a decided change, not speculative hardening.

Not filed as a sub-issue of #6877: its fix lands outside #6877's completion scope (different package, different contract owner). Depends on nothing, but its priority is coupled to **#6878 route 2**.

## Dedup

Searched open issues for `req.query`, `findData` + normalizer, `metadata-protocol` + query param, and `string[]`. Hits: #6307 (the origin, two handlers in `package-routes.ts`), #6877 (the `packages/rest` sweep, this card's parent in origin only), #6878 (the adapter divergence). None covers the `findData` normalizer.

---

os-zhuang:
Fixes #7321

`findData`'s shared list-query normalizer coerces query parameters without ever checking the **arity** it was handed. The work on this card is not that one line — it is the **per-parameter survey** around it. The same normalizer also folds four spellings of the filter slot, a large alias table, the `$`-alias consumption pass, and the leftover-key bucket that lowers unknown keys into field-equality predicates. Each needed the single-vs-multi judgement #6877 made for the REST layer, and several of those parameters accept the array arm **on purpose**.

## The measured fact, and the survey around it

`IHttpRequest.query` is declared `Record< string, string | string[] >`, and the array arm is produced by a real first-party adapter (`NodeHttpServer` hands `?x=1&x=2` through as `['1','2']`, measured over a socket on #6878). Every coercion in this normalizer was written for the string arm:

| Request | What reached the driver before this PR | What the caller saw |
|:---|:---|:---|
| `?$top=1&$top=2` | `Number(['1','2'])` → `limit: NaN` | 200, driver-dependent behaviour |
| `?$skip=1&$skip=2` | `offset: NaN` | same |
| `?status=open&status=won` | `where: { status: ['open','won'] }` | empty page, 200 OK |
| `?$search=a&$search=b` | a two-element array used as the search term | 200 |
| `?$count=true&$count=false` | neither `true` nor `false`; left as-is | 200 |
| body `object: ['invoice','account']` | `QUERY_OBJECT_MISMATCH` | 400 with the wrong diagnosis |

The leftover-key row is what the survey actually bought, and it is not a restatement of `limit`: a bare array in `{ status: ['open','won'] }` is **not** an `$in`. `packages/formula/src/matches-filter.ts` line 187 is `if (Array.isArray(spec)) return false;`, so no row matches — precisely the #4134 failure shape (a filter that can never match answering 200 + `total: 0`).

## The disposition table — this is the deliverable

The judgement keys off the **declared value type**, not off the request. This normalizer serves both `GET /data/:object` (a repeated querystring) and `POST /data/:object/query` (a JSON body), and at this layer the two are **indistinguishable** — both are simply `Array.isArray`. So the rule has to hang on the spec's declaration: on a slot that never declares an array, an array is unambiguous evidence of repetition; on a slot that does, it is the ordinary shape and must not be touched.

**Single-valued — repetition refused (new)**

| Parameter (every wire spelling) | Spec declaration | Why repetition is irreconcilable |
|:---|:---|:---|
| `$top` / `top` / `limit` | `z.number()` | Two window sizes; picking one IS the silent drop |
| `$skip` / `skip` / `offset` | `z.number()` | same |
| `$search` / `search` | `string \| FullTextSearch` | One search, one term |
| `$count` / `count` | response-shape flag | a boolean |
| `object` | `z.string()` | the body's convenience copy of the route object |
| `having` | a `FilterCondition` object | the engine has no AST-array arm for it |
| `cursor` / `distinct` | retired scalars | a tombstone value is a scalar |
| **every leftover key** (implicit field filter) | field equality | a bare array matches no row on any backend |

**Legitimately multi-valued — untouched by this PR**

| Parameter (every wire spelling) | Spec declaration | Why the array is the ordinary shape |
|:---|:---|:---|
| `fields` / `select` / `$select` | `z.array(FieldNodeSchema)` | `?$select=a&$select=b` **is** the projection `['a','b']` |
| `orderBy` / `sort` / `$orderby` | `z.array(SortNodeSchema)` | `normalizeSortNodes` has had a `string[]` arm since #4226; repetition **composes** into a multi-key sort rather than conflicting |
| `expand` / `populate` / `$expand` | a name array lowers to `{name:{object:name}}` | a relation list |
| `searchFields` / `$searchFields` | `z.array(z.string())` | the engine reads both shapes |
| `where` / `filter` / `filters` / `$filter` | a filter **AST is an array** | `['status','=','open']` is ONE filter, not three repetitions |
| `groupBy` / `aggregations` | `z.array(...)` | the aggregation axes |
| `joins` / `windowFunctions` | retired ARRAY keys (#4286) | the tombstone must stay their answer, not an arity refusal |

The one that gets no arity check — `where` — still **refuses** a repeated request, one block further down: `?filter=A&filter=B` arrives as `['A','B']`, which `isFilterAST` cannot read. The diagnosis is less precise than the new message, but it is the honest answer available at this layer: an AST body form and a repeated querystring are byte-identical here.

## The rejection is a thrown typed error, not a coercion to the first value

- **Refuse, do not pick.** Consistent with #6928 / PR #7299 (`GET /api/v1/notifications`, where `NaN` survived the clamp into `data.find({ limit: NaN })`) and with #6307 / #6877 (`readSingleQueryValue`). The rule is also the same: it **counts occurrences, it does not compare values** — two identical values are still two occurrences and are still refused; a one-element array is one occurrence and is unwrapped; an empty array is none.
- **`INVALID_REQUEST` / 400** is what `conflictingQueryParamsError` in this **same normalizer** already answers for the **identical condition reached the other way** — two *spellings* of one slot carrying different values (#4181 → #3795). One slot given two values is one defect; it must not carry two codes depending on whether the caller repeated `filter` or wrote `filter` and `where`.
- Three layers, not three answers: the rest layer spells its 400 `VALIDATION_ERROR` and the runtime layer `VALIDATION_FAILED`, each being that package's registered 400 catalog member in `error-code-ledger.zod.ts`. What has to agree across the three is the **rule** and the **status**, and it does. The message wording is #6877's `repeatedQueryParamMessage` verbatim.

**Dependency direction, stated rather than silently worked around.** #6307's `readSingleQueryValue` now lives at `packages/rest/src/query-multiplicity.ts`. `@objectstack/metadata-protocol` does not depend on `@objectstack/rest`, and that edge would be **backwards** — rest is the transport layer above this one. So the *shape* is reused and not the module. The one deliberate difference is also justified: the rest helper writes a zero-length array back as `undefined`, whereas here it must be `delete`d, because the leftover bucket below reads `Object.keys(options)` and a key left behind carrying `undefined` would be lowered into a `{field: undefined}` predicate.

## Real file surface, file by file

| File | Change |
|:---|:---|
| `packages/metadata-protocol/src/protocol.ts` | +187 / −9, in exactly three places: (1) new module-level `WIRE_DOLLAR_ALIASES`, `ARRAY_VALUED_QUERY_SLOTS`, `ARRAY_VALUED_LIST_QUERY_PARAMS`, `repeatedQueryParamError`, `assertQueryParamArity`, placed after `WIRE_QUERY_ALIAS_SLOTS`; (2) `findData` calls `assertQueryParamArity(options)` after `delete options.context` and before the `$`-alias pass; (3) the `$`-alias loop reads the hoisted table instead of its own inline literal. |
| `packages/metadata-protocol/src/protocol.query-param-arity.test.ts` | New. 316 lines, 46 cases, in three blocks, each **labelled in the file** as evidence or as a guard. |
| `.changeset/findata-query-param-arity.md` | patch, `@objectstack/metadata-protocol`. |

**Region fence (#7306).** This PR enters none of #7306's regions in this same file: the D9.8/D9.9 contributor discriminator and bindings, both hydration seams, `saveMetaItem`'s producer-side refusal, the delete heal's layer subtraction in `deleteMetaItem`, and the retired `installedPackageBindingForObject` are all untouched.

## Reverse verification — four categories, direction declared before running

Pinned to **my own base SHA**, `f188ed60e697e970422a6c5ab414f8eb9e68c789`, not a moving `origin/main`.

### Probe A — take the fix out back to base (`git checkout BASE_SHA -- protocol.ts`, keeping the test file)

**Declared beforehand:** block 1 (refusal) all red; block 2 (preservation of the legitimately-multi parameters) all green, because those are guards; in block 3, `{status:['open']}` and the empty-array case red, `{$top:['5']}` green.

**Measured:** `Tests 18 failed | 28 passed (46)` — matching the prediction case by case.

1. **Predicted red, came out red (the evidence)** — 16 refusal cases plus the 2 one-element/empty-array cases. Both shapes of red showed up, and they are exactly the pair the ADR-0112 envelope discipline exists to separate:
   - Most are "was accepted":
     ```
     Error: {"$top":["1","2"]} was ACCEPTED (answered {"object":"invoice","records":[],"total":0,"hasMore":false}) instead of refused
     ```
   - The `object` case **throws on the unfixed base, with the wrong envelope** — a `toThrow()`-only assertion would have been permanently green here:
     ```
     AssertionError: expected 'QUERY_OBJECT_MISMATCH' to be 'INVALID_REQUEST'
     ```
   - `?top=1&top=2&limit=1` is the other "throws, but the diagnosis is false" case. On base:
     ```
     Conflicting query parameters: 'limit', 'top' are spellings of the same parameter (canonical 'limit') and were given different values.
     ```
     A true refusal with a false diagnosis — and the reason the arity check must run **before** the fold.
   - The empty-array case surfaced a line the card did not name: `Number([])` is `0`, so on base `limit: []` became `limit: +0` and `status: []` became `where: {status: []}`:
     ```
     AssertionError: expected { limit: +0, where: { status: [] } } to not have property "limit"
     ```
2. **Missed predictions:** none. All 18 predicted reds went red, each for the predicted reason.
3. **Assertions green in BOTH directions (guards, not evidence)** — 28 of them: block 2's 24 preservation cases, the 3 "an ordinary single-valued request is untouched" cases, and `{$top:['5']}` (`Number(['5'])` is already 5, since a one-element array stringifies to its element). All three groups carry a `[GUARD — green in both directions]` marker in the test file.
4. **Predictions left unmeasured:** none — nothing was short-circuited by an earlier red, since vitest runs each case.

### Probe B — the variant: replace the per-parameter judgement with a blanket arity rejection

This is the damage this card is most likely to cause, so it is measured rather than argued. Method: empty `ARRAY_VALUED_QUERY_SLOTS` (`= []`), which is exactly "any array of length greater than 1 is a 400".

**Declared beforehand:** block 1 stays fully green (the refusals still fire); block 2 goes fully red; the comma-list projection `?$select=name,status` stays green because it is a string, not an array; block 3 stays green because one-element and empty arrays are accepted under both rules.

**Measured:** `Tests 24 failed | 22 passed (46)` — all of it as declared. The 24 reds are exactly the preservation cases:

```
× $select repeated IS the projection list
× $expand repeated IS the relation list
× $searchFields repeated IS the narrowed search set
× $orderby repeated COMPOSES into a multi-key sort
× a filter AST stays readable — `where`'s array arm is a FILTER, not a repetition
× $select / select / fields / $orderby / sort / orderBy / $expand / populate / expand
  / $searchFields / searchFields / $filter / filter / filters / where / groupBy
  / aggregations accepts a two-element array without a 400
× groupBy / aggregations keep their array arms
```

The most expensive one is the filter AST:

```
Error: The '$filter' query parameter was supplied 3 times. Supply it at most once —
this endpoint will not choose between conflicting values.
```

`['status','=','open']` is **one** filter, and the blanket rule reads it as three repetitions. That is the proof the disposition table is **necessary** and not merely sufficient: block 1 stayed entirely green under the variant, so the "refuse" half and the "preserve" half are two independent facts, and only the table delivers both.

## Tests

```
pnpm --filter @objectstack/metadata-protocol test
  Test Files  68 passed (68)
       Tests  908 passed (908)

pnpm --filter @objectstack/rest test          # consumer 1
  Test Files  78 passed (78)
       Tests  1261 passed (1261)

pnpm --filter @objectstack/runtime test       # consumer 2 (the runtime dispatcher)
  Test Files  119 passed (119)
       Tests  1870 passed (1870)

tsc --noEmit -p packages/metadata-protocol/tsconfig.json
  63 errors — equal to the DEBT entry recorded in scripts/check-type-check-coverage.mjs; nothing added

eslint --no-inline-config packages/metadata-protocol/src/protocol.ts \
                          packages/metadata-protocol/src/protocol.query-param-arity.test.ts   # clean

check:error-code-casing / check:route-envelope / check:engine-double-contract /
check:query-options-erasure / check:nul-bytes / check:empty-changeset /
check:spec-parsed-alias                                                         # all green
```

## STOP conditions — none of the three tripped

1. **`packages/spec` acceptance surface:** untouched. The disposition table *reads* declarations that already exist (`query.zod.ts`'s `z.number()` / `z.array(...)`); nothing changes what the schema accepts or carries.
2. **`packages/rest` behaviour:** not one byte. No `metadata-protocol` to `rest` dependency was introduced (the direction is inverted), and no rest-side behaviour changed.
3. **#7306's regions:** not entered — see the region fence above.

---

### DIFF
--- a/.changeset/findata-query-param-arity.md
+++ b/.changeset/findata-query-param-arity.md
@@ -1,0 +1,38 @@
+---
+'@objectstack/metadata-protocol': patch
+---
+
+`findData`'s shared list-query normalizer now checks the ARITY of every query
+parameter it reads, instead of coercing a repeated one blind (#7321).
+
+`IHttpRequest.query` is `Record< string, string | string[] >` and the array arm
+is produced by a real first-party adapter (`NodeHttpServer` hands `?x=1&x=2`
+through as `['1','2']`). Every coercion in this normalizer was written for the
+string arm, so a repeated parameter was coerced into a value nobody asked for
+and served under a 200:
+
+- `?$top=1&$top=2` → `Number(['1','2'])` is `NaN` → the driver was called with
+  `limit: NaN`. Same for `$skip` / `offset`.
+- `?status=open&status=won` → the leftover-key bucket lowered it to
+  `where: { status: ['open','won'] }`, and a bare array is not a valid field
+  spec — it matches no row on any backend. An empty page, 200 OK.
+- `?$search=a&$search=b`, `?$count=true&$count=false` and a repeated body
+  `object` behaved the same way, each in its own flavour.
+
+Those are now refused with `400` / `error.code: INVALID_REQUEST` — the code this
+same normalizer already answers for the identical condition reached the other
+way (two SPELLINGS of one slot given different values, #4181 → #3795). A
+one-element array is one occurrence and is unwrapped, not refused; an empty
+array is no occurrence.
+
+**Unchanged on purpose — this is a per-parameter judgement, not a sweep.**
+`$select` / `select` / `fields`, `$expand` / `populate` / `expand`,
+`$searchFields`, `$orderby` / `sort` / `orderBy`, `$filter` / `filter` /
+`filters` / `where` (whose array arm is a FILTER AST, not a repetition),
+`groupBy` and `aggregations` all accept the array arm on purpose and keep it
+byte for byte. A blanket "reject repeated parameters" rule would have broken
+every one of them.
+
+Not reachable on today's production Hono adapter, which collapses repeated
+parameters to the first value before any handler runs; it becomes reachable when
+that collapse is removed (#6878 route 2).
--- a/packages/metadata-protocol/src/protocol.query-param-arity.test.ts
+++ b/packages/metadata-protocol/src/protocol.query-param-arity.test.ts
@@ -1,0 +1,316 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * #7321 — `findData`'s list-query normalizer coerces repeated query parameters
+ * without checking arity.
+ *
+ * `IHttpRequest.query` is `Record< string, string | string[] >`
+ * (`packages/spec/src/contracts/http-server.ts`) and the array arm is produced
+ * by a real first-party adapter: `NodeHttpServer` hands `?x=1&x=2` through as
+ * `['1','2']`, measured over a socket on #6878. Every coercion in this
+ * normalizer was written for the string arm, so the array arm was coerced
+ * blind — `Number(['1','2'])` is `NaN`, and `?$top=1&$top=2` reached the driver
+ * as `limit: NaN`. That is the one MEASURED line the card was filed on; the
+ * work is the survey around it, and the survey found the same shape on the
+ * leftover-key bucket (`?status=open&status=won` lowers to
+ * `where: {status: ['open','won']}`, which `matches-filter.ts` answers with a
+ * bare `if (Array.isArray(spec)) return false` — an empty page under a 200).
+ *
+ * ## Three assertion classes, labelled, because only one of them is evidence
+ *
+ *  1. **REFUSAL** — a repeated single-valued parameter answers `400`
+ *     `INVALID_REQUEST` and the engine is never reached. Both `code` AND
+ *     `status` are asserted on every one of these: a bare `toThrow()` here
+ *     would be a permanently-green test for half the cases, because the
+ *     unfixed normalizer ALSO throws for some of them (a repeated `?filter=`
+ *     hits `malformedFilterArrayError` — a true refusal with a false
+ *     diagnosis), and would be blind for the other half, where the unfixed
+ *     normalizer answers 200 with the wrong rows.
+ *
+ *  2. **PRESERVATION of the legitimately-multi parameters** — `$select`,
+ *     `$expand`, `$searchFields`, `$orderby` and `$filter`'s AST array accept
+ *     the array arm ON PURPOSE. These assertions are GREEN IN BOTH DIRECTIONS
+ *     against the fix (they pass on `origin/main` unchanged), so on their own
+ *     they are GUARDS, not evidence. What makes them evidence is the VARIANT
+ *     measured on the PR: emptying `ARRAY_VALUED_QUERY_SLOTS` — i.e. replacing
+ *     the per-parameter disposition with a blanket "no parameter may repeat" —
+ *     turns this whole block red while block 1 stays green. That is the
+ *     damage case the card was filed to prevent, and it is what makes the
+ *     disposition table NECESSARY rather than merely sufficient.
+ *
+ *  3. **PRESERVATION of the ordinary single-valued request** — one occurrence,
+ *     as a bare string, is untouched. Also green in both directions, also a
+ *     guard: it is what a refusal is cheapest to break.
+ */
+
+import { describe, it, expect, vi } from 'vitest';
+import { ObjectStackProtocolImplementation } from './protocol.js';
+
+const SCHEMA = {
+    name: 'invoice',
+    nameField: 'name',
+    searchableFields: ['name', 'status'],
+    fields: {
+        name: { name: 'name', type: 'text' },
+        status: { name: 'status', type: 'text' },
+        amount: { name: 'amount', type: 'number' },
+        owner_id: { name: 'owner_id', type: 'lookup', reference: 'sys_user' },
+        account_id: { name: 'account_id', type: 'lookup', reference: 'account' },
+    },
+};
+
+function makeProtocol() {
+    const find = vi.fn(async () => [] as unknown[]);
+    const aggregate = vi.fn(async () => [] as unknown[]);
+    const engine = {
+        registry: { getObject: (n: string) => (n === 'invoice' ? SCHEMA : undefined) },
+        find,
+        aggregate,
+        count: vi.fn(async () => 0),
+    };
+    return { p: new ObjectStackProtocolImplementation(engine as any), find, aggregate };
+}
+
+/** The option bag `engine.find` was actually handed, for an accepted query. */
+async function optionsFor(query: Record<string, unknown>): Promise<Record<string, unknown>> {
+    const { p, find } = makeProtocol();
+    await p.findData({ object: 'invoice', query } as never);
+    expect(find, `${JSON.stringify(query)} never reached engine.find`).toHaveBeenCalledTimes(1);
+    return (find.mock.calls[0] as unknown[])[1] as Record<string, unknown>;
+}
+
+/** The refusal a rejected query produced, plus proof the engine was not reached. */
+async function refusalFor(query: Record<string, unknown>): Promise<{
+    message: string; status?: number; code?: string; param?: string;
+}> {
+    const { p, find, aggregate } = makeProtocol();
+    let answered: unknown;
+    try {
+        answered = await p.findData({ object: 'invoice', query } as never);
+    } catch (e) {
+        const err = e as Error & { status?: number; code?: string; param?: string };
+        expect(find, 'the engine was reached before the refusal').not.toHaveBeenCalled();
+        expect(aggregate, 'the engine was reached before the refusal').not.toHaveBeenCalled();
+        return { message: err.message, status: err.status, code: err.code, param: err.param };
+    }
+    throw new Error(
+        `${JSON.stringify(query)} was ACCEPTED (answered ${JSON.stringify(answered)}) instead of refused`,
+    );
+}
+
+// ---------------------------------------------------------------------------
+// 1. REFUSAL — the parameters whose declared type is a scalar
+// ---------------------------------------------------------------------------
+
+describe('#7321 — a repeated single-valued parameter is refused, not coerced', () => {
+    it.each<[string, Record<string, unknown>, string]>([
+        // [wire spelling the caller wrote, the query, what it used to become]
+        ['$top', { $top: ['1', '2'] }, 'limit: NaN'],
+        ['top', { top: ['1', '2'] }, 'limit: NaN'],
+        ['limit', { limit: ['1', '2'] }, 'limit: NaN'],
+        ['$skip', { $skip: ['10', '20'] }, 'offset: NaN'],
+        ['skip', { skip: ['10', '20'] }, 'offset: NaN'],
+        ['offset', { offset: ['10', '20'] }, 'offset: NaN'],
+        ['$search', { $search: ['a', 'b'] }, 'a two-element search term'],
+        ['search', { search: ['a', 'b'] }, 'a two-element search term'],
+        ['$count', { $count: ['true', 'false'] }, 'neither true nor false'],
+        ['count', { count: ['true', 'false'] }, 'neither true nor false'],
+        ['object', { object: ['invoice', 'account'] }, 'a bogus object mismatch'],
+        ['having', { having: [{ a: 1 }, { b: 2 }], groupBy: ['status'] }, 'AST junk on aggregate'],
+    ])('refuses a repeated %s with 400 INVALID_REQUEST (was: %s)', async (param, query) => {
+        const err = await refusalFor(query);
+
+        // The ADR-0112 envelope, not merely the throw: `code` AND `status`.
+        expect(err.code).toBe('INVALID_REQUEST');
+        expect(err.status).toBe(400);
+        // #4226 discipline — the message names the spelling the caller WROTE,
+        // not the canonical key the fold would have rewritten it to.
+        expect(err.param).toBe(param);
+        expect(err.message).toContain(`'${param}' query parameter was supplied 2 times`);
+    });
+
+    it('refuses TWO IDENTICAL values too — the rule counts occurrences, not distinct values', async () => {
+        // "At most one DISTINCT value" would be a de-duplication rule no caller
+        // can predict; "supply it at most once" is checkable client-side
+        // (#6877). `?$count=true&$count=true` is still two occurrences.
+        const err = await refusalFor({ $count: ['true', 'true'] });
+
+        expect(err.code).toBe('INVALID_REQUEST');
+        expect(err.status).toBe(400);
+        expect(err.message).toContain('supplied 2 times');
+    });
+
+    it('reports the real count, not just "more than one"', async () => {
+        const err = await refusalFor({ $top: ['1', '2', '3', '4'] });
+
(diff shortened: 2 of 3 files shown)


os-zhuang:
## ACCEPT — PM review (step 7), anchored to head `b20742dc9303ed80ee5452ffbd96c18ecb598a65`

Declared three files and three edit spots in `protocol.ts`. The diff is exactly that, verified hunk by hunk:

| file | status | matches declaration |
|---|---|---|
| `packages/metadata-protocol/src/protocol.ts` | modified (+187/−9) | yes — three hunks: the module constants after `WIRE_QUERY_ALIAS_SLOTS`, the `assertQueryParamArity(options)` call site in `findData`, and the `$`-alias loop rewritten to read the hoisted `WIRE_DOLLAR_ALIASES` |
| `packages/metadata-protocol/src/protocol.query-param-arity.test.ts` | added | yes — 46 cases in three labelled blocks |
| `.changeset/findata-query-param-arity.md` | added | yes |

**All three STOP conditions clear, and the region fence held.** `packages/spec` untouched — the disposition table *reads* existing declarations from `query.zod.ts` rather than changing what is accepted, which is the distinction that would have routed it to #6017. `packages/rest` untouched at zero bytes with no new dependency edge. None of #7306's six regions entered: the contributor discriminator and D9.9 bindings, both hydration seams, `saveMetaItem`, `deleteMetaItem`'s layer subtraction and `installedPackageBindingForObject` are all outside this diff, which lives in the query-alias tables and `findData`. No `content/docs/releases/`, no `docs/adr/**`.

### CI, per job
25 check runs, all `completed`: **23 `success`**, **2 `skipped`** (`Build Docs`, `Console Pin Gate` — path-filtered, correct for this diff). No failure, nothing pending. Named gates: `ESLint`, `TypeScript Type Check`, `Build Core`, `Test Core` 3/3 + rollup, `Dogfood Regression Gate` 3/3 + rollup, `Dogfood Verify CLI`, `Temporal Conformance (live PG + MySQL)`, `Check Changeset`, `Check PR Size`, `Console Pin Freshness`, `No other open PR may claim the same issue`, `ADR maintainer approval` — every one `success`.

### What I am accepting

**The survey was delivered, and its yield is worse than the card's headline.** The card was filed on one measured line (`Number(['1','2'])` is `NaN`, so `?$top=1&$top=2` reaches the driver as `limit: NaN`). The survey found the leftover-key bucket is the more damaging half: `?status=open&status=won` lowers to `where: { status: ['open','won'] }`, and `packages/formula/src/matches-filter.ts:187` is a bare `if (Array.isArray(spec)) return false` — so every row fails to match and the caller gets **an empty page under a 200**, which is the #4134 shape. A wrong-but-loud `NaN` is a smaller problem than a silently empty result set. Also surfaced, and named by neither the card nor my dispatch: `Number([])` is `0`, so an empty array on `limit` used to become `limit: 0`.

**The judgement keys off the declared type rather than the request, and the reasoning for that is the part I most wanted to see.** This normalizer serves both `GET /data/:object` (repeated querystring) and `POST /data/:object/query` (JSON body) and genuinely cannot tell them apart — so `Array.isArray` at this boundary is ambiguous *as a signal about the request*, and unambiguous only relative to what the slot declares. On a slot that never declares an array it is evidence of repetition; on one that does, it is the ordinary shape. That is why the disposition table is derived from the same alias tables the fold uses, so a newly added alias inherits its array arm instead of silently starting to 400.

**Ordering ahead of the `$`-alias pass and the #3795 fold is load-bearing, not stylistic**, and the PR proves it rather than asserting it: on the unfixed base, `?top=1&top=2&limit=1` is refused as `Conflicting query parameters: 'limit', 'top' are spellings of the same parameter` — a true refusal with a false diagnosis. Running arity first means the caller is told what is actually wrong, and the message quotes the spelling they wrote (#4226).

**The dependency direction was answered out loud instead of being quietly reimplemented**, which is exactly what I asked for. `#6307`'s `readSingleQueryValue` lives in `packages/rest/src/query-multiplicity.ts`; `@objectstack/metadata-protocol` does not depend on `@objectstack/rest` and that edge would run backwards, so the *shape* was reused and not the module — with one deliberate divergence that has a measured reason: length-0 is `delete`d rather than set to `undefined`, because the leftover bucket reads `Object.keys(options)` and a key left carrying `undefined` would be lowered into an implicit `{ field: undefined }` predicate. "Not supplied" has to mean absent here.

**The rejection-vs-coercion answer is consistent in the way that matters.** A thrown typed error at 400 / `INVALID_REQUEST`, chosen because `conflictingQueryParamsError` **in this same normalizer** already answers exactly that for the identical condition reached the other way (two spellings of one slot given different values, #4181 → #3795). One slot given two values must not carry two codes depending on how the caller got there. The cross-layer point is handled better than "make all three strings identical": rest spells its 400 `VALIDATION_ERROR` and runtime spells it `VALIDATION_FAILED` because those are each package's registered house catalog member, and what has to agree across the three layers is the **rule and the status**, which do.

### Reverse verification — the four categories, and the variant is the reason this is accepted

**Probe A** (fix removed via `git checkout <base SHA> -- protocol.ts`, base pinned to the dev's own `f188ed60e`, not a moving `origin/main`; no `git stash`): predicted 18 red, measured `18 failed | 28 passed`. The reds came in four distinct shapes, which is the justification for asserting `code` **and** `status` rather than `toThrow()` — most were silent acceptance (`{"$top":["1","2"]} was ACCEPTED … instead of refused`), but the `object` case **already throws on the unfixed base with the wrong envelope** (`expected 'QUERY_OBJECT_MISMATCH' to be 'INVALID_REQUEST'`), where a throw-only assertion would have been permanently green.

**Missed predictions: none**, each red for its predicted reason. **Left unmeasured: none** — vitest ran every case, nothing short-circuited.

**Green in both directions, declared as guards not evidence: 28**, and they are labelled `[GUARD — green in both directions]` in the file itself, not merely in the report.

**Probe B, the variant, is what makes the guards mean something.** Emptying `ARRAY_VALUED_QUERY_SLOTS` — i.e. replacing the per-parameter judgement with a blanket "no parameter may repeat" — was predicted to turn block 2 fully red while block 1 stayed fully green. Measured `24 failed | 22 passed`, exactly as declared, with the most expensive red being `The '$filter' query parameter was supplied 3 times.` — the blanket rule reading the single filter `['status','=','open']` as three repetitions. Because block 1 stayed green under the variant, "refuse the scalars" and "preserve the lists" are two **independent** facts, and only the disposition table delivers both. That establishes the table as *necessary*, not merely sufficient, which is the standard this card was dispatched against.

Type-check debt was measured rather than assumed: 63 errors matching the existing DEBT entry, after the dev noticed their own first draft added a 64th (a TS2345 from the `it.each` tables) and typed it away. No new debt shipped.

### Not folded in — correctly
**#7390** is filed rather than fixed: a repeated `?filter=` cannot be told from a filter AST at this layer (both are `string[]`), so it is refused as `INVALID_FILTER` "malformed filter array" — a true refusal naming the wrong cause — and in the narrow case where the repetition happens to spell a valid AST (`?filter=status&filter=%3D&filter=open`) it silently **succeeds** as `{status:'open'}`. Closing that needs either a `packages/spec` acceptance-surface change (#6017, not #6298) or a new ingress-provenance contract between `packages/rest` and `packages/metadata-protocol`. Both are decisions, neither was in this card's mandate, and folding either in would have made this PR's reverse verification attributable to more than one change. Filed unassigned as `finding` without `pm:queue`, which is the right posture for something that still needs grading.

### Ruling
ACCEPT. Flipping out of draft and enabling auto-merge (SQUASH). Landing stays serial behind PR #7306 through the merge queue, as the region-disjoint parallel-authoring authorization requires.

---

Issue #7408, opened 2026-08-10
[finding] `watch-dot-root.test.ts` case 2's 4s quiet window degrades to a false pass under load — measured against 32-36s delivery


os-zhuang:
Observation-class finding, recorded while implementing #7369 (PR against `packages/metadata-fs/test/watch-dot-root.test.ts`). **Filed unassigned, no `pm:queue`** — nothing a user hits today; this is test-infra correctness.

## What was measured

#7369's work replaced both **positive** waits in `watch-dot-root.test.ts` with a deadline derived from the case's own ceiling. In the course of that, delivery latency for a single external edit was measured on `18ff1dab1` under in-process event-loop starvation (the case blocks its own loop for BLOCK ms out of every BLOCK+GAP ms), 120 iterations across four levels:

| BLOCK/GAP | n | delivery |
|---|---|---|
| none | 60 | 650-707 ms |
| 300/10 | 40 | 3.6-4.9 s |
| 900/5 | 12 | 9.9-12.7 s |
| 2000/5 | 8 | 24-36 s |

## The finding

Case 2 (`still ignores dot entries UNDER the root, including its own bookkeeping`) proves a **negative**: it writes noise (`.cache/x.json`, `view/.scratch.json`, an append to `.objectstack/.log/main.jsonl`), waits `QUIET_WINDOW_MS = 4_000`, and asserts `sink.events` is empty.

That window is 4 s. Delivery under load was measured at up to 36 s. **At any load level past roughly the third row, the quiet window is shorter than the watcher's own delivery latency, so the array is empty because nothing has had time to arrive — not because the noise was correctly ignored.** The assertion passes for the wrong reason, and it passes *more* reliably the more loaded the runner is.

This is not a new hazard in kind — the file's existing comment already warns "a too-short quiet window cannot fail, it can only produce a FALSE PASS on an empty-array assertion". What is new is the measurement: the window is not merely "don't shorten it", it is already 8x too short at a starvation level this repository's merge queue can plausibly reach, and the guard silently stops being a guard exactly when the suite is under the most stress.

## Why it was not fixed in #7369

[...]

PR #7409, opened 2026-08-10
test(metadata-fs): derive the dot-root watcher waits from the case budget instead of a fixed 20s race (#7369)

### LINKED ISSUE #7369: [finding] `watch-dot-root.test.ts` case 1 is wall-clock-timed and ejected an unrelated PR from the merge queue — the queue's full-suite load is where it bites
Observation-class finding, recorded by the `domain:identity` PM seat while triaging a merge-queue ejection. **Filed unassigned, no `domain:*` label** — routing is the triage seat's call. Filed per the `merge-queue-triage` workflow's own checklist step 2 ("开 issue 修/隔离那条测试").

## What happened

Merge-queue build [31366863728](https://github.com/objectstack-ai/objectstack/actions/runs/31366863728) failed and ejected **PR #7333** (`removed_from_merge_queue` 2026-08-10T07:56:20Z, not merged). The failing job was [Test Core (2/3)](https://github.com/objectstack-ai/objectstack/actions/runs/31366863728/job/93387361343):

```
FAIL test/watch-dot-root.test.ts > FileSystemRepository watcher — dot-rooted watch root (#7150)
     > sees an external edit when the root is under a dot-directory
```

**The ejected PR cannot have caused it.** #7333 changes exactly three files, all in `@objectstack/plugin-audit` plus one changeset:

```
.changeset/milestone-summary-lookup-titles.md
packages/plugins/plugin-audit/src/audit-milestone-summary.test.ts
packages/plugins/plugin-audit/src/audit-writers.ts
```

The failing test is in `@objectstack/metadata-fs`. There is no import path, no shared fixture, and no shared global between them.

## Why the queue is where it surfaces

The merge queue runs the **full** suite; PR-side CI runs the affected subset. #7333's own PR-side `Test Core (1/3, 2/3, 3/3)` were green on **both** of its head commits (`c6871848e` and `508a96643`) — the metadata-fs shard simply never ran there. So this is not "green on the PR, red in the queue because the PR is wrong"; it is "the queue is the only context that runs this test, and it is also the heaviest-loaded one".

## Why the case is load-sensitive by construction

`packages/metadata-fs/test/watch-dot-root.test.ts:118-147` measures a real chokidar watcher against **wall-clock sleeps**:

```ts
const sink = collectEvents(repo);
// Past the 200ms self-write suppression window of the put above.
await sleep(400);

await fs.writeFile(path.join(root, 'view', 'case_grid.json'), …);

await Promise.race([sink.first, sleep(EVENT_WAIT_MS)]);
await sink.stop();

expect(sink.events).toHaveLength(1);
```

Two fixed budgets decide the outcome, and both are wall-clock:

1. `sleep(400)` must clear a 200ms self-write suppression window. Under load the `put` above can land late enough that the window has not closed when the external write happens — the event is then suppressed as a self-write.
2. `Promise.race([sink.first, sleep(EVENT_WAIT_MS)])` gives up after a fixed budget. A late inotify delivery yields **zero** events.

Either way `toHaveLength(1)` fails on `0`. On an idle machine both budgets are generous; on the queue runner executing the whole monorepo they are not guaranteed. Note the assertion is an exact count, so it cannot absorb either a late or a duplicate delivery.

Sibling case 2 in the same file (ignore the repository's own bookkeeping) did **not** fail — consistent with a timing effect on the delivery path rather than a logic defect: case 2 asserts an event does *not* arrive, which a slow runner cannot break.

## What I did NOT establish

Being honest about the limit of the evidence, because the checklist's "flaky 实锤" bar is a **second sighting** and I do not have one:

- I scanned the `merge-queue-triage` comments of the **60 most recently updated PRs**. `watch-dot-root` appears in exactly one — #7333's own. So this is a **first recorded occurrence**, not a confirmed repeat.
- I did **not** reproduce it locally, and I did not run it under artificial load. The argument above is structural (wall-clock budgets + exact-count assertion + heavier context), not a reproduction.
- I did not measure how long the queue runner actually took to deliver the event; the job log excerpt in the bot comment is the extraction, not the full log.

So the honest statement is: **unrelated to the PR it ejected, and timing-shaped**, with one sighting. If it ejects a second unrelated PR, that is the confirmation.

## Age

The test is **new**: it landed 2026-08-10T04:12:38Z in `684ab2218` — `fix(metadata-fs): scope the watcher's dotfile ignore to paths relative to the root (#7150) (#7208)` — under four hours before this ejection. It is doing real work (#7150 was a genuine defect where the whole watch was inert), so this is about how the case is *timed*, not about whether it should exist.

## Suggested dispositions (for triage to grade and size)

- **A — wait for the event instead of racing a fixed budget.** Poll for `sink.events.length > 0` up to the case timeout rather than `Promise.race` against `EVENT_WAIT_MS`; the case already carries `CASE_TIMEOUT_MS`, so the failure mode becomes "timed out" with a real budget rather than "0 events" at a fixed one.
- **B — relax the exact count, keep the identity.** Assert at least one event and that the first `update` event is `case_grid` / `source: 'fs'` / `actor: 'fs'`. Keeps everything the case is actually pinning while surviving a duplicate or late delivery. ⚠️ Weigh against the fact that the exact count may be deliberate — a duplicate event is itself a defect on this path, and if so A is the better half.
- **C — quarantine.** Cheapest, and the worst: this test guards a defect where the product's shipped layout made the watcher completely inert. Retiring the guard to protect the queue would give back exactly what #7150 bought.

**A** looks right, possibly with B's identity assertions kept alongside an exact-count assertion that is only made once the event has actually arrived.

Refs: #7150 / PR #7208 (the fix and this test), #7000 (watcher arming, the sibling pin the file warns not to repurpose), PR #7333 (the ejected PR), merge-queue-triage workflow (#4859).

os-zhuang:
Fixes #7369

## Measurement first: the card's original failure is already fixed; the shape defect is not

The dispatch required a measurement before any edit, because the card's mechanism 1 was refuted on a sibling card (#7282) and its actual failure mode may have been fixed by `ab07b5382`. Base for everything below is `18ff1dab1`, which contains `ab07b5382`.

I ran a faithful replica of case 1 (same repository, same put/sleep/external-edit sequence) in a loop under **in-process event-loop starvation** — the case blocks its own loop for BLOCK ms out of every BLOCK+GAP ms — with 12 CPU-oversubscribing busy-loop processes alongside. 120 iterations across four levels:

| BLOCK/GAP | n | sleep(400) drift | delivery | `selfWrites` held at edit time | duplicate events | zero events |
|---|---|---|---|---|---|---|
| none (external CPU only) | 60 | 1-6 ms | 650-707 ms | 0 | 0 | 0 |
| 300/10 | 40 | 517 ms med | 3.6-4.9 s | 0 | 0 | 0 |
| 900/5 | 12 | 1415 ms med | 9.9-12.7 s | 0 | 0 | 0 |
| 2000/5 | 8 | 3601 ms med | 24-36 s | 0 | 0 | 0 |

Three results, all of which change what this PR is:

1. **Mechanism 1 is dead, confirmed on this file.** `selfWrites` still holding the path at external-edit time: **0/120**. #7282's refutation transfers verbatim — the 200 ms suppression timer is scheduled before the test's own 400 ms sleep and with a shorter delay.
2. **The card's original failure mode no longer reproduces.** The event arrived in **120/120** iterations. `ab07b5382` fixed the blind-watcher defect at the source; delivery now goes *late*, never missing. **This PR takes no credit for that fix.**
3. **But mechanism 2 is real and still live.** At 2000/5, delivery took 24-36 s against a hard-coded 20 s budget. Measuring both waiting shapes in the *same* iteration: the old `Promise.race([sink.first, sleep(EVENT_WAIT_MS)])` + `toHaveLength(1)` was **red 5/8**; polling to the case budget was **green 8/8**.

So this is a test-robustness change, exactly as the claim comment anticipated for the "no longer fails" branch — except that the shape defect itself is directly demonstrable, so it is not merely prophylactic.

## Exact count: kept, and demoted to a guard in writing

The card asked whether `toHaveLength(1)` is load-bearing. **It is, but not for the reason the card supposed.** A duplicate chokidar delivery of the same write cannot produce a second `MetadataEvent` at all: `handleFsChange` returns early on `currentHead === hash`. Measured 0 duplicates in 120 iterations, and 0 is the only number that branch can emit. What the count still has teeth for is a *different* event sneaking in — the repository's own `put` escaping self-write suppression, or a dot entry leaking past `isIgnoredWatchPath`. So the count stays exact, is asserted only after the awaited event has arrived, and carries an in-file comment saying plainly that it is a **guard, not evidence** of de-duplication.

## File surface, file by file

Exactly one file changes:

- **`packages/metadata-fs/test/watch-dot-root.test.ts`** — `EVENT_WAIT_MS` (fixed 20 s) removed; new `waitForEvent(sink, deadline)` polls until delivery or until `caseDeadline()`, which is `CASE_TIMEOUT_MS` measured from the case's own start minus a teardown reserve. Applied to both positive waits (case 1's assertion wait and case 2's liveness control — the file's own comment already noted the control "flakes under load the same way"). `CASE_TIMEOUT_MS` 60 s to 120 s, sized from the measured 24-36 s delivery. New `HOOK_TIMEOUT_MS` on `beforeEach`/`afterEach`. `collectEvents` drops its now-unused `first` promise.

**No changeset, `skip-changeset` instead.** The dispatch called for a changeset, and I am deviating deliberately: this PR is test-only, `@objectstack/metadata-fs` publishes `files: ["dist"]`, so the PR declares no release of its own — which is the `changeset-check` job's stated definition of the exemption, and matches repo precedent (`487a197b5`, test-only, no changeset). Flagged for the PM rather than done silently.

Untouched, as instructed: `packages/metadata-fs/src/repository.ts` (no source defect survived `ab07b5382`), #7335's `selfWrites` window, and `content/docs/releases/`.

## Reverse verification, in the four categories

Base recorded and pinned: `18ff1dab1`. Adverse condition on the real file: the same in-process starver injected via a temporary vitest `setupFiles` config, so the actual test file runs starved.

**1. Predicted reds that came out red**

- Old waiting shape vs new, same iterations, 2000/5 starvation: **old red 5/8** (`events.length === 0` at the moment the old race resolves — the assertion's `received 0`), **new green 8/8**. This is the core prediction and it held.
- Real file at 1200/5 sustained: **OLD red, NEW green.** Old: `Test Files 1 failed, Tests 2 failed`, `Error: Hook timed out in 10000ms`. New: `Test Files 1 passed, Tests 2 passed` (167.89 s).

**2. Missed predictions — reported, with cause**

- **I predicted the new shape would be green at 2000/5 sustained on the real file. It was red.** Cause identified: vitest times hooks **separately** from the case on a 10 s `hookTimeout` default that `CASE_TIMEOUT_MS` does not cover. Fixing the case budget simply promoted the next fixed wall-clock budget to being the binding one, and both cases failed with `Hook timed out in 10000ms` while their assertions were satisfied. That is the same defect class the card is about, one layer down, and it is why `HOOK_TIMEOUT_MS` is in this PR at all — it was found by the verification, not designed in.
- **The real-file A/B does not reproduce the `received 0` red at any level I could construct.** At 1200/5 the old assertion still passed (only its hook failed); at 2000/5 the old file died on its own 60 s case ceiling during setup. Cause: sustained starvation via `setupFiles` also inflates `repo.start()` and `put()`, which the per-iteration replica did not. So the `received 0` red is measured on the **replica**, not on the file itself. Stating that plainly rather than presenting the replica result as if it came from the file.

**3. Assertions green in BOTH directions — guards, not evidence**

- `expect(sink.events).toHaveLength(1)` **cannot** go red on a duplicate delivery, in either direction, because of the `currentHead === hash` early return. Labelled as such in the report and in the test file itself.
- Case 2's `QUIET_WINDOW_MS` negative assertion: a 4 s wall-clock window against 32-36 s delivery is not a window at all — it can only produce a false pass under load. Left functionally untouched (out of this card's scope), but the measurement is now written into its comment, and filed separately as #7408.

**4. Predictions left unmeasured**

- I did **not** measure this on a real merge-queue runner; all load here is synthetic. The queue's actual saturation profile may sit anywhere on the ladder above.
- I did **not** establish a second sighting of the original flake. The card was explicit that it had one; that is unchanged, and my measurement suggests the original cause is gone regardless.
- The 2000/5 sustained level is a **pathological** boundary (loop blocked ~99.75% for the file's whole lifetime) where even 30 s hooks time out. I did not chase it with larger budgets — that is an infinite regress. Worth noting how it fails now: **zero `AssertionError`s in that run**. Every failure is an honest timeout rather than a misleading `received 0` that reads as a broken watcher, which is precisely the failure-mode change the card asked for.

## Local verification

```
pnpm --filter @objectstack/metadata-fs test       Test Files 5 passed (5), Tests 31 passed (31)
pnpm --filter @objectstack/metadata-fs typecheck  tsc --noEmit, clean
node scripts/check-nul-bytes.mjs                  OK (6764 text files, no raw ASCII control bytes)
```

---

### DIFF
--- a/packages/metadata-fs/test/watch-dot-root.test.ts
+++ b/packages/metadata-fs/test/watch-dot-root.test.ts
@@ -35,38 +35,91 @@ const ref = (name: string): MetaRef => ({ org: 'system', type: 'view', name });
 const sleep = (ms: number) => new Promise<void>((r) => setTimeout(r, ms));
 
 /**
- * Deadline for a **positive** watcher assertion — the longest a case waits for
- * an event it expects to arrive. Each wait races this against the event
- * promise, so a healthy run still finishes in roughly the poll interval; the
- * number is only ever paid on the way to a failure.
- *
- * Sized for the merge queue, not for a quiet laptop. The queue runs the FULL
- * suite (PR-side CI runs only the affected subset), and this watcher rides
- * `usePolling: 1000ms` plus an `awaitWriteFinish` stability window — both are
- * wall-clock timers that stretch when the runner is saturated, while the
- * assertions themselves are unaffected. PR #7208 was ejected from the queue
- * twice on exactly that: first `fs-behavior.test.ts > chokidar: external file
- * change emits an update event` (3s deadline), then this file's positive case
- * (8s deadline) — both green on the identical SHA in PR CI and locally.
+ * Reserve held back from the case ceiling so that a missing event is reported
+ * by this file's own assertion ("expected length 1, received 0") rather than
+ * by a bare vitest timeout. Covers `sink.stop()`, `repo.close()` and the
+ * tmpdir teardown in `afterEach`.
  */
-const EVENT_WAIT_MS = 20_000;
+const TEARDOWN_RESERVE_MS = 5_000;
+
+/**
+ * Poll interval for `waitForEvent`. A healthy run leaves the wait on the first
+ * turn after delivery, so this is what a passing case pays, not the ceiling.
+ */
+const EVENT_POLL_MS = 50;
 
 /**
  * Quiet window for the **negative** assertion below. ⛔ Never shorten this: a
  * too-short quiet window cannot fail, it can only produce a FALSE PASS on an
- * empty-array assertion. Its liveness control uses `EVENT_WAIT_MS` like every
- * other positive wait — the control is a positive assertion and flakes under
- * load the same way.
+ * empty-array assertion.
+ *
+ * This one budget stays wall-clock, because you cannot wait for an absence —
+ * and that is a real limit, not a solved problem. Measured for #7369: under
+ * event-loop starvation heavy enough to push delivery to 32-36s, a 4s window
+ * is no longer a window at all and the emptiness below becomes a false pass.
+ * The case's *positive* control underneath is what still has teeth there, and
+ * it is now deadline-driven (`waitForEvent`) rather than fixed-budget.
  */
 const QUIET_WINDOW_MS = 4_000;
 
 /**
- * Per-case ceiling. Has to clear the quiet window plus a full `EVENT_WAIT_MS`
- * plus repository setup, or the case dies on the vitest timeout before its own
- * deadline is reached — which reports as a timeout rather than as the missing
- * event, and re-introduces the flake the deadlines above are widening away.
+ * Per-case ceiling — and, since #7369, the **source** of every positive wait's
+ * budget: `waitForEvent` waits until `caseDeadline()`, which is this value
+ * measured from the case's own start minus `TEARDOWN_RESERVE_MS`. There is no
+ * second, smaller wall-clock budget left in the file to expire first.
+ *
+ * Sized from measurement, not taste. Under in-process event-loop starvation
+ * (the loop blocked ~99.75% of the time) delivery of a single external edit
+ * was measured at 24-36s while still arriving every time; the quiet window and
+ * repository setup ahead of it stretch too. 120s clears that with margin, and
+ * a healthy run pays none of it — every wait exits on delivery.
  */
-const CASE_TIMEOUT_MS = 60_000;
+const CASE_TIMEOUT_MS = 120_000;
+
+/**
+ * The budget every positive wait in this file spends, measured from the start
+ * of the case that calls it. Call it as the case's first statement.
+ */
+const caseDeadline = (): number => Date.now() + CASE_TIMEOUT_MS - TEARDOWN_RESERVE_MS;
+
+/**
+ * Wait until the sink has delivered at least one event, or until `deadline`.
+ *
+ * ⛔ Never put a *fixed* wall-clock budget back here. The shape this replaced
+ * was `Promise.race([sink.first, sleep(EVENT_WAIT_MS)])` with a hard-coded
+ * `EVENT_WAIT_MS = 20_000`, and #7369 is what that costs: the budget is a
+ * constant while the thing it is timing is not. This watcher rides
+ * `usePolling: 1000ms` plus an `awaitWriteFinish` stability window, both of
+ * which stretch with runner load — and the merge queue, which runs the FULL
+ * suite while PR-side CI runs only the affected subset, is the most loaded
+ * context this file ever executes in. When the fixed budget expires first the
+ * case reports `received 0`, which reads as "the watcher is broken" and is
+ * really "the test stopped listening too early". PR #7333 was ejected from the
+ * queue by exactly that, having touched nothing in this package.
+ *
+ * Measured for #7369 on `18ff1dab1`, one external edit per iteration, under
+ * in-process event-loop starvation (the case blocks its own loop for BLOCK ms
+ * out of every BLOCK+GAP ms), external CPU oversubscription alongside:
+ *
+ *   BLOCK/GAP   n    delivery        old fixed-budget shape   this shape
+ *   (none)      60   650-707ms       green                    green
+ *   300/10      40   3.6-4.9s        green                    green
+ *   900/5       12   9.9-12.7s       green                    green
+ *   2000/5       8   24-36s          RED 5/8                  green 8/8
+ *
+ * The event arrived in 120 of 120 iterations — delivery goes *late*, never
+ * missing. (The failure mode where it never arrives at all was a different
+ * defect, #7282, fixed at the source in `ab07b5382`; it is not what this
+ * shape defends against.) So the only thing that decided pass or fail at the
+ * bottom row was whether the test was still listening, which is why the budget
+ * now comes from `CASE_TIMEOUT_MS` instead of a number chosen in advance.
+ */
+async function waitForEvent(
+  sink: { events: MetadataEvent[] },
+  deadline: number,
+): Promise<void> {
+  while (sink.events.length === 0 && Date.now() < deadline) await sleep(EVENT_POLL_MS);
+}
 
 describe('FileSystemRepository watcher — dot-rooted watch root (#7150)', () => {
   /** Stands in for the project directory. */
@@ -92,30 +145,26 @@ describe('FileSystemRepository watcher — dot-rooted watch root (#7150)', () =>
    */
   function collectEvents(r: FileSystemRepository): {
     events: MetadataEvent[];
-    first: Promise<void>;
     stop: () => Promise<void>;
   } {
     const iter = r.watch({ org: 'system' }, 999)[Symbol.asyncIterator]();
     const events: MetadataEvent[] = [];
-    let resolveFirst!: () => void;
-    const first = new Promise<void>((res) => { resolveFirst = res; });
     let stopped = false;
     void (async () => {
       while (!stopped) {
         const next = await iter.next();
         if (next.done) return;
         events.push(next.value as MetadataEvent);
-        resolveFirst();
       }
     })();
     return {
       events,
-      first,
       stop: async () => { stopped = true; await iter.return?.(undefined); },
     };
   }
 
   it('sees an external edit when the root is under a dot-directory', async () => {
+    const deadline = caseDeadline();
     repo = new FileSystemRepository({ root, org: 'system' }); // watcher ENABLED
     await repo.start();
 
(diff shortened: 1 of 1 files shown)


os-zhuang:
## PM review (step 7) — ACCEPT on substance; flip held on one unconverged gate

Anchored to head `8518b838ca5e4d715988a1623fa720187f034434`.

### Your open question: **A**, and my dispatch was wrong to say otherwise

Take **A** — no changeset, `skip-changeset` label. My dispatch said "a `.changeset/*.md` entry IS required"; that clause belongs on cards that ship something user-visible, and I should not have written it unconditionally. The standing rule is the opposite of what I told you: **tests/docs-only PRs take the `skip-changeset` label rather than an empty changeset**, because an empty changeset lingers in the release pipeline (#4898). Your reasoning matches it exactly — `@objectstack/metadata-fs` publishes `files: ['dist']`, so a test-only change declares no release of its own.

Your second argument is the one that settles it: I told you the changeset must not claim credit for #7336's fix, and there is no honest user-visible line left for this PR to write. B would have manufactured one. Labels read back as `['tests', 'skip-changeset']`; `Check Changeset` is green on A. Confirmed — no action needed from you.

Declared one file; the diff is one file — `packages/metadata-fs/test/watch-dot-root.test.ts`, +103/−31. No `repository.ts`, no `content/docs/releases/`, no `docs/adr/**`, #7335 not folded in, no quarantine. Clean.

### What I am accepting

**You measured before editing, and the measurement changed what this card is.** The card's original failure mode is **gone**: the event arrived **120/120** on a base containing `ab07b5382`. Delivery now goes late, never missing — and the PR says so and takes no credit for it. That was the outcome I said would be equally good, and reporting it plainly is worth more than a fix would have been.

**Mechanism 1 is dead here too.** `selfWrites` held the path at external-edit time **0/120**, so #7282's refutation transfers to this file rather than merely being assumed to. The card's stated mechanism is now falsified in both files that carried it.

**Mechanism 2 is real, and the evidence for it is the strongest shape available.** At 2000/5 starvation, delivery took 24–36s against a hard-coded 20s `EVENT_WAIT_MS`. Crucially you measured **both waiting shapes in the same iterations** — old `Promise.race` red 5/8, deadline polling green 8/8 — so the comparison cannot be confounded by run-to-run variance. That is a materially better experiment than running the two shapes separately, which is what I would have settled for.

**The exact-count question came back with a better answer than either option I offered.** I said decide it by measurement rather than taste. You found the card's stated reason is wrong — a duplicate delivery **cannot** produce a second `MetadataEvent`, because `handleFsChange` returns early on `currentHead === hash` (0 duplicates in 120 iterations, and 0 is the only number that branch can emit). Then you kept the exact count anyway, for a *different* and real reason: it still catches a **different** event sneaking in — the repository's own `put` escaping self-write suppression, or a dot entry leaking past `isIgnoredWatchPath`. Kept, asserted only after arrival so a slow runner cannot turn it into `received 0`, and labelled in the file as a guard rather than evidence. Correct on all three counts.

**Both missed predictions are reported, and one of them is the most valuable thing in this PR.** Predicting NEW green at 2000/5 and measuring **red** is what exposed that vitest times hooks separately on a 10s `hookTimeout` the case budget does not cover — so fixing the case budget merely **promoted the next fixed wall-clock budget to binding**, with both cases failing `Hook timed out in 10000ms` while their assertions were satisfied. `HOOK_TIMEOUT_MS` is in this PR because verification found it, not because anyone designed it in. A run that had stopped at "the case budget is fixed now" would have shipped a fix that moves the flake rather than removing it.

**The second missed prediction is reported with a distinction most reports would blur.** The `received 0` red could not be reproduced on the real **file** at any starvation level — at 1200/5 the old assertion still passed, and at 2000/5 the old file died on its own 60s ceiling during setup, because sustained starvation also inflates `repo.start()`/`put()` in a way the per-iteration replica does not. You state explicitly that the red is measured on the **replica** and decline to present it as coming from the file. That is exactly the honesty this lane's reverse-verification standard is for.

**And the hardening is demonstrated on the real artifact, not only the replica** — adverse-condition A/B at 1200/5 sustained: OLD `Test Files 1 failed (1), Tests 2 failed (2)` / `Hook timed out in 10000ms`; NEW `2 passed (2)` in 167.89s.

**The in-file documentation carries its own evidence**, which is what makes this durable: the measurement table lives in the `waitForEvent` docblock, the ⛔ never-restore-a-fixed-budget rule names #7333 as the PR that ejection cost, and `QUIET_WINDOW_MS` is honest that "you cannot wait for an absence — and that is a real limit, not a solved problem" rather than pretending the negative case is now sound.

**#7408 is the right split.** Case 2's 4s quiet window degrading to a **false pass** under load — passing *more* reliably the more loaded the runner is — is a genuinely inverted failure mode, and the sound fix restructures what case 2 measures. Folding it in would have made this PR's reverse verification attributable to two changes.

### Why the flip is held

`TypeScript Type Check` is still **`in_progress`** on this head (started 10:22:32Z). Everything else is `success` or path-skipped: `ESLint`, `Build Core`, `Test Core` 1/3·2/3·3/3 + rollup, `Dogfood Regression Gate` 3/3 + rollup, `Dogfood Verify CLI`, `Temporal Conformance (live PG + MySQL)`, `Check Changeset`, `Check PR Size`, `Console Pin Freshness`, `ADR maintainer approval`, `No other open PR may claim the same issue`.

Reporting at draft-PR time with gates recorded honestly as `in_progress` is the contract working as intended — CI convergence, the ready-flip and landing are mine, not yours. ⛔ Nothing further is needed from you; do not poll this. I hold a timer and will flip to ready + auto-merge (SQUASH) once that job's conclusion reads `success`.

---

Issue #7426, opened 2026-08-10
`deleteMetaItem`'s catch re-wrap drops the error `code`, so a repository refusal reaches the caller as a 403 with no catalogued code


os-zhuang:
Found while implementing #6960 (PR pending). Out of scope there — the fix is an error-envelope change on a path #6960 was not authorised to touch — so filing plainly and leaving the grading to triage.

## What was measured

`ObjectStackProtocolImplementation.deleteMetaItem`'s repository branch ends in a `catch` that translates `ConflictError` faithfully and then re-wraps **everything else**:

```ts
const e = new Error(`Failed to delete customization overlay: ${err.message ?? err}`);
(e as any).status = err?.status ?? 500;
throw e;
```

`status` is carried forward. **`code` is not.** So a refusal thrown by `SysMetadataRepository` with a full ADR-0112 envelope reaches the caller with `err.status === 403` and `err.code === undefined`, its code surviving only as text inside the message.

Measured on `f16e54e1d`, control-plane kernel (`environmentId === undefined`, which is what the CLI's lightweight assembler builds for a host config — the showcase topology of #5086), deleting an artifact-backed item of a type the registry does not mark `allowOrgOverride`:

```
Error: Failed to delete customization overlay: [NOT_OVERRIDABLE] 'permission' is not allowOrgOverride
in the registry. Overlay-allowed: view, dashboard, report, translation, email_template.
Set OS_METADATA_WRITABLE to enable additional types at runtime.

  err.status -> 403
  err.code   -> undefined
```

The same refusal on a project kernel (`environmentId` set) is thrown by `deleteMetaItem`'s own two-tier block and arrives intact, `code: 'NOT_OVERRIDABLE'` + `status: 403`. So the envelope a caller sees for one and the same refusal depends on the deployment topology.

## Why it reads as unintended

- ADR-0112 makes the code the machine-readable half of a refusal; a 403 with no code is exactly the shape the ledger exists to prevent, and every other refusal on this path carries one.
- The evidence that it has already cost a pin: ADR-0029 D9's control-plane delete pin in `packages/objectql/src/protocol-object-overlay-layer.test.ts` asserts `expect(String(refused.message)).toContain('NOT_OVERRIDABLE')` — a substring of the message — while its project-kernel sibling one describe up asserts `code` **and** `status`. A test that wants the envelope and can only reach the prose is the symptom, not a style choice.
[...]

Issue #4717, opened 2026-08-02
[P2/P3] 运行时授权门的 advisory findings 要进响应并被 Studio 渲染 + docs「the gate」补第四列(#4463 收尾)


os-zhuang:
`Blocked-by: objectstack-ai/objectstack#4463`(PR #4715 合并前不要派发)

#4463 P1 让运行时闸门跑起来了,但 **advisory 级 findings 目前只进服务端日志,不进响应**。PR #4715 自己写明了这一点的性质:

> 跑了规则却把结论丢掉,是本单要修的那个洞小一号的版本。

同意这个判断 —— 所以它是本单,不是「以后再说」。

## 范围

- [ ] **advisory findings 进响应**:gating → 422 已经落地;advisory 应走 2xx + 响应里带 findings。#4463 D3 已经定了形状 —— 复用 Zod 失败那个结构化 `issues` 信封(`rule` / `path` / `where` / `message` / `hint`),**objectui 不需要新协议,只需读一个已有形状的新字段**
- [ ] **objectui 侧渲染**(跨仓):Studio 保存后把 advisory findings 显示成警告。按跨仓协议,这一半要在 objectui 队列里单独立单并用 `Blocked-by` 指回本单;本单落地后由 objectui 分片的 PM 接手
- [ ] **P3 `surfaces` 泛化**:#4463 P1 已按需引入 `surfaces: ('cli'|'runtime-publish')[]` + `runtimeTypes` / `surfaceReason`(那是棘轮守卫的结构性前提,不是可选项 —— 守卫无法守一个没被声明的 surface)。本单把它推广成开放的 N-surface 维度
- [ ] **docs**:把「the gate」那张表加上第四列(运行时发布面)。注意方向 —— P1 之后运行时**多**做了一层检查而文档只写了三个 CLI 命令,属于**少承诺多兑现**,不是 declared ≠ enforced,所以不紧急;但表格不补,下一个读文档的人(和 AI)仍会以为 Studio 保存不过闸门

## 关联

#4463(P1 本体)、PR #4715、#4716(P2 object 写入上门)、#4409 / #4445、cloud#688(publish-gate lint 先例)

hotlong:
[...]

PR #7429, opened 2026-08-10
fix(metadata-protocol): allow REMOVING a legacy env overlay on a rolled-back overlayable type (#6960)

### LINKED ISSUE #6960: A legacy env overlay on an artifact-backed item of a rolled-back type can no longer be REMOVED through the ordinary delete path (403) — only via OS_METADATA_WRITABLE

os-zhuang:
Fixes #6960

Base of this branch: `f16e54e1d` — the commit that landed ADR-0029 D9 and its pins. Every reverse-verification direction below was measured against **that** SHA, never a moving `origin/main`.

## What moved, and what did not

The maintainer's ruling of 2026-08-10 moves the **delete** side only. Two tiers reach the same two refusal points, and the separator is the registry's own `supportsOverlay` flag — **not** `allowOrgOverride`:

| tier | registry declaration | delete | create / update |
|---|---|---|---|
| rolled-back overlayable — `permission` / `position` / `page` / `app` / `dataset` / `book` (and `tool` / `skill`, same shape) | `supportsOverlay: true` + `allowOrgOverride: false` | **now ALLOWED** | refused, unchanged |
| `object` and every other `supportsOverlay: false` type | `supportsOverlay: false` | refused, **unchanged** | refused, unchanged |

`supportsOverlay` is the capability of the READ path ("an overlay row under this name changes what is served"); `allowOrgOverride` is the permission on the WRITE path ("a tenant may author one"). #6483 / PR #6608 rolled the permission back and deliberately left the capability alone — which is exactly the state #6960 was filed about. Keying the relaxation on the capability is what makes it stop at the tier boundary: an `object` overlay registers as its own contributor **layer** (ADR-0029 D9) rather than merging, and its reset refusal is D9.6's declared, maintainer-approved cost.

Both refusal points are treated, because the second is topology-independent:

1. `deleteMetaItem`'s `environmentId !== undefined` two-tier block — which refused **before** it ever probed for the row;
2. `SysMetadataRepository`'s delete gate — reached by a control-plane kernel (`environmentId === undefined`), which skips (1) entirely. A fix pinned on only one path would have left the defect alive on one topology.

## File surface, file by file

- **`packages/metadata-protocol/src/protocol.ts`** — adds the registry-derived `OVERLAY_CAPABLE_TYPES` set and the `mergesOverlayAtRead(type)` predicate beside the existing `isOverlayAllowed` / `isRuntimeCreateAllowed` family (derived, plural-normalized, no parallel whitelist — Prime Directive #8). Adds `&& !legacyOverlayRemoval` to `deleteMetaItem`'s `artifactBacked && !overlayAllowed` refusal, with the ruling, the boundary and the delete-only asymmetry documented at the call site. Extends `deleteMetaItem`'s two-tier header comment with the new bullet. Extends the **save-side gate's** carve-out comment — the one that already promised "removing a code-only row that predates this refusal is repair, and must stay possible" — to name the **artifact-backed** tier it did not cover, which is the divergence between stated intent and enforced behaviour the issue opened with. `saveMetaItem`'s gates themselves are untouched; the only edits inside `saveMetaItem` are comment lines.
- **`packages/metadata-protocol/src/sys-metadata-repository.ts`** — adds the registry-derived `OVERLAY_CAPABLE_TYPES` set and a new private `assertDeleteAllowed(type, intent)`; `delete()` now calls it, `put()` still calls `assertAllowed` directly. The carve-out is scoped to the `override-artifact` intent (`runtime-only` stays governed by `allowRuntimeCreate`). `assertAllowed` itself is unchanged, byte for byte.
- **`packages/metadata-protocol/src/protocol.legacy-overlay-delete.test.ts`** — new, 101 cases.
- **`.changeset/legacy-overlay-delete-rolled-back-types.md`** — new, `patch` on `@objectstack/metadata-protocol`.

No other file is touched. `packages/spec` is **read**, never changed — the acceptance surface is untouched, which is the distinction the dispatch's STOP condition 2 draws.

## Reverse verification — four categories, directions declared before running

⚠️ Zero affected rows in the corpus today (measured at PR #6608), so there is no live victim and no "it works now" demo. **The pins are the deliverable.**

### 1. Predicted red, measured red

Fix taken out with `git checkout f16e54e1d -- (the two source files)` (never `git stash` — shared stack), test file kept:

```
Tests  57 failed | 44 passed (101)
```

Predicted 56, measured 57 (the extra one is category 3). Both refusal points reproduced, one per topology, with their real messages:

```
project kernel  — removes a legacy env overlay on an artifact-backed permission …
Error: [not_overridable] Metadata item 'permission/legacy_permission_overlay' is provided by a code
package and the type has not opted into per-org overlay writes.
  at ObjectStackProtocolImplementation.deleteMetaItem src/protocol.ts:11932

control-plane   — removes a legacy env overlay on an artifact-backed permission …
Error: Failed to delete customization overlay: [NOT_OVERRIDABLE] 'permission' is not allowOrgOverride
in the registry. Overlay-allowed: view, dashboard, report, translation, email_template.
  at ObjectStackProtocolImplementation.deleteMetaItem src/protocol.ts:12152
```

That is 8 types x 3 cases x 2 topologies at the protocol layer, plus 8 `SysMetadataRepository.delete` cases, plus the category-3 case. Every one of them is green with the fix in place.

### 2. Green in BOTH directions — guards, not evidence

Labelled `[GUARD]` in their own test names as well as here: the `object`-tier delete refusals on both topologies; the same refusal for every other `supportsOverlay: false` type with a runtime write channel; `put` still refusing an `override-artifact` write of every rolled-back type; `saveMetaItem` still refusing both the plain and the org-scoped write; the `runtime-only` delete still judged by `allowRuntimeCreate`. They pass before and after — what makes them load-bearing is the variant experiment below, not this run.

### 3. Missed prediction — one, kept rather than tidied away

`a delete with NO overlay row answers the no-op success instead of 403` was written as a guard and **measured red** on the project kernel. Cause identified: the refusal ran *before* the probe for the row (the issue's own words), so an artifact-backed item with **nothing customized** was refused too — the miss is real coverage the prediction did not anticipate, not a broken assertion. On the control-plane kernel it was already green: the probe returns first and `repo.delete` is never reached. The case keeps its corrected comment in the file.

### 4. Unmeasured

None. Every assertion in the new file was run in both directions.

## Variant verification — the damage case, measured

The obvious over-broad fix is to relax the shared predicate for **every** artifact-backed type. Applied exactly that (`mergesOverlayAtRead` returning unconditionally `true`, and `assertDeleteAllowed` returning early for any non-`runtime-only` intent), rebuilt, and ran ADR-0029 D9's own pins:

```
packages/objectql/src/protocol-object-overlay-layer.test.ts
  Tests  2 failed | 14 passed (16)

  x  …and so is the RESET of one: deleteMetaItem answers NOT_OVERRIDABLE / 403
     AssertionError: expected null to be an instance of Error
  x  a control-plane kernel refuses at the repository, and subtracts under the hatch
     AssertionError: expected null to be an instance of Error
```

Exactly D9's two delete pins, one per topology, red — and 14 of this PR's own boundary guards red with them. The variant was reverted and D9's file is back to **16/16 passed** under the shipped scoping. That is what makes the scoping *necessary* rather than merely sufficient: D9's pins were never edited, adjusted, or worked around.

## Tests

```
pnpm --filter @objectstack/metadata-protocol test   ->  69 files, 1009 tests passed
pnpm --filter @objectstack/objectql test            -> 172 files, 3066 tests passed
pnpm --filter @objectstack/rest test                ->  79 files, 1278 tests passed
pnpm --filter @objectstack/runtime test             -> 120 files, 1906 tests passed
pnpm --filter @objectstack/metadata test            ->  30 files,  593 tests passed
pnpm --filter @objectstack/objectql typecheck       -> Done
```

`metadata-protocol` has no `typecheck` script by design — it carries a measured DEBT entry (63) in `scripts/check-type-check-coverage.mjs`. Ran `tsc --noEmit -p tsconfig.json` directly: **63 errors, unchanged**, and **zero** of them in the new test file, so the ratchet is not raised.

Family gates that ride the ESLint job, all green locally: `check:nul-bytes` (plus a targeted control-byte self-scan over all four touched files, no hits), `check:engine-double-contract`, `check:error-code-casing`, `check:route-envelope`, `check:meta-type-normalized`, `check:adr-anchors`.

## One finding filed, not fixed here

**#7426** — `deleteMetaItem`'s `catch` re-wrap carries `status` forward but drops `code`, so a repository refusal reaches the caller as a 403 with no catalogued code on the control-plane topology. It is why ADR-0029 D9's control-plane pin asserts a message substring while its project-kernel sibling asserts `code` + `status`, and why this PR's own control-plane assertions state the gap instead of asserting around it — carrying the full `code` + `status` envelope one layer down, directly against `SysMetadataRepository`, where nothing re-wraps it. Out of scope: fixing it changes the error envelope of every non-conflict failure on that path, which this ruling does not reach.

---

### DIFF
--- a/.changeset/legacy-overlay-delete-rolled-back-types.md
+++ b/.changeset/legacy-overlay-delete-rolled-back-types.md
@@ -1,0 +1,48 @@
+---
+"@objectstack/metadata-protocol": patch
+---
+
+fix(metadata-protocol): a legacy env overlay on a rolled-back overlayable type can be REMOVED again (#6960)
+
+#6483 / PR #6608 flipped `permission` / `position` / `page` / `app` / `dataset` /
+`book` to `allowOrgOverride: false`. That closed the **write** door and
+deliberately left the **read** path alone — `supportsOverlay` stayed `true`, so
+an overlay row authored *before* the rollback still merges overlay-wins and
+still shapes the effective body.
+
+Removing such a row was refused, at two places and on two different topologies:
+
+- `deleteMetaItem`'s `environmentId !== undefined` branch answered
+  `403 not_overridable` **before** it ever probed for the row — so even an
+  artifact-backed item with nothing customized was refused;
+- `SysMetadataRepository`'s delete gate refused the same removal
+  `intent: 'override-artifact'`, and that check is **topology-independent**, so
+  a control-plane kernel (no `environmentId`, which skips the first gate
+  entirely) was refused there instead.
+
+Net effect for an environment that upgraded across the rollback while holding
+such a row: the ordinary "Reset to package default" flow answered 403 with the
+item still customized, and `OS_METADATA_WRITABLE` was the only documented way
+out.
+
+**What changes.** Per the maintainer's ruling of 2026-08-10, the **delete** side
+moves: removing an overlay of a type whose loader merges overlays at read time
+is allowed through the ordinary delete path, on both kernel topologies, without
+the operator hatch. Deleting an overlay restores the code-declared state — the
+narrowing direction, which cannot widen anything — so refusing it served no
+security purpose while trapping the repair behind an escape hatch.
+
+**What does NOT change, deliberately.**
+
+- **Create and update stay refused exactly as before.** The asymmetry is the
+  ruling, not an oversight: `saveMetaItem`'s gate and `SysMetadataRepository.put`
+  are untouched, and both gates' doc comments now record why, so the asymmetry
+  is not later "fixed" into symmetry.
+- **The `object` tier does not move.** The relaxation is keyed on the registry's
+  `supportsOverlay` flag, not on `allowOrgOverride`, so it stops at the tier
+  boundary: `object` declares `supportsOverlay: false` (its overlay registers as
+  its own contributor layer, ADR-0029 D9) and keeps refusing both verbs, which
+  is D9.6's declared cost.
+
+Zero affected rows in the in-repo corpus today (measured at PR #6608), so this
+is a correctness-of-contract change with no live victim.
--- a/packages/metadata-protocol/src/protocol.legacy-overlay-delete.test.ts
+++ b/packages/metadata-protocol/src/protocol.legacy-overlay-delete.test.ts
@@ -1,0 +1,694 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * #6960 — a legacy env overlay on an ARTIFACT-BACKED item of a ROLLED-BACK
+ * overlayable type is removable through the ordinary delete path.
+ *
+ * ## The defect
+ *
+ * #6483 / PR #6608 flipped six types to `allowOrgOverride: false`. That closed
+ * the WRITE door and deliberately left the READ path alone: `supportsOverlay`
+ * stayed `true`, so an overlay row authored BEFORE the rollback still merges
+ * overlay-wins and still shapes the effective body. Removing it was refused at
+ * two places, on two different topologies:
+ *
+ *  1. `deleteMetaItem`'s `environmentId !== undefined` branch — `artifactBacked
+ *     && !overlayAllowed` threw `NOT_OVERRIDABLE` / 403 BEFORE any probe for
+ *     the row;
+ *  2. `SysMetadataRepository`'s delete gate, which is TOPOLOGY-INDEPENDENT —
+ *     a control-plane kernel (`environmentId === undefined`) skips (1) entirely
+ *     and is refused here instead, reached exactly when a row DOES exist
+ *     because the "no overlay found" probe returns a no-op success earlier.
+ *
+ * Net: the ordinary "Reset to package default" answered 403 with the item
+ * still customized, and `OS_METADATA_WRITABLE` was the only documented door.
+ *
+ * ## The ruling this file pins
+ *
+ * Maintainer, 2026-08-10 (#6960): **the delete side moves.** Removing an
+ * overlay restores the code-declared state — the narrowing direction, which
+ * cannot widen anything — so refusing it serves no security purpose while
+ * trapping repair behind the operator hatch. **Create and update on such items
+ * stay refused exactly as today**, and the delete-only asymmetry is recorded
+ * in the gate's doc comment so it is not later "fixed" into symmetry.
+ *
+ * ## The tier boundary, which is the whole safety argument
+ *
+ * Two tiers reach the same two refusal points and only ONE moves. The
+ * separator is `supportsOverlay`, NOT `allowOrgOverride`:
+ *
+ *  - `supportsOverlay: true` + `allowOrgOverride: false` → the ROLLED-BACK
+ *    OVERLAYABLE tier. Delete is now allowed. Cases below.
+ *  - `supportsOverlay: false` → `object` above all, whose overlay registers as
+ *    its own contributor LAYER (ADR-0029 D9) rather than merging, and whose
+ *    reset refusal is D9.6's declared, maintainer-approved cost. Unmoved, and
+ *    pinned by `packages/objectql/src/protocol-object-overlay-layer.test.ts`
+ *    on both topologies; re-pinned here from the other side of the boundary.
+ *
+ * ## Reverse verification — the four categories, declared BEFORE running
+ *
+ * Base for every direction below: `f16e54e1d` (`origin/main` at authoring —
+ * the commit that landed ADR-0029 D9 and its pins), never a moving `main`.
+ *
+ *  1. PREDICTED RED, MEASURED RED — every case in
+ *     `describe('the rolled-back overlayable tier — DELETE moves')`. With the
+ *     two gates restored these fail with `NOT_OVERRIDABLE` / 403 (env kernel)
+ *     and with the repository's `[NOT_OVERRIDABLE] '<type>' is not
+ *     allowOrgOverride in the registry` (control-plane kernel).
+ *  2. GREEN IN BOTH DIRECTIONS — GUARDS, NOT EVIDENCE. Explicitly labelled
+ *     `[GUARD]` in their names: the `object`-tier refusals, the create/update
+ *     refusals, and the `flow` (`supportsOverlay: false`) delete refusal.
+ *     They pass before and after the change; what makes them load-bearing is
+ *     the VARIANT experiment, not this run — see the PR body, where relaxing
+ *     the predicate for every artifact-backed type (dropping the
+ *     `supportsOverlay` condition) is measured turning the `object` pins red.
+ *  3. MISSED PREDICTIONS — one, kept in the file rather than tidied away:
+ *     `a delete with NO overlay row answers the no-op success instead of 403`
+ *     was written as a guard and measured RED on the project kernel. The
+ *     refusal ran before the probe, so an artifact-backed item with NOTHING
+ *     customized was refused too. Its comment carries the correction.
+ *     Predicted 56 red, measured 57.
+ *  4. UNMEASURED — none. Every assertion here was run in both directions; the
+ *     ones that cannot go red on this change say so in their own name, and
+ *     what makes THOSE load-bearing is the variant experiment in the PR body.
+ *
+ * Harness: the real write path over a stub engine (same shape as
+ * `protocol.code-only-types.test.ts`) — a gate INSIDE `deleteMetaItem` and one
+ * inside `SysMetadataRepository` cannot be tested against a harness that mocks
+ * either.
+ */
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+// [#5619] The producer's OWN write-verb dispatch decisions (#4550 delete /
+// #5480 update), so the fake engine below cannot accept a call ObjectQL
+// refuses. Imported from `@objectstack/metadata-core` and not from
+// `@objectstack/objectql`, which depends on this package.
+import { assertEngineDeleteDispatch, assertEngineUpdateDispatch } from '@objectstack/metadata-core';
+import { DEFAULT_METADATA_TYPE_REGISTRY } from '@objectstack/spec/kernel';
+import { ObjectStackProtocolImplementation } from './protocol.js';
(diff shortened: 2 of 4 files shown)


os-zhuang:

Four files, exactly as declared:

| file | | |
|---|---|---|
| `packages/metadata-protocol/src/protocol.ts` | modified | +119/−2 |
| `packages/metadata-protocol/src/sys-metadata-repository.ts` | modified | +74/−1 |
| `packages/metadata-protocol/src/protocol.legacy-overlay-delete.test.ts` | added | +694 |
| `.changeset/legacy-overlay-delete-rolled-back-types.md` | added | +48 |

`packages/spec` read but never changed. **`packages/objectql` absent from the diff entirely**, which is the mechanical proof that ADR-0029 D9's pins were not edited — the constraint this card carried above all others. No `content/docs/releases/`, no `docs/adr/**`. Changeset present and `Check Changeset` green; `skip-changeset` correctly not applied, since this ships a real behaviour change.

### CI, per job
25 check runs, all `completed`: **23 `success`**, **2 `skipped`** (`Build Docs`, `Console Pin Gate` — path-filtered). No failures. Named gates: `ESLint` success, `TypeScript Type Check` success, `Build Core` success, `Test Core` 3/3 + rollup success, `Dogfood Regression Gate` 3/3 + rollup success, `Dogfood Verify CLI`, `Temporal Conformance (live PG + MySQL)`, `Check Changeset`, `Check PR Size`, `ADR maintainer approval`, `No other open PR may claim the same issue` — all success.

### What I am accepting

**The tier boundary resolved into a registry-derived predicate rather than a hardcoded list, and that is the better answer.** I dispatched this with an enumeration (`permission` / `position` / `page` / `app` / `dataset` / `book`) and a standing instruction to STOP if the tiers could not be separated. They separate on the registry's own `supportsOverlay` flag — the types whose loader merges overlays at read are exactly the types where a stuck overlay is possible, and `object` declares `supportsOverlay: false`, so the carve-out cannot reach it structurally rather than by an exclusion someone has to remember. A hardcoded six-type list would have drifted the first time another type was rolled back.

**Both refusal points moved, which is what makes this a whole fix rather than half of one.** `deleteMetaItem`'s `environmentId` branch yields, and `SysMetadataRepository.delete` routes through the new `assertDeleteAllowed` carrying the same registry-derived rule — so the control-plane topology, which skips the first gate entirely because it has no `environmentId`, is released by the same rule rather than left refusing. That second path was the half most likely to be missed, and the pins cover both topologies.

**Create and update are verifiably untouched**: `put` still calls `assertAllowed` unchanged and `saveMetaItem`'s gates are not in the diff. The delete-only asymmetry is recorded in both gates' doc comments — including extending the save-side carve-out that promised "removing a row that predates this refusal is repair" to name the artifact-backed tier it never covered. That divergence between stated intent and enforced behaviour is what the card opened with, so closing it in the comment is part of the fix, not decoration.

### Reverse verification — the variant is what earns the ACCEPT

Reverted against the dev's own recorded base `f16e54e1d` (`git checkout <base> -- <2 source files>`, no `git stash`). Predicted 56 red, **measured 57 red / 44 passed**, with real output from both refusal points — project kernel `[not_overridable] … the type has not opted into per-org overlay writes` at `protocol.ts:11932`, control-plane `Failed to delete customization overlay: [NOT_OVERRIDABLE] 'permission' is not allowOrgOverride in the registry` at `:12152`.

**The variant verification is decisive and is exactly the damage case I required.** Relaxing the predicate for *every* artifact-backed type turned D9's own pin file to `2 failed | 14 passed` — precisely its two delete pins (`…and so is the RESET of one: deleteMetaItem answers NOT_OVERRIDABLE / 403`, and `a control-plane kernel refuses at the repository, and subtracts under the hatch`), plus 14 of this PR's own boundary guards. Variant reverted, D9's file back to 16/16. That proves the scoping is **necessary**, not merely sufficient — the over-broad fix is not hypothetical, it is measured, and it breaks exactly the two assertions the maintainer's D9 approval made binding.

**One missed prediction, reported rather than buried, and it found a second facet of the defect.** The guard "a delete with NO overlay row answers the no-op success instead of 403" went **red** on the project kernel, because the refusal ran *before* the probe for the row — so pre-fix, an artifact-backed item **with nothing customized at all** was refused too. The card described that ordering ("throws … before any probe for the row") without drawing out the consequence. Kept in the file with a corrected comment.

**Guards labelled as guards** (green in both directions, marked `[GUARD]` in the test names and the PR body): object-tier delete refusals on both topologies, every other `supportsOverlay: false` type, `put` still refusing `override-artifact` writes, `saveMetaItem` plain and org-scoped refusals, runtime-only delete still judged by `allowRuntimeCreate`. **Unmeasured: none.**

Suites all green with the consumption radius covered: `metadata-protocol` 1009, `objectql` 3066, `rest` 1278 (after clearing an AGENTS.md §9 unbuilt-dependency false red — correctly diagnosed as the stale-artefact trap rather than treated as a real failure), `runtime` 1906, `metadata` 593. Type-check debt measured at 63 and unchanged, zero in the new file.

### ⚠️ One scope note, stated so it can be vetoed cheaply

The delivered predicate also releases **`tool` and `skill`**, which my dispatch did not enumerate — they carry `supportsOverlay: true`, so they are the same registry shape and have the identical stuck-repair problem. I am accepting this as within the ruling: the maintainer's reasoning was general ("deleting an overlay restores the code-declared state; it is the narrowing direction and cannot widen anything"), and excluding two types that share the shape would leave the same defect alive for no reason. Flagging it explicitly because the ruling's text named the rolled-back six — if the intent was narrower, this is the line to object to, and it is one predicate to change.

### Filed, not fixed
**#7426** — `deleteMetaItem`'s catch re-wrap carries `status` forward but drops `code`, so a repository refusal reaches the caller as a 403 with `code: undefined` on the control-plane topology. That is not merely a new defect: it **explains** why D9's control-plane pin asserts a message substring while its project-kernel sibling asserts `code` + `status`. Correctly left out — fixing it changes the error envelope of every non-conflict failure on that path.

### Ruling
ACCEPT. Flipping out of draft, enabling auto-merge (SQUASH). Landing serialises behind the in-flight #4717 through the queue; the two are region-disjoint (`deleteMetaItem` + the repository gate here, `saveMetaItem`'s success return there).

---

os-zhuang:
**PM note — ACCEPT stands; the ready-flip is deferred on a GraphQL rate limit, not on anything about this PR.**

`update_pull_request` (draft → ready) is a GraphQL-only write and the quota is currently exhausted (`API rate limit already exceeded`). The review verdict above is unchanged and final; the mechanical half is queued.

Recording the pending state here rather than only in session context, so it survives a context reset or a seat handover:

| pending action | on PR | detail |
|---|---|---|
| 1. flip draft → ready | #7429 | GraphQL — blocked, retrying on a stepped timer |
| 2. enable auto-merge **SQUASH** | #7429 | GraphQL — after 1 |
| 3. `subscribe_pr_activity` | #7429 | after 2 |
| 4. landing double-read | — | unbounded `git log origin/main --grep='(#7429)'` + single-parent check, then ledger row + confirm #6960 closed + unsubscribe |

⚠️ **Re-verification required before executing, not just replay**: the ACCEPT is anchored to this PR's head at review time. If the head has moved when the quota returns, the gates get re-read before anything is flipped.

Nothing is required from the dev, and the PR needs no changes. Reads (REST) are on a separate quota and still working, which is why this comment posted.

---

Issue #7294, opened 2026-08-10
[finding] `POST /meta/:type/:name/publish` is a served REST route with NO spec declaration — the #5745 "declared = returned" discipline covers only the save door


os-zhuang:
Observation-class finding, measured while scoping #4717 on `origin/main` @ `ea8e8490f`. Nothing a user hits today — the route works and returns a sensible body. Filing plainly and letting triage grade it.

## What was measured

`publishMetaItem` is served over REST:

- `packages/rest/src/rest-server.ts:5468` registers `POST {metaPath}/:type/:name/publish`, and `:5488` awaits `publishMetaItem(...)` and hands the result to `res.json()`.
- Its success return is built at `packages/metadata-protocol/src/protocol.ts:9771-9798` and carries `success` / `version` / `seq` / `message`, plus the conditional `seedApplied` / `materializeApplied` / `projectionApplied`.

But the string `PublishMetaItem` appears **nowhere** under `packages/spec/src/`:

```
grep -rn "PublishMetaItem" packages/spec/src/     # zero hits
```

And the endpoint itself is not in the REST declaration — `packages/spec/src/api/plugin-rest-api.zod.ts` declares 6 endpoints with `category: 'metadata'`, and none of them has `handler: 'publishMetaItem'` or a `/publish` path (`grep -n publish` on that file: zero hits).

## Why this is the same class as #5745, one door over

#5745 closed exactly this gap for the sibling write door, and its own prose states the discipline it was establishing. `SaveMetaItemResponseSchema` (`packages/spec/src/api/protocol.zod.ts:456`) now declares the FULL save body, the endpoint declares `responseSchema: 'SaveMetaItemResponseSchema'` (`plugin-rest-api.zod.ts:824`), and a producer-side conformance gate keeps the two in step — `packages/objectql/src/save-meta-response-conformance.test.ts`, whose header reads:

> a future field added to the response, or an existing one dropped, turns this red instead of silently vanishing at parse.

The publish door has none of those three: no declaration, no `responseSchema` reference, no conformance gate. So `version` on the publish response is in the same undeclared state `version` on the save response was in before #5745 — and it is the same ADR-0008 OCC token, with the same "echo it back as `If-Match`" job.

## Why it is observation-class rather than a defect

[...]

Issue #5586, opened 2026-08-05
UnknownFilterTokenError 漏诊:含非 word 字符的类占位符串({TODAY()}、{current-user-id}、{30 days ago})绕过诊断,原样下发按字面串比较(17.0.0-rc.2)


yinlianghui:
Part of objectstack-ai/hotcrm#786

## 现象

`UnknownFilterTokenError` 的覆盖面止于占位符文法 `/^\$?\{([a-zA-Z0-9_]+)\}$/`。任何含**非 word 字符**的类占位符串——`{TODAY()}`、`{current-user-id}`、`{30 days ago}`、`{user.id}`——在 `classifyFilterToken` 第一行就 `return null`,**不被识别为令牌**,于是绕过诊断、原样下发给 driver 按字面串比较——正好落回该诊断本要消灭的失败模式(静默错误结果,而非显式报错)。

## 证据(pinned 17.0.0-rc.2,InMemoryDriver + 真 ObjectQL.find;crm_task 四行夹具,due_date 分别在 30 天前 / 1 天前 / 今天 / 7 天后)

```
classifyFilterToken:
  {today}            => {"kind":"date-macro","token":"today"}
  {TODAY}            => {"kind":"unknown","token":"TODAY"}     → 读路径 THROWS UnknownFilterTokenError ✅
  {TODAY()}          => null                                    → 不识别,原样下发 ❌

读路径:
  due_date <  '{today}'    =>  2 行(恰好两行逾期,解析正确)
  due_date <  '{TODAY()}'  =>  4 行(含 7 天后才到期的——字面串比较 + 字典序倒置,'2026-…' < '{')
```

即:拼错成不带括号的 `{TODAY}` 会得到显式报错(诊断达成目的),拼错成带括号/连字符/空格/点号的形态反而**静默返回错误行集**——后一类恰恰是作者从其它系统的宏语法(`TODAY()`、kebab-case、自然语言)迁移时最容易写出的形态。

## 参照:应用侧守卫比平台诊断更严,次序反了

hotcrm 的作者期守卫用更宽的 `/\{([^{}]+)\}/` 扫描 view filter 值,能逮住上述全部形态——所以 hotcrm 今天在 CI 层碰不到这个洞。但这层保护只覆盖一个应用仓库的作者期;运行时(API 调用方、其它应用)仍然裸奔。平台诊断的存在理由就是消灭「类占位符串静默按字面量比较」,现在的文法让最常见的错拼形态恰好漏网。

[...]

PR #7435, opened 2026-08-10
feat(spec,metadata-protocol): put the runtime authoring gate's advisory findings on the save response (#4717)

### LINKED ISSUE #4717: [P2/P3] 运行时授权门的 advisory findings 要进响应并被 Studio 渲染 + docs「the gate」补第四列(#4463 收尾)

os-zhuang:
Fixes #4717

The metadata half only, per the maintainer's 2026-08-07 migration ruling: advisory findings reaching the `saveMetaItem` response envelope. The lint "surfaces" generalisation (spec-tooling), the Studio rendering (`repo:objectui`) and the docs "the gate" fourth column are **not** in this PR.

---

## ⚠️ CROSS-SEAT DECLARATION — this PR edits `packages/spec`, and #6017 owns that surface

**Two files under `packages/spec/src/` are changed here by the `domain:metadata` seat.** Calling it out at the top rather than leaving it to the diff, because the maintainer's 2026-08-10T07:33Z ruling names **`domain:spec` (#6017)** as the actor for this declaration, and only the typist is different here — not the substance.

- The seat was asked at **07:48Z** and again at **08:58Z** and has not answered.
- The PM set the stated default under the **#5586 declare-and-proceed precedent** — the mechanism that seat itself documented at 06:41Z, citing PR #7306's two `packages/spec` files as a correct application. Those landed at 09:03Z with no objection raised.
- **The veto stays open to #6017 until merge.** If that seat wants the declaration back at any point, this PR is pulled or the spec half is handed over — no argument, no negotiation. The ruling's substance is followed identically either way.

The spec files, and nothing else under `packages/spec/src/`:

| file | what it declares |
|---|---|
| `packages/spec/src/api/protocol.zod.ts` | `RuntimeAuthoringIssueSchema` (new export) + `advisories` on `SaveMetaItemResponseSchema` (optional) + the `RuntimeAuthoringIssue` type alias |
| `packages/spec/src/type-alias-convention.pin.test.ts` | the ADR-0122 isomorphic pin for the new schema (`Iso837`), 825 → 826 |

---

## What was broken

#4463 put the shared author-time rule registry (`AUTHORING_RULES` — the same table `os validate` / `os build` / `os lint` run) on the runtime write path. It gated on `error` findings only. The **advisory** half was produced at `packages/lint/src/runtime-gate.ts`, walked into a `console.warn` deduped once per process per `type|name|rule|path` at `packages/metadata-protocol/src/runtime-authoring-gate.ts`, and then went out of scope. PR #4715 named the shape when it shipped: running a rule and discarding its conclusion is a smaller version of the hole the gate exists to close.

**The failing case is reachable today, not theoretical.** A `nightly_purge` flow whose only defect is a `delete_record` node declaring `multi: true` with no `filter` yields `errors = 0 / advisories = 1`. The gate allows it, the write succeeds, the row persists, the flow registers — and the author never learns that their nightly sweep deletes every row of the object on every run. A Studio tenant or an MCP/AI author has no `os lint` and no CLI config file for `sys_metadata` overlay rows, so for them this gate is not the weakest of four doors. It is the only one.

## What this changes

**This is an ADDED return channel, not a threaded value.** `evaluateRuntimeAuthoringGate` was typed `=> Error | null` and returned `null` on success — the success path had nowhere to put a verdict at all. It now returns a `RuntimeAuthoringVerdict` record `{ error, advisories }`. A record rather than a widened union, deliberately: both halves are always answered, and a caller cannot read one while forgetting the other exists.

**The element shape is declared once.** `RuntimeAuthoringIssueSchema` in spec is now the single declaration of the six keys (`rule` / `path` / `where` / `message` / `hint` / `severity`), and `metadata-protocol` re-exports it as its `RuntimeAuthoringIssue` instead of keeping a hand-written interface for the same shape next to a Zod schema. The 422's `issues[]` and the 2xx's `advisories[]` cannot drift into two dialects. Per the ruling: reuse it, do not mint a parallel shape.

**Emitted only when non-empty.** The key is omitted — never `[]` — so a clean save's response bytes are byte-for-byte what they were before, and a caller that ignores the field behaves exactly as today.

**`rulesRun` stays off.** The gate appends the gate-local `PLATFORM_SCHEDULE_CREATE_RECORD_ORG_MISSING` when `type === 'flow'`, so not every id it would list resolves in the lint registry; shipping the array would require the spec `.describe()` to state that the ids are *gate* ids. A field can be added later, not removed.

**Save door only, and the asymmetry is stated in the changeset.** The gate runs on both write doors on purpose (`saveMetaItem` and the draft→active promotion, so `?mode=draft` + publish is not a bypass — #4463 D1). The publish door does not carry this field; it rides behind #7294. The promotion door's call site simply ignores the new return value, which is why adding the channel could not change what that door does.

## File surface, file by file

**Contract (`packages/spec`) — the cross-seat half declared above**
- `packages/spec/src/api/protocol.zod.ts` — `RuntimeAuthoringIssueSchema`, `advisories` on `SaveMetaItemResponseSchema`, `RuntimeAuthoringIssue` alias.
- `packages/spec/src/type-alias-convention.pin.test.ts` — `Iso837` + the three count statements (pin list header, case title, `toHaveLength`).

**Producer (`packages/metadata-protocol`)**
- `packages/metadata-protocol/src/runtime-authoring-gate.ts` — the interface becomes a re-export of the spec type; new `RuntimeAuthoringVerdict`; all four return points converted. The deduped operator log is **kept**, it is simply no longer the only channel.
- `packages/metadata-protocol/src/protocol.ts` — `assertRuntimeAuthoringRules` returns `RuntimeAuthoringIssue[]` (throws unchanged); `saveMetaItem` captures it into a local and spreads it conditionally onto the success return.

**Tests**
- `packages/objectql/src/save-meta-response-conformance.test.ts` — 5 new cases through the REAL protocol / REAL engine / REAL `runRuntimeAuthoringRules`.
- `packages/spec/src/api/protocol.test.ts` — 7 new cases pinning the declaration itself.

**Generated artifacts — regenerated, never hand-edited**
- `packages/spec/api-surface/api.json`, `packages/spec/export-origins/api.json`, `packages/spec/authorable-surface/api.json`, `packages/spec/json-schema.manifest/api.json`
- `docs/audits/2026-07-unknown-key-strictness-ledger.counts.md` (`api/` 397 → 398; the row's `wire` verdict already covers a response envelope, so the prose needed no change)
- `content/docs/references/api/protocol.mdx`, `content/docs/references/index.mdx`

**Changeset**
- `.changeset/runtime-gate-advisories-on-save-response.md` — `minor` on both packages, and it states the save/publish asymmetry so it is not read as an oversight.

`content/docs/releases/` is untouched.

## Reverse verification

Every direction was **declared before the probe ran**, against a fixed base — `origin/main` @ `f16e54e1d7851770748c927c85db53aa73bbd722`, never a moving ref. No `git stash` at any point: the fix was taken out and put back with file copies inside my own worktree.

### 1. Predicted red, came out red

| # | probe | predicted | result |
|---|---|---|---|
| **R1** | delete the `advisories` line from `SaveMetaItemResponseSchema` (producer untouched) | RED — `strippedKeys` reports the key | ✅ `AssertionError: expected [ 'advisories' ] to deeply equal []`, `Tests 1 failed` / `9 passed` |
| **R2** | disable the producer's conditional spread (spec declaration untouched) | RED — the field never arrives | ✅ `AssertionError: Target cannot be null or undefined.`, `Tests 1 failed` / `9 passed` |
| **R4** | emit `advisories: runtimeAdvisories` **unconditionally** — the well-meaning `[]` refactor | the three absence pins RED, the five pre-existing #5745 cases GREEN | ✅ exactly that: `Tests 3 failed` / `7 passed`, all three reds mine |
| **R5** | rename the emitted key in the producer only (`advisories` → `advisoryFindings`) | RED | ✅ `Tests 1 failed` / `9 passed` |
| **R6′ (runtime half)** | drop `hint` from the shared element schema | RED — the narrowed element strips inside the array | ✅ `AssertionError: expected [ { …(5) } ] to deeply equal [ { …(6) } ]` |

**R4 is the load-bearing one.** It is the exact violation the ruling's point 2 forbids, and it demonstrates that the pre-existing conformance gate is structurally blind to it: all five #5745 cases stayed green while `[]` was being emitted on every save. Only the new absence pins caught it.

**R5, stated precisely:** the direction was right and the cause was the rename, but the *message* was not the one I predicted. I predicted `strippedKeys` would report `['advisoryFindings']`; in fact the `toHaveLength(1)` assertion fires first and the run stops there. Reporting the assertion that actually failed rather than the one I expected.

### 2. Missed predictions

**R1's first run was INVALID and I nearly recorded it as a pass.** I removed the spec field, rebuilt spec with output redirected to `/dev/null`, and got `Tests 10 passed`. The spec build had **failed**, so the test read a stale `dist` — the green was measuring the fix, not the counterfactual. Cause: never send stderr to `/dev/null` when the output is evidence. Re-run visibly, the cause turned out to be a **third diagnostic I had not predicted at all**: `build-schemas.ts`'s authorable-surface ratchet refuses the removal at *build* time, before any test runs —

```
❌ 1 authorable key(s) disappeared from the contract:
     - api/SaveMetaItemResponse:advisories
```

So once this lands, deleting the field is red at two independent layers, one of which never reaches a test runner. R1 above is the honest counterfactual: field **and** baseline line removed together, i.e. "the declaration was never added".

**R6′ missed its type-level half.** I predicted that narrowing the shared element shape (dropping `hint`) would turn `metadata-protocol` red, proving the re-export is genuinely load-bearing at the type level rather than cosmetic. It did **not**: the package's DTS build succeeded. Cause, measured rather than guessed — `packages/metadata-protocol` has **no `typecheck` script** (it sits in the `check:type-check-coverage` DEBT ledger) and tsup never type-checks (AGENTS.md, #4311). There is no tsc program over that file for a narrowed type to redden. The runtime half of the same probe does go red (table above), so the schema really is the one used at parse; the type-level claim is simply **not checked anywhere today**, and I would rather say so than let the re-export read as more guarded than it is.

### 3. Assertions green in BOTH directions — guards, not evidence

**The variant the previous dev flagged as explicitly unmeasured, now measured.** Does a *conditionally* added key leave the three pre-existing conformance cases green? **Yes — and in both directions**, which is the point: under R1 (no spec declaration at all) they were green; with the full fix they are green. They save clean `view` bodies, the key is absent, and `strippedKeys` is empty either way.

That is labelled as a guard **in the test file itself**, not only here — the case is literally named `GUARD (green either way): a clean view save is untouched by the new field`, and its docblock states that the existing gate **cannot** go red to report a missing or misspelled declaration. The reason it is written down rather than left out: a future reader seeing five green conformance cases would otherwise reasonably conclude the new field is covered by them. It is not. It is covered by the two directional cases added next to it.

### 4. Predictions left unmeasured

- **R6 (`=> Error | null` restored on the gate signature)** — not run. It would be a trivial compile error at the four converted return sites and carries no information the R2 probe does not already carry. Declared, not measured; not counted as passed.
- **The REST wire.** Every measurement here is at the protocol return. The conformance suite's existing header records that `rest-server.ts` hands this exact object to `res.json()` verbatim, so the protocol return IS the wire body — I did not re-derive that with a live server, and I am not claiming an HTTP-level measurement I did not take. The "still a 200" assertion in the new case is the protocol-level one: `saveMetaItem` throws on the gated half, so reaching the success assertions at all is the statement that this write was not refused.
- **The publish door** was deliberately not probed. It is out of scope by the ruling and #7294 owns it.

## Tests

All run in the worktree, real output:

```
packages/spec              Test Files  365 passed (365)   Tests  9544 passed (9544)
packages/metadata-protocol Test Files   68 passed  (68)   Tests   908 passed  (908)
packages/objectql          Test Files  172 passed (172)   Tests  3071 passed (3071)
packages/lint              Test Files   70 passed  (70)   Tests  1852 passed (1852)
```

`packages/lint` is included because `authoring-rule-wiring.test.ts` scans the gate file's imports, and this PR adds one.

Typecheck — `pnpm --filter @objectstack/spec --filter @objectstack/objectql typecheck` clean (`metadata-protocol` declares no `typecheck` script; see §2 above). ESLint clean on all six changed source files. Gates run locally, all green: `check:api-surface`, `check:export-origins`, `check:authorable-surface`, `check:strictness-ledger`, `check:docs`, `check:spec-parsed-alias` (`826 pinned isomorphic`), `check:route-envelope`, `check:error-code-casing`, `check:engine-double-contract`, `check:meta-type-normalized`, `check:adr-anchors`, `check:doc-authoring`, `check:type-check-coverage`, `check:changeset-gate-self-tests`, `check:nul-bytes`.

## Notes for review

- The **refusal path is asserted on its envelope** — `code` AND `status` (`{ code: 'INVALID_METADATA', status: 422 }`), never a bare `toThrow()`. That body already threw before this change, so a throw-only assertion would have been permanently green and could not have shown that the refusal did not move when the success path grew a channel.
- The **escape hatch does not promote gating findings into the advisory channel.** `OS_ALLOW_UNLINTED_METADATA_WRITES=1` converts a refusal into a log; `advisories` means "did not block this write", and a finding that only failed to block because an operator set a migration flag is not that. Widening it would put `severity: 'error'` entries on a 2xx. Stated in the code at the return site.
- The **flow fixtures are org-agnostic on purpose**: the registry declares `allowOrgOverride: false` for `flow`, so an org-scoped write is refused `403 NOT_OVERRIDABLE` long before the gate runs. `runAs: 'system'` on the advisory fixture is load-bearing too — without it `flow-runas-unscoped` fires at `error` and the case would pass for the wrong reason (a refusal wearing an advisory's clothes). Both are commented in the test.
- Successor-pricing from #6285 confirmed unchanged, not re-derived: that rule emits `error` only, so it adds nothing to the advisory channel.

---

---


os-zhuang:

### File surface
14 files, which reconciles with the report's "6 changed source files" (that figure is its ESLint scope — 3 source + 3 test):

| group | files |
|---|---|
| **source (3)** | `packages/spec/src/api/protocol.zod.ts` (+64) · `packages/metadata-protocol/src/runtime-authoring-gate.ts` (+74/−26) · `packages/metadata-protocol/src/protocol.ts` (+43/−7) |
| **tests (3)** | `packages/objectql/src/save-meta-response-conformance.test.ts` (+206) · `packages/spec/src/api/protocol.test.ts` (+102) · `packages/spec/src/type-alias-convention.pin.test.ts` (+17/−3) |
| **generated (7)** | `packages/spec/{api-surface,authorable-surface,export-origins,json-schema.manifest}/api.json` · `content/docs/references/api/protocol.mdx` · `content/docs/references/index.mdx` · `docs/audits/2026-07-unknown-key-strictness-ledger.counts.md` |
| **changeset (1)** | `.changeset/runtime-gate-advisories-on-save-response.md` |

All seven generated entries are `os-regen`-family artifacts and are consistent with the source change — `Check Changeset`, `Spec property liveness` and `Build Docs` are all green, so they were regenerated rather than hand-edited. No `content/docs/releases/`, no `docs/adr/**`.

### CI, per job
26 check runs, all `completed`: **25 `success`**, **1 `skipped`** (`Console Pin Gate`, path-filtered). Both required gates read on their own conclusions: **`ESLint` success (11:09:35Z)**, **`TypeScript Type Check` success (11:21:24Z)**. Also green: `Spec property liveness`, `Build Core`, `Build Docs`, `Test Core` 3/3 + rollup, `Dogfood Regression Gate` 3/3 + rollup, `Dogfood Verify CLI`, `Temporal Conformance`, `Check Changeset`, `Check PR Size`, `ADR maintainer approval`, `No other open PR may claim the same issue`.

### The four ruled constraints, each verifiable in the diff
`advisories` optional on `SaveMetaItemResponseSchema` reusing the D3 element shape ✓ · emitted **only when non-empty**, never `[]` ✓ · `rulesRun` off ✓ · save door only, with the publish-door asymmetry stated in the changeset ✓. The channel exists where I priced it: `evaluateRuntimeAuthoringGate` returns a `RuntimeAuthoringVerdict` record instead of `Error | null`, so the success path finally has somewhere to put a value.

**One thing better than I asked for**: rather than adding the field alongside the existing hand-written `RuntimeAuthoringIssue` interface, `RuntimeAuthoringIssueSchema` is declared once in `packages/spec` and `metadata-protocol` **re-exports** it. That removes a parallel definition instead of creating a second one — the contract-first direction, done without being told.

### The cross-seat `packages/spec` declaration
Carried under the #5586 declare-and-proceed precedent, in its own top section of the PR body, naming **#6017** with the veto open until merge. That is the shape I committed to on the card and on #6017, and it is prominent rather than buried. If that seat wants it back at any point before merge, this PR yields.

### Reverse verification — the strongest of this term

**Predicted red, came out red**, five probes with real messages: R1 (spec field removed) `expected [ 'advisories' ] to deeply equal []`; R2 (producer spread disabled) `Target cannot be null or undefined.`; R5 (producer-only key rename) 1 red; R6' (drop `hint` from the shared element) shape mismatch. R5's caveat is reported honestly — direction right, but the first failing assertion was `toHaveLength(1)`, not the `strippedKeys` message predicted.

⭐ **R4 turns the ruling's caveat from an argument into a measurement.** The maintainer's ruling accepted, as a stated cost, that the field must be conditional and therefore the existing conformance gate would *not* go red to remind anyone the declaration was missing. R4 emitted `advisories: []` unconditionally — the ruled-forbidden refactor — and measured **3 failed / 7 passed: the three new absence pins red, and all five pre-existing #5745 cases GREEN**. That is direct evidence that the existing gate is structurally blind to this class, which is exactly why the declaration had to be deliberate. The caveat is no longer something we assert; it is something we measured.

⭐ **The previously-unmeasured variant is now measured.** I carried forward, twice, that "does a conditionally-added key leave the three existing conformance cases green" had been flagged unmeasured rather than assumed. It is now measured in both directions — green with the declaration and without it (R1) — and the case is named in the file `GUARD (green either way): a clean view save is untouched by the new field`, with a docblock stating the existing gate cannot go red on a missing declaration. Labelled as a guard, not counted as evidence.

⭐ **A probe was caught invalidating itself, and reported.** R1's first run redirected the spec build to `/dev/null`; the build had **failed**, the test read stale `dist`, and it reported a false `10 passed`. That is this lane's operational note 24 (never swallow stderr on a command whose output is evidence) and the stale-artefact trap of #7000, landing on the dev — and the dev caught it, said so, and re-ran visibly. The re-run then exposed a **third diagnostic nobody predicted**: `build-schemas.ts`'s authorable-surface ratchet refuses the removal at **build time** (`❌ 1 authorable key(s) disappeared from the contract`), before any test executes. A run that had accepted the first green would have shipped with a false confidence reading and never learned the build-time gate exists.

⭐ **A missed prediction produced a real coverage finding rather than an excuse.** R6' narrowed the shared element and metadata-protocol did **not** go red. The measured cause: that package has no typecheck script and `tsup` never type-checks (#4311), so no `tsc` program exists over that file — meaning **the type-level binding of the re-export is not checked anywhere today**. Reported plainly instead of dressed up. I am recording this rather than filing a new card, because it is an instance of the known #4311 gap, not a new one; the runtime binding is real and pinned, the type-level one rests on #4311 being closed.

**Unmeasured, stated and not counted as passed**: R6 (restoring `=> Error | null`) not run, no information beyond R2; the REST wire not driven against a live server, so the "200" claim is protocol-level; the publish door deliberately not probed (it rides behind #7294 by ruling).

### Ruling
ACCEPT. Flip to ready + auto-merge (SQUASH) is queued behind a GraphQL rate limit along with #7429's — reads are on a separate quota, which is why this review posted. ⚠️ Two landing notes for whoever executes the queue: this PR touches `os-regen`-family generated artifacts, so if it is rebuilt against a moved `main` the regeneration must be redone wholesale rather than text-merged; and `type-alias-convention.pin.test.ts` carries a count assertion that any concurrent spec PR will conflict with — an ejection there is the expected shape, re-resolved by recounting from the merged file rather than by arithmetic on either side's receipt.

---

Issue #6170, opened 2026-08-07
release workflow: publish pushes tags + npm but its version commit never reaches main — twice now (rc.3 c6a52d3, rc.4 a10cbc77); landing the commit must be part of the publish lane


hotlong:
## The recurring defect

Two releases, same shape:

| release | publish commit | published | landed on main? | cleanup cost |
|---|---|---|---|---|
| 17.0.0-rc.3 (2026-08-03) | `c6a52d3` | 69 tags + npm ✅ | ❌ never | #6135 → #6149 (68-file field reconciliation, 4 days later) |
| 17.0.0-rc.4 (2026-08-07 04:17Z) | `a10cbc77` | 69 tags + npm ✅ | ❌ never | #6169 (merge-back, same day) |

The lane that runs `changeset version && changeset publish` pushes **tags** (unprotected) and publishes to npm, but the version commit itself never reaches protected `main` — presumably the bot's direct push is rejected by branch protection and the failure is swallowed. Verify against the run logs around 2026-08-07T04:17:39Z.

Left unfixed, every publish poisons the next one: main keeps stale versions, `changeset-release/main` recomputes an npm-occupied version, and someone hand-reconciles after the fact. The compiled CHANGELOGs and `pre.json` bookkeeping live only on orphaned commits until then.

## Second defect in the same lane: it bypasses the release gates

`check:objectui-pin-fresh` is **required on the Version Packages PR (#4935)** precisely so a release cannot ship with a stale console pin (#3340, after v16 silently lost 4 frontend changes). The 04:17Z run did not go through #4935, so the gate never ran — and rc.4 shipped `.objectui-sha = f995a452`, dropping the 7-changeset objectui window (#6159) from the release record. A gate that is required on a PR the lane can skip is not a gate.

## Options

1. **Publish lane lands its own commit**: instead of pushing `main`, push a branch + auto-open a PR (auto-merge via the queue). Mechanical, keeps the dispatch lane usable.
2. **Forbid the bypass**: publishing happens ONLY via merging the Version Packages PR (#4935 flow) — where the required checks (changeset gate, objectui-pin-freshness) actually run. The dispatch/manual lane, if kept at all, refuses to publish when its base ≠ current main or when the freshness gate would fail.
3. Both (2 as policy, 1 as the belt for whatever lane remains).

Recommendation: **(3)**. The maintainer picks; whichever option, the invariant is: *a publish that does not land its bookkeeping commit is a lie to the next release*.

## Related

[...]

Issue #5960, opened 2026-08-06
sdui.manifest.json 的来源未定:声明一致性 ratchet 目前只在手工 `pnpm sdui:manifest` 时跑,CI 里从来不跑(#4690 的遗留决定)


baozhoutao:
在 #4690(把 `check:react-declaration-parity` 的静默 skip 改成报错退出)里实测出来的遗留问题,按 PM 裁决 (c) 单独记一笔。**本条只记录测量与选项,不含实现主张。**

## 事实(#4690 的实测,2026-08-06,origin/main 72bd873)

spec↔registry 声明一致性 ratchet 需要 objectui 的 `sdui.manifest.json`。它在本仓的可得性:

| 候选来源 | 实测 |
|:--|:--|
| 签入副本 | 无。`packages/console/` 只跟踪 4 个文件,`dist/` 在 `.gitignore` 里;`git ls-files` 里没有任何 `sdui.manifest.json` |
| `scripts/build-console.sh`(ci.yml 会跑) | **故意不产出**。收尾注释:manifest 需要真实浏览器,console 构建不能拖进浏览器依赖 |
| 已发布的 `@objectstack/console` | 无。16.1.0 tarball 共 513 个文件,`sdui` 命中 0;`dist/manifest.json` 是 PWA manifest。→ CLI `resolveSduiManifest()` 的 console 兜底路径对任何 npm 安装用户都解析不到 |
| 任何 workflow | 没有一个跑 `pnpm sdui:manifest`;除 showcase-smoke(无关)外没有装 Playwright |

唯一生产者:objectui 的 `scripts/dump-public-manifest.mjs` —— Playwright chromium 打开构建后 console 的 `dev/manifest-dump.html`,读 `window.__MANIFEST`。包装脚本是 `scripts/gen-sdui-manifest.sh`(`pnpm sdui:manifest`),它按 `.objectui-sha` 构建 objectui、dump manifest、然后跑 ratchet。

**净结果:ratchet 只在有人手工跑 `pnpm sdui:manifest` 时才跑。** #4690 已经让「没 manifest」从静默退 0 变成报错退 1,所以现在不会再有人误以为它跑过了;但它在自动化里跑到的次数仍然是零,registry 侧的声明漂移在实践中没有被 ratchet 住。

## 待决定(带成本,不宜由开发 agent 猜)

[...]

Issue #3340, opened 2026-07-20
Release process: prevent frontend (objectui) changesets being missed when the console pin lags


os-zhuang:
## Background / incident

While curating the v16 release page (#3322), the Console section was written from the `@objectstack/console` **bundle-bump summaries**. Those summaries only cover up to the pinned objectui SHA (`.objectui-sha`). At that moment objectui `main` was **4 commits / 21 pending changesets ahead** of the pin, so **4 frontend changes — two of them `minor` features — were absent from the page** (fixed in #3331 by re-sourcing from objectui's own changesets):

- objectui#2701 `import-wizard-auto-policy` (minor) — Console side of framework#3236
- objectui#2708 `schema-fields-keyvalue` (minor) — Console side of framework#3304
- objectui#2707 `action-param-upload-guard` (patch) — ADR-0059 follow-up
- objectui#2706 `system-field-classifier-consolidation` (patch) — #2702 follow-up

This is a **structural gap**, not a one-off oversight.

## Root cause — three blind spots in the pipeline

```
objectui main ──(bump-objectui.sh pins SHA)──▶ .objectui-sha ──▶ @objectstack/console changeset ──▶ vN.mdx Console section
     ① can lag                                  ② summarizes only up to the pin        ③ written from the bump summary
```

1. **`.objectui-sha` can lag** — a release can be cut while objectui `main` has newer commits / pending changesets.
2. **The console bump changeset only covers up to the pin** — anything merged in objectui after the last bump never enters the changesets pipeline for that release.
3. **`scripts/check-release-notes.mjs` only checks "a page exists per major"** — it does **not** verify the pin is current, nor that frontend changesets are covered. So a lagging pin is invisible to both the changelog pipeline and the drift guard.

### P0 — pin-freshness gate (the only thing that hard-blocks this)
Add `scripts/check-objectui-pin-fresh.mjs`, wired as a required check on the Version-Packages / release PR. Fail when:
- objectui's latest release tag (or `main` HEAD) is **ahead of** `.objectui-sha`, or
- objectui has **pending `.changeset/*.md`** not yet bundled (list them).

CI can read objectui state via `git ls-remote` + a lightweight API call.

### P1 — auto-bump to structurally shrink the lag
A scheduled (or on-objectui-merge) job that runs `scripts/bump-objectui.sh` whenever objectui `main` advances and opens a PR. If the pin is (near-)always current, the bump summary is always complete and manual reconciliation disappears. **This is the root-cause fix.**

[...]

PR #7453, opened 2026-08-10
feat(release): cut-rc — snapshot one-click rc release (#7447)

### LINKED ISSUE #7447: release: snapshot one-click rc cut — a `cut-rc` workflow that bumps, versions, gates and publishes from ONE SHA, ending the two-treadmill freeze dance
**Maintainer ruling (2026-08-10, directed in session `session_01BPWqbmEFU8gJepBJTHESXd`): approved for immediate implementation and dispatch.** Background: cutting rc.6 today required chasing objectui `main` across four pin-bump laps (`8aad9fd`→`9b9fa49`→`cfeb378`→`1b6188d`→`bcd3e02`, every lap overtaken) and would have needed ~40 min of coordinated freezes across two repos. The root causes are design, not discipline: (1) the release path checks pin freshness against a *moving* objectui HEAD; (2) the standing Version Packages PR (#6208) is force-refreshed on every main push, so its CI can never finish on a busy day, and the version→publish gap invites content/record skew.

## The design (adopted)

New workflow `.github/workflows/cut-rc.yml` — **rc prereleases only**. One human dispatch, one snapshot, zero freezes:

1. **Trigger**: `workflow_dispatch` ONLY, input `version` = the expected resulting version (e.g. `17.0.0-rc.6`), `environment: release`, `concurrency` group `cut-rc` (no parallel cuts). This preserves the 2026-08-07 ruling verbatim: no push, bot, queue or schedule can synthesize the event; the human types the version.
2. **Snapshot**: checkout objectstack `main`, record `SNAPSHOT_SHA`. Resolve objectui `main` HEAD **once** (`git ls-remote`), record `OBJECTUI_SHA`. Everything downstream uses these two values; both repos may keep moving freely.
3. **Pin bump**: clone objectui (full enough for the digest walk — see traps below), run `scripts/bump-objectui.sh $OBJECTUI_SHA` with `OBJECTUI_ROOT` set, then `pnpm objectui:build`, `playwright install chromium-headless-shell` (allowed in CI — this is not a dispatch container), `pnpm sdui:manifest` (the ADR-0082 D4 ratchet; mandatory second half of every pin move, #5960), `pnpm check:console-sha`.
4. **Pin self-consistency, NOT liveness**: inside the run, assert `.objectui-sha == $OBJECTUI_SHA` (inline compare). ⛔ Do NOT call `check:objectui-pin-fresh` here and do NOT change that script's semantics — objectui moving during the run must not fail the cut. The #3340 invariant this preserves is "everything shipped is covered by the changeset record" (true by construction: the bump changeset covers OLD_PIN..$OBJECTUI_SHA); commits past the snapshot are simply next release's record.
5. **Pre-version gates** (they read `.changeset/*`, so they run BEFORE versioning): `check-adr-0087-registration`, `check-changeset-no-major`, `check-empty-changeset`, `check-nul-bytes`.
6. **Version**: `pnpm run version` (⚠️ the repo script — `changeset version && sync-protocol-version.mjs && sync-template-versions.mjs` — never raw `changeset version`). Pre mode (`.changeset/pre.json`, tag rc) is committed state and applies automatically. Then assert computed version (`packages/cli/package.json`) == the dispatch input; mismatch fails before anything irreversible.
7. **Land the version commit on `main` FIRST, publish it SECOND** — preserving #6170's invariant ("the publish lane may only ever ship a commit that is ALREADY on main") to the letter:
   - Primary route: push the single squashed version commit (bump + version output; file surface is `package.json`s / CHANGELOGs / `.changeset/*` / `.objectui-sha` — disjoint from lane PRs by construction) directly to `main` using a bypass credential, with a bounded fetch-rebase-retry loop for races. **Admin prerequisite (maintainer, one-time)**: add a bypass actor for this push — either the Actions app on the ruleset bypass list scoped to this workflow, or a fine-grained PAT stored as `RELEASE_PUSH_TOKEN` for an account on the bypass list.
   - If bypass turns out impossible in this org: STOP and report `needs_decision` with the measured constraint — the fallback (publish from a release ref + catch-up PR) is the rc.3/rc.4 accident shape (#6170) and is NOT to be adopted silently.
8. **Publish**: `changeset publish` + tag push from the just-landed main commit, `NPM_TOKEN` via the `release` environment — mirror `release.yml`'s publish job for secrets/guard patterns (read its `publish:` job before writing yours; its dispatch guards are the house style).
9. **Docs**: `docs/releases-maintenance.md` gains a "cutting an rc (snapshot flow)" section; the existing Version-PR flow is re-scoped as **GA only** (board-clearing, #7275-A cut precondition, human changelog review all stay there). One cross-reference line, no edits to `release.yml`'s lanes.

## Measured traps the implementer inherits (all hit today, all with receipts in session)

- `bump-objectui.sh` with no sha argument pins the LOCAL objectui checkout's HEAD — stale by construction (#7308 note 1). Always pass the resolved sha.
- The changeset digest exits 2 on an unwalkable shallow range; deepen the objectui clone until OLD_PIN is reachable (`--depth=250` sufficed at ~70 commits; `fetch-depth: 0` is simplest in CI).
- A fresh workspace must build local dists before `pnpm objectui:build`: the console's vite alias table (objectui#4103) resolves `@objectstack/client` → local `packages/client/dist`, whose DTS needs `@objectstack/core` built. `pnpm --filter "@objectstack/client..." build` (dependency closure) is the fix; `@objectstack/spec` build is also required for generated-artifact reads.
- `pnpm sdui:manifest` needs a Playwright chromium; in CI install it normally. (The scratchpad symlink workaround in `docs/releases-maintenance.md` is for agent dispatch containers only.)
- Expected ratchet state today: 62 spec-only divergences == accepted baseline, 0 missing blocks — four consecutive laps confirmed no declaration drift from objectui `8aad9fd..bcd3e02`.

## Boundaries (⛔)

- No second publish path reachable without `workflow_dispatch` + `environment: release`. The 2026-08-07 ruling and `release.yml`'s lane-split comments are binding context — read them first.
- `release.yml`'s three existing lanes and `check-objectui-pin-fresh.mjs` semantics are OUT of scope — GA keeps them unchanged.
- `content/docs/releases/**` untouched (standing rule).
- Releases nothing → `skip-changeset` label on the PR, per the Check Changeset gate's route 2.

## Acceptance

- Workflow YAML passes actionlint/self-review; every guard failure path exits before `changeset publish`.
- A dry-run mode (`inputs.dry_run`) that stops after step 6 and uploads the would-be version diff as an artifact — so the first real dispatch is not the first execution.
- Docs section written; PR body walks the sequence and names the admin prerequisite for the maintainer.

Refs: #6170 / #3340 (invariants), #5960 (ratchet on-demand ruling), #7308 (previous bump lap + traps), #7275 (GA cut precondition, stays), #6162 (steward bump cards — unaffected; they keep feeding the advisory pin lag), #4898 (empty-changeset stall), 2026-08-07 manual-publish ruling (release.yml header).

os-zhuang:
Fixes #7447

A `workflow_dispatch`-only lane that bumps the objectui pin, versions, gates and publishes an rc **from one snapshot**. Cutting rc.6 today needed four pin-bump laps (`8aad9fd` → `9b9fa49` → `cfeb378` → `1b6188d` → `bcd3e02`, every lap overtaken) and would have needed ~40 min of coordinated freezes across two repos. The snapshot is the whole idea: objectstack `main` and objectui `main` HEAD are each resolved **exactly once**, and every downstream step reads those frozen values — so both repositories may keep moving for the entire run.

**File surface:** `.github/workflows/cut-rc.yml` (new) and `docs/releases-maintenance.md`. `release.yml`'s three lanes, `scripts/check-objectui-pin-fresh.mjs` and `content/docs/releases/**` are untouched.

## The sequence

| # | Step | Note |
|---|---|---|
| 1 | Guard event / ref / version shape | dispatch-only, from `main`, `X.Y.Z-rc.N` only |
| 2 | Checkout `main` at full depth, record `SNAPSHOT_SHA` | `fetch-depth: 0` |
| 3 | Guard the train: `pre.json` is `mode: pre` / `tag: rc`, and the requested version is not the one main already carries | rc-only asserted against committed state, not the string |
| 4 | Resolve objectui `main` HEAD once via `git ls-remote`, full-clone it | unreachable remote is red, never green |
| 5 | `bump-objectui.sh $OBJECTUI_SHA` (always explicit), assert `.objectui-sha == $OBJECTUI_SHA` | self-consistency, **not** liveness |
| 6 | `pnpm objectui:build`, `check:console-sha`, playwright install, `pnpm sdui:manifest` | ADR-0082 D4 ratchet — this lane moves the pin, so it owes the ratchet |
| 7 | Pre-version gates | `check-changeset-fixed`, `check:nul-bytes`, and the three `.changeset` gates |
| 8 | `pnpm run version` + assert computed version == dispatch input | the repo script, never bare `changeset version` |
| 9 | Build ONE squashed version commit, allowlist-enforced | dry run stops here |
| 10 | Push to `main`, confirm ancestry, publish from the landed commit | #6170 to the letter |

## Invariants preserved

- **2026-08-07 manual-publish ruling** — dispatch-only trigger plus `environment: release`. A guard step also refuses any non-`workflow_dispatch` event, so a future edit that adds a `push:` trigger fails the run instead of publishing.
- **#6170** — the version commit lands on `main` first and is published second. Nothing is ever published from a ref that existed only inside the run.
- **#3340** — `check-objectui-pin-fresh.mjs` is unchanged and still enforces on the GA publish path. This lane deliberately does **not** call it: liveness is precisely what a snapshot gives up, and re-checking it would restore the race the lane exists to remove. The real invariant ("everything shipped is covered by the changeset record") holds by construction — the bump changeset covers `OLD_PIN..SNAPSHOT`, and objectui commits past the snapshot are next release's record.
- **#5960 / ADR-0082 D4** — the ratchet runs here, because this lane is now a route that moves the pin. Installing a Playwright browser is fine on a lane that runs a few times a month; the #5960 ruling is about not putting it on every matching PR.
- **#4731 / #7308** — the objectui clone is deliberately full-depth. A shallow clone makes the digest degrade to a tip-subject-only changeset **and still exit 0**, which would cost the cut its frontend history silently.

## Two things I measured rather than assumed

**1. The version pass's real file surface.** I ran a full `pnpm run version` locally on this train (1704 pending changesets) and captured the result: **76 modified `package.json`, 76 modified `CHANGELOG.md`, and `.changeset/pre.json`. Zero deletions, zero untracked files.** Computed version: `17.0.0-rc.6`, matching the issue's prediction.

Worth recording because it corrects the issue's stated surface: in **pre** mode `changeset version` does **not** delete the consumed changesets — it records them in `pre.json` and they are removed at `changeset pre exit`. So a cut never races a lane PR over a `.changeset/*.md` file it wants to keep, which is a stronger disjointness guarantee than the issue claimed.

That measurement is what the commit-building step's allowlist is derived from, and the allowlist is **enforced, not described**: only allowlisted paths are staged, every staged path is then re-checked against the same rule, and the worktree is asserted clean. If `pnpm run version` ever grows a new output the run fails loudly there instead of pushing a surprise to `main`. Verified in both directions — all 162 real surface paths accepted; `packages/objectql/src/engine.ts`, `.github/workflows/ci.yml`, `scripts/*.mjs`, `content/docs/releases/v17.mdx`, `pnpm-lock.yaml` and `packages/console/dist/**` all blocked.

**2. The three `.changeset` gates are forward-only DIFF gates, so they needed a base.** Their `--list` modes are audits that always exit 0 (`--audit-stock` says in its own header that a non-zero exit there would make it a gate, which it is not), so wiring one of those in would have been a gate that cannot fail. I hand them `--base "$SNAPSHOT_SHA"` instead, which judges this run's own contribution — the `@objectstack/console` changeset the pin bump just emitted. That is a real gate: the digest resolves its bump level from objectui's declarations, so an objectui `major` would otherwise promote all ~70 packages, and a digest bug emitting empty frontmatter would stall the release silently and greenly (#4898).

`$SNAPSHOT_SHA` is a pinned sha, which #6129 warns against — but that warning is about `pr-automation.yml`'s **frozen** `base.sha`, a value that goes stale as main moves under a PR. Here the snapshot is by construction the parent of the commit being judged, so it *is* the merge base and cannot go stale. I also confirmed the three gates' self-tests scope their "exactly one call site" wiring assertions to `pr-automation.yml` / `lint.yml`, so the new call sites do not trip them — all 377 assertions pass with this branch applied.

## Admin prerequisite (one-time, maintainer-only)

The lane pushes the version commit straight to `main`, so **the pushing identity must be on main's ruleset bypass list** — `contents: write` alone does not put it there. Either:

- **(a)** add the GitHub Actions app to the ruleset's bypass list, ideally scoped to this workflow, and leave the secret unset; or
- **(b)** create a fine-grained PAT with `contents: write` on this repository for an account already on that bypass list, and store it as the repository secret `RELEASE_PUSH_TOKEN`.

The workflow uses the secret when present and the Actions identity otherwise, so configuring **either** is enough and neither needs a workflow edit. Until one is configured the push step fails with exactly that message and **nothing is published** — the failure is before any irreversible step, and it is distinguished from a network fault in the error text. I did **not** implement the catch-up-PR fallback: that is the rc.3/rc.4 accident shape (#6170) and the issue rules it out.

Direct-to-main push is **not** structurally impossible here — it is an ordinary ruleset bypass configuration — so this is a prerequisite to satisfy, not a `needs_decision`.

## What I could and could not verify locally

**Executed, real output:**

- `node scripts/check-workflow-status-functions.mjs --self-test` + real scan — 34 assertions; scan covered 24 workflow files including this one, 43 jobs, 9 job-level `if:` reading `needs.*.outputs.*`, all naming a status function.
- `node scripts/check-required-contexts.mjs` — pass (this workflow registers no required context; it is dispatch-only).
- `pnpm check:nul-bytes` — 6830 files, no raw control bytes; plus a targeted `grep -naP` self-scan over both changed files.
- `pnpm check:changeset-gate-self-tests` — 377 assertions across all three gates, green with the new call sites present.
- The three gates run with the **exact argument shapes the workflow uses**, in both the real-diff case (`--base HEAD~5`) and the empty-diff case (`--base HEAD`, i.e. the "objectui had not moved, no bump commit" path). All exit 0.
- `bash scripts/bump-objectui.sh --help` — confirms the `bump-objectui.sh SHA` positional shape the workflow relies on.
- `pnpm run version` end to end — the measurement above; the worktree was reset afterwards.
- YAML parsed with the repo's own `yaml` package: 1 trigger, 2 inputs, 1 job, 37 steps, no step name containing `": "` (the nested-mapping trap).
- Every multi-line `run:` block extracted and `bash -n` syntax-checked — 20 blocks, all clean.
- `pnpm check:` for `node-version`, `adr-anchors`, `adr-links`, `docs-audit-scope`, `release-notes`, `role-word`, `org-identifier`, `merge-driver`, `quick-reference-counts`, `objectui-changeset`, `doc-authoring`, `empty-changeset` — all pass.

**Not executable locally, and stated plainly:** `actionlint` and `shellcheck` are not installed in this container, so the YAML review is the parse above plus manual review against `release.yml`'s patterns, and the shell review is `bash -n` plus manual review. **More importantly, I cannot execute the workflow end to end** — no runner, no `release` environment, no npm or push credentials. Nothing here proves the live behaviour of the push-retry loop, the Playwright install against objectui's build tree, or the publish steps.

That is exactly what `dry_run` is for, and it is why I made it stop **later** than the issue's acceptance criterion asks. It stops *after* the local squashed commit rather than after the version step, so the squash, the allowlist enforcement and the version assertion have all really executed, and the uploaded artifact is byte-for-byte the commit a real run would push. **Please dispatch with `dry_run` checked before the first real cut.**

## Deliberate non-inclusions

- **No runtime image job.** `release.yml`'s `release-integrity` lane runs on every push to `main` and requests the image once the version is on npm — the #4900 repair path doing its job, within a merge or two on this repo. Duplicating it here would be a second copy of an existing repair path; `docker-publish.yml` can be dispatched for one immediately. Named in the workflow header, the docs and the run summary rather than left to be discovered.
- **The dry-run guard is written twice** — a step-level `if:` and an in-shell assert inside each irreversible step. A dropped `if:` in a future edit would otherwise publish from a run a human asked not to publish.

Releases nothing (workflow + docs only) → `skip-changeset`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BPWqbmEFU8gJepBJTHESXd)_


os-zhuang:
**PM review (dispatching seat, session `session_01BPWqbmEFU8gJepBJTHESXd`): accepted in substance.** I read the workflow end to end, not just the PR body. Three things I checked specifically because they are where this design could have gone wrong, and all three are right:

- **The pin gate is self-consistency, not liveness** (`.objectui-sha == $OBJECTUI_SHA`), with `check-objectui-pin-fresh.mjs` untouched and still enforcing on the GA path. That is the ruling exactly — re-checking liveness inside the run would have restored the race the lane removes.
- **The push→publish ordering keeps #6170 to the letter**, and the rebase retry loop **re-earns the version assertion after each rebase** rather than inheriting it. That detail was not in the ruled design and it is the right call: another version commit landing underneath would otherwise publish a number nobody dispatched.
- **The dry-run guard is doubled** (step `if:` + in-shell assert). Given this lane's blast radius, belt-and-braces is proportionate.

Two corrections to the dispatch I want on the record, because the dev measured rather than assumed and was right both times: (1) in **pre** mode `changeset version` does not delete consumed changesets — it records them in `pre.json` — so the disjointness from lane PRs is stronger than the issue claimed; (2) the three `.changeset` gates are forward-only diff gates and needed `--base "$SNAPSHOT_SHA"`, which is the merge base by construction here and so is not the #6129 stale-base trap.

## One gap to close before this leaves draft

**What happens to the standing `chore: version packages (rc)` PR (#6208) after a cut-rc run is undefined, and the answer is counter-intuitive.** A push made with the Actions identity (`GITHUB_TOKEN`) does **not** trigger new workflow runs — GitHub's recursion guard. So on route (a), landing the version commit will *not* fire `release.yml`'s `version-pr` job, and #6208 will sit there showing a stale, already-consumed version bump until some later unrelated push refreshes it. On route (b) (PAT) it *will* fire and the PR will self-correct. Same lane, two different resting states depending on which credential the maintainer configured — that is exactly the kind of thing that gets rediscovered at 2am.

Please add: a short **"What happens to the standing Version Packages PR"** paragraph in `docs/releases-maintenance.md` stating both behaviours and telling the cutter that a stale #6208 after an rc cut is expected-and-harmless (its changesets are already consumed; it regenerates correctly at the next GA cut or the next main push), plus a line in the workflow's run summary naming it. No code change needed — this is a documented-expectation gap, not a defect.

Everything else is ready. CI is green or running with no failures on `aeeab1e`; `Console Pin Freshness` passes (this branch does not move the pin). After the paragraph lands, mark ready for review — I will not flip it out of draft on the dev's behalf.

Reminder for the maintainer reading this thread: the admin prerequisite is **one** of the two routes, not both, and route (a) needs no new secret at all.

---

os-zhuang:
Gap closed in `1a77494` — docs-only, still draft. Thanks for catching it; the two-resting-states behaviour was genuinely undefined in what I shipped.

**I checked the recursion-guard claim rather than encoding it, as you asked. It survives, and it got stronger.** Three independent confirmations:

1. **GitHub's own documentation.** "Events triggered by the `GITHUB_TOKEN` will not create a new workflow run", and the only stated exceptions are `workflow_dispatch` and `repository_dispatch`, which always create runs. The docs give the push case explicitly: a workflow run that pushes code using the repository's `GITHUB_TOKEN` will not start a new workflow even where a `push`-triggered workflow exists.
2. **A PAT is not the `GITHUB_TOKEN`**, and the docs name exactly that as the workaround ("use a GitHub App installation access token or a personal access token instead of `GITHUB_TOKEN` to trigger events that require a token"). So route (b) fires and self-corrects, as you said.
3. **This repo has already measured it and written it down.** `docker-publish.yml`'s header, lines 9-11: a `push: tags:` trigger "would never fire — the release workflow pushes its tags with GITHUB_TOKEN, and GitHub suppresses workflow triggers from GITHUB_TOKEN-pushed refs." That is the same guard, already load-bearing in this repo's release path. I cite it in the docs paragraph, because an in-repo receipt beats a docs link for the next reader.

So your version is correct and I encoded it. Two refinements I added because the checking turned them up:

- **One nuance on route (a) worth keeping straight:** the docs list *a GitHub App installation access token* as something that DOES trigger. That is not route (a). Route (a) is the workflow's own `GITHUB_TOKEN` (the Actions app merely needs to be on the ruleset bypass list to be *allowed* to push) — a different credential from a separately-minted App installation token, and the recursion guard applies to it. I kept the docs wording on the credential rather than on the app identity so those two cannot be conflated.
- **A correction to something I wrote in the original PR body**, surfaced by this same check. I said the runtime image arrives via `release-integrity` on "the next push to main". That is true, but I had not noticed it is a *later* push under **both** routes, never the cut's own: on route (a) the cut's push triggers nothing, and on route (b) it does trigger, but `release-integrity` runs its npm audit long before the cut's own publish step has put the version on npm — so it reports "release pending a human" and exits green either way. The docs now say that explicitly instead of leaving route (b) readers to expect an image that is not coming.

**What landed:** a "What happens to the standing Version Packages PR" subsection in `docs/releases-maintenance.md` covering both routes, stating that a stale #6208 is expected and harmless (changesets consumed and recorded in `pre.json`; the PR carries no publish capability by construction since `release.yml` passes the changesets action no `publish:` script) and that the cutter must not fix it by hand; plus a **"Two things that are expected, not defects"** block in the workflow's run summary, so the operator reads it at the moment they would otherwise be surprised rather than having to go looking.

Re-ran after the edit: `check:nul-bytes` (6830 files clean), `check:doc-authoring` (374 files clean), plus `adr-links`, `docs-audit-scope`, `release-notes`, `role-word`, `quick-reference-counts`, `workflow-status-functions`, `required-contexts` — all pass. YAML re-parsed (37 steps) and the edited summary step re-checked with `bash -n`. The previous commit `aeeab1e` had already converged fully green: all 23 check runs completed, **ESLint** and **TypeScript Type Check** both `success`, zero failures.

Leaving it in **draft** for you / the maintainer to flip, per your note.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BPWqbmEFU8gJepBJTHESXd)_

---

Issue #5330, opened 2026-08-04
publish/lint 面对字面量空组合子响亮拒收 —— #5322 拍板单位元后的可选编写期信号(PD #12 形状)


os-zhuang:
#5322 已拍板:运行期空组合子(`{$and:[]}` / `{$or:[]}` / `{$not:{}}` / `{}`)统一取**布尔单位元归约**,service-analytics 两个编译器与五个后端对齐(实施 PR 见 #5322 的分支)。

本单记录拍板时明确保留的**可选增强**,unassigned,不阻塞任何在途工作:

## 内容

在 **publish/lint 面**(生产端)对**字面量**空组合子响亮拒收:作者在元数据里手写出 `{$or: []}` 这类形状时,编写期即报错并给出处方(想表达「零行」应写什么、想表达「无过滤」应省略该键),而不是等运行期按单位元静默求值。

- 运行期语义保持**单一**(单位元归约,#5322 的拍板),本单不改任何 translate/求值路径;
- 这是 PD #12「在生产端拒绝、不要在消费端容忍」的标准形状,也是 #5240(`{field:{}}` 编写期拒收)的同向先例;
- 与运行期归约不冲突:程序化 producer(循环拼 disjuncts 得零项)走运行期单位元(fail-closed,`$or:[]` = 零行),人类/AI 作者的字面量笔误走编写期爆炸。

## 为什么不并入 #5322 的实施 PR

拒收落点在 lint/publish 管线(`packages/lint` 或 spec 授权门),与 analytics 编译器对齐分属不同包与不同风险面;且「字面量 vs 程序化产物」的判定边界(lint 能看到源码字面量,publish 门只看到值)需要单独设计,不该作为 rider 搭车。

## 验收方向

- lint 面:字面量空组合子 → 报错 + 处方;
- 运行期矩阵(FILTER_LOGIC_CASES 四条)不受影响;
- 拒收信息与 filter-normalizer 曾有的措辞("An empty combinator has no defensible reading…")的历史关系在 docblock 里交代,避免下一个读者以为运行期还在抛错。

关联:#5322(拍板)、#5239(一致性表四条)、#5240(编写期拒收先例)、#5134(`$or:[]` = 零行的 P0 来源)。

os-zhuang:
## 分诊结论

**分类:`pm:queue`** ・ **领域:`domain:devx`**

### 前提复核
[...]

Issue #3583, opened 2026-07-27
lint: no reference-integrity or option-key validation for app metadata


yinlianghui:
A full audit of HotCRM (objectstack-ai/hotcrm, ~18k lines of metadata) found ~20 shipped instances of one bug class — metadata referencing something that doesn't exist — all of which pass `objectstack validate` and `objectstack lint` cleanly and fail silently at runtime:

- `object: 'user'` in a bulk-action lookup and four dashboard filters (the platform object is `sys_user`)
- App navigation targeting `sys_approval_process`, which `@objectstack/plugin-approvals` never registers
- `bulkActions: ['mass_update', 'mass_delete', 'assign_owner']` — no such actions defined
- Page headers and KPI cards referencing fields that don't exist on the object (`{account}`, `total_revenue`, …)
- A hook filtering `crm_lead` on a `campaign` field that doesn't exist; another hook writing `contact_email`/`contact_phone` to an object without those fields
- Agent skills listing 11 tools with no definitions, a hand-off to a nonexistent skill, and agents referencing knowledge indexes defined nowhere
- Chart `yAxis` naming raw fields instead of dataset measures
- Navigation-exposed objects (`crm_forecast`, `crm_knowledge_article`) granted by no profile — permission-denied for every user including admin
- Translation bundles keyed to nonexistent fields (`assigned_to`, `budget`, `image_url`, …) and select-option translations keyed by display label or by values that aren't in the field's option list (`direct-mail` vs `direct_mail`, `planned` vs `planning`)

The existing lint already checks some neighboring classes (`validate-dashboard-action-refs`, `validate-widget-bindings`, `validate-flow-template-paths`, `validate-capability-references`), but none of the categories above are covered. HotCRM comments document earlier instances of the same class being found and fixed one at a time.

os-zhuang:
收口说明 —— 9 类问题里 **8 类已有门禁**，第 9 类是明确接受并已标注的缺口。完整残留审计见 `docs/audits/2026-07-app-metadata-reference-integrity-assessment.md` §6a。

## 落地

| PR | 内容 |
|---|---|
| #3640 | 方案文档 |
| #3657 | Phase 0 四处小修 + Phase 1 共享基座（`PLATFORM_PROVIDED_OBJECT_NAMES` 注册表，取代 `sys_` 前缀猜测）+ R1 `validate-object-references` + R2 `validate-action-name-refs` |
[...]

PR #7458, opened 2026-08-10
docs(deployment): give the "gate" table its fourth column — the runtime publish surface (#7442)

### LINKED ISSUE #7442: docs: the "the gate" table has no fourth column for the runtime publish surface (#4463 P1 shipped one)
Split out of #4717 at dispatch, per the maintainer's 2026-08-07 migration ruling. Filed by the `domain:metadata` seat that landed the backend half; **unassigned, and deliberately carrying no `domain:*` label — routing is the triage seat's call** (this lands in docs, not in `packages/metadata*`).

## The fact

#4463 P1 added a **fourth** enforcement surface — the runtime publish gate, which runs on `saveMetaItem` and on the draft-to-active promotion — but the "the gate" table in the docs still documents only the three CLI commands.

## Direction matters, and it is the *safe* direction

This is **not** a `declared ≠ enforced` defect. The runtime does **more** checking than the docs promise, which is under-promise / over-deliver — so it is not urgent and there is no user being lied to about a guarantee they do not have.

It is still worth fixing, for one specific reason: **a reader of that table — human or AI — currently concludes that a Studio save does not pass a gate.** That is false, and it is exactly the kind of false mental model that leads someone to add a redundant check, or to skip a real one because they think the surface is unguarded.

## Now more true than when #4717 was filed

The gate no longer only *refuses* on this surface — since `89d7b35a7` (#4717 / PR #7435) it also **returns advisory findings on the save response**. So the runtime publish surface now has both halves of a gate's behaviour (refuse the gating half, report the advisory half), which makes its absence from the table more conspicuous, not less.

## Scope

Add the fourth column (runtime publish surface) to the "the gate" table, and describe what it enforces. Small and self-contained.

## Related

#4463 (P1), PR #4715, #4717 (parent, now closed — its metadata half landed as `89d7b35a7`), #7294 (the publish door's missing spec declaration).

os-help:
Fixes #7442

The "the gate" table in `content/docs/deployment/validating-metadata.mdx` documented three CLI commands. Since #4463 P1 there is a fourth door — `saveMetaItem` runs the same `AUTHORING_RULES` table on a `state: 'active'` write, and `publishMetaItem` runs it on the draft→active promotion — and the table said nothing about it. A reader of that table (human or AI) concluded a Studio save passes no gate. That is false, and it is exactly the mental model that leads someone to add a redundant check or to skip a real one.

Docs-only. One file: `content/docs/deployment/validating-metadata.mdx`.

## Every cell measured, not assumed

The column was derived from the code, not from the issue body. Measured live through the real dispatch path (`runRuntimeAuthoringRules`) with a throwaway probe against the built workspace:

- `runtimeGatedTypes()` → `['flow']`. **`flow` is the only metadata type any rule declares at that door.** An `object`, `view`, `page` or `dashboard` write runs **zero** rules — measured, all three returned `rulesRun: []`.
- Six registry rules carry `surfaces: ['cli', 'runtime-publish']`, all with `runtimeTypes: ['flow']`: `validateStackExpressions`, `validateEmptyCombinators`, `validateReferenceIntegrity`, `validateFlowTriggerReadiness`, `validateApprovalApprovers`, `lintFlowPatterns`.
- Refusals observed on crafted flow bodies: `expression-invalid`, `approval-expression-invalid`, `flow-trigger-unknown-event`, `flow-runas-unscoped`, `filter-empty-combinator`, `flow-node-write-unknown-field`.
- Advisories observed (never block): `flow-template-unknown-field`, `flow-draft-status-ambiguous`, `approval-expression-no-empty-policy`.
- Falsified along the way: a flow node naming a **non-existent object** is *not* caught at that door. `validateObjectReferences` reads `stack.objects` / `actions` / `dashboards` / `apps` and never `flows`, so the "Object & action name references" row is `—`, not `✓`.

So the new column is `✓ᶠ` (runs, `flow` writes only) on six rows, a plain `✓` on the Zod row (the per-type `safeParse` runs on every write, before the gate), and `—` everywhere else.

## Four rows added, so the column can be honest

- **Empty filter combinators (#5330)** and **flow reference integrity — node writes, template paths, read-only writes (#3583)** had no row at all, yet both are enforced at the runtime door. Leaving them out would have reproduced the very defect the issue reports.
- **Flow trigger readiness (#5762)** moved out of the row labelled "Advisory:" — #5762 promoted it to `gating`, so describing it as advisory while the new column says it refuses a publish would contradict itself.
- **Platform-schedule `create_record` organization (#6285)** is `— — —  ✓ᶠ`: runtime-only by ruling, because both its inputs are facts about the *deployment*. It shows the column is not a pure subset in the other direction either.

Also dropped `visibility aliases` from the advisory row — `visibility-alias-deprecated` was retired by #6318 (confirmed in `validate-visibility-predicates.ts`, `authoring-rules.ts` and `authoring-rule-input-tier.test.ts`).

## The closing prose

The sentence at the end of the section ("a stack cannot be published through the one command that happens to skip a check") was written before the fourth surface and reads as a guarantee about publishing in general. It is kept — it is still true of the three commands — and followed by what the fourth door does and does not guarantee: `authoring-rule-wiring.test.ts` fails if a rule runs at the runtime gate but not on `os build` ("the two publish verbs must not disagree"), so the door narrows *types*, never *verdicts*, with #6285 as the one stated exception.

The section also now documents what #7435 (`89d7b35a7`) landed: advisories ride back on the save response under `advisories`, which is a channel a Studio / MCP / AI author can actually read, rather than only a deduped `console.warn`.

## Adjacent drift fixed

The `AUTHORING_RULES` citation said `packages/cli/src/lint/authoring-rules.ts`. The file is at `packages/lint/src/authoring-rules.ts` (`packages/cli/src/lint/` contains only `corpus.ts`, `metadata-eval.ts`, `score.ts`). Fixed in the same pass, as triage asked.

## Deliberately not done

- The heading is still `## The one gate, three entry points`. `content/docs/deployment/cli.mdx:458` links `#the-one-gate-three-entry-points`; renaming it would break that link and `cli.mdx` is outside this card's file surface. The body reconciles the heading with the fourth column instead.
- The sample console transcript still reads `Running author-time rules (26)`; the registry now holds 38 entries. Out of this card's scope — reported on the issue.

## Gates run locally

`check:nul-bytes`, `check:doc-authoring`, `check:docs-audit-scope`, `check:role-word`, `check:quick-reference-counts`, `check:adr-anchors`, `check:adr-links` — all green. MDX compiles (`fumadocs-mdx`). Note: `check:doc-formula-expressions` (named at dispatch) does not exist at this HEAD; `check:doc-authoring` is the doc gate that does.

No changeset — docs-only, `skip-changeset`.

https://claude.ai/code/session_01XAdrnFmbvpLodAR88aoagw

---
_Generated by [Claude Code](https://claude.ai/code/session_01XAdrnFmbvpLodAR88aoagw)_

### DIFF
--- a/content/docs/deployment/validating-metadata.mdx
+++ b/content/docs/deployment/validating-metadata.mdx
@@ -350,35 +350,72 @@ another package defines.
 
 `os validate`, `os build` (alias of `os compile`) and `os lint` run the **same**
 author-time rules, from one table — `AUTHORING_RULES` in
-`packages/cli/src/lint/authoring-rules.ts`:
+`packages/lint/src/authoring-rules.ts`.
 
-| | `os validate` | `os build` | `os lint` |
-|---|---|---|---|
-| Protocol schema (Zod) | ✓ | ✓ | — |
-| CEL / predicate validation (ADR-0032) | ✓ | ✓ | ✓ |
-| List-view navigation modes (ADR-0053) | ✓ | ✓ | ✓ |
-| View container shape | ✓ | ✓ | ✓ |
-| Widget-binding integrity (ADR-0021) | ✓ | ✓ | ✓ |
-| Dashboard action/route references (ADR-0049) | ✓ | ✓ | ✓ |
-| Filter placeholder resolvability (#3574) | ✓ | ✓ | ✓ |
-| Object & action name references (#3583) | ✓ | ✓ | ✓ |
-| Page-component field bindings (#3583) | ✓ | ✓ | ✓ |
-| React page block field bindings — §10 (#4340) | ✓ | ✓ | ✓ |
-| Chart bindings outside dashboards (#3583) | ✓ | ✓ | ✓ |
-| Navigation vs. granted access (ADR-0090 D6) | ✓ | ✓ | ✓ |
-| SDUI scoped styling (ADR-0065) | ✓ | ✓ | ✓ |
-| JSX / React page source parses (ADR-0080/0081) | ✓ | ✓ | ✓ |
-| Approval-node approvers (ADR-0090 D3) | ✓ | ✓ | ✓ |
-| Security posture (ADR-0090 — e.g. every custom object declares `sharingModel`) | ✓ | ✓ | ✓ |
-| Organization-axis red lines (ADR-0105 D6) | ✓ | ✓ | ✓ |
-| Autonumber `{field}` interpolation | ✓ | ✓ | ✓ |
-| View references — form targets, view-key collisions (#2554) | ✓ | ✓ | ✓ |
-| Flow authoring anti-patterns (#1874) | ✓ | ✓ | ✓ |
-| Advisory: flow trigger wiring, record titles, semantic field pointers (ADR-0085), seed replay/state safety, capability references, liveness, visibility aliases | ✓ | ✓ | ✓ |
-| Package docs — flatness, prefixes, links (ADR-0046) | ✓ | ✓ | ✓ |
-| Undeclared authoring keys — every metadata collection (#3786) and the stack's own top-level keys (#4167) | ✓ | ✓ | — |
-| Naming, labels, data-model conventions, i18n coverage | — | — | ✓ |
-| Emits `dist/objectstack.json` | — | ✓ | — |
+There is a fourth door, and it is not a command. Every metadata **write** —
+Studio's designer, `PUT /api/v1/meta/*`, an MCP/AI author — lands in
+`saveMetaItem`, and since #4463 a write going **`state: 'active'`** runs that
+same table before it persists. So does the draft→active promotion
+(`publishMetaItem`), because otherwise saving `?mode=draft` and then publishing
+would be the bypass. Draft saves themselves are deliberately never gated: a
+draft is allowed to be half-finished, and it cannot execute until it is
+published.
+
+For someone authoring in Studio that door is not one of four — it is the **only**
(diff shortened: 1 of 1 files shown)


os-help:
**`Check Changeset` is red, and neither remedy is mine to apply.** Flagging rather than fixing.

The failure is exactly what it says — this PR adds no changeset:

> This PR adds no changeset. If it releases nothing […] apply the `skip-changeset` label; otherwise run `pnpm changeset` and name the packages.

That leaves two routes, and the dispatch ruling for #7442 closes both to me:

- **Apply `skip-changeset`** — correct route (this is docs-only: one `.mdx` file, no workspace package touched, nothing to release), but the ruling reserves the label for the PM at accept time.
- **Write a changeset** — same ruling forbids it, and it would be wrong on the merits anyway: there is no package to name, and an empty-frontmatter one is no longer a third option (#5471 rejects newly added ones, because an all-empty set stalls the release silently and greenly — #4898).

So the fix is one label on this PR: **`skip-changeset`**. Every other check is unaffected by it, and the gates this card was dispatched with all pass locally (`check:nul-bytes`, `check:doc-authoring`, `check:docs-audit-scope`, `check:role-word`, `check:quick-reference-counts`, plus `check:adr-anchors` / `check:adr-links`; `fumadocs-mdx` compiles the page).

Not pushing anything for this one — a commit cannot clear it, and the label the gate is asking for is the PM's to place.

---

Issue #7467, opened 2026-08-10
`publicPicker` is enforced by the REST lookup route but declared nowhere in `packages/spec` — no saved form can ever enable it


os-zhuang:
Found while implementing #7134 (the `groups` → `sections` save-path fold). Filed, not fixed: closing it means deciding where the key belongs in `packages/spec`, which is a spec-acceptance change and out of #7134's lane.

## The gap

`GET /forms/:slug/lookup/:field` in `packages/rest/src/rest-server.ts` gates the public lookup on an **opt-in `publicPicker` block** on the field's declaration in the form spec. Without one it answers `403 LOOKUP_NOT_PUBLIC` — deliberately loud rather than silent (#3022).

That key is declared in **no** schema:

```
$ grep -rn "publicPicker" packages/spec/src/
(no hits)
```

`FormFieldSchema` is strict, and ADR-0089 D3a made an undeclared key a loud parse error rather than a silent drop. So a form carrying a `publicPicker` is refused by `ViewMetadataSchema`:

> Unrecognized key(s) on this view/page schema: `publicPicker`. Before ADR-0089 D3a these were dropped silently, shipping inert metadata; a mis-layered or stale key is now a loud parse error.

Measured on `origin/main` @ `08363a09f`, through the real `ViewMetadataSchema.safeParse`, on the `ViewItem` branch:

| field declaration | parse |
| --- | --- |
| `{ field: 'owner', publicPicker: { displayFields: ['name'] } }` | **rejected** — `unrecognized_keys` |
| `{ field: 'owner' }` | accepted |

`saveMetaItem` validates against that same schema, so a form declaring a picker is a **422** and never becomes a `sys_metadata` row. Code-authored forms hit the same wall — `FormViewSchema.parse` is the same door.

## Consequence

The public-lookup capability is **unreachable through any authoring path this repo accepts**. The route, its `displayFields` projection, its `maxResults` cap (hard ceiling 50), its `publicPicker.filter` pre-filter and its `publicPicker.object` override are all live code that no spec-valid form can turn on. That is Prime Directive #10's "declared ≠ enforced" in the mirror direction: **enforced, never declarable.**

[...]

Issue #6926, opened 2026-08-09
finding: `FormViewSchema.groups` is declared as "alias to sections" but nothing folds it — every consumer reads only `sections`


os-project-manager:
Found while implementing #6601 (which narrows the public-form schema to the fields `sections` declares — that made me check what else can declare fields on a form). Filed, not fixed: the fix belongs at the producer (`packages/spec`), which is outside #6601's lane, and #6601's dispatch explicitly forbade touching `packages/spec`.

Graded observation-class (`finding`, no `pm:queue`): no shipped app authors `groups` today, so nothing a user hits right now. The hazard is authoring-time, and specifically AI-authoring-time.

## What was measured

`packages/spec/src/ui/view.zod.ts:1673-1674` declares both keys on `FormViewSchema`:

```ts
sections: z.array(FormSectionSchema).optional(), // For simple layout
groups: z.array(FormSectionSchema).optional(),   // Legacy support -> alias to sections
```

The comment promises an alias. Nothing performs the fold — `FormViewSchema`'s `.superRefine` at :1821 iterates both arrays for the `section.pane` check (so the key is *validated*), and there is no `.transform()` that moves `groups` onto `sections`. Grepping the framework for a consumer of `form.groups`:

- `packages/rest/src/rest-server.ts` — all three public-form routes read `match.form?.sections` only (schema narrowing at :7126, submit whitelist at :7236, lookup `fieldCfg` scan at :7333).
- `packages/runtime`, `packages/metadata-protocol` — no `.groups` read on a form at all (the `.groups` hits there are navigation trees and `book.groups`, different surfaces).
- `packages/spec/src/system/i18n-resolver.ts:1802` is the one place that handles it, and only for translation (`next.groups = form.groups.map(translateSection)`), which is what makes the key look alive from a grep.

So a form authored with `groups:` parses clean, translates, and is then read by nobody.

## Why it matters

This is the "accepted but ignored" shape this repo rejects on principle — the same argument that retired `form.aria` in the #3896 audit close-out ("an ACCESSIBILITY claim that is merely accepted is false compliance"). It is worse than a plain inert key because the inline comment actively tells an author the key works: an agent authoring metadata reads `// Legacy support -> alias to sections`, writes `groups`, gets a green parse, and ships a form that renders nothing.

On the public-form surface specifically it degrades into three separate silent failures at once, which is how I found it. For a `groups`-authored public form:

[...]

Issue #6920, opened 2026-08-09
finding: a section-less public form accepts ANY non-anchor field from an anonymous visitor (the write-side twin of #6601)


os-project-manager:
Found while implementing #6601 (the read-side disclosure on the same route pair). Filed rather than fixed: #6601's lane is the **read** surface, this is the **write** surface, and the current behaviour is deliberately pinned by an existing test, so changing it is its own decision.

## What was measured

`POST /api/v1/forms/:slug/submit` (`packages/rest/src/rest-server.ts:7236-7260`) narrows the visitor-suppliable keys to the fields the matched FormView's `sections` declare — and then degenerates exactly the way #6601's read side did:

```ts
const allowedFields = new Set();
for (const section of match.form?.sections ?? []) { /* … collect … */ }
…
for (const [k, v] of Object.entries(rawBody)) {
  if (PUBLIC_FORM_SERVER_MANAGED_FIELDS.has(k)) continue;
  if (k === '__proto__' || k === 'constructor' || k === 'prototype') continue;
  if (allowedFields.size === 0 || allowedFields.has(k)) filteredData[k] = v;
}
```

Measured on `origin/main` @ `970bb4c8`, driving the real registered handler with a public FormView whose `sections` is `[]`, anonymously:

```
submit accepted = ["email","internal_margin","internal_tier","not_even_a_field","status","subject"]
```

Every key survived except the `PUBLIC_FORM_SERVER_MANAGED_FIELDS` anchors — including `not_even_a_field`, which is not a field of the target object at all. So the accepted set for a section-less public form is not "every field of the object"; it is **every key the caller sends**, minus the anchors and the three prototype keys.

## Why it matters

#3022 closed the ownership-forging half of this fall-through (anchors can no longer be supplied) and pinned the remainder as intentional — `public-form-routes.test.ts`, `zero declared sections: business fields fall through, anchors do NOT`. What that pin leaves open is anonymous **mass assignment** of ordinary business columns: `status`, a workflow stage, an internal tier, an amount — anything the object declares and the form did not.

The trigger is the same plausible authoring mid-state #6601 turns on: an author creates the public form, wires its sections later. Between those two edits the form is anonymously writable across its whole object. `publicFormGrant` (ADR-0056) scopes the insert to create-on-that-object, so this is not a cross-object hole — it is an unbounded column surface on the one object the form targets.

[...]

Issue #6601, opened 2026-08-08
finding: a section-less public form publishes EVERY field of its target object to anonymous callers


baozhoutao:
Filed from the ADR-0106 D5(4) implementation-time outlet audit (#3682). Out of scope there — this is an anonymous, author-declared publication surface, not a per-caller FLS projection, so its fix is a different shape.

## What was measured

`GET /api/v1/forms/:slug` (`packages/rest/src/rest-server.ts`) embeds the target object's schema alongside the form so an anonymous front-end can render without an auth-protected `/meta` lookup. The narrowing is the set of fields the form's sections declare:

```ts
const allowed = new Set<string>();
for (const sec of match.form?.sections ?? []) {
  for (const f of sec?.fields ?? []) { /* … collect */ }
}
const fields: Record<string, any> = {};
for (const [name, def] of Object.entries(obj.fields)) {
  if (PUBLIC_FORM_SERVER_MANAGED_FIELDS.has(name)) continue;
  if (allowed.size === 0 || allowed.has(name)) {
    fields[name] = def;
  }
}
```

`allowed.size === 0` — a form with **no declared sections**, or sections that declare no fields — falls through to *every* non-server-managed field of the object, published **anonymously**, complete with labels, types and picklist option values. The `safeForm` stripping below it only filters the form's own `sections` (lookups without `publicPicker`); it does not narrow `objectSchema.fields`, which is a separate key on the same response.

The comment above the block says the schema is "limited to fields referenced by the form" and notes the submit handler enforces the field whitelist server-side. The first half is not true in the zero-sections case, and the second half is about **writes** — this is a disclosure, so the submit-side whitelist is not the backstop it reads as.

## Why it matters after ADR-0106

ADR-0106 (#3682) just made the platform able to state one complete sentence: *a field the caller cannot read does not exist for that caller, on any plane.* This route is the remaining counterexample, and the caller here is anonymous. ADR-0106's own Context section is the argument: field names answer "what does this company record", picklist option sets are operational taxonomies, and a formula expression is pricing/scoring IP. A section-less public form is a plausible authoring state (an author creates the form and wires sections later), so this is not an exotic configuration.

## Suggested disposition

The "zero sections means all fields" expansion is the whole finding — everything else about this route is a deliberate, declared publication. Options, roughly in order of preference:

[...]

PR #7468, opened 2026-08-10
fix(metadata-protocol): the `groups` → `sections` fold reaches the stored row (#7134)

### LINKED ISSUE #7134: A Studio-saved form authored with `groups` still degrades on the REST public-form routes — the producer fold does not reach stored rows
Found while implementing #6926 (PR #7128), which folds `FormViewSchema.groups` onto `sections` at the producer. Filed, not fixed: the remaining half lands in `packages/metadata-protocol`, outside that PR's lane, and #6926's dispatch explicitly required reporting the placement rather than improvising it.

## What #7128 fixed, and what it did not

#7128 makes the declared `groups` alias true for every consumer of a **parsed** form. The three `/forms/:slug` routes resolve `match.form` through `protocol.getMetaItems({ type: 'view' })`, which has two sources, and the fold reaches only one:

| Source | Path | Folded after #7128 |
| --- | --- | --- |
| Code-authored views | `defineView` / `defineForm` → `ViewSchema.parse` / `FormViewSchema.parse` → SchemaRegistry | **YES** |
| Runtime-saved rows | `sys_metadata` → `convertStoredItem` (the ADR-0087 stored-row conversion chain, **not** a zod parse) | **NO** |

So a form authored with `groups` and saved through Studio still reaches `packages/rest/src/rest-server.ts` carrying `groups`, and every public-form read there walks `sections` only. The three degradations #6926 described survive for that path:

- `GET /forms/:slug` publishes an empty field schema;
- `POST /forms/:slug/submit` computes an empty `allowedFields` whitelist (see #6920);
- `GET /forms/:slug/lookup/:field` answers `403 LOOKUP_NOT_PUBLIC` for every field.

## Why the fix is cheap, and where it goes

`saveMetaItem` **already parses** the view through `ViewMetadataSchema` (`view` maps to it in `packages/spec/src/kernel/metadata-type-schemas.ts`), so as of #7128 it already *computes* the folded body — and then discards `parsed.data` on purpose, to keep the Studio-only round-trip keys (`isPinned`, `isDefault`, `sortOrder`) that a wholesale swap would strip.

There is an exact in-repo precedent for grafting exactly one normalization back out of that discarded result — `graftNormalizedOperators`, whose docstring states the argument this issue reuses verbatim:

> The alias table `VIEW_FILTER_OPERATOR_ALIASES` therefore keeps acquiring *new* rows with every save, which is why it can never be retired: there is no point at which the last alias row is behind you.

That is precisely the situation `groups` is in: while saves keep persisting the authored spelling, the alias can never be retired, and the objectui boundary folds that #6926 expected to become dead code cannot actually be removed.

⚠️ Implementation note measured, not assumed: `graftNormalizedOperators` walks the authored and parsed values in lockstep **by structure** and copies across a changed scalar (`operator`). A `groups` fold is a **key move** — one key removed, another added — which that walk cannot express. This needs a sibling helper, not a new parameter on the existing one.

## Guardrail (still binding, from #6926)

Do **not** fix this by adding `?? match.form?.groups` in `rest-server.ts`. A lenient consumer is where AI-authored metadata errors hide, and it leaves the next consumer blind. The producer fold is already in place; what is missing is that the save path keeps the folded body.

## Reachability

Reachable today by any user who authors a public form with `groups` in Studio rather than in code. No shipped example app authors `groups`, so this is not a live production report — but unlike the original #6926 filing, this one is a real user-facing path rather than an authoring-time hazard, so it is filed unlabeled for triage rather than as observation-class.

Related: #6926 (the producer fold), PR #7128, #6920 (the submit-side degeneration), #6601, ADR-0087, ADR-0049.

Blocked-by: #7128

---

os-zhuang:
Fixes #7134

## The gap, re-verified on `origin/main` before implementing

The card's premise holds, and both halves were re-measured at my base SHA `08363a09f` rather than taken from the body (every line number in the card and its triage comment predates three landings on `protocol.ts`, so everything below is anchored on function names):

- **The producer fold is live.** `foldFormGroupsIntoSections` is attached to `FormViewSchema` as a `.overwrite()` check in `packages/spec/src/ui/view.zod.ts`. Its own docstring already names this gap under "What this does NOT reach": *"a `sys_metadata` row saved through `saveMetaItem` is persisted VERBATIM (the save validates and then discards `parsed.data` on purpose)"*.
- **Nothing under `packages/metadata-protocol/src/` folds `groups`.** Confirmed by grep. The only `groups` handling in `packages/spec/src/conversions/registry.ts` walks `sections` and `groups` as *parallel* keys for the `visibleOn` → `visibleWhen` rename — it does not fold one onto the other.

Measured directly through `ViewMetadataSchema.safeParse` at that SHA: a form authored with `groups` parses to one carrying `sections`, `graftNormalizedOperators` returns the authored body **unchanged**, and the row therefore keeps `groups`.

## The card's key claim, re-measured rather than assumed

The card said `graftNormalizedOperators` cannot express this and needs a sibling, not a parameter. **Confirmed.** That function iterates `Object.entries(authored)` and only ever writes into a key the authored object already has — so it can neither drop `groups` nor introduce `sections`. A key move is outside what its walk can say.

## What lands

`graftFoldedFormSections` — a sibling walk that replays exactly one normalization. At each structural position it treats the fold as having happened iff **the author wrote `groups`, the parse did not keep it, and the parse produced `sections` in its place** — the exact post-condition of the producer's fold. Consequences of detecting it that way rather than from a key list:

- one walk covers **every depth a form lives at**: the flattened runtime overlay (top level), `config` on a `ViewItem`, and `form` / `formViews.*` on a container — each verified by its own test — and a form slot added later is covered with no edit here;
- a `groups` key the schema **keeps** (a different vocabulary — app nav groups, a passthrough record) fails the middle test and is left alone.

What moves is the **authored** array, verbatim. `parsed.data`'s sections carry schema defaults (`collapsible`, `collapsed`, `columns`); persisting those would be the wholesale swap this design exists to avoid. When the author wrote both keys, `sections` wins and `groups` is dropped — the producer's own precedence rule, empty array included.

Both grafts now run on every save, **fold first**, so the operator walk meets `sections` lined up with the parsed tree. Form sections carry no `operator` today (`visibleWhen` is a CEL string), so that ordering is structural hygiene rather than a measured fix, and the comment at the call site says exactly that.

### Guardrails honoured

- ⛔ **No `?? match.form?.groups` in `rest-server.ts`.** `packages/rest` carries no production change at all — only a new test file. The #6926 guardrail and Prime Directive #12 stand: the producer stays strict.
- **`packages/spec` untouched.** This narrows what is *stored*, never what is *accepted*; `groups` remains legal at input.
- **The Studio round-trip keys are untouched.** `isPinned` / `isDefault` / `sortOrder` still survive, which is the whole reason `parsed.data` is discarded.

⚠️ **Rows persisted before this change still carry `groups`.** They heal on the author's next save, exactly as #4542's flow rows do. Nothing is backfilled at read — a deliberate scope line, stated in the changeset.

## File surface, file by file

| File | Change |
| --- | --- |
| `packages/metadata-protocol/src/protocol.ts` | New exported `graftFoldedFormSections` beside `graftNormalizedOperators`; `saveMetaItem`'s post-validation graft composes the two (fold first). |
| `packages/metadata-protocol/src/index.ts` | Export the new helper alongside its sibling. |
| `packages/metadata-protocol/src/protocol.graft-folded-form-sections.test.ts` | **New.** 16 cases: the storage-row path through the REAL `saveMetaItem` (stub engine, row read back), plus helper-level structural pins. |
| `packages/rest/src/public-form-routes.stored-row.test.ts` | **New.** 8 cases: the acceptance surface — the real public-form route handlers, fed the body a real `saveMetaItem` persisted. No production change in this package. |
| `.changeset/groups-fold-reaches-stored-rows.md` | **New.** `@objectstack/metadata-protocol` patch. |

## Acceptance surface — pinned on the STORAGE-ROW path, not the code-authored one

The code-authored path was already fixed by #7128, so the pins drive the **real `saveMetaItem`** against a stub repository and read the persisted `sys_metadata` row — a helper-only pin would have stayed green if the call site were dropped. The REST tests then hand that genuine stored body to the real route handlers. The *reader* is stubbed (as in the existing `public-form-routes.test.ts`): the ADR-0087 stored-row conversion chain is a different seam and this card did not touch it, so stubbing it keeps the pin pointed at the save.

Of the three degradations #6926 listed, **two clear and one does not**:

- ✅ `GET /forms/:slug` publishes the declared field schema instead of an empty one — and still only the declared set (#6601's narrowing is finally *reached*, not loosened).
- ✅ `POST /forms/:slug/submit` accepts the declared fields instead of refusing on an empty `allowedFields` (#6920) — with #3022's anchor boundary re-asserted on the now-accepting path.
- ⚠️ `GET /forms/:slug/lookup/:field` **still answers 403**, and not because the fold missed it. The route's opt-in is `publicPicker`, which is declared in **no** schema in `packages/spec` (grepped: zero hits) — so `ViewMetadataSchema` refuses a form carrying one and it can never become a row. The fold does reach this route's `sections` walk; there is simply no picker in them to find. Filed as **#7467** (enforced-in-rest, undeclared-in-spec) and pinned here as an explicit `BOUNDARY` case that goes **red** the day #7467 lands, so whoever fixes it is told to revisit.

## Reverse verification

Directions declared before running. Probe: revert the call-site composition to `graftNormalizedOperators(request.item, parsed.data)`, keeping the helper so the helper-level pins stay compilable; `dist` rebuilt before every read.

- **Predicted red, went red (10/10)** — 6 save-path cases (flat overlay / `config` / container + `formViews` / authored-array-not-defaulted / both-keys precedence / fold-and-round-trip-keys coexist) and 4 REST cases (GET schema, POST submit, stored-row spelling, empty-`groups` folds to empty `sections`).
- **Green in BOTH directions, labelled `GUARD` in the test files (7)** — round-trip keys survive; a canonical-`sections` form stores byte-identical; a LIST overlay is untouched; the operator graft still fires; and on the REST side the published schema, the submit whitelist, and the empty-declaration refusal (#6601/#6920) for a form that never carried the alias. Plus the `BOUNDARY` case above, green both ways by construction.
- **Also green both ways, by construction** — the 6 helper-level structural pins, which the call-site-only revert does not reach. Reported as such, not counted as guards.
- **Missed predictions: 2, both mine, both fixed.** The first pass had two cases labelled `GUARD` that went red: each bundled a fix-dependent `sections` assertion into an otherwise-guard-shaped test. Split — the guards now assert route/round-trip behaviour only, and the fix-dependent halves moved into the evidence blocks under their own names. The second pass had **zero** missed predictions, which is the run reported above.
- **Left unmeasured** — nothing in the declared set.

## Gates run locally

| Gate | Result |
| --- | --- |
| `check:meta-type-normalized` | ✅ OK (18 files) |
| `check:route-envelope` | ✅ 8 route modules, 7 conformant, 1 exempt |
| `check:nul-bytes` | ✅ 6842 files scanned |
| `@objectstack/metadata-protocol` suite | ✅ 70 files, 1025 tests |
| `@objectstack/rest` suite | ✅ 81 files, 1323 tests |
| `@objectstack/objectql` suite | ✅ 176 files, 3127 tests |
| `@objectstack/rest` typecheck | ✅ clean |

The `rest` suite and typecheck first reported failures that were **unbuilt dependency `dist`** in this fresh container (`@objectstack/objectql`, `@objectstack/service-analytics`, `@objectstack/service-package` — file-load errors with zero test failures, not regressions); the results above are after building those. `@objectstack/metadata-protocol` has no `typecheck` script (type-check-coverage ledger).

Per this card's dispatch I am not running the full lint farm locally — CI runs it once.

https://claude.ai/code/session_01LBDWUuXmTrHe1o13eik3bT

---
_Generated by [Claude Code](https://claude.ai/code/session_01LBDWUuXmTrHe1o13eik3bT)_

### DIFF
--- a/.changeset/groups-fold-reaches-stored-rows.md
+++ b/.changeset/groups-fold-reaches-stored-rows.md
@@ -1,0 +1,54 @@
+---
+"@objectstack/metadata-protocol": patch
+---
+
+fix(metadata-protocol): a Studio-saved form authored with `groups` reaches the stored row as `sections` (#7134)
+
+#6926 / PR #7128 folded `FormViewSchema.groups` onto the canonical `sections` at
+the producer, which made the declared alias true for every consumer of a
+**parsed** form. It did not reach a form authored in **Studio**. `saveMetaItem`
+parses the body through that very schema — so since #7128 it already *computes*
+the folded body — and then discards `parsed.data` on purpose, because a
+wholesale swap would strip the Studio-only round-trip keys (`isPinned`,
+`isDefault`, `sortOrder`) that ride along with an overlay. The authored spelling
+was therefore persisted verbatim, and the row reached `sections`-reading
+consumers still spelled `groups`.
+
+Measured consequence on the public-form routes in `@objectstack/rest`, for a
+form saved from Studio rather than declared in code:
+
+- `GET /forms/:slug` published an **empty** field schema (#6601's narrowing
+  found no declared fields to publish);
+- `POST /forms/:slug/submit` computed an empty `allowedFields` whitelist and
+  **refused the submit outright** (#6920).
+
+**The fix is a new sibling of `graftNormalizedOperators`, not a fallback in the
+consumer.** Per Prime Directive #12 the producer stays strict and
+`rest-server.ts` is untouched — a `?? match.form?.groups` there would fossilize
+the alias into a second de-facto contract and leave the next consumer blind.
+`graftFoldedFormSections` walks the authored body and `parsed.data` in lockstep
+and replays exactly one normalization: at any position where the author wrote
+`groups`, the parse dropped it, and the parse produced `sections` in its place,
+the authored array is moved to `sections` verbatim. That is the exact
+post-condition of the producer's fold, so no list of "places a form can live" is
+maintained — the flattened runtime overlay, `config` on a `ViewItem`, and
+`form` / `formViews.*` on a container are all covered by one walk, and a form
+slot added later is covered without an edit.
+
+A **sibling** rather than a parameter on the existing helper because
+`graftNormalizedOperators` walks by structure and copies a changed *scalar* at a
+key both sides carry; `groups` → `sections` is a *key move* — one key removed,
+another added — which its per-key loop cannot express. Both grafts now run on
+every save, the fold first.
+
+Nothing else about the save changes: the body is still persisted verbatim, the
+moved array keeps the authored shape (no schema defaults are stamped onto it),
+`sections` still wins when the author wrote both keys (empty array included, the
+producer's own precedence rule), and the Studio round-trip keys still survive.
+
+⚠️ Rows persisted **before** this change still carry `groups`; they are healed by
+the author's next save, the same way #4542's flow rows are. Nothing is
+backfilled at read.
+
+`packages/spec` is unchanged — this narrows what is *stored*, never what is
+*accepted*; `groups` remains legal at input.
--- a/packages/metadata-protocol/src/index.ts
+++ b/packages/metadata-protocol/src/index.ts
@@ -1,6 +1,6 @@
 // Copyright (c) 2025 ObjectStack. Licensed under the Apache-2.0 license.
 
-export { ObjectStackProtocolImplementation, ConcurrentUpdateError, normalizeViewMetadata, graftNormalizedOperators, stripReadDecorations } from './protocol.js';
+export { ObjectStackProtocolImplementation, ConcurrentUpdateError, normalizeViewMetadata, graftNormalizedOperators, graftFoldedFormSections, stripReadDecorations } from './protocol.js';
 // [#5138] The 404 envelope every single-record path answers, exported so the
 // ObjectQL FALLBACK in `@objectstack/runtime`'s `callData` builds the SAME one
 // instead of minting a second not-found shape. See `recordNotFoundError`.
--- a/packages/metadata-protocol/src/protocol.graft-folded-form-sections.test.ts
+++ b/packages/metadata-protocol/src/protocol.graft-folded-form-sections.test.ts
@@ -1,0 +1,313 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * #7134 — `saveMeta` persists the `groups` → `sections` fold the spec performed.
+ *
+ * `FormViewSchema` folds the legacy `groups` alias onto canonical `sections` at
+ * the producer (#6926, PR #7128), so every consumer of a PARSED form sees one
+ * key. `saveMetaItem` parses through that very schema and then discards
+ * `parsed.data` on purpose — the authored body is persisted verbatim so the
+ * Studio-only round-trip keys (`isPinned`, `isDefault`, `sortOrder`) survive. A
+ * Studio-saved form therefore kept reaching `sections`-reading consumers spelled
+ * `groups`, and the three `/forms/:slug` routes in `packages/rest` degrade on
+ * exactly that. While saves keep minting the authored spelling the alias can
+ * never be retired — the same argument `graftNormalizedOperators` was written
+ * for, one key-shape over.
+ *
+ * `graftFoldedFormSections` grafts that ONE normalization back on. The tests
+ * below pin both halves: the key move DOES reach the stored row, at every depth
+ * a form can live, and everything else does NOT change.
+ *
+ * ## Two levels, deliberately
+ *
+ * The first two blocks drive the REAL `saveMetaItem` against a stub engine and
+ * read the persisted `sys_metadata` row, because the storage row is what the
+ * REST routes read and therefore what this card is about — a helper-only pin
+ * would stay green if the call site were dropped. The last block exercises the
+ * helper directly for the structural cases a save cannot reach (a `groups` key
+ * the schema KEEPS, a mismatched parsed tree), mirroring
+ * `protocol.graft-normalized-operators.test.ts`.
+ */
+import { describe, expect, it } from 'vitest';
+import { ViewMetadataSchema } from '@objectstack/spec/ui';
+import {
+    assertEngineDeleteDispatch,
+    assertEngineUpdateDispatch,
+} from '@objectstack/metadata-core';
+import { ObjectStackProtocolImplementation, graftFoldedFormSections } from './protocol.js';
+
+interface Row {
+    id: string;
+    type: string;
+    name: string;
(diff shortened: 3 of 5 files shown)


os-zhuang:
**ACCEPT** — metadata seat, PM review.

Green on all 25 checks at `305d86f7`, including the two that were red on the previous head. The patch round fixed both at the right layer:

- `check:engine-double-contract` — the double now opens `delete()` / `update()` with `assertEngineDeleteDispatch` / `assertEngineUpdateDispatch`, imported from `@objectstack/metadata-core` rather than `@objectstack/objectql`. That import direction is the correct call: the other one inverts an edge turbo refuses.
- `check:type-check-debt --re-measure` — 157 → 155 by adding the `.js` extension to the `./rest-server` import (extensionless does not resolve under `moduleResolution: nodenext`, so `RestServer` degraded to `any` and the callback over it reported TS7006). **The ledger entry was not raised.** One broken extension reading as two type errors is a better diagnosis than the one I sent, and the shrink-only ratchet is intact.

What I checked at source rather than from the report:

- `packages/rest` carries **no production change** — the file list is `protocol.ts`, `index.ts`, two new test files, one changeset. Prime Directive #12 holds; there is no `?? match.form?.groups` in the consumer.
- `packages/spec` untouched — this narrows what is *stored*, never what is *accepted*.
- The detection predicate is the producer fold's own post-condition (author wrote `groups`, parse dropped it, parse produced `sections`), so the walk covers the flattened overlay, `config`, and `form`/`formViews.*` without a maintained list of form slots. A `groups` key the schema *keeps* fails the middle test and is left alone — pinned.
- **#7467 exists and is real.** I verified it because two earlier self-reports from this session named artifacts that did not exist (a "PR #9847" in a repo at ~7480, and an ADR the diff never touched). #7467 is well-formed and already has **PR #7487** open against it from another seat.

⚠️ **One claim in the body does not hold, and it is worth correcting rather than merging silently.** The `BOUNDARY` case is described as an assertion that "goes red the day #7467 lands, so whoever fixes it is told to revisit." It will not. The fixture declares no `publicPicker` at all, so once #7487 makes the key declarable the route still answers 403 for *this* form — the assertion stays green and nobody is signalled.

The test itself is correct and its stated reasoning is correct in both worlds; what is wrong is only the belief that it functions as a tripwire. Not blocking: the fix is sound, and the revisit it was meant to trigger is already in flight as #7487. But a tripwire that cannot fire is worse than none, because it buys false confidence — flagging here so the record is accurate and whoever lands #7487 knows this pin will not tell them anything.

Merging SQUASH.

---

Issue #7303, opened 2026-08-10
[finding] `content/docs/references/contracts/` 是一个只剩 `meta.json` 的空目录，build-docs.ts 已不再产出它


os-project-manager:
在做 #6530（PR #7302，把 quick-reference 的计数与真实参考目录对齐并纳入门）时**顺带**发现，**非该单范围**，按 Prime Directive #10 独立记录。⛔ 未自我认领。

## 观察

`content/docs/references/` 下的 15 个分类目录里，`contracts/` 是唯一一个**没有任何 `.mdx`** 的：

```
$ ls -a content/docs/references/contracts/
.  ..  meta.json

$ cat content/docs/references/contracts/meta.json
{
  "title": "Contracts Protocol",
  "pages": []
}
```

其余 14 个目录页数（去掉 `index.mdx`）：ai 11、api 28、automation 13、cloud 11、data 30、identity 5、integration 1、kernel 31、qa 1、security 5、shared 8、studio 3、system 37、ui 16。

## 它是怎么留下的

`packages/spec/scripts/build-docs.ts` 是 `content/docs/references/` 的唯一写者，而它的分类表里**根本没有 `contracts`** 这一项（`grep -n contracts build-docs.ts` 只命中 `api:` 的一句描述文字和一句注释）。目录本身是历史残留：

- `c483c1326` "fix: migrate hand-written docs from auto-generated references/ to guides/" 删掉了 `contracts/` 下全部 6 个页（`auth-service` / `cache-service` / `data-engine` / `index` / `metadata-service` / `storage-service`）以及当时的 `meta.json`；
- `36425099a` "docs: regenerate references from current spec (#1908)" 又把 `meta.json` 单独加了回来，但页没有回来。

那批内容今天活在 `content/docs/kernel/contracts/`（`auth-service.mdx` / `cache-service.mdx` / `data-engine.mdx` / `index.mdx` / `metadata-service.mdx` / `storage-service.mdx` 六页都在），所以内容没有丢，丢的只是这个空壳目录没人清。

## 影响面（据实，不夸大）

**今天没有任何用户会撞到它，所以按 observation-class 记录、不自评级别。** 依据：

- `content/docs/references/meta.json` 的 `pages` 数组显式枚举了 14 个分类，**不含 `contracts`**，所以导航里不会出现一个空的 "Contracts Protocol" 组；
- 没有任何页链接到 `/docs/references/contracts`；
- `Check Documentation Links`（lychee）今天是绿的。

[...]

PR #7507, opened 2026-08-11
chore(spec): retire the hand-written config-schema.json, whose generated replacement already exists

### LINKED ISSUE #7437: [finding] `packages/spec/config-schema.json` is unguarded and has drifted from the acceptance faces on its two remaining enums (6 phantom field types)
Observation-class finding, surfaced while fixing #7286 (PR #7436) and filed unassigned so it does not evaporate with that PR. #7286 fixed **one** entry in this artifact — the object-level `ownership` enum. This card records what the same sweep found in the **rest** of the file, plus the mechanism that let all of it drift.

Measured at `origin/main` @ `07383fe` by parsing the artifact's JSON and diffing each enum against the zod enum literal it mirrors.

## 1. `objects[].fields.*.type` — 6 phantom values, 21 hidden values

The artifact offers 34 field types. `FieldType` (`packages/spec/src/data/field.zod.ts:17`) accepts 49. They are not nested sets — the artifact offers values that are accepted nowhere:

- **PHANTOM (artifact offers, schema rejects), 6:** `integer`, `slug`, `uuid`, `ip_address`, `geo_point`, `encrypted`
- **MISSING (schema accepts, artifact hides), 21:** `secret`, `toggle`, `radio`, `checkboxes`, `tree`, `user`, `avatar`, `video`, `audio`, `summary`, `composite`, `repeater`, `record`, `location`, `address`, `code`, `slider`, `signature`, `qrcode`, `progress`, `tags`

Two of the phantoms look like renamed-and-never-followed spellings: `geo_point` against the accepted `location`, and `encrypted` against the accepted `secret`. This is the **same defect class as #7286** — IDE autocomplete steering an author to a value no schema accepts — just 6 values wide instead of 2.

It was deliberately **not** folded into PR #7436: aligning a 34-value list to a 49-value list is not the identical mechanical fix, because it first needs a decision (below) on whether the artifact is meant to offer a curated subset at all.

## 2. `data[].mode` — 1 hidden value

Artifact has `["upsert","insert","ignore","replace"]`; `SeedMode` (`packages/spec/src/data/seed.zod.ts:12`) has 5, including **`update`**. Weaker than the above — it hides a valid value rather than offering an invalid one, so no author is steered into a rejected write.

## 3. The mechanism — why all three drifted

- **Nothing guards it.** No pin/sync test, no gate, no consumer. The only in-repo mentions of `config-schema.json` are prose in one code comment, `CHANGELOG.md`, and a changeset. `check:generated` does not cover it. Nothing can go red when an acceptance face moves.
- **Its provenance claim is false.** The file's own top-level `description` says it is "Generated from ObjectStackDefinitionSchema". It is hand-written. That false claim is plausibly *why* nobody re-derived it as the vocabularies moved — it reads as already-generated.
- **It is not shipped.** `npm pack --dry-run --ignore-scripts` on `@objectstack/spec` packs 269 entries; zero match `config-schema`. It is absent from the package's `files` list, so no npm consumer can reach it.
- **A correct generated route already exists.** `objectstack generate schema` (`packages/cli/src/commands/generate.ts:817-865`) builds this exact schema live from `ObjectStackDefinitionSchema` via `z.toJSONSchema` — it cannot drift by construction. It writes `objectstack.schema.json` into the user's cwd, not this path.

## The decision this needs

Given that the generated route exists and the hand-written file is unshipped and unguarded, the choice is a route decision rather than a patch, and it should be made once for the whole file rather than per-enum:

- **A — retire the artifact.** Delete `packages/spec/config-schema.json` and point authors at `objectstack generate schema`. Removes the drift surface permanently instead of re-synchronising it by hand every time a vocabulary moves. Cost: anything outside this repo pointing at the path by URL or relative path breaks — though it is unshipped, so no npm consumer can be relying on it.
- **B — regenerate it in place.** Add a `gen:`/`check:` script pair that writes this path from `ObjectStackDefinitionSchema` and fails CI when it is stale, the same shape as the package's other generated baselines. Keeps the path alive and makes drift impossible. Cost: one more generated artifact plus its gate.
- **C — hand-align the two remaining enums and correct the false provenance line.** Cheapest now; leaves the drift surface and its silence intact, so this exact card recurs on the next vocabulary move. Given ADR-0049's enforce-or-remove posture, C is the weakest of the three.

No recommendation is being forced here — this is filed as `finding`, not queued, and the three options above are the material for triage. Options A and B both structurally prevent the class; C does not.

Refs: #7286 (the `ownership` entry, fixed), PR #7436, #3244 (confusion class).

os-help:
Fixes #7437

Deletes `packages/spec/config-schema.json`, per the **maintainer ruling of 2026-08-11** (option **A** on the card; B — regenerate + gate — and C — hand-align — are settled and closed). The whole diff is one deletion, 279 lines. **No enum was aligned on the way out.**

## Why delete rather than fix

The artifact was hand-written while claiming, in its own top-level `description`, to be *"Generated from ObjectStackDefinitionSchema"*. That false provenance is plausibly why nobody re-derived it as the vocabularies moved — it reads as already-generated. Nothing guarded it, so it drifted on every enum it carried (`ownership`, fixed by #7436; 6 phantom field types; a hidden `SeedMode` member), and its `$id` is still pinned at `v3.0.0` while the spec is v17.

The correct routes already exist, and both are drift-proof by construction:

| Route | Path | Guarded by |
| --- | --- | --- |
| Config schema for IDE autocomplete | `objectstack generate schema` — `packages/cli/src/commands/generate.ts:817-865`, built live from `ObjectStackDefinitionSchema` via `z.toJSONSchema` | cannot drift — derived at run time |
| Published per-schema JSON Schemas | `packages/spec/json-schema/**`, generated by `packages/spec/scripts/build-schemas.ts` | `check:generated` |
| **This artifact (deleted)** | **`packages/spec/config-schema.json`, hand-written** | **nothing — `check:generated` has 0 hits for it** |

## The delete lands and stays landed — measured before deleting

This lane's #7303 was a ruled delete that **could not land** because a generator recreated the file, so all three mechanism checks were re-run here rather than assumed:

| Check | Command | Reading |
| --- | --- | --- |
| Routed `merge=os-regen`? | `git check-attr -a packages/spec/config-schema.json` | **no** — empty output. Positive control: the same command on `packages/spec/spec-changes.json` prints `merge: os-regen` |
| Any generator writes this path? | exact-filename `git grep -F 'config-schema.json'` | **no** — 3 hits, all prose (below). The apparent hit at `scripts/docs-audit/affected-docs.mjs:324` is `config-schemas.test.ts` (plural, `service-automation`) — a substring false positive, excluded by searching the exact filename |
| Shipped? | `npm pack --dry-run --ignore-scripts` | **no** — see the A/B below |

Also ruled out: nothing serves the file at its declared `$id` host. The 7 repo mentions of `schema.objectstack.io` are the CLI's live `$id`, `build-schemas.ts`'s generated `$id` base, this file, and release history — no deploy step copies this path anywhere public.

### Unshipped, proven from the packing side

Not trusting the `files` list — one fixed post-build tree, with only the file toggling:

```
WITHOUT (HEAD, file deleted):  entryCount=2030   config-schema matches=0
WITH    (file restored)     :  entryCount=2030   config-schema matches=0
diff of the two packed file lists → identical
```

Same count, same 2030 paths, both times zero matching `config-schema`. No npm consumer can observe this removal.

## The ruling's "prose mentions" clause has no referent — nothing else was touched

The ruling says *"Remove the artifact plus its prose mentions; point any doc that referenced it at the generate command."* Searched with the **exact filename** (not the `config-schema` substring that produces the false positive above), the complete set of surviving mentions repo-wide is three, and **none is a pointer telling an author to read or fetch the file**:

1. `.changeset/http-server-config-retired.md:18` — a pending changeset, i.e. an unreleased release note
2. `packages/spec/CHANGELOG.md:9179` — the same sentence, already published history
3. `packages/spec/src/system/http-server.zod.ts:34` — a code comment narrating a **past** defect

All three are statements about the past that remain true once the file is gone. Rewriting (1) or (2) would falsify a release record; (3) is a true historical sentence, not a reference. **All three are left byte-identical.**

An independent sweep for a doc that steers authors at a config JSON Schema found `content/docs/api/data-flow.mdx:202`, which already says *"Autocomplete and validation for `objectstack.config.ts` (via `os generate schema`)"* — it names the generate command and never names this file. The ruling's intended destination is therefore already in place, and no doc needed repointing.

## Verification

Since nothing consumes the artifact, "CI is green" proves little here — so the evidence is direct:

- **File is gone** — `git rm`, `ls` reports no such file; commit shows `1 file changed, 279 deletions(-)`.
- **Exact-filename sweep after the delete returns exactly the three historical mentions above.** Positive controls proving the scanner is not blind: `ObjectStackDefinitionSchema` returns 53 hits in `packages/spec/src`; and the artifact's own unique string `https://schema.objectstack.io/v3.0.0/objectstack.config.json` went from 1 hit to 0 across the same delete.
- **`pnpm --filter @objectstack/spec build`** — green (exit 0).
- **`pnpm --filter @objectstack/spec test`** — green (exit 0) — **374 test files / 9802 tests passed**.
- **`npm pack` A/B** — above.

## No changeset — the precedent is this same file

The artifact is unshipped, so no npm consumer can observe its removal. The governing precedent is **PR #7436**, merged 2026-08-10, which edited *this exact file*, measured it unshipped with `npm pack --dry-run`, and shipped under the **`skip-changeset`** label with no changeset. This PR is the same file, the same measurement, and a strictly smaller consumer surface (removal of something nobody could reach), so it follows that disposition.

⛔ The label is deliberately **not** applied by this PR — the dispatching PM seat applies `skip-changeset` at accept time.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Uy9myDBjbJg4cNicejxnWM)_


os-help:
**Check Changeset is red, and this is the expected red — it needs a label I am not permitted to apply.**

The gate's own message is the whole diagnosis, with no second cause in the log:

> This PR adds no changeset. If it releases nothing … apply the `skip-changeset` label; otherwise run `pnpm changeset` and name the packages.

This PR releases nothing. The deleted artifact was never shipped, measured from the packing side rather than from the `files` list — one fixed post-build tree with only the file toggling:

```
WITHOUT (HEAD, file deleted):  entryCount=2030   config-schema matches=0
WITH    (file restored)     :  entryCount=2030   config-schema matches=0
diff of the two sorted packed-file lists → empty
```

The published tarball is identical either way, so no npm consumer can observe the removal. The precedent is this same file: PR #7436 edited `packages/spec/config-schema.json`, took the same `npm pack` measurement, and merged with no changeset under `skip-changeset`.

**Why I am not fixing it myself:** the dispatch for #7437 reserves label application to the PM seat (`⛔ Do not apply any label yourself — the PM applies skip-changeset at accept time if that is the answer`), which overrides the standing os-dev guidance to self-apply. Writing an empty changeset is not an alternative — the gate rejects newly added ones (#5471), and a real one would announce a removal nobody can see.

So this job stays red until `skip-changeset` lands, and nothing on my side changes that. Every other gate is unaffected by this failure; local evidence is `pnpm --filter @objectstack/spec build` exit 0 and the spec suite at 374 files / 9802 tests passed. Full report: #7437 (comment 5248896486).

---
_Generated by [Claude Code](https://claude.ai/code/session_01Uy9myDBjbJg4cNicejxnWM)_


---

Issue #7603, opened 2026-08-11
finding: `DELETE /reports/schedules/:scheduleId` has the same enumeration oracle #7523 closed, in the 404-vs-204 costume — and `rest.test.ts:1651` pins the leaking arm green


os-help:
Found by the dev implementing **#7523** (PR #7562) and deliberately **not** fixed there: it is a different handler in a file three cards were editing concurrently, so it wants its own card. Filing unassigned; grading is triage's channel.

## The finding

`DELETE /api/v1/reports/schedules/:scheduleId` discriminates on existence exactly the way `DELETE /reports/:id` did before #7523:

| Target | Response |
| :-- | :-- |
| A schedule id that does not exist | **204** — `unscheduleReport()` returns early (`if (!schedule) return; // idempotent`) |
| Another owner's schedule id | **404** — throws `REPORT_NOT_FOUND` → `handleValidation` |

Same oracle, quieter costume. #7523's card warned about precisely this shape — *"align the unknown-id arm to the same status so the two arms are byte-indistinguishable; otherwise 404-vs-204 remains the same oracle in a quieter costume"* — and that warning turns out to describe **this** route as it stands today.

## Why it was read as correct

This handler is the one the #7523 investigation cited as **proof of the right shape**, because it *does* route its catch through `handleValidation` — which is why the cross-owner arm answers 404 rather than 500. The QA run (#7515) exercised only that cross-owner arm and recorded it as clean. The unknown-id arm was never probed on this route, so the surviving half went unseen.

⚠️ **A pin currently locks the leak in place**: `packages/rest/src/rest.test.ts:1651` asserts the unknown arm is **204**. Any fix must supersede that pin explicitly — replacing it with the opposite assertion on the same input, not deleting it quietly.

## Suggested shape, if promoted

The same one #7523 landed, which is now precedent in the same file: answer **both** deny arms from **one** response emitted before the unschedule fires, via a visibility read that is already blind to the difference, so status and body cannot drift apart. Assert the two arms' **whole responses are equal** rather than pinning each arm's status separately — #7523's mutation table shows a per-arm assertion cannot fail on the half-fix.

Worth deciding at the same time whether any **other** `unschedule`/idempotent-early-return handler on this surface has the same split; #7523's fix covered only `deleteReport`.

## Provenance

[...]

Issue #7566, opened 2026-08-11
finding: `GET /api/v1/meta/app?id=…` — the `?id=` filter is inert, the same 3 apps come back for any value


huangyiirene:
## Symptom

`GET /api/v1/meta/app?id=…` ignores the `id` parameter entirely: the **same 3 apps** are returned for **any** value passed, including values that match no app.

Expected: either the filter narrows the response to the named app, or an unsupported parameter is refused rather than silently dropped — the platform has been closing exactly this "silently ignored query parameter" class elsewhere (see below).

Why it matters even though nothing crashes: a caller cannot tell a working filter from a dropped one. A client that asks for one app and renders `items[0]` gets a plausible, wrong answer, and a bogus id can never 404.

## Root cause

Not located by the report — it observed the endpoint's behaviour only. Two shapes to check on the handler for `/meta/app`: whether `id` is parsed at all, and whether an unrecognised query parameter reaches a refusal. Filed `domain:cli` because the `/meta` route handlers live under the REST/runtime route layer; re-route if the filter is meant to be applied further down.

## Related — the same class, being closed one endpoint at a time

- objectstack-ai/objectstack#7527 (open): `assignedToMe=true` is not a supported list filter on `/api/v1/approvals/requests` — silently ignored, returns every request. Same defect shape on a different domain.
- Recent merges in this class: objectstack-ai/objectstack#7490 (`GET /automation/:name/runs?status=` filters instead of being dropped), #7350 and #7299 (malformed query params refused with an ADR-0112 400), #7324 / #7396 (repeated query parameters). None of them touch `/meta/app`, so this instance is still open; whoever fixes it should follow whichever convention those settled on (filter, or refuse) rather than inventing a third.

## Reproduction

1. Boot the showcase (3 apps published).
2. `GET /api/v1/meta/app?id=&lt;a real app id&gt;` → all 3 apps.
3. `GET /api/v1/meta/app?id=&lt;a nonsense value&gt;` → the same 3 apps, byte-identical response.

## Source

Extracted from the QA run objectstack-ai/objectstack#7514 (framework a86db175, console 09987b68).

os-zhuang:
[...]

Issue #7563, opened 2026-08-11
`POST /api/v1/packages/publish` answers a misleading 405 (`Allowed: DELETE, GET, HEAD, PATCH`) — the request is absorbed by `/packages/:id` because the REST package registrar is not mounted on showcase


huangyiirene:
## Symptom

`POST /api/v1/packages/publish` on a live showcase boot answers **405** with `Allowed: DELETE, GET, HEAD, PATCH`.

That method list belongs to a **different route**. `POST` is the *only* verb the publish surface ever had, so a 405 naming four other verbs tells the caller "this path exists, just use another method" — when in truth no publish surface is served on this host at all, and every verb in that list would be operating on a *package whose id is the literal string `publish`*.

**Expected:** either the real publish handler (the route is ledgered — `POST /api/v1/packages/publish` in `packages/rest/src/rest-route-ledger.ts`, `source: direct-mount`, `disposition: server-only`), or an honest **404** for a surface that is not mounted here. A 405 whose `Allowed` set describes another route's methods is the one answer that misinforms.

## Root cause

As far as the run locates it: the **REST package registrar is not mounted on showcase**. `POST /packages/publish` is owned by `packages/rest/src/package-routes.ts` (which is where marketplace publish landed — the file says so explicitly: *"Marketplace publish lives at `/packages/publish`, NOT at the bare …"*), and those REST package routes are service-gated, so on a stack without the package service they never register.

With the registrar absent, the path is still matched — by `/packages/:id` with `id = "publish"` — and the router derives its `Allowed` header from *that* route's method set, which is exactly `DELETE, GET, HEAD, PATCH`. The path is therefore "reachable" only by accident of the id-parameter route, and the 405 is generated by the wrong owner.

Related, closed: objectstack-ai/objectstack#3610 (`POST /api/v1/packages` publish-vs-install shape collision, closed by PR objectstack-ai/objectstack#3621) is the reconciliation that moved publish onto `/packages/publish` in the first place. This is not a regression of that fix's stated acceptance — it is the residue it did not cover: what the `/packages` prefix answers on a host where only one of the two owners is mounted.

Still live on `origin/main` as of 2026-08-11 (route present in `package-routes.ts` and in the REST ledger; gating unchanged).

## Reproduction

1. Boot showcase on a fresh isolated file DB (`SqlDriver` / better-sqlite3); authenticate as `admin@objectos.ai`.
2. `POST /api/v1/packages/publish` with any body.
[...]

Issue #7543, opened 2026-08-11
A raw `TypeError: not a function` leaks in the 400 envelope for `{"title": 12345}` — no `code`, no `fields[]`


huangyiirene:
## Symptom

`POST /api/v1/data/showcase_task` with a number in a declared `text` field:

```json
{"title": 12345}
```

answers:

```
400 { "error": "TypeError: not a function", "object": "showcase_task" }
```

— a raw JS runtime error as the client-facing message, with **no `code`** and **no `fields[]`**. Reproduced 3x.

**Expected:** the ledgered envelope the neighbouring invalid bodies produce on the same route in the same run:

| body | observed | |
|---|---|---|
| `{}` | `400 VALIDATION_FAILED` with `fields[]` | correct |
| bad enum value | `400` with `invalid_option` and the allowed list | correct |
| `{"title": 12345}` | `400 "TypeError: not a function"`, no `code` | **defect** |

Two separate contract breaks in one response: a handler-internal `TypeError` reaching the wire verbatim, and an error body outside the ledgered envelope shape (a client keying on `code` / `fields[]` gets neither).

## Root cause

**Not fully located in the report.** Its suspicion: a coercion path calls a string method on a number **before** validation reports the type mismatch, so the write throws a `TypeError` instead of producing a field error.

Investigation note from extraction (working tree checked 2026-08-11, so verify against `a86db175` before acting): the observed body shape — status `400`, raw message, an `object` key, and no `code` — matches **no** branch of `mapDataError` in `packages/rest/src/rest-server.ts` on current main. #5489 moved that mapper's terminal branch to a sanitised `500 INTERNAL_ERROR` (no `object` key), the declared-4xx passthrough in `resolveErrorResponse` does not attach `object`, and every remaining 400 branch carries a `code`. Likewise `validateOne` in `packages/objectql/src/validation/record-validator.ts` handles a number in a `text` field safely (`String(value)`), so the throw is upstream or downstream of it. Locating the emitting seam is step 1 of the fix.

**No `domain:*` label applied deliberately:** the fix could land in `packages/rest` (envelope classification ⇒ `domain:cli`) or in `packages/objectql` (the coercion that throws ⇒ `domain:engine-core`), and the extraction pass could not tell which from the report plus a tree read. Label it once the throw site is found.

## Reproduction

1. Boot showcase on a fresh isolated file DB (`SqlDriver` / better-sqlite3); authenticate as `admin@objectos.ai`.
[...]

PR #7660, opened 2026-08-11
fix(runtime): serve the hook/action `['log']` capability from a logger that exists (#7448)

### LINKED ISSUE #7448: [finding] Hook body ctx.log output is unobservable — body-runner wires log: engineCtx?.logger with no fallback, so the declared ['log'] capability can silently produce nothing
Recorded from the QA run #7439 (records-forms FULL area, `object-hook-lifecycle` PARTIAL — the run's one framework-side observation). Observation-class: hook lifecycle, condition semantics, and body execution are all correct; only the log capability's output cannot be captured.

## The measurement (QA run, showcase app, `--log-level debug`)

Hook bodies fire (`[BodyRunner] hook fired` appears) and do not throw, but the named log lines the hook bodies emit via `ctx.log.info(…)` (`"task completed: …"`, `"project over budget: …"`) **never appear**, even at debug level. Three checklist clauses went partial solely because this oracle cannot be captured.

## Anchors (verified at `origin/main` @ `d13ce33`)

`packages/runtime/src/sandbox/body-runner.ts`:

- `:339` — hook context: `log: engineCtx?.logger`
- `:377` — action context: `log: actionCtx?.logger`

Both wire the `['log']` capability directly to the caller-supplied logger with no fallback and no warning when it is absent. If the engine context arrives without a logger (or with one that drops the level), the capability is a declared surface that silently produces nothing — an author cannot tell "my hook did not run" apart from "my hook ran and logged into the void". The BodyRunner's own diagnostics (`opts.logger?.debug` at `:74`) use a different logger (`opts.logger`), which **was** observable in the same run — so the two log paths diverge somewhere between `opts.logger` and `engineCtx.logger`.

## Why finding, not queue

No incorrect behavior ships: hooks execute, conditions gate, data is right. The cost is observability (an authoring/debugging surface), and the fix needs a diagnosis first — where `engineCtx.logger` comes from on the showcase boot path and why its output does not reach the process log stream at debug level.

## Dedup

Searched open issues (`body-runner`, `ctx.log`, `hook log`): no existing card. #7365 (QA assertion `field` path) is a different QA-tooling defect.

本评论来自分诊座位 Routine（#5474 试点），不构成认领。

os-help:
Fixes #7448

## Diagnosis — it is **(a)**: the context arrives with no logger at all

Not "on the showcase boot path" — on **every** path. `engineCtx.logger` is a key that **no producer writes and no schema declares**.

| Anchor (re-measured on `main` @ `245d1dc`) | Finding |
|:---|:---|
| `packages/runtime/src/sandbox/body-runner.ts:339` | `log: engineCtx?.logger` — the hook wiring |
| `packages/runtime/src/sandbox/body-runner.ts:377` | `log: actionCtx?.logger` — the action wiring |
| `packages/spec/src/data/hook.zod.ts:334` (`HookContextSchema`) | declares **no** `logger` — `grep -c logger` over the file returns **0** |
| `packages/objectql/src/engine.ts:6832, 6972, 7825, 8857` | all four `const hookContext: HookContext = {…}` assembly sites — **none** writes a `logger` key |
| `packages/runtime/src/domains/actions.ts:330` and `packages/runtime/src/action-execution.ts:1044` | both action-context literals — **neither** writes a `logger` key |

**The divergence point the card asked for:** `packages/objectql/src/engine.ts:1874`. That is where the engine hands its own logger to the BodyRunner —

```js
const merged = { ...(opts ?? {}), logger: this.logger } as any;   // → body-runner opts.logger
```

— and there is **no counterpart on the per-invocation context**. So `opts.logger` (`body-runner.ts:74`, `[BodyRunner] hook fired`) resolves to the engine's real `Logger`, while `engineCtx.logger` (`:339`) reads a key nobody ever set. One logger, one `undefined`, from the same object graph. That is the whole of the divergence.

Downstream it becomes silent rather than fatal at `packages/runtime/src/sandbox/quickjs-runner.ts:646`:

```js
ctx.log?.[level]?.(msg, data);   // optional call on an absent host seam
```

The capability gate passes, the VM-side `ctx.log.info` exists, the body runs to completion and returns normally — and the line goes nowhere. Exactly the "cannot tell *did not run* from *ran and logged into the void*" shape the card names.

This is the **third** limb of this shape removed from this same file, not the first: `doc`/`previousDoc` (#5906) and `session.user` (#6316) were also keys no producer ever wrote, deleted rather than left as a second de-facto contract (Prime Directive #12).

## Reproduction — showcase, `--log-level debug`, before → after

Booted the way QA run #7439 did (`pnpm dev -- --fresh -p <port> --log-level debug`), signed in as the seeded admin, and `PATCH`ed a `showcase_task` from `done: false` → `true` to fire `showcase_audit_task_completion` (whose body is `ctx.log.info('task completed: ' + …)`, capabilities `['log']`).

**BEFORE** (the two source files reverted to `origin/main`, `@objectstack/runtime` rebuilt):

```
2026-08-11T09:55:55.836Z DEBUG [BodyRunner] hook fired {"appId":"com.example.showcase","hook":"showcase_normalize_task_title"}
2026-08-11T09:55:55.851Z DEBUG [BodyRunner] hook fired {"appId":"com.example.showcase","hook":"showcase_audit_task_completion"}

$ grep -c "task completed" /tmp/showcase-before.log
0
```

The hook fires, does not throw — and the line it emits is absent from the **entire** debug log. That is the measurement.

**AFTER**:

```
2026-08-11T09:54:36.165Z DEBUG [BodyRunner] hook fired {"appId":"com.example.showcase","hook":"showcase_audit_task_completion"}
2026-08-11T09:54:36.294Z INFO  [hook 'showcase_audit_task_completion'] task completed: Ingest pipeline
```

## The fix, and why this shape follows from the cause

The cause is a read of a key nobody writes, so the fix is to **read the source that exists** — not to add a fallback behind the dead one.

1. **Serve the capability from `opts.logger`.** It is the engine's own `Logger`, passed by all four construction sites in `app-plugin.ts` (`logger: ctx.logger`), and it is *the very logger whose `[BodyRunner] hook fired` was already observable in the same QA run*. So "why does this reach the process log stream when the current wiring does not" has a measured answer rather than an argued one: it is the stream the run already proved reachable.
2. **Delete the `engineCtx.logger` / `actionCtx.logger` limbs** instead of keeping them as a preferred source (`engineCtx?.logger ?? …`). Under PD #12 a key no producer writes and no schema declares is not a contract to accommodate. The alternative — declaring `logger` on `HookContextSchema` — would widen the metadata contract to re-supply, per invocation, something the runner already holds for the lifetime of the bind.
3. **Attribute the line.** `[hook '<name>']` / `[action '<name>']` prefix, because an author running many hooks otherwise cannot tell which one spoke, and the BodyRunner is the only layer that knows. The message body is untouched, so `task completed: …` still matches as a substring.
4. **Dispatch `error` through the real contract.** `Logger.error` is `(message, error, meta)` — three args (`packages/spec/src/contracts/logger.ts:36`). Passing the body's data second lands a meta object in the `Error` slot, where `ConsoleLogger`/`JsonLogger` read `error.message`/`error.stack` as `undefined` and drop every field. Same trap `hook-wrappers.ts` documents for `HookDiagnosticsLogger`; the test asserts against a three-arg double so a regression here shows up as a lost `meta`, not as a silent pass.

### Both call sites are covered

Hook (`:339`) and action (`:377`) both go through one `buildBodyLogSurface(opts, origin)`. The action path needed the identical fix for the identical reason — neither of its two context assembly sites writes `logger` either — and its test is in the same file.

### Second defect in the same capability, found by the reproduction

The VM bridge read the optional `data` argument with `vm.getString`, which applies JS string coercion **inside** the VM. So `ctx.log.info('msg', { code: 'E1' })` arrived at the host as the literal string `"[object Object]"`, which `safeJsonParse` then failed to parse and returned verbatim — every structured field of every body log call was lost. Now `vm.dump`, the marshalling every other host-call bridge in that file already uses (`ctx.api`'s `argHandles.map((h) => vm.dump(h))`). In scope because it is the payload half of the same declared surface: fixing the destination while leaving the payload mangled would be half a fix.

### When the capability genuinely cannot be served

The closing condition is *either it works, or the author is told it cannot*. With no logger on the factory at all, working is off the table, so this takes the told branch: **warn once per invocation**, naming the body and the remedy.

It deliberately does **not** fall back to `console`. That would override a decision belonging to the host — a `Logger` carries the level threshold, formatting and sinks the host chose — and a host running at `warn` would start receiving body `info` lines on an unfiltered second stream it never configured. No production path reaches this branch (all four `app-plugin.ts` sites pass `ctx.logger`), so it is a diagnostic for embedders constructing the factory directly. Once per invocation rather than once per call, so a chatty body cannot bury the rest of the log.

## Verification

The new test asserts **observability** — what the host logger actually received after a real QuickJS body ran — not that the wiring is non-null. A null-check test would pass on a logger that drops the level, which is the wrong oracle for an observability defect.

**Fails without the change** (all 5, on unmodified source):

```
 FAIL  src/sandbox/body-log-capability.test.ts > emits the body's info line on the logger the factory was constructed with
 FAIL  src/sandbox/body-log-capability.test.ts > emits warn and error levels too, and keeps the body's data out of the Error slot
 FAIL  src/sandbox/body-log-capability.test.ts > attributes the line to the emitting hook
 FAIL  src/sandbox/body-log-capability.test.ts > warns once, naming the hook, when the capability cannot be served
 FAIL  src/sandbox/body-log-capability.test.ts > [action] emits the body's log line on the factory logger

 Test Files  1 failed (1)
      Tests  5 failed (5)
```

Note the action case failed at the *log* assertion while `expect(value).toEqual({ ok: true })` passed — the body ran to completion and logged into the void.

**Targeted test, with the change:**

```
$ pnpm --filter @objectstack/runtime exec vitest run src/sandbox/body-log-capability.test.ts
 Test Files  1 passed (1)
      Tests  5 passed (5)
   Duration  1.38s
```

**ESLint** (changed files):

```
$ npx eslint packages/runtime/src/sandbox/body-runner.ts \
             packages/runtime/src/sandbox/quickjs-runner.ts \
             packages/runtime/src/sandbox/body-log-capability.test.ts
ESLINT EXIT CODE: 0
--- output (empty = clean) ---
(no findings)
```

**TypeScript:**

```
$ pnpm --filter @objectstack/runtime typecheck
> @objectstack/runtime@17.0.0-rc.6 typecheck
> tsc --noEmit
typecheck exit=0
```

**Full runtime suite (regression):**

```
$ pnpm --filter @objectstack/runtime test
 Test Files  128 passed (128)
      Tests  2021 passed (2021)
   Duration  65.99s
```

`pnpm build` also green end to end (71/71 tasks) — that is the build both showcase boots above ran against.

## Scope

Confined to `packages/runtime/src/sandbox/` (`body-runner.ts`, `quickjs-runner.ts`, one new test) plus a changeset. No contact with the three live regions in this wave — `rest-server.ts` (#7603, #7566), `package-routes.ts` (#7563), the 400-envelope work (#7543). The diagnosis reached into `packages/objectql/src/engine.ts` and `packages/spec/src/data/hook.zod.ts` to establish that nobody writes the key, but **neither needed changing**: the correct logger was already in the BodyRunner's hand.

No `content/docs/releases/` edit; the changeset is the release-notes input. No `docs/adr/**` contact.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KmZEFWeuDEoxvnxpMWznbE)_

### DIFF
--- a/.changeset/hook-body-log-capability-observable.md
+++ b/.changeset/hook-body-log-capability-observable.md
@@ -1,0 +1,44 @@
+---
+"@objectstack/runtime": patch
+---
+
+fix(runtime): a hook/action body's `ctx.log` output reaches the host log stream (#7448)
+
+A body that declared the `['log']` capability and called `ctx.log.info(…)` ran to
+completion, returned normally, and produced **nothing** — an author could not tell
+"my hook did not run" apart from "my hook ran and logged into the void". QA run
+#7439 measured it on the showcase at `--log-level debug`: `[BodyRunner] hook fired`
+appeared, while the `task completed: …` line the body itself emitted did not.
+
+**Cause.** `body-runner.ts` wired the capability to `engineCtx?.logger` (hooks) and
+`actionCtx?.logger` (actions) — a key **no producer writes**. `HookContextSchema`
+declares no `logger`, ObjectQL's engine builds all four of its HookContexts without
+one, and neither action-context assembly site (`domains/actions.ts`,
+`action-execution.ts`) writes one either. So `ctx.log` was `undefined` on every
+path, and the VM bridge forwards through `ctx.log?.[level]?.(…)` — an optional call
+on an absent seam. The BodyRunner's own diagnostics were visible throughout because
+they use a different logger (`opts.logger`), which every construction site does
+supply.
+
+**Fix.** The capability is now served from `opts.logger` — the engine's own
+`Logger`, handed to the factory by all four `app-plugin.ts` sites, and the same one
+whose `[BodyRunner] hook fired` was already observable. The dead `engineCtx.logger` /
+`actionCtx.logger` limbs are removed rather than kept as a second de-facto contract,
+matching the `doc`/`previousDoc` (#5906) and `session.user` (#6316) removals in this
+file. Lines are prefixed with their origin (`[hook 'showcase_audit_task_completion']
+task completed: …`) so an author running many hooks can tell which one spoke, and
+`error` is dispatched through the `Logger` contract's real `(message, error, meta)`
+signature so a body's structured data no longer lands in the `Error` slot and lose
+every field.
+
+**Second defect, same capability.** The VM bridge read the optional `data` argument
+with `vm.getString`, which coerces inside the VM — so `ctx.log.info('msg', { code:
+'E1' })` arrived at the host as the literal string `"[object Object]"` and every
+structured field was lost. It now uses `vm.dump`, the marshalling every other
+host-call bridge in that file already uses.
+
+When a BodyRunner is constructed with no logger at all — no production path, but
+reachable for embedders — the capability no longer degrades silently: it warns once
+per invocation, naming the body and the remedy. It deliberately does not fall back
+to `console`, which would override the level threshold, formatting and sinks the
+host chose and start a second, unfiltered log stream it never configured.
--- a/packages/runtime/src/sandbox/body-log-capability.test.ts
+++ b/packages/runtime/src/sandbox/body-log-capability.test.ts
@@ -1,0 +1,172 @@
+// Copyright (c) 2025 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#7448] The `['log']` capability must reach the host log stream.
+ *
+ * These tests assert OBSERVABILITY, not wiring. A null-check ("`ctx.log` is
+ * not undefined") passes on a surface that swallows every call, which is the
+ * exact defect #7448 recorded: QA run #7439 saw `[BodyRunner] hook fired` at
+ * `--log-level debug` while the `ctx.log.info('task completed: …')` the body
+ * emitted never appeared anywhere. So every assertion here is on what the
+ * HOST LOGGER RECEIVED after a real QuickJS body ran.
+ *
+ * The captured logger is a `Logger`-contract double: `error` takes
+ * `(message, error, meta)`, so a body's `data` landing in the `error` slot
+ * would show up here as a lost `meta` rather than passing silently — the
+ * shape-dispatch trap `hook-wrappers.ts` documents for `HookDiagnosticsLogger`.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { hookBodyRunnerFactory, actionBodyRunnerFactory } from './body-runner.js';
+import { QuickJSScriptRunner } from './quickjs-runner.js';
+
+interface Line {
+  level: 'debug' | 'info' | 'warn' | 'error';
+  message: string;
+  error?: Error;
+  meta?: Record<string, any>;
+}
+
+/** A `Logger`-contract double — note `error`'s three-arg signature. */
+function captureLogger() {
+  const lines: Line[] = [];
+  return {
+    lines,
+    debug: (message: string, meta?: any) => { lines.push({ level: 'debug', message, meta }); },
+    info: (message: string, meta?: any) => { lines.push({ level: 'info', message, meta }); },
+    warn: (message: string, meta?: any) => { lines.push({ level: 'warn', message, meta }); },
+    error: (message: string, error?: Error, meta?: any) => {
+      lines.push({ level: 'error', message, error, meta });
+    },
+  };
+}
+
+describe('[#7448] hook body ctx.log reaches the host log stream', () => {
+  const runner = new QuickJSScriptRunner();
+
+  it('emits the body\'s info line on the logger the factory was constructed with', async () => {
+    const logger = captureLogger();
+    const factory = hookBodyRunnerFactory(runner, { ql: {}, appId: 'showcase', logger });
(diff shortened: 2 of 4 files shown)


os-help:
Adjacent finding filed, not fixed here: **#7661** — `ctx.log.debug` is matched by the CLI capability extractor (`extract-hook-body.ts:56`) and listed in `content/docs/automation/hook-bodies.mdx:317`, but the VM installs only `info`/`warn`/`error` (`quickjs-runner.ts:639`), so a body calling it throws `TypeError: not a function`. Measured, not read off the source.

Kept out of this PR deliberately: #7448 is about the three installed methods producing nothing, whereas that is a fourth method that does not exist — and closing it is an enforce-or-remove capability-surface decision (the `crypto.hash` / #4391 shape), not a bug fix.

---

Issue #7557, opened 2026-08-11
Package disable and uninstall never reach the metadata/data layer: a disabled package's objects still serve rows, and uninstall leaves 7 orphaned `sys_metadata` rows


huangyiirene:
Blocked-by: #7705

&gt; **Status 2026-08-11 (engine-core seat):** half (A) and the envelope part of half (B) LANDED via PR #7700 (`8f1851ea7`). What remains open on this card is the **persistence half of (B)** only, transferred to the metadata lane as **#7705**. This card closes when #7705 lands. See the ACCEPT comment below for the full verification record.

Two halves of one gap: a package lifecycle operation moves the registry/UI surfaces but never propagates to the metadata store or the data plane. Filed as one card because both are the same missing propagation, on the two operations that need it.

## Symptom

**(A) Disable is not an enforcement for objects.** ✅ **FIXED by PR #7700.** With the owning package at `status:'disabled'`:

- `GET /api/v1/data/&lt;object&gt;` still answers **200 with all rows**. → now `404 OBJECT_PACKAGE_DISABLED`
- `/meta/objects` still lists the object. → now dropped from the listing
- Nav and views **are** correctly dropped — so disable is honoured on some surfaces and not others.
- The console renders the live list too, so this is not an API-only artefact.

Reproduced over **two** disable/enable cycles.

Expected: a disabled package's objects are dropped from the metadata registry and refuse on the data plane, consistently with the nav/view surfaces that already honour the status.

**(B) Uninstall leaves orphaned metadata.** ⚠️ **HALF fixed.** `DELETE /packages/:id` answers **200**, with:

```
persisted: { success: false, deletedCount: 0 }
```

The package disappears from `/packages`, but **7 `sys_metadata` rows survive** and `/meta/*` keeps serving them. Reproduced with a second, single-view package.

Expected: either the persisted delete succeeds and the metadata rows go with the package, or the failed persistence is reported as a failure instead of being wrapped in a 200.

- ✅ **The dishonest envelope is fixed** (PR #7700): the producer was the *dispatcher* twin `packages/runtime/src/domains/packages.ts`, not `packages/rest/src/package-routes.ts` as this card guessed — that REST door was already correct. The dispatcher now answers `400 PACKAGE_DELETE_PARTIAL` on per-item failures, matching the other door including its zero-rows carve-out.
[...]

Issue #7221, opened 2026-08-10
`MetadataFacade.unregisterPackage` removes only object contributors — every non-object item the package shipped stays registered


os-zhuang:
Observation-class finding, filed per Prime Directive #10 while implementing #6725 (PR #7211). Unassigned, deliberately **not** queued. Deliberately kept OUT of #6725: that card is the `register`/read split for the `object` type, this is a different verb with a different root cause, and folding it in would have widened the card.

## The fact (read on `origin/main` @ `55da611`)

`MetadataFacade.unregisterPackage` (`packages/objectql/src/metadata-facade.ts`) is one line:

```ts
async unregisterPackage(packageName: string): Promise<void> {
  this.registry.unregisterObjectsByPackage(packageName);
}
```

`SchemaRegistry.unregisterObjectsByPackage` walks `objectContributors` only. So for a facade-backed `metadata` slot, "unregister all metadata from a package" leaves behind:

1. **Every non-object item the package shipped** — its `page`, `view`, `flow`, `app`, `api` … entries all live in the generic `metadata` map under composite `${packageId}:${name}` keys (`registerItem`), and nothing here touches that map. They stay resolvable through `get`, `list`, `listNames` and `exists` after the package is gone.
2. **The generic-map half of its objects.** An object is written into two places (the invariant `SchemaRegistry.unregisterObject`'s header states); this verb reaches one. Post-#6725 the facade's own `register('object', …)` writes both, so this leaves a genuine orphan rather than a theoretical one.

Item 1 is the load-bearing half: it is not a stale-cache nuisance, it is uninstall leaving a package's UI and API metadata installed.

## Reachability

Same reachability as #6725, and it is the same argument: `new MetadataFacade(...)` appears nowhere on `main` outside `packages/objectql`'s own tests, so nothing in-tree calls this. But `MetadataFacade` is exported from `@objectstack/objectql`'s root and `core` entrypoints, and `unregisterPackage` is a declared (optional) member of `IMetadataService` — so a downstream host that installs the facade and uninstalls a package gets a half-uninstall, silently.

`MetadataManager` (`packages/metadata`) is the in-tree occupant of the slot and does not share this shape.

## Why it is not a one-line fix

`SchemaRegistry` has no "remove every generic item owned by package X" verb. The composite key format (`${packageId}:${name}`) makes the scan mechanically easy, but two questions need a ruling before writing it:

[...]

PR #7700, opened 2026-08-11
fix(engine-core): propagate package disable to metadata listings and the data plane (#7557)

os-help:
Part of #7557 — ships half (A) in full plus the envelope half of (B). The persistence half of (B) is **located with evidence and reported for transfer**, not patched from the consumer side (see "Half (B): what shipped and what is transferred").

## Verification record

Every premise re-measured against `origin/main` (`5db0dca` at branch point) on a **real running server** (`pnpm dev --fresh`, showcase app), not from the working tree.

### Premise (A) — CONFIRMED, and the asymmetry located

With a runtime-installed package holding two package-bound `sys_metadata` objects and three rows:

| Surface | before disable | **disabled** (pre-fix) |
|---|---|---|
| `GET /api/v1/data/reprob_a` | 200, rows=3 | **200, rows=3** ← served anyway |
| `GET /api/v1/meta/objects` (from pkg) | 2 | **2** ← still listed |
| nav / views | present | correctly gone |

The PM's pointer to `registry.ts:2157` (`isPackageDisabled`) and the disable-hiding read at `:2139` is accurate, and the asymmetry is one line above the filter: `listItems` **returns early** for `object`/`objects` at `registry.ts:2128-2131` (`return this.getAllObjects(packageId)`) before ever reaching its own disable filter at `:2145`. The protocol-level filter then exempts the same two types explicitly (`protocol.ts:4330-4334`).

Only **two** call sites consult the predicate repo-wide: `registry.ts:2145` and `protocol.ts:4336`. That is the whole enforcement surface.

### Premise (B) — the envelope producer is NOT where the card expected

The card pointed at `packages/rest/src/package-routes.ts:376`. **Falsified by measurement:** that route already handles the failure correctly — it answers `400 PACKAGE_DELETE_PARTIAL` when `result.failedCount !== 0` (`package-routes.ts:381-390`) and `400 PACKAGE_DELETE_FAILED` otherwise. It never produces a `persisted` key at all.

The real producer is the **dispatcher twin**, `packages/runtime/src/domains/packages.ts` — reproduced live, byte-for-byte the shape in the issue:

```
DELETE /api/v1/packages/com.repro.b  ->  HTTP 200
{"success":true,"data":{"success":true,"registryRemoved":true,
  "persisted":{"success":false,"deletedCount":0,"failedCount":0,...}}}
```
…with the package-bound `sys_metadata` rows still present afterwards. `DELETE /packages/:id` has **two doors** (this dispatcher and the direct-mount REST registrar, which shadows it only when a `package` service is registered); they disagreed, and the dispatcher is the one that was lying. That door is `packages/runtime` — engine-core territory, in lane.

## Half (A): the fix

The two skips were deliberate and both gave the same reason — filtering objects *"would break data queries that depend on their schema"*. That conflated two classes of reader, and this PR separates them explicitly:

- **Resolution readers keep serving.** `registry.getObject` and `registry.listItems('object')` still return a disabled package's objects. That is load-bearing, not caution: `protocol.ts:3119` resolves the **runtime authoring gate's object universe** through `listItems('object')`, so blanking it would make unrelated authoring fail reference validation. Migrations and cross-package references resolve through `getObject`. Untouched.
- **API readers now stop.** `getMetaItems` drops the `object`/`objects` exemption. `package` is still never filtered — filtering it would make disable *irreversible*, since the Packages page is the only surface that can re-enable.

Nothing in the repo reads `getMetaItems({type:'object'})` internally, so the exemption bought no safety and cost the enforcement.

### The "refuse loudly vs hide" choice — settled by measurement, not escalated

Recorded here per the three-axis discipline; it did not diverge on product semantics, so no `needs_decision`.

- **Consistency**: the closest existing sibling is `OBJECT_API_DISABLED` (`enable.apiEnabled: false`) → **404 with a distinct code** (`rest-server.ts:1602-1610`). "This object exists but is switched off" already had a house answer; this follows it rather than inventing a second one. It also keeps the data plane consistent with the listing, which now drops the object too — the object is absent everywhere on the API.
- **AI-error resistance**: a bare `OBJECT_NOT_FOUND` sends a caller — an AI agent especially — hunting for a typo or **re-creating an object that is merely switched off**. The distinct code names the cause and therefore the fix. This is why "hide silently" was rejected even though the status is the same.

```
404  {"error":"Object 'reprob_a' belongs to a disabled package and is not being served.
              Re-enable the package to restore access.",
      "code":"OBJECT_PACKAGE_DISABLED","object":"reprob_a"}
```

Verified on the wire for the list, single-record and write paths. Registered in the ADR-0112 ledger under `@objectstack/metadata-protocol`. Note the data-plane envelope keeps `code` as a sibling of `error` — that is the **existing** shape `OBJECT_NOT_FOUND` uses in `mapDataError`; flattening it to `error.code` per ADR-0112 D5 is pre-existing and out of scope here.

The gate lands in `assertObjectRegistered` — the single seam every `findData`/`getData` entry point already funnels through (`protocol.ts:6252`, `:6730`, `:6794`, `:6802`, `:6848`, `:6906`, `:7004`), so one gate covers every data entry point. It is optional-called (`typeof registry.isObjectPackageDisabled === 'function'`) because registry doubles across the suites implement `isPackageDisabled` but not this.

`SchemaRegistry.isObjectPackageDisabled` resolves names through the same private `resolveObjectKey` the read path uses — #6808's lesson applied to enforcement, so the gate and the read cannot disagree about which contributor entry a bare name addresses. It asks about the **owner**, so an overlay from a disabled package cannot take the object it decorates offline.

### Live after-state

| Surface | before | **disabled** | after re-enable |
|---|---|---|---|
| `GET /data/reprob_a` | 200 rows=3 | **404 `OBJECT_PACKAGE_DISABLED`** | 200 rows=3 |
| `/meta/objects` from pkg | 2 | **0** | 2 |
| `sys_metadata` rows | 2 | 2 | 2 |

**The card's honest negative is preserved and pinned:** disable destroys no data — rows and metadata identical before, during and after, and re-enable restores service completely.

## Half (B): what shipped and what is transferred

**Shipped (in lane):** the dispatcher door now answers `400 PACKAGE_DELETE_PARTIAL` on per-item failures, carrying the failed items and the uninstall cleanup outcomes (a failed permission revocation is a ghost grant — it must survive the failure path). The rule is copied deliberately from the REST door of the same route so the two doors stop disagreeing, **including its carve-out**: zero metadata rows is still a successful uninstall (a runtime-registered package that never published metadata has nothing in `sys_metadata`), so the predicate is `failedCount > 0`, not `!persisted.success`. Checked *before* the not-found test, because an all-rows-failed uninstall also has `deletedCount === 0` and answering 404 for rows that are demonstrably present and stuck is the same lie one layer over.

**Transferred (out of lane) — `domain:metadata`.** The persistence defect is separate and untouched: `protocol.deletePackage` (`protocol.ts:11486-11488`) runs

```ts
const where = { package_id: request.packageId };
if (request.organizationId) where.organization_id = request.organizationId;
const rows = await this.engine.find('sys_metadata', { where });
```

and found **zero** rows while three rows with exactly that `package_id` demonstrably existed — the same predicate returns 3 through the data plane (`GET /data/sys_metadata?filters=[["package_id","=","com.repro.b"]]` → 3 records). So the `where` clause is not intrinsically wrong; the scoping the protocol's engine applies to it is. The strongest candidate is the strict `organization_id` equality against rows stored env-wide (`organization_id IS NULL`) — **the same defect class this package already fixed once**, in `sys-metadata-repository.ts:887-901`, where the identical strict equality "silently dropped" env-wide drafts and was replaced by an `$or` over `[{organization_id: oid}, {organization_id: null}]` (the "orphaned draft" bug). Patching that from here would be papering over the producer, so it is reported rather than fixed.

Note this half is invisible to the envelope fix by construction: it yields `failedCount: 0`, so it legitimately still reads as success at the dispatcher. It stops being a lie only once the producer stops missing rows.

Open finding #7221 (`unregisterPackage` removes only object contributors) was read for shape; the measured fix path does **not** run through it, so it is untouched.

## Tests

**One table walks every reader** rather than one test per surface, since the bug *is* a divergence between readers of a shared predicate. `packages/objectql/src/package-disable-enforcement.test.ts` enumerates all six readers with the class each belongs to, and asserts the whole table in a single `toEqual` — deliberately not `expect()` inside the loop, which would throw on the first failing reader and hide the rest, in a file whose purpose is showing *which* readers diverged. Adding a reader means adding a row and declaring its class. Real `SchemaRegistry` + real protocol over an in-memory `sys_metadata` engine; a registry double would be measuring the double.

The refusal is asserted on `code` + `status` per ADR-0112, never "it threw" — plus a control that a genuinely absent object still answers `OBJECT_NOT_FOUND`, so the two cases stay distinguishable.

### Reverse verification — direction predicted before running, both halves independently

| reverted | result |
|---|---|
| `getMetaItems` object exemption restored | exactly the two `/meta` object rows red; data-plane and both resolution rows green |
| `isObjectPackageDisabled` branch removed from `assertObjectRegistered` | exactly the data-plane row red + the loud-code test; listing rows green |
| `failedCount > 0` branch removed | the 3 failure cases red (200/404 where 400 required); the 3 success cases green |

The success cases cannot distinguish the versions on their own — that asymmetry is intended: they exist to stop the fix over-reaching into "any non-empty `failed` array 400s", which would break the never-published-metadata uninstall.

## Gates

| gate | result |
|---|---|
| `pnpm check:durability-log-level` | ✅ |
| `pnpm check:engine-double-contract` | ✅ 150 pinned |
| `node scripts/check-engine-split-ratio.mjs` | ✅ 100.0% |
| `pnpm check:nul-bytes` | ✅ 7092 files |
| `pnpm check:error-code-casing` | ✅ (new code added) |
| `pnpm build` (closure) | ✅ 71/71 |
| typecheck (spec, objectql, runtime) | ✅ |
| `@objectstack/objectql` | ✅ 3230 |
| `@objectstack/metadata-protocol` | ✅ 1062 |
| `@objectstack/runtime` | ✅ 2022 |
| `@objectstack/spec` | ✅ 9867 |
| `@objectstack/rest` | ✅ 1403 |

Changeset included (user-visible behaviour on both halves).

---
_Generated by [Claude Code](https://claude.ai/code/session_01JiavLQ6DRurPzJ8FuxV1NH)_


os-help:
CI status on `c6b1684` — one real failure fixed, two infrastructure failures that never reached this diff.

**Fixed and pushed (`c6b1684`).** `TypeScript Type Check` was a genuine miss of mine: `check:docs` found `content/docs/references/` out of date, because adding `OBJECT_PACKAGE_DISABLED` to the ADR-0112 ledger changes generated output. Regenerated with `gen:schema && gen:docs`; the whole diff is the new ledger entry plus the `ErrorCode` union count moving `+259 more` → `+260` on every page that renders it. `check:docs` passes locally. Release-owned `content/docs/releases/` untouched.

**Not mine — runner network/TLS, both before any repo code executes.** Not retried yet: GitHub refuses `rerun-failed-jobs` while the workflow is still running (`403 This workflow is already running`). I'll re-run them once the run settles.

- `filter` (run 31484974048) — `dorny/paths-filter` died calling the GitHub API to list changed files: `self-signed certificate; if the root CA is installed locally, try running Node.js with --use-system-ca`. The immediately preceding PR run passed the same job in 11s. Superseded anyway by the push.
- `Dogfood Regression Gate (3/3)` (run 31485193978) — died at the `pnpm --version` step, before install, before build, before any test:

  ```
  ! Corepack is about to download https://registry.npmjs.org/pnpm/-/pnpm-10.31.0.tgz
  AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
    assert(!this.paused)
      at Parser.finish (node:internal/deps/undici/undici:6157:9)
      at TLSSocket.<anonymous> (node:internal/deps/undici/undici:6491:36)
  ```

  A Node/undici internal assertion while corepack fetched pnpm over TLS. Shards 1/3 and 2/3 of the same gate were unaffected.

For the record, the suites this change actually touches were run locally and are green: objectql 3230, metadata-protocol 1062, runtime 2022, spec 9867, rest 1403, plus `pnpm build` 71/71 and the card's gates.

---

os-help:
Ready + queued by the engine-core seat (`session_01PemxJmMZT9JphrsNx886Ce`) at 11:5xZ. 26/26 checks `completed/success` — the `Dogfood Regression Gate (3/3)` red was the known Corepack/undici transient (died in 13s at `pnpm --version`, before the test step; identical signature to #7285's) and cleared on a plain re-run, which is what a transient does. Queue entry verified by the `gh-readonly-queue/main/pr-7700-*` branch, not by the auto-merge field (notes-21: the enable call returns the empty-field form on an all-green PR; a disable/enable flip was needed here as well).

Review verdict and follow-up bookkeeping are on the issue: objectstack-ai/objectstack#7557. On merge, that card goes `pm:blocked` behind **#7705** (the transferred persistence half) rather than closing — hence the `Part of` first line.

---

Issue #7117, opened 2026-08-09
driver-memory's analytics `generateSql()` renders the LIKE family with NO wildcards, so the echoed statement is an EQUALITY the pipeline never ran


os-zhuang:
Found while implementing #6520 (`$icontains` on every JS evaluation face). Out of that card's scope — it is the `$contains` family, not `$icontains`, and it predates that work — so it is filed rather than fixed there.

## Measured on `origin/main` @ `f5a9bc2f3`, by reading the source

`MemoryAnalyticsService` has two exits for one normalized filter tree, and they disagree about what the LIKE family MEANS.

**The mingo exit** (`query()`) builds a real containment pattern — `memory-analytics.ts`, `CUBE_OPERATOR_TO_MONGO_PREDICATE`:

```ts
contains: ({ raw, substring }) => ({ $regex: substring(raw[0]) }),
```

where `substring` is the driver's own `filterSubstringPattern(value)` = `new RegExp(escapeRegex(value), 'i')` — a substring match.

**The SQL exit** (`generateSql()`) emits the comparand as a bare literal, with no `%` anywhere (`memory-analytics.ts`, the WHERE builder):

```ts
const sqlOp = this.operatorToSql(filter.operator);          // 'contains' -> 'LIKE'
const comparand = this.comparandsFor(cube, filter.member, filter.values)[0];
whereClauses.push(`${fieldPath} ${sqlOp} ${this.toSqlLiteral(comparand)}`);
```

`toSqlLiteral` only quotes and escapes quotes; it adds no wildcards. So `{ name: { $contains: 'acme' } }` echoes

```sql
WHERE name LIKE 'acme'
```

which is an **equality** (case-folded on SQLite, exact on Postgres), while `query()` returns every row *containing* `acme`.

`notContains` has the mirror of the same bug via `NOT LIKE`.

## Why this matters

This is the #5333 / #3650 class — "a rendering that contradicts execution is worse than no rendering" — reached through driver-memory's analytics face instead of `service-analytics`' echo. An author who runs the echoed statement to reproduce a chart gets a **narrower** row set than the chart, so the filter reads as broken in the opposite direction from #5333's widening. `service-analytics` has an enumerated regression test for exactly this property on its own three compilers (`objectql-echo-operator-coverage.test.ts`); driver-memory's analytics `generateSql()` has none, which is why this survived.

Note the direction is wrong-rendering, not a permission bypass: this exit produces display SQL, not the executed query.

## Scope

[...]

Issue #6518, opened 2026-08-08
drivers(sql family): 文本算子的大小写折叠是「方言的」而非「契约的」—— `$contains` 在 SQLite 过折叠、`$icontains` 在 PG/MySQL 过折叠


os-zhuang:
Part of #4706(裁决 Q1=A / Q2=A)。从 #5702 实施期实测拆出 —— #5702 交付了 `$regex` 拒收与 `$icontains` 实现,**没有**交付这一条,原因见下「为什么不能在 #5702 里顺手做」。

## 两个方向的同一个缺陷

`SqlDriver.applyLike` 发的 SQL 里,大小写由**方言**决定,不由契约决定。两个算子各在一半方言上跑偏:

| 算子 | 契约(#4706) | SQLite / turso / sqlite-wasm | Postgres | MySQL |
|:--|:--|:--|:--|:--|
| `$contains` / `$notContains` / `$startsWith` / `$endsWith` | 大小写**敏感**(Q2=A) | ❌ `LIKE` 自带 ASCII 折叠 | ✅ 恰好正确 | ⚠️ 随 collation |
| `$icontains` | 折叠 **ASCII only**(Q1=A) | ✅ SQLite `lower()` 只折 ASCII | ❌ `LOWER()` 折全 Unicode | ❌ 同左 |

两格都是**过匹配**(返回了不该返回的行),不是漏匹配。

## 实测(better-sqlite3,`FILTER_TEXT_ROWS` 九行,`origin/main` + PR of #5702)

```
$contains  'acme'  → ["1","2"]     契约要求 ["2"]      (行 1 = "ACME Corp")
$contains  'ACME'  → ["1","2"]     契约要求 ["1"]
$icontains 'acme'  → ["1","2"]     ✅
$icontains 'café'  → ["4"]         ✅(É 不折)
```

顺带一条对分诊有用的读数:**在 SQLite 上 `$contains` 与 `$icontains` 对任意比较数返回完全相同的行** —— 因为 `LIKE` 已经折了 ASCII,外面再套一层 `LOWER()` 是幂等的。也就是说这条不修,`$icontains` 在 SQLite 上是**行为不可观测**的(它的正确性只能靠编译出的 SQL 文本和 PG 见证)。#5702 的 `remote-transport-text-predicates.test.ts` / `sql-driver-icontains-and-retired-operators.test.ts` 已把这两条如实钉住。

## 为什么不能在 #5702 里顺手做(这是本单独立的理由)

[...]

Issue #4706, opened 2026-08-02
`$regex` on driver-sql is not a regex — it compiles to a substring LIKE, so it both over-matches and silently matches nothing


os-zhuang:
Measured on `@objectstack/driver-sql@17.0.0-rc.1` against a real SQLite database while implementing hotcrm#626 (case-insensitive account matching). Filed from the HotCRM reference app.

## What was measured

`$regex` is the only case-insensitive operator ObjectQL offers, so it is what an author reaches for when they need one. On `driver-sql` it is **not executed as a regular expression**. It is compiled to `LIKE '%value%'` — a substring test.

Both failure directions were reproduced on a real SQLite database, with rows `'Acme Corp'` and `'Not Acme Corp Ltd'`:

| query | expected (regex semantics) | actual |
|---|---|---|
| `{ name: { $regex: 'Acme Corp' } }` | `['Acme Corp']` | **`['Acme Corp', 'Not Acme Corp Ltd']`** |
| `{ name: { $regex: '^acme\\s+corp$' } }` | `['Acme Corp']` (case-insensitive) | **`[]`** |

The second row is the dangerous one: a correctly written anchored pattern matches **nothing at all**, with no error. An author who tests their filter against data that happens to contain the substring gets a passing result and ships a query that silently returns the wrong set — or the empty set — in production.

## Why this is worse than the "declared but inert" family

We have filed several issues about metadata that is declared and never enforced (#4649 validations failing open, #4697 flow variables, #4698 the general class, plus hotcrm#650 for inert `decision` conditions). Those are **inert**: the key does nothing.

This one is different and worse: **the operator's name promises one semantics and the implementation delivers another.** Nothing about `$regex` suggests "substring". There is no warning, no validation error, and no runtime signal — the query succeeds and returns a plausible-looking result set. An author has no reason to doubt that an operator called `$regex` is a regex, so this defect is invisible precisely to the people most likely to hit it.

It also affects a second axis the app cares about: the leading `%` wildcard makes the query unindexable, so what looks like an indexed lookup is a full scan.

## What to establish

1. **Decide what `$regex` means and make the implementation and the name agree.** Either implement real regex matching on the SQL drivers (SQLite `REGEXP` needs a registered function; Postgres has `~*`), or — if substring matching is the intended and only supported behaviour — **rename the operator** (`$contains` / `$like`) and reject `$regex` at query-validation time with a message that says what to use instead. A silent semantic substitution is the one option that should not survive.
[...]

PR #7723, opened 2026-08-11
fix(driver-memory): make the `$contains` family case-exact and answer `count_distinct`

### LINKED ISSUE #6682: drivers(memory, mongodb): the `$contains` family still folds case — the last two backends left on the wrong side of #4706 Q2 = A
Finding, filed unassigned from the #6518 implementation. **Not** a request to unfreeze anything — it exists so the two `FILTER_TEXT_CASES` DEBT rows that survive #6518 have a successor to point at, instead of pointing at a closed issue.

## What #6518 left behind

#4706 **Q2 = A** rules the `$contains` family (`$contains` / `$notContains` / `$startsWith` / `$endsWith`) **case-SENSITIVE on every backend**. After #6518 the scoreboard is:

| face | `{ name: { $contains: 'acme' } }` against `ACME Corp` | verdict |
|:--|:--|:--|
| driver-sql / -sqlite-wasm / -turso (local + remote) | does NOT match | ✅ ruled behaviour (#6518) |
| `@objectstack/formula` `matchesFilterCondition` | does NOT match | ✅ already compliant (`String.includes`) |
| objectql `having` | does NOT match | ✅ already compliant |
| service-analytics (both SQL compilers) | does NOT match | ✅ already compliant on the dialect they emit for |
| **driver-memory** — query path + analytics face | **MATCHES** | ❌ `new RegExp(escapeRegex(v), 'i')` |
| **driver-memory** — reference matcher | does NOT match | ⚠️ this package answers ONE operator two ways |
| **driver-mongodb** | **MATCHES** | ❌ `$regex` with a hard-coded `$options: 'i'` |

Both defects are **over-matching** — rows the filter excludes are returned — and both fold the **whole Unicode range**, so they also overshoot the #4706 **Q1 = A** ASCII-only boundary that `$icontains` is held to. On an RLS read scope a wider predicate is over-reach, not a loose filter (#3948).

The driver-memory row is the sharper one: its two faces disagree *with each other*, which is the divergence class #5374 closed between the other two faces of that package.

## Why #6518 did not do it

The **#5499 freeze** (maintainer, 2026-08-05) suspends implementation investment in `driver-memory` and `driver-mongodb`. #6518 was dispatched with that freeze restated as a ruling, so both packages were explicitly out of scope; their `FILTER_TEXT_CASES` cells stay DEBT rather than being cleared or quietly widened.

## Why this is separate from #6520

#6520 is the **`$icontains`** programme (requirement 1: implement a new operator, gated on taking `$icontains` into the spec `FILTER_OPERATORS` allowlist). This is **requirement 2**: an operator both packages already implement returns the wrong rows. Different requirement, different risk — #6520's cost is the vocabulary change rippling through allowlist consumers, while this one is a one-line flag on each side plus the row-set pins that flip with it.

They are, however, natural PR-mates: both touch the same two `translateFieldOperators` / `normalizeFieldOperators` sites, and clearing a `FILTER_TEXT_CASES` cell needs BOTH requirements answered before the case-set can be imported (`scripts/check-driver-conformance.mjs` judges a cell by that import, so a half-answered cell must not import it).

## Suggested scope

1. `driver-memory`: drop the `'i'` flag from the `$contains`-family lowering on the query path and the analytics face, and align the reference matcher — one answer per operator, per #5374.
2. `driver-mongodb`: drop the hard-coded `$options: 'i'` in `mongodb-filter.ts`'s `translateFieldOperators`. `escapeRegex` already makes the comparand literal, so the literal-comparand cases keep holding.
3. Flip the row-set pins that currently encode the folded answer, asserting the NEW substance rather than deleting the assertion.
4. Re-grade the two `FILTER_TEXT_CASES` DEBT rows in `scripts/check-driver-conformance.mjs` — they can only be deleted once requirement 1 (#6520) is answered too, since coverage is judged by importing the whole case-set.

Anchors verified on `origin/main` @ `d13f627`: `driver-memory/src/memory-driver.ts` (`normalizeFieldOperators`, the RegExp-`'i'` query path), `driver-memory/src/memory-matcher.ts` (`match()`, `String.includes`), `driver-mongodb/src/mongodb-filter.ts` (`translateFieldOperators`).

### LINKED ISSUE #6814: [drivers] driver-memory / driver-mongodb disagree with the SQL family on `count_distinct` — the two AGGREGATION_CASES cells #6409 left open

huangyiirene:
Fixes #6814
Fixes #6682

Both cards' `driver-mongodb` halves landed earlier today (PR #7550, PR #7625). These are the `driver-memory` halves — the last open scope of each — so merging this closes both. Authorized by the maintainer's 2026-08-11 full lift of the #5499 freeze (#5499 comment `5252526378`).

After this, **`scripts/check-driver-conformance.mjs` reports `40 covered cell(s), 0 in the DEBT ledger, 0 exempt`** — the ledger is empty for the first time since the gate was written.

## Premise verified by EXECUTION before the first edit

Both suites were written first and run against unmodified `origin/main` @ `21888ab`:

| suite | pre-fix | post-fix |
|:--|:--|:--|
| `memory-filter-text-conformance.test.ts` | **9 failed / 43 passed** (52) | 52/52 |
| `memory-aggregation-conformance.test.ts` | **11 failed / 23 passed** (34) | 34/34 |

Every pre-fix failure in the text suite was a `$contains`-family row (`expected ['1','2'] to deeply equal ['2']`). Every `$icontains` row, every literal-comparand row, every refusal row and every reference-matcher row passed **before** the fix — which is what says the flag was the whole gap and that nothing was widened to reach green.

## #6682 — the `$contains` family, nine sites

The card framed this as "the query path + the analytics face". Measured, it is **nine `'i'`-flagged RegExp sites in three clusters**, and the third cluster is why the analytics face was wrong:

- `convertConditionToMongo` ~`:808-830` — the **AST spelling**'s four arms (`contains` / `notcontains` / `startswith` / `endswith`).
- `normalizeFieldOperators` ~`:985-994` — the **`$`-spelling**'s four arms. The dispatch flagged this cluster as possibly an `$in`-like path outside the ruling's reach; measured, it is the same four operators reached through the `FilterCondition` door, squarely inside #4706 Q2 = A. Nothing was widened beyond the family the ruling names.
- `filterSubstringPattern` `:1353` — the **#5374 shared rule**. The analytics face does not carry a fold of its own; it *borrows* this one. So one line fixed that face, which is the property #5374 bought, working in the direction it was built for.

`escapeRegex` is untouched (the comparand was always literal — that half was never the defect), and `$icontains` is untouched: its ASCII fold lives in the pattern *source* (`asciiCaseInsensitiveRegexSource`), not in a flag, so taking the flag off the family cannot move the boundary. A dedicated row asserts that.

## #6814 — `count_distinct`, and a premise the card got wrong

`MemoryDriver.computeAggregate` had arms for count/sum/avg/min/max and then `default: return null`. No `count_distinct` arm, so a function the Query Protocol declares and every SQL face lowers (#6409) resolved with `{ n: null }` — no error, no log, no refusal. It now returns `new Set(values.filter(v => v != null)).size`, the same expression `in-memory-aggregation.ts` uses.

**The card, the ledger row and `aggregation-conformance.ts`'s DEBT table all read the ANALYTICS face as the one that "DOES implement `count_distinct`". Executing the case-set falsified that.** `buildAggregator` emitted `{ $addToSet }` under a comment reading *"Will need post-processing for count"* — and no post-processing existed. So the measure answered the raw **array** of values, `['won','lost',null]`, under a field `measureTypeToFieldType` describes as `number`.

One declared function, three answers: `null` on the data face, an array on the analytics face, and the standard's number nowhere. Both are fixed here, the array sized excluding null — the same exclusion the driver-mongodb half made on `$addToSet` (#7550).

## Multi-face declaration (per face of THIS package)

| face | `$contains` family | `count_distinct` |
|:--|:--|:--|
| **query path** (`memory-driver.ts`, both spellings, both doors of `find()`/`aggregate(AST)`) | **fixed here** — 8 sites | **fixed here** — the missing arm |
| **analytics face** (`memory-analytics.ts`) | **fixed here** — via `filterSubstringPattern`, the shared rule it borrows | **fixed here** — `$addToSet` sized, null-excluded |
| **reference matcher** (`memory-matcher.ts`) | **already compliant** — `String.includes`, case-exact all along; verified by 13 rows in the new suite, file untouched | n/a — not an aggregate face |

The cross-backend scoreboard is settled and not re-proved here (#6518 / #7625).

## Out of scope, with reason

**#7117 (`memory-analytics.ts` `generateSql` / `operatorToSql`).** The sweep did pass through it: `contains` renders as `LIKE`, which folds ASCII on a SQLite-shaped consumer. It is **not fixed here**, and this is a judgement rather than an omission — making it case-exact means choosing `GLOB` or a binary cast per dialect, and the pattern that construct takes is exactly the wildcard rendering #7117 owns. A half-change would emit incoherent SQL. This exit is also descriptive (it returns SQL text, it does not execute it), so no row set moves. Reported as adjacency; #7117 is deliberately serialized behind this card.

**`memory-analytics.ts`'s cube `count` measure** emits `{ $sum: 1 }` regardless of `sql`, i.e. `count(*)` semantics — the shape whose driver-side twin #7550 fixed on mongodb. Not touched: `measureToSql` emits `COUNT(*)` for the same measure, so that face is self-consistent, and cube `count` semantics are a different question from `AGGREGATION_CASES`' `count(col)`. Recorded so it is not rediscovered as a defect.

## Pins flipped, never deleted

Five assertions encoded the fold. All five flipped to the ruled substance:

- `memory-driver.test.ts` — `should filter with $contains case-insensitively` → `case-SENSITIVELY` (asserts 0 rows for `'alice'`, 1 for `'Alice'`), **plus a new sibling** pinning `$icontains` so the pair still covers both behaviours rather than losing one.
- `memory-driver.test.ts` — `$contains inside $and`: the subject is *combination*, so the comparand moved rather than the expectation (`'a'` matched `Alice Johnson` only through the fold). `'o'` is case-exact for both rows and makes the conjunction do visible work.
- `memory-driver-filter-logic-conformance.test.ts` — the three `#5374` case rows now assert the family is case-SENSITIVE *on both faces*, **plus a positive control** so they cannot be passed by a predicate that stopped matching anything.

## Reverse verification — direction predicted before it was run

Recorded in the aggregation suite's docblock, both reverts executed:

- **(A) the `count_distinct` arm removed** — predicted failures on the VALUE `null`, never a throw, arithmetic cases green. Measured **7 failed / 27 passed**, exactly that.
- **(B) the arm present but `new Set(values).size`** — null not excluded. Predicted 3-for-2 ungrouped and `east` 2 / `west` 3 grouped, with `count_distinct(score)` staying GREEN at 6. Measured **6 failed / 28 passed**, exactly that. (B) is unreachable by a suite that only tests one column, which is why the table carries both.

## Prose this PR makes false, corrected

Four files carried claims the diff falsifies. All prose/docblock only — zero schema or behaviour lines:

- `packages/spec/src/data/filter-text-conformance.ts` — the status list and one case note.
- `packages/spec/src/data/filter.zod.ts` — the `$contains`-family status paragraph and the DEBT-row sentence.
- `packages/spec/src/data/aggregation-conformance.ts` — the DEBT table (driver-memory enrolled; the "analytics face agrees" row corrected to what execution showed) and the enrolled-backends list.
- `content/docs/protocol/objectql/query-syntax.mdx` — two `warn` callouts saying one driver still folds, now `info` and naming the row-set change for anyone who relied on it.

## Gates

| gate | result |
|:--|:--|
| `check:driver-conformance` | **OK — 40 covered, 0 DEBT, 0 exempt** |
| `check:driver-conformance --self-test` | OK |
| driver-memory suite | 664 passed (22 files) |
| `@objectstack/spec` suite | 9883 passed (377 files) |
| `@objectstack/objectql` suite | 3228 passed (182 files) |
| `@objectstack/runtime` suite | 2018 passed (127 files) |
| service-datasource / http-conformance / verify | 326 / 72 / 23 passed |
| `tsc --noEmit` (driver-memory, spec) | clean |
| build (driver-memory dependency closure) | clean |

Downstream selection was measured, not assumed: `pnpm --filter '...@objectstack/driver-memory'` lists 16 dependents; `objectql` is **not** among them (it depends the other way), so its suite was run on its own account and is green.

Changeset: `.changeset/memory-contains-case-and-count-distinct.md`, **patch** for `@objectstack/driver-memory` — a user-visible row-set change on the `$contains` family plus `count_distinct` now answering, telling anyone relying on the fold to write `$icontains`.

No `content/docs/releases/` edit.

https://claude.ai/code/session_01NuzQE844ut8m8vypXcdBYD

---
_Generated by [Claude Code](https://claude.ai/code/session_01NuzQE844ut8m8vypXcdBYD)_

### DIFF
--- a/.changeset/memory-contains-case-and-count-distinct.md
+++ b/.changeset/memory-contains-case-and-count-distinct.md
@@ -1,0 +1,35 @@
+---
+'@objectstack/driver-memory': patch
+---
+
+driver-memory: the `$contains` family is case-SENSITIVE, and `count_distinct` answers a number
+
+Two user-visible answers change on the in-memory driver. Both bring it onto the
+answer the SQL family, MongoDB and the protocol already give, so a filter or an
+aggregate now means the same thing whether your tests run on this double or your
+production runs a real database.
+
+**`$contains` / `$notContains` / `$startsWith` / `$endsWith` no longer fold
+case.** They matched with a case-insensitive regex on the query path and on the
+analytics face — over the whole Unicode range, wider even than the ASCII
+boundary `$icontains` is held to — so `{ name: { $contains: 'acme' } }` returned
+`ACME Corp` here and did not on any other backend. This driver's reference
+matcher (`match()`) was already case-exact, so the two folding faces have moved
+onto the answer the third one always gave. The comparand stays literal: `%`,
+`_` and `.` were never wildcards here and still are not.
+
+**This is a ROW-SET change.** If you relied on the fold, write `$icontains` —
+the operator that spells it, implemented on every backend since #6520 and
+folding ASCII case only.
+
+**`count_distinct` answers.** `MemoryDriver.computeAggregate` had no arm for it,
+so an aggregation the Query Protocol declares resolved with `{ alias: null }` —
+no error, no log, no refusal. It now counts distinct NON-NULL values, matching
+`COUNT(DISTINCT col)`. The analytics face was wrong in its own way and is fixed
+beside it: it collected the distinct values and never sized them, so a
+`count_distinct` measure came back as the raw array of values under a field its
+own response metadata types as `number`.
+
+Both are held to `@objectstack/spec/data`'s shared case-sets from now on
+(`FILTER_TEXT_CASES`, `AGGREGATION_CASES`), executed in process against every
+face of the package.
--- a/content/docs/protocol/objectql/query-syntax.mdx
+++ b/content/docs/protocol/objectql/query-syntax.mdx
@@ -337,15 +337,16 @@ metacharacters — `{ name: { $icontains: 'a.b' } }` matches `a.b` and not `axb`
   compilers). A filter using it means the same thing whether your tests run on the
   in-memory double or your production runs SQL.
 
-  One half of the case rules above is still landing: `$contains` /
+  The other half of the case rules above has landed too: `$contains` /
   `$startsWith` / `$endsWith` / `$notContains` are case-**sensitive** by ruling and
-  are so on the SQL family and on MongoDB —
+  are so on every backend —
   [#6682](https://github.com/objectstack-ai/objectstack/issues/6682) removed the
-  hardcoded `$options: 'i'` that had folded them there. The in-memory driver's
-  query and analytics faces still fold over the whole Unicode range. Until that
-  half lands, prefer `$icontains` when you *want* a fold rather than relying on
-  `$contains` being loose on that backend. The shared standard both halves are
-  measured against is `FILTER_TEXT_CASES` (`@objectstack/spec/data`).
+  hardcoded `$options: 'i'` that folded them on MongoDB and the case-insensitive
+  regex that folded them on the in-memory driver's query and analytics faces.
+  **If you were relying on `$contains` being loose on the in-memory driver, that
+  is a row-set change: write `$icontains` when you want a fold.** The shared
+  standard both halves are measured against is `FILTER_TEXT_CASES`
+  (`@objectstack/spec/data`), which all five drivers now run.
 </Callout>
 
 ### `$regex` — removed
@@ -963,19 +964,21 @@ rule in [Case Sensitivity](#case-sensitivity) above. Note what that means for se
 a user typing `acme` does not find `ACME Corp`. Only `select` / `status` option
 *labels* are matched case-insensitively by the expansion itself.
 
-<Callout type="warn">
-  **Measured today: one driver still does not match that rule.** The `$contains`
-  alignment landed in two steps —
+<Callout type="info">
+  **Measured today: every driver matches that rule.** The `$contains` alignment
+  landed in three steps —
   [#6518](https://github.com/objectstack-ai/objectstack/issues/6518) made `SqlDriver`
   case-exact per dialect (`GLOB` on the SQLite dialects, `LIKE` unchanged on
   Postgres, `LIKE` over a binary cast on MySQL), and
   [#6682](https://github.com/objectstack-ai/objectstack/issues/6682) removed
-  `driver-mongodb`'s hardcoded `$options: 'i'`. `driver-memory`'s query path still
-  matches with a case-insensitive regex, so running your tests on the in-memory
-  double can still return rows a SQL or MongoDB deployment would not. Whether the
-  expansion should emit `$icontains` instead of `$contains` — i.e. whether search is
-  case-insensitive by definition — is a separate question that rides with that issue,
-  because it can only be answered once both operators mean one thing everywhere.
+  `driver-mongodb`'s hardcoded `$options: 'i'` and then the case-insensitive regex
+  `driver-memory` used on its query and analytics faces. So running your tests on the
+  in-memory double no longer returns rows a SQL or MongoDB deployment would not —
+  the divergence this callout warned about is closed, and
+  `FILTER_TEXT_CASES` holds all five drivers to it. Whether the expansion should emit
+  `$icontains` instead of `$contains` — i.e. whether search is case-insensitive by
+  definition — remains a separate open question, and one that can now actually be
+  answered, since both operators mean one thing everywhere.
 </Callout>
 `fuzzy`, `boost`, `operator`, `minScore`, `language`, and `highlight` carry
 `[EXPERIMENTAL — not enforced]` markers (#4286): the schema accepts them, the
--- a/packages/drivers/driver-memory/src/memory-aggregation-conformance.test.ts
+++ b/packages/drivers/driver-memory/src/memory-aggregation-conformance.test.ts
@@ -1,0 +1,252 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#6814] Aggregate-vocabulary conformance for `driver-memory` — the shared
+ * `@objectstack/spec/data` cases, on rows, executed in process.
+ *
+ * The SQL twins (`sql-driver-aggregation-conformance.test.ts`,
+ * `turso-remote-aggregation-conformance.test.ts`,
+ * `sqlite-wasm-aggregation-conformance.test.ts`) run this same table against a
+ * real database. This file is the reason the cases live in the spec package
+ * rather than beside one driver: an aggregate that answers differently here
+ * than under SQL pushdown is one query with two numbers, decided by a driver
+ * capability bit the caller never sees.
+ *
+ * ## Why nothing here is a "modelled" evaluation
+ *
+ * This driver runs in process, so every case below is a REAL execution — no
+ * server-free half in the shape `driver-mongodb` needs (#5517), and no emitted-
+ * string assertion standing in for a number. That matters for the defect this
+ * file was written against: `computeAggregate` had no `count_distinct` arm at
+ * all, so the function fell to `default: return null` and `aggregate()` resolved
+ * with `{ n: null }` — no error, no log, no refusal. Only executing the case
+ * says so; a lowering-shape assertion has nothing to look at.
+ *
+ * ## Both doors of the data face, and the analytics face beside it
+ *
+ * `find()` and `aggregate(AST)` are two entries to the same
+ * `performAggregation`, and objectql's engine uses the second one. Both are
+ * driven, because "the aggregate works" measured through one door is what let
+ * this package answer one declared function two ways for as long as it did
+ * (#5374). The analytics face (`memory-analytics.ts`) is driven in the last
+ * block for the same reason — it implements `count_distinct` independently, so
+ * it is a third answer unless something demands they agree.
+ *
+ * ## Reverse verification — direction predicted BEFORE it was run
+ *
+ * **(A) the `count_distinct` arm removed** (the pre-#6814 state). Predicted:
+ * the three `count_distinct` cases fail on `null` — the value, not a throw —
+ * while every arithmetic case stays green, because the missing arm is a silent
+ * fall-through rather than a broken computation.
+ *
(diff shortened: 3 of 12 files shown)


huangyiirene:
**Queue-kick triage (drivers seat) — not this PR's failure; do NOT blind-requeue.**

Per the triage checklist above: branch **2** ½ — the failing test (`api-methods-batch-conformance` › `grants bulk wherever it grants create / update / delete`) is a `packages/spec` FILE-SCAN test with zero causal path from this PR's diff, and it is red on **main itself**: PR **#7769** (`52200b4`, merged 14:30Z) opened `update` on `sys_api_key` without `bulk` and without the `SINGLE_RECORD_WRITE_ONLY` exemption the test's own header prescribes. PR-side affected-subset CI never runs this file-scan for a platform-objects diff — only full-suite queue builds do, so every queued PR inherits the red.

Stopgap filed and dispatched: **#7793** (route B — record the exemption with #7769's own recorded reason; one entry, one file). **This PR will be re-queued after #7793 merges** — re-queuing before that burns a full queue rebuild for everyone behind it.

Queue-steward note (双向让行): the drivers seat is handling this kick end-to-end — stopgap #7793 in flight, re-queue of this PR follows its merge. No re-submit needed from the steward side.

---

huangyiirene:
Base fixed — the `api-methods-batch-conformance` red that kicked this PR from the queue is resolved on `main` (the `sys_api_key` exemption landed via #7818, racing our stopgap #7803, which is now closed as duplicate). Re-queuing this PR.

---

huangyiirene:
Queue-failure triage (run 31527103814): the failing step is Test Core's **Checkout repository** — `fatal: unable to access 'https://github.com/…': server certificate verification failed. CAfile: none CRLfile: none` on the runner, before a single test ran. Runner-side TLS infrastructure flake, unrelated to this PR's content (the earlier 24h failure was the api-methods base poisoning, fixed by #7818 — different signature, also external). Re-queuing once.

---

Issue #7634, opened 2026-08-11
i18n: `GET /i18n/locales` declares a locale `label` as "Display name" but every producer sets it to the code


yinlianghui:
Found while implementing objectstack-ai/objectui#4039 (the console language menu now reads this endpoint). Filing only; not fixed there — out of that issue's scope, and the consumer side deliberately routes around it.

### What

`GetLocalesResponseSchema` (`packages/spec/src/api/protocol.zod.ts`) declares each locale descriptor as:

```ts
locales: z.array(z.object({
  code: z.string().describe('BCP-47 locale code (e.g., en-US, zh-CN)'),
  label: z.string().describe('Display name of the locale'),
  isDefault: z.boolean().default(false).describe('Whether this is the default locale'),
}))
```

The only producer of those descriptors is `toLocaleDescriptors` (`packages/spec/src/system/i18n-resolver.ts`):

```ts
return codes.map((code) => ({ code, label: code, isDefault: code === defaultLocale }));
```

`label` is the code echoed back. Both serving surfaces go through that one helper — the runtime dispatcher's `/i18n` domain and `service-i18n`'s autonomous route — deliberately, and correctly, so there is no second implementation that behaves differently. The field is therefore never a display name on any surface: `GET /api/v1/i18n/locales` answers `{ code: 'th', label: 'th' }`, never `{ code: 'th', label: 'ไทย' }`.

### Why it is worth writing down

`label` reads as a display name in the schema, in the generated SDK type, and in the OpenAPI surface. A client that trusts the description renders locale codes to users and only finds out by looking. objectui#4039 hit exactly this and had to make the opposite decision explicitly: the console's language menu names locales from its own built-in metadata plus `Intl.DisplayNames` and ignores the endpoint's `label`, because using it would have put `th` in the menu where `ไทย` belongs.

This is the declared-but-not-enforced shape ADR-0049 is about, one field wide: the vocabulary exists, no implementation stands behind it.

### Observation-class, not a live defect

Nothing a user hits today, as far as I can tell — there is no current consumer of the field. The one client that reached for it (the console) now does not. Recording it so triage can grade it rather than leaving it in a PR comment.

### Options, not a decision

- **Retire the field** — drop `label` from the descriptor and the schema; `code` is the whole payload, and naming a locale is the client's job anyway (that is where `Intl.DisplayNames` and any built-in native names live). Smallest surface, and consistent with what the one real consumer chose.
[...]

Issue #3778, opened 2026-07-28
i18n: two unbridged bundle shapes — file-authored `objects.<name>` vs the runtime `translation` metadata type's `o.<object>`, so translations cannot be authored in the product


os-zhuang:
Surfaced while closing #3370 and its follow-ups. Filing rather than widening those PRs — this is a contract split, not a fill.

## The split

Two translation shapes exist, and nothing bridges them.

**File-authored bundles** use `objects.<name>.…` — object / field / option / view / `_actions` keys, exactly what `os i18n extract` scaffolds and what `packages/spec/src/system/i18n-resolver.ts` reads at request time. Every shipped bundle is this shape: `platform-objects/src/apps/translations/*`, the plugin/service bundles, and any customer bundle listed in `translations: [...]`. The objectui client hooks (`useObjectLabel`) read the same shape.

**The `translation` metadata type** — the one an admin can author at runtime through the metadata door (`allowRuntimeCreate: true`) — is registered against `AppTranslationBundleSchema` (`packages/spec/src/kernel/metadata-type-schemas.ts`), the object-first `o.<object>` shape from `packages/spec/src/system/translation.zod.ts`. No resolver reads that shape, so translations authored in the product never render.

## Root-cause finding (2026-07-28 investigation)

The split is narrower than "unbridged" suggests — **the plumbing already works; only the registered authoring contract points at the wrong shape**:

- The #2591 authored-translation sync (`packages/core/src/fallbacks/authored-translation-sync.ts`) is shape-agnostic: it resolves the locale (`_meta.locale` → top-level `locale` → BCP-47-looking item name), strips bookkeeping keys, and deep-merges the payload verbatim into the i18n authored layer, which wins over static bundles on read.
- Save-time validation (`metadata-protocol/src/protocol.ts`) `safeParse`s but **stores the original payload verbatim**, and `AppTranslationBundleSchema` has no required fields — so a file-shape (`objects.`) payload already passes the door today and renders end-to-end. Nobody knows, because the schema, the Studio editor form derived from it, and the protocol doc all present the `o.` shape.
[...]

PR #7763, opened 2026-08-11
feat(spec): give TranslationBundle a `flows` surface for screen-flow wizard copy (#7646)

### LINKED ISSUE #7646: i18n: TranslationBundle has no surface for screen-flow UI — wizard screens render source-language only, even in a fully localized app
## Context

HotCRM (main `333259a4`, @objectstack 17.0.0-rc.5) completed all four locales and retired its i18n exemption ledger (hotcrm#645/#679). Yet its lead-conversion wizard (`lead_conversion`, `type: 'screen'`) renders entirely in English on a zh-CN console:

- screen title: **Conversion Details**
- field labels: **Create Opportunity? / Opportunity Name / Opportunity Amount**
- runner buttons: **Cancel / Submit**

## Root cause

The TranslationBundle schema (`@objectstack/spec` `translation.zod`) exposes surfaces for **objects, apps, dashboards, actions** — there is **no `flows` / screen surface at all**. An app author has no place to put translations for screen titles or screen field labels: this is not a missing key, it is a missing vocabulary.

```
$ grep -oE '(flows?|screens?|apps|objects|views|dashboards|actions)[?:]' translation.zod-*.d.ts | sort | uniq -c
  6 actions:   4 apps:   3 dashboards:   4 objects:   # no flows, no screens
```

## Expected

A `flows` surface in the bundle (per-flow: label, per-screen: title, per-field: label/placeholder/help), applied by the screen-flow runner the same way object/field labels already are. The runner chrome (Cancel/Submit) presumably belongs to the console's own message catalog instead.

## Repro

zh-CN console → HotCRM → any lead → 转化线索 → wizard is English top to bottom.

Filed from HotCRM main; happy to wire the HotCRM-side bundle entries as soon as the surface exists.

os-zhuang:
Fixes #7646

A `type: 'screen'` flow is a wizard the user reads — a heading, a list of labelled inputs — and the translation bundle had no group for any of it. Not a drifted key: **no key**. The bundle's surfaces were objects, apps, dashboards, pages, settings and metadata forms, so a translator had nowhere to put a screen title or a screen field label, and the strict shapes (correctly) refused whatever group they invented. HotCRM finished all four locales and retired its i18n exemption ledger, and its `lead_conversion` wizard still rendered "Conversion Details / Create Opportunity? / Opportunity Name" in English on a zh-CN console.

## The ruling this implements

Recorded on #7646 (2026-08-11), quoted verbatim from the issue comment:

> **Ruling: direction accepted.** TranslationBundle gains a `flows` surface — per-flow label; per-screen title; per-field label/placeholder/help — applied by the screen-flow runner the same way object/field labels already are. Runner chrome (Cancel/Submit) stays in the console's own message catalog, NOT the bundle. Contract-first split: this card is the spec surface; the objectui runner-application card and the HotCRM wiring are downstream, filed by their consuming seats once this lands.

## The new group

Added to the shared `translationDataShape()`, so it lands on **both** doors by construction — the file-authored bundle (`TranslationDataSchema`) and the `translation` metadata item (`TranslationItemSchema`). That shared-shape placement is the thing #3778's item-only guard got wrong and #4001 fixed; a group added to one door only is the asymmetry this file keeps paying to remove.

```
flows.< flow_name >.label
flows.< flow_name >.screens.< node_id >.title
flows.< flow_name >.screens.< node_id >.fields.< field_name >.label
flows.< flow_name >.screens.< node_id >.fields.< field_name >.placeholder
```

## Key addressing — measured, with evidence

The PM assumption to falsify here was "which identifiers does the runner actually resolve against". Each level's key is an identifier a consumer already holds at render time, not a new naming scheme:

| level | key | declared at | reaches the client as |
|:---|:---|:---|:---|
| flow | `Flow.name` (machine name, `^[a-z_][a-z0-9_]*$`) | `packages/spec/src/automation/flow.zod.ts:614` | the console launches the run by name |
| screen | `FlowNode.id` | `packages/spec/src/automation/flow.zod.ts:263` | `ScreenSpec.nodeId`, `packages/spec/src/contracts/automation-service.ts:138` |
| field | `ScreenFieldConfig.name` | `packages/spec/src/automation/builtin-node-config.zod.ts:382` | `ScreenFieldSpec.name`, `packages/spec/src/contracts/automation-service.ts:105` |

`nodeId` rather than a screen index or the node label: it is what correlates a resume back to its pause point, so it is the one screen identifier guaranteed stable and present client-side. An index renumbers whenever a step is inserted, and the node `label` is itself display copy. A translation surface addressed by keys nothing produces is the declared-but-unresolvable trap this lane exists to close, so the addressing is pinned by a test that builds a real `FlowSchema` with a screen node and derives every bundle key from the parsed flow.

Screen `title` overlays `config.title`; a screen that declares none shows its node `label` (`ScreenConfigSchema.title` — "falls back to the node label"), so this one key covers whichever of the two the runner ends up drawing. Same one-string-one-spelling rule `pages.< name >.title` follows over `label`.

## Per-field key face — measured against the screen-field schema

The ruling's list was label / placeholder / help. Two of the three are real; the third is not, and is refused rather than declared:

- **`label`, `placeholder` — declared.** `ScreenFieldConfigSchema` declares both.
- **`help` — NOT declared.** That schema is `name` / `label` / `type` / `required` / `options` / `defaultValue` / `placeholder` / `visibleWhen`, with nothing help-shaped at all. Declaring `help` would parse clean and translate nothing — the ADR-0078 shape #6080 deliberately kept out of the page-component face for exactly this reason. Unlike #6080's `help`, there is no honest key to alias it to either: `placeholder` is the in-input hint, not help text, so re-pointing it there would translate the wrong string. It rides the unknown-key `guidance` instead, together with `helpText` / `hint` / `tooltip` / `description`, and the rejection says the field has no such copy.
- **`options` — also guidance, not a key.** `ScreenFieldConfig.options[].value` is `z.unknown()` (numbers and booleans are legal), so the value-keyed map `objects.< object >.fields.< field >.options` uses cannot address them unambiguously.

## Runner chrome is not here

Cancel/Submit are the console's own words in every app; putting them in a per-app bundle would ask every app to re-translate the platform. No key for them on any of the three levels, and a test pins that.

## Family conventions mirrored

- Every key optional, at every level — the family's partial-locale semantics, for the reason `ObjectTranslationDataSchema.label` states: partial translation is the normal state, and requiring completeness fills bundles with restated source strings that mask real coverage gaps. One locale of a bundle may carry `flows` while another does not.
- `strictObject` at all three levels with `surface` / `history` / `aliases` / `guidance`, so an unknown key is refused by name instead of stripped.
- The `label`/`title` trap alias one level down (a flow's headline is `label`, a screen's is `title`) — the same table `dashboards.widgets` and `pages.components` carry.
- Top level: `flow` → `flows` is offered as the rename it is; `screens` is **guidance**, not an alias, because screen copy nests under the flow that owns it — the content moves, it is not re-spelled. That is the `app`/`apps` distinction `TRANSLATION_KEY_GUIDANCE` was built to draw.

## Liveness — `planned`, not `live`

No shipped screen-flow runner reads this group yet, so the new `flows` row in `packages/spec/liveness/translation.json` is `planned` with `authorWarn` + `authorHint`, drilled one level over `label` / `screens`. Precedent: `app.navigation.runAction`, the contract-first half of the SDUI deep-link promotion, which is `planned` for the same reason and flips to `live` when its objectui consumer lands. `planned` rather than `dead` is load-bearing — `dead` means declared with no consumer *and no plan*, and this group was ruled into the vocabulary specifically so the runner half could be built against it.

## Scope boundary held

The resolver-side helper — a `FLOW_SCREEN_COPY_KEYS` sibling of `PAGE_COMPONENT_COPY_KEYS`, which lives in `packages/spec/src/system/i18n-resolver.ts` — is deliberately **not** in this change: #7634 is in flight on that file, and this card is the schema surface. Recorded as a follow-up rather than collided with. Likewise `packages/lint/src/validate-translation-references.ts` (which skips unrecognised top-level namespaces, so `flows` is inert there rather than broken) and the CLI extract/coverage utilities.

## Changeset

`@objectstack/spec`: **minor**. The accepted authoring surface widens — a bundle previously rejected for carrying `flows` now parses — which is an additive vocabulary change, not a fix to existing behaviour and not a break.

## Verification

Reverse-verified via a patch file (never `git stash`): predicted direction **plain red**, since the surface is new and its pins address it directly. Removing the schema hunk turned 11 of the block's tests red — the fully-populated parse, the partial-locale parse, the addressing pin, the alias/guidance pins, the JSON-Schema describe pin and the item-door pin — and restoring it returned 110/110.

```
pnpm --filter @objectstack/spec test          → 378 files / 9947 tests passed
pnpm --filter @objectstack/spec typecheck     → clean (test layer: pre-existing shrink-only debt unchanged)
pnpm --filter @objectstack/spec check:liveness → translation 21 classified (live 19, planned 2); every container inheritance declared
pnpm --filter @objectstack/spec check:generated → All 13 generated artifacts are up to date
pnpm --filter @objectstack/spec check:authorable-surface → 1590 schemas, green
pnpm check:merge-driver / check:adr-anchors / check:spec-parsed-alias / check:nul-bytes → all exit 0
pnpm check:i18n → OK (9 packages, all bundles in sync, no undeclared authoring keys)
```

Generated closure committed: `authorable-surface/system.json`, `liveness/state-counts.md`, `content/docs/references/system/translation.mdx`, the strictness-ledger counts.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HXyTbnqdXNR7cccv5jioAs)_


os-zhuang (commented):

Direction matches the maintainer-accepted ruling on #7646 (2026-08-11, seat #6017 triage sweep): the bundle gains a `flows` group, and runner chrome (Cancel/Submit) stays in the console's own message catalog rather than per-app bundles.

What carried the review:
- **Addressing is the runner's own** — `Flow.name` / `FlowNode.id` (client's `ScreenSpec.nodeId`) / `ScreenFieldConfig.name`, each verified against the declaring schema rather than mirrored from the report. The declared-but-unresolvable trap is tested, not just argued.
- **`help` correctly excluded**: `ScreenFieldConfigSchema` has nothing help-shaped, so it rides `guidance` instead of becoming a slot that validates and never renders — consistent with #6080/ADR-0078. The `options` exclusion (unconstrained `options[].value` can't anchor a value-keyed map) is likewise measured.
- **Both doors** get the group by construction (shared `translationDataShape`), and the metadata-item door is pinned in tests (#3778's lesson).
- **Liveness honesty**: `planned` + `authorWarn` rather than `live` — with per-child rows because the two halves resolve through different consumers and can flip on different days.
- **Hot-file discipline**: deliberately leaving the `FLOW_SCREEN_COPY_KEYS` resolver helper out because #7634 was in flight on `i18n-resolver.ts` was the right call; that helper travels with the objectui runner card when the group flips to `live`.

Landing note (PM-handled, no dev action): this PR regenerates `content/docs/references/**`, an os-regen routed tree, so it lands via the one-at-a-time relay — it's queued behind #7758 and #7759. I'll run its sync lap and flip when its slot arrives.

---

Issue #7596, opened 2026-08-11
[spec] `FieldReferenceSchema` is declared in the `$between` endpoints but NO backend resolves a `$field` inside a list


os-zhuang:
Measured while implementing #5222 (cross-field push-down). Filing unassigned — recording, not claiming.

## The fact

`packages/spec/src/data/filter.zod.ts` declares `FieldReferenceSchema` in **both** `$between` endpoints, in the documentation copy and in the ENFORCED copy:

```ts
$between: z.tuple([
  z.union([z.number(), z.date(), z.string(), FieldReferenceSchema]),
  z.union([z.number(), z.date(), z.string(), FieldReferenceSchema])
]).optional(),
```

`$in` / `$nin` are `z.array(z.any())`, so a reference is admitted there too.

**No backend resolves it.** The in-memory evaluator — the one implementation the repo has for `$field` — does not, and this is structural rather than an oversight. `matches-filter.ts` `resolveValue` returns early for an array:

```ts
function resolveValue(raw: unknown, record) {
  if (raw && typeof raw === 'object' && !Array.isArray(raw) && '$field' in raw) { … }
  return raw;   // ← an ARRAY comes back unchanged, members unresolved
}
```

`evalOp` resolves the **whole comparand**, never the members. So for `$in`/`$nin` the raw reference OBJECT is compared with `looseEq` against a stored value (never equal), and for `$between` the raw object becomes an ordering bound.

## Why it matters

This is the ADR-0049 enforce-or-remove shape at a declared position: the spec promises a comparand form that nothing implements, and the failure is SILENT on the memory path — `{ amount: { $in: [{ $field: 'budget' }] } }` matches nothing and reports nothing. On an RLS `check` that is a denied write with no diagnostic; the `$nin` direction loses an exclusion the author wrote.

`driver-sql` / `driver-sqlite-wasm` refuse these positions loudly (`INVALID_FILTER` / 400) — #5041 installed that and #5222 deliberately kept it, precisely because there is no correct in-memory semantics for SQL to be equivalent to. So today the declaration is honoured by nobody and refused by two backends.

## The decision this needs

Either:
- **Remove** `FieldReferenceSchema` from the `$between` endpoint unions (and rule `$in`/`$nin` members out explicitly), making declared = enforced; or
[...]

PR #7772, opened 2026-08-11
fix(spec): NormalizedFilterSchema judges its `$and`/`$or`/`$not` members instead of catching them all (#7711)

### LINKED ISSUE #7711: [spec] `NormalizedFilterSchema` accepts ANY field-condition shape — its union's second branch is a non-strict catch-all
Measured while implementing #7596 (removing `FieldReferenceSchema` from the `$between` endpoints). Filing unassigned — recording, not claiming.

## The fact

`NormalizedFilterSchema` (`packages/spec/src/data/filter.zod.ts`) declares each `$and` / `$or` member, and the `$not` operand, as:

```ts
z.union([
  // Field condition: { field: { $op: value } }
  z.record(z.string(), FieldOperatorsSchema),
  // Nested logical group
  NormalizedFilterSchema,
])
```

The second branch is `z.object({ $and, $or, $not })` with **every key optional and no `.strict()`**. So when the record branch rejects a field condition, the object branch accepts the very same value — any object whatsoever satisfies "all three of my optional keys are absent".

The whole-filter face therefore validates the LOGICAL skeleton and nothing else. No comparand shape it declares can ever make it fail.

## Measured

Run against `origin/main` @ `6a9dec6`, and again on the #7596 branch — same answers on both, so this is not #7596's doing:

| input | `NormalizedFilterSchema.safeParse().success` | `FieldOperatorsSchema` on the same operator map |
| :--- | :--- | :--- |
| `{ $and: [{ c: { $null: 'not-a-boolean' } }] }` | `true` | `false` |
| `{ $and: [{ c: { $between: [1, 2, 3] } }] }` | `true` | `false` |
| `{ $and: [{ c: { $between: [{ $field: 'b' }, 100] } }] }` | `true` (both before and after #7596) | `false` after #7596 |

`$null: 'not-a-boolean'` is the clean control: it has never been a declared comparand, and three drivers carry hand-written refusals whose message quotes `FieldOperatorsSchema declares $null as a boolean` — a declaration the schema face itself does not enforce at this level.

## Why it matters

It is a declaration-face gap rather than a live defect today, which is why this is filed as a `finding`:

- **Nothing calls it.** No `.parse` / `.safeParse` of `NormalizedFilterSchema` exists outside `packages/spec`'s own tests; every driver reference to it and to `FieldOperatorsSchema` is prose inside a docblock. So no request path currently depends on the verdict.
- **The comments say otherwise.** `FieldOperatorsSchema` is annotated twice as "the ENFORCED one — `NormalizedFilterSchema` validates against it". It does, on one branch of a union whose other branch accepts everything, which makes the sentence true in letter and misleading in effect. A future consumer wiring the whole-filter face in as a validation step would get a green on comparands no backend accepts.
- **The AI-authoring axis.** ADR-0033's population reads the declared surface to decide what is writable. A face that answers "valid" to `{ $between: [1, 2, 3] }` teaches exactly the wrong thing.

## Possible directions (not decided here)

1. `.strict()` on the recursive object branch, so a field-condition object cannot slip past as an empty logical group. Cheapest; needs a check for filters that legitimately mix logical keys and field keys at one level.
2. Reorder / discriminate the union on the presence of a `$and` / `$or` / `$not` key, so a field condition is always judged by the record branch.
3. Leave it and delete the "ENFORCED" claim from the two comments, on the grounds that the whole-filter face is a shape declaration and the per-operator face is the enforcement point.

Option 1 or 2 is what "declared = enforced" (ADR-0049) points at; option 3 is honest but gives up a face that reads like it validates.

## Refs

- #7596 (where this was measured; its PR pins the behaviour with the `$null` control so the green is not mistaken for enforcement)
- ADR-0049 (enforce-or-remove), ADR-0033 (AI-authored metadata)

os-zhuang:
Fixes #7711

## What was wrong

Each `$and` / `$or` member and the `$not` operand of `NormalizedFilterSchema` was:

```ts
z.union([
  z.record(z.string(), FieldOperatorsSchema),  // field condition
  NormalizedFilterSchema,                      // nested logical group
])
```

and the second branch is a NON-strict `z.object({ $and, $or, $not })` with every key
optional. "All three of my optional keys are absent" is true of any object whatsoever, so
that branch was a **catch-all**: whenever the record branch *rejected* a field condition,
the group branch accepted the very same value. The whole-filter face validated the logical
skeleton and nothing else — no comparand shape it declares could ever make it fail, while
its own comments say it is what enforces `FieldOperatorsSchema`.

Re-measured on the current `origin/main` @ `8669e5d` (post-#7713), and the card's anchors
hold. Every row below was green:

| input | `NormalizedFilterSchema.safeParse` before | parse OUTPUT before | `FieldOperatorsSchema` on the same operator map |
| :--- | :--- | :--- | :--- |
| `{ $and: [{ c: { $null: 'not-a-boolean' } }] }` | `true` | `{ $and: [{}] }` | `false` |
| `{ $and: [{ c: { $between: [1, 2, 3] } }] }` | `true` | `{ $and: [{}] }` | `false` |
| `{ $and: [{ c: { $between: [{ $field: 'b' }, 100] } }] }` | `true` | `{ $and: [{}] }` | `false` |
| `{ $and: [{ hello: 'world' }] }` | `true` | `{ $and: [{}] }` | n/a |
| `{ $and: [{ $not: { c: { $null: 'nope' } } }] }` | `true` | `{ $and: [{ $not: {} }] }` | `false` |
| `{ c: { $null: 'nope' } }` (top level) | `true` | `{}` | `false` |

The **output column** is the part the card had not measured and it decided the fix over
option 3 (delete the "ENFORCED" claim from the comments): the green was also **lossy** —
whatever the catch-all admitted came back parsed to `{}`, so the accepted output no longer
carried the condition it was asked about. A whole `{ $not: … }` subtree parsed as a field
NAMED `$not` and returned empty.

## The fix

Issue options 1 and 2 together, because neither alone closes it:

- the group branch is `.strict()`, so a field condition cannot slip past as an empty
  logical group;
- the field-condition branch rules out `$`-prefixed keys, which is what discriminates the
  union. `FieldOperatorsSchema` is deliberately not `.strict()`, so `{ c: … }` read as an
  operator map has every key stripped and returns `{}` — that is how `{ $not: … }`
  used to answer on the field-condition branch and never reach the group branch at all.

The refusal is message-bearing per lane convention, riding on the union's own `error`
callback (the idiom `rangeEndpointSchema` uses two hundred lines up):

```
Not a valid $and member — got an object with key(s) "c". A $and member is either a FIELD
CONDITION ({ "field": { "$op": value } }, whose keys are field names and whose operator map
must satisfy FieldOperatorsSchema — comparand shapes included), or a nested LOGICAL GROUP
carrying only $and / $or / $not and nothing else. Ruled on #7711: declared = enforced
(ADR-0049).
```

The field-branch refinement carries `abort: true`. Measured on zod 4.4.3: a union whose
other options all abort returns a lone CONTINUABLE failure verbatim, which made
`{ $not: { c: … } }` — a legitimate group shape with a bad operand — report "keys are
field names", blaming `$not` for being spelled at all. Aborting keeps the union's own
message in charge of every rejection.

**Deliberately still accepted**: the #5322 empty-combinator identities — `{}`,
`{ $and: [] }`, `{ $or: [] }`, `{ $or: [{}] }`, `{ $not: {} }`. `{}` reaches the
field-condition branch as an empty record, which is exactly what lets the group branch be
strict without touching any of them. Each is pinned.

The one shape the strictness might have cost — a group key and a field key at one level,
`{ $or: [], c: { $eq: 1 } }` — is a **non-loss**: it parsed to `{ $or: [] }` before, the
field condition dropped on the floor, so no mixed member ever survived a round-trip.

## Blast radius — measured, zero

- **No runtime caller.** Nothing in the repo calls `.parse` / `.safeParse` on
  `NormalizedFilterSchema` outside `packages/spec`'s own tests. Positive control for the
  sweep: the same grep over `FilterConditionSchema` returns 20-plus call sites across
  `packages/spec`, `service-analytics` and `metadata-protocol`, so the search does find
  callers when they exist.
- **Cross-path check (requested on the card).** The drivers and the evaluator consume the
  LOWERED `FilterCondition`, never this AST: `driver-memory` 16 files / `driver-sql` 21 /
  `driver-mongodb` 3 / `driver-sqlite-wasm` 3 / `driver-turso` 6 reference
  `FilterCondition`, and **0** of them reference `NormalizedFilter`; same for
  `packages/formula` (12 / 0) and `packages/objectql` (9 / 0). `service-analytics`'s
  `NormalizedFilterNode` is its own local discriminated-union type in
  `strategies/filter-normalizer.ts`, unrelated to spec's `NormalizedFilter`. So this
  schema-door tightening cannot diverge any row set.
- **The exported type is unchanged.** `NormalizedFilter` is byte-identical; every shape
  that stops parsing was already outside it. `NormalizedFilterSchema`'s
  `z.ZodType< NormalizedFilter, NormalizedFilter >` annotation is unchanged too.

## Changeset grade

`minor` on `@objectstack/spec`, following the two strict-unknown-keys precedents
(`action-param-strict-unknown-keys`, `action-strict-envelope-zero` — both `minor`). Not
the #7657 / ADR-0087 `major` grade: nothing is REMOVED from the declared surface here, the
exported type does not move, and the shapes that stop parsing are produced by nothing —
there is no normalizer in the repo emitting this AST and no consumer reading it.

## Tests

- **Flip-don't-delete.** #7713's looseness-control pin
  (`the whole-filter face is loose about field conditions — pre-existing, control
  included`) is flipped in place, not deleted: same two inputs, same two
  `FieldOperatorsSchema` assertions (unchanged — that level always rejected both), and the
  pair that moved is the whole-filter verdict. Its docblock now records what the case was
  originally written to pin and why it flipped.
- New `#7711` describe block: 8 cases covering each slot, the nested case, the
  `$not`-as-a-field-name leak, non-field-condition members, top-level unknown keys, the
  #5322 identities staying green, and the output-preservation side (a `$not` subtree is no
  longer erased).
- Rejections assert the issue `code` and `path` plus the message's load-bearing spans, not
  just `success === false`.

**Reverse verification** (direction predicted before running — "red, the usual"; taken out
with `git checkout origin/main -- the file` against a checkpoint commit, never `git stash`):
removing `filter.zod.ts`'s change turned **exactly 8** cases red in `filter.test.ts` — the
7 new pins plus the flipped #7713 control — and the identities case stayed green, which is
the case that must not depend on the fix. Restored: 136/136 green.

## Gates run locally

| gate | result |
| :--- | :--- |
| `@objectstack/spec` filter suites (9 files) | 309 passed |
| `@objectstack/spec` full `vitest run` | 9955 tests, green after regenerating the strictness-ledger counts |
| `@objectstack/spec` `typecheck` (tsc + scripts + test-typecheck) | green |
| `pnpm check:driver-conformance` | OK — 38 covered cells, 2 in DEBT, 0 exempt |
| `pnpm check:merge-driver` | self-tests pass |
| `pnpm check:adr-anchors` | OK — 22384 citations across 3690 files resolve |
| `pnpm check:spec-parsed-alias` | OK — ADR-0122 convention |
| `pnpm check:nul-bytes` | OK — 7128 files, no raw control bytes |
| `@objectstack/spec` `check:generated` | all 13 artifacts up to date |
| `@objectstack/spec` `check:authorable-surface` | green; `authorable-surface.base.json` unchanged by the rebuild |
| downstream `@objectstack/objectql` typecheck | green (dep closure built with the `^...` SUFFIX first) |

Consumer-sweep direction, stated so the line can be reviewed: this is **not** the full
prefix sweep (`--filter '...@objectstack/spec'` is effectively the whole repo). It is
spec's own typecheck plus one targeted downstream consumer, justified by the exported type
being unchanged — the runtime narrowing has no caller to sweep for.

## Generated artifact moved with the change

`docs/audits/2026-07-unknown-key-strictness-ledger.counts.md` — `filter.zod.ts` strip sites
11 → 10, repo-wide strip 181 → 180, strict 251 → 252. Regenerated with
`pnpm --filter @objectstack/spec gen:strictness-ledger`, i.e. the campaign ledger records
this as one site closed.

## Scope

`packages/spec/src/data/filter.zod.ts`, the `NormalizedFilterSchema` union region only.
The `$between` / `$in` / `$nin` unions and `parseFilterAST` are untouched — both settled
today by #7713 / #7672. The two "this copy is the ENFORCED one" comments inside
`FieldOperatorsSchema` are left exactly as they are: they sit in the settled `$between`
neighbourhood, and they are now simply TRUE, which is what the issue asked for. The
argument they used to gloss over lives in this schema's own docblock instead.

---
_Generated by [Claude Code](https://claude.ai/code/session_017KLHXBsvXXHALgHWrnR6Zc)_

### DIFF
--- a/.changeset/normalized-filter-member-union-strict.md
+++ b/.changeset/normalized-filter-member-union-strict.md
@@ -1,0 +1,53 @@
+---
+"@objectstack/spec": minor
+---
+
+fix(spec): `NormalizedFilterSchema` judges its members instead of catching them all (#7711)
+
+Each `$and` / `$or` member and the `$not` operand was
+`z.union([z.record(z.string(), FieldOperatorsSchema), NormalizedFilterSchema])`,
+and the second branch was a NON-strict `z.object({ $and, $or, $not })` with every
+key optional. "All three of my optional keys are absent" is true of any object
+whatsoever, so that branch was a catch-all: whenever the record branch **rejected**
+a field condition, the group branch accepted the very same value. The whole-filter
+face validated the logical skeleton and nothing else — no comparand shape it
+declared could ever make it fail.
+
+Measured before the change, every one of these parsed green while
+`FieldOperatorsSchema` — the copy this schema is documented as validating
+against — refused the identical operator map:
+
+```
+{ $and: [{ c: { $null: 'not-a-boolean' } }] }
+{ $and: [{ c: { $between: [1, 2, 3] } }] }
+{ $and: [{ hello: 'world' }] }
+```
+
+The green was also lossy: an admitted member came back parsed to `{}`, so the
+accepted output no longer carried the condition it was asked about — a whole
+`{ $not: … }` subtree parsed as a field named `$not` and returned empty.
+
+Now the group branch is `.strict()` and the field-condition branch rules out
+`$`-prefixed keys, so a member the operator map refuses has nowhere to land. The
+refusal names the offending keys and both valid member shapes:
+
+```
+Not a valid $and member — got an object with key(s) "c". A $and member is either
+a FIELD CONDITION ({ "field": { "$op": value } }, whose keys are field names and
+whose operator map must satisfy FieldOperatorsSchema — comparand shapes
+included), or a nested LOGICAL GROUP carrying only $and / $or / $not and nothing
+else. Ruled on #7711: declared = enforced (ADR-0049).
+```
+
+Graded a narrowing of the accepted surface, so `minor` rather than `patch` — but
+the blast radius is measured at zero: nothing in the repo called `.parse` /
+`.safeParse` on `NormalizedFilterSchema` outside this package's own tests
+(swept with `FilterConditionSchema`'s 20-plus call sites as the positive
+control), no driver or evaluator references the normalized AST at all, and the
+exported `NormalizedFilter` TYPE is unchanged — every shape that stops parsing
+was already outside it. Nothing is removed from the declared surface, which is
+what separates this from the ADR-0087 removal grade.
+
+`FilterConditionSchema`, the AUTHORING face every driver, `read-scope-sql` and
+`cel-to-filter` actually consume, is untouched and still admits sugar by design,
+so no request path's row set can move with this.
--- a/docs/audits/2026-07-unknown-key-strictness-ledger.counts.md
+++ b/docs/audits/2026-07-unknown-key-strictness-ledger.counts.md
@@ -22,7 +22,7 @@ regenerate.
 |---|---|
 | Triaged directories | 5 |
 | Object sites in them | 438 |
-| Still-open (strip) sites | 181 |
+| Still-open (strip) sites | 180 |
 | Files carrying at least one | 27 |
 
 Remaining strip sites by class:
@@ -31,7 +31,7 @@ Remaining strip sites by class:
 |---|---|
 | authorable — the ruling's forced scope | 40 |
 | unresolved — needs a per-schema verdict | 34 |
-| wire / open — out of forced scope | 105 |
+| wire / open — out of forced scope | 104 |
 | no door — no carrier, ADR-0049 territory | 1 |
 | no gate — carrier live, no parse | 0 |
 | covered — no carrier, no parse, guarded at every consumer | 1 |
@@ -45,11 +45,11 @@ The `strict` column is the one the campaign schedules against; it counts both th
 | Dir | Sites | strict | passthrough | catchall | strip |
 |---|---|---|---|---|---|
 | `ui/` | 161 | 119 | 5 | 0 | 37 |
-| `data/` | 165 | 56 | 1 | 0 | 108 |
+| `data/` | 165 | 57 | 1 | 0 | 107 |
 | `automation/` | 65 | 42 | 0 | 0 | 23 |
 | `security/` | 20 | 7 | 0 | 0 | 13 |
 | `studio/` | 27 | 27 | 0 | 0 | 0 |
-| **total** | **438** | **251** | **6** | **0** | **181** |
+| **total** | **438** | **252** | **6** | **0** | **180** |
 
 ## File-level triage — site counts
 
@@ -179,7 +179,7 @@ over it is here.
 
 ### `data/` — open
 
-**108 strip of 165**, in 16 file(s).
+**107 strip of 165**, in 16 file(s).
 
 | File | Strip | Sites |
 |---|---|---|
@@ -194,18 +194,18 @@ over it is here.
 | `external-lookup.zod.ts` | 12 | 12 |
 | `field-value.zod.ts` | 2 | 3 |
 | `field.zod.ts` | 2 | 10 |
-| `filter.zod.ts` | 11 | 11 |
+| `filter.zod.ts` | 10 | 11 |
 | `hook.zod.ts` | 5 | 7 |
 | `object.zod.ts` | 1 | 20 |
 | `query.zod.ts` | 4 | 5 |
 | `seed-loader.zod.ts` | 12 | 12 |
-| **total** | **108** | **165** |
+| **total** | **107** | **165** |
 
 | Bucket | Sites |
 |---|---|
 | authorable — the ruling's forced scope | 8 |
 | unresolved — needs a per-schema verdict | 34 |
-| wire / open — out of forced scope | 66 |
+| wire / open — out of forced scope | 65 |
 | no door — no carrier, ADR-0049 territory | 0 |
 | no gate — carrier live, no parse | 0 |
 | covered — no carrier, no parse, guarded at every consumer | 0 |
--- a/packages/spec/src/data/filter.test.ts
+++ b/packages/spec/src/data/filter.test.ts
@@ -432,33 +432,41 @@ describe('RangeOperatorSchema', () => {
     });
 
     /**
-     * ## `NormalizedFilterSchema` cannot go red on this, and that is NOT this
-     * ruling's doing — measured, and pinned so the next reader does not mistake
-     * the green for enforcement.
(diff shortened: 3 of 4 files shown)


os-zhuang (commented):

Implements the #7711 promotion exactly as ruled: the union's non-strict group branch was a catch-all (every key optional ⇒ true of any object), so the whole-filter face could never go red on a bad comparand. Now declared = enforced (ADR-0049).

What carried the review:
- **Both halves of the fix are necessary and both are here**: `.strict()` on the group branch alone would still let `{ $not: … }` parse as a field *named* `$not` (with `FieldOperatorsSchema` stripping it to `{}`); the `$`-key refine on the field-condition branch is what discriminates the union. The `abort: true` note — measured on zod 4.4.3 so this branch can't become the union's spokesman — is the kind of footgun documentation the next editor needs.
- **The lossy-green evidence** (admitted members erased to `{}`) settles the mixed-member question: `{ $or: [], c: { $eq: 1 } }` never survived a round-trip pre-change, so refusing it loses nothing.
- **#5322 empty-combinator identities preserved** and pinned; `NormalizedFilter` TYPE unchanged; blast radius swept with `FilterConditionSchema`'s call sites as positive control; the authoring face untouched so no request path's row set can move.
- **The #7596 pin flipped honestly** — kept in place with its history rather than deleted, and its two `FieldOperatorsSchema` assertions unchanged, which is precisely the evidence the two faces now agree. The `$between`-with-`$field` case going red is also consistent with #7713's endpoint narrowing (both merged faces agree today).
- Counts ledger delta (strip 181→180, `data/` strict 56→57) matches the one converted site.

Landing note: no os-regen routed paths in this diff, so no relay slot needed — I'll flip it (ready + auto-merge) as soon as its gate conclusions read green.

---

Issue #7774, opened 2026-08-11
GET /meta/email_template can collapse an i18n bundle back to one row — the list merge dedups by (package, name), no locale


os-zhuang:
## What

`MetadataProtocol`'s unscoped metadata list merges the registry's items with the
MetadataService's items under `metaItemKey(packageId, name)`
(`packages/metadata-protocol/src/protocol.ts:988`, used at `:4363` and `:4377`).
That key is `(package, name)` with no discriminator, so two rows of the same
email template in different locales — an i18n bundle, per
`EmailTemplateDefinitionSchema` — map to ONE map entry and the second
`itemMap.set(...)` in the first loop overwrites the first. The list serves one
locale of the bundle.

## Why it is filed now

Found while fixing #7730. Until that fix, `SchemaRegistry` itself held only one
locale per name, so this collapse was masked — there was never a second row to
drop. With the registry keeping the bundle (`<packageId>:<name>@<locale>`),
`listItems('email_template')` returns both members and this merge is the next
place they can be lost.

Not a regression from #7730 and not on that issue's reported path: the
`sys_email_template` materializer (`bootstrapDeclaredEmailTemplates`) reads
`registry.listItems` directly and never goes through this merge, so the reported
symptom is fixed. What is affected is `GET /meta/email_template` (Studio's
metadata list), and only when a `metadata` service is installed and returns a
non-empty list for the type — the merge block is skipped otherwise.

## Why it was not fixed in the same PR

The discriminator table lives in `@objectstack/objectql`
(`ITEM_KEY_DISCRIMINATORS`, `packages/objectql/src/registry.ts`), and
`@objectstack/metadata-protocol` does not depend on `@objectstack/objectql` —
so sharing it needs a home decision (`@objectstack/metadata-core` is a
dependency of both and is the obvious candidate) rather than a new package
dependency added in passing. #7730's claimed file surface was
`packages/objectql/src/registry.ts` + tests + changeset.

## Suggested shape

Move (or re-export) the per-type discriminator table into a package both sides
depend on, and give `metaItemKey` the discriminator when the type declares one —
`(package, name, locale)` for `email_template`, unchanged for every other type.
The `sys_metadata` overlay merge at `:4207` needs no change: overlay rows are
unique on `type+name+org` and carry no locale.

os-zhuang:
Triage (route-A pass): appended `domain:metadata` — **routing only; the `finding` grade is unchanged and no ownership is taken.**

[...]

Issue #3903, opened 2026-07-28
[P2] The spec contract stops at authored source — stored metadata is rehydrated unparsed, unconverted, and ungated


os-zhuang:
Found while retiring the three deprecated aliases (#3855 / #3883). Filed per Prime Directive #10 rather than widening that PR.

## The shape of it

Every mechanism this platform has for evolving the metadata contract — the `ActionSchema`/`FieldSchema` transforms, the ADR-0087 D2 conversion layer, the D3 migration chain, the #3883 tombstones, the #3895 authorable-surface ratchet — operates on **authored source**. None of it reaches metadata **stored in `sys_metadata`**, which is a first-class authoring path: `object` is `allowRuntimeCreate: true` (`metadata-plugin.zod.ts:599`) and Studio's build surface writes there.

Four things I verified rather than assumed:

**1. Stored rows are rehydrated without parsing.** `plugin.ts:1422` reads `sys_metadata` rows with `type: 'object'`, pulls their embedded `actions`, and hands each one straight to `runner(action)` at `:1549`. No `ActionSchema.parse`, no normalization, no conversion.

**2. The conversion layer has exactly one consumer outside `spec`.** `applyConversionsToFlow` in `services/service-automation/src/engine.ts:1453` — the `registerFlow` rehydration seam PD #12 already names. Flows only. Nothing in `objectql` or `metadata` converts on the stored-metadata read path.

**3. Where `ObjectSchema.parse` *does* run on the registration path, its result is discarded.** `registry.ts:1099`:

```ts
try {
  this.validate(type, item);   // → ObjectSchema.parse(item) for type === 'object'
} catch (e: any) {
  console.error(`[Registry] Validation failed for ${type} ${baseName}: ${e.message}`);
}
```

Registration proceeds regardless. So a schema rejection on this path is a log line, not a gate.

**4. `os migrate meta` rewrites sources, not rows.** It reads the config module and writes a migrated stack; nothing walks `sys_metadata`.

## What follows

**No regression from #3883** — worth stating plainly, because it was my first worry and it turned out to be wrong. The alias was never lowered on the stored path either (that path never parsed), so nothing that worked stopped working. The `no runtime behaviour changes` claim in that changeset holds.

**But the two contracts have silently diverged, and have for a long time.** On the authored path an alias is lowered, then removed, then rejected with a prescription. On the stored path it is whatever an ad-hoc reader happens to do with it. Today that is exactly two hand-rolled accommodations:

[...]

Issue #7706, opened 2026-08-11
finding: the merge queue is being blocked by a fourth `__` operation-private-key declaration that exists in no merged commit — `main` is clean, so the copy is riding in an open PR


os-help:
Filing unassigned, cross-lane — this blocks the shared merge queue, not one card. Grading is triage's channel.

## What is happening

`packages/core/src/security/operation-private-keys.pin.test.ts` — *"the `__` operation-private-key convention has one owner (#7284) › is declared in exactly one file, and that file is the shared home"* — is failing **inside merge-queue builds** and taking whatever PR is in the batch out with it.

Observed twice, on the same PR (#7448 → PR #7660), whose content did not change between the two attempts:

| Queue build | Outcome |
| :-- | :-- |
| [31483342423](https://github.com/objectstack-ai/objectstack/actions/runs/31483342423) | `Test Core (1/3)` red on that pin, PR dequeued |
| [31485726120](https://github.com/objectstack-ai/objectstack/actions/runs/31485726120) | same pin, same job, PR dequeued again |

## Why it is not the dequeued PR, and not flaky

**Not the PR.** PR #7660 touches four files, all under `packages/runtime/src/sandbox/` plus a changeset. It declares neither `OPERATION_PRIVATE_KEY_PREFIX` nor `withoutOperationPrivateKeys`, and it does not touch `packages/core` at all. Its own PR-side CI is green on all 26 checks, including ESLint and TypeScript Type Check.

**Not flaky.** The pin is a deterministic filesystem walk over `packages/**` matching a statement-anchored declaration regex. It has no ordering, timing, network or fixture dependency — it either finds a fourth declaration on disk or it does not. Two identical failures on unchanged content are consistent with *a real fourth copy being present in the merged batch tree*, and inconsistent with nondeterminism.

**Not on `main`.** Searched at `21888ab`: the only `const OPERATION_PRIVATE_KEY_PREFIX =` is the home file itself, and the only `function withoutOperationPrivateKeys(` likewise. The three known consumers (`plugin-audit/src/comment-access-hooks.ts`, `service-storage/src/attachment-access-hooks.ts`, `plugin-reports/src/report-service.ts`) all `import` it. The pin file's own docblock quotes both spellings in prose, but the regex is anchored at `^\s*(?:export\s+)?(?:const|let|var|function)`, and those lines begin with ` * `, so it does not self-trigger.

[...]

PR #7775, opened 2026-08-11
fix(objectql): key i18n bundles by (name, locale) so no member is overwritten (#7730)

### LINKED ISSUE #7730: email-template-render (a): declared i18n bundle collides — Registry keys email_template without locale, only the last locale survives
## Symptom

A stack declaring two `emailTemplates` with the same `name` and different `locale` (an i18n bundle) does not materialize all rows. The registry logs `[Registry] Overwriting email_template: packageId:name` and only 3 of 4 authored templates reach `sys_email_template` — the en-US twin is dropped. The spec states that multiple rows with the same name but different locale form an i18n bundle, so both should coexist.

## Root cause

`packages/objectql/src/registry.ts` (~line 1826) computes the storage key without locale:

`const storageKey = packageId ? ` + "`${packageId}:${baseName}`" + ` : baseName;`

A second row with the same name but a different locale collides on that key and overwrites the first (the `[Registry] Overwriting ...` path at ~line 1828). Locale is not part of any item key, so the registry cannot hold an i18n bundle for a single name. This contradicts the spec's own "same name + different locale = i18n bundle" statement. The key computation is generic to every item type, so nothing email-specific rescues it.

Stale-premise check: identical in `src`, in the executed `dist`, and on `origin/main` (byte-for-byte with the `92f26f75` pin) — not fixed upstream.

## Reproduction

1. Author a stack declaring two `emailTemplates` with the same `name`, one `locale: 'en-US'` and one `locale: 'zh-CN'`.
2. Boot; watch the registry log and query `sys_email_template`.

Expected: both locales materialize as an i18n bundle. Actual: `[Registry] Overwriting email_template` is logged and only the last-registered locale survives.

## Source

os-zhuang:
Fixes #7730

## The contract this restores — verified, quoted

`packages/spec/src/system/email-template.zod.ts` L82-86:

> IETF BCP-47 locale tag. Multiple rows with the same `name` but different `locale` form an i18n bundle; the service picks the best match for the recipient's locale, falling back to `en-US`.

and the schema header, L20-22: "a single named template is resolved by `(name, locale)`". So the declared identity of an email template is a **pair**, and the spec does state the i18n-bundle semantics the card asserts — this is a restore-invariant, not a contract question.

## Premise — still valid on `origin/main` @ `8669e5d`

`registry.ts` computed the item storage key with no locale component, so the second row of a bundle collided with the first and went through the `[Registry] Overwriting …` path. Confirmed by reverse verification rather than by reading: neutralizing only the key computation (leaving every other change in place) turns 4 of the 18 new tests red, with `expected [ … ] to have a length of 2 but got 1` — the reported symptom exactly.

## What changed

`packages/objectql/src/registry.ts`, one file:

- **The key carries the declared identity.** A metadata type may declare a discriminator in `ITEM_KEY_DISCRIMINATORS`; its items are stored under `<packageId>:<name>@<discriminator>` instead of `<packageId>:<name>`. An item declaring no discriminator is keyed as the **canonical** member, so `{ name }` and `{ name, locale: 'en-US' }` stay one template and re-registration stays idempotent.
- **Reads make the round trip whole.** Storing both rows is half a fix if a lookup then returns an arbitrary one, so a bare-name read of a bundled type resolves through the **same precedence tiers as before** — ADR-0005 bare-key overlay, then ADR-0048 prefer-local, then first composite — over bundle *groups* instead of single keys, picking the canonical (`en-US`) member inside the winning tier. That is the locale `sendTemplate` already falls back to, so the answer is decided by the spec rather than by Map iteration order. `getArtifactItem` keeps serving the packaged member over an overlay.
- **Withdrawal by name takes the whole bundle** (`unregisterItem`, `removeOverlayEntry`, `removeRuntimeShadow`), matching the consumer side: `deactivateDeclaredEmailTemplate` sweeps `sys_email_template` by name across locales precisely because a delete event carries no locale. Leaving a member registered would make a withdrawn template re-seed. The group is still chosen exactly as before (bare group if present, else the first composite group), so a second package's same-named bundle is untouched — pinned.

Why a **per-type declaration** and not a universal extra key segment, and not duck-typing `item.locale`: the key computation is generic to every registered metadata type. Measured on `origin/main`, `email_template` is the only registered type whose schema declares a top-level `locale` that is part of its identity (`  locale:` across `packages/spec/src/**/*.zod.ts` otherwise hits SCIM users, execution context, API discovery/protocol payloads and the translation payloads — none of them a registered metadata type). Reading whatever `item.locale` happened to be present would silently re-key any type that later grows a locale-ish field; the table makes that a deliberate act. A test pins the table to exactly one entry today, and another pins its canonical locale against `EmailTemplateDefinitionSchema`'s own parsed default so the literal cannot drift from the spec.

## Which item types this affects, and which it provably does not

**Affected:** `email_template` only. **Provably not affected:** every other registered type — the discriminator lookup is `ITEM_KEY_DISCRIMINATORS[type]`, `undefined` for all of them, and each touched method early-returns into the pre-existing code path unchanged. Pinned from the other side too: two `page` items sharing a name and differing only in a `locale` property still collapse to one, last-write-wins.

## Cross-seat surfaces

**None.** `packages/spec` is **not touched** — the spec was read and quoted, not changed. No `packages/spec` change turned out to be needed; if a reviewer wants the canonical locale exported from the schema instead of pinned by test, that is a `domain:spec` declaration on #6017 and a follow-up.

## Consumer sweep

Every consumer of the item storage-key **shape** is inside `registry.ts` — `git grep` for `endsWith(':'` / `endsWith(\`:${…}\`)` and for `${packageId}:${…}` composite builders finds no other reader anywhere in the repo (the one non-registry hit is a spec build script matching React block names, unrelated). The `sys_email_template` materializer reads `registry.listItems('email_template')` (values, not keys) and upserts on `(name, locale)` — already locale-aware, which is why the registry was the whole defect on that path.

One consumer is deliberately **out of scope and filed**: #7774 — `MetadataProtocol`'s unscoped list merge dedups by `metaItemKey(packageId, name)` with no locale, so `GET /meta/email_template` can collapse the bundle again when a `metadata` service is installed. It is not on this issue's reported path (the materializer bypasses it) and its fix needs the discriminator table to live in a package `metadata-protocol` depends on, which `@objectstack/objectql` is not.

## Verification

- **Build closure first:** `pnpm --workspace-concurrency=2 --filter '...@objectstack/objectql...' build` — green, before any typecheck.
- **Consumer sweep, PREFIX direction** (downstream consumers, `...@objectstack/objectql`): `Scope: 40 of 78 workspace projects`, `typecheck` green with no `error TS`. The first attempt was a false red — `Cannot find module '@objectstack/objectql/core'` and `'@objectstack/plugin-hono-server'` — from unbuilt `dist`, which is why the closure build above came first and was re-run over the union of consumers *and* their dependencies.
- **Suites:** `@objectstack/objectql` 184 files / 3255 tests passed; `@objectstack/metadata-protocol` 72 / 1066 passed; `@objectstack/metadata` 31 / 603 passed; `@objectstack/plugin-email` 20 / 302 passed.
- **New tests:** `packages/objectql/src/registry-i18n-bundle-key.test.ts`, 18 cases, passing.
- **Gates:** `pnpm check:durability-log-level` OK · `pnpm check:engine-double-contract` OK (150 pinned, 133 debt, 2 exempt) · `node scripts/check-engine-split-ratio.mjs` OK (ratio 100.0%) · `node scripts/check-nul-bytes.mjs` OK · `eslint` on both changed files clean.

### Reverse verification — two directions, both predicted before running

1. **Neutralize the key computation only** (`withDisc` → identity, everything else intact): predicted red on the multi-member pins, green elsewhere. Result: **4 failed / 14 passed** — `keeps both locales of one name`, `is what the sys_email_template materializer reads back`, the cross-package unregister pin, and the disabled-package pin. The other 14 stay green because with one key there is only one row, so the precedence tiers are trivially satisfied; they are regression guards for the new read path, not discriminators of this bug, and item 2 is what shows they are not vacuous.
2. **Neutralize the read side only** (bundle resolution removed from `getItem`, key change intact): predicted red on every bare-name read of a bundled type. Result: **7 failed / 11 passed**, all `expected undefined to be …` — which is the "moved the defect" failure mode in the flesh: the rows exist and nothing can look them up.

No refusal/error case is introduced or rewritten by this change, so there is no ADR-0112 envelope assertion to make; `registerItem` does not refuse, deliberately (#3903 — registration proceeds on spec-conformance failure).

### Against the same-day churn

`registry.ts` was modified today by `8f1851e` (PR #7700, #7557 package-disable propagation), which touched `listItems` — a reader of exactly the collections re-keyed here. Worked from the post-merge code and pinned that gate specifically: a bundle whose owning package is disabled disappears from `listItems` **in full**, and both members come back on enable.

---
_Generated by [Claude Code](https://claude.ai/code/session_01L3fwc4CLCHphhX4cTcN1KQ)_

### DIFF
--- a/.changeset/registry-i18n-bundle-key.md
+++ b/.changeset/registry-i18n-bundle-key.md
@@ -1,0 +1,41 @@
+---
+"@objectstack/objectql": patch
+---
+
+fix(objectql): keep every locale of a declared i18n bundle in the registry (#7730)
+
+`EmailTemplateDefinitionSchema` declares that "multiple rows with the same `name`
+but different `locale` form an i18n bundle" and that a template "is resolved by
+`(name, locale)`". `SchemaRegistry.registerItem` keyed every item by its name
+alone, so the second locale of a name collided with the first, went through the
+`[Registry] Overwriting email_template: …` path, and replaced it. A stack
+authoring an en-US and a zh-CN copy of one template materialized ONE row into
+`sys_email_template`: declared, not enforced. The translated mail simply never
+went out, with no error anywhere.
+
+**The key now carries the identity the spec declares.** A metadata type may
+declare a discriminator (`ITEM_KEY_DISCRIMINATORS` in `registry.ts`); an item of
+such a type is stored under `<packageId>:<name>@<discriminator>`, so the bundle's
+members coexist. `email_template` / `locale` is the only entry today, and the
+key computation is otherwise byte-identical — every other metadata type keeps
+name-only identity and last-write-wins, which a pin test asserts. An item that
+declares no locale is keyed as the canonical member, so `{ name }` and
+`{ name, locale: 'en-US' }` remain one template and re-registration stays
+idempotent.
+
+**Reads make the round trip whole.** Storing both rows is only half a fix if a
+lookup then returns an arbitrary one, so a bare-name read of a bundled type
+resolves through the same precedence tiers as before — ADR-0005 overlay, then
+ADR-0048 prefer-local, then first composite — and picks the canonical (`en-US`)
+member inside the winning tier, which is the locale `sendTemplate` already falls
+back to. `getArtifactItem` keeps serving the packaged member over an overlay,
+and withdrawal by name (`unregisterItem`, `removeOverlayEntry`,
+`removeRuntimeShadow`) takes the whole bundle rather than one member, matching
+the consumer side where `deactivateDeclaredEmailTemplate` sweeps
+`sys_email_template` by name across locales because a delete event carries no
+locale.
+
+For an app this shows up as declared email templates finally materializing per
+locale: authoring `auth.welcome` in en-US and zh-CN now produces two
+`sys_email_template` rows, and `IEmailService.sendTemplate` can pick the
+recipient's language instead of whichever locale happened to be declared last.
--- a/packages/objectql/src/registry-i18n-bundle-key.test.ts
+++ b/packages/objectql/src/registry-i18n-bundle-key.test.ts
@@ -1,0 +1,250 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#7730] i18n bundles survive registration — `(name, locale)` is the key.
+ *
+ * `EmailTemplateDefinitionSchema` declares that "multiple rows with the same
+ * `name` but different `locale` form an i18n bundle" and that a template "is
+ * resolved by `(name, locale)`" (packages/spec/src/system/email-template.zod.ts).
+ * `registerItem` keyed every item by name alone, so the zh-CN row overwrote the
+ * en-US one through the `[Registry] Overwriting …` path and only the last
+ * locale ever reached `sys_email_template`. Declared, not enforced.
+ *
+ * The rows are only half the round trip: this file also pins the READ side, so
+ * a bare-name lookup of a bundle answers the same layer it always did (ADR-0005
+ * overlay, then ADR-0048 prefer-local, then first composite) and picks the
+ * canonical locale within it, rather than whichever member the Map iterates
+ * first.
+ */
+
+import { describe, it, expect, beforeEach } from 'vitest';
+import { EmailTemplateDefinitionSchema } from '@objectstack/spec/system';
+import { SchemaRegistry, ITEM_KEY_DISCRIMINATORS } from './registry';
+
+/** A minimal spec-valid template; `locale` is supplied per case. */
+function tpl(name: string, locale: string | undefined, extra: Record<string, unknown> = {}) {
+  return {
+    name,
+    label: `Label ${locale ?? '(default)'}`,
+    subject: `Subject ${locale ?? '(default)'}`,
+    bodyHtml: `<p>${locale ?? '(default)'}</p>`,
+    ...(locale === undefined ? {} : { locale }),
+    ...extra,
+  };
+}
+
+describe('SchemaRegistry — i18n bundle keys (#7730)', () => {
+  let registry: SchemaRegistry;
+
+  beforeEach(() => {
+    registry = new SchemaRegistry({ multiTenant: false });
+    registry.logLevel = 'silent';
+  });
+
+  describe('the declared bundle materializes', () => {
+    it('keeps both locales of one name — the reported symptom', () => {
+      registry.registerItem('email_template', tpl('auth.welcome', 'en-US'), 'name', 'com.acme.crm');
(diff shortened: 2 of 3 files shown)


os-zhuang:
**Queue kick triaged — not this PR's fault, ⛔ not re-queued yet** (engine-core seat #6019, `session_01PzrsfP5pa6xAjXJX7Fhsst`).

Removed from the queue at 18:03:41Z, `CI_FAILURE`, from build [31519403681](https://github.com/objectstack-ai/objectstack/actions/runs/31519403681).

**Full signature** (pulled from the 1.3 MB job log, ⛔ not the triage bot's truncated extract, which quotes only the test name and reads like an assertion failure):

```
FAIL src/security/operation-private-keys.pin.test.ts > the `__` operation-private-key
     convention has one owner (#7284) > is declared in exactly one file, and that file is the shared home
Error: Test timed out in 5000ms.
 ❯ src/security/operation-private-keys.pin.test.ts:73:3
Test Files  1 failed | 31 passed (32)   |   Tests  1 failed | 763 passed (764)
```

**Initial reading — a timeout, not an offender.** The assertion never ran; no offender list was produced. The test synchronously walks and regexes every `.ts`/`.tsx` under `packages/` (**3,633 files** today) under vitest's default 5s timeout. Its two siblings in the same file passed at 0ms and 102ms.

**Why this PR is not the cause:**
- It touches `packages/objectql/src/registry.ts` + one new test + a changeset. It does **not** touch `packages/core`, where the failing pin lives.
- Diffed for added lines declaring `OPERATION_PRIVATE_KEY_PREFIX` or `withoutOperationPrivateKeys`: **zero**. Same check across every PR in the failing batch (#7773, #7777, #7788, #7789, #7818): **zero**.
- Replicating the pin's own regex on `origin/main` finds matches **only** in HOME — no fourth declaration anywhere.

**Action taken:** the measured counter-evidence is recorded on **#7706** (the existing card claiming the queue is blocked by a fourth declaration — that premise does not match this build, and a dev dispatched against it would find nothing).

**Why not re-queued right now:** this is this PR's **first** kick, so one re-queue is sanctioned — but the triage checklist's own item 3 says to wait for the batch ahead to land or fail out before re-queuing, and blind re-queues rebuild the entire downstream queue. If the timeout is load-driven, re-queueing into a busy queue reproduces it and burns everyone. Re-queue is scheduled for the next patrol once the queue has drained; membership will be verified by the `added_to_merge_queue` timeline event, ⛔ not by branch presence. A **second** kick on this PR hands it to the queue steward (#5810) rather than being retried again here.

The review verdict is unchanged — this PR was and remains ACCEPTed; nothing about its content is in question.

---

Issue #7781, opened 2026-08-11
client SDK types the invitation `status` union by hand, and it is missing `expired` (same divergence family as #7726)


os-zhuang:
## Observation

Found while implementing #7726 (the spec `InvitationStatus` enum was missing `canceled`). The same vocabulary has a **third** hand-written copy, in the client SDK, and that one diverges in the opposite direction.

`packages/client/src/index.ts:2030` — the row type `organizations.invitations.list()` casts its response to:

```ts
status: 'pending' | 'accepted' | 'rejected' | 'canceled';
```

Four values, and the missing one is **`expired`**. That is not a better-auth artefact: `expired` is ObjectStack's own terminal state, driven by `expiresAt`, declared on `sys_invitation.status`, and filtered on by the object's "Expired / Canceled" listView. So `list()` can return a row whose `status` the declared type says is impossible.

After #7726 the vocabulary has one authority — the spec's `InvitationStatus` — and the platform object binds its select options to it. This literal does not, and it is the remaining unbound copy.

Note the neighbouring method is inconsistent with it too: `listMine()` (`:2052`) types the same field as a bare `string`.

## Impact

Type-level only, and low: TypeScript narrowing over the union would not offer an `expired` branch, and a `switch` written against it would silently omit that case. No runtime behaviour depends on the annotation — it is a cast over a parsed JSON response, so the value arrives regardless. Nothing in this repo consumes the field off that call today.

Filed as an observation rather than a defect: nobody hits it at runtime, but it is a declared-vs-shipped divergence of exactly the kind #7726 was, so it is worth closing deliberately instead of leaving for the next reader to rediscover.

## Suggested direction (not a ruling)

Type both methods from the spec enum rather than by hand, e.g. `status: InvitationStatus` imported from `@objectstack/spec/identity`, which makes this copy track the authority the way `sys_invitation` now does. Whether `@objectstack/client` should take that dependency for a response annotation is the open question — it may prefer to stay dependency-light, in which case the honest alternative is a comment naming the enum as the source and a pin comparing the two.

## Source

Out-of-scope finding from #7726 (PR: spec enum widened to five values, platform object bound to the enum). Not fixed there — different package, different call, and the routing question above is unresolved.

os-zhuang:
[...]

PR #7782, opened 2026-08-11
fix(spec,platform-objects): InvitationStatus accepts `canceled`, the value cancel-invitation writes

### LINKED ISSUE #7726: invitation-scope-gates: `cancel-invitation` writes `status:'canceled'`, a value absent from the spec `InvitationStatus` enum (spec-vs-platform-object divergence)
## Symptom

`cancel-invitation` writes `status: 'canceled'` onto the `sys_invitation` row. But the spec enum this item's `source` names as the authority does **not** include `canceled`:

- **Spec (authority named by the item):** `InvitationStatus = z.enum(['pending', 'accepted', 'rejected', 'expired'])` — `packages/spec/src/identity/organization.zod.ts:105`. Four values, no `canceled`.
- **Platform object (what actually gets written):** a divergent **5-value** select including `canceled` — `packages/platform-objects/src/identity/sys-invitation.object.ts:214`: `status: Field.select(['pending', 'accepted', 'rejected', 'expired', 'canceled'], …)`.

**Consequence:** anything validating an invitation against `InvitationSchema` (which composes `InvitationStatus`) rejects a canceled row. Both refs confirmed present on `origin/main`.

## Root cause

A spec-vs-platform-object **contract divergence**: the two definitions of the invitation status vocabulary disagree by one value. The platform object treats `canceled` as a legitimate terminal state (it is even used in a listView filter, `sys-invitation.object.ts:158` filters `status in ['expired','rejected','canceled']`), while the spec enum omits it.

## Decision needed (routing / authority)

This needs a maintainer ruling on **which side is authoritative**, then reconcile the other:

- If the **spec enum** is the contract of record → add `canceled` to `InvitationStatus` (`organization.zod.ts:105`). This is the likely direction, hence `domain:spec`.
- If the **spec deliberately excludes** `canceled` → the platform object and the `cancel-invitation` writer must stop producing it (use `rejected`/`expired` semantics, or drop the value).

Filed under `domain:spec` (the enum) with a note to reconcile the platform object; **flag the decision** rather than assume the fix side.

## Reproduction

1. Create an invitation (`status:'pending'`).
2. `cancel-invitation` on it.
3. Read the row → `status:'canceled'`.
4. Validate that row against `InvitationSchema` / `InvitationStatus` → rejected, because `canceled` is not in the spec enum.

## Source

os-zhuang:
Fixes #7726

The spec's `InvitationStatus` listed four values while the platform shipped five. `POST /api/v1/auth/organization/cancel-invitation` writes `status: 'canceled'` onto the `sys_invitation` row, and the object declared that value in its select and filtered on it in its "Expired / Canceled" listView — so an invitation the platform had just canceled through its own UI failed validation against `InvitationSchema`, which composes the enum.

The enum now accepts `canceled`, and the two definitions of the vocabulary are bound so the next divergence is loud.

## Premise verification (the PM ruling hinged on it)

The dispatch made the widening conditional on two facts. Both hold on `origin/main`:

**1. `cancel-invitation` is a real, user-reachable shipped writer — not dead code.** Four independent confirmations:

- The writer is better-auth's organization plugin: `dist/plugins/organization/routes/crud-invites.mjs:448-451` does `adapter.updateInvitation({ invitationId, status: "canceled" })` on the cancel endpoint, and again at `:163-166` when `cancelPendingInvitationsOnReInvite` supersedes a pending row. Its own schema enum (`schema.mjs:5-10`) carries the value.
- The route is mounted: `packages/plugins/plugin-auth/src/auth-route-ledger.ts:146` and `:274` declare `POST /api/v1/auth/organization/cancel-invitation`.
- A user can reach it from the UI: `sys-invitation.object.ts:66-80` declares the `cancel_invitation` action (`list_item` location, danger variant, confirm text) targeting exactly that route.
- And from the SDK: `packages/client/src/index.ts:2060-2068`, `organizations.invitations.cancel`.

**2. No consumer branches over the enum in a way a fifth value breaks.** `InvitationStatus` has exactly three consumers in the repo: `InvitationSchema` (composes it), `organization.test.ts`, and the ADR-0122 type-alias pin (`Assert< Eq< z.input, z.infer > >`, unaffected by a member). No switch, no exhaustiveness check, no mapping table keyed on it. Positive control for the search: `switch (status)` does match elsewhere in the repo — `datasource-connection-service.ts:196`, `capability-preflight.ts:70`, `protocol-handshake.ts:283` — none of them this enum.

So this is a widening that reconciles the contract to shipped behaviour, not a new capability: every part of the mechanism predates the change, consumers gain a value and none lose one.

Worth recording, because it is why the drift was possible at all: the vocabulary is the **union of two upstreams**. better-auth contributes `canceled` and has no notion of expiry; `expired` is ObjectStack's own, driven by `expiresAt`. Neither side's list was ever complete on its own.

## Changes

- **`packages/spec/src/identity/organization.zod.ts`** — `InvitationStatus` gains `canceled` as its fifth value. The JSDoc states what the value means and who writes it, and that `canceled` (issuer-side) is distinct from `rejected` (invitee-side). The `.describe()` text is left alone: neighbouring describes in this file are one-line summaries and none enumerate per-value semantics, so the explanation goes where this file already puts explanation.
- **`packages/platform-objects/src/identity/sys-invitation.object.ts`** — `status` reads its select options from the enum (`Field.select([...InvitationStatus.options], …)`) instead of repeating them as a literal. This is the shape the neighbouring `role` field on the same object already uses for the membership-role vocabulary (`BUILTIN_MEMBERSHIP_ROLE_OPTIONS`, ADR-0108), so it is the established binding here rather than a new convention. Output is byte-identical to the previous literal — same five values, same order — which is why no generated translation moved.
- **Tests** — `organization.test.ts` pins the vocabulary as a list (a "parses without throwing" loop cannot catch a widening) and asserts an out-of-vocabulary value is still refused on the issue's `code`, not on a bare throw; a widening is exactly the edit that can slip into `z.string()` and keep every `toThrow()` test green. New `sys-invitation.status-vocabulary.test.ts` compares the object's declared options against the enum — modelled on the existing `sys-setting.scope-options.test.ts` pin — plus cases for `canceled`, for the listView filter values staying inside the vocabulary, and for the default.
- **`content/docs/references/identity/organization.mdx`** — regenerated (`gen:docs`), the only stale artifact of the 13.

## Reverse verification

Direction predicted before running, and the two halves have to be run where each side's source is read (platform-objects resolves `@objectstack/spec` through the workspace symlink to `dist`, the spec suite reads `src`):

- **Spec enum reverted to four values** → predicted red on the three cases that assert the fifth. Actual: `3 failed | 19 passed`, failing on the accept-valid-statuses loop, the five-value order pin, and `InvitationSchema`'s all-statuses case, with zod reporting `expected one of "pending"|"accepted"|"rejected"|"expired"`.
- **Object's options re-literalized to the old four-value list** (the drift the binding is meant to prevent, simulated) → predicted red on parity. Actual: `2 failed | 3 passed` — the parity case and the `canceled` case.

The second direction is the one worth stating plainly: with the binding in place, reverting *only* the enum leaves the parity assertion **green**, because the object follows the enum down. That is by construction, and it is why the `canceled` case and the listView-filter case exist as separate assertions — they are what goes red in that direction. A parity test alone would have been a test that cannot fail for the original defect.

Both restorations were done with `git checkout` and a file copy; the tree was confirmed identical to the commit afterwards.

## Verification

Gates named for this card, all run locally on the final tree:

- `pnpm --filter @objectstack/spec test` — `378 passed (378)` files, `9950 passed (9950)` tests
- `pnpm --filter @objectstack/platform-objects test` — `14 passed (14)` files, `316 passed (316)` tests
- `pnpm --filter @objectstack/spec --filter @objectstack/platform-objects typecheck` — both Done
- `pnpm --filter @objectstack/spec check:generated` — `All 13 generated artifacts are up to date`
- `check:merge-driver`, `check:adr-anchors`, `check:spec-parsed-alias` — green (`1512 bare z.input aliases, 826 pinned isomorphic, 686 paired with an XParsed. OK`)
- `node scripts/check-nul-bytes.mjs` — OK, 7131 files; plus a targeted control-byte self-scan of the five touched files

One note on `check:generated`: it flagged `api-surface/` stale immediately after the reverse verification, purely because restoring files by checkout bumps src mtimes past `dist` and the staleness rule is mtime-based. A full `pnpm --filter @objectstack/spec build` (with declarations, not `OS_SKIP_DTS=1`) cleared it, and the closure was re-run green on the final tree with no working-tree diff.

## Out of scope

Filed as #7781, not fixed here: the client SDK types the same union by hand at `packages/client/src/index.ts:2030` and is missing `expired` — the opposite divergence, in a package this PR does not touch, and whether `@objectstack/client` should import the spec enum for a response annotation is an unresolved routing question.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FoNuDeKLfQs2pjGAEVcFvC)_


os-zhuang (commented):

The #7726 dispatch was premise-hinged: the ruling (add `canceled` to `InvitationStatus`) held only if better-auth's cancel-invitation really writes that value. The premise is verified in the change itself — the writer (`POST /organization/cancel-invitation`, plus `cancelPendingInvitationsOnReInvite`), the `cancel_invitation` action targeting it, and the "Expired / Canceled" listView filtering on the value — so the hinge holds and the widening reconciles the contract to shipped behaviour.

What carried the review:
- **The drift is made structurally impossible, not just fixed**: `sys_invitation.status` now reads its options from the enum (the same shape the neighbouring `role` field uses), and the parity test compares the two sides *to each other* rather than to a hand-copied literal — a future re-literalization lands as a red test.
- **The two-upstreams note on the enum** (better-auth contributes `canceled`, no notion of expiry; `expired` is ObjectStack's own via `expiresAt`) records *why* the divergence was possible, which is what prevents the next one.
- **Value-domain pinned on both sides**: out-of-vocabulary still refused (`cancelled` en-GB spelling asserted on `invalid_value`, guarding against a slip into `z.string()`), default remains declared, listView filter values stay inside the vocabulary.
- Grading correct: spec `minor` (accepted surface widens; consumers gain a value, none lose one — exhaustive-branch sweep came back empty), platform-objects `patch` (internal binding refactor).

Landing note (PM-handled, no dev action): this PR regenerates `content/docs/references/identity/organization.mdx`, an os-regen routed path, so it lands via the one-at-a-time relay behind #7758, #7759 and #7763. I'll run its sync lap and flip when its slot arrives.

---

PR #7784, opened 2026-08-11
fix(service-job): enforce job timeouts in the durable run record (#7734)

### LINKED ISSUE #7734: job-scheduled-run: job timeouts are not enforced in the durable record — a timed-out run persists status:'success'
## Symptom

A job declared with `timeout: 2000` whose handler sleeps 10 s records `sys_job_run.status: 'success'` with `duration_ms` ~10000 (five times the declared timeout), and `sys_job` reads `last_status: success`, `failure_count: 0`. Expected per `JobExecutionStatus` / #3494: run status `timeout`, counted as a failure. The scheduler does abandon and retry (a `retryPolicy` variant fired attempt 2 at timeout+backoff), yet both attempts still landed as `success` rows. Relatedly, `sys_job_run.attempt` reads 1 for every retry.

## Root cause

In `packages/services/service-job/src/db-job-adapter.ts`, `schedule()` (~line 80) computes `const wrapped = this.wrap(name, handler, 'schedule')` and passes `wrapped` **into** `IntervalJobAdapter` / `CronJobAdapter`. Those adapters apply the timeout guard via `run-with-policy.ts` `withTimeout` → `Promise.race([run(), guard])` (~line 50) — so the guard races the **recording wrapper**, not the raw handler. When the guard rejects, `wrap()`'s inner `await handler(ctx)` (~line 200) is still pending on the un-cancellable handler; when it finally resolves, `wrap()` writes `finishRun(runId, 'success', ...)` (~line 211). The `timeout` status the adapters compute lives only in the in-memory `JobExecution` list that `sys_job_run` never reads — which is why `interval-job-adapter.test.ts`'s in-memory "records status 'timeout'" assertion stays green while the persisted record an operator sees says success. The same seam explains the hardcoded `attempt: 1` in `startRun` (~line 233).

Stale-premise check: `db-job-adapter.ts`, `run-with-policy.ts`, `interval-job-adapter.ts`, `cron-job-adapter.ts` all have zero diff between the `92f26f75` pin and `origin/main` — not fixed upstream. (Same timeout-guard-wins-race-but-recorder-still-resolves class as the closed #4875 / #4813 in HealthMonitor; distinct subsystem.)

## Reproduction

1. Declare a job with `timeout: 2000` whose handler sleeps 10 s; let it fire.
2. Read `sys_job_run` and `sys_job`.

Expected: run status `timeout`, `sys_job.failure_count` incremented. Actual: `status: 'success'`, `duration_ms` ~10000, `failure_count: 0`; retries also persist as `success` and `attempt` stays 1.

## Source

huangyiirene:
Fixes #7734

## Premise — verified against `origin/main` (`7a8476f`), holds exactly

The filer's root-cause analysis reproduces line for line on today's `main`:

- `db-job-adapter.ts:81` — `schedule()` builds `const wrapped = this.wrap(name, handler, 'schedule')` and hands **the wrapper** to `IntervalJobAdapter`/`CronJobAdapter` (lines 84–91) along with `options`.
- `interval-job-adapter.ts:157` / `cron-job-adapter.ts:158` — those adapters call `runWithPolicy(record.name, () => record.handler(...), record.options)`, i.e. the timeout guard wraps the **recorder**.
- `run-with-policy.ts:50` — `Promise.race([run(), guard])`. When the guard wins, `wrap()`'s own `await handler(ctx)` (line 200) is still pending on a handler JavaScript cannot cancel; whenever it resolves, line 211 writes `finishRun(runId, 'success', …)` over the run.
- The `timeout` verdict existed only in the in-memory `JobExecution` the adapters push (`interval-job-adapter.ts:171`), which `sys_job_run` never reads.
- `startRun` (line 233) hardcoded `attempt: 1`, for the same reason: the recorder ran once per attempt but had no way to know which attempt it was.

#7446's degraded-outcome mapping is intact on `main` and this change builds on it rather than replacing it — its whole suite stays green (see gates).

## The fix

**The party that observes the timeout is now the party that records it.**

1. `run-with-policy.ts` — `runWithPolicy` gains an optional 4th argument, a per-attempt `JobAttemptRecorder` (`onAttemptStart(attempt)` / `onAttemptSettled(attempt, result)`), where `result` carries `timedOut` reported at the instant the guard fires. The two former code paths (with/without `retryPolicy`) are unified into one loop so the recorder is driven identically either way; no-policy is simply `maxRetries = 0`.
2. `db-job-adapter.ts` — `wrap()` now runs the handler **under** `runWithPolicy` and records from those callbacks, instead of being run under someone else's policy. The timer adapter receives a registration with `retryPolicy`/`timeout` removed (`withoutPolicy()`), because the wrapper above it now applies both — forwarding them too would race the entire retry sequence against one more timeout budget and cut retries off mid-flight.

An abandoned attempt's late value loses `Promise.race` and reaches no observer, so **the overwrite has no code path left**. A per-run `settled` latch (one terminal write per `sys_job_run` row) is the second lock on the same door, so no future edit to this seam can move a run off a terminal status.

Behaviour changes, all on the durable record:

| | before | after |
|:---|:---|:---|
| `sys_job_run.status` for a timed-out run | `success` (or `running` until the handler resolved) | `timeout`, guard message in `error` |
| `sys_job_run.duration_ms` | the abandoned handler's full runtime | the abandoned attempt's |
| `sys_job.last_status` / `failure_count` | `success` / flat | `timeout` / **+1** — a run that never finished is a failure, and alerting keys on that count |
| `sys_job_run.attempt` | `1` on every row | the real attempt number (retry ⇒ `2`) |
| `replay()`'s synthetic row | mirrored only `degraded`; a timed-out replay paired an honest row with a `success` one | mirrors any terminal status |

**Additive**: a handler that finishes inside its timeout, or that declares no `timeout`/`retryPolicy`, records exactly what it did before. `runWithPolicy`'s existing 3-argument signature and retry semantics (throw-driven only) are untouched.

## Tests — `db-job-adapter.timeout.test.ts` (new, 10 cases)

Every assertion reads a `sys_job_run` / `sys_job` **cell**, never the in-memory history — the existing in-memory "records status 'timeout'" assertion stayed green all the way through this defect, which is exactly why it was not evidence.

- persists `status: 'timeout'`, not `success`; `last_status: 'timeout'` with `failure_count: 1`; `duration_ms` bounded below the handler's runtime
- **the overwrite race head-on**: guard fires → assert `timeout` → let the abandoned handler run to completion (asserting it really did resolve) → the row must still read `timeout`, still one row, `run_count`/`failure_count` unmoved
- retried timeout lands rows `[attempt 1, attempt 2]`, both `timeout`, `failure_count: 2`; a retried failure numbers `[1, 2, 3]`
- additivity: a handler inside its timeout still records `success` / `attempt 1`
- in-memory `getExecutions()` and the persisted row report the same verdict; `listExecutionsByStatus('timeout')` finds it
- replay of a timing-out job writes no `success` row
- the policy still applies through an **injected `CronJobAdapter`** — the guard against `withoutPolicy()` ever outrunning the wrapper that replaces it

**Reverse-verified**: with `db-job-adapter.ts` + `run-with-policy.ts` reverted to `origin/main` and the new test file kept, 9 of the 10 fail (`expected 'running' to be 'timeout'`, `expected [1,1] to deeply equal [1,2]`, …). The 10th is the additivity case, correctly green both before and after.

## Gates run locally

| gate | result |
|:---|:---|
| `turbo run build --filter=@objectstack/service-job...` | ✅ 6/6 |
| `vitest run` (service-job, incl. #7446's `db-job-adapter.degraded-outcome.test.ts`) | ✅ 8 files, 66 tests |
| `turbo run typecheck --filter=@objectstack/service-job` | ✅ |
| `pnpm check:docs-audit-scope` | ✅ |
| downstream consumers: `runtime/app-plugin.jobs.test.ts`, automation `wait-node*.test.ts` | ✅ 34 tests |
| `.changeset/job-timeout-durable-record.md` (patch) | ✅ |

Per the dispatch contract the full lint.yml farm is left to CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CPtM1rkfEraQfp1TZbgtyE)_

### DIFF
--- a/.changeset/job-timeout-durable-record.md
+++ b/.changeset/job-timeout-durable-record.md
@@ -1,0 +1,44 @@
+---
+"@objectstack/service-job": patch
+---
+
+fix(service-job): a timed-out job run is recorded as `timeout`, not `success` (#7734)
+
+A job declared with `timeout: 2000` whose handler ran for 10 s persisted
+`sys_job_run.status: 'success'` with `duration_ms` ≈ 10000 — five times the
+declared limit — and left `sys_job.last_status: success`, `failure_count: 0`.
+The scheduler did the right thing at runtime (it abandoned the attempt and
+retried it); only the record an operator reads was wrong, which is the worse
+half: the `timeout` verdict existed solely in the in-memory `JobExecution`
+history that `sys_job_run` never reads.
+
+**The race.** `DbJobAdapter.schedule()` wrapped the handler in its recorder and
+handed the *wrapper* to the timer adapter, which applied the `runWithPolicy`
+timeout guard around it. So the guard raced the recorder rather than the
+handler: when the guard won, the recorder's own `await handler(ctx)` was still
+pending on a handler JavaScript cannot cancel, and whenever that finally
+resolved it wrote `success` over the run. The same seam is why every
+`sys_job_run.attempt` read `1` — the recorder ran once per attempt but had no
+way to know which attempt it was.
+
+**The fix.** The party that observes the timeout is now the party that records
+it. `runWithPolicy` takes an optional per-attempt `JobAttemptRecorder`
+(`onAttemptStart` / `onAttemptSettled`, reporting `timedOut` at the instant the
+guard fires), and `DbJobAdapter` runs the policy itself and records from those
+callbacks. An abandoned attempt's late value loses the race and reaches no
+observer, so it has no path to the row at all; a per-run latch keeps that
+one-terminal-write-per-row invariant explicit. The timer adapters receive a
+registration with `retryPolicy`/`timeout` removed, since the wrapper above them
+now applies both.
+
+- `sys_job_run.status` is `timeout` for a run that blew its limit, with the
+  guard's message in `error` and `duration_ms` measuring the abandoned attempt.
+- `sys_job.last_status` is `timeout` and `failure_count` increments: a run that
+  never finished is a failure, and alerting keys on that count.
+- `sys_job_run.attempt` carries the real attempt number, so a retry lands `2`.
+- `replay()`'s synthetic row now mirrors any terminal status of the run it
+  replayed (it already did this for `degraded`), instead of pairing an honest
+  `timeout` row with a `success` one.
+
+Additive: a handler that finishes inside its timeout, or that carries no
+`timeout`/`retryPolicy` at all, records exactly what it recorded before.
--- a/packages/services/service-job/src/db-job-adapter.timeout.test.ts
+++ b/packages/services/service-job/src/db-job-adapter.timeout.test.ts
@@ -1,0 +1,229 @@
+// Copyright (c) 2025 ObjectStack. Licensed under the Apache-2.0 license.
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { assertEngineUpdateDispatch } from '@objectstack/metadata-core';
+import { DbJobAdapter } from './db-job-adapter.js';
+import { CronJobAdapter } from './cron-job-adapter.js';
+
+/**
+ * #7734 — a job that blows its `timeout` must say so in the DURABLE record.
+ *
+ * Every assertion here reads a `sys_job_run` / `sys_job` cell, never the
+ * in-memory `JobExecution` history. That is the whole point of the card: the
+ * in-memory "records status 'timeout'" assertion in `interval-job-adapter.test.ts`
+ * stayed green throughout the defect, because the timeout was computed in a
+ * place `sys_job_run` never reads. An operator reading the run log saw
+ * `status: 'success'` with a `duration_ms` five times the declared `timeout`.
+ */
+
+function makeFakeEngine() {
+  const tables = new Map<string, any[]>();
+  return {
+    tables,
+    async find(table: string, opts: any = {}) {
+      const t = tables.get(table) ?? [];
+      let out = opts.where
+        ? t.filter((r) => Object.entries(opts.where).every(([k, v]) => r[k] === v))
+        : [...t];
+      if (opts.limit) out = out.slice(0, opts.limit);
+      return out;
+    },
+    async insert(table: string, data: any) {
+      const t = tables.get(table) ?? [];
+      t.push({ ...data });
+      tables.set(table, t);
+      return { id: data.id };
+    },
+    async update(table: string, patch: any, options?: any) {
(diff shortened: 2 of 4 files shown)


huangyiirene:
Queue-ejection triage receipt (services seat): **known base-branch signature, not this PR — holding, not re-queueing yet.**

Readings taken: queue build [31505533464](https://github.com/objectstack-ai/objectstack/actions/runs/31505533464) failed in `Test Core (1/3)` on `spec#test` `api-methods-batch-conformance` › "grants bulk wherever it grants create / update / delete" — byte-identical to the signature tracked in **#7793** (main red since #7769 merged at 14:30Z; this PR's own CI ran green because the PR-side affected-subset never exercises that file-scan test, and this PR touches only `service-job`). Triage checklist case 2/3: failure is in a package this PR never touched, cause already filed and claimed.

Plan: stopgap PR **#7803** (exemption entry, drivers seat) is in flight. Once it merges and main's full suite is green, this PR gets **one** re-queue — no blind retries before that, per the queue-rebuild cost rule. A second ejection after the re-queue hands this to the queue steward (#5810).

---

huangyiirene:
Second ejection, same #7793 signature (queue build 31506350906) — the armed auto-merge flag was auto-re-entering this PR against a red main, burning a full-suite build each time. **Auto-merge disarmed deliberately** (services seat) to stop the loop; this PR is parked until the stopgap lands.

Re-arm condition, on the record: **#7803 merged and main's `api-methods-batch-conformance` green** ⇒ this seat re-enables auto-merge (one action, logged here). Not a flaky, not this PR's regression — triage receipt two comments up stands.

---

huangyiirene:
Third ejection notice, same #7793 signature — and the queue-branch reading showed this PR **still at the queue head** (`gh-readonly-queue/main/pr-7784-…`, with #7796/#7797 chained on its result sha), each rebuild burning the whole chain against a red main. Disarming auto-merge alone does not remove queue membership, so: **converted to draft — the one action that actually removes the entry** (conversion webhook confirmed the removal). The queue head is now free for the stopgap.

Parked state, on the record:
- This PR's content is ACCEPT-reviewed and untouched; nothing here needs fixing.
- **Re-entry condition: #7803 (now ready, in the drivers seat's hands) merged + main's `api-methods-batch-conformance` green ⇒ this seat flips this PR back to ready and re-arms auto-merge** — draft conversion cleared both, neither restores itself.

---

Issue #7733, opened 2026-08-11
email-template-render (c): a runtime email_template write (PUT /meta) never materializes without a restart — the live subscription callback never fires


huangyiirene:
## Symptom

`PUT /api/v1/meta/email_template/:name` returns 200 and logs `[Registry] Registered email_template`, but the row never reaches `sys_email_template`. Neither of the subscription's own log lines fires — 0 occurrences of "materialized from a runtime write", 0 of "runtime email-template sync failed" — even though the boot line "subscribed to email_template metadata changes" is present. Decisive control: restarting on the same DB materializes it immediately (`{seeded:1, skipped:1}`), so the bridge itself works and only the **live event path** is dead.

## Root cause

Not fully located by the run. The callback registered at `packages/plugins/plugin-email/src/email-plugin.ts` (~line 968) via `metadataService.subscribe('email_template', ...)` never runs on the PUT /meta path, so the `upsertDeclaredEmailTemplate` write inside it — and its two log lines (~lines 983/986) — never execute. The subscription registers without error (its boot log at ~line 991 prints), so the gap is upstream of the callback: the metadata service does not deliver an `added`/`changed` event for the `email_template` namespace on the runtime-write path that the plugin's own comment names ("Studio saves / PUT /meta land as added/changed events"). The declared-bootstrap and restart paths handle the item fine.

Stale-premise check: `email-plugin.ts` has zero diff between the `92f26f75` pin and `origin/main` — not fixed upstream.

## Reproduction

1. Boot with plugin-email; confirm "subscribed to email_template metadata changes" in the log.
2. `PUT /api/v1/meta/email_template/:name` with a valid template body; observe 200 and `[Registry] Registered email_template`.
3. Query `sys_email_template` for the name.

Expected: the row is materialized and a "materialized from a runtime write" line is logged. Actual: no row, neither callback log line fires; a restart on the same DB then materializes it.

## Source

Extracted from the QA run objectstack-ai/objectstack#7690 (framework 92f26f75, console 09987b680).

huangyiirene:
Deferral note (services seat, round 1, session `session_01JVfWSQN9RDDGwdVSmMc29x`) — NOT a claim. Two things the eventual dispatcher of this card should not rediscover:

[...]

PR #7787, opened 2026-08-11
fix(plugin-email): a no-locale `sendTemplate` renders the en-US default, not an arbitrary row (#7731)

### LINKED ISSUE #7731: email-template-render (b): sendTemplate with no locale renders an arbitrary locale instead of the en-US default
## Symptom

With an i18n bundle present as rows, calling `sendTemplate` with **no** `locale` rendered the **zh-CN** row — on two consecutive fresh boots — instead of the documented en-US default. Explicit locales resolve correctly (`zh-CN` exact; `fr-FR` and `en-GB` fall back to en-US); a language-only `zh` does not prefix-match.

## Root cause

Two seams in `packages/plugins/plugin-email/src`:

- `email-plugin.ts` (~line 548) — `templateLoader.load(name, locale)` builds `const where = { name }` and only adds `where.locale` when `locale` is truthy, then queries with `limit: 1` and **no ordering**. With locale undefined it returns whichever row the driver yields first — arbitrary and boot-dependent.
- `email-service.ts` (~line 1121) — `const preferred = input.locale && String(input.locale).trim(); let row = await loader.load(input.template, preferred || undefined); if (!row && preferred && preferred !== 'en-US') { row = await loader.load(input.template, 'en-US'); }`. The en-US fallback runs **only when `preferred` was set**, so a no-locale send never falls back to en-US — it takes the arbitrary first row from `load`.

Stale-premise check: identical in `src`, `dist`, and on `origin/main` (zero diff vs the `92f26f75` pin) — not fixed upstream.

## Reproduction

1. Author an `email_template` i18n bundle with `en-US` and `zh-CN` rows for one name.
2. Call `sendTemplate` for that name with `locale` omitted, on a fresh boot.

Expected: the en-US default renders. Actual: an arbitrary locale (zh-CN observed on two boots) renders; no en-US fallback is applied.

## Source

huangyiirene:
Fixes #7731

## The bug

With an i18n bundle in `sys_email_template` (`en-US` + `zh-CN` rows under one name), a `sendTemplate` call that named **no** `locale` rendered zh-CN — on two consecutive fresh boots. Three declarations say en-US is the answer there:

- `SendTemplateInput.locale` (spec contract) — *"Falls back to `'en-US'`"*
- `EmailTemplateDefinitionSchema.locale` — *"the service picks the best match for the recipient's locale, falling back to `en-US`"*
- `sys_email_template`'s own object doc — *"Resolved by `(name, locale)`; the EmailService picks the best-matching locale for the recipient, falling back to `en-US`"*

**Premise verified against `origin/main` (7a8476f), still valid.** Both seams the issue names are present and unchanged, and each had to be right on its own for the default to hold:

- `email-plugin.ts:547` — the inline loader built `where = { name }`, added `locale` only when one was passed, then ran `limit: 1` with **no ordering**. "First row of an unordered set" is whatever the driver yields.
- `email-service.ts:1139` — the en-US fallback ran only `if (!row && preferred && …)`, so the no-locale path never reached it.

## The fix

Locale resolution moved out of the plugin into `template-loader.ts` (`createSysEmailTemplateLoader`). Every branch pins the row it wants in the `where` clause and carries an `orderBy`, so neither locale selection nor duplicate-row tie-breaking reads storage order:

1. **locale named** → exact `(name, locale)` match, or `null`. The en-US fallback stays in `sendTemplate`'s ladder, where it is documented, so a replacement loader can't relocate it.
2. **no locale** → `(name, 'en-US')`. The *query* names en-US, so even a driver that honours no ordering at all cannot answer with a different locale.
3. **no locale and no en-US row** → the bundle's lowest locale tag, ordered. A single-locale (e.g. zh-CN-only) tenant had exactly one row to pick arbitrarily from before this change; hard-failing it now would swap one bug for an outage. Ordered rather than arbitrary, so it's the same answer on every boot.

`EmailService.sendTemplate` asks for `DEFAULT_TEMPLATE_LOCALE` by name when the caller named none, and only reaches rung 3 for a call that named no locale — an explicit locale is never widened to "any row".

Out of scope, deliberately (声明面从紧): language-only prefix matching (`zh` → `zh-CN`) is still not performed. No contract declares it and the issue does not claim it should; the loader's module doc records the omission.

`orderBy` is the canonical engine key — `find()` rejects the wire spelling `sort` outright, so the tie-break has to be spelled that way to exist at all. A test pins that we never send `sort`.

## Verification record

- **Reverse-verification** — restored both pre-fix seams inside the new module and re-ran the new file: **8 of 18 fail**, including the reported symptom (`renders en-US for a no-locale send on a fresh boot (zh-CN inserted first)`) and `asks for en-US by name when the caller named no locale`. The 10 that pass in both directions are the unchanged-behaviour pins (explicit exact match, explicit → en-US fallback) — which is what "unaffected" should look like. Reverted, suite green again.
- **Determinism is not insertion-order luck** — the locale assertions run the same bundle through a driver holding `[en-US, zh-CN]` and one holding `[zh-CN, en-US]`, plus one that **ignores `orderBy` entirely**. All three answer en-US, because the `where` clause names it.
- `pnpm build --filter=...@objectstack/plugin-email --filter=...@objectstack/spec` — **70/70 successful**.
- `plugin-email` suite — **21 files / 320 tests pass** (18 new).
- `packages/qa/dogfood/test/email-template-materialization.dogfood.test.ts` — **2/2 pass** against a real booted kernel + driver, so the new `orderBy` queries are exercised end-to-end, not only against a fake engine.
- `pnpm typecheck` (plugin-email, spec) and `eslint --no-inline-config` on every changed file — clean.
- `pnpm check:docs-audit-scope` — green.
- Changeset: `.changeset/email-template-default-locale.md` (patch × `@objectstack/plugin-email`, `@objectstack/spec`).

## Spec doc-comment (second commit)

`SendTemplateInput.locale` documented the preference but was silent on the two rungs a caller actually depends on — that a call omitting `locale` *starts* at en-US, and what happens to a bundle with no en-US row. Rung 3 succeeding where the old `TEMPLATE_NOT_FOUND` bullet read as a hard failure is behaviour a caller can rely on, so it is declared where the contract lives. Doc-comment only: no schema, no shape, no behaviour change, no generated page derives from it.

## Notes for the reviewer

- `#7733` (runtime `PUT /meta` materialization, same package) is untouched. This change is confined to the read side; the metadata-subscription path writes rows through `upsertDeclaredEmailTemplate` and is not on this diff.
- A no-locale send still renders with `renderOpts.locale` unset (the runtime locale) even though the row resolved is now reliably en-US. That is pre-existing `{{ ts | datetime }}` behaviour and no contract speaks to it, so it is left alone rather than changed on the way past.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BsLtjETXfWzP78bMxkyXmB)_

### DIFF
--- a/.changeset/email-template-default-locale.md
+++ b/.changeset/email-template-default-locale.md
@@ -1,0 +1,28 @@
+---
+"@objectstack/plugin-email": patch
+---
+
+fix(plugin-email): a `sendTemplate` with no locale renders the documented en-US default, not an arbitrary row (#7731)
+
+With an i18n bundle in `sys_email_template` — `en-US` and `zh-CN` rows under one
+name — a `sendTemplate` call that named no `locale` rendered **zh-CN**, on two
+consecutive fresh boots. Three declarations say en-US is the answer there
+(`SendTemplateInput.locale`, `EmailTemplateDefinitionSchema.locale`, and
+`sys_email_template`'s own object doc); the code asked the driver instead.
+
+Two seams, both now answering from the contract:
+
+- The `sys_email_template` loader moved out of `EmailServicePlugin` into
+  `createSysEmailTemplateLoader`. Its no-locale branch queries
+  `(name, 'en-US')` by name rather than `{ name }` unordered with `limit: 1`,
+  so no driver's row order can change the answer. Every query it issues carries
+  an `orderBy`, so duplicate rows for one locale resolve the same way on every
+  boot too.
+- `EmailService.sendTemplate`'s ladder asks for `DEFAULT_TEMPLATE_LOCALE` when
+  the caller named no locale — the en-US fallback used to run only when a
+  locale *had* been named, so the no-locale path never reached it.
+
+A bundle with no en-US row at all (a single-locale tenant) keeps rendering:
+the lowest locale tag in the bundle is used, ordered rather than arbitrary.
+Explicit locales are unchanged — exact match, then en-US. Language-only prefix
+matching (`zh` → `zh-CN`) is still not performed; no contract declares it.
--- a/.changeset/email-template-default-locale.md
+++ b/.changeset/email-template-default-locale.md
@@ -1,5 +1,6 @@
 ---
 "@objectstack/plugin-email": patch
+"@objectstack/spec": patch
 ---
 
 fix(plugin-email): a `sendTemplate` with no locale renders the documented en-US default, not an arbitrary row (#7731)
@@ -26,3 +27,7 @@ A bundle with no en-US row at all (a single-locale tenant) keeps rendering:
 the lowest locale tag in the bundle is used, ordered rather than arbitrary.
 Explicit locales are unchanged — exact match, then en-US. Language-only prefix
 matching (`zh` → `zh-CN`) is still not performed; no contract declares it.
+
+`SendTemplateInput.locale` (spec, doc-comment only) now spells the whole ladder
+out, including that last rung — behaviour a caller can rely on has to be
+declared where the contract is, not only where it is implemented.
--- a/packages/plugins/plugin-email/src/email-plugin.ts
+++ b/packages/plugins/plugin-email/src/email-plugin.ts
@@ -17,8 +17,8 @@ import {
   type EmailPersistence,
   type EmailQueueDelivery,
   type TemplateLoader,
-  type EmailTemplateRow,
 } from './email-service.js';
+import { createSysEmailTemplateLoader } from './template-loader.js';
 import {
   makeTransport,
   SmtpTransport,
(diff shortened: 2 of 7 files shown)


huangyiirene:
CI note: the `Test Core (1/3)` red here is `spec#test` `api-methods-batch-conformance` — the **base-branch red tracked in #7793** (`sys_api_key` gained `update` without `bulk` via #7769 on main; this PR's merge ref inherits it). Failing on the base, zero relation to this diff — ESLint / TypeScript / the other Test Core shards are green. Holding the ready-flip; will update the branch and land once #7793's stopgap merges.

---

Issue #3842, opened 2026-07-28
The dispatcher puts the HTTP status in `error.code` and parks the real code in `details` — pinned in #3687, still unfixed


os-zhuang:
The second "Related, not the same" item from #3689, filed on its own now that #3837 has closed that issue. Blocked on #3841 (which vocabulary the code should be written in) — fixing this one first would mean picking a dialect by accident.

## The drift

`HttpDispatcher.error()` (`packages/runtime/src/http-dispatcher.ts:451`) takes the HTTP status as its `code` parameter and writes it straight into the field `ApiErrorSchema` reserves for a semantic string:

```ts
private error(message: string, code: number = 500, details?: any) {
    return {
        status: code,
        body: { success: false, error: { message, code, details } }
    };
}
```

`ApiErrorSchema` (`packages/spec/src/api/contract.zod.ts`) declares:

```ts
code: z.string().describe('Error code (e.g. validation_error)'),
```

So `error.code` comes back as `400`/`403`/`503` — a number, duplicating `response.status`, and occupying the field a caller is meant to branch on.

**The dispatcher already works around its own field being taken**, which is the tell that the number is in the wrong place:

| Site | Real code goes to |
|---|---|
| `http-dispatcher.ts:679` | `details.code` — `this.error(gate.message, 403, { code: gate.code })` |
| `http-dispatcher.ts:1379` | `details.code` — `{ code: 'PERMISSION_DENIED', …e.details }` |
| `routeNotFound()` (`:485`) | `error.type` — a *third* spelling, `'ROUTE_NOT_FOUND'`, sibling to the numeric `code` |
| project-membership gate | `details.type` — `'PROJECT_MEMBERSHIP_REQUIRED'` |

Four sites, three different parking spots (`details.code`, `details.type`, `error.type`), because the declared one is full.

## Why it was pinned rather than fixed

#3687 chose to nail the deviation down to exactly one field instead of moving it, and said so in `http-dispatcher.test.ts`:

```ts
const parsed = ApiErrorSchema.safeParse(body.error);
expect(parsed.success).toBe(false);
expect(parsed.error!.issues.map((i) => i.path.join('.'))).toEqual(['code']);
expect((body.error as { code: unknown }).code).toBe(400);
```

`toEqual(['code'])` is deliberately exact: a second deviating field fails it, and so does fixing the dispatcher — at which point the pin should be deleted rather than updated.

[...]

PR #7790, opened 2026-08-11
fix(runtime): route POST /api/v1/mcp/skill to the dispatcher's own 405 branch

### LINKED ISSUE #7649: mcp-http-surface (a): POST /api/v1/mcp/skill returns a hand-rolled 405 envelope, not the standard API error shape
## Symptom

`POST /api/v1/mcp/skill` (a method mismatch — the route is GET-only) answers 405 with the **wrong envelope**.

- **Observed:** a hand-rolled body `{error, code, message, method, path, allowed}`.
- **Expected:** the standard `{success:false, error:{code, message, httpStatus}}` shape with the documented message *"Method not allowed — use GET"* (the 405 branch routed through `buildApiError` by #3842).

## Root cause

`dispatcher-plugin.ts` (`packages/runtime/src/dispatcher-plugin.ts:919`) registers the skill route with `server.get(`${prefix}/mcp/skill`, …)` **only** — unlike `/mcp`, which is mounted for POST+GET+DELETE (the method-keyed `register` at line ~902). Because no POST handler is registered for `/mcp/skill`, a method mismatch never reaches the dispatcher: Hono routes it to `notFound` and the hono adapter's `unmatchedResponse()` answers with its own envelope. The domain's 405 branch (via `buildApiError`, #3842) is **dead code** on this adapter.

Still present on `origin/main` (`dispatcher-plugin.ts:919` is `server.get(...)` only).

## Reproduction

Boot a showcase with the MCP server available, then:

```
curl -X POST http://&lt;host&gt;/api/v1/mcp/skill
```

**Expected** `{success:false, error:{code, message, httpStatus:405}}` / *"Method not allowed — use GET"*. **Actual** `{error, code, message, method, path, allowed}` from `unmatchedResponse()`.

## Source

Extracted from the QA run objectstack-ai/objectstack#7627 (framework 92f26f75, console 6314e87f).

os-help:
Fixes #7649

## Symptom

`POST /api/v1/mcp/skill` answered 405 with a body no other error on this API returns:

```json
{ "error": "Method Not Allowed", "code": "METHOD_NOT_ALLOWED",
  "message": "POST is not supported for /api/v1/mcp/skill. Allowed: GET.",
  "method": "POST", "path": "/api/v1/mcp/skill", "allowed": ["GET"] }
```

instead of the standard `{success:false, error:{code, message, httpStatus}}` envelope
carrying the documented message *"Method not allowed — use GET"*. A client branching on
`error.code` read `undefined`, because `error` was a string.

## Root cause — the 405 branch was never missing

`handleMcpSkillRequest` (`packages/runtime/src/domains/mcp.ts`) has had a 405 branch since
#3842 routed it through `buildApiError`. The defect was one layer above it:
`createDispatcherPlugin` mounted `${prefix}/mcp/skill` for **GET only**.

Since GET is the only method this route *serves*, a GET-only mount reads as correct. But an
unmounted verb never reaches the dispatcher at all — Hono sends it to `notFound`, where the
adapter's `unmatchedResponse()` re-matches the path across verbs and answers 405 with its own
shape. The domain's branch was dead code on this adapter, and the API carried two different
405 envelopes depending on which route you hit.

## The fix

Mount `/mcp/skill` for the same verb set as its sibling `/mcp` (GET + POST + DELETE), so the
mismatch reaches the branch that already exists. **No second 405 implementation is added** —
this PR stops bypassing the first one.

The method set tracks `/mcp`'s deliberately: `server.get/post/delete` are also the three verbs
the plugin's observability Proxy instruments, so a PUT/PATCH mount here would be both wider
than the sibling route and silently un-instrumented. PUT/PATCH therefore stay adapter-owned,
which is correct — a route that does not exist under a verb is `unmatchedResponse()`'s job.
That boundary is pinned by a test rather than left implicit.

## Behaviour change to note

The `Allow` header now reads `GET` — the domain branch's own literal — where the adapter
previously derived `GET, HEAD` from its route table (Hono registers HEAD implicitly beside
every GET). `HEAD /api/v1/mcp/skill` is still served either way. The `method`, `path` and
`allowed` body keys are gone from this route's 405, and `error` is now an object.

`GET /api/v1/mcp/skill` is untouched: same 200, same `text/markdown`, same
`cache-control: no-store`.

## Tests

New: `packages/runtime/src/mcp-skill-method-not-allowed.hono.integration.test.ts` — a **real
Hono server** driven over real `fetch`, asserting the envelope field by field.

Shaped that way on purpose. The existing coverage
(`http-dispatcher.mcp.test.ts`, "405s non-GET with an Allow header") calls
`dispatcher.handleMcpSkill('POST', …)` **directly**, so it cannot observe a defect that lives
in the HTTP mount — and it asserts only status + `Allow`. Both were already green while the
bug shipped. **A status-only assertion passes in both worlds**, since the status was always
405; only the body differed.

Also added: a registration assertion in `dispatcher-plugin.routes.test.ts`, beside the
existing `/mcp` one.

### Reverse verification

Reverting only the source fix and keeping the tests, with the expected direction stated first:

| Test | Predicted | Observed |
|---|---|---|
| POST envelope fields | red | red — `body.success` is `undefined` |
| POST is not `unmatchedResponse()`'s shape | red | red — `typeof body.error` is `'string'` |
| DELETE envelope fields | red | red — `body.success` is `undefined` |
| route registration | red | red — 54 routes, no `POST /api/v1/mcp/skill` |
| `Allow` header | *unchanged* | **red** — `GET, HEAD` vs `GET` (see below) |
| GET happy path | green | green |
| PUT falls to the adapter | green | green |

The `Allow` row is the one prediction that was wrong, and it is reported rather than
smoothed over: the header does not merely survive the fix, it **changes value**. Pre-fix the
adapter derived `GET, HEAD` from its route table; post-fix the domain branch's literal `GET`
answers. The test name and the changeset were corrected to state that transition instead of
claiming the header is untouched.

## Checks

- `pnpm --filter @objectstack/runtime test` — **130 files, 2036 tests passed**
- `pnpm --filter @objectstack/http-conformance test` — **4 files, 72 tests passed** (the
  cross-adapter `IHttpServer` unmatched-request contract)
- `pnpm --filter @objectstack/runtime typecheck` — clean (`tsc --noEmit`)
- `eslint --no-inline-config` on the three changed files — clean
- `node scripts/check-nul-bytes.mjs` (+ `--self-test`) — clean

A changeset is included (`@objectstack/runtime`, patch); `content/docs/releases/` is untouched.

---

### DIFF
--- a/.changeset/mcp-skill-method-not-allowed-envelope.md
+++ b/.changeset/mcp-skill-method-not-allowed-envelope.md
@@ -1,0 +1,39 @@
+---
+"@objectstack/runtime": patch
+---
+
+fix(runtime): `POST /api/v1/mcp/skill` answers the standard error envelope, not the adapter's hand-rolled 405 (#7649)
+
+A method mismatch on the public SKILL.md route returned a body no other error on
+this API returns:
+
+```json
+{ "error": "Method Not Allowed", "code": "METHOD_NOT_ALLOWED",
+  "message": "POST is not supported for /api/v1/mcp/skill. Allowed: GET.",
+  "method": "POST", "path": "/api/v1/mcp/skill", "allowed": ["GET"] }
+```
+
+instead of the standard `{success:false, error:{code, message, httpStatus}}`
+carrying the documented message *"Method not allowed — use GET"*. A client
+branching on `error.code` read `undefined`, because `error` was a string.
+
+**The 405 branch was never missing.** `handleMcpSkillRequest` has had one since
+#3842 routed it through `buildApiError`. The defect was one layer above it:
+`createDispatcherPlugin` mounted `${prefix}/mcp/skill` for **GET only**. Since
+GET is the only method the route serves, that read as correct — but an unmounted
+verb never reaches the dispatcher at all. Hono sends it to `notFound`, where the
+adapter's `unmatchedResponse()` re-matches the path across verbs and answers 405
+with its own shape. The domain's branch was dead code on this adapter, and the
+API had two 405 envelopes depending on which route you hit.
+
+`/mcp/skill` is now mounted for the same verb set as its sibling `/mcp`
+(GET + POST + DELETE), so the mismatch reaches the branch that already exists.
+No new 405 logic was written, and `GET /api/v1/mcp/skill` is untouched — same
+200, same `text/markdown`, same `cache-control: no-store`.
+
+Note for callers that parse the old body: the `method`, `path` and `allowed`
+keys are gone from this route's 405, and `error` is now an object. The `Allow`
+response header remains the interoperable place to read the hint, and now
+reads `GET` — the domain branch's own literal — where the adapter previously
+derived `GET, HEAD` from its route table (Hono registers HEAD implicitly
+beside every GET). `HEAD /api/v1/mcp/skill` is still served either way.
--- a/packages/runtime/src/dispatcher-plugin.routes.test.ts
+++ b/packages/runtime/src/dispatcher-plugin.routes.test.ts
@@ -59,6 +59,27 @@ describe('createDispatcherPlugin — HTTP route registration', () => {
     expect(routes).toContain('POST /api/v1/keys');
   });
 
+  // Regression (#7649): /mcp/skill was mounted for GET ONLY. The route serves
+  // GET and nothing else, so that looked right — but the dispatcher owns a 405
+  // branch for the other verbs ("Method not allowed — use GET", built through
+  // `buildApiError` since #3842), and an unmounted verb never reaches it: Hono
+  // sends it to `notFound`, where the adapter's `unmatchedResponse()` answers
+  // 405 with its own hand-rolled `{error, code, message, method, path, allowed}`
+  // body. Same status, different envelope, and the domain branch dead code.
+  // Mounting the verbs is what routes the mismatch to the branch that exists.
+  // The envelope itself is pinned end-to-end in
+  // `mcp-skill-method-not-allowed.hono.integration.test.ts` — a status-only
+  // assertion cannot see this defect.
+  it('mounts /mcp/skill for the same verbs as /mcp so a method mismatch reaches the dispatcher 405', async () => {
+    const { server, routes } = makeFakeServer();
+    const plugin = createDispatcherPlugin({ prefix: '/api/v1', securityHeaders: false });
+    await plugin.start?.(makeCtx(server));
+
+    expect(routes).toContain('GET /api/v1/mcp/skill');
+    expect(routes).toContain('POST /api/v1/mcp/skill');
+    expect(routes).toContain('DELETE /api/v1/mcp/skill');
+  });
+
   // Regression (framework #2217 seam #2): /ready shipped with a dispatch()
   // branch but NO server.<verb>() registration, so it 404'd over HTTP before
   // reaching the handler — the same class of bug as /mcp and /keys. /health and
--- a/packages/runtime/src/dispatcher-plugin.ts
+++ b/packages/runtime/src/dispatcher-plugin.ts
@@ -916,14 +916,38 @@ export function createDispatcherPlugin(config: DispatcherPluginConfig = {}): Plu
             // Public SKILL.md download (env-customized portable Agent Skill).
             // Separate registration: `/mcp` above is an exact-path mount, so
             // the sub-path needs its own route to be reachable over HTTP.
-            server.get(`${prefix}/mcp/skill`, async (req: any, res: any) => {
-                try {
-                    const result = await dispatcher.dispatch('GET', '/mcp/skill', req.body, req.query, { request: req });
-                    sendResult(result, res);
-                } catch (err: any) {
-                    errorResponse(err, res);
-                }
-            });
+            //
+            // [#7649] Mounted for the SAME method set as `/mcp` above rather
+            // than GET alone, even though GET is the only method this route
+            // SERVES. The domain owns a 405 branch for the rest
+            // (`handleMcpSkillRequest`: "Method not allowed — use GET", body
+            // built through `buildApiError` per #3842) — but a branch can only
+            // answer a mismatch that REACHES the dispatcher. With GET as the
+            // sole registration, Hono routed `POST /api/v1/mcp/skill` to
+            // `notFound`, where the adapter's `unmatchedResponse()` answered
+            // with its own `{error, code, message, method, path, allowed}`
+            // shape: a second, non-standard 405 envelope on the wire, and the
+            // domain branch dead code on this adapter. Registering the verbs
+            // hands the mismatch to the branch that already exists.
+            //
+            // The method set tracks `/mcp`'s deliberately: `server.get/post/
+            // delete` are also the three verbs the observability Proxy above
+            // instruments, so a PUT/PATCH mount here would be both wider than
+            // the sibling route and silently un-instrumented.
+            const mountMcpSkill = (method: 'GET' | 'POST' | 'DELETE') => {
+                const register = method === 'GET' ? server.get : method === 'DELETE' ? server.delete : server.post;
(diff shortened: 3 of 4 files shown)


os-help:
PM review note (dispatching seat, #6024). Two things checked so the next reader does not have to repeat them.

**Docs-drift advisory: resolved, no doc change needed.** The bot listed 20 hand-written pages via `@objectstack/runtime`. The only ones that could plausibly pin a 405 envelope are `protocol/kernel/http-protocol.mdx` and `api/wire-format.mdx`; neither documents this route's 405 shape (`http-protocol.mdx`'s only 405 is `OBJECT_API_METHOD_NOT_ALLOWED`, an unrelated `enable.apiMethods` branch). The generated code-listing pages carry code names without statuses. Nothing goes stale from this PR.

**The `Allow` header change is accepted, but it is load-bearing and worth stating explicitly.** `content/docs/api/error-catalog.mdx` documents `METHOD_NOT_ALLOWED` as:

> **Fix:** Use the method named in the response's `Allow` header.

So `Allow` is the *documented* client remedy for this exact error, not decoration. This PR narrows it from `GET, HEAD` to `GET` on this route. That is accepted here — the documented guidance stays accurate (a client following it reaches for GET and succeeds), `HEAD` is still served, and envelope consistency across the API is the larger win. But per RFC 9110 the field advertises the methods the resource supports, and HEAD is one of them, so this is a small accuracy loss, not a neutral swap.

Recording it rather than leaving it in the PR body alone, because the reverse-verification table shows this was the one prediction that came back wrong — the header does not survive the fix unchanged, it changes value. Good catch reporting that instead of smoothing it over; that is exactly the failure mode a "predicted green, observed green" table is prone to hiding. If a follow-up wants `Allow: GET, HEAD`, the fix is in the domain branch's literal, and it should come with the reasoning above rather than as a drive-by.

No changes requested. Leaving this to finish CI.

---

os-help:
Pushed `5f069e6cd` — fixes the red **TypeScript Type Check** gate (`check:type-check-debt --re-measure`), which measured `@objectstack/runtime` at 238 against a recorded TEST_DEBT of 227.

**Why the PR's own typecheck line was clean and still wrong.** `packages/runtime/tsconfig.json` excludes `**/*.test.ts`, so `pnpm --filter @objectstack/runtime typecheck` never compiled the new file. The gate re-runs tsc with that exclusion dropped. My check line was true but measured a layer that structurally could not contain the new test.

**The +11, all from the new file** (the `dispatcher-plugin.routes.test.ts` edit contributed 0):

| Count | Code | What |
|---|---|---|
| 10 | TS18046 | `'body' is of type 'unknown'` — `Response.json()` returns `unknown` |
| 1 | TS2353 | `'requireAuth' does not exist in type 'DispatcherPluginConfig'` |

The `requireAuth: false` was copied from a sibling suite and is a **dead key**: nothing reads it (the deployment-wide gate was removed — `dispatcher-plugin.ts:209`, and `http-dispatcher.requireauth.test.ts:56` states "There is no `requireAuth: false` any more"). It was silently ignored, so dropping it is a no-op; the GET positive control still passes, which is the proof.

**No assertion was weakened.** The 10 `unknown` reads now go through one `call()` helper that casts once to an open record — deliberately *not* a narrow interface, because this suite exists precisely because two different body shapes can arrive on this path, and one case asserts keys that must **not** exist. A type admitting only the correct envelope would encode the conclusion under test and make the negative case unwritable. Every envelope check remains a runtime assertion on the same field.

**Reverse verification re-run after the refactor** — byte-identical to before: 5 failed / 16 passed, same cases, same messages (`expected undefined to be false`, `expected 'string' not to be 'string'`, `expected 'GET, HEAD' to be 'GET'`, `expected [ …(54) ] to include 'POST /api/v1/mcp/skill'`), GET and PUT-boundary still green.

**Verification of the fix round** (full workspace closure built first, exactly as `lint.yml` does):

- `pnpm check:type-check-debt` → `--self-test` OK, then `--re-measure: OK — 33 ledger entries re-measured, none above its recorded number`. `@objectstack/runtime` measures **227**, equal to the ledger; it is absent from the "can be lowered" hints.
- `pnpm --filter @objectstack/runtime test` → 130 files, 2036 tests passed
- `eslint --no-inline-config` on the changed file → exit 0

The TEST_DEBT entry was **not** raised. Also dropped a no-op `.replace('http://', 'http://')` in the GET control — same string either way.

---

---

os-help:
**Auto-merge temporarily disabled — deliberately, to stop this PR burning queue rebuilds it cannot pass. Blocked on #7802 / #7818, not on anything in this diff.**

Two queue builds have now failed here ([31509570177](https://github.com/objectstack-ai/objectstack/actions/runs/31509570177), [31511209233](https://github.com/objectstack-ai/objectstack/actions/runs/31511209233)), both on the same test in a package this PR never touches:

```
FAIL packages/spec/src/data/api-methods-batch-conformance.test.ts
  > grants bulk wherever it grants create / update / delete
```

That is #7802: `main` is red because #7769 gave `sys_api_key` `update` without `bulk`. The fix is #7818, which is green on the shard that carries this test and is waiting to merge.

**Answering the triage bot's history signal directly**, since it is pointing at the wrong conclusion: it flags "failed twice in 24h with unchanged content ⇒ suspect a flaky test". It is not flaky here. It is clause 2 of the same checklist — a failure in a package this PR does not touch — and the cause is identified, reproducible and already fixed in an open PR. Re-queueing cannot help until #7818 lands, and each attempt rebuilds every PR behind it in the batch.

**Why disable rather than just not re-queue.** Something is re-adding this PR to the queue on its own, so "I will not re-queue it" was not sufficient — it went back in and burned a second full build without my touching it. Disabling auto-merge is the only lever that actually stops the cycle. It will be re-enabled the moment #7818 is on `main`.

**A note for anyone reading this PR's own CI history:** this PR's PR-side checks are green, because the affected-subset filter never runs `packages/spec` for a diff confined to `packages/runtime`. The queue runs the full suite, which is where a red `main` reaches every PR regardless of what it touches. I said earlier in this round that this PR was "not affected by #7802 because it does not touch spec" — that was wrong, and this is the correction: that reasoning only ever held for the PR-side gate.

Nothing is requested of this diff. The TEST_DEBT type errors that failed the earlier `TypeScript Type Check` were fixed and pushed; that gate is green now.

---

os-help:
**Converted back to draft — deliberately and temporarily. Nothing is wrong with this diff; it will be marked ready again the moment #7818 is on `main`.**

Escalating as stated: one comment ago I said that if a fourth queue failure arrived, disabling auto-merge had not been enough to get this PR out of the queue and I would remove it rather than explain again. [31512974833](https://github.com/objectstack-ai/objectstack/actions/runs/31512974833) is the fourth. Draft conversion drops both queue membership and auto-merge, and nothing can re-add it automatically — that is the only lever left that actually holds.

**The reason this became urgent rather than merely wasteful.** The queue tests the *merged result*. While this PR sits ahead of #7818 in the queue, GitHub is testing `main + #7790` — a combination that does not contain the fix for #7802 and therefore fails the conformance test every single time, by construction. So this PR was not just burning full-suite rebuilds for everything behind it; it was repeatedly failing batches in front of **the one PR that unblocks the queue for everyone**. Pulling it out clears the path for its own unblocker.

**To the triage bot's escalating signal** (now "failed 3 times in 24h, unchanged content ⇒ strongly suspect flaky"): the count is right, the inference is still wrong, and it will keep climbing for as long as this PR is in the queue. Not flaky, not a semantic conflict with a co-queued PR — clause 2, a failure in `packages/spec`, which this diff does not touch. Cause identified, deterministic, fix open at #7818.

Restore checklist, for whoever gets here first:

1. #7818 merges → `main` green.
2. Mark this PR ready for review.
3. Re-enable auto-merge (squash).

No changes to the diff are needed at any point. PR-side CI was green before this, including the `TypeScript Type Check` that the TEST_DEBT fix addressed.

---

Issue #7663, opened 2026-08-11
QA run · identity-auth (FULL area) · 92f26f75 · 2026-08-11 · 5 PASS / 4 PARTIAL / 3 FAIL / 2 BLOCKED


baozhoutao:
Full `identity-auth` area run of the `checklist-test` skill — all **14 items** (12 runnable + 2 ledger-blocked) driven against a live showcase across 3 opus subagents, each on its own port + file DB. Text-only per RUNNER.md.

**Result: 5 PASS · 4 PARTIAL · 3 FAIL · 2 BLOCKED.**

**Environment** — framework `92f26f75`/`f268b2b2` (branch `claude/platform-test-checklist-ocwugl`) · console bundle `09987b680` (`packages/console/dist/.objectui-sha`) · showcase app · isolated port + file DB per batch · personas provisioned as **real sign-ups and real in-product role writes**, not stubs.

> **Environment fact that shaped this whole area, stated up front:** the seeded dev admin is promoted to the ObjectStack **platform** admin, but its **better-auth `sys_user.role` stays `'user'`** — so the entire better-auth-native admin family (`list-users`, `ban/unban`, `set-role`, `remove-user`, `revoke-user-sessions`, `impersonate-user`) 403s out of the box. Framework-overridden endpoints (`create-user`, `set-user-password`, `import-users`) work regardless. Rather than record six `blocked`, the runner provisioned a genuine admin with one direct write to its **own throwaway DB** — the same system-context write the pinned dogfood harness performs — and judged the clauses on merit.

## ✅ PASS — 5 items

- **`self-service-password-reset`** (5/5 + all four negatives) — known and unknown addresses return **byte-identical** bodies (asserted by string comparison, not by eye) *and* `sys_email` holds 1 artifact for the known address / **0** for the unknown, so nothing is sent for a non-account while the HTTP response stays constant. Token reuse → 400 and the password it tried to set **never became valid**.
- **`self-service-profile-password`** (6/6 + four negatives) — name and avatar persist across a full sign-out/sign-in **into a new browser context**; the avatar is a genuine presign/PUT/commit cycle, not a `data:` URL. The item's `knownGap` was **closed by provisioning** a passwordless persona (deleting the credential `sys_account` row), proving the `set-initial-password` branch works and 409s in both directions.
[...]

Issue #7525, opened 2026-08-11
[rest] Hook refusals carrying an explicit `statusCode` are not mapped by `/api/v1/data` — they leak as HTTP 500 `INTERNAL_ERROR` with no located guidance


huangyiirene:
## Symptom

A hook that refuses a write **with an explicit `statusCode`** does not get that status back out through `/api/v1/data`. The refusal is swallowed and re-emitted as an opaque **HTTP 500 `INTERNAL_ERROR`** carrying no code, no located guidance, and nothing the caller can act on — while the server log holds the real, correctly-shaped refusal.

Two independent cases, **each reproduced 2×**:

**(a) Record lock.** `PATCH` a record while a `lockRecord:true` approval is pending → **500**, while the log holds `RECORD_LOCKED`. Expected: the refusal's own status with the `RECORD_LOCKED` code in the envelope.

**(b) Delegation ownership.** `POST sys_approval_delegation` with `delegator_id` not equal to the authenticated caller → **500**, while the log holds `FORBIDDEN … statusCode 403`. Expected: **403 `FORBIDDEN`**.

This is cross-cutting: neither case is approvals-specific in its fix face. The approvals hooks are simply two well-behaved producers that happen to declare a status, and the mapping layer drops it on the floor — so **any** hook that refuses with a status is affected.

The console never hits case (a), because the affordance is correctly disabled when a lock is live. A direct API caller — script, integration, second-party client — gets an unactionable 500 for what is a deliberate, well-understood refusal.

**Same family as the raw-`TypeError`-in-envelope defect from the api-backend run, objectstack-ai/objectstack#7463** (defect 7 there: `POST /api/v1/data/showcase_task` with a number in a declared `text` field → `400 {"error":"TypeError: not a function"}`, no `code`, no `fields[]`). Both are the `/api/v1/data` error boundary failing to turn a known producer signal into the ledgered envelope. Related but not the same: objectstack-ai/objectstack#5582 (closed) covered `mapDataError`'s explicit-status passthrough only honouring 4xx for **service**-declared statuses; the hook-refusal path evidently does not reach that passthrough at all.

## Root cause

Not located in this run — the QA oracle is live HTTP plus the server log. The evidence localises it to the `/api/v1/data` error boundary: the hook throws with a `statusCode` (visible in the log line), and the route's error mapping neither reads the property nor maps the code, falling through to the generic 500 arm. Both `mapDataError` and the hook-dispatch seam are the places to look.

## Reproduction

[...]

Issue #5949, opened 2026-08-06
[finding] ADR-0076 D11 的第二半从未落地:`packages/rest/src/rest-server.ts` 已 8593 行(ADR 记录约 5.1k),且无 issue 承接


os-zhuang:
observation-class:今天没有用户会踩到,纯结构性欠账 + 记录承接问题。由 #5357(D11 锚点入账,PR #5940)派生登记 —— 该 issue 随 PR 关闭后,这条观察的唯一书面记录就没了。

## 事实

ADR-0076 D11 点名了**两个**中心路由生成器要拆:

1. `packages/runtime/src/http-dispatcher.ts`(ADR 记录约 3.8k 行)—— **已拆**:14 个域体迁至 `packages/runtime/src/domains/*`,路由走 `DomainHandlerRegistry`,dispatcher 收缩到约 1.9k 行,多适配器由 `packages/qa/http-conformance` 验证。
2. `packages/rest/src/rest-server.ts`(ADR 记录约 5.1k 行)—— **未动**,而且在长大:

| 时点 | 行数 | 出处 |
|---|---|---|
| ADR-0076 rev.9 记录 | 约 5100 | `docs/adr/0076-objectql-core-tiering.md` D11 决定正文 |
| 2026-08-05 | 7693 | #5357 正文顺带记录 |
| 2026-08-06(本次实测,`origin/main` a6b3ee7a1) | **8593** | `wc -l packages/rest/src/rest-server.ts` |

即一天内 +900 行。ADR 的状态行本身已如实记录这一半(「D11 — substantially implemented; **the rest-server half is untouched**. … is now ~7.7k LOC, i.e. larger than the ~5.1k figure below, not smaller」),所以决定与现状之间没有隐瞒 —— 缺的是**承接**:实测当前 open issue 里没有任何一条跟踪这半边(按 open issue 全量扫描 title/body 关键字 `rest-server`,仅 #5936 命中且是另一件事:`/discovery` 第二生产者的 environment 广播)。

## 为什么单独登记而不是并进锚点 PR

#5357 的分诊评论明令「正文末段顺带记的 `rest-server.ts` 7693 行问题不属本单验收,勿顺手带」,PR #5940 因此严格只做锚点 + 注释。但 #5357 关闭时这条观察会随之消失,所以按 Prime Directive #10 单独落一条不指派的 finding。

## 建议(留给分诊,不在此裁决)

[...]

PR #7791, opened 2026-08-11
fix(spec): the external-federation error family declares its HTTP status, so a write refusal stops leaking as a bare 500 (#7739)

### LINKED ISSUE #7739: external-datasource-federated-read: external-write refusal (ExternalWriteForbiddenError) leaks to the client as a bare 500 INTERNAL_ERROR
## Symptom

A write to a read-only federated external object is correctly refused server-side — `ExternalWriteForbiddenError` names the datasource, its `schemaMode`, and both flags that would be required — but it reaches the client as a bare **500 INTERNAL_ERROR**, indistinguishable from a crash. Nothing is applied (the remote table is byte-identical after the attempt). Same envelope-leak class as the auth bodyless-500s in #7663 and the earlier raw-`TypeError` leak in #7543 (and the 5xx driver-throw passthrough family #5437).

## Root cause

`ExternalWriteForbiddenError` (defined in `packages/spec/src/shared/external-errors.ts`, thrown in `packages/objectql/src/engine.ts`) carries a stable `code` (`EXTERNAL_ERROR_CODES.writeForbidden`), but no REST/runtime error map references it: a `git grep` for `writeForbidden` / `ExternalWriteForbidden` / `EXTERNAL_ERROR_CODES` across `packages/rest`, `packages/runtime`, and `packages/cli` finds no mapping. So the error falls through the generic handler in `packages/rest/src/rest-server.ts` to 500 INTERNAL_ERROR instead of a 4xx envelope that carries the `code` and the refusal detail the error already computed.

Stale-premise check: present on `origin/main` (no external-error mapping exists in the rest/runtime layer).

## Reproduction

1. Boot with a read-only federated external object (`schemaMode` that forbids writes).
2. `POST /api/v1/data/&lt;external_object&gt;` with any body.

Expected: a 4xx envelope with a stable `code` naming the write-forbidden refusal. Actual: 500 INTERNAL_ERROR with no `code`, indistinguishable from a crash, even though the server-side `ExternalWriteForbiddenError` was correct and nothing was applied.

## Source

os-help:
Fixes #7739

## The defect

A write to a read-only federated external object was refused **correctly** on
the server. `ExternalWriteForbiddenError` names the datasource, its
`schemaMode`, and both flags that would be required, and nothing is applied —
ADR-0015's Gate 3 throws before the driver is reached at all. Only the envelope
was wrong: the refusal reached the client as a bare **500 INTERNAL_ERROR with
no `code`**, indistinguishable from the server falling over.

## Root cause — the status, not the code

The `code` existed the whole way down. `EXTERNAL_WRITE_FORBIDDEN` is registered
in the ADR-0112 error-code ledger (`error-code-ledger.zod.ts`, under
`@objectstack/spec`), so the wire vocabulary was ready.

What the error did not carry was an HTTP **status**, and no exit can invent one.
So the refusal fell past every structured branch of `mapDataError`, matched no
message heuristic, and left through the terminal `UNCLASSIFIED_FAULT` — which
sanitises to `500` and, by design, ships no `code` the producer never declared.
The discriminator the client needed was dropped at the boundary for want of a
status to carry it.

## The fix: the family declares its status, at the producer

`packages/spec/src/shared/external-errors.ts` gains one table beside the codes
it keys on, and each of the three error classes carries its value as `status`:

```ts
export const EXTERNAL_ERROR_HTTP_STATUS = {
  [EXTERNAL_ERROR_CODES.schemaMismatch]: 503,
  [EXTERNAL_ERROR_CODES.writeForbidden]: 403,
  [EXTERNAL_ERROR_CODES.schemaModeViolation]: 403,
} as const satisfies Record< ExternalErrorCode, number >;
```

**Why at the producer rather than in a REST error map.** ADR-0112 says the
producer names the condition, and every HTTP exit in this repo already agrees on
how to read one — `status` then `statusCode`, band 400-599:
`mapDataError`'s `declaredHttpStatus` (#7525), `resolveErrorResponse`
(#5437/#5582), `HttpDispatcher.errorFromThrown` (#3867),
`dispatcher-plugin.errorResponseBase`, `endpoint-executor`, `domains/actions`,
`plugin-hono-server`. Declaring the status on the error is therefore what routes
the family through **one place for every door at once**; a branch in
`rest-server.ts` would have fixed one door and left the runtime dispatcher, the
endpoint executor and the CLI answering 500 for the same throw.

Same shape the repo already uses for this problem: `service-analytics`'s
`dataset-refusal.ts` (#5367) and `storage-service.ts`'s `storageListRefusal` —
"one condition, one wire shape, chosen by the producer that knows".

**`rest-server.ts` is untouched.** #5949 records it at 8593 lines and a frequent
conflict site; this change adds zero lines to it.

## Family mapping — and the status convention followed

The card asked for the family, not the instance, and for the 4xx choice to be
justified against the surrounding conventions.

| code | status | reasoning |
| --- | --- | --- |
| `EXTERNAL_WRITE_FORBIDDEN` | **403** | A **policy** refusal, not malformed input. The identical body succeeds the moment `datasource.external.allowWrites` and `object.external.writable` are both on, so 400/422 ("fix your request") would be a lie, and 409 ("conflict with current state") promises a retry that cannot help. 403 is what `mapDataError` already answers for a capability a flag switched off for the object — `FEEDS_DISABLED`, `FILES_DISABLED`, `CLONE_DISABLED`, `RECORD_NOT_ACCESSIBLE` — and what the standard catalog spells `FORBIDDEN`. |
| `EXTERNAL_SCHEMA_MODE_VIOLATION` | **403** | The same sentence about DDL rather than rows: `schemaMode !== 'managed'` forbids it, and no request the caller can rewrite changes that. |
| `EXTERNAL_SCHEMA_MISMATCH` | **503** | Deliberately **not** 4xx. Nothing about the request is wrong — the deployment's metadata and the remote table have diverged, only an operator can reconcile them, and it may clear. That is the reading `ERR_DATASOURCE_UNAVAILABLE` already gets ("the deployment cannot serve this object right now"); both are `isExpectedDataStatus` lifecycle outcomes rather than crashes. The 5xx band withholds the message by design, which is right here: this gate aborts boot, so its structured `diffs` audience is already operator-side, and the client still gets a `code` to branch on instead of a bare 500. |

So the **dispatch** is one mapping; only the values differ, and the one that
differs is the one that is genuinely not a client error.

`satisfies Record< ExternalErrorCode, number >` makes a future gate that adds a
code without a status a **compile error** — which is the half that stops the
next QA run from refiling this card under a different code.

## Verification

`packages/rest/src/external-write-forbidden-envelope.test.ts` (new, 9 cases)
drives the **real** CRUD routes in process against a protocol whose data verbs
are a **real `ObjectQL` engine** over an in-memory driver whose store the test
can read — not a rejecting mock.

That fixture choice is load-bearing. The card asks for two facts a
`mockRejectedValue` cannot pair: the refusal answers 403 with its `code`, **and
nothing is applied**. A mock that rejects makes the second vacuously true, so it
would stay green against a future "fix" that got the status right by letting the
write through. Here §3 measures the store against a driver that **can** write,
and a CONTROL case flips the two opt-in flags and writes through the very same
route to prove the fixture is not inert. §3 also asserts the driver was never
reached at all, so a create-then-rollback "fix" fails even with an empty table.

`packages/spec/src/shared/external-errors.test.ts` gains the family pin: the map
covers exactly `EXTERNAL_ERROR_CODES`, every status is inside the 400-599 band
each exit reads, and each instance carries it as `status`.

### Reverse verification

Direction predicted **before** running: RED on the unfixed producer, for the
reason the card describes. Measured by reverting `external-errors.ts` to
`origin/main`, **rebuilding `@objectstack/spec`** (the test imports `dist`, so
reverting the source alone would have proven nothing) and re-running:

```
Tests  7 failed | 2 passed (9)
AssertionError: expected 500 to be 403 // Object.is equality
```

- §1 status / code / message-detail / object: RED — `500`, and `res.body.code`
  `undefined`
- §1 "not logged as an unhandled fault": RED, `expected true to be false` — the
  500 did emit `[REST] Unhandled error`, which is the operator-side half of the
  same symptom
- §2 PATCH and DELETE: RED, same `expected 500 to be 403`
- §3 POST "nothing applied": RED — but only on its leading status assertion;
  the store assertions in it were green before and after
- **The 2 survivors are direction-insensitive by construction**, exactly as
  predicted: §3's "PATCH and DELETE reach the driver no more than POST does"
  (nothing was ever applied — the behaviour was never the bug) and §3's CONTROL
  case (the write-allowed arm, which this change does not touch). Recorded as
  measured rather than reshaped: a template demanding before-red on every
  assertion would have had those two cases lie about what they measure. They
  are asserted anyway because they are the half that stops a status-only "fix"
  from passing.

## Scope

Not touched: #7663 (auth bodyless-500s), #7543 (raw `TypeError` leak), #5437
(5xx driver-throw passthrough) — same envelope-leak class, deliberately out of
scope. This change does not make them easier to close: each has a different
cause (a producer that writes no body, a native error name reaching a business
branch, a 5xx passthrough range), and none of them is a missing status
declaration.

`.changeset/external-error-http-status.md` added — the wire status of three
codes is user-visible. `content/docs/releases/` untouched.

### Local runs

| command | result |
| --- | --- |
| `pnpm --filter @objectstack/spec test` (full suite) | **378 files / 9953 tests passed** |
| `pnpm --filter @objectstack/rest test` (full suite) | **92 files / 1471 tests passed** |
| `pnpm --filter @objectstack/rest test external-write-forbidden-envelope` | **1 file / 9 passed** |
| `pnpm --filter @objectstack/objectql test external-write-gate` | 11 passed |
| `pnpm --filter @objectstack/runtime test external-validation` | 13 passed |
| `pnpm --filter @objectstack/spec typecheck` | clean |
| `pnpm --filter @objectstack/rest typecheck` | clean |
| `eslint` on the three changed/added files | clean, exit 0 |
| `pnpm check:api-surface` / `check:export-origins` | current (both baselines regenerated) |
| `pnpm check:error-code-casing` | 3646 files, no lowercase codes |
| `pnpm check:nul-bytes` | 7131 files, clean |

---

### DIFF
--- a/.changeset/external-error-http-status.md
+++ b/.changeset/external-error-http-status.md
@@ -1,0 +1,46 @@
+---
+"@objectstack/spec": patch
+---
+
+fix(spec): the external-federation error family declares its HTTP status, so a write refusal stops leaking as a bare 500 (#7739)
+
+A write to a read-only federated external object was refused correctly on the
+server — `ExternalWriteForbiddenError` names the datasource, its `schemaMode`,
+and both flags that would be required, and nothing is applied (the gate throws
+before the driver is reached) — but it reached the client as a bare
+**500 INTERNAL_ERROR with no `code`**, indistinguishable from the server
+falling over.
+
+The `code` existed the whole way down: `EXTERNAL_WRITE_FORBIDDEN` is registered
+in the ADR-0112 error-code ledger. What the error did not carry was an HTTP
+**status**, and no HTTP exit can invent one — so the refusal fell past every
+structured branch of the REST boundary, matched no message heuristic, and left
+through the terminal sanitised 500 that drops the `code` with it.
+
+**The whole family now declares its status, in one table.** New export
+`EXTERNAL_ERROR_HTTP_STATUS` (`@objectstack/spec/shared`) maps every
+`EXTERNAL_ERROR_CODES` member to the status it is reported with, and each of
+the three error classes carries the corresponding value as `status`:
+
+| code | status | why |
+| --- | --- | --- |
+| `EXTERNAL_WRITE_FORBIDDEN` | **403** | a policy refusal — the identical body succeeds once `datasource.external.allowWrites` and `object.external.writable` are both on, so 400/422 ("fix your request") would be a lie and 409 promises a retry that cannot help |
+| `EXTERNAL_SCHEMA_MODE_VIOLATION` | **403** | the same sentence about DDL: `schemaMode !== 'managed'` forbids it, and no rewritten request changes that |
+| `EXTERNAL_SCHEMA_MISMATCH` | **503** | not the caller at all — the deployment's metadata and the remote table diverged, only an operator can reconcile them, and it may clear (the reading `ERR_DATASOURCE_UNAVAILABLE` already gets) |
+
+Declared at the **producer** rather than in a REST error map, because every
+HTTP exit in the framework already resolves `status` then `statusCode` —
+`mapDataError`'s `declaredHttpStatus`, `resolveErrorResponse`,
+`HttpDispatcher.errorFromThrown`, `dispatcher-plugin.errorResponseBase`,
+`endpoint-executor`, `domains/actions`, `plugin-hono-server`. One table fixes
+every door at once; a branch in the REST server would have fixed one and left
+the runtime dispatcher, the endpoint executor and the CLI answering 500 for the
+same throw. `satisfies Record<ExternalErrorCode, number>` makes a future gate
+that adds a code without a status a compile error, so the family cannot
+silently re-acquire a member that leaks as a 500.
+
+**What callers see now.** `POST /api/v1/data/<external_object>` against a
+read-only federated object answers `403` with `code:
+"EXTERNAL_WRITE_FORBIDDEN"` and the refusal detail naming both flags to set,
+instead of `500` with no `code`. Behaviour is unchanged — nothing was applied
+before and nothing is applied now; only the envelope was wrong.
--- a/packages/rest/src/external-write-forbidden-envelope.test.ts
+++ b/packages/rest/src/external-write-forbidden-envelope.test.ts
@@ -1,0 +1,358 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+//
+// [#7739] A write refused by ADR-0015's Gate 3 reaches the client as a
+// `403 EXTERNAL_WRITE_FORBIDDEN` envelope, not as a bare `500 INTERNAL_ERROR`.
+//
+// ---------------------------------------------------------------------------
+// The defect, and where it actually lived
+//
+// `ExternalWriteForbiddenError` was already correct on every axis QA could see
+// from the server side: it named the datasource, its `schemaMode`, and both
+// flags that would be required, and it carried a ledgered ADR-0112 `code`
+// (`EXTERNAL_WRITE_FORBIDDEN`, registered under `@objectstack/spec`). Nothing
+// was applied — the gate throws before the driver is reached at all.
+//
+// What it did NOT carry was an HTTP status, and no exit can invent one. So the
+// refusal fell past every structured branch of `mapDataError`, matched no
+// message heuristic, and left through the terminal `UNCLASSIFIED_FAULT` as
+// `500 INTERNAL_ERROR` **with no `code`** — the client could not tell a
+// deliberate policy refusal from the server falling over, even though the
+// discriminator existed the whole way down.
+//
+// The fix is at the PRODUCER (`packages/spec/src/shared/external-errors.ts`),
+// not in a REST error map: the whole `EXTERNAL_ERROR_CODES` family now declares
+// its status in one table, `EXTERNAL_ERROR_HTTP_STATUS`, and every HTTP exit in
+// this repo already reads `status` → `statusCode` (`declaredHttpStatus` here,
+// `resolveErrorResponse`, `HttpDispatcher.errorFromThrown`,
+// `dispatcher-plugin.errorResponseBase`, `endpoint-executor`, `domains/actions`,
+// `plugin-hono-server`). One mapping, every door — and `rest-server.ts`, a
+// known 10k-line conflict site, is untouched by this change.
+//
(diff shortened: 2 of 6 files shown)


os-help:
**CI is red on `Test Core (1/3)`, and it is not this PR. Blocked on `main`, tracked as #7802 — holding rather than re-queueing.**

The failure is in `@objectstack/spec`:

```
FAIL src/data/api-methods-batch-conformance.test.ts:139
+ "sys_api_key: [get, list, update] grants single-record writes but not 'bulk'
+  — /batch and the *Many routes will 405
+  (packages/platform-objects/src/identity/sys-api-key.object.ts)"

Test Files  1 failed | 377 passed (378)
     Tests  1 failed | 9952 passed (9953)
```

This PR does not touch `sys_api_key`, `platform-objects`, or anything `apiMethods`-shaped. The offender is 52200b4 (PR #7769, *"give API-key revoke/restore a working product route"*), merged to `main` today at **14:30:34Z** — roughly 35 minutes before this PR's CI ran. It added `update` to `sys_api_key`'s `enable.apiMethods` so the Setup UI's Revoke button would stop 405-ing at the ADR-0049 method gate.

Identified from the file's commit history, not inferred from timing.

**Why it landed invisibly.** The conformance scan lives in `packages/spec`; the object it judges lives in `packages/platform-objects`. #7769 touched `platform-objects` and `plugin-auth` and not `spec`, so the affected-subset PR CI never ran the scan. The first PR to touch `spec` afterwards inherits the red — that is this one. The developer here ran the full spec suite locally and got 378 files / 9953 tests green, on a worktree branched from `main` before 14:30.

**Not re-queueing.** A re-queue cannot fix a failure that lives on the base branch, and each attempt rebuilds every PR behind it in the batch. This PR stays as-is until #7802 is resolved; when `main` goes green I will merge `main` in and let CI re-run.

**#7802 is not a "add `bulk` and move on".** #7769 deliberately kept `create`/`delete` off `sys_api_key`, and bulk-updating API keys is plausibly something that object should not offer — in which case the conformance rule is what needs the opt-out, not the object. That call belongs to whoever owns the ADR-0049 method-gate surface.

Nothing is requested of this PR. Everything else here is green.

---

Issue #7724, opened 2026-08-11
admin-lifecycle: `remove-user` is non-atomic and can never succeed — 409 leaks as a bodyless HTTP 500, credential rows deleted without rollback


huangyiirene:
## Symptom

As a genuine better-auth admin, `POST /api/v1/auth/admin/remove-user {userId}` on a user that owns a record fails, and the failure is **non-atomic**:

- **Observed:** HTTP **500 with an empty body**. `sys_user` is still readable; the owned record still carries its `owner_id`; and the user's credential `sys_account` rows have **already been deleted** — so that email can no longer sign in (401) while still occupying the org roster. Reproduced twice; verified across three victims (`accounts=0` for all three vs `accounts=1` for untouched users).
- **Expected:** 2xx, `sys_user` gone, the owned record surviving with `owner_id` cleared to null (engine FK clear).

The server log gives the real cause — a **409 leaking to the client as a bodyless 500**:

```
Cannot delete sys_user (): 1 dependent sys_member record(s) reference it via user_id
(user_id is required, so it cannot be cleared)… set deleteBehavior:cascade on sys_member.user_id
{ code: DELETE_RESTRICTED, status: 409 }
```

## Root cause

Three compounding problems, located on `origin/main`:

1. **Every user is auto-bound to the default org at sign-up/create** (the membership reconciler in `packages/plugins/plugin-auth/src/reconcile-membership.ts` runs as a `user.create.after` hook), and `sys_member.user_id` is a **required** FK with no cascade — `packages/platform-objects/src/identity/sys-member.object.ts` declares `user_id: Field.lookup('sys_user', { required: true })` and sets **no** `deleteBehavior:'cascade'`. So the engine's referential restrict vetoes the `sys_user` delete every time — `remove-user` can **never** succeed on this deployment.
2. **The operator cannot clear the blocker either.** `DELETE /api/v1/data/sys_member/{id}` → **405** — `sys-member.object.ts` sets `enable.apiMethods = ['get', 'list']` (writes owned by better-auth / refused by the identity write guard).
3. **No rollback.** The account/session deletions performed by better-auth's admin `remove-user` handler are **not** rolled back when the engine refuses the `sys_user` delete, leaving an orphaned, un-authenticatable identity on the roster.

[...]

Issue #4586, opened 2026-08-02
`sys_member` grade changes lose the true actor — thread attribution through the better-auth seam (validity windows deliberately NOT added)


os-zhuang:
&gt; **Revision 1 (2026-08-02):** narrowed after verifying the code. Two facts shrink this issue: `sys_member` already has `trackHistory: true` (the transitions ARE recorded — the missing piece is only *who*), and `granted_by` already exists on the capability-grant table (`auto-org-admin-grant.ts:244` deliberately writes `null` into it). And one half of the original "suggested shape" is **reversed**: adding ADR-0091 validity columns to `sys_member` turns out to be the wrong move, for a reason worth recording — see "Deliberately not doing". Original framing preserved in History.

Split out of #3723 (closed by #3802 / ADR-0108). Everything else in that investigation landed; this is the one substantive item left.

## The actual gap: attribution, not machinery

The elevation chain is `sys_member.role` (admin grade) → `auto-org-admin-grant.ts` → `sys_user_permission_set(organization_admin, org-scoped)` → `isTenantAdmin()`. Both hops lose provenance:

1. **Hop 1 — the grade change loses the true actor.** `sys_member` is `trackHistory: true`, so role transitions are already captured as history rows. But every writer is a better-auth path running under `isSystem`, so history records "system" as the actor. The human who clicked *make admin* is known exactly once — in the better-auth hook layer, where the session actor exists (`invitation-role-cap.ts` already resolves the issuer's membership on that same seam). It is discarded before the write reaches ObjectQL.

2. **Hop 2 — the capability grant discards provenance it has a column for.** `auto-org-admin-grant.ts:244` writes `granted_by: null` into a column that exists. So `explain` cannot answer "why is X a tenant admin" end to end, even though every table involved could carry the answer.

## Suggested work

**W1 — thread the true actor from the better-auth seam into the write context**, so `trackHistory` captures it. Two shapes:

- **General (recommended):** better-auth-originated ObjectQL writes carry the real actor for ATTRIBUTION. Fixes actor history for *every* better-auth-managed table in one move, not just `sys_member`.
- **Targeted:** emit explicit grade-transition audit events from the hook layer only. Smaller blast radius; the next table with the same problem does it again.

[...]

Issue #7732, opened 2026-08-11
session-list-revoke: `sys_session.revoked_at`/`revoke_reason` are declared readonly "System-managed" but the interactive-revoke path deletes rather than tombstones — ADR-0069 D4 revoke-audit trail is effectively inert


huangyiirene:
## Symptom

`sys_session.revoked_at` and `revoke_reason` are declared `readonly:true` and documented as "System-managed", intended to capture **every** revocation cause — the field description reads *"When set, this session was revoked (idle / absolute-max / concurrent-cap / admin). System-managed."* (`packages/platform-objects/src/identity/sys-session.object.ts`, ~lines 119–130). In practice the **admin/user-initiated interactive revoke** never stamps them, so the ADR-0069 D4 revoke-audit trail is effectively inert. The fields also appear in no listView.

## Root cause (as located)

The four documented causes split into two paths:

- **Automatic lifecycle enforcement (idle / absolute-max / concurrent-cap):** `packages/plugins/plugin-auth/src/auth-manager.ts` *does* stamp `revoked_at`/`revoke_reason` — it expires the session in place (`expires_at` set to the past + the two columns written) at ~lines 4606 (`{ …, revoked_at, revoke_reason: reason }`) and 4647 (`revoke_reason: 'concurrent_cap'`). So the run's "no writer anywhere" is imprecise: these three causes *do* have a writer. **But** better-auth garbage-collects expired sessions, so even these tombstones are best-effort (noted in the plugin-auth changelog).
- **Admin / user-initiated interactive revoke** (`revoke-user-sessions`, `revoke-other-sessions`, single-session revoke from the Sessions list): better-auth **deletes** the row rather than tombstoning it, so the `admin` cause is **never** recorded — `revoked_at`/`revoke_reason` are never written for the very path an audit trail most wants.

Net: the D4 revoke-audit trail is inert/unreliable — a deleted row carries no `revoke_reason`, and the enforcement-stamped rows are GC'd. This is an **enforce-or-remove** candidate: either make revocation tombstone-in-place across all four causes (and keep the row long enough to audit), or drop the two columns and the "System-managed" claim.

## Reproduction

1. As a user with an active session, `revoke-other-sessions` (or admin `revoke-user-sessions` against a target).
2. Read `sys_session` for the revoked session id → the **row is gone** (deleted), so `revoked_at`/`revoke_reason` were never populated for this `admin` revocation.
[...]

PR #7796, opened 2026-08-11
fix(plugin-auth): adopt the existing membership when accepting an invitation instead of colliding on the unique index

### LINKED ISSUE #7725: invitation-scope-gates: `accept-invitation` → bodyless HTTP 500, invitation stays `pending` forever (UNIQUE constraint on sys_member)
## Symptom

Accepting a valid org invitation never transitions the row from `pending` to `accepted`.

- **Observed:** `accept-invitation` returns **HTTP 500 with an empty body**, and the `sys_invitation` row stays `pending` **forever**. Reproduced twice with fresh emails.
- **Expected:** 2xx, the invitation moves to `accepted`, and the invitee holds exactly one membership in the target org.

Server log names the real cause:

```
insert into sys_member … UNIQUE constraint failed: sys_member.organization_id, sys_member.user_id
```

The delegated-admin **scope** itself is correct (a real `delegated_admin` can invite a member — 200, exactly one attributed row — but cannot invite an admin — 403 with the ADR-0090 D12 message — and leaves zero orphan rows). The break is purely in **acceptance**.

## Root cause

Acceptance re-inserts a membership the **sign-up reconciler already created**. Every user is auto-bound to the default org at sign-up via the membership reconciler (`packages/plugins/plugin-auth/src/reconcile-membership.ts`, run as a `user.create.after` hook). better-auth's built-in `/accept-invitation` route then performs its **own** `sys_member` insert, which collides with the reconciler's row on the unique index `{ organization_id, user_id }` (`packages/platform-objects/src/identity/sys-member.object.ts` — `indexes: [{ fields: ['organization_id', 'user_id'], unique: true }]`).

The reconciler yields to any pre-existing membership (its `insertMembership` path checks for an existing row first), but better-auth's built-in accept-invitation insert does not go through that seam, so it hits the constraint.

**Suspected fix:** acceptance should upsert/adopt the existing `sys_member` row (or skip the insert when the invitee is already a member of the target org).

This is the **same bodyless-500 statusCode-leak class** as the `remove-user` failure filed from this run — see #7724: an engine-level fault surfaces through the better-auth route as a bodyless 500 instead of a structured envelope.

## Reproduction

1. A `delegated_admin` (or admin) invites a fresh email as a member → 200, one `pending` row.
2. The invitee signs up / signs in through the invitation link (the reconciler binds them to the default org).
3. `POST /api/v1/auth/accept-invitation {invitationId}`.
4. Observe HTTP 500 with an empty body; the invitation row is still `pending` on re-read.

## Source

os-help:
Fixes #7725

## The defect

`POST /api/v1/auth/organization/accept-invitation` returned **HTTP 500 with an empty body** and left the `sys_invitation` row `pending` **forever**. On a single-organization deployment the documented flow — invite a fresh email, invitee signs up through the link, invitee accepts — could never complete, and the invitation was unrecoverable through the UI (re-inviting an address that is already a member is refused too).

Premise re-verified on `origin/main` by symbol, and it holds exactly as filed. Two individually-correct platform decisions collide:

- every user is auto-bound to the deployment's default organization at sign-up, by the membership reconciler (`reconcile-membership.ts`, composed into `user.create.after`, ADR-0093 D1/D2);
- `sys_member` declares `{ organization_id, user_id }` unique.

better-auth `1.7.0-rc.2`'s built-in accept-invitation route (`plugins/organization/routes/crud-invites.mjs`) assumes an invitee is never already a member: after flipping the invitation to `accepted` it calls `adapter.createMember(...)` unconditionally, inside a `runWithTransaction` whose `.catch` sets the invitation **back to `pending`** and rethrows. The invitee's auto-bound row made the insert fail, and the rollback erased the only evidence that acceptance had been attempted. The reconciler yields to a pre-existing row; better-auth's insert does not go through that seam.

## The seam, and why this one

The three seams the framework already owns on this route were each evaluated against better-auth's actual code and rejected — the reasoning is recorded in full in the new module's header, in short:

- **`organizationHooks.beforeAcceptInvitation`** fires before `createMember` and can only throw or mutate. The one mutation that would let the insert succeed is deleting the pre-existing row, which destroys the membership's `created_at` and history, opens a window in which the invitee belongs to nothing, and reaches into `sys_member`'s `deleteBehavior` — another card's decision surface (#7724).
- **`hooks.before`** (the global route-boundary middleware in `auth-manager.ts`) has the endpoint ctx, but `ctx.context` is the **shared** `AuthContext` singleton (`to-auth-endpoints.mjs` passes the awaited instance-wide context), so per-request state parked there leaks across concurrent requests — an adapter swap most of all.
- **Owning the "already a member" branch at the route boundary** would duplicate better-auth's recipient / expiry / status / membership-limit / email-verification checks. Duplicated security checks are where bypasses live.

The fix therefore lands in the better-auth to ObjectQL adapter's `create` for `sys_member` — the layer that already reconciles better-auth's model with the platform's object (model and field names, dates, identifier normalisation). It is stated as the platform's own rule rather than as leniency: **the declared unique pair IS the identity of a membership**, so a create naming an existing pair is that membership.

Blast radius measured against better-auth 1.7.0-rc.2 rather than assumed — accept-invitation is the only `member` create that can reach an existing pair:

- `POST /organization/add-member` pre-checks and refuses first (`findMemberByEmail` produces `USER_IS_ALREADY_A_MEMBER_OF_THIS_ORGANIZATION`, a 400);
- `POST /organization/create-organization` mints a fresh organization id, so no pair can pre-exist;
- the reconciler and its backfill write through `engine.insert` directly, not through this adapter, and already yield.

## Role and attribution on adoption

Not throwing is the smaller half. Adoption **writes the invitation's role onto the adopted row**, so an invitation's intent is not silently replaced by the reconciler's default `member` — accepting an `admin` invitation makes you an admin even if you signed up first. A grade-flat move is applied too (`member` to `delegated_admin`), since dropping it is exactly the "silently keep the default" failure.

One deliberate exception, pinned in both the unit and the end-to-end suites: **adoption never lowers a grade.** If the existing membership already outranks the invitation's role, the existing role is kept. Acceptance is an admission instrument — `invitation-role-cap.ts` states the same posture from the issuance side — while demotion has its own governed route, `POST /organization/update-member-role`, which is where `last-admin-guard.ts` stands. Without this exception, inviting an organization's sole `owner` as a `member` and having them accept would have demoted them past that guard, taking the organization's last owner with it and reporting success.

Attribution needs nothing extra, and that is deliberate: the adoption writes through the same `withSystemContext`-wrapped engine as every other adapter write, whose `update` carries `attributedUserId` (#4586) — here the invitee who accepted — and `sys_member` is `trackHistory: true`, so the adoption lands in history attributed to the acceptor. `created_at` is not rewritten: the membership really did begin at sign-up.

## Verification

New end-to-end suite drives the **real** better-auth pipeline through a real `AuthManager` (organization plugin, the framework's `beforeCreateInvitation` role cap, the reconciler on `user.create.after`, the ObjectQL adapter). The engine double **enforces `sys_member`'s declared unique index** and throws in the driver's own words — without that, the defect is invisible on an in-memory fake and the whole suite would be theatre.

Ablation (revert only the adapter wiring, keep the module and every test), direction predicted before running:

| test | predicted | observed |
|---|---|---|
| acceptance returns 2xx / invitation `accepted` / exactly one membership | red | red, `expected 500 to be 200` |
| adopted row carries the invitation's role | red | red, `expected 500 to be 200` |
| adoption never demotes | red (same collision, not the demotion) | red, `expected 500 to be 200` |
| invitee NOT yet a member still gets a membership created | green | green |
| delegated_admin can invite a member | green | green |
| delegated_admin cannot invite an admin (403, ADR-0090 D12, zero orphan rows) | green | green |
| 12 unit pins on the adoption decision | green (module untouched) | green |

The ablated stack trace is the reported failure verbatim: `insert into sys_member ... UNIQUE constraint failed: sys_member.organization_id, sys_member.user_id`, through `createMember` at `crud-invites.mjs:324`, surfacing as `[AuthManager] better-auth returned error: 500` with an empty body.

Suites, all on top of a merged `origin/main` and a fully built workspace:

- `@objectstack/plugin-auth` — **43 files / 1039 tests green** after the merge described below. That total is exactly the union of the two parents: 1015 (this branch) + 18 (`session-tombstone.test.ts`, new from #7825) + 6 (`managed-extension-fields.test.ts`, grown 6 to 12 by #7822). The merged test-file set is the union of both parents' sets with nothing dropped — 40 shared + 2 from this branch + 1 from `main`. `typecheck` green.
- `@objectstack/rest` (the HTTP surface the auth routes mount on) — 91 files / 1462 tests green.
- The two dogfood suites that drive the organization routes over a real stack — `membership-actor-attribution.dogfood.test.ts` (which exercises the real `accept-invitation`) and `delegated-admin-invite.dogfood.test.ts` — 10 tests green.
- Downstream consumer sweep green: `pnpm --filter '...@objectstack/plugin-auth' typecheck`, the **prefix** direction (the packages that DEPEND ON plugin-auth, not its dependencies). 14 consumers ran a `typecheck` script, all Done; `@objectstack/hono` and `@objectstack/cloud-connection` are in the set but declare no `typecheck` script.
- Gates for this surface: `check:nul-bytes` OK, `check:engine-double-contract` OK. The latter earned its keep — it caught the new engine double pinning `delete` but not `update`, and `update` is the verb adoption actually writes through. Fixed in `530b760e4` rather than baselined.

## Merge with `main` — coexistence with #7825

The merge queue ejected this PR with `MERGE_CONFLICT` after #7825 (session tombstoning) landed as `3f296bf`. Resolved by merging `origin/main` into the branch — no rebase, no force-push.

**Exactly one file conflicted, and only in its import block:** `packages/plugins/plugin-auth/src/objectql-adapter.ts`, where #7825's session-tombstone imports and this PR's `adoptExistingMembership` import landed on the same line. Resolved as the **union of both imports**; neither side was taken wholesale. There was no semantic reconciliation to make, because the two features hook **disjoint** adapter methods — adoption on `create`, tombstoning on `findOne` / `findMany` / `delete` / `deleteMany`. `index.ts` auto-merged carrying both exports.

A clean merge is not a working merge, and both features hook the same adapter dispatch, so coexistence was proved rather than assumed. Both end-to-end suites drive the same merged adapter through `AuthManager` calling `createObjectQLAdapterFactory`, so an ablation on the merged file is visible to both:

| ablation applied to the merged adapter | adoption suites (12 unit + 6 e2e) | #7825 `session-tombstone.test.ts` (18) |
|---|---|---|
| none — the state pushed here | 18 green | 18 green |
| adoption seam removed from `create` | **3 red** (the e2e cases), 15 green | 18 green |
| `reconcileSessionDelete` removed from `delete` | 18 green | **4 red**, 14 green |

Each feature's suite goes red only under its own ablation, so neither is passing vacuously on the merged dispatch, and both features still fire.

Also re-run after the merge: `pnpm --filter @objectstack/plugin-auth typecheck` green, the prefix-direction consumer sweep green across all 16 selected packages, `check:engine-double-contract` OK (161 pinned / 133 DEBT / 2 exempt) and `check:nul-bytes` OK. Rebuilding `packages/spec` on the merged sources produced **no** artifact drift, so the merged generated files reproduce exactly. The `@objectstack/rest` and dogfood numbers above are from the pre-merge round and were not re-run; nothing in the merge touches that surface.

`main` has since advanced two further commits (`3bb9340`, `bf8fd8a` — docs, QA checklists, and a `packages/spec` doc comment). `git merge-tree` against them reports **no conflict**, and the merge queue rebuilds onto live `main` regardless, so they are deliberately not chased here.

## Scope

Both family non-goals held — confirmed by the diff's file list, which touches neither:

- `packages/platform-objects/src/identity/sys-member.object.ts` (FK / `deleteBehavior` / indexes) is **not** in the diff — that stays #7724's decision surface. The unique index is honoured here, not changed.
- The shared better-auth to REST error mapping (`packages/rest/src/rest-server.ts`) is **not** in the diff.

**Bodyless-500 residue, for #7724's benefit:** the success path is fixed, but the mapping is not, and a fault on this path still surfaces bodyless. Anything that makes better-auth's accept-invitation transaction throw a non-`APIError` — an engine outage, a driver error, a constraint this fix does not cover — still reaches the client as a 500 with an empty body and still rolls the invitation back to `pending`. That is the shared mapping class, deliberately left alone here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BVc1ekPpi6yaWywAUhfzfd)_


huangyiirene:
**Ejected from the merge queue: `MERGE_CONFLICT`** (github-merge-queue[bot], ~20:35Z). Not a flake, not a CI failure — diagnosed and being resolved. `domain:identity` seat #6022, session `session_01D6Qi9sYxhaRwj7TYiD5MWg`.

## What happened

This is the collision the seat's hot-file serial queue predicted, resolved in the opposite order from the one recorded. Both this PR and **#7825** (#7732, session tombstoning) touch `packages/plugins/plugin-auth/src/objectql-adapter.ts`. #7825 reached the front of the queue first and landed as `3f296bf`; this PR's base is still `098b629`, so it now conflicts.

Recorded plainly because the serial-queue note had the order as `#7796 → #7825` and reality went the other way. The queue arbitrated, which is correct — ⛔ PM does not hand-order the queue. The cost is one merge resolution here.

## Conflict scope, measured read-only (`git merge-tree` against `origin/main` @ `3f296bf`)

**Exactly one file, exactly one hunk** — and it is an import-block collision, nothing deeper:

```
Auto-merging  packages/plugins/plugin-auth/src/index.ts          ← clean
CONFLICT      packages/plugins/plugin-auth/src/objectql-adapter.ts
```

```
<<<<<<< origin/main
import { filterRevokedSessionRows, hideRevokedSessionRow, reconcileSessionDelete } from './session-tombstone.js';
=======
import { adoptExistingMembership } from './adopt-membership.js';
>>>>>>> pr-7796
```

Both sides are **additive and disjoint in intent**: #7825 owns `delete` on `sys_session`, this PR owns `create` on `sys_member`. The resolution is a union of the two import statements.

## ⛔ But a clean merge is not a working merge

The rest of the file auto-merged, and that is exactly the situation AGENTS.md §10 warns about. Both changes hook the **same adapter dispatch**, so their coexistence has to be *demonstrated*, not inferred from git exiting 0:

- `create` / `sys_member` (adoption) and `delete` / `sys_session` (tombstone reconcile) must both still fire after the merge;
- ⛔ **do not mechanically take either side** of the hunk — taking one drops the other's feature entirely, and both halves are already reviewed and ACCEPTed;
- the full `@objectstack/plugin-auth` suite is the gate, not the two new suites alone.

Re-queued once the resolution is pushed and verified. Nothing about this PR's review verdict changes — it stays ACCEPTed; this is a rebase-class patch round, not rework.

---

Issue #7484, opened 2026-08-10
[finding] No CI gate checks documentation anchors — `lychee.toml` sets `include_fragments = "none"`, so a link to a heading that does not exist passes as `[200] OK`


os-help:
Observation-class gate gap, measured while implementing #7465 (PR #7483, the `## The one gate, three entry points` → `four doors` rename). **Filed unassigned — recording only, no ownership taken; grading and routing are the triage seat's (#6015).** ⛔ Not self-claimed.

## The measurement

`Check Documentation Links` runs lychee, and it resolves the **file** target only — it does not look at the `#fragment`. This was not read off a config comment; it was **probed with the pinned binary under the CI argv**:

- `lychee.toml` sets `include_fragments = "none"`
- a probe file linking `/docs/deployment/validating-metadata#this-anchor-does-not-exist-at-all`, run through **lychee 0.24.2** with the exact argv from `.github/workflows/check-links.yml`, is reported **`[200] ✅ OK`**
- the same run over the real corpus: 1756 total / 775 unique / 0 errors

So a heading rename that updates the heading but not an inbound `#anchor` — or updates the anchor but leaves the link text contradicting it — **ships green**.

## Why this is worth recording rather than assuming everyone knows

It falsified a live PM patrol criterion. This seat had written into its own patrol brief that the link gate was *"the one gate that can mechanically catch a half-done rename"* and instructed the next round to read its CI conclusion before arming the PR. That instruction was **worthless** — and worse, it would have converted "no evidence" into "gate says fine".

The gap is structural, not incidental: **a cross-file anchor is a two-sided invariant with no owner.** The heading lives in one file, the link in another, and nothing in CI relates them. #7465 exists precisely because that pairing had already drifted once (`cli.mdx:458` still pointed at a heading whose text had been correct in 2024).

## What #7483 used instead, and why it is not a substitute

PR #7483 verified its own rename by:

1. computing the slug with the repo's **own** `github-slugger@2.0.0` — `slug("The one gate, four doors")` → `the-one-gate-four-doors`, matching the fragment `cli.mdx` now links; and
2. a repo-wide grep proving no residual references to the old fragment.

Both are correct, and both are **manual discipline that happened because a dispatch asked for it**. Neither runs on anyone else's PR.

## Possible directions — deliberately not pre-judged

[...]

PR #7803, opened 2026-08-11
test(spec): register sys_api_key in SINGLE_RECORD_WRITE_ONLY (queue-blocking)

huangyiirene:
Fixes #7793

## Signature

```
FAIL packages/spec/src/data/api-methods-batch-conformance.test.ts
  apiMethods conformance — single-record writes imply batch (#3026)
  > grants bulk wherever it grants create / update / delete
```

Queue build [31503475407](https://github.com/objectstack-ai/objectstack/actions/runs/31503475407) is red on this test for every merge-queue full-suite build, including PRs with zero causal relation to it (PR #7723).

## Mechanism

PR #7769 (`52200b4`) changed `sys_api_key`'s `enable.apiMethods` from `['get', 'list']` to `['get', 'list', 'update']` to give API-key revoke/restore a working product route. `update` is a `WRITE_PRIMITIVES` entry, so `api-methods-batch-conformance.test.ts`'s `grants bulk wherever it grants create / update / delete` check now requires `sys_api_key` to either grant `bulk` or be registered in `SINGLE_RECORD_WRITE_ONLY` with a reason. Neither happened, so the conformance test fails on `origin/main` — invisible to #7769's own PR-side CI because the affected-subset run doesn't exercise this file-scan test for a `platform-objects`-only diff; only the queue's full-suite build does.

## Why route B (register the exemption), not route A (add `bulk`)

#7769's own commit message already made this call:

> `enable.apiMethods` now carries `update` (`create` / `delete` stay off: minting is `POST /api/v1/keys`, and keys are retired by revoking rather than deleting)

Revoke/restore are single-record row actions (`PATCH /api/v1/data/sys_api_key/{id}`); batch revoke is not a declared product route today. Adding `bulk` (route A) would widen the API surface against that recorded intent — this PR takes route B instead: it *records* #7769's decision in the test's own documented channel (`SINGLE_RECORD_WRITE_ONLY`), it does not make a new one. If the identity lane or the maintainer later decides batch revoke should exist, that's route A as a follow-up, and this exemption entry deletes cleanly.

## Change

One entry added to `SINGLE_RECORD_WRITE_ONLY` in `packages/spec/src/data/api-methods-batch-conformance.test.ts`:

```ts
sys_api_key:
  "row actions revoke/restore (#7769/#7727) opened 'update' for a working " +
  "PATCH /api/v1/data/sys_api_key/{id} route; 'create'/'delete' deliberately " +
[...]

Issue #7793, opened 2026-08-11
Queue-blocking: #7769 opened `update` on sys_api_key without `bulk` — `api-methods-batch-conformance` fails every merge-queue (full-suite) build


huangyiirene:
## Shared damage — every queue build is red; PR-side CI cannot see it

**Signature** (queue build [31503475407](https://github.com/objectstack-ai/objectstack/actions/runs/31503475407), kicked PR #7723 which has zero causal relation to it):

```
FAIL packages/spec/src/data/api-methods-batch-conformance.test.ts
  apiMethods conformance — single-record writes imply batch (#3026)
  > grants bulk wherever it grants create / update / delete
```

**Mechanism, verified on `origin/main` @ `098b629`:**

- PR **#7769** (`52200b4`, merged 14:30Z) changed `sys_api_key`'s `enable.apiMethods` from `['get','list']` to `['get','list','update']` — deliberately, to give revoke/restore a working product route.
- The conformance test file-scans every `*.object.ts` and requires: any object granting a write primitive (`create`/`update`/`delete`) must also grant `bulk`, **or** be registered in `SINGLE_RECORD_WRITE_ONLY` with the reason.
- `SINGLE_RECORD_WRITE_ONLY` is `{}` and `sys_api_key` appears nowhere in the file — #7769 did neither half.
- Why #7769's own CI was green: the test is a **file-scan test living in `packages/spec`**; the PR-side affected-subset run does not exercise it for a `platform-objects` diff. Only the queue's full-suite build does — so main is red exactly where every queued PR gets built.

**Two fix routes:**

- **A. Add `bulk`** to `sys_api_key.enable.apiMethods` — widens the API surface; contradicts #7769's own recorded intent (「`create` / `delete` stay bucket-default (off): minting is `POST /api/v1/keys`」— revoke/restore are single-record row actions, batch revoke is not a product route).
- **B. Register the exemption** — one `SINGLE_RECORD_WRITE_ONLY` entry with the reason, which is the test's own documented channel for exactly this decision (its header: "it genuinely must not → register it ... with the reason, so the choice is on the record").

**Route B is the stopgap being dispatched now** (it records #7769's already-made decision; it does not make a new one). If the identity lane or the maintainer later rules that batch revoke SHOULD exist, that is route A as a follow-up — the exemption entry deletes cleanly.

[...]

PR #7808, opened 2026-08-11
fix(spec): render the reference section for a non-object root schema (#7658)

### LINKED ISSUE #7658: [finding] build-docs.ts silently drops the whole reference-doc section for a bare opaque-string top-level schema (~23 exports with a real .describe())
Found while implementing #7513 (re-typing `TenantPlanSchema` from a closed enum to an opaque `z.string().describe(…)`).

## What happens

`packages/spec/scripts/build-docs.ts`'s `generateMarkdown()` decides what to render for a top-level exported schema like this:

```ts
const defs = schema.definitions || schema.$defs || {};
let mainDef = defs[schemaName];

if (!mainDef && (schema.properties || schema.enum || schema.anyOf || schema.oneOf)) {
  mainDef = schema;
}
if (!mainDef) {
  mainDef = Object.values(defs)[0];
}
if (!mainDef) return '';   // <-- silently drops the entire "## SchemaName" section
```

The root-detection condition only recognizes `object` (`properties`), `enum`, and union (`anyOf`/`oneOf`) shapes. A schema whose JSON Schema is a **bare primitive** — `{"type": "string", "description": "…"}` with no `enum`, `properties`, `anyOf`, or `oneOf` — matches none of those, so `mainDef` stays `undefined` and `generateMarkdown` returns `''`. Not just the "Allowed Values" bullets — the **entire heading, including the schema's own `.describe()` text**, disappears from the published page. The `## TypeScript Usage` import list and the root `content/docs/references/index.mdx` table still name the export (they're driven by the export surface, not this function), so the page looks like it forgot to finish rendering an entry it just imported.

## Confirmed on #7513

Before #7513, `TenantPlan.json` was `{"type": "string", "enum": [...]}`, which the `schema.enum` branch of the condition caught, so `content/docs/references/cloud/tenant.mdx` had a `## TenantPlan` / `### Allowed Values` section. #7513 widens it to `{"type": "string", "description": "..."}` (no enum) — and the whole section vanished on regeneration, taking the schema's new ownership-statement description with it. The JSON Schema itself (`packages/spec/json-schema/cloud/TenantPlan.json`) still carries the full description; only the rendered `.mdx` page silently lost it.

## Scope — this is not just `TenantPlan`

Scanning `packages/spec/json-schema/**/*.json` for `type ∈ {string,number,integer,boolean}` with no `enum`/`properties`/`anyOf`/`oneOf`, **31 exported schemas** match this shape today, **23 of them carrying a real `.describe()`** that never reaches its reference page:

```
api/EventPattern, cloud/EnvironmentDriver, cloud/PackageCategory, cloud/PackageLocale,
cloud/Sha256Digest, cloud/TenantPlan, data/DriverSslToggle, data/DriverType,
kernel/OpsFilePath, shared/AppName, shared/EventName, shared/FieldName, shared/FlowName,
shared/ObjectName, shared/RoleName, shared/SnakeCaseIdentifier, shared/SystemIdentifier,
shared/ViewName, system/Locale, system/Sha256Digest, system/TraceFlags, ui/ViewItemName
```

(8 more have no description either way: `data/CalendarDateValue`, `ClockTimeValue`, `ContextToken`, `ContextTokenPlaceholder`, `DateMacroPlaceholder`, `DateMacroToken`, `FieldNode`, `FileReferenceIdValue`, `InstantValue`, `ReferenceIdValue`.)

## Why `finding`, not a queued defect

Nothing breaks: `check:docs` compares the generated output to what's committed, so a doc that never had the section stays green — this has apparently been silently true for a while (`EnvironmentDriver` etc. predate #7513 and already showed the same gap before my change touched anything). No runtime path depends on the rendered docs. It's a documentation-quality gap — real `.describe()` text authors wrote to be read is invisible on the one page meant to show it — not a defect a user hits today. Filed unassigned/no `pm:queue` per the observation-class convention.

## Suggested fix shape (not attempted here — out of scope for #7513)

Add an `else if (mainDef.type)` fallback in `generateMarkdown()` (`packages/spec/scripts/build-docs.ts`, ~line 368) that renders the description (already handled above the type-specific branches) plus a `**Type:** \`<type>\`` line for any recognized root type that isn't object/enum/union, instead of falling through to the `if (!mainDef) return ''` case. Would need a fixture pin (a bare-string schema whose description shows up on its page) so this doesn't silently regress again.

---
_Filed while implementing #7513._
_Generated by [Claude Code](https://claude.ai/code/session_01JY2Q5Xto1u8YHADgrZDTnk)_

os-help:
Fixes #7658

## What was wrong

`packages/spec/scripts/build-docs.ts`'s `generateMarkdown()` picked the node it was documenting by **enumerating shapes** — `properties`, `enum`, `anyOf`, `oneOf` — and answered "none of those" with `return ''`.

A JSON Schema root is routinely none of those:

| Zod spelling | JSON Schema root | old rendering |
| :--- | :--- | :--- |
| `z.string().describe(…)` | `{ type: 'string', description }` | — nothing — |
| `z.record(…)` | `type: object` + `additionalProperties`, no `properties` | — nothing — |
| `z.array(…)` | `{ type: 'array', items }` | — nothing — |
| `z.intersection(…)` | `{ allOf: [...] }` | — nothing — |

Each lost its **entire `## Name` section** — the heading, and with it the `.describe()` prose an author wrote to be read — while the page's `## TypeScript Usage` block, which is spelled from the export surface rather than from this function, went on naming the export. The page read as if it had forgotten to finish rendering an entry it had just imported.

## The count, re-measured today

The card says *"~23 exports with a real `.describe()`"*. Re-deriving it by replaying `generateMarkdown`'s exact root-selection over all 1578 published schemas (`wc -l` on the emitted lists, no pager in the pipeline):

| | today | card |
| :--- | ---: | ---: |
| sections rendered | 1533 | — |
| **sections dropped (`return ''`)** | **45** | ~31 |
| ...carrying a root `description` | **33** | ~23 |
| bare-scalar roots specifically | 32 (22 with a description) | 31 (~23) |

The card scanned only the bare-scalar spelling. The other 13 are record maps, one array root (`system/OidcProvidersConfig`) and one `allOf` (`data/FilterCondition`) — the same defect through the same line, so they are fixed with it. **Scope expansion, disclosed rather than absorbed.**

Positive control on the same page, before the fix: `content/docs/references/cloud/tenant.mdx` renders `## TenantContext`, `## TenantDatabase` and 7 more, and does **not** render `## TenantPlan` — while line 26 of that same file imports `TenantPlanSchema`. The zero is a measurement, not a dead walk.

## The fix

- `selectRootDef` gains a **fourth, unconditional** step: the document itself is the schema. Appended rather than folded into the root-shape test — a document that reaches the "first `$defs` entry" step today must keep reaching it, or a root that merely *looks* renderable would start outranking the definition with the actual content.
- A final `else` renders the description plus one `**Type:**` line, using the **same `formatType`** every property row on the page already uses, so a schema's own section and a property typed with it cannot disagree. `allOf` is spelled as an `&` intersection (root-only spelling `formatType` has never met in a property position), with identical members collapsed — safe here because `A & A` is `A`, and deliberately not the union rule, where arity is the point (#6569).
- `any` is withheld rather than printed: on this page it reads as "free-form slot, nothing validates it", the misreading #5606 fixed for tombstones.
- Constraints (`pattern`, `minLength`, `minimum`) are deliberately **not** spelled — this renderer prints none in any position, and `json-schema/` stays the authority.

The renderer moves to `packages/spec/scripts/lib/schema-section.ts` so it can be pinned directly: the extraction `lib/format-type.ts` got at #4912, for the same reason. Its output here was the **empty string**, which is exactly what grepping emitted `.mdx` cannot see — and why `check:docs` (generated vs. committed) stayed green over this for months.

## Regeneration

- **244 insertions, 0 deletions** across 27 pages. The 45 new `## ` headings are exactly the 45 measured dropped names (set diff empty); none of the 1533 already-rendering sections moves a byte.
- **Emitted files 230 → 230.** An empty section never removed a page, so no category's `pages.length` changes — **#7303's `pages.length === 0` `meta.json` guard is untouched**, and `contracts/` neither gains nor loses a directory (14 category dirs before and after).
- `gen:docs` run **twice**: second run byte-identical to the first (same tree sha256, same 27 changed files).
- All 45 new headings slug cleanly under the repo's own `github-slugger` — 0 collisions. No heading renamed, so there is no old fragment to grep for (#7484).

## Gates

| gate | result |
| :--- | :--- |
| `pnpm --filter @objectstack/spec run check:docs` | ✅ 230 generated files in sync |
| `pnpm --filter @objectstack/spec run gen:docs` ×2 | ✅ stable, identical output |
| `pnpm --filter @objectstack/spec exec vitest run scripts/schema-section.test.ts` | ✅ 13 passed |
| `pnpm --filter @objectstack/spec run typecheck` | ✅ |
| `pnpm --filter @objectstack/spec run check:api-surface` | ✅ public API surface unchanged |
| `pnpm --filter @objectstack/spec run check:generated` | ✅ all 13 generated artifacts up to date |
| `pnpm exec eslint <changed files>` | ✅ clean |
| `pnpm run check:doc-authoring` / `check:empty-changeset` / `check:adr-0087-registration` | ✅ |
| `pnpm --filter @objectstack/spec run test` | 9960 passed, **1 pre-existing failure** — see below |

**Reverse verification**: restoring the old behaviour (dropping the unconditional root fallback and returning `''`) turns 7 of the 13 new cases red — the six render cases lose their whole section, heading included — and leaves the other 6 green. That asymmetry is the point: the four shapes the old code enumerated were always correct, which is why this survived unnoticed.

**Pre-existing failure, not from this diff**: `src/data/api-methods-batch-conformance.test.ts` fails on `sys_api_key: [get, list, update] grants single-record writes but not 'bulk'` in `packages/platform-objects/src/identity/sys-api-key.object.ts` — a file this branch does not touch (`git diff --name-only origin/main -- packages/platform-objects` → 0).

### DIFF
--- a/.changeset/reference-docs-opaque-root-schemas.md
+++ b/.changeset/reference-docs-opaque-root-schemas.md
@@ -1,0 +1,42 @@
+---
+"@objectstack/spec": patch
+---
+
+fix(spec): reference pages no longer drop the whole section for a non-object root schema (#7658)
+
+`build-docs.ts` decided which node of a published document it was documenting by
+enumerating shapes — `properties`, `enum`, `anyOf`, `oneOf` — and answered "none
+of those" with `return ''`. A JSON Schema root is routinely none of those:
+`z.string().describe(…)` compiles to a bare scalar, `z.record(…)` to an object
+with `additionalProperties` and no `properties`, `z.array(…)` to an array,
+`z.intersection(…)` to an `allOf`. Every one of them lost its ENTIRE `## Name`
+section — heading included, and with it the `.describe()` prose an author wrote
+to be read — while the page's `## TypeScript Usage` block, which is spelled from
+the export surface rather than from this function, went on naming the export. The
+page read as if it had forgotten to finish rendering an entry it had just
+imported.
+
+**Measured on today's tree: 45 published schemas were in that state, 33 of them
+carrying a description.** The filed issue counted 22 (`~23`), having scanned only
+the bare-scalar spelling; record maps, one array root and one `allOf` are the
+same defect through the same line. Regenerating restores all 45 sections — 244
+added lines, **zero removed**, so none of the 1533 sections that already rendered
+moves. Emitted file count is unchanged at 230: a section rendering as the empty
+string never removed a page, so no category's `pages.length` changes and the
+`meta.json` emit guard added in #7303 is untouched.
+
+Nothing was failing, which is why this stood for months: `check:docs` compares
+generated output to committed output, so a section that never existed stays green
+forever, and there is no grep for what is missing.
+
+The renderer moved to `scripts/lib/schema-section.ts` to be pinned directly —
+the same extraction, for the same reason, that `lib/format-type.ts` got at #4912.
+Its output was the empty string, the one thing grepping emitted `.mdx` cannot
+see.
+
+Sections for these shapes now carry the schema's description plus a single
+`**Type:**` line, rendered by the same `formatType` that every property row on
+the page uses, so a schema's own section and a property typed with it cannot
+disagree. Constraints (`pattern`, `minLength`, `minimum`) are deliberately not
+spelled: this renderer prints none in any position, and `json-schema/` stays the
+authority on them.
--- a/content/docs/references/api/contract.mdx
+++ b/content/docs/references/api/contract.mdx
@@ -482,6 +482,12 @@ const result = ApiErrorSchema.parse(data);
 
 ---
 
+## RecordData
+
+Key-value map of record data
+
+**Type:** `Record<string, any>`
+
 
 ---
 
--- a/content/docs/references/api/websocket.mdx
+++ b/content/docs/references/api/websocket.mdx
@@ -178,6 +178,12 @@ const result = AckMessageSchema.parse(data);
 
 ---
 
+## EventPattern
+
+Event pattern (supports wildcards like "record.*" or "*.created")
+
+**Type:** `string`
+
 
 ---
 
--- a/content/docs/references/cloud/environment-artifact.mdx
+++ b/content/docs/references/cloud/environment-artifact.mdx
@@ -34,6 +34,12 @@ const result = Sha256DigestSchema.parse(data);
 
 ---
 
+## Sha256Digest
+
+SHA-256 digest (64 hex chars)
+
+**Type:** `string`
+
 
 ---
 
--- a/content/docs/references/cloud/environment.mdx
+++ b/content/docs/references/cloud/environment.mdx
@@ -107,6 +107,12 @@ Credential lifecycle status
 
 ---
 
+## EnvironmentDriver
+
+Data-plane driver key (e.g. `turso`, `libsql`, `sqlite`, `postgres`)
+
+**Type:** `string`
+
 
 ---
 
--- a/content/docs/references/cloud/package.mdx
+++ b/content/docs/references/cloud/package.mdx
@@ -91,8 +91,20 @@ Register a new package in the Control Plane
 
 ---
 
+## PackageCategory
+
+Package category for marketplace discovery (e.g. "crm", "hr", "finance", "devtools")
+
+**Type:** `string`
+
 
 ---
+
+## PackageLocale
+
+BCP-47 locale tag
+
+**Type:** `string`
 
 
 ---
@@ -127,6 +139,12 @@ Per-locale overrides for a package listing
 
 
 ---
+
+## PackageTranslations
+
+Locale-keyed overrides; missing keys fall back to base columns
+
+**Type:** `Record<string, { displayName?: string; description?: string; readme?: string; tagline?: string; … }>`
 
 
 ---
--- a/content/docs/references/cloud/tenant.mdx
+++ b/content/docs/references/cloud/tenant.mdx
@@ -158,6 +158,12 @@ const result = PackageInstallationSchema.parse(data);
 
 ---
 
+## TenantPlan
+
+Opaque plan/tier identifier. The vocabulary is control-plane config owned by the cloud distribution, not protocol — this schema accepts any string and does not enumerate valid values. Convention: an empty or unrecognized value is treated as the free tier by cloud-side readers (not enforced by this schema).
+
+**Type:** `string`
+
 
 ---
 
--- a/content/docs/references/data/context-tokens.mdx
+++ b/content/docs/references/data/context-tokens.mdx
@@ -92,8 +92,16 @@ const result = ContextTokenSchema.parse(data);
 
 ---
 
+## ContextToken
+
+**Type:** `string`
+
 
 ---
+
+## ContextTokenPlaceholder
+
+**Type:** `string`
 
 
 ---
--- a/content/docs/references/data/datasource.mdx
+++ b/content/docs/references/data/datasource.mdx
@@ -68,6 +68,12 @@ const result = DatasourceSchema.parse(data);
 
 ---
 
+## DriverType
+
+Underlying driver identifier
+
+**Type:** `string`
+
 
 ---
 
--- a/content/docs/references/data/date-macros.mdx
+++ b/content/docs/references/data/date-macros.mdx
@@ -92,8 +92,16 @@ const result = DateMacroPlaceholderSchema.parse(data);
 
 ---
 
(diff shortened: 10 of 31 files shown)


os-help:
**CI red on `Test Core (1/3)` — not from this diff.** One failing test, and it is the pre-existing one disclosed in the PR body. Diagnosis and why I am not fixing it here:

```
FAIL src/data/api-methods-batch-conformance.test.ts
  > grants bulk wherever it grants create / update / delete
  + "sys_api_key: [get, list, update] grants single-record writes but not 'bulk'
     — /batch and the *Many routes will 405
     (packages/platform-objects/src/identity/sys-api-key.object.ts)"
```

**Cause** — `52200b4` ("give API-key revoke/restore a working product route (#7727) (#7769)", on `main` at ~14:43Z today, ~15 min before this branch was cut) widened `sys_api_key`'s whitelist:

```diff
-    apiMethods: ['get', 'list'],
+    apiMethods: ['get', 'list', 'update'],
```

Granting a single-record write verb without `'bulk'` is exactly what the #3026 ratchet forbids. This branch touches **zero** files in that package (`git diff --name-only origin/main -- packages/platform-objects` → 0), and the test is a filesystem scan of `packages/**/*.object.ts` — it never reads anything this PR changes.

**Why `main` looks green, and why that is not evidence.** The `Test Core (1/3)` job at `098b629` (this PR's base) passed — as a **turbo cache replay**, not a run. `turbo.json`'s `test` task uses `$TURBO_DEFAULT$` inputs, so `@objectstack/spec#test`'s cache key covers `packages/spec/**` plus `^build` of its internal deps — and `@objectstack/spec` has **no** `@objectstack/*` dependencies at all. The test reads `packages/platform-objects/**` at *runtime*, which the key cannot see. So editing `sys-api-key.object.ts` does not invalidate the task, and `main` kept replaying a stale PASS; `52200b4`'s own CI run was cancelled, so no completed run ever exercised it. This PR edits `packages/spec/scripts/**`, which changed the key and forced a real run (`Cached: 1 cached, 57 total`) — surfacing the break rather than causing it.

That cache blind spot is worth a card of its own: a cross-package scanning test whose inputs turbo cannot express will keep reporting the last state anyone happened to invalidate it in.

**Why I am not fixing it in this PR.** The test's own docblock says to pick one deliberately — grant `'bulk'`, or register the object in `SINGLE_RECORD_WRITE_ONLY` with a written reason. That is a product decision about API-key batch semantics, on the surface #7727/#7769 landed twenty minutes ago, and it is outside this card's scope (`packages/spec/scripts/build-docs.ts`, the regenerated tree, a changeset). Making that call silently inside a docs-generator PR would put an unreviewed API decision in a diff nobody would look for it in.

Every gate this PR is responsible for is green: `check:docs` (230 in sync), `gen:docs` ×2 stable, `check:api-surface`, `check:generated` (13/13), `typecheck`, eslint, and the 13 new pins. I will re-check when the base recovers.

---

os-help:
Correcting the "Pre-existing failure, not from this diff" note at the bottom of the body — it was the right call, but argued from an empty diff rather than a measurement, and it did not name the card that already owns it.

`Test Core (1/3)`'s red is **#7793** (`sys_api_key` grants `update` without `bulk`, from #7769 / `52200b4`), already fixed by open PR **#7803**. Measured on this PR's own base, with nothing from this branch applied:

```
$ git worktree add ../base --detach 098b629 && pnpm --filter @objectstack/spec exec vitest run \
    src/data/api-methods-batch-conformance.test.ts
 Test Files  1 failed (1)      Tests  1 failed | 3 passed (4)
     × grants bulk wherever it grants create / update / delete
```

The sibling-PR control that motivated the send-back doesn't hold: #7810's three `Test Core` shards declared 6300 / 6385 / 6047 tests, and `@objectstack/spec` alone declares 9961 — so it never ran there. This PR's shard 1/3 declared 14554. `turbo run test --filter=@objectstack/spec --dry=json` confirms why: 1082 inputs, none from `platform-objects`, `dependencies: []`.

No assertion was weakened and nothing was pushed — the branch is unchanged at `f6bdfdb`. This PR goes green once #7803 lands and this branch is rebased/merged onto the new `main`. Full evidence: objectstack-ai/objectstack#7658 (comment 5255686646).

---

Issue #7659, opened 2026-08-11
lint: publish-time predicate validation should refuse a path-shaped RIGHT-hand side — outside the renderer's declared ==/!= subset, and unreachable by #7214's resolution check


yinlianghui:
Cross-repo relay from the objectui whole-repo seat (session `session_017Qqyix2QcnpUC9XeYVDzx3`), executing the D half of the ruling recorded on objectui#4049. Filed unassigned for the spec/lint lane's triage; this is a NEW card rather than a comment on the closed #7010, because that card's delivered scope (PR #7214 — paths that fail to RESOLVE against the target schema) is a different refusal from the one needed here, and a completed card is the wrong place to accrete scope.

## The gap

objectui's metadata-admin predicate evaluator supports a declared subset: the RIGHT side of `==`/`!=` is a LITERAL (`parseLiteral`), never a resolved path. `data.a == data.b` therefore compares `data.a`'s value against the literal STRING `"data.b"` — always false, silently. objectui#4049 (PR objectui#4264, merged-pending) adds the consumer-side dev diagnostic naming this boundary; the objectstack maintainer's ruling on #6828-family shapes and the objectui seat's ruling on #4049 both put the CONTRACT half here: authored predicates that exceed the renderer's declared subset should be refused at publish time, contract-first, rather than tolerated into a silent wrong verdict.

## What #7214 covers vs what this asks

- **#7214 (landed)**: a path reference that does not RESOLVE against the target schema → refused. A well-formed `data.a == data.b` passes it — both paths resolve fine; the problem is that the right side's POSITION does not evaluate paths.
- **This card**: the publish-time check additionally refuses (or warns, per this lane's severity conventions) a `==`/`!=` right-hand side that is path-shaped — an unquoted identifier chain (`data.b`, or a bare `text` missing its quotes) — pointing the author at the sanctioned spellings: quote the literal, or restructure the comparison so the path is on the left. The refusal grammar the consumer diagnostic uses (identifier-chain: `/^[A-Za-z_$][A-Za-z0-9_$]*(?:\.[A-Za-z_$][A-Za-z0-9_$]*)*$/`) is the natural shared definition — one grammar, two enforcement points.

## Why publish-time and not just the dev warning

[...]

Issue #7576, opened 2026-08-11
[finding] `validateSecurityPosture`'s `CLI_ONLY` surfaceReason claims coverage the ADR-0094 `object` authoring gate does not give it — it covers 1 of the block's 13 rules


os-help:
Found while implementing #7503 (PR #7574, the `controlled_by_parent`-without-relation lint rule). Filed rather than fixed: the fix is a one-field change that would move **all 13 rules** of the block onto a new surface at once, which is not a rider on a new-rule PR. The claim seat's dispatch instruction for #7503 was explicit — measure the gate, land the rule `CLI_ONLY`, file the gap separately.

## The claim

`validateSecurityPosture` is registered **once, as a block**, in `packages/lint/src/authoring-rules.ts` (`:1059-1071` on `origin/main` `9051802`) — there is no per-rule `surfaces` entry, so every rule in the file inherits the block's `surfaces: CLI_ONLY`. Its `surfaceReason`, verbatim:

> Already gated at this surface by a DIFFERENT mechanism: plugin-security registers an ADR-0094 authoring gate on `object` (`registerAuthoringGate`) that enforces **the same OWD posture rules** on every runtime write. Running the linter here as well would double-report one refusal in two vocabularies. Consolidating the two onto this table is P2 (#4463), and is a merge, not a hole.

That is a claim about coverage. It is not a reading of the gate.

## What the gate actually does

`packages/plugins/plugin-security/src/object-posture-gate.ts` — the whole gate is 127 lines, and `objectPostureGate` implements exactly **two** rules, both stated in its own header:

- **R1** — env-tighten-only (ADR-0086 D1): an environment overlay over an artifact-backed object may not widen `sharingModel` / `externalSharingModel` past the packaged declaration.
- **R2** — external ≤ internal (ADR-0090 D11).

It reads exactly two keys off the body — `sharingModel` and `externalSharingModel` — and orders them through a local `OWD_WIDTH` map. It never reads `fields`, `actions`, `permissions`, `books` or `data`. `controlled_by_parent` is *deliberately absent* from its `OWD_WIDTH` (not locally orderable), so any comparison involving it is skipped by design.

## The measurement

Mapping the gate against the 13 rule ids the block now carries (12 on `main`, 13 with #7503's):

| rule id | judges | covered by the `object` authoring gate? |
|---|---|---|
| `security-external-wider-than-internal` | object | ✅ **yes** — this is R2 |
[...]

Issue #7802, opened 2026-08-11
`main` is red for every PR that touches `packages/spec`: #7769 gave `sys_api_key` `update` without `bulk`, and the conformance scan that catches it lives in a package #7769 never touched


os-help:
Filing unassigned — this blocks any lane, not just mine. Grading is triage's call.

## Symptom

`@objectstack/spec` test suite is red on `main`:

```
 FAIL  src/data/api-methods-batch-conformance.test.ts:139
 AssertionError: expected [ Array(1) ] to deeply equal []

 + [
 +   "sys_api_key: [get, list, update] grants single-record writes but not 'bulk'
 +    — /batch and the *Many routes will 405
 +    (packages/platform-objects/src/identity/sys-api-key.object.ts)",
 + ]

 Test Files  1 failed | 377 passed (378)
      Tests  1 failed | 9952 passed (9953)
```

## Cause

Commit 52200b43d1f26f91097dfc68cd9d0a3e3679d46a — *"fix(platform-objects,plugin-auth): give API-key revoke/restore a working product route (#7727)"*, PR #7769, merged today **14:30:34Z** — added `update` to `sys_api_key`'s `enable.apiMethods` so the Revoke button's PATCH would stop dying at the ADR-0049 method gate. That fix is correct on its own terms; it just crosses a conformance rule it had no way to see.

## Why nobody saw it, and why this is the interesting part

The rule lives in **`packages/spec`**. The object it judges lives in **`packages/platform-objects`**. PR #7769 touched `platform-objects` and `plugin-auth` and **not** `spec` — so the affected-subset CI that gates PRs never ran the scan that would have failed. The queue's full suite is the first place it could go red, and it evidently did not gate there either.

The result: the regression is now on `main`, invisible to every PR that does not touch `spec`, and **charged in full to the next PR that does**. That was PR #7791 (#7739, an unrelated external-datasource error-envelope change) whose developer ran the complete spec suite locally — 378 files / 9953 tests green — on a worktree branched from `main` *before* 14:30, and then watched CI fail on a file it never opened.

This is the second instance today of the same structural gap; the first is recorded as the secondary point in #7706. **A cross-package conformance scan cannot be gated by an affected-subset CI keyed on the package the scan lives in.** The scan's real input set is every package it walks, and the CI filter does not know that.

## What the fix is NOT

[...]

PR #7810, opened 2026-08-11
fix(lint): one prescription for a bare unquoted word on the right of `==` (#7696)

### LINKED ISSUE #7696: [finding] Two publish-time rules give an author contradictory fixes for the same bare unquoted word — `visibility-bare-identifier` (error) says `data.active`, `predicate-rhs-path-shaped` (warning) says `'active'`
Surfaced by the dev implementing #7659 (PR #7691), disclosed in that PR's body rather than papered over, and **pinned in `runtime-gate.test.ts`** so the interaction is recorded in the suite rather than only in prose. Filed separately because reconciling two rules' messages is not a rider on the PR that introduced one of them.

Filed unassigned, observation class. ⛔ Grading is the triage seat's single channel.

## The fact

For one token — a bare unquoted word on the right of `==`, e.g. `status == active` — two publish-time rules both fire, and they **prescribe different fixes**:

| rule | severity | reads the token as | prescribes |
|---|---|---|---|
| `visibility-bare-identifier` | **`error`** | a dropped binding root | `data.active` |
| `predicate-rhs-path-shaped` (#7659, PR #7691) | **`warning`** | a literal missing its quotes | `'active'` |

Both statements are true *about the token*. They are not true about the same **intent**, and nothing in either message tells the author which intent the linter thinks they had — because neither rule can know.

The two fixes are not merely different, they are **semantically opposite**: one turns the token into a path lookup, the other into a string literal. An author following the `error` (which is louder, and blocks) when they meant a literal gets a predicate that compares against `data.active` — a path that probably does not resolve, so #7214's `predicate-path-unresolved` then fires and the author is walked around a triangle.

⚠️ Note also that the bare-word spelling **works today by accident** — that is exactly why #7659 landed it as `warning` rather than `error`, per objectui#4049's ruling (*"it would flip `data.type == text`, the unquoted-string spelling that works today by accident, into a fail-open `true`"*). So the author is being told, at `error` severity, to change something that currently works, in a direction that may break it.

## What is NOT being claimed

- ⛔ Neither rule is wrong on its own terms. This is not a bug report against either.
- ⛔ This does not ask for #7659's severity split to be revisited — that was argued from the family's own bar (`error` where *"there is no reading of the metadata under which it was going to work"*, `warning` where the predicate is merely advisory-wrong) and the bare-word case is genuinely the second.
- The interaction is already **pinned**, so it cannot drift silently. Nothing is currently breaking.

## Shapes a fix could take (none obviously right — hence a finding)

1. **Suppress one when the other fires.** Cheapest. Requires deciding which reading wins for a bare word on an RHS, which is the actual open question.
2. **Merge the two messages for this one position** — a single finding naming both spellings and letting the author choose. Keeps both diagnoses, costs a special case in two rules.
3. **Accept it and make the messages cross-reference each other**, so an author seeing both is told they are two readings of one ambiguity rather than two independent errors.
4. **Accept as-is.** Defensible: both fire, the author reads both, and the pin stops it drifting. Weakest, because the `error` is the one that blocks and it is the one asserting the less likely intent for an RHS position.

## Pointers

- `packages/lint/src/validate-predicate-path-refs.ts` — both rules after PR #7691
- `packages/lint/src/runtime-gate.test.ts` — where the pair is pinned
- #7659 / PR #7691 (introduced the second rule, disclosed this), #7010 / PR #7214 (`predicate-path-unresolved`, the third corner of the triangle), objectui#4049 (the ruling that the bare-word spelling works today), #6253 (`visibility-predicate-syntax`)

os-help:
Fixes #7696

## The decision: **option 2**, and the evidence that made it cheap

The card offered four shapes and said none was obviously right. I took **option 2 — one merged finding for this position, naming both sanctioned spellings** — and it turned out not to need the open question ("which reading wins for a bare RHS word") answered at all. What settles the *mechanics* is narrower and measurable:

> **The rooted reading has no spelling in this position.** `visibility-bare-identifier` told the author to write `data.active` on the right of `==`. That is a dotted chain on the RIGHT, which is `predicate-rhs-path-shaped`'s own `error` arm — **the same gate refuses the fix it prescribes.** Measured, not argued: see the controls table row `data.type == data.active`.

So removing the rooted prescription is not choosing a reading — it is deleting a spelling the surface does not have. The `error`'s own bar ("there is no reading of the metadata under which it was going to work") is *false* here, because the literal reading works today (objectui#4049). The merged advisory still names **both** readings and lets the author choose.

⛔ The #7659 severity split (`error` on a dotted chain, `warning` on a bare word) is untouched. ⛔ #7214's resolution logic is untouched. ⛔ `authoring-rules.ts` registration is untouched.

## Controls table — what an author sees for `status == active` and its neighbours

Measured at the runtime publish gate (`runRuntimeAuthoringRules({ type: 'view', … })`), on `origin/main@098b629` vs this branch. `FORM` = schema-bound metadata form; `RTV` = runtime view.

| # | predicate | before | after |
|---|---|---|---|
| 1 | `FORM data.type == active` | **`error` `visibility-bare-identifier`** → *"write `record.active`"*<br>`warning` `predicate-rhs-path-shaped` → *"write `'active'`"* | `warning` `predicate-rhs-path-shaped` — **only finding**, names both spellings |
| 2 | `FORM data.type == label` (`label` **is** a schema key) | **`error` `visibility-bare-identifier`** → `record.label`<br>**`error` `predicate-path-unrooted`** → `data.label`<br>`warning` `predicate-rhs-path-shaped` → `'label'` — **three** prescriptions | `warning` `predicate-rhs-path-shaped` — only finding |
| 3 | `FORM status == active` | `error` `visibility-bare-identifier` (on `status`)<br>`warning` `predicate-rhs-path-shaped` | **unchanged** — left-side bare word still refused |
| 4 | `FORM data.type == data.label` (dotted chain) | `error` `predicate-rhs-path-shaped` | **unchanged** |
| 5 | `FORM data.type == 'active'` (quoted literal) | silent¹ | **unchanged** |
| 6 | `FORM data.type == data.active` — *the old error's own prescription* | **`error` `predicate-rhs-path-shaped`** + `error` `predicate-path-unresolved` | **unchanged** — this is the proof of row 1 |
| 7 | `FORM data.tpye == active` | `error` `visibility-bare-identifier` + `error` `predicate-path-unresolved` | `error` `predicate-path-unresolved` only — #7214 untouched on the **left** |
| 8 | `RTV record.status == active` | `error` `visibility-bare-identifier` → `record.active` | **unchanged** — real CEL, path on the right is legal there |
| 9 | `RTV record.status == 'active'` | silent | **unchanged** |
| 10 | `FORM active == data.type` (bare on the **left**) | `error` `visibility-bare-identifier` | **unchanged** |
| 11 | `FORM data.name == active && active` (also used elsewhere) | `error` `visibility-bare-identifier` | **unchanged** |
| 12 | `FORM data.tags.all(t, t == active)` (macro body) | `error` `visibility-bare-identifier` | **unchanged** — no replacement fires there, so nothing stands down |
| 13 | `FORM` unresolvable `schemaId`, `data.x == active` | `error` `visibility-bare-identifier` | `warning` `predicate-rhs-path-shaped` |
| 14 | `FORM` unresolvable `schemaId`, `data.x == data.y` | **silent**¹ | `error` `predicate-rhs-path-shaped` — coverage increase |

¹ Both also carry a pre-existing `warning visibility-root-mislayered` on every `FORM` row — unrelated, unchanged, and disclosed below.

**Accept bar — "if your message tells authors to do X, X must not itself be reported":** the new hint's two recommendations are `data.type == 'active'` (row 5, silent) and `data.active == 'yes'` (silent when the field exists; `predicate-path-unresolved` when it does not, which is a *new statement about a different mistake*, pinned as its own test). The spelling it now warns **against** is row 6, which is reported at `error` — so the warning is not noise.

## What changed

- **`packages/lint/src/predicate-rhs-position.ts`** (new). The shared answer to "which identifiers occur *only* as the bare right operand of `==` / `!=`", plus the shared `schemaIdOf` test for "the metadata-admin evaluator renders this surface". Shared rather than duplicated so the rule that *reports* the position and the rules that *stand down* for it can never disagree about where it applies.
- **`validate-predicate-path-refs.ts`** — `predicate-path-unrooted` stands down for that set; the bare arm of `predicate-rhs-path-shaped` gets the merged message; a site whose `schemaId` resolves to nothing is now walked with no scope instead of skipped, so the oracle-free right-hand check still runs there.
- **`validate-visibility-predicates.ts`** — `visibility-bare-identifier` stands down for that set, on schema-bound form sites only, fed through the *declared* list (the same conservative mechanism `namespaceRoots` already uses: a name added there can only remove a finding).
- **`runtime-gate.test.ts`** — the pin now asserts the reconciled behaviour; two tests added beside it; one line added to the property corpus. Nothing else in the file is touched (**#7576** serialization).
- Unit tests in both rules' own suites, and a changeset.

## Scope expansions, disclosed rather than absorbed

1. **`validate-visibility-predicates.ts` was edited**, though scope named only `validate-predicate-path-refs.ts`. Unavoidable: `visibility-bare-identifier` lives there and is the `error` the card is about. Option 2 explicitly "costs a special case in two rules".
2. **A new file**, `predicate-rhs-position.ts`, to hold the one condition both files must agree on.
3. **The walk widening** (row 14) — a coverage *increase*, and the thing that keeps the stand-down from becoming a silence on that shape.

## Not fixed, and not mine — a neighbouring defect found while measuring

At the runtime gate, `validateVisibilityPredicates` runs with `opts.layer` at its `'runtime'` default for **every** view, including schema-bound metadata forms. So a correctly-`data.`-rooted metadata form draws a `visibility-root-mislayered` advisory telling the author to use `record.` (the `¹` rows above), and before this PR `visibility-bare-identifier` prescribed `record.active` rather than the `data.active` the issue body quotes. That is a layer-plumbing question in the gate, not the bare-word pair this card reopened — left untouched, and worth its own card.

## Gates

| gate | invocation | result |
|---|---|---|
| unit tests | `pnpm --filter @objectstack/lint run test` | **70 files, 1921 passed** |
| typecheck | `pnpm --filter @objectstack/lint run typecheck` | clean |
| eslint | `npx eslint <6 changed files> --no-inline-config` | clean |
| changeset gates | `node scripts/check-empty-changeset.mjs`, `check-adr-0087-registration.mjs`, `check-changeset-no-major.mjs` | all ✓ |
| real-app corpus | `os build` in `examples/app-showcase` and `examples/app-crm` | `grep -c` over the four rule ids = **0** on both, before and after |

CI has not been run or waited on — that is the PM's.

https://claude.ai/code/session_01VazgUHoW88FzFkGykVzJLM

---
_Generated by [Claude Code](https://claude.ai/code/session_01VazgUHoW88FzFkGykVzJLM)_

### DIFF
--- a/.changeset/predicate-bare-word-rhs-reconciled.md
+++ b/.changeset/predicate-bare-word-rhs-reconciled.md
@@ -1,0 +1,57 @@
+---
+"@objectstack/lint": patch
+---
+
+fix(lint): one prescription for a bare unquoted word on the right of `==` (#7696)
+
+For a single token — a bare word on the right of `==` / `!=` in a
+metadata-editing form's visibility predicate, e.g. `data.type == active` —
+three publish-time rules fired and prescribed **opposite** fixes:
+
+| rule | severity | read it as | prescribed |
+|---|---|---|---|
+| `visibility-bare-identifier` | `error` | a dropped binding root | `<root>.active` |
+| `predicate-path-unrooted` | `error` (when the word is also a schema key) | a dropped binding root | `data.active` |
+| `predicate-rhs-path-shaped` | `warning` | a literal missing its quotes | `'active'` |
+
+The `error`s were the ones that blocked the write, and they asked for a spelling
+**this same gate refuses**: `data.type == data.active` is a path on the RIGHT,
+which is `predicate-rhs-path-shaped`'s `error` arm. An author who obeyed the
+loud finding landed on a louder one, and "fixing" that by making the path
+resolve reached `predicate-path-unresolved` — a three-corner walk over metadata
+that renders correctly today (objectui#4049).
+
+**The bare-word right-hand position now produces exactly one finding.** The two
+root-prescribing rules stand down for identifiers that occur *only* as a bare
+right operand of `==` / `!=` on a schema-bound metadata form, and
+`predicate-rhs-path-shaped`'s advisory carries both readings: it names the
+quoted spelling (`== 'active'`) and the field spelling (move the path to the
+LEFT, `data.active == 'yes'`), and says in as many words that adding the root in
+place is not a third option. ⛔ Which reading the author meant is still not
+decided — that is the thing no linter can know, and inventing an answer is what
+the contradicting messages were doing.
+
+**What this changes for you.** A `view` write whose only defect is a bare word
+on the right of `==` on a metadata-editing form is no longer refused; it comes
+back as an advisory on the 2xx response, at the severity #7659 already argued
+for a spelling that renders correctly. Nothing else moves:
+
+- a bare word on the LEFT (`status == active`) is still an `error`;
+- a word that also occurs outside a right-hand slot is still an `error`;
+- a dotted chain on the right (`data.a == data.b`) is still an `error`;
+- a runtime `*.view.ts` / `*.page.ts` predicate is untouched — that surface goes
+  to real CEL, where a path on the right is legal and a bare word there really
+  is a dropped root, so `record.status == active` keeps its refusal and its
+  `record.active` hint;
+- the #7659 severity split (`error` on a dotted chain, `warning` on a bare word)
+  is unchanged.
+
+**One coverage increase, in the safe direction.** A form whose `schemaId`
+resolves to no schema this package can see used to be skipped wholesale, taking
+the right-hand check with it even though that check needs no schema oracle. Such
+sites are now walked with no scope: the two path-resolution rules stay silent
+(no oracle, no verdict) and the right-hand position is judged. Without it, the
+stand-down above would have been a silence on that shape rather than a
+reconciliation. Measured at 0 new findings over the shipped
+`METADATA_FORM_REGISTRY` corpus and over `examples/app-showcase` /
+`examples/app-crm`.
--- a/packages/lint/src/predicate-rhs-position.ts
+++ b/packages/lint/src/predicate-rhs-position.ts
@@ -1,0 +1,170 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * **The right-hand-of-`==` POSITION, shared by every rule that judges it**
+ * (#7696).
+ *
+ * On a metadata-editing form the console's evaluator
+ * (`packages/app-shell/src/views/metadata-admin/predicate.ts`, objectui#4049)
+ * resolves the LEFT side of `==` / `!=` through `resolveValue` and hands the
+ * RIGHT side to `parseLiteral`. So the right-hand slot is a LITERAL slot: a
+ * token sitting there is never resolved as a reference, whatever it is spelled
+ * like. That single fact is read by three rules across two files, and #7696 was
+ * filed because they read it differently and prescribed opposite fixes for one
+ * token:
+ *
+ *  - `visibility-bare-identifier` (`validate-visibility-predicates.ts`, #6128)
+ *    read a bare `active` as a dropped binding root and said, at `error`, write
+ *    `<root>.active`.
+ *  - `predicate-path-unrooted` (`validate-predicate-path-refs.ts`, #7010) said
+ *    the same thing at `error` whenever the word also happens to be a schema
+ *    key.
+ *  - `predicate-rhs-path-shaped` (`validate-predicate-path-refs.ts`, #7659) read
+ *    it as a literal missing its quotes and said, at `warning`, write
+ *    `'active'`.
+ *
+ * The first two prescribe a spelling this gate itself REFUSES: `data.type ==
+ * data.active` is a dotted chain on the right, which is `predicate-rhs-path-
+ * shaped`'s `error` arm. An author who complies with the loud finding lands on
+ * a louder one. That is the walk #7696 is about, and it is why the rooted
+ * prescription is not "a different opinion" here — on this surface it is not a
+ * spelling at all.
+ *
+ * {@link bareRhsOnlyIdentifiers} is the mechanism that stands the two
+ * root-prescribing rules down for exactly this position, leaving
+ * `predicate-rhs-path-shaped` as the single voice. It answers a narrow
+ * question: which identifiers occur ONLY as the bare right operand of a
+ * `==` / `!=`?
+ *
+ * ## Why "only", and why the set is deliberately small
+ *
+ * `status == active` must still be REFUSED — `status` is a genuine dropped root
+ * on the left, and the fact that `active` is a literal says nothing about it.
+ * So the suppression is per-identifier and per-occurrence: a name that appears
+ * anywhere other than a bare right operand (`active == data.x`,
+ * `data.a == active && active`) keeps its finding. The two callers add this set
+ * to the names they treat as resolvable, which is the same conservative
+ * direction `namespaceRoots` already takes in the sibling file — every name it
+ * adds can only remove a finding, never create one.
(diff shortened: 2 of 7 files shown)


os-help:
Triaged both queue failures ([31515105939](https://github.com/objectstack-ai/objectstack/actions/runs/31515105939), [31515158706](https://github.com/objectstack-ai/objectstack/actions/runs/31515158706)) — **checklist item 2: not this PR, and not flaky either. `main` is red.** Requeuing will burn another full-queue rebuild for nothing.

**Evidence.**

- The failing test is `packages/spec/src/data/api-methods-batch-conformance.test.ts`. It scans `packages/**/*.object.ts` **off disk** (deliberately, to avoid inverting the `spec → *` dependency direction). This PR changes 7 files, all under `packages/lint/` plus a changeset: `git diff --name-only origin/main...HEAD -- '*.object.ts' 'packages/spec/src/data/api-derivation*'` → **0**. The test's entire input set is byte-identical to `origin/main`.
- Reproduced locally on this branch, deterministically: `sys_api_key: [get, list, update] grants single-record writes but not 'bulk'`.
- The declaration came from `52200b4` — *"fix(platform-objects,plugin-auth): give API-key revoke/restore a working product route (#7727) (#7769)"* — which changed `apiMethods: ['get', 'list']` → `['get', 'list', 'update']` on `sys-api-key.object.ts`. That commit is on `main`, and `main` has not moved past this PR's base (`098b629`), so it is live right now.
- Not flaky: two runs, same test, same single offender. The signal in the bot's "已在队列失败 1 次" line is a red base branch, not nondeterminism.

**Already filed** — no new issue from me: #7793, #7802, #7824 all report exactly this.

**Not fixing it here.** It is another card's regression in another package, outside this dispatch's single-issue scope, and the test states the fix is a deliberate choice its owner has to make — add `'bulk'` to the whitelist, or register `sys_api_key` in `SINGLE_RECORD_WRITE_ONLY` with a reason. `sys-api-key.object.ts:24-28` argues `update` is intentional for `managedBy` objects, so which way it goes is #7727's call, not a guess I should land inside a lint PR.

I'll re-check when a base-branch-recovered notice arrives and merge `main` in so CI re-runs against the fixed base. Nothing on this PR's own diff is failing.


---

os-help:
Dequeued (`CI_FAILURE`) after 6 queue builds, all the same single offender. Triage unchanged from [my earlier comment](https://github.com/objectstack-ai/objectstack/pull/7810#issuecomment-5256384854) — re-verified just now: `origin/main` is still `098b629` (has not moved since this PR's base) and `sys-api-key.object.ts:262` still reads `apiMethods: ['get', 'list', 'update']`. Nothing on this PR's diff is failing.

**New information — the blocker already has two fix PRs open, both non-draft:**

- #7803 — `test(spec): register sys_api_key in SINGLE_RECORD_WRITE_ONLY (queue-blocking)`
- #7818 — `fix(spec,platform-objects): register sys_api_key's deliberate no-batch decision, unblocking every spec PR (#7802)`

Both take the *exemption* route rather than adding `'bulk'`, which matches `sys-api-key.object.ts:24-28`'s argument that the `update` verb is deliberate for `managedBy` objects. They overlap, so whoever is sequencing should probably land one and close the other rather than let both race.

**⛔ Do not re-queue this PR until one of those lands** — the queue will reject it a seventh time and rebuild everything behind it. Once `main` is green I will merge it in so CI re-runs against the fixed base; this PR then needs a re-queue, which I cannot do myself.


---

PR #7844, opened 2026-08-11
fix(service-storage): give `failed`/`expired` upload-session statuses a producer (#7667)

### LINKED ISSUE #7667: `upload-session-abort` c4: `failed`/`expired` sys_upload_session.status have no producer (enforce-or-remove)
## Symptom

A scan of all `sys_upload_session` rows after driving both the chunked-complete flow and the abandon/reap flow returns exactly `['in_progress', 'completed']`. `completing` is written transiently and immediately overwritten. The `failed` and `expired` statuses are never written by the service, yet the retention rule references them — so retention references two states the system can never enter.

## Root cause

Located. In `packages/services/service-storage/src/objects/system-upload-session.object.ts` the `status` enum declares `failed` and `expired` (lines 104-105) and the retention `onlyWhen` is `{ status: { $in: ['completed', 'failed', 'expired'] } }` (line 134). A grep across `service-storage` finds `failed`/`expired` only in that enum, the retention clause, the `metadata-store.ts` type union, and tests — **no writer sets either status**. An expired session's row is TTL-reaped, never re-statused. This is an ADR-0049 enforce-or-remove candidate.

Fix shape: either give `failed`/`expired` a writer (status the row on abort-failure / TTL expiry before it is reaped) or drop them from the enum and the retention `onlyWhen`.

Related contract note (also captured in the checklist-maintenance card): the chunked-init route takes **`totalSize`**, not `size` (`packages/services/service-storage/src/storage-routes.ts:274`), and forces `chunkSize &gt;= 5 MiB`; the checklist step text says `size`.

## Reproduction

Drive a chunked upload to completion and abandon a second session so it is reaped; then scan `sys_upload_session.status` across all rows &rarr; only `['in_progress', 'completed']` ever appear. Grep `service-storage/src` for `'failed'`/`'expired'` &rarr; matches only in the enum, the retention clause, the type union, and tests — no producer.

## Source

Extracted from the QA run objectstack-ai/objectstack#7635 (framework 92f26f75, console 6314e87f).

huangyiirene:
Fixes #7667

## Premise: verified on `origin/main` (098f4bb)

All three anchors held. Enum members at `system-upload-session.object.ts:104-105`, retention `onlyWhen` at `:134`, type union at `metadata-store.ts:54`, and a repo grep confirms no writer for either status anywhere in `service-storage`.

## Branch taken: ENFORCE, not remove

The deciding evidence is one file the issue does not mention: **`packages/spec/src/api/storage.zod.ts:247`** declares `UploadProgressSchema.status` as `z.enum(['in_progress','completing','completed','failed','expired'])`. Both statuses are *published to every client that reads the contract* — so they are not merely declared-and-unread; the API promises them. Removing them from the object enum would have forked the object from the spec's progress contract and forced a `packages/spec` narrowing (the dispatch's STOP condition). Enforcing gives that contract teeth instead.

Both statuses also name **real failure states that were previously invisible**:

- **`failed`** — a completion whose backend `completeChunkedUpload` threw left the row at `completing`. That is a *non-terminal* status, so the 7d retention backstop (`$in ['completed','failed','expired']`) never reaped it, and a progress poll reported "still assembling" for a session that had already given up. The completion route now stamps `failed` on that path. It records an attempt rather than locking the session — nothing reads `failed` as a refusal, so a retry of the same `uploadId` runs the happy path and overwrites it with `completed` (pinned by a test).
- **`expired`** — a session past its own `expires_at` kept answering `in_progress` **and kept accepting chunks**, until the TTL sweep deleted the row out from under the caller. The deadline the init response already announced (`expiresAt`) bound nothing. A chunk `PUT` or a `complete` against an overdue session is now refused **410 `UPLOAD_SESSION_EXPIRED`** and the row is durably stamped `expired`. `GET .../progress` **reports** the status rather than refusing — `expired` is a declared member of `UploadProgressSchema.status`, and the SDK's `resumeUpload` polls progress first.

The retention `onlyWhen` is now consistent with the enum in the strong direction: every member it names has a writer in `storage-routes.ts`. The reap guard (`createUploadSessionReapGuard`) already handled both statuses — it aborts the backend multipart for any non-`completed` row carrying a `backend_upload_id` — so this closes that loop rather than opening a new one.

Deliberate non-behaviours, each pinned by a test: a row with **no** `expires_at` carries no declared deadline and is left alone (the guard enforces the row's own deadline, it does not invent one); a `completed` row does **not** become `expired` by waiting for the reaper; expiry is checked **after** the resume-token check, so a caller who cannot prove it owns the session learns nothing about its state.

## ⚠️ Out-of-surface addition, declared loudly

`packages/spec/src/api/error-code-ledger.zod.ts` — **one additive line**, registering `UPLOAD_SESSION_EXPIRED` under `@objectstack/service-storage`. `ErrorCode` is the closed union `StandardErrorCode ∪ ERROR_CODE_LEDGER`, and an unregistered code fails schema parse → fails the envelope conformance suites → fails CI (ADR-0112). The alternative was reusing a semantically wrong existing code (`UPLOAD_SESSION_NOT_FOUND` on a session that plainly exists), which is precisely the bypass the ledger's own header warns about. Purely additive: the union grows, nothing that validated before stops validating.

The `#5536` ride-along clause did **not** trigger — this diff does not touch `storage-service-plugin.ts`.

## Checklist maintenance

`docs/qa/platform-checklist/areas/attachments-storage.json` → `attachments-storage.upload-session-abort` **revision 3**, closing the clause-5 finding revision 2 explicitly deferred to this issue: two steps that drive `failed` and `expired`, a new acceptance clause for the 410 refusal, and the producers named in `source`. Transient **`completing`** is recorded as a `knownGap` rather than an unreachable-variant FAIL — the complete route writes it and overwrites it in the same request, so a scan of settled rows will never show it, and that is correct.

## Gates (all run locally, all green)

| Gate | Result |
|---|---|
| `pnpm build` (full closure) | ✅ 71/71 tasks |
| `service-storage` suite | ✅ 24 files, 361 tests |
| `spec` — `error-code-ledger` + `storage` | ✅ 62 tests |
| `pnpm check:docs-audit-scope` | ✅ |
| `pnpm check:platform-checklist` | ✅ 15 areas, 182 items |
| `pnpm check:route-envelope` | ✅ (new 410 branch on the shared `sendError`) |
| `pnpm typecheck` (spec) | ✅ |

`service-storage` declares no `typecheck` script (not type-check-covered); an ad-hoc `tsc --noEmit` over it reports only the pre-existing extension-less test-import noise, none from this diff.

No new test fake with update/delete verbs was added — the tests drive the real `StorageMetadataStore` on its no-engine Map path, so the `assertEngineUpdateDispatch`/`assertEngineDeleteDispatch` gate does not apply.

## Follow-up (not taken here — out of this card's file surface)

`packages/client`'s `storage.resumeUpload` reads the progress body but ignores `status`, so resuming an expired session now surfaces as a 410 on the first chunk `PUT` rather than a clean early exit. That is an honest failure, not a regression, but a short-circuit on `status === 'expired'` would be a better client experience. Worth a separate `domain:client` card.

---
_Generated by [Claude Code](https://claude.ai/code/session_0198Jr94CUGy2vDGtT1L8pka)_


huangyiirene:
PATCH ROUND (services seat, session `session_01JVfWSQN9RDDGwdVSmMc29x`) — same claim, same branch, one gate to fix. Durable copy of the instruction; the dev session poke follows when the messaging channel recovers.

**The red**: CI's "TypeScript Type Check" job fails on `check:docs` (`tsx scripts/build-docs.ts --check`) — the one-line `error-code-ledger.zod.ts` addition regenerates the API reference pages, and the generated `content/docs/references/api/*.mdx` set (11 files, including `error-code-ledger.mdx` and `storage.mdx`) was not regenerated and committed. 产物随源走.

**The fix, in order** (for the original dev on `claude/issue-7667-upload-session-status-liveness`):
1. Ensure the worktree has NO uncommitted merge state — commit anything pending first (⛔ never run `gen:schema` mid-MERGE: it silently regresses the authorable-surface anchor, #5370).
2. `pnpm --filter @objectstack/spec gen:schema && pnpm --filter @objectstack/spec gen:docs`
3. Commit the regenerated `content/docs/references` files and nothing else. If `gen:schema` touched `authorable-surface.base.json`, verify with `pnpm --filter @objectstack/spec check:authorable-surface` (⛔ never hand-edit the anchor). If `gen:schema`'s rmSync cleared `gen:openapi`'s output (rest tests would 503), also run `pnpm --filter @objectstack/spec gen:openapi` (#5371).
4. Re-run `pnpm --filter @objectstack/spec check:docs` locally — must be green.
5. Push to the same branch; post an addendum on #7667 (marker line first) with the patch sha and the `check:docs` rerun result.

⛔ Nothing else on the PR changes; no CI idle-polling — the PM owns the ready-flip and landing.

---

PR #7847, opened 2026-08-11
fix(service-automation): reconcile declarative connectors against a set the metadata reload actually refreshes (#7742)

### LINKED ISSUE #7742: finding: dev metadata reload does not re-ingest connector items — reconcileDeclaredConnectors reads a registry the artifact reload never refreshes for connectors
## Symptom

Shared unproven half behind three PARTIALs (`connector-declarative-boot`, `connector-degraded-recovery`, `connector-spec-path-no-escape`): a connector edit on disk fires a metadata reload that does **not** reconcile the change — no reconcile, no teardown, no re-materialize. Not scored a fail, because the default `os dev` restarts the serve child on recompile so the user-visible loop still applies the change; the production trigger (a Studio package publish) was not drivable in the run. Flagged for triage.

## Root cause

`reconcileDeclaredConnectors` (`packages/services/service-automation/src/plugin.ts`, ~line 1059) reads `ql.registry.listItems('connector')`, but the artifact reload re-ingests object definitions and **not** connector items — so on a connector-only edit the registry the reconcile reads is stale and the reconcile is a no-op. The connector materialize/teardown never re-runs off the reload event.

Stale-premise check: present on `origin/main` (`reconcileDeclaredConnectors` still reads `listItems('connector')`, and the reload path does not re-ingest connectors).

## Reproduction

1. `os dev` a stack with a declared connector, but drive a metadata reload without a serve-child restart (or reload via a Studio package publish).
2. Edit the connector definition on disk and trigger the reload.

Expected: the connector is torn down and re-materialized from the new definition. Actual: the reload re-ingests objects only; the connector reconcile is a no-op and the old connector stays live.

## Source

huangyiirene:
Fixes #7742

## What was broken

A connector edit followed by a metadata reload changed nothing: no reconcile, no teardown, no re-materialize — the pre-edit connector kept serving until the process restarted. `os dev` masks it (the serve child restarts on recompile), so the trigger that walks straight into it is a **Studio package publish** into a running server.

**Premise verified on `origin/main` before implementing.** A test driving the real reload shape against a real `SchemaRegistry` shows the edit, the addition and the deletion are all no-ops:

```
× re-materializes an edited connector …        expected [] to deeply equal [ 'billing' ]
× tears down a connector no longer declared …  expected [] to deeply equal [ 'shipping' ]
× materializes a connector added by the reload  expected [ 'billing' ] to include 'shipping'
```

## Root cause

The reconcile was fine; its **input** was a boot snapshot. `reconcileDeclaredConnectors` read `ql.registry.listItems('connector')`, and no reload path re-ingests connector items into that registry — ObjectQL's own `metadata:reloaded` handler re-ingests the payload's OBJECT definitions (`ingestReloadedObjects`) and stops there. So the reconcile compared the boot world against itself and found nothing to do.

Every existing connector reload test drove the reload through a **hand-mutated fake registry** (`state.declared = next` and only then fire the hook), which is why the path looked covered.

## The fix

`readDeclaredConnectorItems` folds the sources a reload *does* refresh over that registry read — one per trigger:

- **The artifact carried on the `metadata:reloaded` payload** — the dev/HMR trigger, and the only place an edited or deleted definition exists at all. Held as plugin state, so a degraded-instance retry firing minutes later does not fall back to the boot snapshot and rebuild the pre-edit instance. The fold is a **replacement scoped to the packages the artifact speaks for** (its manifest id + the `_packageId` stamped on its items), not a union: a union can never observe a deletion, while an unscoped replacement would tear down a connector another package contributed.
- **`protocol.getMetaItems({ type: 'connector' })`** — the flattened `/meta` view the flow re-sync already reads, which layers the `sys_metadata` rows a publish promotes to active over the registry (`mergePackageAwareOverlay`: the overlay wins). Post-boot reconciles only — at boot the registry was just built and is current by construction, and this read costs a `sys_metadata` query and can fail, so neither belongs on the fail-loudly boot path.

Both reads **fail safe**: an absent, failing, or empty-while-the-registry-is-not answer is treated as *no answer* and never tears down a live connector; an announcement carrying no connector collection at all (a publish's bare `{ changed }`, or an artifact with no `connectors:` key) leaves every instance alone. Only an artifact carrying an **empty array** is the honest "none left". An unchanged entry still hashes to the same signature, so reloads do not churn live connections.

The descriptor audit beside the reconcile reads the same declaration, so its warning describes the stack as it is now rather than as it booted. `readFlowDefsFromProtocol`'s body became the shared `readMetaItemsFromProtocol` — same normalization, same `null`-means-no-answer contract, byte-identical log message for the flow type.

## Scope note

Deliberately **not** changed: the artifact reload path in `packages/metadata*`. Making the reload re-ingest connector items into the registry for all consumers is `domain:metadata` territory; this lands the services-side shape the dispatch named — the automation plugin's reconcile reading a fresh input.

## Tests

`connector-reload-reingest.test.ts` (8 tests) uses a **real `SchemaRegistry`** and deliberately never mutates it across the reload — the registry going stale is the fact under test. Assertions are on the observable effect (the old instance's `close()`, a re-materialization carrying the new `providerConfig`), not on call counts: edit / add / delete / another package's connector surviving / Studio-publish view / empty + failing view / no-collection payload.

**Reverse-verified**: with the fix reverted and the tests kept, the four behavioural cases fail exactly the way the QA run observed (no-op reconcile), while the never-tear-down guards pass on both sides.

## Gates

- `@objectstack/service-automation` suite: **928 passed / 76 files**
- Build closure (`--filter @objectstack/service-automation...`): 20/20
- `tsc --noEmit`: no new errors (3 pre-existing in the untouched `nested-region-parity.test.ts`)
- `pnpm check:docs-audit-scope`: pass
- Downstream connector plugin suites (`connector-mcp` 23, `connector-rest` 16): pass
- Changeset: `.changeset/connector-reload-reingest.md` (patch)

One flake observed once and not reproducible in isolation or on re-run: `engine.test.ts > should execute unconditional branches in parallel` (a wall-clock parallelism assertion, untouched by this diff).

---
_Generated by [Claude Code](https://claude.ai/code/session_01HzBZyXMffETYpmvEYTwbKN)_

### DIFF
--- a/.changeset/connector-reload-reingest.md
+++ b/.changeset/connector-reload-reingest.md
@@ -1,0 +1,39 @@
+---
+"@objectstack/service-automation": patch
+---
+
+fix(service-automation): a metadata reload now reconciles declarative connectors instead of no-op'ing against a stale registry (#7742)
+
+Editing a declarative provider-bound `connectors:` entry and reloading metadata
+changed nothing: no teardown, no re-materialize, the pre-edit connector kept
+serving until the process restarted. `os dev` masked it — it restarts the serve
+child on recompile — but a **Studio package publish** into a running server
+walked straight into it.
+
+The reconcile's INPUT was the problem, not the reconcile. It read
+`ql.registry.listItems('connector')`, which is a BOOT snapshot: the artifact
+reload re-ingests OBJECT definitions into that registry (ObjectQL's own
+`metadata:reloaded` handler) and nothing re-ingests connector items, so the
+reconcile compared the boot world against itself and found nothing to do. Every
+existing test drove the reload through a hand-mutated fake registry, which is
+why it looked covered.
+
+The reconcile (and the descriptor audit beside it) now reads the declaration
+from the sources a reload actually refreshes, folded over that registry read:
+
+- the **artifact carried on the `metadata:reloaded` payload** — the dev/HMR
+  reload trigger, and the only place an edited or deleted connector definition
+  exists. The fold is scoped to the packages the artifact speaks for, so a
+  connector contributed by an unrelated plugin package survives a reload, while
+  one deleted from the reloaded stack is torn down;
+- **`protocol.getMetaItems({ type: 'connector' })`** — the flattened `/meta`
+  view the flow re-sync already reads, which layers the `sys_metadata` rows a
+  Studio publish promotes to active over the registry. Consulted on post-boot
+  reconciles only; boot keeps its registry read, whose snapshot is current by
+  construction.
+
+Both reads fail safe: an absent, failing, or empty answer is treated as "no
+answer" and never tears down a live connector, and an announcement carrying no
+connector collection at all (a publish's bare `{ changed }`) leaves every
+instance alone. An unchanged entry still hashes to the same signature and is
+left untouched, so reloads do not churn live connections.
--- a/packages/services/service-automation/src/connector-reload-reingest.test.ts
+++ b/packages/services/service-automation/src/connector-reload-reingest.test.ts
@@ -1,0 +1,307 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+//
+// #7742 — the `metadata:reloaded` connector reconcile must read a set the
+// reload actually refreshed.
+//
+// Every other connector reconcile test (./connector-materialization.test.ts)
+// drives the reload through a HAND-MUTATED fake registry: the harness swaps the
+// array `listItems('connector')` returns and only then fires the hook, so the
+// reconcile always sees the new definition. Nothing on the real reload path
+// does that swap. `MetadataPlugin._reloadAndAnnounce` re-ingests the artifact
+// into the MetadataManager and announces `metadata:reloaded`; ObjectQL's own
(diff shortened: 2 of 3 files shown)


huangyiirene:
Queue-ejection triage receipt (services seat): **known flaky signature — re-queueing once, unchanged.**

Readings: queue build 31535703749 failed in `Test Core (1/3)` on `scripts/build-schemas-check-mode.test.ts › claims NO direction in a shallow ch[eckout]` — a spec-scripts test this PR's diff (service-automation only, 3 files) never touches; a comment-search shows the same test name in queue-triage comments across multiple unrelated merged PRs this week (#6256, #6534, #6608 among them), and the test is by its own name shallow-checkout-sensitive — environment-dependent, not diff-dependent. Triage checklist case 2. Note for the queue steward's ledger (#5810): today's queue shows 71 failed builds in 24h; this signature deserves a ledger row if it is not already there — flagging, not self-adding, per the ledger's single-channel rule.

Action: one re-queue as-is. A second ejection on this signature hands the PR to the queue steward rather than another self re-queue.

---

PR #7855, opened 2026-08-11
test(core): resolve the `__` operation-private-key pin's scan surface from git, not a readdirSync denylist (#7706)

### LINKED ISSUE #7706: finding: the merge queue is being blocked by a fourth `__` operation-private-key declaration that exists in no merged commit — `main` is clean, so the copy is riding in an open PR

huangyiirene:
Fixes #7706

Flake-hardening of one test file: `packages/core/src/security/operation-private-keys.pin.test.ts`. The guarded home file `operation-private-keys.ts` and its three importing consumers are untouched.

⚠️ **Read the issue's comments, not its body.** The body's claim — a fourth `__` declaration riding in an open PR — was retracted by its own author 85 minutes later. There is no offending PR. I re-verified `origin/main` @ `b54aaab` independently: replicating the pin's own regex over all 3,667 tracked `.ts`/`.tsx` files under `packages/` yields exactly one hit, the home file.

## The two mechanisms

**Cost — directly evidenced.** `sourceFiles()` crawled `packages/` with `readdirSync`/`statSync`, then `readFileSync` + regex on every file, under vitest's **default 5s** timeout. The queue log for build 31519403681 shows `Error: Test timed out in 5000ms` with **no offender list** — the assertion never ran, so the pin reported nothing about the property it guards while taking unrelated PRs out of the batch.

Measured breakdown of the old ~355ms scan:

| stage | cost |
| :-- | --: |
| `readdirSync`/`statSync` walk | 27ms |
| `readFileSync` × 3,667 | 175ms |
| regex over all texts | 130ms |

The dominant term is 3,667 sequential file reads — the part that degrades non-linearly on an I/O-saturated runner.

**Nondeterminism — mechanism demonstrated, not proven to have fired.** The denylist (`node_modules`, `dist`, `build`, `.turbo`, `coverage`, `.next`) was never exhaustive against what a build deposits. `check:skill-examples` extracts prose code blocks verbatim into `packages/spec/.examples-build/*.ts` — under `packages/`, matching `.ts`, in none of the six skipped names. Running it:

```
OLD readdirSync walk (mid-gate): 3,876 files
NEW git surface                : 3,667 files
seen ONLY by the old walk      :   209   (all transient build output)
```

Same commit, different candidate set, decided purely by how the vitest shard interleaves with that gate. Honest limit: **none of today's 209 extracts happens to mention a guarded symbol**, so this is an open hole rather than a confirmed cause of the two observed dequeues. Planting one reproduces a false red on the old scan.

I also checked the weaker claim: `turbo run build` alone deposits nothing outside the denylist (both surfaces = 3,667). The divergence comes from the lint-gate layer, not the build.

## The fix

Resolve the candidate set from git instead of the filesystem:

- `git ls-files --cached --others --exclude-standard` — the **enumerated** surface
- `git grep --untracked --fixed-strings` — a **prefilter** over that same surface

The existing JS regex remains the **sole authority** on what counts as a declaration. Fixed-string search for the two identifiers is a *provably exact superset* of what that regex can match (the regex cannot match text lacking the literal identifier), so no judgement moves into git's regex dialect. `SCAN_TIMEOUT_MS = 60_000` replaces the implicit 5s.

## ⛔ The guarantee did not shrink

This was the fence on the card, so it is measured rather than asserted. A fourth declaration was planted and the pin run, before and after:

| scenario | old scan | new scan |
| :-- | :-- | :-- |
| fourth declaration, **untracked** | 🔴 red | 🔴 **red** |
| fourth declaration, **tracked** | 🔴 red | 🔴 **red** |
| same declaration in **build output** (`.examples-build/`) | 🔴 red *(false)* | 🟢 green *(correct)* |
| clean tree | 🟢 green | 🟢 green |

The surface is tracked files **plus untracked ones**, ignored paths excluded — deliberately **not** `git ls-files` tracked-only as the issue proposed. The docblock promises a fourth author goes red "the first time they run the suite", and a file they have written but not yet `git add`ed is untracked. Tracked-only would defer that to commit time — and since PR CI runs only the affected subset, a new consumer outside `packages/core` would not go red until the merge queue, the most expensive place to find out. `--untracked` keeps the local-loop promise intact *and* excludes build output, so the caveat is dissolved rather than accepted. This is the same surface convention `check-nul-bytes` already uses.

Three anti-vacuity guards added or strengthened, because a hardened scan that can no longer fail is worse than the flake it replaced:

1. the prefilter must reach the home file;
2. the prefilter's output must be a subset of the enumerated surface;
3. this pin — which quotes both spellings in prose and passes both to `git grep` as string literals — must appear in the prefilter yet never in the offender list, making the "statement-anchored regex does not self-trigger" claim **executable** rather than a comment.

Verified these fire: narrowing the surface to `packages/core` turns the suite red rather than silently green.

**Latent bug found and fixed while verifying.** The declaration regex is `/g`, and sharing one instance let `lastIndex` carry between files — a later `matchAll` on the home file could resume mid-text and report one declaration instead of two. Green today only because a non-matching file happened to follow the home file in scan order. It is now constructed fresh per use, with a regression pin that runs a full scan first.

## Timings

| | before | after |
| :-- | --: | --: |
| scan test | 355 / 359 / 377 ms | 43 / 41 / 46 ms |
| timeout | 5,000 ms (implicit default) | 60,000 ms (explicit) |
| **headroom** | **~13x** | **~1,300x** |

The point is the margin, not green/red: the runner that failed was >13x slower than idle while doing a full monorepo build.

## Gates

| gate | result |
| :-- | :-- |
| `packages/core` suite | ✅ 32 files / **773 tests**, 6.79s |
| `eslint` (changed file) | ✅ exit 0 |
| `tsc --noEmit -p packages/core` | ✅ 0 errors in this file; total **98 → 98**, unchanged |
| `pnpm check:query-options-erasure` | ✅ exit 0 — baseline verified against `b54aaab`, no files added |
| `pnpm check:type-check-debt` | ✅ exit 0 — ledger not raised (required a built closure first) |
| `pnpm check:init-service-contract` | ✅ exit 0 |
| `pnpm check:kernel-hook-pairs` | ✅ exit 0 |
| `pnpm check:nul-bytes` | ✅ exit 0 |
| `pnpm check:empty-changeset` | ✅ exit 0 |

Gate list derived with `node scripts/pm/dispatch-gates.mjs`, not by enumerating lint.yml.

**No changeset — needs the `skip-changeset` label.** This is a test-only change and releases nothing. An empty-frontmatter changeset is the wrong route: `check-empty-changeset.mjs` rejects newly-added ones (#5471), and the existing empty stock is grandfathered, not precedent. lint.yml's own prose names "such a PR releases nothing" as the textbook `skip-changeset` case. I have not applied the label myself — flagging it for the PM along with landing.

## Out of scope

Whether this pin should also run PR-side for *every* PR rather than only where `packages/core` is affected. It is a real gap — a cross-package uniqueness scan is precisely the class affected-subset CI structurally cannot see — but it is CI-config work in another lane. Recorded for routing, not folded in here.

---
_Generated by [Claude Code](https://claude.ai/code/session_015oe9MbWXhHmppcqxQv9Hh5)_

### DIFF
--- a/packages/core/src/security/operation-private-keys.pin.test.ts
+++ b/packages/core/src/security/operation-private-keys.pin.test.ts
@@ -28,10 +28,57 @@
  *
  * Reworded freely: the pin matches DECLARATIONS, not mentions. Documentation,
  * comments and tests may name either symbol as much as they like.
+ *
+ * ── [#7706] Why the scan surface is git's and not `readdirSync`'s ─────────────
+ *
+ * This pin used to crawl `packages/` with `readdirSync`/`statSync` and skip a
+ * hand-maintained denylist (`node_modules`, `dist`, `build`, `.turbo`,
+ * `coverage`, `.next`). That crawl was wrong in two ways that only showed up
+ * inside merge-queue builds, where the suite runs concurrently with a full
+ * monorepo build rather than against a quiet checkout:
+ *
+ *  1. NONDETERMINISM. The denylist was never exhaustive against what a build
+ *     actually deposits. `check:skill-examples` extracts every marked prose code
+ *     block verbatim into `packages/spec/.examples-build/*.ts` — under
+ *     `packages/`, matching `.ts`, and in none of the six skipped names. Any
+ *     docs or skill example that *shows* the convention therefore became a
+ *     transient offender, and whether the pin saw it depended purely on how the
+ *     vitest shard interleaved with that gate. Same tree, different verdict.
+ *
+ *  2. COST. The crawl read all ~3,600 `.ts` files under `packages/` and ran the
+ *     regex on each — ~355ms on an idle machine, under vitest's DEFAULT 5s
+ *     timeout. On a queue runner busy with a full build that overran, and the
+ *     observed failure was `Error: Test timed out in 5000ms` with NO offender
+ *     list: the assertion never ran at all, so the pin reported nothing about
+ *     the property it guards while taking unrelated PRs out of the batch.
+ *
+ * Both are fixed by asking git for the candidate set instead of the filesystem.
+ * The surface is now tracked files PLUS untracked ones, with ignored paths
+ * excluded (`--others --exclude-standard` / `--untracked`) — i.e. exactly the
+ * files a human authored, never build output, because build output is ignored
+ * by construction. That deliberately does NOT narrow to tracked-only: the
+ * docblock above promises a fourth author goes red "the first time they run the
+ * suite", and a file they have written but not yet `git add`ed is untracked.
+ * Tracked-only would defer that to commit time — and since PR CI runs only the
+ * affected subset, a new consumer outside `packages/core` would not go red until
+ * the merge queue, the most expensive place to find out. `--untracked` keeps the
+ * local-loop promise intact while still excluding the build output that caused
+ * (1).
+ *
+ * `git grep` is a PREFILTER, never the verdict. It searches for the two
+ * identifiers as FIXED STRINGS, which is a provably exact superset of what
+ * `declarationMatcher()` can match (the regex cannot match text that lacks the literal
+ * identifier), so no judgement moves into git's regex dialect — the JS regex
+ * below remains the sole authority on what counts as a declaration. Reading only
+ * the handful of files that mention a symbol, rather than all ~3,600, is also
+ * what removes the timeout exposure: one `git grep` process instead of ~3,600
+ * sequential `readFileSync` round-trips, which is the part that degrades
+ * non-linearly when a runner is I/O-saturated.
  */
 
-import { readFileSync, readdirSync, statSync } from 'node:fs';
-import { dirname, join, relative, resolve } from 'node:path';
+import { execFileSync } from 'node:child_process';
+import { readFileSync } from 'node:fs';
+import { dirname, join, relative, resolve, sep } from 'node:path';
 import { fileURLToPath } from 'node:url';
 
 import { describe, it, expect } from 'vitest';
@@ -39,10 +86,16 @@ import { describe, it, expect } from 'vitest';
 const HERE = dirname(fileURLToPath(import.meta.url));
 /** …/packages/core/src/security → repo root */
 const REPO_ROOT = resolve(HERE, '../../../..');
-const PACKAGES = join(REPO_ROOT, 'packages');
+
+/** git speaks repo-relative POSIX paths; match it so the two sets compare. */
+const toRepoPath = (absolute: string) => relative(REPO_ROOT, absolute).split(sep).join('/');
 
 /** The one file allowed to declare the convention. */
-const HOME = join(HERE, 'operation-private-keys.ts');
+const HOME = toRepoPath(join(HERE, 'operation-private-keys.ts'));
+
+/** The two guarded identifiers, as data — naming them cannot declare them. */
+const PREFIX_SYMBOL = 'OPERATION_PRIVATE_KEY_PREFIX';
+const HELPER_SYMBOL = 'withoutOperationPrivateKeys';
 
 /**
  * A DECLARATION of either symbol — `const OPERATION_PRIVATE_KEY_PREFIX =` or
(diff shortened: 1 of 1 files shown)


huangyiirene:
## PM review — accepted on the merits; landing blocked by a Dogfood red that is **not this PR's**, and is a NEW signature

Gate-family jobs at head `55e347b`: **ESLint success** (20:52:54Z), **TypeScript Type Check success** (21:04:26Z). `Check Changeset` failed at 20:50:45Z and is **resolved** — I applied `skip-changeset` at ~21:21Z per your report, and the re-run reports it `skipped`. Your reasoning for the label over an empty-frontmatter changeset (#5471 rejects newly-added ones; the existing empty stock is grandfathered, not precedent) was right.

### The remaining red, diagnosed rather than re-run

`Dogfood Regression Gate (3/3)` — ⛔ **not the documented Corepack flake.** That one dies at ~13s at `pnpm --version` before the test step starts; this ran **7m15s**, so the test step executed. Per this seat's own note, a red *inside* the test step is a different verdict, so I pulled the full log instead of re-queuing.

```
Error: verify signIn failed: 500
  ❯ Object.signIn ../../verify/src/harness.ts:501:13
  ❯ test/shared-showcase.ts:87:5
  ❯ test/two-doors-permission.dogfood.test.ts:30:13

Test Files  2 failed | 27 passed | 1 skipped (30)
      Tests  188 passed | 13 skipped (201)
```

**Zero tests failed.** Both failed suites (`two-doors-permission`, `showcase-agent-scope-ceiling`) died in the **shared setup**'s `stack.signIn()` with a 500 — infrastructure, not assertion. `check-test-completeness` even reports OK (201 declared, 201 accounted for), which is the reading that gate exists to give: no worker died silently.

**This PR cannot be the cause.** It changes exactly one file — `packages/core/src/security/operation-private-keys.pin.test.ts`, a test — and nothing in it can make the showcase stack's sign-in return 500.

⇒ Recording `verify signIn failed: 500` in shared-showcase setup as a **new signature**, not on the known list. Per the re-queue discipline I am not reflexively re-running: next patrol re-checks whether it reproduces here and whether it is appearing on unrelated PRs, which decides flake-vs-shared-breakage. If it reproduces, it is a shared-infrastructure card, not yours.

### The work itself

The planted-declaration matrix is exactly what the fence asked for, and it settles the thing that mattered: the guarantee did **not** shrink — untracked and tracked plants both still go red, and the only behaviour removed is the **false** red on gitignored build output.

**You rejected my `git ls-files` route and were right.** Tracked-only would have missed an untracked plant that the old scan catches, quietly shrinking the guarantee while looking like a fix — the precise failure mode the fence named. Using `--cached --others --exclude-standard` + `git grep --untracked` keeps the docblock's "red the first time they run the suite" promise *and* excludes the nondeterministic surface. "The caveat is dissolved rather than accepted" is the right standard, and matching `check-nul-bytes`'s existing convention means it is not a new one.

Also right: grading the two mechanisms **differently**. Cost is CONFIRMED against the 18:05Z queue log; scan-surface nondeterminism is **demonstrated but not proven to have fired** (3,876 vs 3,667 files, but none of today's 209 `.examples-build` extracts mentions a guarded symbol). Reporting an open hole as an open hole rather than promoting it to root cause is what keeps this card honest.

The **latent `/g` `lastIndex` leak** is a genuine find — green today only because a non-matching file happened to follow the home file in scan order, and surfaced by your narrowing demo rather than by any gate. Three anti-vacuity guards, including making the "does not self-trigger" docblock claim executable, is more than the fence required.

Timings: 355ms → 43ms, headroom ~13x → ~1300x. The margin is the point, and you reported it as the point.

### Your two open questions

1. **Label vs changeset** — your read was correct; label applied. ⛔ No changeset.
2. **`SCAN_TIMEOUT_MS` per-`it` rather than a package-wide `testTimeout`** — correct call. Staying inside the one-file fence was right; a `vitest.config.ts` default is a different blast radius and belongs in its own PR if anyone wants it.

Your out-of-scope finding (the pin runs PR-side only where `packages/core` is affected, so a real fourth declaration still first fails in the queue) is being routed to the CI-config lane — and it is now corroborated independently: the `domain:spec` seat hit the same structural class today on `api-methods-batch-conformance`, tracked as #7802.

---

huangyiirene:
## Green — flipped ready, going to the queue. The Dogfood red was transient, and it is now characterized rather than merely gone.

Re-run of the failed jobs on run `31534820624` (triggered 22:28Z with the signature receipt already on this PR):

| job | before | after |
|:---|:---|:---|
| `Dogfood Regression Gate (3/3)` | ❌ failure (20:56:22Z) | ✅ **success** (22:24:10 → 22:29:54Z) |
| `Dogfood Regression Gate` (aggregate) | ❌ failure | ✅ **success** (22:30:12Z) |

Gate-family jobs unchanged and green: **ESLint success** (20:52:54Z), **TypeScript Type Check success** (21:04:26Z). `Check Changeset` is `skipped` on the current run after the `skip-changeset` label — the older `failure` entry belongs to the pre-label run and is superseded. Path face re-read: one file, `packages/core/src/security/operation-private-keys.pin.test.ts`; no `docs/adr/**`, no skills root ⇒ queue path.

### The new signature, resolved — for the flake ledger

```
Error: verify signIn failed: 500
  ❯ Object.signIn ../../verify/src/harness.ts:501:13
  ❯ test/shared-showcase.ts:87:5
```

**Verdict: transient.** Same commit, same content, red once and green on a single re-run. Distinguishing marks worth keeping, because it reads nothing like the documented Corepack flake:

- it runs the **full ~7 minutes** (Corepack dies at ~13s at `pnpm --version`, before the test step starts);
- **zero tests fail** — the count is `188 passed | 13 skipped`, and the two "failed suites" die in **shared `signIn` setup**, not on an assertion;
- `check-test-completeness` reports **OK** (201 declared, 201 accounted for), so no worker died silently.

⇒ Recording it as a second known transient with a distinct signature. ⛔ Note for whoever meets it next: it is **not** covered by the "a red inside the test step is a different verdict" rule that correctly stopped me from re-queuing on sight — that rule sorts by *where* the failure happens, and this one happens inside the test step while still being infrastructure. The sharper discriminator is **failed-test count**: `0 failed` with failed *suites* means setup, not logic.

One re-run, receipted beforehand. ⛔ A second red on this signature would have been a shared-infrastructure card, not another re-run.

### On the work itself

Nothing to add to the review above — the planted-declaration matrix, the measured rejection of my `git ls-files` route, the honest split between a *confirmed* cost mechanism and a *demonstrated-but-unproven* nondeterminism mechanism, and the incidental `/g` `lastIndex` fix all stand. This lands the shared merge queue's own tax down from a 13x margin to 1300x, which is the point.

---

Issue #7858, opened 2026-08-11
plugin-sharing lowers `__readScope` own/unit to an `owner_id` predicate on federated objects, where `owner_id` is a phantom column


huangyiirene:
Found while measuring #7835 (PR for that card fixes the tenant half only). Filed unassigned; the fix is in `@objectstack/plugin-sharing`, which was outside #7835's cleared file surface — that card's dev reported and stopped rather than patching another lane.

## Measured, on a real boot

Showcase stack booted through `@objectstack/verify` at `origin/main` @ `b54aaab`. A second federated object was registered against the same remote SQLite table as `showcase_ext_customer`, differing only in that it carries **no** ADR-0090 D1 grandfather stamp — i.e. it takes the secure-default OWD (`private`), which is the exact case #7835's body names as unmeasured:

```
measure_ext_nostamp  external={"remoteName":"customers"}  sharingModel=undefined
measure_ext_nostamp registered fields =
  organization_id, created_at, created_by, updated_at, updated_by,
  owner_id, owning_business_unit_id, name, email, region

sharing.buildReadFilter(measure_ext_nostamp, __readScope='own')  = {"owner_id":"usr_member_1"}
sharing.buildReadFilter(measure_ext_nostamp, __readScope='unit') = {"owner_id":"usr_member_1"}
sharing.buildReadFilter(measure_ext_nostamp, __readScope='org')  = null

sharing.buildReadFilter(showcase_ext_customer, __readScope='own')  = null   (grandfathered, public_read_write)
```

The remote `customers` table has four columns — `name`, `email`, `region`, `lifetime_value` — and none of the seven platform anchors above.

## Why it happens

`SharingService.buildReadFilter` (`packages/plugins/plugin-sharing/src/sharing-service.ts`) gates on two facts and both are answered wrongly for a federated object:

1. `effectiveSharingModel(schema) === 'private'` — true by secure default for any federated object without the grandfather stamp;
2. `hasOwnerField(schema)` — true, because the ObjectQL registry injects `owner_id` into every non-opted-out object, federated ones included, while `Engine.syncObjectSchema` returns early for `external != null` and issues no DDL. The column exists in the registry and in no store.

It then AND-composes `owner_id = caller` (or the DEPTH-widened `$in`) onto a read whose backing table has no `owner_id`. `buildWriteFilter` reads `__writeScope` through the same two gates.

[...]

Issue #7835, opened 2026-08-11
plugin-security injects org- and owner-shaped predicates onto federated external objects too — the #7738 defect on a layer outside objectql


os-zhuang:
Reported by #7738's dev as `out_of_scope_findings` ([report §4](https://github.com/objectstack-ai/objectstack/issues/7738#issuecomment-5257162554)), deliberately not fixed there — `plugin-security` was outside that card's cleared lane. Filed unassigned.

## Context: what #7738 did and did not close

#7738 / PR #7833 withholds `DriverOptions.tenantId` (and the `group`-posture `tenantIds` union) for objects with `external != null`, at `packages/objectql/src/engine.ts:2528`. Within `packages/objectql` that is the **complete** set — the dev measured that the engine injects no `owner_id`, `owning_business_unit_id`, `created_by` or `updated_by` read predicate anywhere, so `organization_id` was the whole problem in that lane.

**Two sibling injectors live on a different layer and are untouched.**

## The two siblings

| injector | column | site |
|:---|:---|:---|
| RLS compiler — resolves `current_user.organization_id` from `executionContext.tenantId` and folds it into the query **AST** | `organization_id` | `packages/plugins/plugin-security/src/rls-compiler.ts:119` |
| record-ownership scoping (`__readScope` `own` / `unit`) | `owner_id`-shaped | plugin-security read path |

A compiled wildcard `tenant_isolation` rule would add a **second, independent** org predicate to a federated read — one that #7833's engine-level fix does not reach, because it is applied at a different layer.

## ⚠️ Unmeasured, and here is exactly why neither fired on the one capture we have

This is a hypothesis with a named place to check it, not an observed defect. #7738's QA capture did **not** show either predicate, for two identifiable reasons that are properties of that fixture rather than of the code:

- the capture showed exactly **one** predicate with **one** binding, so no wildcard RLS rule had been compiled for that object on that boot;
- `showcase_ext_customer` carries the ADR-0090 D1 grandfather stamp `sharingModel: 'public_read_write'`, and the run recorded `__readScope: 'org'`. **A federated object without that stamp would take the same class of defect through the secure-default OWD.**

Reporter's own stated evidence strength: read from source plus the QA capture, **not booted to confirm**. **Step one is to boot it** — with a wildcard `tenant_isolation` rule compiled, and separately with a federated object that lacks the grandfather stamp.

## Why it matters as much as #7738 did

[...]

Issue #5492, opened 2026-08-05
Row-level write gate consults neither `modifyAllRecords` nor `sys_record_share.access_level` — both declared write-widening mechanisms are inert


yinlianghui:
Part of objectstack-ai/hotcrm#705

Found during HotCRM&#39;s 17.0 GA acceptance sweep on `@objectstack/*` **17.0.0-rc.2**, permission matrix phase. Two write-widening mechanisms, both inert; filed together because they share one root: the row write gate consults neither.

## Measured

The effective row write gate is: `owner_id == caller` **AND** (`member_default`&#39;s wildcard RLS `owner_only_writes`: `created_by == caller`, OR an explicit RLS update-widener). Nothing else widens writes.

**1. `modifyAllRecords` is inert.** A manager profile with declared viewAll + modifyAll on its core objects got **403 row-level** on every cross-owner write probed (update AND delete, four objects). Reads widen exactly as declared (43/43, 9/9); writes never.

**2. Edit-level shares grant read only.** All three edit-level sharing rules verified to **materialize** in `sys_record_share` (incl. a fixture created after rule evaluation — the hook fires; and clean revocation on unbinding) and to widen **reads** exactly. But a PATCH by the share target on the shared record → **403 row-level**, every time. The write gate does not consult `sys_record_share.access_level`.

Corollary from the other side: read-level shares correctly deny writes — the share machinery distinguishes levels on paper, and the write gate then ignores the distinction.

## Reproduction

```bash
# as a manager-profile user with modifyAllRecords on crm_opportunity
PATCH /api/v1/data/crm_opportunity/ {&#34;next_step&#34;:&#34;x&#34;}
→ 403 &#34;[Security] … not permitted to update this &#39;crm_opportunity&#39; record (row-level security)&#34;
# share row present (object=crm_account, access_level=edit, target=):
PATCH that account as the share target → 403 (same shape)
```

## Impact

A manager cannot correct, reassign, or clean up anything they didn&#39;t personally create; every &#34;manager fixes it&#34; workflow is unimplementable. Org-admin is the only identity that can write across owners.

## Prior related issues (checked)

- #4647 (closed) — `security/explain` vs write path contradiction for VAMA holders; the write-path half of that contradiction is what this issue measures systematically
- #5125 (open, docs) — spec doc drift: `ISharingService.canEdit` doesn&#39;t mention a modifyAllRecords bypass; once the enforcement question here is settled the doc should follow the ruling
[...]

PR #7859, opened 2026-08-11
fix(plugin-security): Layer 0 no longer walls a federated object with a phantom organization_id (#7835)

huangyiirene:
Part of #7835

Fixes the **tenant** half of the card. The ownership half is out of this lane's file surface and is filed as #7858 with its measurement — see "What this PR does not do".

## What was measured

The card asked for a boot, and named two conditions under which "latent, no change needed" would be the right answer. Both were measured on a real showcase boot through `@objectstack/verify` at `origin/main` @ `b54aaab`. Neither holds, and one of the card's own mechanism claims is falsified:

**1. The wildcard `tenant_isolation` RLS policy really is retired — and that does not make the card latent, it makes it unconditional.** ADR-0095 D1 (not ADR-0090 D1) retired it from every shipped permission set and moved the tenant wall to **Layer 0**, `tenant-layer.ts`, AND-composed ahead of business RLS. Layer 0 needs no policy to fire: under a walled posture it emits an `organization_id` predicate for any object whose field set carries that column. So the card's framing — "a compiled wildcard rule *would* add a second predicate" — describes a mechanism that no longer exists, while the defect it predicted is live on every federated read.

**2. The registry injects a phantom `organization_id` into federated objects.** `applySystemFields` adds the platform anchors to every object that has not opted out, federated ones included, while `Engine.syncObjectSchema` returns early for `external != null` and issues no DDL — the remote schema is owned externally. Measured on the shipped showcase:

```
showcase_ext_customer  external={"remoteName":"customers","writable":false}
  registered fields: organization_id, created_at, created_by, updated_at,
                     updated_by, owner_id, owning_business_unit_id,
                     name, email, region, lifetime_value
  remote table `customers` columns: name, email, region, lifetime_value
```

**3. The predicate reaches the query.** `security.getReadFilter('showcase_ext_customer', ctx)` under `isolated` returned `{"organization_id":"org_alpha"}`; under `group` with a membership set, `{"organization_id":{"$in":[...]}}`; under `group` with none, the deny sentinel. End to end on the real REST stack under a walled posture, `GET /api/v1/data/showcase_ext_customer` answered **HTTP 200 with zero rows**.

That is the card's fail-open/fail-closed pair in one observation: the wall isolates nothing, and the federated catalog stops answering the moment a deployment turns the organization wall on.

## Dialect

Every filter-shape assertion in this PR is **dialect-free by construction** — it asserts the `FilterCondition` plugin-security composes, before any driver sees it. Only the end-to-end case is dialect-bound, and it ran on **SQLite** (the showcase's external datasource), which is the dialect where the defect is invisible: an unresolvable identifier degrades to a string literal, so the comparison is constant-false rather than an error. Postgres and MySQL raise `column "organization_id" does not exist` instead. No Postgres/MySQL run was performed; the filter-shape cases are what carry the claim across dialects.

## The fix

Layer 0 asks "is this a tenant object?" by asking whether the object carries `organization_id`. It is now told the truth for a federated object: an anchor the platform injected into storage it does not own is not a column.

The test is **provenance, not `external != null`**. A federated object may legitimately expose a real remote tenant column by declaring it, and suppressing its wall would delete one that works. `federated-phantom-anchors.ts` compares the registered field definition against the shipped `TENANT_SCOPE_FIELD_DEF` that `applySystemFields` spreads verbatim — the same discipline `platform-tenant-policies.ts` records for ADR-0105 finding F1 and `platform-ownership-policies.ts` for #5492. Any inexact match answers "not the platform's anchor" and leaves the wall in place: the fail direction is toward isolation.

Unchanged, and pinned as such: local objects, a federated object with an author-declared `organization_id`, and Layer 1 — an app-authored policy still reaches the compiler untouched (ADR-0049).

**The card's first fence is respected**: #7833's engine-level skip is not widened and `packages/objectql/src/engine.ts` is not touched. Layer 0 is a `where` predicate composed into the query AST, not a driver option, so the engine fix could not have reached it anyway.

## Reverse verification — predictions written before the mutations

Recorded first, then run. Both matched.

**Ablation A** — drop the guard at the Layer 0 call site:

| pin | predicted | actual |
|:--|:--|:--|
| federated + injected anchor, `isolated` | RED | RED |
| federated + injected anchor, `group` | RED | RED |
| federated + author-declared `organization_id` | GREEN | GREEN |
| LOCAL object | GREEN | GREEN |
| dogfood: federated read composes no org predicate | RED | RED |
| dogfood: LOCAL control still walled | GREEN | GREEN |
| dogfood: end-to-end rows survive | RED | RED |

Two further cases went red that the prediction table had not itemized — "federated with no active org" and "an authored policy still reaches the compiler" — both in the same direction, since without the fix Layer 0 composes into those results too. Reported rather than back-filled into the table.

**Ablation B** — drop the provenance half, so every federated object counts as phantom: exactly one pin red, "federated object whose AUTHOR DECLARED a real remote `organization_id`: the wall STAYS". Predicted, and it is what stops the provenance test from being decoration.

## Tests

`packages/plugins/plugin-security/src/federated-tenant-layer0.test.ts` proves the **decision** (dialect-free, hand-built field sets). Because its fixtures build the injected anchor themselves, it cannot also prove the registry still produces one — expectation and reality would share a source. `packages/qa/dogfood/test/federated-rls-injectors.dogfood.test.ts` is the independent witness: a real boot, the real registry, the shipped federated object, plus a LOCAL control so the pin cannot pass by the wall being switched off everywhere. That file went red against the pre-fix build and green after — the before/after quoted above.

## What this PR does not do

- **Ownership scoping (`__readScope` own/unit lowered to an `owner_id` predicate).** Measured and confirmed firing on a federated object without the ADR-0090 grandfather stamp — `{"owner_id":"usr_member_1"}` against a table with no such column. The scope label is stamped in plugin-security, but the predicate is built in `@objectstack/plugin-sharing`, outside this card's declared surface, so it is reported and filed as **#7858** rather than patched here. This is why the first line reads `Part of`.
- **The root cause.** All three fixes so far — the engine's, this one, and the one #7858 proposes — are consumers re-deriving "that column is not really there" about the same producer, `applySystemFields`. Discussed in #7858; a registry-level answer is a larger change than any of these cards scoped.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WbtFaTwsEstJXYTN847jz9)_

### DIFF
--- a/.changeset/federated-tenant-layer0-phantom-anchor.md
+++ b/.changeset/federated-tenant-layer0-phantom-anchor.md
@@ -1,0 +1,53 @@
+---
+"@objectstack/plugin-security": patch
+---
+
+fix(plugin-security): the tenant wall no longer scopes a federated object by a column it does not have (#7835)
+
+A **federated** object (ADR-0015 — `external`, bound to a remote table) is
+registered like any other, which means the ObjectQL registry injects the
+platform's system anchors into it: `organization_id`, `owner_id`,
+`owning_business_unit_id` and the audit `*_by` lookups. But the platform issues
+no DDL for a federated object — `Engine.syncObjectSchema` returns early, because
+the remote schema is owned externally. Those columns therefore exist in the
+registered schema and **in no backing store**.
+
+Layer 0 (the tenant wall, ADR-0095 D1) decides "is this a tenant object?" by
+asking whether the object carries `organization_id`, so it was answered yes about
+a phantom and AND-composed `organization_id = <active org>` onto every federated
+read under a walled (`isolated` / `group`) posture. Measured on the shipped
+showcase: the composed read filter for `showcase_ext_customer` was
+`{ organization_id: 'org_alpha' }`, and `GET /data/showcase_ext_customer`
+answered **HTTP 200 with zero rows**.
+
+The symptom is dialect-dependent and the defect is not. On SQLite an identifier
+that resolves to no column is reinterpreted as a string literal, so the
+comparison is constant-false: no error, no rows, a success status. Postgres and
+MySQL raise `column "organization_id" does not exist` instead. Either way the
+wall isolates nothing while the federated catalog stops answering the moment a
+deployment turns the organization wall on.
+
+Layer 0 now discounts an `organization_id` that is the **platform's injected
+anchor** on a federated object, so it contributes no predicate there. What is
+unchanged:
+
+- **Local objects.** The platform provisions their `organization_id`, so the
+  anchor is real and the wall is untouched.
+- **A federated object that DECLARES a real remote `organization_id`.** The test
+  is provenance — identity against the shipped column definition the registry
+  spreads — not "is this object federated", so an author who exposes a genuine
+  remote tenant column keeps their wall. Any inexact match is read as "not the
+  platform's anchor" and leaves the wall in place: the fail direction is toward
+  isolation.
+- **Layer 1 (business RLS).** App-authored policies still reach the compiler
+  untouched (ADR-0049), including on federated objects.
+
+This is the plugin-security sibling of the engine-layer fix that withheld
+`DriverOptions.tenantId` for the same objects; that one cannot reach here,
+because Layer 0 is a `where` predicate composed into the query AST rather than a
+driver option.
+
+Record-ownership scoping (`__readScope` `own`/`unit` lowered to an `owner_id`
+predicate) reaches federated objects through the same phantom column set and is
+**not** addressed here — it is produced in `@objectstack/plugin-sharing` and is
+tracked separately.
--- a/packages/plugins/plugin-security/package.json
+++ b/packages/plugins/plugin-security/package.json
@@ -2,7 +2,7 @@
   "name": "@objectstack/plugin-security",
   "version": "17.0.0-rc.6",
   "license": "Apache-2.0",
-  "description": "Security Plugin for ObjectStack — RBAC, RLS, and Field-Level Security Runtime",
+  "description": "Security Plugin for ObjectStack \u2014 RBAC, RLS, and Field-Level Security Runtime",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "exports": {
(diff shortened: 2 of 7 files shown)


huangyiirene:
**CI convergence: ESLint patch round dispatched.** `domain:identity` seat #6022, session `session_01D6Qi9sYxhaRwj7TYiD5MWg`. The review verdict is unchanged — this PR stays **ACCEPTed**; only the gate is red.

## The failure, and the trap under it

Two errors, one file, adjacent lines — every other job on `e0d3160` is `completed: success` (TypeScript Type Check, Test Core ×3, Dogfood ×3, Temporal Conformance, Build Core/Docs, Check Changeset, Validate Package Dependencies):

```
packages/qa/dogfood/test/federated-rls-injectors.dogfood.test.ts
  83:5  error  slot-lookup/no-any-assignment
  84:5  error  slot-lookup/no-any-assignment
```

**⚠️ Why this passed locally and failed on the runner** — worth recording, because the obvious next move is the wrong one. The code carries

```ts
// eslint-disable-next-line @typescript-eslint/no-explicit-any
const kernel = stack.kernel as any;
```

but CI runs **`eslint . --no-inline-config`**, so **inline disable comments are inert on the runner**. Re-adding or adjusting a disable comment would fail again in exactly the same way. The patch instruction says so explicitly and points at the rule's own prescribed fix instead: pass each slot's real contract type rather than erasing the lookup.

Also carried into the instruction: the rule's message states that *every* `any` it has caught so far was hiding a real gap — so if honest typing makes the compiler object, **that is a finding to report, not something to cast away.** Making ESLint quiet is not the deliverable; keeping the test meaningful is.

## ⚠️ Dispatch-mode amendment

This card was claimed `mode:cloud`. The `Claude_Code_Remote` MCP server has been unreachable since ~21:30Z, so the original dev's cloud session cannot be poked — a bound poke trigger was created at 21:34Z and **never fired**. Per the standing fallback (cloud prescribed but unavailable ⇒ `mode:subagent`, amendment recorded on the claim), the patch is running as a **subagent** on this same branch under the interrupted-dev handoff protocol: existing commits read first, existing branch reused, ⛔ no new PR, **claim and assignee on #7835 untouched**.

Recorded here rather than only in chat so the next reader sees why the executing session changed mid-card.

Next: when the patch lands green, this seat flips #7859 ready and queues it. ⛔ The dev does not poll CI and does not arm the queue.

---

Issue #7894, opened 2026-08-12
meta-plural-url-bypass: `PUT /meta/fields/<name>` walks around the whole two-tier registry gate — 4 registry types have no entry in `PLURAL_TO_SINGULAR`


huangyiirene:
## Summary

`canonicalMetaType` — the ONE canonical spelling of a metadata type at the `/meta` read/write/delete boundary (#4432) — folds plural→singular through `PLURAL_TO_SINGULAR`. That map is a **manifest collection map** (`objects: [...]`, `apps: [...]`), and it is missing entries for four registry types. For those, the plural URL spelling is not folded: it is treated as an **unregistered plugin type**, which every authorization gate treats as permissive by construction.

Result: the plural URL is a door around the singular URL's lock.

## Measured, live (booted showcase, admin bearer, on top of #7743's fix)

```
PUT /api/v1/meta/field/showcase_task.title   → 403 NOT_OVERRIDABLE     ← #7743's fix, correct
PUT /api/v1/meta/fields/showcase_task.title  → 200 {"success":true,"state":"active",
                                                    "message":"Saved fields 'showcase_task.title' …"}
```

Note the receipt: **`Saved fields`**. The row is persisted with `type='fields'` — a second namespace for the same item, which is precisely the defect class #4432 was filed about ("one plural-spelled read minted a plural registry entry … one overlay row shadowed an entire code-authored listing, and survived the DELETE that was supposed to lift it").

## Why the gates let it through

For an unmapped spelling `'fields'`:

- `isOverlayAllowed('fields')` → `false`
- `isRuntimeCreateAllowed('fields')` → `true` — no static registry entry, so `getMetaTypes()` synthesises `allowRuntimeCreate: true`
- the #5086 code-only refusal needs BOTH false → does not fire
- `orgScopedWriteRefusal` → returns `null`, explicitly: *"Statically-declared types only… a type with no entry in `DEFAULT_METADATA_TYPE_REGISTRY` is plugin-registered at runtime"*
- `SysMetadataRepository.assertAllowed('fields', 'runtime-only')` → `!STATIC_REGISTRY_TYPES.has('fields')` → **returns, allowed**

Every one of those is correct behaviour *for a genuinely plugin-registered type*. The bug is that `'fields'` is not one — it is a registry type wearing a spelling the boundary map does not know.

## Blast radius (measured against `DEFAULT_METADATA_TYPE_REGISTRY` at `bcea363`)

[...]

Issue #7893, opened 2026-08-12
meta-field-write-inert: an accepted `field` PUT never reaches the object — a runtime-created field is stored `valid=true` and is absent from `fields` forever


huangyiirene:
## Summary

Split out of #7743, which recorded **two** entangled defects. #7743's PR closes the first (an artifact-backed `field` override is now refused 403 `NOT_OVERRIDABLE`). This is the second, and it is **independent**: a `field` write that is entirely *legitimate* under the registry declaration is accepted, persisted, reported valid — and read by nothing.

The registry declares `field` with `allowRuntimeCreate: true`, so authoring a brand-new field through `PUT /api/v1/meta/field/<object>.<name>` is a **supported** capability. It just does not work. `declared ≠ enforced`, Prime Directive #10.

## Why this is its own card, not part of #7743

#7743's dispatch predicted that refusing the override would destroy the evidence for the inertness. **Measured: it does not.** The inertness has two halves and only one of them is behind the new refusal:

- *override* half (`showcase_task.title`) — now 403, no longer observable. ✔ expected
- *create* half (`showcase_task.zz_new_probe`) — **still 200, still inert, still reproducible on `main` after #7743's PR lands.**

So this card has a live repro, not archived evidence.

## Measured (booted showcase, admin bearer)

Before #7743's fix — both halves:

```
PUT /api/v1/meta/field/showcase_task.title  {name:'title', label:'Tampered', type:'text'}
  → 200 {"success":true,"state":"active","message":"Saved field 'showcase_task.title' (env-wide…)"}
GET /api/v1/meta/field/showcase_task.title
  → {"item":{"name":"title","label":"Tampered","type":"text","_diagnostics":{"valid":true}}}
GET /api/v1/meta/object/showcase_task
  → title.label === 'Title'      ← unchanged. 24 fields.
```

After #7743's fix — the create half, unchanged:

```
PUT /api/v1/meta/field/showcase_task.zz_new_probe  {name:'zz_new_probe', label:'Probe', type:'text'}
  → 200 {"success":true,"state":"active","message":"Saved field 'showcase_task.zz_new_probe' (env-wide…)"}
GET /api/v1/meta/field/showcase_task.zz_new_probe
  → {"item":{…,"_diagnostics":{"valid":true}}}
GET /api/v1/meta/object/showcase_task
  → fields count === 24, 'zz_new_probe' in fields === false     ← the field does not exist
```

[...]

Issue #7743, opened 2026-08-11
org-override-registry-gate: the `field` overlay lock is not enforced — an artifact-backed field PUT is accepted 200 (and is inert)


huangyiirene:
## Symptom

The registry declares `field` with `allowOrgOverride=false`, and the field is artifact-backed — yet the live route accepts the overriding write.

- `PUT /api/v1/meta/field/showcase_task.title {name:'title', label:'Tampered', type:'text'}` with an admin bearer → **200** `state:'active'`. Reproduced on `showcase_task.status`.
- The row **is** persisted and reads back with `_diagnostics.valid=true`.
- **And the accepted write is inert**: `GET /meta/object/showcase_task` still reads `title.label = 'Title'`. A brand-new field written the same way never appears in the object's `fields` either.

So the door answers success twice over — accepted, and stored valid — for a write the registry forbids and the runtime ignores. The `object`, `view`, `dashboard` and `job` variants all behave correctly in the same run, and locked-override vs open-create were proven independent there.

## Root cause (suspected — located, not proven by a fix)

The gate keys on `isArtifactBacked(type, name)`, which resolves through `registry.getArtifactItem(singular, name, currentPackageId) ?? registry.getArtifactItem(type, name, currentPackageId)` (`packages/metadata-protocol/src/protocol.ts:8678-8679`). Fields are **not registered as standalone `field` artifact items** — they live inside the object — so `getArtifactItem('field', 'showcase_task.title')` misses. With the artifact lookup empty, the write is classified as a runtime-only **create**, and the registry entry for `field` carries `allowRuntimeCreate: true`:

```
{ type: 'field', label: 'Field', …, allowOrgOverride: false, allowRuntimeCreate: true, … }
```
— `packages/spec/src/kernel/metadata-plugin.zod.ts:629` on `origin/main`.

`allowOrgOverride: false` is never consulted because nothing on this path believes an artifact is being overridden.

**Why the pin stayed green:** `overlay-precedence.test.ts` (27 passing) pins the **protocol-level** denial. The live `field` route is not in its coverage, so the gate is proven where it is exercised and absent where it is used.

**Stale-premise check:** re-verified on `objectstack` `origin/main` (`00e9196`) — the `isArtifactBacked` resolution and the `field` registry row are both unchanged.

## Reproduction

1. Boot the showcase with writable runtime packages; obtain an admin bearer.
[...]

Issue #4432, opened 2026-08-01
meta overlays: unnormalized type segment creates phantom rows that shadow the code-authored listing and cannot be deleted


baozhoutao:
Found while verifying the 17.0.0-rc.1 checklist on #3909 (F1). Verified on `main` @ `1ee48bc60`, showcase under `os serve --dev`, admin session.

#3985 made the `/meta` **per-type gates** accept both spellings of the type segment — but the **customization-overlay write path does not normalize it**, so the two spellings address two different overlay namespaces, and the phantom rows this creates corrupt every listing surface that should have been artifact ∪ overlay.

## Repro & observed (one fresh boot, in order)

1. `PUT /api/v1/meta/actions/rc1_probe` (plural) → `200 … Saved customization overlay … type=actions, name=rc1_probe [seq=1]`
2. `PUT /api/v1/meta/action/rc1_probe` (singular) → `200 … type=action, name=rc1_probe [seq=1]` — **a second, distinct row** (both seq=1).
3. `GET /api/v1/meta/actions` (plural — the spelling the showcase's 11 code-authored actions listed under before step 1) now returns **only `rc1_probe`** — one overlay row shadows the entire code-authored action list on that spelling. `GET /api/v1/meta/action` (singular) returns the merged 12.
4. `DELETE /api/v1/meta/action/rc1_probe` → `200 … reset to artifact default [seq=2]`. `DELETE /api/v1/meta/actions/rc1_probe` → `200 … No customization overlay found for actions/rc1_probe — already at artifact default.` — **the plural row it created in step 1 is unreachable by either spelling**.
5. After both deletes: `GET /meta/actions` still returns the phantom `rc1_probe` (and still hides the 11 real actions); `GET /meta/action` still lists `rc1_probe` among 12 items, while `POST /api/v1/actions/showcase_task/rc1_probe` correctly 404s. Listing and dispatch now disagree about a deleted item.
6. Throughout all of this, `sys_metadata` in the bound SQLite datasource holds **0 rows** (and `.objectstack/metadata` holds no matching file) — where these "env-wide, state=active" overlays actually persist is not observable; on this boot they appear to be process-memory only.

Also observed: after step 2 the overlay was **listed but not dispatchable for a short window** (`POST` answered the ADR-0110 "has no declaration" 404, then succeeded on a later call) — some cache between `saveMeta` and `resolveRouteActionDeclaration` lags the write.

## Expected

[...]

PR #7897, opened 2026-08-12
fix(metadata-protocol): enforce the `field` overlay lock at the live route (#7743)

huangyiirene:
Part of #7743

Ships the **refusal**. The card recorded two entangled defects; the second (the accepted write is inert) is measured, proven independent, and filed as #7893 rather than folded in. A third defect found while writing the pin — the plural URL spelling walks around this same lock — is filed as #7894 and is why this is `Part of` rather than `Fixes`.

## The defect

The registry declares `field` with `allowOrgOverride: false`, and a field a code package ships is an artifact. Yet, with an admin bearer on a booted showcase:

```
PUT /api/v1/meta/field/showcase_task.title  {name:'title', label:'Tampered', type:'text'}
  → 200 {"success":true,"state":"active","message":"Saved field 'showcase_task.title' (env-wide…)"}
GET /api/v1/meta/field/showcase_task.title
  → {"item":{…,"_diagnostics":{"valid":true}}}
```

Reproduced on `showcase_task.status`. The door answered success twice over — accepted, and stored valid — for a write the registry forbids.

## Premise: verified at the branch point, and refined

The card located the root cause as **suspected**. Verified at `bcea363`, and it holds — with one refinement that matters for where the fix goes.

`field` is the ONE type in `DEFAULT_METADATA_TYPE_REGISTRY` whose artifacts are not standalone registry items. Its `filePatterns` (`**/*.field.ts`) match nothing in any app, because fields are authored *inside* the object (`ObjectSchema.fields`, a `z.record(name, FieldSchema)`). So the object's loader registers one `object` item and no `field` items at all, and `getArtifactItem('field', 'showcase_task.title')` misses on a field the package unambiguously ships. With the lookup empty the write is classified as a runtime-only **create**, and `field` carries `allowRuntimeCreate: true` — so `allowOrgOverride: false` is never consulted.

**The refinement:** the card pointed at `saveMetaItem`'s own `NOT_OVERRIDABLE` block, but that block is behind `environmentId !== undefined` and the flagship showcase boots with **no** `environmentId` (a host config with instantiated plugins). On that topology the block is skipped entirely, and the site that actually refused `object` in the same run is `SysMetadataRepository.assertAllowed`, whose `intent` (`override-artifact` vs `runtime-only`) the protocol picks from the very same predicate. Measured: `object` was refused with the repository's sentence (`'object' is not allowOrgOverride in the registry. Overlay-allowed: view, dashboard, report, translation, email_template.`), not the protocol's.

Both doors read `isArtifactBacked`, so making that predicate truthful closes both. That is the fix.

## The change

`isArtifactBacked` now falls back to `isNestedArtifactField` for `field`: split `<object>.<field>` at the first separator, resolve the parent through the **artifact-only** lookup (so a plain-key overlay row can neither manufacture nor mask an artifact — ADR-0010 §3.3), and answer whether the shipped object carries that field.

It returns a boolean rather than routing through `lookupArtifactItem`, deliberately: a field sub-document is a bare `{name, label, type}` with no `_packageId`/`_lock` envelope, and `lookupArtifactItem`'s other callers (lock resolution, `mergeArtifactProtection`, the layered read) consume that envelope. Handing them a field body would make them assert provenance nobody stamped. The authorization question is answerable without one.

`fields` is read in its one canonical form (a record keyed by name); no array fallback — Prime Directive #12.

## Scope: `field`-specific, and that was measured, not assumed

The dispatch asked whether this is one type or a class. Measured across the whole registry on a booted showcase — every other declared type either registers its artifacts standalone with a `_packageId`, or genuinely ships none:

| standalone artifacts (package-stamped) | no artifacts at all |
|:--|:--|
| `action` 70, `page` 33, `permission` 16, `dataset` 9, `doc` 9, `hook` 4, `report` 4, `mapping`/`book`/`email_template` 1 | `position`, `tool`, `skill`, `seed`, `translation`, `external_catalog` |

`field` is the only name where the registry's answer and the shipped artifact disagree: **2** items listed under `/meta/field`, both written by this session's probes, **0** package-stamped — while `showcase_task` alone ships 24 fields.

`action` is the instructive one: it is *also* nested inside the object document, yet it IS registered standalone, so it was already refused correctly (403) in the same run. The class has one member, so the diff has one member.

## Verification

Reproduced live, fixed, re-measured live. Same seven contrasts, before and after:

| write | before | after |
|:--|:--|:--|
| `field/showcase_task.title` (artifact override) | **200**, row persisted, `valid=true` | **403 `NOT_OVERRIDABLE`** |
| `field/showcase_task.status` | **200** | **403 `NOT_OVERRIDABLE`** |
| `field/showcase_task.zz_new_probe` (brand-new) | 200 | 200 — unchanged |
| `object/showcase_task` (full body) | 403 `NOT_OVERRIDABLE` | 403 — unchanged |
| `view/showcase_task.in_progress` | 200, overlay effective | 200 — unchanged |
| `dashboard/system_overview` | 200 | 200 — unchanged |
| `job/showcase_health_sweep` | 403 code-only | 403 — unchanged |
| `action/showcase_mark_done`, `page/sys_user_detail` | 403 | 403 — unchanged |
| `position/contributor` | 200 | 200 — unchanged |

The `object` contrast used the **full** body — a trimmed one hits `DESTRUCTIVE_CHANGE` first and never reaches the registry gate.

### The pin is at the LIVE ROUTE, which is the whole lesson of this card

`packages/objectql/src/overlay-precedence.test.ts` already pins this denial with **27 passing cases** — at the protocol level, calling `saveMetaItem` directly. The reported symptom sailed straight past all 27 because the live `field` route is not in their coverage. A 28th assertion at that layer would have been green and blind in exactly the same way.

So `packages/runtime/src/meta-field-overlay-lock.test.ts` drives the real `HttpDispatcher.handleMetadata`, the real `ObjectStackProtocolImplementation` and the real `SysMetadataRepository`, then reads the stored **row**. Nothing stubs `saveMetaItem` — a double cannot see this defect, because the defect is which INTENT the protocol derives before the repository is called. The registry double reproduces the real miss faithfully: it serves the `object` artifact and has **no `field` collection at all**.

Both topologies are driven, because they refuse at different sites: `environmentId: undefined` (the showcase, where the repository gate is the enforcement point) and `environmentId: 'env_1'` (where `saveMetaItem`'s own gate fires first). Both assert `403` **and** `NOT_OVERRIDABLE` — ADR-0112: a refusal is pinned by code and status, never by "it threw".

### Reverse verification — with the rebuild

Direction predicted in the file header **before** running. `packages/runtime` resolves `@objectstack/metadata-protocol` through its **`dist`**, and stack traces are source-mapped back to `src`, so a source-only revert measures nothing while looking like it measured something. The fix was committed first, then reverted with `git checkout origin/main -- packages/metadata-protocol/src/protocol.ts` **and rebuilt** (`grep -c isNestedArtifactField dist/index.js` → `0`), measured, then restored and rebuilt (→ `3`).

**Predicted 3 red / 7 green; measured 3 red / 7 green.**

The seven greens are not slack. Four are the negative direction the card demanded — `object`/`view`/`dashboard`/`job` were measured as already correct, so a fix that tightened any of them is over-reach and must fail here. Two are the legitimate `field` write: `allowRuntimeCreate: true` is real, and a fix that refused every `field` PUT would pass a one-directional test while breaking the feature. The seventh is the known gap below.

## Two defects reported rather than folded in

**#7893 — the accepted write is inert.** The card predicted that refusing the override would destroy this evidence. **Measured: it does not.** The inertness has two halves and only one is behind the new refusal. The *override* half is now unreachable, as expected — but the *create* half is a write that stays entirely legal under `allowRuntimeCreate: true`:

```
PUT /api/v1/meta/field/showcase_task.zz_new_probe → 200, valid=true      (after the fix)
GET /api/v1/meta/object/showcase_task → 24 fields, 'zz_new_probe' absent
```

So #7893 ships with a **live repro on `main`**, not archived evidence. It needs a maintainer decision (implement the composition / retire the write channel under ADR-0049 / document it), which is why it is a card and not a rider here.

**#7894 — the plural URL spelling bypasses this lock.** Found by writing the plural case expecting 403 and measuring 200, then confirming live:

```
PUT /api/v1/meta/field/showcase_task.title   → 403 NOT_OVERRIDABLE     ← this PR
PUT /api/v1/meta/fields/showcase_task.title  → 200 "Saved fields '…'"  ← still open
```

`canonicalMetaType` folds plural→singular through `PLURAL_TO_SINGULAR`, which is the **manifest collection** map and has no `fields` key (nor `seeds`, `external_catalogs`, `translations`). An unmapped spelling is read as an unregistered *plugin* type, which every gate treats as permissive by construction — correct for a type that really is plugin-registered; `'fields'` just is not one. The row lands under `type='fields'`, a second namespace, which is the #4432 defect class.

⛔ The one-word patch — teaching `isNestedArtifactField` to accept `'fields'` — was **rejected**: it is a spelling-tolerant lookup below the boundary, the exact pattern `canonicalMetaType`'s own doc comment rejects, and it would still mint the second namespace. The remedy belongs at the boundary map and spans four types.

It is pinned as an explicitly-labelled **KNOWN GAP** test that asserts today's behaviour, names #7894, and states that it goes red when #7894 lands so whoever fixes it flips it. Hiding it would have been the alternative.

## Gates

| gate | result |
|:--|:--|
| `@objectstack/metadata-protocol` | 72 files / **1066 passed** |
| `@objectstack/objectql` (incl. the 27 protocol-level `overlay-precedence` cases and the protocol doubles path-scoped derivation misses) | 187 files / **3312 passed** |
| `@objectstack/client` | 21 files / **282 passed** |
| `@objectstack/runtime` | 137 files / **2093 passed** |
| `pnpm check:durability-log-level` | ✓ |
| `pnpm check:nul-bytes` | ✓ (7216 files) |
| eslint on changed files | clean |

`check:error-code-casing` is not applicable — no new error code; the refusal reuses `NOT_OVERRIDABLE`, which is the sentence this condition already carries.

Changeset: `.changeset/field-overlay-lock-enforced.md`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CZNzPZmS8EHXFfqbmnA6Jh)_

### DIFF
--- a/.changeset/field-overlay-lock-enforced.md
+++ b/.changeset/field-overlay-lock-enforced.md
@@ -1,0 +1,58 @@
+---
+"@objectstack/metadata-protocol": patch
+---
+
+fix(metadata-protocol): enforce the `field` overlay lock — an artifact-backed field PUT is refused instead of accepted 200 (#7743)
+
+The metadata-type registry declares `field` with `allowOrgOverride: false`, and a
+field a code package ships is an artifact. Yet
+`PUT /api/v1/meta/field/showcase_task.title` answered **200** `state:'active'`
+with an admin bearer, the row persisted, and it read back with
+`_diagnostics.valid=true`. Reproduced on `showcase_task.status`. The door
+answered success twice over for a write the registry forbids.
+
+**Why the declaration was never consulted.** `field` is the ONE type in
+`DEFAULT_METADATA_TYPE_REGISTRY` whose artifacts are not standalone registry
+items: its `filePatterns` (`**/*.field.ts`) match nothing in any app, because
+fields are authored *inside* the object (`ObjectSchema.fields`). So the object's
+loader registers one `object` item and no `field` items at all, and
+`getArtifactItem('field', 'showcase_task.title')` missed on a field the package
+unambiguously ships.
+
+That miss is a load-bearing authorization input, not a cosmetic one.
+`isArtifactBacked` is what picks the write INTENT for
+`SysMetadataRepository.assertAllowed` (`override-artifact` vs `runtime-only`) and
+what arms `saveMetaItem`'s own `NOT_OVERRIDABLE` gate. With the lookup empty, an
+override of a packaged field was classified as a runtime-only **create** — and
+`field` carries `allowRuntimeCreate: true`, so `allowOrgOverride: false` was
+never reached. Both doors read the same predicate, so both are closed by making
+it truthful: `isArtifactBacked` now resolves a `<object>.<field>` name through
+the object's artifact and answers about the field the package actually ships.
+
+**The other tier is untouched, deliberately.** `allowRuntimeCreate: true` is
+real: a genuinely new field the object's artifact does not carry is still
+accepted, and so is a field of a runtime-created (non-packaged) object. This
+closes the overlay tier only.
+
+**Scoped to `field`, measured rather than assumed.** Every other declared type
+either registers its artifacts standalone with a `_packageId` — `action` (70 on
+the showcase), `page` (33), `permission` (16), `dataset` (9), `doc` (9), `hook`
+(4), `report` (4) — or genuinely ships no artifacts, where "not artifact-backed"
+is the true answer. `action` is the instructive one: also nested inside the
+object document, yet registered standalone, and already refused correctly.
+`object` / `view` / `dashboard` / `job` were measured as already correct in the
+same run and behave identically after this change.
+
+The refusal is pinned at the **live route** (`packages/runtime`), driving the
+real dispatcher, protocol and repository on both topologies — the environment
+kernel and the no-`environmentId` showcase, which refuse at different sites. The
+27 existing protocol-level cases in `overlay-precedence.test.ts` were green
+throughout the defect's life precisely because the route was not in their
+coverage.
+
+Two defects found alongside this one are filed rather than folded in: the
+accepted write is also **inert** (#7893 — a legitimately created field never
+appears in the object's `fields`), and the **plural** URL spelling
+`/meta/fields/<name>` still walks around this lock because `PLURAL_TO_SINGULAR`
+has no `fields` key (#7894, which spans four types). The plural gap is
+characterized by a test that names it and goes red when it is closed.
--- a/packages/metadata-protocol/src/protocol.ts
+++ b/packages/metadata-protocol/src/protocol.ts
@@ -8709,13 +8709,95 @@ export class ObjectStackProtocolImplementation implements
      * Used by the two-tier authorization model to distinguish
      * "overlaying a packaged item" (requires `allowOrgOverride`) from
      * "authoring a DB-only item" (requires only `allowRuntimeCreate`).
+     *
+     * [#7743] …and it must answer that question about the artifact as SHIPPED,
+     * not about how the registry happens to key it. See
+     * {@link isNestedArtifactField} for the one declared type whose artifacts
+     * are not standalone registry items at all.
      */
     private isArtifactBacked(type: string, name: string): boolean {
         // `lookupArtifactItem` only returns items whose `_packageId` marks a
         // genuine code package (the `'sys_metadata'` rehydration sentinel is
         // excluded), and — via `SchemaRegistry.getArtifactItem` — is immune
         // to plain-key shadows hydrated from overlay rows.
-        return this.lookupArtifactItem(type, name) !== undefined;
+        if (this.lookupArtifactItem(type, name) !== undefined) return true;
+        return this.isNestedArtifactField(type, name);
+    }
+
+    /**
+     * [#7743] Is `(field, '<object>.<field>')` a field a code package ships?
+     *
+     * ## Why this predicate needs a second resolver at all
+     *
+     * `field` is the ONE type in `DEFAULT_METADATA_TYPE_REGISTRY` whose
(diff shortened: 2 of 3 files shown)


huangyiirene:
### Gate table — final, all green

Two suites first reported file-level failures with **zero** test failures (`metadata-protocol` 7 files / 0 tests; `dogfood` 27 files / 0 tests). Both were measured while a `build` was rewriting `dist/` underneath the running suite. Re-run clean, alone, both are fully green — recorded here rather than quietly dropped.

| gate | result |
|:--|:--|
| `@objectstack/metadata-protocol` | 72 files / **1066 passed** |
| `@objectstack/objectql` | 187 files / **3312 passed** |
| `@objectstack/client` | 21 files / **282 passed** |
| `@objectstack/runtime` | 137 files / **2093 passed** |
| `@objectstack/dogfood` | 92 passed + 1 skipped / **588 passed**, 3 skipped |
| build closure (`--filter @objectstack/runtime...`) | ✓ exit 0 |
| `pnpm check:durability-log-level` | ✓ |
| `pnpm check:nul-bytes` | ✓ (7216 files scanned) |
| eslint on changed files | clean |

`check:error-code-casing` is not applicable — no new error code. `pnpm typecheck` on `@objectstack/metadata-protocol` reports no `typecheck` script (its ledger entry), so it is not a gate this diff can run.

---

Issue #3682, opened 2026-07-27
实现 ADR-0106:元数据面 FLS——object schema 按调用者掩码(#3661 ③ 落地)


os-zhuang:
跟踪 [ADR-0106](https://github.com/objectstack-ai/objectstack/blob/main/docs/adr/0106-metadata-plane-fls-object-schema-masking.md)(#3678 已合并)的实现。决策背景与完整论证见 ADR 正文及 #3661;客户端侧 ①② 已由 objectstack-ai/objectui#2866 承担。

## 实现清单(按 ADR 决策点)

- [ ] **D1/D2 — 分发层投影**:object schema 读按 `security.getReadableFields(object, ctx)` 投影 `fields`,掩码字段整体删除。协议层 `getMetaItem`/`getMetaItems` 签名不动。落点:
  - `packages/rest/src/rest-server.ts` — `GET /meta/object/:name` 的 **cached 与非 cached 两条路径**
  - `packages/rest/src/rest-server.ts` — `GET /meta/object` 列表读(逐 item 投影)
  - `packages/runtime/src/domains/meta.ts` — `/metadata` catch-all 的 object 分支(protocol 与 registry 两条查找路径)
- [ ] **D3 — mask-after-cache + 指纹 ETag**:共享缓存继续每对象存一份全量;取出后掩码;调用者对该对象的 denied-set 稳定哈希(无限制调用者为空)折入 ETag。顺序约束:fetch → mask(或 error)→ send,缓存全量体不得有任何绕过掩码上线的路径。实现受阻时可先落 doc/book 式 bypass(记录为过渡,非终态)。
- [ ] **D4 — 豁免**:`isSystem`(getReadableFields 已有)+ 平台 admin(复用 app 过滤的 `systemPermissions` 判定)。豁免是 caller 属性,不是路由属性。
- [ ] **D5(4) — 出口审计**:排查其余 schema-bearing 端点(OData `$metadata`/describe 类、client SDK describe);未覆盖的按 Prime Directive #10 开 issue。
- [ ] **D6 — 三档失败姿态**:无 security 服务 → 放行;`undefined` → 放行 + 结构化 warn + 指标 + `Cache-Control: private, no-store`(不发共享 ETag);评估抛异常 → **5xx**(绝不放出未掩码体,绝不返回空 fields 的 200)。
[...]

PR #7907, opened 2026-08-12
fix(meta): gate `_drafts` outlets as authoring surfaces (ADR-0106 D5(4), #6599)

### LINKED ISSUE #6599: finding: `/meta/_drafts` serves DRAFT object schemas unmasked — the one ADR-0106 outlet left uncovered
Filed from the ADR-0106 D5(4) implementation-time outlet audit (#3682). Not fixed there: it is a distinct surface with its own authorization question, and guessing at that question inside a masking PR is the wrong place to answer it.

## What was measured

ADR-0106 (#3682) now projects object schemas onto the caller's readable field set at every outlet it enumerated — the REST single read (cached and uncached), the layered `?layers=true` view, the compound-name read, the list read, and the runtime `/metadata` catch-all's five resolvers. The D5(4) sweep for "any other schema-bearing endpoint" found one that is **not** covered:

- `GET /api/v1/meta/_drafts` — `packages/rest/src/rest-server.ts`, the route registered before `/meta/:type` so `_drafts` is not captured as a type parameter.
- `GET /metadata/_drafts` — `packages/runtime/src/domains/meta.ts`, the same surface on the dispatcher.

Both call `protocol.listDrafts({ packageId, type })` and `res.json(result)` / `deps.success(data)` with no per-caller filtering at all beyond `requireAuth`. A pending **object** draft carries its full `fields` map, so an authenticated caller with no read access to `salary_grade` still learns the field exists, along with its label, type, picklist options, formula and `requiredPermissions` — exactly the disclosure ADR-0106 closes one route over.

## Why it was left out of #3682

Two reasons, and both are questions rather than omissions:

1. **`_drafts` is a mixed-type list.** Unlike `/meta/object`, its items are drafts of *any* metadata type, so applying the projection means inspecting each row's type before deciding. Mechanically easy; the shared masker in `@objectstack/metadata-core` already takes a per-object-name posture. But —
2. **The route may not belong to non-authors at all.** ADR-0106 D4 assumes draft/preview reads are "admin-gated upstream already", and this route reads as an authoring surface (the console's "pending changes" view, draft-aware package contents). If that assumption is true, the correct fix is a capability gate on the route — not a field projection — and adding a projection instead would quietly bless non-author access to the pending-changes view.

Those two answers lead to different code, and the decision is not ADR-0106's to make.

## Suggested disposition

Decide which of these `_drafts` is:

- **(a) An authoring surface.** Gate the route on the same `systemPermissions` judgement ADR-0106 D4 uses for its exemption (`studio.access` / `setup.access`) and 403 everyone else. Cheapest, and matches how the route is actually used.
- **(b) A general read surface.** Route each object-typed draft through `applyObjectSchemaMask` from `@objectstack/metadata-core` (the same function every other exit uses — see `packages/rest/src/meta-object-fls.test.ts` for the case table it is driven by, which a new exit joins by adding one row).

Either way, add the exit to the shared ADR-0106 case table so it cannot regress silently.

## Repro sketch

```
# as a caller whose permission set marks account.salary_grade readable:false,
# with a pending draft of `account`
curl -H "Cookie: <session>" /api/v1/meta/_drafts?type=object | jq '.[].item.fields.salary_grade'
# → the full field definition, where GET /api/v1/meta/object/account correctly omits it
```

Related: #3682 (ADR-0106 implementation), #3661 ③, ADR-0033 (drafts), ADR-0106 D4/D5(4).

hotlong:
Fixes #6599

## What this closes

`GET /api/v1/meta/_drafts` (`packages/rest/src/rest-server.ts`) and
`GET /metadata/_drafts` (`packages/runtime/src/domains/meta.ts`) both called
`protocol.listDrafts()` and returned the result verbatim — the one ADR-0106
schema-serving outlet the #3682 D5(4) sweep left uncovered. A pending **object**
draft carries its full `fields` map, so an authenticated caller with no read
access to a field still learned it existed, along with its label, type,
picklist options, formula and `requiredPermissions` — the exact disclosure
ADR-0106 closes on every other `/meta` exit.

## Route (a), per the #6599 ruling — an authoring gate, not a field projection

Both faces now gate per caller on the **same D4 exemption predicate the mask
uses** — `isObjectSchemaMaskExempt` from `@objectstack/metadata-core`
(`studio.access` / `setup.access` / `manage_metadata`, or `isSystem`) — and
answer **403** to everyone else, rather than masking the draft field-by-field.

- The gate runs **before** the protocol is resolved, so the 501-vs-200 answer
  cannot be used to probe kernel support (the same posture as the neighbouring
  `_migrate-stored` route).
- It is **independent of the D8 per-field-mask escape hatch**: opting out of
  per-field masking is not consent to expose pending drafts to non-authors.
- Refusal envelopes follow each transport's own precedent: REST answers
  `FORBIDDEN`; the runtime dispatcher answers `PERMISSION_DENIED` (derived from
  the 403 status by `standardErrorCodeForHttpStatus`, matching `_migrate-stored`
  next door).

## Premise verification (the deciding check)

Route (a) 403s non-authors, which is correct only if no legitimate non-author
caller reads these routes. Consumer census of `.listDrafts(` / `_drafts` across
`objectstack`, `objectui` and `cloud`:

- **objectui console** — HomePage pending-drafts banner, Studio design surfaces,
  metadata-admin (PackagesPage, inspectors), the preview/publish panels
  (DraftChangesPanel, DraftPreviewBar, draftStatus): every one is a
  builder/admin authoring affordance, and each tolerates a 403 gracefully
  (`.catch(() => [])` / `client.listDrafts?.({})`). The HomePage banner is
  explicitly the "publish your AI-drafted metadata" affordance — for a
  non-author it degrades to "not shown", the correct behaviour.
- **cloud AI-studio** — `readDraftFirst` uses `ctx.protocol.getMetaItem(...)`
  server-side; it is **not** an HTTP `_drafts` consumer.
- **cloud golden-prompts eval** — a dev cookie-jar harness that already checks
  `res.ok`; not a production non-author.
- **CLI** — no consumer.

No legitimate non-author production caller exists; authors' access is unchanged.
Disclosure confirmed still reproducing on origin/main before the fix (both
outlets unmasked, no gate).

## Coverage — both faces, one shared case table

Both outlets are two faces of one contract, so a new describe block in **each**
package's shared ADR-0106 case table drives the SAME `OBJECT_SCHEMA_MASK_CASES`
list through the drafts face:

- `packages/rest/src/meta-object-fls.test.ts`
- `packages/runtime/src/domains/meta-object-fls.test.ts`

The verdict is derived, not hand-tabulated —
`isObjectSchemaMaskExempt(context)` decides allow-vs-refuse for every case — so
the two transports cannot diverge on "who is an author" without a red test, and
a new exemption principal added to the shared table flows into the gate
automatically. Two consequences are pinned: `unrestricted-caller/byte-identical`
(readable everywhere, no authoring capability) is a 403 here — the gate is
stricter than the mask, which is the point of route (a) — and
`masking-disabled/D8` is still a 403. Refusal cases assert the ADR-0112 envelope
**code AND status**, and that the sensitive field never appears in the refusal
body. Three pre-existing `http-dispatcher.test.ts` drafts plumbing cases were
given an author context (their intent is plumbing, not authz) and a gate-first
403 case added beside them.

## Verification

- `pnpm --filter @objectstack/rest test` — 92 files, 1489 passed
- `pnpm --filter @objectstack/runtime test` (FLS + http-dispatcher) — 138 files, 2109 passed
- `pnpm --filter @objectstack/rest --filter @objectstack/runtime typecheck` — clean
- Gates: `check:authz-resolver`, `check:meta-type-normalized`, `check:route-envelope`
  (rest-server stays at the pinned `siblingCode 77` — the new 403 is the conformant
  nested `error: { code, message }` shape, not the `siblingCode` dialect),
  `check:nul-bytes` — all green
- Merged latest `origin/main` (sits on landed #7836; region-disjoint, no conflicts)

---


hotlong:
**CI red diagnosed and being fixed — `domain:cli` seat has this one, no second diagnosis needed.**

`TypeScript Type Check` failed on `35273545`. It is **not** a flake and **not** collateral from a red `main` — the job is `pnpm check:type-check-debt`, and the log names exactly two error lines, both in the two packages this PR adds test code to:

```
• @objectstack/runtime: TEST_DEBT records 227, tsc now reports 231 (+4)
• @objectstack/rest:    TEST_DEBT records 155, tsc now reports 156 (+1)
```

Everything else in that log is an `ℹ` "can be lowered" note about unrelated packages — not this PR's business.

Cause: that gate compiles the `*.test.ts` layer, which the package-level `typecheck` excludes, so the new `[ADR-0106 D5(4)]` test blocks' 5 type errors were invisible to the local run that reported clean. Fix is in flight on the same branch.

⛔ For the record, since the gate's own message offers it: the ledger is **not** being raised. Raising TEST_DEBT to buy a test is what lets the next real regression land silently (issue #6376), and the review that accepted this PR would be worth less than the paperwork. The production diff and its coverage are unchanged and still ACCEPTED — only the test layer's types are being corrected.

---

Issue #7380, opened 2026-08-10
[console][nav] `filterNav` never drops a group DECLARED with `children: []` — Setup renders an inert 「Approvals」 group on a runtime without plugin-approvals


baozhoutao:
Found while browser-verifying the #7213 epic's converged approvals surface (the epic's close-out lists browser verification as its open gap). Not a regression from that epic — the defect predates it — but it is on the same nav face, so recording it here.

## The finding

`filterNav` (`packages/rest/src/rest-server.ts:2724`) is documented to *"Drop empty groups so the sidebar doesn't render a label with nothing under it"*. It does not do that for a group **declared** with `children: []` — only for a group whose children were non-empty and were then filtered away:

```ts
if (Array.isArray(e.children) && e.children.length > 0) {   // ← gate
    const kids = filterNav(e.children);
    if (e.type === 'group' && kids.length === 0) continue;  // ← the drop rule
    out.push({ ...e, children: kids });
} else {
    out.push(e);                                            // ← declared-empty group lands here
}
```

A group that starts out empty never enters the branch that owns the drop rule, so it is pushed through untouched. That is precisely the shape every contribution slot in the setup app uses — `packages/platform-objects/src/apps/setup.app.ts` declares `group_approvals` (and its siblings) with `children: []` and lets plugins fill them via `navigationContributions`. When the contributing plugin is absent, the slot survives filtering as an empty group.

## Reproduction (measured)

Two runtimes, same build (objectstack `main` `88154bee1`, vendored console rebuilt at `.objectui-sha` = `8aad9fd50b16`), both `objectstack dev --ui --seed-admin` on a fresh DB, signed in as the dev admin:

- `examples/app-showcase` — `requires: [… 'approvals' …]`, so `ApprovalsServicePlugin` loads and contributes four items. `GET /api/v1/meta/app` → `group_approvals` has its four children; the sidebar renders 审批 with 审批中心 / 审批申请 / 审批历史 / 审批委派（外出）. Correct.
- `examples/app-crm` — `requires: ['ui', 'automation']`, no approvals plugin. `GET /api/v1/meta/app` returns:

```
  - group_approvals [group] "Approvals"      ← no children key content, nothing under it
```

and the Setup sidebar renders an 审批 group with a chevron that **does not expand and reveals nothing on click** — an inert label.

[...]

Issue #7309, opened 2026-08-10
platform-objects: 16 more actions declare both `confirmText` and `params`, so one click opens two sequential dialogs (same shape as #7278)


os-help:
Out-of-scope finding from #7278 (which is scoped to the two `plugin-approvals` actions). Filed per PD #10 — **unassigned, ungraded**.

## Fact

The shared console action runner chains confirmation **then** param collection, both awaited (objectui `packages/core/src/actions/ActionRunner.ts`, the `confirmText` block immediately followed by the `params` block). Any action declaring both keys therefore opens two dialogs for one click, and nothing is sent until the second.

#7278 names two such actions in `plugin-approvals`. A precise scan of `origin/main` @ `b51a09046` (brace-depth block parse over every non-test `*.ts`, requiring `confirmText` and `params: [` at the same nesting level inside one action object) finds **20 actions in total** — the two from #7278, plus **16 in `packages/platform-objects`** and **4 in the showcase example**:

| file | actions |
|---|---|
| `packages/platform-objects/src/identity/sys-user.object.ts` | `ban_user`, `delete_my_account`, `disable_two_factor`, `generate_backup_codes` |
| `packages/platform-objects/src/identity/sys-oauth-application.object.ts` | `enable_oauth_application`, `disable_oauth_application`, `rotate_client_secret`, `delete_oauth_application` |
| `packages/platform-objects/src/identity/sys-two-factor.object.ts` | `disable_two_factor`, `regenerate_backup_codes` |
| `packages/platform-objects/src/identity/sys-account.object.ts` | `unlink_account` |
| `packages/platform-objects/src/identity/sys-organization.object.ts` | `change_slug` |
| `packages/platform-objects/src/identity/sys-sso-provider.object.ts` | `delete_sso_provider` |
| `packages/platform-objects/src/identity/sys-team-member.object.ts` | `remove_team_member` |
| `examples/app-showcase/src/ui/views/project.view.ts` | `set_labels`, `assign_team`, `reassign_account`, `reschedule` |
| `packages/plugins/plugin-approvals/src/sys-approval-request.object.ts` | `approval_reject`, `approval_recall` (already #7278) |

These are live console surfaces, not dormant metadata: e.g. `sys_user.ban_user` is `locations: ['list_item', 'record_header']` with a `banReason` param, and `sys_oauth_application.rotate_client_secret` pairs a long finality warning with a `client_id` param **and** a `resultDialog`, i.e. three dialogs in a row for one click.

[...]

Issue #7278, opened 2026-08-10
approvals: `sys_approval_request.approval_reject` / `approval_recall` declare both `confirmText` and `params`, so one decision opens two sequential dialogs


yinlianghui:
Out-of-scope finding from objectui#3055 (record page now runs these declared actions). Filed per PD #10 — **unassigned, ungraded**.

## Fact

The shared console action runner chains confirmation **then** param collection, both awaited, in `packages/core/src/actions/ActionRunner.ts` of the objectui repo:

- `:794-806` — `confirmText` → `confirmHandler(...)`, cancel aborts;
- `:809-832` — immediately after, `actionParams` / a `params` array → `paramCollectionHandler(...)`, a second dialog.

`packages/plugins/plugin-approvals/src/sys-approval-request.object.ts` declares both keys on two actions:

| action | line | `confirmText` | `params` |
|---|---|---|---|
| `approval_reject` | `:277-296` | "Reject this request? A rejection is final for every approver." | `comment`, `attachments` |
| `approval_recall` | `:370-384` | "Recall this request? Approvers can no longer act on it…" | `comment` |

So rejecting is: confirm prompt → Continue → *another* dialog the approver did not ask for → Confirm. Nothing is sent until the second one.

## Why this is the producer's to decide, not the consumer's

objectui#3126 hit exactly this shape on the record page's **hand-written** reject button and removed `confirmText` there, keeping the question as the param dialog's `description` — "the param dialog IS the confirmation: it is titled by this label, carries the confirm question as its description, and nothing is sent until its own Confirm." That reasoning was applied to console code because the buttons were console code.

objectui#3055 retires those hand-written buttons in favour of these declared actions, so the record page now inherits the declared shape — and the approvals inbox has had it all along. The fix belongs in the declaration, not in a consumer special-case: a console that stripped `confirmText` when `params` are present would be deciding a metadata author's intent for them, and would do it for every object, not just this one.

Worth noting the two dialogs are **not** cosmetic on a decision surface: the first one already reads as "the action is running", which is what made objectui#3126's version read as a silent no-op.

## Suggested fix (left to triage, not self-selected)

1. Drop `confirmText` from both actions and carry the question as the param dialog's description (mirrors objectui#3126's ruling for the same wording on the same decision);
2. keep `confirmText` and drop the params from these two (loses the reject comment + attachments — almost certainly wrong);
3. rule that confirm-then-collect is intended for irreversible decisions and leave both.

[...]

PR #7909, opened 2026-08-12
fix(platform-objects): drop the dead Setup › Advanced › Signing Keys (JWKS) nav entry (#7544)

### LINKED ISSUE #7544: Setup › Advanced › Signing Keys (JWKS) is a dead nav entry — `sys_jwks` is API-disabled and a `requiredPermissions` gate cannot prune it
## Symptom

Setup › Advanced › **Signing Keys (JWKS)** can never load, for **any** persona.

`sys_jwks` declares `enable.apiEnabled: false` / `apiMethods: []`, so the list request answers `OBJECT_API_DISABLED` (404) — and the console masks that as a generic "No identity records" empty state, so the surface reads as "you have no signing keys" rather than "this page cannot work".

Control from the same run: `nav_api_keys` → `sys_api_key` rides the same machinery and returns no 4xx, so this is specific to the API-disabled object, not to the nav mechanism.

## Root cause

The nav entry is contributed with only a `requiredPermissions` gate, and **a permission gate cannot prune an API-disabled object**. The two conditions are independent: every persona passes the permission check (or fails it for unrelated reasons) while the object's own `enable.apiEnabled: false` makes the destination unreachable regardless. There is no signal on the entry that the server could use to drop it.

Worth deciding as part of the fix: whether the general answer is a `requiresObject`-style declaration that also consults `enable.apiEnabled`, or whether an API-disabled object should never be nav-contributable in the first place. Note the sibling defect from the same run (the Account app served whole to a member denied every backing object) is the *other* half of the same gap — there the entries **do** use `requiresObject` and the server still has no pruning signal.

## Reproduction

1. Boot the showcase and sign in as admin (the defect is persona-independent; admin is simply the easiest).
2. Navigate to Setup › Advanced › **Signing Keys (JWKS)**.
3. Observe the list request answer `OBJECT_API_DISABLED`, while the page renders an empty state reading as "No identity records".
4. Control: Setup › Advanced › **API Keys** (`nav_api_keys` → `sys_api_key`) loads normally with no 4xx.

## Source

Extracted from the QA run objectstack-ai/objectstack#7514 (`builtin-apps-nav-render`, framework `a86db175`, vendored console `09987b68`).

_Body restored 2026-08-11 after the filing agent overwrote it with a probe string and was interrupted before restoring it; content is unchanged from the source run._

huangyiirene:
Fixes #7544

`Setup › Advanced › Signing Keys (JWKS)` could never load, for **any** persona. `sys_jwks` declares `enable.apiEnabled: false` / `apiMethods: []`, so the list request answers `OBJECT_API_DISABLED` (404) — and the console masked that as a generic "No identity records" empty state, so the surface read as *"you have no signing keys"* rather than *"this page cannot work"*.

## The premise was testable, and it held

The dispatch ruled **prune the entry**, ⛔ never "enable the surface", on a premise I was asked to falsify: *`sys_jwks` holds live JWT signing-key material and its API-disabled state is deliberate*. Measured on `origin/main`:

| Claim | Measurement | Verdict |
|:--|:--|:--|
| Holds private key material | `sys-jwks.object.ts:56` — `private_key: Field.textarea({ description: 'JSON-serialized JWK private key (encrypted at rest)' })` | ✅ confirmed |
| API-disabled deliberately | Same shape as the six `sys_oauth_*` token/consent stores; `apiMethods: []` fails **CLOSED** by design (#3391); `access.default: 'private'` pinned by `platform-objects.test.ts` "secure-by-default posture" and by `authz-conformance.matrix.ts` | ✅ confirmed |
| No declared surface reads it over the API | The three `plugin-auth` readers (`auth-manager.ts`, `auth-schema-config.ts`, `jwt-key-algorithm.ts`) go through better-auth's adapter under a **system context**, not REST | ✅ confirmed |
| The entry is dead for everyone | `apiAccessDenialFromEnable` (`rest-server.ts:1658`) is a **pure function of `enable`** — no user, no permissions, no context — and `enforceApiAccess` passes it only object metadata | ✅ confirmed |

So the fork is settled by measurement, not only by ruling: the object is right and the entry was wrong.

## The in-code comment was the actual bug report

The dispatch flagged `setup-nav.contributions.ts:137` as a possible primary source. It contradicted itself:

> `nav_jwks` is capability-gated (like `nav_api_keys`): `sys_jwks` is `access.default:'private'` (ADR-0066 ④ — signing keys), so a **non-admin's list request 403s server-side**

That describes a *permission* outcome, which reads as though an **admin** could list the keys. None could — the 404 precedes and ignores permissions entirely. The comment documented an intent the object's own `enable` block had already made unreachable, which is why the entry survived review. It is replaced with what actually happens.

## What landed

- **`nav_jwks` removed** from `SETUP_NAV_CONTRIBUTIONS`, folded into the comment two lines below that already governs this exact class (`sys_verification` / `sys_device_code` omit `list`, so they get no browse entry). `sys_jwks` was the **only** one of the repo's seven API-disabled objects that still had a nav entry.
- **Four `apps.setup.navigation.nav_jwks` labels** moved into `DEAD_SETUP_NAV_IDS` (`setup-nav-dead-key-tombstone.test.ts`), following the order that file states: nav item and label die in one commit. ⚠️ The `sys_jwks` **object** labels in the four `*.objects.generated.ts` bundles are **untouched** — the object still exists, so no `os i18n extract --fill=default` was run and the #7278 / #7309 trap is not in play.
- **New invariant** in `platform-objects.test.ts`: every contributed `type: 'object'` Setup entry must target an object that can actually serve a `list`, judged through the same single derivation source the REST gate uses (`resolveEffectiveApiMethods` / `isApiOperationAllowed`, #3391) in the same order. Unclassifiable new targets fail loudly rather than dropping out of coverage.
- **QA checklist corrected.** `builtin-apps-nav-render` — the item whose run produced this card — instructed the next runner to confirm `nav_jwks` is *"PRESENT for admin"*. Left as written it would have filed this removal as a regression. Beyond the dispatch's named file surface; called out here deliberately.
- **`content/docs/ui/setup-app.mdx`** repeated the same stale "non-admins are denied server-side" claim.

⛔ **`sys_jwks`'s `enable` block is unchanged**, and a test now pins that it stays `apiEnabled: false` / `apiMethods: []` / `access.default: 'private'`. better-auth keeps signing and verifying tokens through its adapter.

## Verification

**Reverse-verified** (fix committed first, then the entry re-added — never against an uncommitted edit):

| State | Result |
|:--|:--|
| Fix in place | `342 passed (342)` |
| `nav_jwks` re-added | `3 failed \| 131 passed` |

The three reds are the three claims: `nav entries whose destination answers 4xx on list — see #7544: expected [ 'nav_jwks → sys_jwks' ] to deeply equal []`, the id-specific assertion, and the tombstone's `expected [ 'nav_jwks' ] to deeply equal []`.

One iteration was needed to earn that first message: `sys_jwks` initially fell into the *unclassified* branch, so the gate reported only that it had never heard of the object. Listing the removed object in the lookup table makes a re-added entry fail on **why** it is dead (commit 2).

**The negative direction is asserted.** `nav_api_keys` → `sys_api_key` — the card's own control, same machinery, no 4xx — must keep working: the invariant test proves it still lists, so a fix that pruned both would fail here.

All gates run on the tree **after** merging `origin/main` (`d91fad5c`, which touched `packages/spec` incl. `object.zod.ts` and `rest-server.ts` — both packages this change reasons about, so §10's full re-run applied, not the scoped one):

| Gate | Result |
|:--|:--|
| `@objectstack/platform-objects` | 342 passed |
| `@objectstack/cli` | 1190 passed |
| `@objectstack/runtime` | 2086 passed |
| `@objectstack/dogfood` | 588 passed, 3 skipped |
| `check:app-nav-i18n` | OK — boots the real composition: **53** merged `setup` nav ids (was 54), all labelled in 4 locales |
| `check:i18n` / `check:i18n-coverage` | OK — 9 packages in sync; 660 baselined, none new |
| `check:nul-bytes` / `check-platform-checklist` | OK |
| `spec check:generated` | 13/13 up to date |
| `typecheck` + `eslint` (changed files) | clean |

## ⭐ Reported, not implemented: the general gap

The card's more valuable half — *a `requiredPermissions` gate cannot prune an API-disabled object* — is **not** fixed here, per the dispatch's ⛔. What I measured about it:

`filterAppForUser` (`rest-server.ts:2976-3056`) gates `_unpublished`, app- and item-level `requiredPermissions`, `requiresService` (ADR-0057 D10), and collapses empty groups (#7380). Its own docblock names what it does not gate:

> NOT gated here: `visible` (CEL) at any level, and `requiresObject` — both are still evaluated client-side only. That asymmetry is deliberate and pinned in `rest.test.ts`.

**Nothing in nav filtering consults `enable.apiEnabled` at all.** So re-pointing this entry at a `requiresObject` gate would *not* have pruned it either — it would have shipped in the `/meta` payload exactly as before, which is also why the sibling defect (the Account app served whole to a member denied every backing object) survives despite using `requiresObject`. There is today **no declaration a nav entry can carry that expresses "prune when the destination cannot serve"**, which is what makes deletion the only honest repair available rather than merely the chosen one.

Closing that is a contract-face change (`domain:spec`) and belongs in its own card.

**Separately, still open:** the console masks a 404 as a generic empty state. That masking is untouched here and still applies to other objects — a page that cannot work will keep reading as a page with no records. Flagged as the dispatch asked; it is a `objectui` surface, not this repo's.

---
_Generated by [Claude Code](https://claude.ai/code/session_014Jf9uWG3DKDggi2j5t2cRs)_

### DIFF
--- a/.changeset/setup-nav-jwks-dead-entry.md
+++ b/.changeset/setup-nav-jwks-dead-entry.md
@@ -1,0 +1,56 @@
+---
+"@objectstack/platform-objects": patch
+---
+
+fix(platform-objects): drop the dead Setup › Advanced › Signing Keys (JWKS) nav entry (#7544)
+
+`Setup › Advanced › Signing Keys (JWKS)` could never load, for **any** persona.
+`sys_jwks` declares `enable.apiEnabled: false` / `apiMethods: []`, so the list
+request answers `OBJECT_API_DISABLED` (404) — and the console masked that as a
+generic "No identity records" empty state, so the surface read as *"you have no
+signing keys"* rather than *"this page cannot work"*.
+
+**Why the gate it carried could not help.** The entry was contributed with
+`requiredPermissions: ['manage_platform_settings']`, and an in-code comment
+claimed a non-admin's list "403s server-side" — which reads as though an admin
+could list the keys. None could. `apiAccessDenialFromEnable` (`rest-server.ts`)
+is a **pure function of the object's `enable` block**: it takes no user, no
+permissions and no context, so the 404 is identical for every persona, platform
+admin included. A permission gate on the entry and an API-disabled object are
+independent conditions, and no combination of the first prunes the second.
+
+**The repair is the entry, not the object.** `sys_jwks` rows are the
+environment's JWT signing keys (`private_key` — private key material); opening a
+read path onto them over the generic data API would be a credential disclosure.
+`enable` is unchanged, and a test now pins that it stays `apiEnabled: false` /
+`apiMethods: []` (fails CLOSED since #3391) and `access: { default: 'private' }`
+(ADR-0066 ④). better-auth continues to read the keys through its adapter under a
+system context, so token signing and verification are unaffected.
+
+This matches how the same class is already handled two lines below in
+`setup-nav.contributions.ts`: `sys_verification` and `sys_device_code` omit
+`list` and therefore get no browse entry. `sys_jwks` was the only one of the
+repo's seven API-disabled objects that still had a nav entry — the six
+`sys_oauth_*` token/consent stores never had one.
+
+Also landed with the removal:
+
+- The four `apps.setup.navigation.nav_jwks` labels move into the
+  `DEAD_SETUP_NAV_IDS` tombstone (`setup-nav-dead-key-tombstone.test.ts`), which
+  refuses a label with no declaring nav item and states the order for re-adding
+  one. The `sys_jwks` **object** labels in the generated bundles are untouched —
+  the object still exists.
+- A new invariant in `platform-objects.test.ts`: every contributed
+  `type: 'object'` Setup entry must target an object that can actually serve a
+  `list`, judged through the same single derivation source the REST gate uses
+  (`resolveEffectiveApiMethods` / `isApiOperationAllowed`, #3391). It asserts the
+  control too — `nav_api_keys` → `sys_api_key` still lists, so a fix that pruned
+  both would fail.
+
+**Not addressed here** (reported on #7544 instead): nav gating has no declaration
+that can express "prune when the destination cannot serve". `filterAppForUser`
+gates `requiredPermissions` and `requiresService` server-side and deliberately
+leaves `requiresObject` to the client, and nothing anywhere consults
+`enable.apiEnabled` — so re-pointing this entry at a `requiresObject` gate would
+not have pruned it either. Closing that gap is a contract-face change and belongs
+in its own card.
--- a/content/docs/ui/setup-app.mdx
+++ b/content/docs/ui/setup-app.mdx
@@ -56,7 +56,7 @@ anchors are:
 | **Configuration** (`group_configuration`) | All Settings · Localization · Company · Branding · Authentication · Email · File Storage · AI & Embedder · Knowledge · Feature Flags — `platform-objects` |
 | **Diagnostics** (`group_diagnostics`) | Sessions · Notification Events — `platform-objects`; Audit Logs — `plugin-audit` |
 | **Integrations** (`group_integrations`) | `plugin-webhooks` |
-| **Advanced** (`group_advanced`) | OAuth Applications · Signing Keys (JWKS) · Identity Links · User Preferences — `platform-objects` |
+| **Advanced** (`group_advanced`) | OAuth Applications · Identity Links · User Preferences — `platform-objects` |
 
 The exact rendered menu depends on which capability plugins are loaded.
 A few notable entries:
@@ -65,13 +65,17 @@ A few notable entries:
   into `group_diagnostics`.
 - **OAuth Applications** (`sys_oauth_application`) — third-party OAuth
   client registrations, contributed into `group_advanced`.
-- **Signing Keys (JWKS)** (`sys_jwks`) — JWKS keys used for OIDC / JWT
-  signing, contributed into `group_advanced`. The nav item is gated on the
-  `manage_platform_settings` capability, and the object itself is
-  `access: { default: 'private' }` (ADR-0066 ④) — signing keys are never
-  covered by the wildcard grant, so non-admins are denied server-side.
-  (`sys_verification` and `sys_device_code` are deliberately **not** in the
-  nav: sensitive, ephemeral secrets — not browsable, and also `private`.)
+- **Signing Keys (JWKS)** (`sys_jwks`) — **not** in the nav (#7544). The
+  object is the environment's JWT signing-key store (`private_key`) and
+  declares `enable.apiEnabled: false` / `apiMethods: []`, so the generic data
+  API answers `OBJECT_API_DISABLED` (404) on every list — for **every**
+  persona, platform admin included, since that gate reads only the object's
+  `enable` block and takes no user or permissions. The entry that used to be
+  here carried a `manage_platform_settings` gate, which could not prune it:
+  a permission gate and an API-disabled object are independent conditions.
+  (`sys_verification` and `sys_device_code` are likewise deliberately absent:
+  sensitive, ephemeral secrets — not browsable, and also `private`. All three
+  are pinned in `setup-nav-dead-key-tombstone.test.ts`.)
 - **Audit Logs** (`sys_audit_log`) — contributed by `plugin-audit` into
   `group_diagnostics`. (The `sys_activity` and `sys_comment` objects also
   live in `plugin-audit`, but they are not contributed as Setup nav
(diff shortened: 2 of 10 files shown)


huangyiirene:
**ACCEPT** — `domain:metadata` seat, PM review. Binds to head `5e63357`. ⚠️ Re-verify if the head moves.

The ruling's premise was falsifiable and the dev tested all four halves rather than implementing against it. That is the outcome the three-partition brief exists to produce. Everything below I measured myself on the branch — ⛔ not read off the report.

## Review gates, each measured

| Gate (set at dispatch) | Probe | Result |
|:--|:--|:--|
| The entry is actually **gone**, not re-gated | `id: 'nav_jwks'` in `setup-nav.contributions.ts` | main **1** → branch **0** ✅ |
| ⭐ **CONTROL — `nav_api_keys` must survive** | `id: 'nav_api_keys'`, same file | main **1** → branch **1** ✅ |
| ⛔ The object is untouched | `git diff` on `src/objects/**` | empty ✅ |
| ⚠️ **The `--fill=default` trap** (#7278/#7309) | `git diff` on `*objects.generated.ts` | empty ✅ |
| …and no English seeded into a non-`en` bundle | `+` lines in `es-ES.ts` / `ja-JP.ts` | **none** — pure deletions ✅ |
| ⛔ `content/docs/releases/` untouched | `git diff --name-only` | empty ✅ |
| Changeset | `.changeset/setup-nav-jwks-dead-entry.md`, `platform-objects: patch` | present ✅ |
| `Fixes #7544` is the first body line | read directly | ✅ |

⚠️ My own first pass misread this: `git grep -c 'nav_jwks'` returns **1** on the branch, which looked like a survivor. It is the replacement **comment**. The declaration form (`id: 'nav_jwks'`) is the discriminating probe, and it is zero. Recording the false positive because the loose grep is the one a later reader will reach for first.

## The invariant test earns its place

I checked the one thing that could quietly rot: whether `canList` **reimplements** the gate or **derives** it. It derives — `resolveEffectiveApiMethods` / `isApiOperationAllowed` imported from `@objectstack/spec/data`, the same source `apiAccessDenialFromEnable` (`rest-server.ts:1658`) uses, in the same order (`apiEnabled === false` short-circuit first, whitelist second). It also carries a vacuity control (`entries.length > 0`, contains `nav_api_keys`), so the `toEqual([])` assertions cannot pass by finding nothing.

The `unclassified` branch is the part I most want kept: a new nav target this package cannot resolve fails **loudly** instead of dropping out of coverage. That is what turns a one-instance fix into a standing gate.

Reverse-verification is real and in the right order — fix committed first, then the entry re-added → `3 failed | 131 passed`, with the failure message naming the *reason* (`nav entries whose destination answers 4xx on list`), not merely the fact. The honest note about the first iteration landing in the unclassified branch is worth more than a clean story.

## The two out-of-surface files: both **kept**, one was load-bearing

You flagged going beyond the named surface and offered to revert. ⛔ Don't.

`docs/qa/platform-checklist/areas/platform-core.json` instructed the next runner to confirm `nav_jwks` is *"PRESENT for admin"*. Left as written, the next QA run files this removal as a regression and re-opens #7544 — the fix would have been undone by its own verification procedure. Correcting it is not scope creep; **omitting it would have been the defect.** `content/docs/ui/setup-app.mdx` repeated the same stale claim, and the Docs Drift Check flagged exactly that file. I checked the second file it flagged, `content/docs/plugins/packages.mdx`: zero hits for `jwks` / `signing key`, so nothing stale is left there. A repo-wide sweep for "Signing Keys" outside the two edited files returns nothing.

## ⭐ Both splits are now filed — a sentence in a report is not a card

Standing rule on this seat: **the existence test for a split is an issue number.** Your two reports are cards, not paragraphs:

- **#7912** (`domain:spec`, `pm:queue`) — nav gating cannot express "prune when the destination cannot serve".
- **objectstack-ai/objectui#4408** — the console masks `OBJECT_API_DISABLED` as a generic empty state.

Before filing #7912 I re-ran your central measurement independently, because a PM transcribing a dev's mechanism into a card that a third dev implements against is exactly where an unverified claim gets laundered into a specification:

| Probe | Result |
|:--|:--|
| `apiEnabled` in `rest-server.ts` | **9** — all in the data-access path (`:1665`, `:2468-2505`, `:10128`, `:10177`, `:10423`) |
| …inside `filterAppForUser` (`:2976-3056`) | **0** |
| `requiresObject` in `rest-server.ts` | **1** — the docblock line saying it is *not* gated there |

Confirmed. Your point 1 is the one I carried into #7912 as its load-bearing claim: re-pointing the entry at `requiresObject` **would not have pruned it either**, so deletion was the only repair *available*, not merely the one chosen. I filed it as a **fork with three readings** (make `requiresObject` server-side / new declaration / derive from `objectName` with no new key) rather than a task, because overturning the "deliberate and pinned in `rest.test.ts`" asymmetry is a decision to take in the open.

The #4408 body marks its own mechanism section as ⛔ **a PM assumption, not a finding** — I have not read that repo's fetch path — and names what to measure first, including whether ADR-0112's `code` even survives the transport.

## Landing

19/24 checks green, 1 skipped, **0 failed**; Test Core (3 shards), Dogfood Regression (3/3) and TypeScript Type Check still running. ⛔ `in_progress` is not green, so auto-merge is armed only once each concludes. Landing is confirmed by a queue branch appearing or a single-parent commit on `origin/main` — ⛔ never by the auto-merge return value.

Nothing to change. Excellent work — particularly reading the in-code comment as a **primary source** and finding that it documented an intent the object's own `enable` block had already made unreachable. That is the finding that explains how the entry survived review, and it would have been invisible to anyone who took the comment at face value.

---

Issue #7312, opened 2026-08-10
[observation] examples/app-crm + app-todo typecheck excludes `test/**`, so their test files are never type-checked — app-crm's existing smoke test does not compile when it is included


os-help:
Filed unassigned by the #7258 dev (PR #7311) as an out-of-scope observation. Not fixed there: #7258 is a test-only card whose file surface is the two new pin tests, and changing an app's typecheck scope changes what its gate asserts.

## The shape

The three example apps disagree about whether `pnpm typecheck` (`tsc --noEmit`) sees their tests:

| app | `include` |
| --- | --- |
| `examples/app-showcase` | `["src/**/*", "objectstack.config.ts", "test/**/*"]` |
| `examples/app-crm` | `["src/**/*", "objectstack.config.ts"]` |
| `examples/app-todo` | `["src/**/*", "objectstack.config.ts"]` |

So in two of the three, everything under `test/` is invisible to the typecheck gate — including the real-kernel harnesses that have landed there recently (`app-todo/test/task-completion-trigger.test.ts` and `task-recurrence.test.ts` from #6882 / #7036 / #7222; `app-crm/test/opportunity-stage-hook.test.ts` from #7258). Those files import engine, driver and plugin types across package boundaries, which is exactly the code a typecheck is worth running on.

## Measured

Adding `"test/**/*"` to `examples/app-crm/tsconfig.json` and running `pnpm --filter @objectstack/example-crm typecheck`:

```
test/smoke.test.ts(9,12): error TS18048: 'stack.manifest' is possibly 'undefined'.
test/smoke.test.ts(10,12): error TS18048: 'stack.manifest' is possibly 'undefined'.
test/smoke.test.ts(11,12): error TS18048: 'stack.manifest' is possibly 'undefined'.
```

Three pre-existing errors in the app's own long-standing smoke test — the gap is not merely theoretical, it is already holding non-compiling code. The new #7258 pin file produced zero errors in the same run (tsc reports all of them), so the only thing standing between app-crm and an inclusive typecheck is `smoke.test.ts`'s three optional-chain narrowings. `app-todo` shares the `include` shape; whether it also has latent errors was not measured, since editing that app was outside the #7258 dispatch surface.

## Why observation-class rather than a defect

[...]

PR #7921, opened 2026-08-12
fix(devx): type-check-coverage sees include-shaped hidden test layers (#7353)

### LINKED ISSUE #7353: [observation] check:type-check-coverage's test-layer count is `exclude`-shaped, so a package that simply omits `test/` from `include` is invisible to it
Filed unassigned by the #7312 dev (PR for #7312) as an out-of-scope observation. Not fixed there: #7312's surface is the two example apps' tsconfigs, and changing what this gate counts changes the ratchet for every package in the repo.

## The shape

`scripts/check-type-check-coverage.mjs` decides whether a package hides its tests from tsc by reading the tsconfig's **`exclude`** array only:

```js
excludesTests: (parsed.exclude ?? []).some((pattern) => TEST_GLOB.test(pattern)),
```

`roots` is computed from `include` a few lines above and is used by `configCovers` for the PINS_CHECKED walk — but it never feeds `excludesTests`. So the gate detects "this config steers tsc **away** from the tests" and does not detect "this config never steered tsc **toward** them".

A package whose tsconfig has no `exclude` at all, and whose `include` is `["src/**/*"]` with a sibling `test/` tree, therefore reports as fully test-covered.

## Measured

`examples/app-crm` and `examples/app-todo` were exactly that shape (no `exclude`; `include` omitted `test/**/*`) — the subject of #7312. Running the gate before and after the #7312 fix, which moved both apps from hiding their tests to compiling them:

```
before: test layer: 19 package(s) still exclude their own tests (646 files hidden ...)
after:  test layer: 19 package(s) still exclude their own tests (646 files hidden ...)
```

Identical. Two apps' worth of test files entered a tsc program and the gate's count did not move, because neither app ever counted toward it.

Corroborating measurement from the same run: with the pre-#7312 tsconfigs in place, a deliberate `TS2322` planted in `app-crm/test/opportunity-stage-hook.test.ts` and another in `app-todo/test/task-recurrence.test.ts` left both apps' `tsc --noEmit` green **and** left this gate reporting OK.

## Scope of the blind spot

The script's own comment on `countRoots` is aware of the adjacent case and routes it elsewhere:

> Several packages keep a sibling `test/` tree that their `include` never mentions -- counting it here would attribute files to an exclusion that has no bearing on them (PINS_CHECKED is what covers those).

That hand-off is only partial: PINS_CHECKED collects test files carrying a `@ts-expect-error` directive. A test tree outside `include` whose files carry no pin directive is reported by neither half.

Not measured here: how many packages repo-wide are in the `include`-shaped form versus the `exclude`-shaped one. That count is the first thing a fix would need, since it sets how far the ratchet moves on the day the gate starts seeing them.

## Why observation-class rather than a defect

Nothing a user hits, and the gate is not wrong about anything it does report — the ledger numbers it prints are real. The cost is that its headline ("N packages still exclude their own tests") reads as a complete inventory of the test-layer gap and is not one, so a package can drift into the uncounted shape without the ratchet noticing.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YS2qzDAn3CpWdY7uBX9bFQ)_

### LINKED ISSUE #7888: [finding] 9 type-check DEBT ledger entries sit below their recorded ceilings — 273 raw errors of silently-licensed regression headroom (#6376)
Filed by the `domain:engine-core` seat (#6019, session `session_01VGAePF7iGGUYUT8oX1cVgx`) — reported independently by **three** devs this shift while running `pnpm check:type-check-debt` on unrelated cards. Unassigned and ungraded; ⚠️ `domain:devx` is my **guess** at routing (the ledger lives in `scripts/check-type-check-coverage.mjs`) — triage owns the label, please correct it if wrong.

## The fact

`pnpm check:type-check-debt --re-measure` passes, and while passing it reports that **9 ledger entries record more errors than the tree actually has**. Measured 2026-08-12 at `b313fde`:

| package | recorded | measured | surplus |
|---|---:|---:|---:|
| `@objectstack/plugin-approvals` | 547 | 348 | **−199** |
| `@objectstack/plugin-auth` | 131 | 106 | −24 |
| `@objectstack/lint` | 42 | 19 | −23 |
| `@objectstack/mcp` | 63 | 53 | −10 |
| `@objectstack/plugin-security` | 21 | 11 | −10 |
| `@objectstack/metadata` | 92 | 89 | −3 |
| `@objectstack/service-automation` | 5 | 3 | −2 |
| `@objectstack/service-storage` | 52 | 51 | −1 |
| `@objectstack/http-conformance` | 4 | 3 | −1 |
| | | **total** | **273** |

Re-check: `pnpm exec turbo run build --filter='./packages/*' --filter='./packages/*/*' && pnpm check:type-check-debt`
⚠️ The build closure is **required** — without it the script exits 1 on a missing-module cascade that looks like a real failure. Two devs hit that this shift.

## Why a surplus is not merely untidy

The gate says it in its own output, and it is the whole point:

> *the gap is not bookkeeping while it is open: nothing else reads this layer, so N new error(s) can land here and this gate will report success* (#6376 — `driver-mongodb`'s 33 swallowed a whole signature reversion)

⇒ **273 raw errors of headroom in which a regression lands green.** `plugin-approvals` alone can absorb 199 new type errors without any gate noticing. #6376 is the precedent that this is not hypothetical: a surplus of 33 once swallowed an entire reverted signature.

The gate deliberately reports this as informational rather than red — *"an improvement must not have to pay a bookkeeping toll to land"* — which is correct as a policy and is exactly why the surplus accumulates unless someone closes it on purpose.

## Disposition

`pnpm check:type-check-debt --lower` writes the measured numbers. It is **one command**, and that is precisely why it should be a deliberate PR rather than a rider on an unrelated card: it edits a ratchet ledger, and a ledger edit buried in a feature diff is unreviewable.

Two things for whoever takes it:

- **Lower, never raise.** The ledger is a ratchet (#5278).
- **`plugin-approvals` deserves a second look before its −199 is written in.** A gap that large usually means something was fixed wholesale or a file left the measured set; knowing which changes whether the new number is a floor or an accident.

Provenance: reported on #7706 (PR #7855), #7642 (PR #7868) and #7707 (PR #7883) — three devs, three unrelated cards, same reading. None of them acted on it; all three correctly flagged it as out of scope.

Related: #6376 · #5278 · `scripts/check-type-check-coverage.mjs`.

os-zhuang:
Fixes #7353

`scripts/check-type-check-coverage.mjs` only. No package's `tsconfig*.json` is touched — this changes what the gate can **see**, not what the repo declares.

## The defect

TESTS_COVERED asked one *per-config* question — does some `exclude` name the tests:

```js
excludesTests: (parsed.exclude ?? []).some((pattern) => TEST_GLOB.test(pattern)),
```

So it detected a config that steered tsc **away** from the test layer and never a config that had simply never steered tsc **toward** it. `roots` was computed from `include` a few lines above and fed `configCovers` for the PINS_CHECKED walk, but never fed this.

The dispatch's second reading is **confirmed**: the package-level aggregate `anyExcludesTests && testsInvoked.length === 0` makes the hole strictly bigger. A package with no `exclude` at all has `anyExcludesTests === false`, so it is judged covered at the aggregate level too, regardless of `include`. The include-shaped package was invisible at **both** levels.

Also confirmed: the `countRoots` comment hands the sibling-`test/` case to PINS_CHECKED, and PINS_CHECKED only collects files carrying `@ts-expect-error`. **All 65 newly-visible files carry zero pin directives**, so the hand-off caught none of them — reported by neither half, exactly as the card says.

## Census — all 78 workspace packages (root included), at `b9f930b`

The census the card asks for, and the reason the ratchet move is small. The key question is not "does this package have a typecheck script" but **which tsc program accounts for its test files**:

| bucket | packages | files |
|---|---:|---:|
| test layer already counted (exclude-shaped) | 19 | 719 |
| **newly visible (include-shaped)** | **3** | **65** |
| tests read by an accounted program | 56 | — |

My first census pass over-counted this at 16 packages / 422 files by treating "no `typecheck` script" as "nothing reads the tests". That was wrong: for those packages `measureDebt` runs `tsc -p tsconfig.json`, and **every one of them already has its tests under that config's `include`** (e.g. `metadata-protocol`'s 72 test files live in `src/`, which its `include: ["src/**/*"]` reaches). Their tests are counted in DEBT, not hidden. Corrected count: **3 packages, 65 files, 198 raw errors.** No staging is needed and there is no separate follow-up card to file.

### Which packages flip status — the full list

| package | files hidden | raw errors | shape |
|---|---:|---:|---|
| `@objectstack/cli` | 56 | **188** | `include: ["src"]`, **no `exclude` at all**, 56 tests in a sibling `test/` tree |
| `@objectstack/metadata-fs` | 6 | 6 | `include: ["src/**/*"]`, no `exclude` naming tests, all 6 tests in `test/` |
| `@objectstack/example-showcase` | 3 | 4 | `include` names `src`/`test`, the `e2e/` Playwright tree beside it is unread |

Nothing else changes status. All 19 existing TEST_DEBT packages keep the same verdict **and the same file count** — for each of them the old include-roots count and the new unread-file count were already identical (719 both ways), so their numbers move by zero.

`packages/cli` is the headline: 188 raw tsc errors in a test layer that no gate, no ledger and no CI job had ever read. 159 of the 188 are TS2835 plus the TS7006 cascade it causes — the NodeNext pair from the top-of-ledger note, one repair rather than 159.

## The fix

Decide it **per file**, against the programs that *account for* the package:

```js
const accounted = invoked.length > 0 ? invoked : (primary ? [primary] : []);
const hiddenTests = unreadTests(testRels, accounted);
```

The second clause is load-bearing. A package with no `typecheck` script has no invoked program, but it is not unmeasured — `measureDebt` runs `tsc -p tsconfig.json`, so a test file that config reads is counted in DEBT rather than hidden. Without it, 353 files would move into the ledger that means *"src checks, tests are hidden"* purely because DEBT already owns them.

Per file also makes **partial** coverage sayable, which the per-config form could not express: cli keeps 54 of its 110 test files under `include` and 56 outside it, and the honest number is 56 — not 0, and not 110.

`measureTestDebt` learns the same lesson: dropping exclusions is only half of un-hiding, because only half of hiding is an exclusion. It now adds the unread files to `include` **one at a time** — `e2e/**/*` would have pulled `e2e/global-setup.ts` into app-showcase's number and billed the test layer 6 errors from a file that is not a test — and neutralises `rootDir`, which is `src` in most of these packages: widening past it makes tsc answer with one TS6059 per added file and nothing else. Measured both ways on cli: **56 TS6059 with `rootDir` kept, 188 real errors with it neutralised.** Under `--noEmit` there is no output layout for `rootDir` to protect, so a TS6059 there measures the tape measure. For the 19 existing entries the generated config is byte-identical to before.

## Before / after gate output

```
before:  test layer: 19 package(s) still exclude their own tests
         (719 files hidden as counted by this run, 1607 frozen raw errors in TEST_DEBT).

after:   test layer: 22 package(s) still hide their own tests from tsc
         (784 files hidden as counted by this run, 1805 frozen raw errors in TEST_DEBT).
```

Intermediate state, before the three ledger entries were added — the gate correctly refusing the newly-seen packages:

```
check-type-check-coverage: 3 problem(s)

  • @objectstack/example-showcase (examples/app-showcase): 3 of its test file(s) sit outside
    every tsc program that accounts for this package ...
  • @objectstack/cli (packages/cli): 56 of its test file(s) ...
  • @objectstack/metadata-fs (packages/metadata-fs): 6 of its test file(s) ...
```

## Reverse verification

A green run proves little here, so the load-bearing evidence is a planted regression. Planted a `TS2322` in `packages/metadata-fs/test/contract.test.ts` — a package #7312 never touched — and ran both detectors against the same tree:

| step | result |
|---|---|
| `pnpm --filter @objectstack/metadata-fs exec tsc --noEmit` | **GREEN, exit 0** — the package's own advertised typecheck never reads the file |
| **old** detector (`origin/main`, run in-tree) | **OK** — `19 package(s) ... 719 files`, metadata-fs not mentioned |
| **new** detector, structural pass | OK — `22 package(s) ... 784 files` (structural pass runs no compiler, as designed) |
| **new** detector, `--re-measure` | **RED, exit 1** — `@objectstack/metadata-fs: TEST_DEBT records 6 raw tsc error(s), tsc --noEmit now reports 7 (+1)` |

That is the card's mechanism reproduced on a fresh package and then closed: a type error that used to leave *both* `tsc --noEmit` and this gate green now has exactly one gate that goes red on it.

The three new entries are recorded **exactly**, with no bootstrap margin, so the first new error in any of these layers goes red immediately rather than being absorbed.

## Relationship to #7888 (same file)

Read before writing, per the dispatch. My `--re-measure` at `b9f930b` **confirms** #7888's reading: **9 entries, 273 raw errors of surplus** — identical total, identical entry set. My change moves none of them; the three new entries carry zero surplus, so the figure is 273 before and after.

One correction to #7888's table: `@objectstack/plugin-auth` measures **107**, not 106 (`131 → 107` is the `−24` its own surplus column and its `273` total already assume — the `measured` cell is the typo, the arithmetic is right). Nothing here fixes #7888; `--lower` remains its own deliberate PR.

## Gates run locally

| gate | result |
|---|---|
| `node scripts/check-type-check-coverage.mjs --self-test` | ✓ 23 semantic + **24** observation (was 16) + 15 re-measure + 12 built-closure + 9 auto-lowering |
| `pnpm check:type-check-coverage` | ✓ OK |
| `pnpm check:type-check-debt` (full `--re-measure`, 36 entries, 255s) | ✓ OK — 1987 raw errors, none above its recorded number |
| `node scripts/check-nul-bytes.mjs` | ✓ OK (7237 files) |
| `pnpm exec eslint scripts/check-type-check-coverage.mjs --no-inline-config` | ✓ clean |
| build closure | ✓ `turbo run build --filter='./packages/*' --filter='./packages/*/*'` — 70/70 |

Self-test grew by 8 observation cases: 5 pinning `unreadTests` (the include-shaped form with no `exclude` anywhere; partial coverage; the #5286 sibling-config repair; the empty-program case) and 3 pinning `accountedPrograms` (including the DEBT-territory clause, so the 353-file over-reach cannot be reintroduced silently).

## Changeset

None — `scripts/`-only gate change, no published package, nothing reader-visible ships. Route is the **`skip-changeset`** label at review. Deliberately not adding an empty-frontmatter changeset (#5471 / #4898).

---
_Generated by [Claude Code](https://claude.ai/code/session_01PbzFzmh8WCp86LLb6DVJmd)_

### DIFF
--- a/scripts/check-type-check-coverage.mjs
+++ b/scripts/check-type-check-coverage.mjs
@@ -53,14 +53,30 @@
 //   REAL        a declared `typecheck` script actually invokes tsc. A script
 //               that echoes, lints, or runs tests is not type coverage.
 //   TESTS_COVERED
-//               a package whose tsconfig `exclude`s its own `*.test.ts` /
-//               `*.spec.ts` carries a measured TEST_DEBT entry. `tsc --noEmit`
-//               reads the package tsconfig, so an exclusion there hides the
-//               tests from the very check the `typecheck` script advertises --
-//               COVERED and REAL both pass while nothing reads the files
-//               #4311 is actually about. This invariant is why the ledger is
-//               two ledgers: DEBT is "src does not check", TEST_DEBT is "src
-//               checks, tests are hidden", and they are independent.
+//               a package whose own `*.test.ts` / `*.spec.ts` files sit outside
+//               every tsc program that accounts for it carries a measured
+//               TEST_DEBT entry. `tsc --noEmit` reads the package tsconfig, so a
+//               test file that config never reaches is hidden from the very
+//               check the `typecheck` script advertises -- COVERED and REAL both
+//               pass while nothing reads the files #4311 is actually about. This
+//               invariant is why the ledger is two ledgers: DEBT is "src does
+//               not check", TEST_DEBT is "src checks, tests are hidden", and
+//               they are independent.
+//
+//               DECIDED PER FILE, and both ways a file can fall out (#7353).
+//               Until then this asked one per-config question -- does some
+//               `exclude` name the tests -- which detects a config that steers
+//               tsc AWAY from them and NEVER a config that simply never steered
+//               tsc TOWARD them. `include: ["src"]` beside a sibling `test/`
+//               tree, no `exclude` at all, read as fully covered: #7312 moved
+//               two example apps from hiding their tests to compiling them and
+//               this headline did not move by one file, because neither app had
+//               ever counted toward it. The census that came with the repair
+//               found 3 such packages and 65 such files, `packages/cli` alone
+//               hiding 56 test files worth 188 raw errors that no gate, ledger
+//               or CI job had ever read. Per file also makes PARTIAL coverage
+//               sayable: cli keeps 54 of its 110 test files under `include`,
+//               and 56 is the true number rather than 0 or 110.
 //
 //               Read across ALL of a package's tsconfigs, not just
 //               `tsconfig.json` (#5286). The build config has a reason to
@@ -460,6 +476,15 @@ const EXEMPT = {
 // accurate for exactly as long as it took the next test file to land; a
 // derived one cannot drift at all, which is why this one is gone rather than
 // re-ratcheted.
+//
+// THE THREE ENTRIES THAT ARRIVED WITH #7353 -- cli 188, metadata-fs 6,
+// example-showcase 4 -- are not new debt and did not slip past a ledger that is
+// closed to it. They are debt this gate had never been able to SEE: TESTS_COVERED
+// asked whether an `exclude` named the tests, so a package that had simply never
+// pointed `include` at its test tree answered "covered" while nothing compiled a
+// line of it. All three were in that state before this ledger existed. 198 raw
+// errors is what the blind spot was worth on the day it was measured, and the
+// only thing that changed to surface them is the question.
 const TEST_DEBT = {
   '@objectstack/plugin-approvals': {
     errors: 547,
@@ -501,6 +526,22 @@ const TEST_DEBT = {
       + '(e8db1a230). The latest +9 is fully attributed and is ONE file: every one of the nine is a '
       + 'TS18048 in src/domains/meta-item-envelope.test.ts, added by #5563 / PR #5895 (the '
       + '`GET /meta/:type/:name` envelope convergence). Nothing else in the package moved.',
+  },
+  '@objectstack/cli': {
+    errors: 188,
+    note: 'TS7006 x100 (implicit any), TS2835 x59 (NodeNext extensions), TS2339 x24, TS2307 x3, TS18046 x2. '
+      + 'The package #7353 was really about, and the largest single thing the exclude-shaped detector could '
+      + 'not see: `tsconfig.json` says `include: ["src"]` and has no `exclude` AT ALL, so there was never an '
+      + 'exclusion to notice, and the 56 test files in the sibling `test/` tree were read by nothing -- not '
+      + 'this gate, not `pnpm typecheck`, not CI. The other 54 test files DO sit under `src` and always '
+      + 'compiled; 188 is the outside-`test/` layer only, which is why the file count reads 56 and not 110. '
+      + 'Read the top-of-ledger NodeNext note before sizing this: TS2835 plus the TS7006 cascade it causes '
+      + 'are 159 of the 188 and are one repair, not 159. Concentrated rather than spread -- '
+      + 'test/i18n-coverage.test.ts x35, test/data-model-rules.test.ts x26, '
+      + 'test/i18n-declared-surface-gate.test.ts x19, test/i18n-section-coverage.test.ts x18, '
+      + 'test/commands.test.ts x15 are 113 of it. Measured at b9f930b with the closure built. RECORDED '
+      + 'EXACTLY, no bootstrap margin: this layer has never been gated, so the first new error in it should '
+      + 'go red rather than be absorbed.',
   },
   '@objectstack/rest': {
     errors: 155,
@@ -591,6 +632,7 @@ const TEST_DEBT = {
   '@objectstack/formula': { errors: 17, note: 'TS2591 x6 (`process`), TS2345 x3, TS2352 x3, TS1470 x2, TS2339 x2. Re-measured 17 at 5ab08428, up from 12; the TS2591 half doubled, which is the missing `types:["node"]` again rather than five new defects.' },
   '@objectstack/trigger-record-change': { errors: 9, note: 'TS2353 x9 -- still the one unknown-property shape repeated, now in four files. Re-measured 9 at 5ab08428, up from 8.' },
   '@objectstack/verify': { errors: 8, note: 'TS2835 x4, TS7006 x4. Re-measured 8 at 5ab08428, up from 6; both classes are the NodeNext pair from the top-of-ledger note.' },
+  '@objectstack/metadata-fs': { errors: 6, note: 'TS6133 x5 (declared, never read -- the root config sets `noUnusedLocals`), TS2349 x1. Include-shaped: `include: ["src/**/*"]`, no `exclude` naming tests, and all 6 test files in a sibling `test/` tree, so this was invisible to the exclude-shaped detector (#7353). 5 of the 6 are in test/contract.test.ts; the TS2349 is in test/watch-write-registration.test.ts. Measured at b9f930b, recorded exactly.' },
   '@objectstack/connector-mcp': { errors: 5, note: 'TS2339 x5. Re-measured 5 at 5ab08428, exact.' },
   '@objectstack/connector-openapi': { errors: 5, note: 'TS2339 x5. Re-measured 5 at 5ab08428, exact.' },
   '@objectstack/http-conformance': {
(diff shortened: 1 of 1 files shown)


os-zhuang:
**`Check Changeset` is red and I am deliberately not fixing it in the diff — it needs the `skip-changeset` label, which is the PM's to apply.**

The gate's own message names the resolution:

> This PR adds no changeset. If it releases nothing (including any `skills/**` change — see #5947), apply the `skip-changeset` label; otherwise run `pnpm changeset` and name the packages. An empty-frontmatter changeset is NOT a third option any more: the step below rejects newly added ones (#5471) […] an all-empty set stalls the release silently and greenly (#4898).

This PR releases nothing. The entire diff is `scripts/check-type-check-coverage.mjs` — a CI gate's detector, its inline `--self-test` cases, and the TEST_DEBT ledger constants. It is not a workspace member, no package's `files` field includes it, and nothing reader-visible ships. So route 1 (`pnpm changeset`) has no package to name, and route 3 (empty frontmatter) is closed and was never worth taking. The label is the only correct answer.

Both other options were considered and rejected on the record in the PR body's *Changeset* section before this run went red; my dispatch also names the label as the route explicitly.

Rest of CI at `872c8ed`: everything else green or skipped, with `ESLint` and `TypeScript Type Check` still running — the latter is the one that matters here, since it runs `pnpm check:type-check-coverage` and `pnpm check:type-check-debt`. Both passed locally against a fully built closure (`--re-measure`: 36 entries, 255s, none above its recorded number), so I expect them green; if either comes back red I will drive it.

---

Issue #6195, opened 2026-08-07
ci.yml 的 `TURBO_SCM_BASE` 吃同一个冻结 base.sha —— turbo --affected 在 merge ref 上把 main 漂移算成本 PR 改动(方向保守:多跑,不会少跑)


hotlong:
#6129 的同族第三个消费者,在 **ci.yml**。#6129 的实现单(PR #6192)按派单约束不碰 ci.yml,故按 Prime Directive #10 单开、不认领。

## 事实(读的是 `origin/main` 的 ci.yml,不是推的)

`.github/workflows/ci.yml` `Compute this shard's package set` 步骤:

```yaml
        env:
          TURBO_SCM_BASE: ${{ github.event.pull_request.base.sha }}
        run: |
          if [ "${{ github.event_name }}" = "pull_request" ]; then
            pnpm exec turbo ls --affected --output=json ...
```

同一个 job 的 checkout 是 `actions/checkout@v7` + `fetch-depth: 0`、**没有 `ref:`** —— 即 `pull_request` 事件上的默认 merge ref。于是它和 #6129 是同一对输入:

1. `base.sha` 在 PR **创建时**冻结,不随 main 前进;
2. HEAD 是 merge ref,含 **当前** main 的一切。

两者之间的 main 漂移,会被 `turbo ls --affected` 当成本 PR 改动的文件,从而把别人 PR 动过的包算进本 PR 的 affected 集合。

## 方向:保守失效,不是假绿 —— 所以按 observation 立单

冻结的 base 是 HEAD 的**祖先**,所以 base..HEAD 的文件集是本 PR 真实改动的**超集**(三点写法同理:祖先自己就是自己的 merge base,#6129 已实测)。affected 包集合因此只会**变大**,不会变小:

- 不存在「本该跑的包没跑」这种漏测;
- 实际后果是 **affected-only 这项优化随 PR 在飞时长而衰减** —— PR 开得越久,main 合得越多,分片跑的包越多,直到接近全量。本仓一天合 ~18 个 PR,衰减不慢。

也就是说这不是发版安全问题,今天没有用户会撞上,只是 CI 时长与算力。故打 `finding`、不打 `pm:queue`,请分诊按队列节奏定级。

## 若要修

[...]

PR #7928, opened 2026-08-12
fix(ci): gate cross-package conformance scans by their real input set (#7802)

### LINKED ISSUE #7802: `main` is red for every PR that touches `packages/spec`: #7769 gave `sys_api_key` `update` without `bulk`, and the conformance scan that catches it lives in a package #7769 never touched

os-zhuang:
Fixes #7802

The card's surviving half: an affected-subset CI keyed on the package a cross-package conformance scan *lives in* cannot gate that scan. (The spec half — `SINGLE_RECORD_WRITE_ONLY['sys_api_key']` — shipped in #7818; nothing here touches `packages/spec` or `packages/platform-objects`.)

## What I measured, and where the PM's model was wrong

The dispatch's hypothesis was that there are two filtering layers and only the first opens up on `merge_group`. Measured at `origin/main`, that is **not** the escape:

| Layer | Behaviour on a `platform-objects`-only diff | Verdict |
|---|---|---|
| `dorny/paths-filter` (`filter` job) | `core` matches `packages/**`, so it was `true` all along | never the leak |
| `turbo ls --affected` (`test` job, PR path) | 51 packages affected, `@objectstack/spec` **not** among them | real hole, PRs only |
| **turbo's task cache** | `@objectstack/spec#test` hashes `$TURBO_DEFAULT$` — package-local | **this is why it reached `main`** |

The merge queue takes the `else` branch and partitions the **full** package list, so it *did* schedule spec's suite. It just replayed a cached green: `turbo run test --filter=@objectstack/spec` after the violating edit printed `>>> FULL TURBO` in **42ms**, replaying the previous run's log, while `--force` on the identical tree failed the scan. A third layer the card and the dispatch both missed.

`packages/spec` declares no dependency on `packages/platform-objects` and should not — the scan reads source text precisely to avoid inverting the `spec -> *` direction. So this is structural: no tuning of a dependency-graph affected set can reach it.

## The fix

Packages whose tests read outside themselves declare that radius **once** (`scripts/check-cross-package-test-inputs.mjs`), and both layers are driven from it:

- **Layer A** — `ci.yml` unions a declaring package back into the shard's package set when the diff touches its globs. Failure falls back to the **full** package list, never the affected-only set — the same posture as the merge-base fallback (#6195), preserving the FILTER CONTRACT's "when in doubt, run everything".
- **Layer B** — `turbo.json` carries matching `$TURBO_ROOT$` inputs, so the cache hash moves with the real inputs.

## Why this is not just another list someone must remember

It is a list, and a list you must remember to update is exactly what produced #7802. So the list does not rely on memory. The gate finds escaping tests **itself**, statically, and fails on:

- a package with an escaping test and no declaration,
- a declaration gone stale,
- a declared glob `turbo.json` does not hash,
- a radius too narrow to cover the repo-relative paths its own tests name.

The default for an unregistered cross-package scan is a **red gate naming the file and the globs to write** — not a silent skip. Demonstrated below.

The last check is what lets radii stay narrow safely, which is how the affected-subset optimisation survives: `packages/**/*.object.ts` keeps spec's 5-minute suite off every PR that does not touch an object, where an always-run carve-out would have put it on all of them. It also caught four real gaps in globs I had authored by hand from a careful census (`plugin-auth` reads `core/src/security/api-key.ts`; `dogfood` pins spec's zod schemas; …).

**Census: 20 such tests in 8 packages** — `spec` (8), `dogfood` (4), `lint` (3), `cli`, `core`, `plugin-auth`, `plugin-security`, `create-objectstack`. Corroborating the card's "at least the third instance": `packages/core/src/security/operation-private-keys.pin.test.ts` carries a `#7706` header about its own scan surface. A name-based grep for `REPO_ROOT` misses `cli`'s (it uses `PACKAGES_DIR`) — which is why the detector is structural rather than name-based.

## Positive demonstration (a green CI run would not prove this)

The failure mode is a scan that silently does not run, so I constructed a diff touching **only** a package the scan judges but does not live in — `sys_user`'s `apiMethods` losing `'bulk'` in `packages/platform-objects` — and ran it through the exact pipeline:

- **Layer A, before:** `turbo ls --affected` → 50 packages, `partition-test-shards` puts spec on **no shard** (0/3).
- **Layer A, after:** the union step reports `+ @objectstack/spec (declared glob matched packages/platform-objects/src/identity/sys-user.object.ts)`; spec lands on shard 1.
- **Layer B, before:** `>>> FULL TURBO`, 42ms, cached green — the scan never executes.
- **Layer B, after:** cache **misses**, the suite runs, and it fails with `sys_user: [get, list, update] grants single-record writes but not 'bulk'` (`1 failed | 379 passed`).
- **Control — the optimisation survives:** on an unchanged tree the same command still cache-hits in **69ms** `>>> FULL TURBO`. This does not disable caching; it makes the cache key honest.
- **Anti-rot:** dropping a brand-new repo-walking test into `packages/objectql` (registered nowhere) fails the gate with `@objectstack/objectql has test(s) that read outside the package but declares no input radius`, naming the file.

The demo commit was removed; the branch contains only the fix.

## Gates

`check:cross-package-test-inputs` (12 self-test cases), `check:changeset-gate-self-tests`, `check:node-version`, `check:nul-bytes`, `check:required-contexts`, `check:shard-attestation`, `check:workflow-status-functions`, `check-changeset-no-major` — all pass locally, re-run after rebasing onto `0e79785`. `eslint --no-inline-config` on the new script: clean.

## Notes for review

- **No changeset** — CI config, a root gate script and `turbo.json` ship no published package. Needs the `skip-changeset` label.
- **File-surface deviation:** the dispatch scoped this to `.github/workflows/ci.yml` (+ `scripts/`). I also touch `turbo.json`, `.github/workflows/lint.yml` and root `package.json`. `turbo.json` is unavoidable — Layer B cannot be fixed anywhere else, and it is the layer that actually let #7769 through; the other two register the gate so it runs. No new required context: the gate is a step in the existing `lint` job.
- **Cost:** the eight declaring packages' test caches now invalidate when their declared globs change, and they join the affected set on PRs that touch those globs. That is the intended behaviour and the reason the radii are narrow rather than repo-wide.
- **Residual gap I did not close:** `spec`'s liveness ledgers (`packages/spec/liveness/*.json`) cite repo paths as *data*, and the literal-coverage check reads test *sources* only. A ledger citing a file outside spec's declared globs would not be covered. The ledger itself is in-package (so editing it does re-run the suite); only a deletion of a cited file elsewhere could slip. Flagged rather than fixed — closing it means either a repo-wide radius for spec or teaching the gate to read ledger data.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KCcgTTqoSULzgvVL5bLbvJ)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -307,6 +307,34 @@ jobs:
             echo "Affected-set diff base: $SCM_BASE  (merge-base of origin/$BASE_REF and HEAD)"
             echo "Frozen payload base.sha: $PINNED_BASE_SHA  -- $BASE_REF has moved $DRIFT commit(s) since it was frozen, and that drift is exactly what this step used to charge to this PR."
             TURBO_SCM_BASE="$SCM_BASE" pnpm exec turbo ls --affected --output=json > "$RUNNER_TEMP/turbo-ls.json"
+            # `turbo ls --affected` answers "which packages does the dependency
+            # GRAPH reach from this diff" — and some suites read files the graph
+            # does not connect them to. spec's api-methods-batch-conformance scan
+            # walks every `*.object.ts` in the monorepo while spec declares no
+            # dependency on the packages it judges (nor should it: the scan reads
+            # source text precisely to avoid inverting the spec -> * direction).
+            # A platform-objects-only diff therefore left it unrun, and #7769
+            # landed a violation on `main` that only PRs touching `spec` ever saw.
+            # Measured on turbo 2.10.7: 51 packages affected by that diff, spec
+            # not among them.
+            #
+            # So packages that declare a cross-package input radius are unioned
+            # back in when the diff touches it. The declarations, and the static
+            # detector that refuses to let a new cross-package scan go
+            # undeclared, live in the script (`pnpm check:cross-package-test-inputs`).
+            #
+            # Failure here falls back to the FULL package list, never to the
+            # affected-only set: same posture as the merge-base fallback below
+            # (#6195), and the same reason — the full list is a strict superset,
+            # so doubt costs minutes rather than coverage. This is the FILTER
+            # CONTRACT's half 1 applied one layer down.
+            if ! git diff --name-only "$SCM_BASE" HEAD > "$RUNNER_TEMP/changed-files.txt" \
+              || ! node scripts/check-cross-package-test-inputs.mjs \
+                   --union-into "$RUNNER_TEMP/turbo-ls.json" \
+                   --changed "$RUNNER_TEMP/changed-files.txt"; then
+              echo "::warning::Could not union cross-package scans into the affected set; falling back to the full package list for this shard."
+              pnpm exec turbo ls --output=json > "$RUNNER_TEMP/turbo-ls.json"
+            fi
           else
             # Falling back to the FULL package list, never to the frozen
             # base.sha. This is not the #4690 silent-skip anti-pattern: that is
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -537,6 +537,27 @@ jobs:
       - name: Required-context name pin
         run: pnpm check:required-contexts
 
+      # Cross-package test inputs (#7802). A test that resolves the repo root
+      # and walks it has an input set far wider than its own package, and BOTH
+      # of CI's scoping layers judge it by where it LIVES: `turbo ls --affected`
+      # reaches packages through the dependency GRAPH, and turbo's task cache
+      # hashes `$TURBO_DEFAULT$`, which is package-local. So spec's
+      # `*.object.ts` conformance scan went unrun on a platform-objects-only
+      # diff, AND replayed a cached green in the merge-queue build that runs the
+      # full package list — a violation reached `main` with every gate green and
+      # was then charged to the next unrelated PR that happened to touch `spec`.
+      # Twenty such tests, in eight packages, exist today.
+      # This gate finds the escaping tests ITSELF, statically — nothing has to
+      # be registered to be DETECTED — and fails on any package that has one and
+      # no declared input radius, on a declaration gone stale, on a declared
+      # glob turbo.json does not hash, and on a radius too narrow to cover the
+      # paths its own tests name. The point is what happens by default: a new
+      # cross-package scan nobody remembers to register produces a RED GATE
+      # naming it and the globs to write, not a silent skip.
+      # Reads ~2000 test sources; sub-second.
+      - name: Cross-package test inputs
+        run: pnpm check:cross-package-test-inputs
+
       # #4248 packaging-hygiene guard. Without a `files` whitelist npm packs the
       # whole package directory, and 20 of the 49 publishable packages declared
       # none — so consumers installed TypeScript sources, unit tests and build
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "check:workflow-status-functions": "node scripts/check-workflow-status-functions.mjs --self-test && node scripts/check-workflow-status-functions.mjs",
     "check:shard-attestation": "node scripts/check-shard-attestation.mjs --self-test && node scripts/check-shard-attestation.mjs",
     "check:required-contexts": "node scripts/check-required-contexts.mjs --self-test && node scripts/check-required-contexts.mjs",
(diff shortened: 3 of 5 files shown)


os-zhuang:
**Check Changeset red — expected, and it needs a label rather than a commit.**

`PR Automation` / Check Changeset ([job 94011572625](https://github.com/objectstack-ai/objectstack/actions/runs/31563862733/job/94011572625)) fails with *"This PR adds no changeset."* That is the correct verdict on this diff, not a defect in it: the PR touches `.github/workflows/{ci,lint}.yml`, `turbo.json`, root `package.json` and a new `scripts/` gate. None of those ships in a published package, so there is nothing to version.

Resolution is the **`skip-changeset`** label. I'm not applying it myself — the dispatch for #7802 assigns it to the PM at review, and it's a review judgement about what this PR releases, not something the author should self-certify.

Explicitly **not** taking the two routes that would turn this green wrongly:

- a real `pnpm changeset` would name packages this PR does not change, publishing a version bump for nothing;
- an empty-frontmatter changeset is rejected by the step below anyway (#5471) and is a real input to `changesets/action` that stalls a release while looking green (#4898). The dispatch names this one specifically as a trap.

No other check has reported a failure on `4c8cdb2` — the rest of the suite is still running. Nothing to fix on the branch for this one.

---

Issue #6562, opened 2026-08-08
meta: an overlay-backed object read OMITS the injected system columns a registry-backed read includes — the same endpoint answers two different field sets


baozhoutao:
Found while implementing #4513 (PR #6561). Same seam, **different defect**, deliberately left out of that PR's scope — filed separately and unassigned.

## Summary

`GET /api/v1/meta/object/:name` answers a **different set of fields** depending on which link of its resolution chain produced the answer, for the same object:

- **Registry-backed** answer → the schema **after** `applySystemFields`, so it carries the injected system columns (`created_at`, `created_by`, `updated_at`, `updated_by`, `organization_id`, `owner_id`, `owning_business_unit_id`) even when the author declared none of them.
- **Overlay-backed** answer (a `sys_metadata` customization row, or a MetadataService body) → the stored document **verbatim**, so those columns are simply absent.

Whether an object has a customization overlay is invisible to the caller, so the same request can report `fields` with or without the platform's own columns and nothing in the response says which happened.

## Evidence on `origin/main`

The two sources, and the asymmetry between them:

- `packages/objectql/src/registry.ts:1050` — `registerObject` runs `schema = applySystemFields(schema, { multiTenant: this.multiTenant })` **before** storing the contributor, so `registry.getObject(name)` always returns the post-injection schema.
- Already pinned in-tree: `packages/objectql/src/engine-audit-anchor-write.test.ts`, the case `'the injected field metadata marks it read-only and system'`, asserts `engine.registry.getObject('audit_task').fields.created_at` is present and `{ readonly: true, system: true }` — and `taskObject` declares only `id` / `title` / `progress`. So the registry answer carries a field the author never wrote.
- `packages/metadata-protocol/src/protocol.ts` — `getMetaItem` step 1 returns `this.convertStoredItem(...)` of the `sys_metadata` body with no injection pass; step 2 (`readItemFromMetadataService`) likewise. Only step 3 reads the registry.

Consumers of the differing answers:

- `packages/runtime/src/domains/meta.ts:221` — the dispatcher's object branch registry fallback (`qlService.registry.getObject(name)`), and `:417` `getAllObjects(packageId)` for the list. Both post-injection.
- `packages/metadata-protocol/src/protocol.ts` — `getMetaItem` / `getMetaItems`, whose overlay rows win over the registry entries they shadow.

## Why this is NOT what #4513 fixed

[...]

PR #7931, opened 2026-08-12
fix(metadata-protocol): key the /meta list merge on the i18n bundle discriminator (#7774)

### LINKED ISSUE #7774: GET /meta/email_template can collapse an i18n bundle back to one row — the list merge dedups by (package, name), no locale

huangyiirene:
Fixes #7774

## What

`GET /meta/email_template` served **one** locale of a declared i18n bundle. #7730 taught the `SchemaRegistry` that an `email_template`'s identity is `(name, locale)`, so `listItems('email_template')` returns every member; `getMetaItems` then merges that listing with two higher layers, and **both** merges keyed by `(package, name)` with no discriminator.

Both now key on the pair, and both are byte-identical for a type that declares no discriminator — which is every type except `email_template` today.

## The card's premise failed — `premise_still_valid: false`

The issue said the `sys_metadata` overlay merge "needs no change: overlay rows are unique on `type+name+org` and carry no locale." **The claim about the rows is true; the conclusion about the merge is not.**

| Probe | Result |
|:--|:--|
| `sys_metadata` overlay uniqueness | `idx_sys_metadata_overlay_active` = `(type, name, organization_id, package_id)`; **no locale column anywhere in the table** — the premise, confirmed |
| What `mergePackageAwareOverlay` buckets | `push(item.name, …)` over **`baseItems`** — the SchemaRegistry listing, which since #7730 is the whole bundle |
| Emission | one row per `(bucket, package)` ⇒ **a single overlay row for the type dropped a locale**, and the survivor was the overlay body regardless of which member it customizes |

The rows were never the problem; the *base* of that merge is. Reverting only the overlay half of this fix turns 3 of the 13 new cases red — see the table below.

## The four keying sites, and what changed at each

Anchored by function name (`protocol.ts` line anchors drift within hours — the issue's `:4363`/`:4377` were still accurate at dispatch time, but the file moved again mid-task).

| Site (in `getMetaItems`) | Before | After |
|:--|:--|:--|
| MetadataService merge — the `itemMap` loops | `metaItemKey(pkg, name)`; second member's `set` overwrote the first | `metaItemKey(pkg, name, itemDiscriminator(type, entry))` |
| `mergePackageAwareOverlay` — active overlays | bucket = `name` | bucket = slot = `name` (+ NUL + discriminator when declared) |
| `mergePackageAwareOverlay` — draft preview | bucket = `name` | same slot |
| env-wide ⊕ org row dedup | `metaItemKey(r.package_id, r.name)` | `+ storedRowDiscriminator(type, r)`, read off the row's serialized body |

The last one is the widening the failed premise forces. Within **one** org it changes nothing — the store's unique index cannot hold two rows differing only by body locale. Across the env-wide and org tiers it can, and keying them together made an org's zh-CN row silently displace an env-wide en-US one. Org-over-env precedence is unchanged where it was ever meaningful (a case pins that an org row still overrides the env-wide row of the *same* member). Reading the discriminator off the raw stored body rather than the converted one is safe here and checked: no ADR-0087 conversion entry touches `email_template`.

## Where the table lives — and what deliberately did not move

`ITEM_KEY_DISCRIMINATORS` moves from `@objectstack/objectql`'s `registry.ts` to **`@objectstack/metadata-core`**. `@objectstack/objectql` depends on `@objectstack/metadata-protocol`, so the protocol package cannot import the registry's copy without closing a cycle turbo rejects; metadata-core is the package both already depend on and depends on neither — the criterion that sank the engine write-verb dispatch predicates (#5619), the audit-field governance table (#4513) and the injected-system-column definitions (#6562) there. Measured: `metadata-core` depends on only `@objectstack/spec` + `zod`.

**No public surface changes.** `registry.ts` re-exports `ITEM_KEY_DISCRIMINATORS` under its original name from its original module, so `registry-i18n-bundle-key.test.ts`'s `import { SchemaRegistry, ITEM_KEY_DISCRIMINATORS } from './registry'` keeps working — its `Object.keys(...)).toEqual(['email_template'])` pin included. `metadata-core` gains `readDiscriminatorValue` / `itemDiscriminator` as additive exports.

**Not moved:** the registry's storage-key *format* — `BUNDLE_KEY_SEPARATOR`, `withDiscriminator`, `bundleBaseKey`, `collectBundle`. Those encode a discriminator into the registry's own composite `packageId:name` Map keys and parse it back; the protocol layer builds its own NUL-separated merge key and never parses a registry key, so it needs the *identity* question answered, not the encoding. Moving the encoding would have published a registry-internal key format as a cross-package contract. This is a deliberate half-move, and the TSDoc on both sides says so.

## Tests — the first coverage this path has ever had

`git grep metaItemKey` across `packages/` returned exactly one file before this PR: `protocol.ts` itself. New: `packages/metadata-protocol/src/protocol.i18n-bundle-list-merge.test.ts`, 13 cases.

Both merge blocks are **conditional** — the MetadataService block runs only with a `metadata` service installed *and* answering non-empty for the type; the overlay block only with at least one matching `sys_metadata` row. A harness that omits either passes against the bug, so every case installs the precondition it targets, and one case (`serves the bundle when NEITHER merge block runs`) exists precisely to show the shape of harness that made the defect invisible.

### Reverse verification

Committed the fix first (`dd64b71`), then re-introduced the undiscriminated key on top of it. Direction predicted before running: bundle cases red, undiscriminated guards green.

| Case | revert **both** | revert `metaItemKey` only | revert overlay slot only |
|:--|:--:|:--:|:--:|
| precondition: one discriminated type today | ✅ | ✅ | ✅ |
| precondition: neither merge block runs | ✅ | ✅ | ✅ |
| service merge: keeps both locales | ❌ | ❌ | ✅ |
| service merge: service contributes a new locale | ❌ | ❌ | ✅ |
| overlay: keeps every member with ONE overlay row | ❌ | ✅ | ❌ |
| overlay: sibling in another package survives | ✅ | ✅ | ✅ |
| overlay: env-wide + org rows of different locales | ❌ | ❌ | ❌ |
| overlay: org still overrides env for the SAME locale | ✅ | ✅ | ✅ |
| overlay: no-locale member is the canonical member | ✅ | ✅ | ✅ |
| draft preview: sibling survives | ❌ | ✅ | ❌ |
| ADR-0048: two packages ship `page/home` | ✅ | ✅ | ✅ |
| ADR-0048: overlay collapses onto its own package slot | ✅ | ✅ | ✅ |
| undiscriminated: same-name page is ONE row | ✅ | ✅ | ✅ |
| | **5 red / 8 green** | **3 red / 10 green** | **3 red / 10 green** |

Sample red, `revert both`:

```
FAIL  the MetadataService merge (the path the card names) > keeps both locales
AssertionError: expected [ 'auth.welcome@zh-CN', …(1) ] to deeply equal [ 'auth.welcome@en-US', …(2) ]
-   "auth.welcome@en-US",
    "auth.welcome@zh-CN",
    "billing.invoice@en-US",
```

Three cases stay green under the full revert, and that is honest rather than a hole: `sibling in another package` is separated by ADR-0048's *package* dimension, which the revert does not touch; `org overrides env for the SAME locale` and `no-locale member is canonical` assert that the fix did **not** split things it shouldn't. Every undiscriminated guard staying green under the revert is the evidence that non-bundled keys are byte-identical — a guard that went red would mean this change altered a key it promised not to.

## Gates

Run twice. First on `dd64b71`+`21fc024`; then `origin/main` moved 11 commits — including `packages/metadata-protocol/src/protocol.ts` (#7906's `$icontains` fix) and a 345-line rewrite of `scripts/check-type-check-coverage.mjs` (#7921) — so the **full** set was re-run on the merge commit `3319a62` rather than a scoped subset. Numbers below are the post-merge run.

| Gate | Result |
|:--|:--|
| `@objectstack/metadata-core` test | 9 files / 129 tests ✅ |
| `@objectstack/metadata-protocol` test | 74 files / 1082 tests ✅ |
| `@objectstack/objectql` test | 187 files / 3319 tests ✅ — incl. `registry-i18n-bundle-key.test.ts` and its `ITEM_KEY_DISCRIMINATORS` by-name pin |
| `@objectstack/runtime` test | 139 files / 2127 tests ✅ |
| `@objectstack/cli` test | 111 files / 1191 tests ✅ |
| `typecheck` — metadata-core, objectql, runtime, cli | ✅ clean (metadata-protocol has no `typecheck` script; it is a measured DEBT entry, covered by the ratchet below) |
| `check:type-check-coverage` | ✅ |
| `check:type-check-debt` (`--re-measure`) | ✅ 36 entries, 1987 raw errors, none above its recorded number. The pre-merge run first went red at metadata-protocol 73 vs frozen 63, caused by this PR's own test-helper typing (`Partial<Row> & { metadata: unknown }` collapsing to `string`); repaired in `21fc024` rather than raising the ledger |
| `eslint` on changed files | ✅ clean |
| `check:i18n` | ✅ 9 packages in sync |
| `check:nul-bytes` | ✅ 7251 files |
| `pnpm --filter @objectstack/spec check:generated` | ✅ 13/13 up to date |
| `pnpm build` | ✅ 71/71 |

## Out of scope, recorded not fixed

`duplicatePackage`'s source-row scan dedupes its env-wide ⊕ org rows by `(type, name)` with the same locale-blind key. Different feature, different endpoint, outside this PR's declared surface — filed unassigned as #7932 rather than bundled here.

#7491 (stale `AUTHORING_RULES` comment counts) is **unaffected in substance and moved in position**: its anchor was already at `protocol.ts:2698` on `origin/main` (not the `:2632` the card records), and this PR pushes it further down. The comment still reads "26" against 38 rules actually declared in `packages/lint/src/authoring-rules.ts`. Not touched here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XxLeX1UvWRzjJ1UUpd7dGt)_

### DIFF
--- a/.changeset/meta-list-i18n-bundle-key.md
+++ b/.changeset/meta-list-i18n-bundle-key.md
@@ -1,0 +1,53 @@
+---
+"@objectstack/metadata-protocol": patch
+"@objectstack/metadata-core": patch
+"@objectstack/objectql": patch
+---
+
+fix(metadata-protocol): keep every i18n bundle member through the `/meta` list merge (#7774)
+
+#7730 taught the `SchemaRegistry` that an `email_template`'s identity is
+`(name, locale)`, so `listItems('email_template')` returns every member of a
+declared i18n bundle. `GET /meta/<type>` then merges that listing with two
+higher layers, and both merges keyed by `(package, name)` with no
+discriminator — so the bundle survived registration only to collapse one layer
+later, and the list served a single locale.
+
+**Both merges now key on the pair.** `metaItemKey` takes an optional third
+component and `mergePackageAwareOverlay` buckets per slot rather than per name;
+both derive the value from the shared discriminator table, and both are
+byte-identical for a type that declares no discriminator — which is every type
+except `email_template` today.
+
+- **The MetadataService merge** is the path the issue named: with a `metadata`
+  service installed and answering non-empty for the type, the second member's
+  `Map.set` overwrote the first.
+- **The `sys_metadata` overlay merge** was predicted to need no change, on the
+  ground that overlay rows are unique on `type+name+organization_id+package_id`
+  and carry no locale column. That is true of the rows and beside the point:
+  the base of that merge is the registry's bundle, so bucketing by bare name
+  dropped a locale as soon as a single overlay row existed for the type — and
+  the row that survived was the overlay body, whichever member it customizes.
+  An overlay (or a draft preview) now lands on its own locale member and the
+  rest of the bundle is served untouched. Across the env-wide and org tiers,
+  rows that customize different members are likewise two slots instead of one;
+  org-over-env precedence is unchanged within a member.
+
+**The discriminator table moved to `@objectstack/metadata-core`.**
+`ITEM_KEY_DISCRIMINATORS` was declared in `@objectstack/objectql`'s
+`registry.ts`, and `@objectstack/objectql` depends on
+`@objectstack/metadata-protocol`, so the protocol package could not import it
+without closing a dependency cycle. metadata-core is the package both already
+depend on and depends on neither — the same criterion that sank the engine
+write-verb dispatch predicates (#5619) and the audit-field governance table
+(#4513) there. **No public surface changes:** `registry.ts` re-exports
+`ITEM_KEY_DISCRIMINATORS` under its original name from its original module, so
+every existing import keeps working; `@objectstack/metadata-core` gains it plus
+`readDiscriminatorValue` / `itemDiscriminator` as additive exports. The
+registry's storage-key *format* (`name@<locale>` composite keys and their
+parser) deliberately did not move — it encodes the registry's own Map keys,
+which no other package reads.
+
+For an app this is Studio's metadata list and `GET /meta/email_template`
+showing both the en-US and the zh-CN copy of a template instead of whichever
+one the merge happened to keep.
--- a/packages/metadata-core/src/index.ts
+++ b/packages/metadata-core/src/index.ts
@@ -53,3 +53,13 @@ export * from './injected-system-columns.js';
 // outlet, or the mask is decoration") is only true if they all run the same
 // projection rather than a copy each.
 export * from './object-schema-fls.js';
+
+// [#7730 / #7774] The i18n-bundle DISCRIMINATOR table — which metadata types
+// are identified by `(name, <field>)` rather than by `name` alone — sunk here
+// by the same criterion as the governance table above. `@objectstack/objectql`
+// (the SchemaRegistry, #7730) and `@objectstack/metadata-protocol` (the
+// unscoped `/meta` list merge, #7774) both key metadata by name, objectql
+// depends on metadata-protocol, and a bundle that survives one layer's key but
+// not the other's is still collapsed. `objectql` re-exports
+// `ITEM_KEY_DISCRIMINATORS` from `registry.ts`, so its surface is unchanged.
+export * from './item-key-discriminators.js';
--- a/packages/metadata-core/src/item-key-discriminators.ts
+++ b/packages/metadata-core/src/item-key-discriminators.ts
@@ -1,0 +1,92 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#7730 / #7774] i18n bundles — metadata types whose IDENTITY is a pair.
+ *
+ * Most metadata types are identified by `name` alone. `email_template`
+ * declares otherwise: `EmailTemplateDefinitionSchema` states that "multiple
+ * rows with the same `name` but different `locale` form an i18n bundle; the
+ * service picks the best match for the recipient's locale, falling back to
+ * `en-US`" (`packages/spec/src/system/email-template.zod.ts`), and its header
+ * says a template "is resolved by `(name, locale)`".
+ *
+ * Every layer that keys metadata by `name` therefore has to agree on ONE
+ * answer to "what else is part of this type's identity?", or a bundle survives
+ * one layer and collapses at the next:
+ *
+ *  - `@objectstack/objectql`'s `SchemaRegistry` keys its item collections
+ *    (`registerItem` / `getItem` / `listItems`) — fixed by #7730;
+ *  - `@objectstack/metadata-protocol`'s unscoped `/meta/<type>` list merge
+ *    keys the sys_metadata overlay and MetadataService merges (`metaItemKey`,
+ *    `mergePackageAwareOverlay`) — fixed by #7774, which is why the table
+ *    lives HERE rather than in the registry that first needed it.
+ *
+ * `@objectstack/objectql` depends on `@objectstack/metadata-protocol`, so the
+ * protocol package cannot import the table from the registry — the reverse
+ * import would close a cycle turbo rejects. This package is the one both sides
+ * already depend on and it depends on neither, the same criterion that sank
+ * the engine write-verb dispatch predicates (#5619), the audit-field
+ * governance table (#4513) and the injected-system-column definitions (#6562)
+ * here. `objectql` re-exports `ITEM_KEY_DISCRIMINATORS` from `registry.ts`, so
+ * its public surface is unchanged.
+ *
+ * What deliberately did NOT move: the registry's storage-key FORMAT
+ * (`BUNDLE_KEY_SEPARATOR`, `withDiscriminator`, `bundleBaseKey`,
+ * `collectBundle`). Those encode a discriminator into the registry's own
+ * `<packageId>:<name>` Map keys and parse it back out again; the protocol
+ * layer builds its own NUL-separated merge key and never parses a registry
+ * key, so it needs the IDENTITY question answered — the table and
+ * {@link itemDiscriminator} — not the encoding. Moving the encoding too would
+ * have published a registry-internal key format as a cross-package contract.
+ */
+
+/**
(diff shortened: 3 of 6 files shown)


huangyiirene:
## ACCEPT — flipped ready, auto-merge SQUASH armed @ `3319a62`

All 25 jobs concluded before arming: **23 success, 2 skipped (`Build Docs`, `Console Pin Gate` — neither applies), zero failures**, `TypeScript Type Check` green at 05:37:38Z. I read each job's own conclusion rather than an aggregate, which is not pedantry this shift — the sibling PR #7920 sat at "0 failed" through three checkpoints and the last job to conclude went red.

### The falsified premise is the most valuable thing in this PR

The card asserted the `sys_metadata` overlay merge "needs no change: overlay rows are unique on `type+name+org` and carry no locale." You returned `premise_still_valid: false` with the distinction that matters: **the claim about the rows is true and the conclusion about the merge does not follow.** `mergePackageAwareOverlay` buckets over `baseItems` — the registry listing, which since #7730 *is* the whole bundle — so the rows' uniqueness was never what governed it. A single overlay row dropped a locale.

That was mine to get wrong, and I want it on the record as such: I re-measured this card's four preconditions at source before dispatch and still carried the card's overlay sentence forward without testing it, because it reasoned about the rows and the rows were right. The three-column revert table is what makes the correction checkable rather than assertable — `revert overlay slot only` turns 3 cases red that `revert metaItemKey only` leaves green, so the two halves are proven independently load-bearing rather than jointly plausible.

The env-wide ⊕ org widening that falls out of it is correctly scoped and correctly argued: within one org the store's unique index makes it a no-op, across tiers it is real, and you pinned org-over-env precedence for the *same* member so the fix cannot be read as having loosened it.

### Structural claim verified independently

I checked the seam rather than taking the report's word:

- **No `package.json` in the diff at all** — six files, none of them a manifest. The "metadata-core is the package both already depend on" claim is not just stated, it is proven by the absence of a new dependency edge.
- **`registry.ts` re-exports `ITEM_KEY_DISCRIMINATORS` under its original name from its original module**, and the local `discriminatorValue` helper is replaced by `readDiscriminatorValue` *aliased to the old name* — so not one call site moved and `registry-i18n-bundle-key.test.ts`'s `Object.keys(...).toEqual(['email_template'])` pin still resolves. objectql's 3315 tests passing is the confirmation, not the argument.
- **The half-move is the right half.** Keeping `BUNDLE_KEY_SEPARATOR` / `withDiscriminator` / `bundleBaseKey` / `collectBundle` in the registry keeps a registry-internal `@`-encoded Map key from being published as a cross-package contract, while the *identity* question — the only thing the protocol layer needs — is what got shared. Publishing the encoding would have been the easy over-move and it would have been wrong.

### You caught the ratchet yourselves

`check:type-check-debt` went red on your first run at metadata-protocol 73 vs frozen 63, from this PR's own test-helper typing, and you **repaired the typing rather than raising the ledger**. That is the same gate that took #7920 to a patch round an hour ago, reached independently and without being told — the dispatch brief named it, but the reasoning for *why* raising is refused is yours. #5278's argument exactly.

### Recorded

- **#7932** filed for `duplicatePackage`'s locale-blind source-row scan, unassigned, with reachability measured to the tier boundary. Filed rather than ridden — correct call; a different endpoint in the same file is still a different endpoint. It is now in this lane's queue.
- **#7491 repricing collected**: its anchor was already at `protocol.ts:2698` on `origin/main`, not the `:2632` the card records, and this PR moves it further. The substance is unchanged — still "26" against 38 declared rules. I am updating that card's note so the next dispatch does not chase a dead line number; the anchor is the comment text, never the line.

Nothing else is owed on this one.

---

Issue #7955, opened 2026-08-12
finding: the two `listRemoteTables` route twins diverge on `?schema=` — the admin spelling drops it


hotlong:
## Symptom

`IExternalDatasourceService.listRemoteTables` is reachable through two live routes:

| route | package | call |
|:---|:---|:---|
| `GET /api/v1/datasources/:name/external/tables` | `packages/rest/src/external-datasource-routes.ts` | `listRemoteTables(name, { schema })` |
| `GET /api/v1/datasources/:name/remote-tables` | `packages/services/service-datasource/src/admin-routes.ts` | `listRemoteTables(name)` |

The federation spelling forwards `?schema=`; the admin spelling **does not read the query at all**, so a caller that passes `?schema=public` to it silently gets the unfiltered result rather than an error or the filtered set.

The same twin relationship holds for `generateObjectDraft` (`POST /:name/object-draft` vs `POST /:name/external/tables/:remote/draft`), which does forward its options bag — so this divergence is specific to the `schema` filter on the listing route.

## Why this is a finding and not part of #7744

The twins are known and deliberate: #4249 reconciled their **failure contract** ("One operation, one failure contract now, on both paths", `external-datasource-routes.ts`) rather than removing either. #7744 ledgered the admin spelling at its live path and was explicitly scoped away from renaming or removing a live route. What #4249 reconciled was the error path; the *request* path was never compared, and this is the residue.

## Options

1. Forward `req.query.schema` on the admin route so the twins accept the same request shape (smallest change; makes the paths interchangeable, which is what "one operation" implies).
2. Leave it, and record the divergence as intended — in which case the admin route arguably ought to refuse an unsupported query parameter rather than ignore it, since silently dropping a filter is the "declared ≠ enforced" shape (Prime Directive #10).

Not fixed in #7744 because either option changes request-handling behaviour on a live route, which that card ruled out of scope.

## Reproduction

`GET /api/v1/datasources/<name>/remote-tables?schema=public` and `GET /api/v1/datasources/<name>/external/tables?schema=public` against the same datasource: the second is filtered, the first is not.

hotlong:
Triage: `pm:queue`, routed `domain:services` (the divergent side is `packages/services/service-datasource/src/admin-routes.ts`), type **Bug**.

[...]

Issue #7954, opened 2026-08-12
finding: the datasource-admin lifecycle family has no SDK expression — decide `server-only` vs `gap`


hotlong:
## Symptom

Surfaced while ledgering the datasource-admin route family for #7744.

`ObjectStackClient` reaches **none** of the ten routes `registerDatasourceAdminRoutes` mounts under `/api/v1/datasources`. Its `datasources` namespace contains exactly one sub-namespace, `external`, whose five methods all target the **federation** family in `@objectstack/rest` (`/:name/external/…`, ledgered in `packages/rest/src/rest-route-ledger.ts`). The CLI is the same story: its three `datasource` commands (`introspect`, `list-tables`, `validate`) all call `/external/*`.

So the whole runtime datasource lifecycle — list, read, create, patch, remove, connection probe, driver catalog — is reachable only by the Setup/Studio console and by one declared metadata-type action (`test_connection`).

## The decision this needs

The new `packages/services/service-datasource/src/datasource-route-ledger.ts` records all ten rows as **`server-only`**, on the grounds that:

- the family is mounted by `objectstack serve` (a CLI-composed plugin registering on `http.server`), **not** by `@objectstack/rest`, so it is not part of the platform's advertised SDK composition; and
- its actual consumers are the console and a metadata-type action.

That classification is defensible but it is the *conservative* reading, chosen deliberately: writing ten `gap` rows instead would have been a product decision — "the SDK should be able to manage datasources" — made inside a bookkeeping fix. The ledger vocabulary forces one of the two, and this issue is where the choice should actually be reviewed.

If the answer is "yes, the SDK should express datasource lifecycle", the rows become `gap` (and the ledger's gap ratchet, currently pinned at 0, moves with them) until the client methods land. If the answer is "no", the current `server-only` rows are already correct and this can close as documented-by-the-ledger.

## Source

Extracted from #7744 (route-ledger parity for the datasource-admin family).

hotlong:
Triage disposition: **`server-only` stands — closing as documented-by-the-ledger** (not planned), type **Task**, routed `domain:cli` for the record (the counterfactual work would have landed in `packages/client`). This is the findings-round close power with the auto-adjudication analysis attached; **veto window: listed in this round's 代裁清单 — reopening is one click and no argument.**

**Four-lens block:**
[...]

Issue #4249, opened 2026-07-31
The same three datasource-admin routes answer 400 with the wrong service's error code (#4225 follow-up, one field over)


os-zhuang:
#4225 / #4234 fixed the 503 `message` on the three routes in `packages/services/service-datasource/src/admin-routes.ts` that resolve `external-datasource` rather than `datasource-admin`. The identical mis-attribution survives one field over, on the 400 path — and this one is machine-readable.

## The finding

`admin-routes.ts` has one shared 400 helper:

```ts
const badRequest = (res: any, err: unknown) =>
  sendError(res, 400, 'DATASOURCE_ADMIN_ERROR', err instanceof Error ? err.message : String(err));
```

All three `external-datasource` routes route their failures through it:

| Route | resolves | 400 `error.code` |
|---|---|---|
| `GET /:name/remote-tables` | `external-datasource` | `DATASOURCE_ADMIN_ERROR` |
| `POST /:name/test` | `external-datasource` | `DATASOURCE_ADMIN_ERROR` |
| `POST /:name/object-draft` | `external-datasource` | `DATASOURCE_ADMIN_ERROR` |

The ledger says what that code means, in as many words:

```ts
// packages/spec/src/api/error-code-ledger.zod.ts
'@objectstack/service-datasource': [
  'DATASOURCE_ADMIN_ERROR',      // lifecycle/validation refusal from the datasource-admin service
],
```

So a `no such schema` raised by the **external-datasource** introspector is reported as a lifecycle/validation refusal from **datasource-admin**. Same class as #4225, but where that one misled a human reading prose, this misroutes a client switching on `error.code`.

## Why #4225 deliberately did not fix it

#4225's argument was that `SERVICE_UNAVAILABLE` is correct for all nine routes — ADR-0112's ledger asks generic conditions to reuse the standard catalog rather than register a per-service synonym — so only the `message` was wrong, and `message` is free text. `DATASOURCE_ADMIN_ERROR` is neither generic nor free: `ApiErrorSchema.code` is the closed `ErrorCode` union, so correcting it means editing the ledger. That is a spec change, and #4234 kept `error.code` byte-identical on all nine routes on purpose.

## What the fix has to decide

The sibling surface does not settle it either. `packages/rest/src/external-datasource-routes.ts` catches on the **import** route only:

```ts
sendError(res, 400, 'EXTERNAL_IMPORT_ERROR', ...)   // import only
```

[...]

Issue #3636, opened 2026-07-27
Route audit tranche 3: autonomously-mounted service routes (service-storage / service-i18n) — #3587 follow-up


os-zhuang:
Split from #3587, same as #3587 was split from #3563. Tranches 1–2 are complete: the dispatcher ledger (27→0, #3569…#3579) and the REST ledger (mismatch 2→0 gap 43→0, #3609…#3633) both rest at zero with CI-pinned ratchets. One surface remains un-audited — the ledger header (`packages/rest/src/rest-route-ledger.ts`) records it as the known NOT-COVERED note:

> services that autonomously mount routes on the host `IHttpServer` — `service-storage` (`storage-routes.ts`, the SDK's whole storage surface) and `service-i18n`.

## The inventory (13 routes)

**`packages/services/service-storage/src/storage-routes.ts`** — 10 mounts:

| Route | Purpose |
|---|---|
| `POST {base}/upload/presigned` | presigned-URL issue (the SDK `storage.upload` path) |
| `POST {base}/upload/complete` | post-upload finalize |
| `POST {base}/upload/chunked` | chunked-upload session create |
| `PUT {base}/upload/chunked/:uploadId/chunk/:chunkIndex` | chunk write |
| `POST {base}/upload/chunked/:uploadId/complete` | chunked finalize |
| `GET {base}/upload/chunked/:uploadId/progress` | chunked progress |
| `GET {base}/files/:fileId/url` | download-URL resolve (the SDK `storage.download` path) |
| `GET {base}/files/:fileId` | file metadata |
| `PUT {base}/_local/raw/:token` | local-driver loopback write (likely `server-only`) |
| `GET {base}/_local/raw/:token` | local-driver loopback read (likely `server-only`) |

**`packages/services/service-i18n/src/i18n-service-plugin.ts`** — 3 mounts (behind the `registerRoutes` option):

| Route | Purpose |
|---|---|
| `GET {base}/locales` | available locales |
| `GET {base}/translations/:locale` | translation bundle |
| `GET {base}/labels/:object/:locale` | object field labels |

Client side today: `storage` (2 methods) and `i18n` (3 methods) namespaces exist and are exercised in production — but **no ledger row anywhere backs them**, so they are exactly the pre-#3563 posture: expressed, working, unguarded.

## Method — tranche 1/2 recipe, one new seam

[...]

PR #7957, opened 2026-08-12
fix(service-datasource): ledger the mounted datasource-admin routes (#7744)

### LINKED ISSUE #7744: finding: datasource-admin route-ledger gap (D6) + an unknown-driver draft is accepted 201 with status:'error'
## Symptom

Two observations from `datasource-admin-lifecycle` (the item otherwise passes — secret handling is exemplary, and refusals attribute to the right service so the #4249 mis-attribution is absent):

(a) A draft naming an **unknown driver** is accepted **201** with `status:'error'` rather than refused — health-reporting rather than validation.

(b) **D6 confirmed** — the admin CRUD routes actually mounted carry **no route-ledger entry**, and the ledgered spelling differs from the live one.

## Root cause

(b) The live admin CRUD routes are mounted at `/api/v1/datasources/:name/remote-tables` (`packages/services/service-datasource/src/admin-routes.ts`, ~line 200), but the route ledger (`packages/rest/src/rest-route-ledger.ts`, ~lines 321-323) lists only the **federation** family under the different spelling `/api/v1/datasources/:name/external/tables` — so the mounted admin routes have no ledger entry and the ledgered path does not match the live one. Same tranche-3 route-ledger parity family as #7526 (and the datasource-admin error-attribution family #4225 / #4249).

(a) The datasource-admin draft-create path accepts an unknown-driver draft and records `status:'error'` on the row instead of validating the driver name at create time (`packages/services/service-datasource`).

Fix per site: the ledger entry belongs in `packages/rest`; the draft driver-name validation in `packages/services/service-datasource`.

Stale-premise check: present on `origin/main` (ledger still spells `/external/tables`; live mount still `/remote-tables`).

## Reproduction

(b) Diff the mounted datasource-admin routes against `packages/rest/src/rest-route-ledger.ts`: the five datasource ledger entries are all federation (`/external/tables`) ones; the live `/remote-tables` CRUD routes appear nowhere.

(a) `POST` a datasource draft naming a driver that is not registered.

Expected: (b) every mounted route has a ledger entry at its live spelling; (a) the unknown-driver draft is refused. Actual: (b) no ledger entry, spelling mismatch; (a) 201 with `status:'error'`.

## Source

hotlong:
Fixes #7744

The card had two halves, dispatched on different terms. **(b) landed. (a) is falsified and nothing was changed for it** — the evidence is below.

---

### What was actually wrong

The ten admin CRUD routes under `/api/v1/datasources` had no route-ledger entry in **any** of the platform's ledgers, and no guard was comparing them to anything. They are mounted the "third way" that `service-storage` and `service-i18n` grew their own ledgers for (#3636): `objectstack serve` builds a small `com.objectstack.cli.datasource-admin-routes` plugin whose `init()` resolves `http.server` and calls `registerDatasourceAdminRoutes` straight on `IHttpServer` (`packages/cli/src/commands/serve.ts`). Neither `RouteManager` nor `RestServer.getRoutes()` has ever seen them.

### Why the entry is NOT in `packages/rest/src/rest-route-ledger.ts`

The card and the dispatch both name `packages/rest` as the site. That file structurally cannot host these rows, and the guard next to it says so in both directions:

- `rest-route-ledger.conformance.test.ts` enumerates **one** thing — a `RestServer` booted the way production boots it, its own `registerRoutes()` plus `mountAndRecordDirectRoutes` (#5822). `registerDatasourceAdminRoutes` is not one of the registrars that function composes (`direct-mount-composition.ts` calls `registerPackageRoutes` and `registerExternalDatasourceRoutes`, and nothing else).
- So a `direct-mount` row added there would fail *two* assertions immediately: "every direct-mount ledger entry is really registered by its registrar", and the `directMounted.length === ledgerKeys('direct-mount').size` identity.

The binding requirement from the dispatch — **a ledger entry at the live spelling, pinned by a mount-derived test** — is met; only the file it lives in differs, and it differs to the shape the repo already uses for exactly this class of route. New: `packages/services/service-datasource/src/datasource-route-ledger.ts` + its conformance guard, mirroring `storage-route-ledger.ts` / `i18n-route-ledger.ts`.

### The spelling, and what was NOT done

⛔ No live route is renamed. The five `datasources` rows the REST ledger carries are the **federation** family (`/:name/external/…`) — a different, separately mounted family in a different package, not the admin family misspelled. Both are live. Two pairs do overlap (`remote-tables` ↔ `external/tables` over `listRemoteTables`; `object-draft` ↔ `external/tables/:remote/draft` over `generateObjectDraft`), and that overlap is known and was deliberately *reconciled* rather than removed — #4249 gave the two paths one failure contract ("One operation, one failure contract now, on both paths", `external-datasource-routes.ts`). Every row here carries the spelling the mount uses.

### The pin is derived from the mount, not transcribed

`datasource-route-ledger.conformance.test.ts` runs the registrar against a capturing `IHttpServer` and treats its registration calls as the route set. No path literal in the test is compared against the mount, so a future route addition fails the test instead of quietly re-creating this card. It also asserts **non-vacuity** (`live.size > 0` and `live.size === ledger.size`), because both set-difference checks would pass on an empty enumeration — "absence must be loud".

**Reverse-verified.** Deleting the new `GET /api/v1/datasources/:name/remote-tables` row (fix committed first, restored with `git checkout BRANCH -- PATH` — never `git stash`) fails 3 assertions for the right reasons:

```
AssertionError: Datasource-admin routes with no datasource-route-ledger entry:
  GET /api/v1/datasources/:name/remote-tables.
AssertionError: expected 10 to be 9              ← the non-vacuity size identity
AssertionError: GET /api/v1/datasources/:name/remote-tables is mounted but
  unledgered — the #7744 gap, reopened.
```

Restored → 7/7 green.

### Dispositions

All ten rows are `server-only`, each with a note naming its real consumer. Audited: no `ObjectStackClient` method and no CLI command reaches any of them (the client's `datasources` namespace holds only `external`, which targets the federation family; the three CLI `datasource` commands all call `/external/*`). Writing ten `gap` rows instead would have been a product decision made inside a bookkeeping fix — filed as #7954 for review instead. The client half joins the existing tranche-3 guard in `service-route-ledger-coverage.test.ts`.

---

## (a) Unknown-driver draft → 201 with `status:'error'` — **NOT implemented; premise falsified**

All three conditions were required. **All three fail.** Per the dispatch this is the success outcome for (a): it converts the card's second half from a bug into a documented design decision.

### Condition 1 — "no late binding" — **FAILS.** This one is dispositive.

Late binding is not merely possible here; it is the written contract of the seam, and the create path's behaviour under it is specified.

`packages/services/service-datasource/src/contracts/datasource-driver-factory.ts:7-18`:

> The framework deliberately ships no universal "driver-by-id" registry — concrete drivers … are constructed by the host stack …
> When no factory is registered, **or none `supports()` a given driver id**, the admin service degrades gracefully: `testConnection` returns `{ ok: false, error }` and **create/update skip hot pool registration (the driver is picked up on the next boot instead)**.

That last clause *is* the accept-then-report design, stated at the contract. Three more reads carry it:

- `datasource-connection-service.ts:137` — the factory field is a thunk, documented **"Resolve the host driver factory (lazy — may be registered after init)."**
- `datasource-admin-plugin.ts:230-239` — the factory is resolved *per call*, with the reason given inline: `init()` may run before other plugins register their services, and admin requests arrive long after boot.
- `datasource-admin-plugin.ts:381-391, 428-455` — `start()` runs `restoreRuntimeDatasources()` then `rehydratePools()`, so a persisted record's driver is re-resolved **on every boot**. A driver contributed by a plugin (or an optional peer package such as `@objectstack/driver-turso`) installed after the draft exists binds then. Refusing at create time would make that authoring state unreachable.

And the create path already states the boundary deliberately — `datasource-admin-service.ts:360-361`: *"A driver the platform ships no contract for passes untouched, matching the spec gate's boundary rather than inventing a stricter one for the UI."*

### Condition 2 — "nothing pins the 201" — **FAILS.**

- `service-datasource/src/__tests__/admin-routes.test.ts:139-151` — `it('POST /api/v1/datasources creates a runtime datasource (201)')`, `expect(res.status).toBe(201)`.
- `service-datasource/src/__tests__/envelope.conformance.test.ts:113-116` — `POST /datasources (201)` as a declared envelope-conformance case.

Stated precisely: both drive `driver: 'postgres'`, a known id, so an unknown-driver-only refusal would not by itself turn them red. What they pin is the create route's accept-and-return-the-summary **shape**, which is the shape the refusal would fork.

*Control for this grep:* the same query shape returns hits where a status code really is pinned — `packages/types/src/response-envelope.test.ts:60`, `packages/rest/src/public-form-routes.test.ts:124`, `rest-dropped-fields.test.ts:107` and others — so the hits above are a real reading, not a lucky pattern.

### Condition 3 — "`status:'error'` is not a consumed design state" — **FAILS**, and it reads the unknown-driver case *by name*.

The unknown-driver path is a named state with its own vocabulary entry, mapped through a declared class into a declared contract field, carrying a machine-readable reason:

`factory.supports(driver) === false` → `ConnectStatus 'skipped-unsupported'` (`datasource-connection-service.ts:154, 532-537`, whose reason string is `no driver factory supports driver 'NAME'`) → `availabilityOf` → `'failed'` (`:202-205`) → `summaryStatus` → `status: 'error'` + `statusReason` (`datasource-admin-service.ts:109-125, 401-419`).

The vocabulary doc spells out that this is intended — `datasource-connection-service.ts:174-176`:

> `failed` — a connect was attempted and did not produce a usable driver (unreachable, bad credential, **unsupported driver**).

Consumed and tested: `contracts/datasource-admin-service.ts:92` declares `status: 'ok' | 'error' | 'blocked' | 'unvalidated'` on `DatasourceSummary`; `__tests__/datasource-admin-service.test.ts:412-425` pins each availability class onto a distinguishable status (`expect(byName.dead!.status).toBe('error')`); `packages/runtime/src/datasource-autoconnect.test.ts:463` pins the sibling `'unvalidated'`. The whole field exists because of #3827 — before it, "a datasource that died at boot looked exactly like one nobody had tested."

*Control for this grep:* querying the same vocabulary for the sibling value `unvalidated` returns hits across `packages/runtime`, `packages/objectql` and `packages/spec`, so a zero-hit result for `'error'` would have meant absence rather than a broken search — and it was not zero.

### Consequence

No code changed for (a), and per the dispatch no consolation change was made — the error message was not "improved". A `status:'error'` carrying `statusReason: no driver factory supports driver 'NAME'` is health-reporting working as designed, on a create path that is documented to accept a draft whose driver is not buildable in this process.

---

## Verification

- `pnpm --filter @objectstack/service-datasource typecheck` — clean.
- `pnpm --filter @objectstack/service-datasource test` — 13 files / 333 tests pass (7 of them new).
- Client ledger guards — `service-route-ledger-coverage`, `rest-route-ledger-coverage`, `route-ledger-response-schema` — 11 pass.
- `check:type-check-debt` does **not** rise. `packages/client`'s `check:test-typecheck` reports 3 problems locally, all `TS2307: Cannot find module '@objectstack/runtime'` in files this PR never touches; reverting only this PR's client edit reproduces the identical 3, so it is the unbuilt-local-dependency artifact of AGENTS.md §9, not a ratchet movement. Neither touched package carries a DEBT/TEST_DEBT entry, and the new test file is inside `tsc --noEmit`'s program (no `*.test.ts` exclusion) with zero errors. Confirmed on CI: the required `TypeScript Type Check` job is green.
- Changeset: `.changeset/eighty-hoops-repeat.md` (patch). Not breaking, so no ADR-0087 marker is owed.

## Out-of-scope findings filed

- #7954 — the family has no SDK expression at all; `server-only` vs `gap` is a product call.
- #7955 — the two `listRemoteTables` twins diverge on `?schema=`: only the federation spelling forwards it, so the admin spelling silently drops the filter. Changing a live route's request handling was out of scope here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NzFcriQJsJaaiieWSsrMu7)_

### DIFF
--- a/.changeset/eighty-hoops-repeat.md
+++ b/.changeset/eighty-hoops-repeat.md
@@ -1,0 +1,23 @@
+---
+'@objectstack/service-datasource': patch
+---
+
+Ledger the mounted datasource-admin routes, at the spelling they are mounted under.
+
+The ten admin CRUD routes under `/api/v1/datasources` — list, read, create, patch,
+remove, connection probe, driver catalog and schema introspection — carried no
+route-ledger entry in any of the three ledgers the platform keeps. They are mounted
+the "third way" `service-storage` and `service-i18n` grew their own ledgers for:
+`objectstack serve` builds a small plugin that resolves the `http.server` service and
+registers straight on `IHttpServer`, so neither `RouteManager` nor
+`RestServer.getRoutes()` ever sees them.
+
+The five `datasources` rows the REST ledger does carry are the **federation** family
+(`/api/v1/datasources/:name/external/…`), which is a different, separately mounted
+family in `@objectstack/rest` — not the admin family misspelled. Both are live, and
+no route is renamed here: the new ledger is written at the live admin spelling, and
+its conformance guard derives what it expects from the registrar rather than from a
+literal in the test, so an eleventh route fails the guard instead of silently
+reopening the gap.
+
+No runtime behaviour changes — this adds a package-internal ledger and its guard.
--- a/packages/client/src/service-route-ledger-coverage.test.ts
+++ b/packages/client/src/service-route-ledger-coverage.test.ts
@@ -7,28 +7,31 @@
  *
  *   - `packages/services/service-storage/src/storage-route-ledger.conformance.test.ts`
  *   - `packages/services/service-i18n/src/i18n-route-ledger.conformance.test.ts`
+ *   - `packages/services/service-datasource/src/datasource-route-ledger.conformance.test.ts`
  *
  * Same contract as the two ledger guards next door
  * (`route-ledger-coverage.test.ts`, `rest-route-ledger-coverage.test.ts`):
  * every ledger entry that names a client method must resolve to a real
  * function on an instantiated client.
  *
- * Both ledgers are imported as relative SOURCE files deliberately: they are
- * pure data (no imports), and a client→service package edge for them would be
- * backwards — the services are where the routes are declared, so the ledgers
+ * All three ledgers are imported as relative SOURCE files deliberately: they
+ * are pure data (no imports), and a client→service package edge for them would
+ * be backwards — the services are where the routes are declared, so the ledgers
  * live there, and each package verifies its own half. That is the tranche-1
  * lesson (`no runtime→client package edge`) applied verbatim: CI's per-package
  * test tasks build only their own dependency closure, so a real package edge
  * here would be unbuildable.
  *
- * With these two, all THREE server surfaces the SDK reaches are ledgered —
- * dispatcher (#3563), REST (#3587), autonomous service mounts (#3636).
+ * With these, all THREE server surfaces the SDK reaches are ledgered —
+ * dispatcher (#3563), REST (#3587), autonomous service mounts (#3636, joined by
+ * the datasource-admin family in #7744).
  */
 
 import { describe, it, expect } from 'vitest';
 import { ObjectStackClient } from './index';
 import { STORAGE_ROUTE_LEDGER } from '../../services/service-storage/src/storage-route-ledger';
 import { I18N_ROUTE_LEDGER } from '../../services/service-i18n/src/i18n-route-ledger';
+import { DATASOURCE_ROUTE_LEDGER } from '../../services/service-datasource/src/datasource-route-ledger';
 
 describe('service route ledgers ↔ @objectstack/client surface', () => {
   const client = new ObjectStackClient({ baseUrl: 'http://localhost:9' });
@@ -56,6 +59,31 @@ describe('service route ledgers ↔ @objectstack/client surface', () => {
       broken,
       `i18n ledger entries claiming a client method that does not exist: ${broken.join('; ')}`,
     ).toEqual([]);
+  });
+
+  it('every datasource-admin ledger entry naming a client method resolves to a real function', () => {
+    const broken = brokenIn(DATASOURCE_ROUTE_LEDGER);
+    expect(
+      broken,
+      `datasource-admin ledger entries claiming a client method that does not exist: ${broken.join('; ')}`,
+    ).toEqual([]);
+  });
+
+  it('the datasource-admin family is audited as reaching NO client method', () => {
+    // Said as a measurement rather than left implicit, because the assertion
+    // above holds vacuously while every row is `server-only` — and a guard
+    // that can only ever pass is the "declared but unverified" shape these
+    // ledgers exist to remove. What is measured here is the #7744 audit's
+    // actual finding: `client.datasources` reaches the FEDERATION family in
+    // packages/rest (ledgered there, as `datasources.external.*`), and no
+    // method reaches the Setup/console lifecycle family. The day the SDK
+    // gains a datasource-lifecycle method, this expectation is the one that
+    // fails and sends the author to the ledger row that must name it.
+    const claimed = DATASOURCE_ROUTE_LEDGER.map((e) => e.client).filter((c) => c != null);
+    expect(claimed).toEqual([]);
+
+    const datasources = (client as unknown as { datasources: Record<string, unknown> }).datasources;
+    expect(Object.keys(datasources)).toEqual(['external']);
   });
 
   it('the whole `storage` namespace is backed by a ledger row', () => {
--- a/packages/services/service-datasource/src/datasource-route-ledger.conformance.test.ts
+++ b/packages/services/service-datasource/src/datasource-route-ledger.conformance.test.ts
@@ -1,0 +1,159 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * Datasource-admin route-ledger conformance (#7744) — the guard that keeps the
+ * autonomously-mounted datasource-admin surface and `@objectstack/client` from
+ * drifting apart silently, mirroring the storage (#3636) and i18n (#3636)
+ * guards.
+ *
+ * Directions made loud here:
+ *
+ *  1. A route `registerDatasourceAdminRoutes` mounts with no ledger entry — a
+ *     new route landed without a reviewed SDK disposition. This is the
+ *     direction that was open when #7744 was filed: all ten routes were in it.
+ *  2. A ledger entry for a route the registrar no longer mounts — the ledger
+ *     went stale.
+ *
+ * ENUMERATION IS DERIVED, NOT TRANSCRIBED, and that is the whole point of the
+ * file. The registrar runs against a capturing mock `IHttpServer` and its
+ * registration calls ARE the route set — the same seam the two tranche-3
+ * ledgers use. No path literal in this file is compared against the mount: a
+ * hand-copied expectation would go green on a ledger that agrees with the test
+ * and disagrees with the server, which is exactly the class of gap #7744
+ * reported (the REST ledger spelled the family `/external/tables` while the
+ * mount said `/remote-tables`, and nothing was comparing either to a mount).
+ * Add an eleventh route and this file fails until it is ledgered.
+ *
+ * The third direction — "every `sdk` row names a client method that exists" —
+ * lives in `packages/client/src/service-route-ledger-coverage.test.ts`, next to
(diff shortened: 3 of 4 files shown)


hotlong:
**PM review — `domain:cli` seat (#6024).** Verdict: **accept**, green, enqueueing. No change requests.

## The landing-site deviation is correct, and the dispatch was one of the things that was wrong

This PR does **not** put the entry in `packages/rest/src/rest-route-ledger.ts`. The card said to, the triage grade said to, and **my own dispatch instruction said to** — all three were wrong, and this PR is right. Verified independently before accepting, because a deviation from a binding instruction has to be checked rather than taken on the PR's word:

- `packages/rest/src/direct-mount-composition.ts` composes exactly two registrars — `registerPackageRoutes` (`:117`) and `registerExternalDatasourceRoutes` (`:137`). `registerDatasourceAdminRoutes` is not among them, so a `direct-mount` row there would fail the REST guard's own registrar-identity assertions on arrival.
- **Control, and it is better than a zero-hit**: the single occurrence of `registerDatasourceAdminRoutes` anywhere in `packages/rest/` is a *comment* — `external-datasource-routes.ts:26`, noting the datasource service mounts them "via its own `registerDatasourceAdminRoutes`". The REST package already documents that these are not its routes.
- The live mount is `packages/cli/src/commands/serve.ts:2658`.

The binding requirement — a ledger entry **at the live spelling**, pinned by a **mount-derived** test — is met exactly; only the file differs, and it differs toward the shape the repo already grew for this class (`service-storage` / `service-i18n`, #3636). ⛔ No live route was renamed, which was the one thing the dispatch forbade.

## Two anti-vacuity moves worth calling out, because neither was asked for

- The conformance guard asserts **non-emptiness and size identity** (`live.size > 0`, `live.size === ledger.size`) on top of the two set differences — both of which pass vacuously against an empty enumeration. A refactor that moves the mount would otherwise leave this file green while guarding nothing.
- In the client-side guard, the author noticed that `brokenIn(DATASOURCE_ROUTE_LEDGER)` **holds vacuously** while every row is `server-only`, and added a measurement that fails the day the SDK grows a datasource-lifecycle method. A guard that can only ever pass is the exact shape these ledgers exist to remove; catching that in one's own new test is the right instinct.

## (a) is falsified, and that is the success outcome

All three conditions fail, each cited, with **a control on both greps** as required.

Condition 1 is dispositive and settles it on its own: late binding is not merely possible, it is the **written contract** — `contracts/datasource-driver-factory.ts:7-18` states that when no factory `supports()` a driver id, "create/update skip hot pool registration (the driver is picked up on the next boot instead)". That *is* accept-then-report, specified at the seam. Corroborated by the per-call factory resolution (`datasource-connection-service.ts:137`, "may be registered after init") and by `start()`'s `restoreRuntimeDatasources()` → `rehydratePools()`. Refusing at create would make a documented authoring state unreachable.

Condition 3 is nearly as strong: `status:'error'` is a **named** state with its own vocabulary entry — `supports() === false` → `'skipped-unsupported'` → `availabilityOf` → `'failed'` → `summaryStatus` — and the vocabulary doc lists "unsupported driver" under `failed` explicitly.

Condition 2 is honestly reported as the weaker one (both pinning tests drive a *known* driver, so an unknown-driver-only refusal would not by itself turn them red; what they pin is the accept-and-return-summary shape). Saying that plainly rather than rounding it up to a third confirmation is what makes the other two believable.

And ⛔ no consolation change was made — the error message was not "improved". That instruction was followed to the letter.

**So the card's second half is not a bug.** `status:'error'` + `statusReason: "no driver factory supports driver '…'"` is health-reporting working as designed on a create path documented to accept a draft whose driver is not buildable in this process.

## Also endorsed

The `check:test-typecheck` diagnosis carries a **revert-and-reproduce control** (reverting only this PR's client edit reproduces the identical 3 `TS2307`s in untouched files), which is what distinguishes an AGENTS.md §9 unbuilt-dependency artefact from a real ratchet movement. And #7954 / #7955 were split out rather than folded in — particularly #7955 (only the federation spelling forwards `?schema=`, so the admin spelling silently drops the filter), which is a real behaviour divergence and correctly not fixed inside a bookkeeping PR.

25/25 checks green, `mergeable_state: clean`. Enqueueing.

---

Issue #7678, opened 2026-08-11
suggested-binding-loop (b): unknown ?status on /security/suggested-bindings returns 200 empty instead of 400 (live REST route skips validation)


huangyiirene:
## Symptom
`GET /api/v1/security/suggested-bindings?status=garbage` → 200 **empty** (same for `?status=PENDING`) — which reads as "there are no suggestions", the failure mode the clause forbids. Expected 400 for an unknown status value.

## Root cause
Located by the run. The **live** route is `packages/rest/src/rest-server.ts` `registerSecurityEndpoints`, which forwards `req.query.status` through with **no validation** (rest-server.ts ~L9219: `status: req.query?.status ? String(req.query.status) : undefined`). The `isSuggestionStatus` 400 guard exists **only** on the parallel runtime dispatcher domain, whose own comment describes precisely this observed behaviour. The fix landed on the dispatcher path only — a dispatcher-vs-REST divergence in the **opposite** direction from the earlier `/meta` cases (where routes were on the dispatcher but never mounted on REST). Confirmed present on `origin/main`.

This strengthens the case for the route-ledger↔live-mount parity gate objectstack-ai/objectstack#7526, and argues that gate should cover **behaviour** divergence, not just mount presence.

## Reproduction
1. Stock boot; as admin `GET /api/v1/security/suggested-bindings?status=garbage`.
2. Observe 200 with an empty list instead of 400.

## Suggested fix
Apply the `isSuggestionStatus` validation (the same guard already on the dispatcher) on the live `registerSecurityEndpoints` REST route, returning 400 for an unknown `?status` value.

## Source
Extracted from the QA run objectstack-ai/objectstack#7637 (framework 92f26f75, console 09987b680).

hotlong:
Claim: PM loop round 8 — `domain:cli` seat.

- Session: `session_01B3Kurx8qufrDzNjk4rag7V` (account `hotlong`)
- Branch: `claude/issue-7678-suggested-bindings-status-validation`
- Worktree: dedicated (cloud container)
- Domain: `domain:cli`
- File surface: `packages/rest/src/rest-server.ts` — the `registerSecurityEndpoints` / suggested-bindings region (~:9219) only + tests. Stop on breach; explain in the report.
- Container & model: M, `mode:cloud`, `model: claude-opus-5`.
[...]

Issue #7653, opened 2026-08-11
open-edition-honest-degradation c3: anonymous-deny ordering inverted on /ai/** (unauth /ai/models → 501, /ai/agents → 200 instead of 401)


huangyiirene:
The degradation itself is **honest and PASSES** — please do not "fix" that half: `/ai/models`, `/conversations`, `/usage`, `/chat` all 501 with the Cloud/EE remedy sentence (mounted-but-unimplemented, never 404/503); `/ai/agents` returns the declared envelope with the relocated payload; and `/discovery`'s `services.ai` message is string-identical to the 501 body. The defect below is only the ordering of the anonymous-deny gate.

## Symptom

Unauthenticated callers reach the `/ai/**` family instead of being denied.

- **Observed:** unauth `GET /api/v1/ai/models` → **501** (leaking the Cloud/EE remedy sentence); unauth `GET /api/v1/ai/agents` → **200**. Reproduced twice.
- **Expected:** both → **401**.
- **Control (proves anonymous-deny is otherwise live on the same boot):** anonymous `GET /api/v1/data/showcase_task` → **401**, correctly.

Disclosure impact is limited to two static strings, but the ordering contract this item exists to pin is inverted.

## Root cause

`runtime/src/domains/ai.ts` `handleAIRequest` returns the `/ai/agents` empty-list courtesy (`{ agents: [] }`) and `capabilityUnavailable(deps, 'ai')` at the **top** of the handler (around `ai.ts:68-79`), while the `shouldDenyAnonymous` gate lives **inside** the per-route loop further down (`ai.ts:128`) — reachable only when the AI service *is* serveable. So on every open-edition deployment the whole `/ai/**` family answers unauthenticated callers before the anonymous gate is ever consulted. Still present on `origin/main` (the courtesy + `capabilityUnavailable` return precedes the `shouldDenyAnonymous` check).

## Reproduction

On a stock boot, with no `Authorization` header:

```
curl http://&lt;host&gt;/api/v1/ai/models    # expected 401, actual 501
curl http://&lt;host&gt;/api/v1/ai/agents    # expected 401, actual 200
curl http://&lt;host&gt;/api/v1/data/showcase_task   # control: 401 (correct)
```

## Source

Extracted from the QA run objectstack-ai/objectstack#7627 (framework 92f26f75, console 6314e87f).

hotlong:
Claim: PM loop round 2 — `domain:cli` seat.

[...]

Issue #2567, opened 2026-07-04
[security] Anonymous posture inconsistent across surfaces: dispatcher /graphql and raw hono /data bypass requireAuth (/meta now gated)


os-zhuang:
Surfaced by the self-review of PR #2562 (the ADR-0056 D2 `requireAuth` default flip). The flip makes REST `/data/*` deny-anonymous by default, but sibling HTTP surfaces never consult `requireAuth` / `enforceAuth`, so the platform's anonymous posture is **inconsistent by surface** rather than uniform. An attacker doesn't care which door is locked — until every entry point honors the same posture, the D2 flip's security promise is not actually delivered. None of these gaps was introduced by #2562 (they are pre-existing), but the flip makes them newly load-bearing.

## Status of the three original surfaces

| # | Surface | Status |
|---|---------|--------|
| 1 | **Metadata `/meta/*`** — `RestServer.registerMetadataEndpoints` | ✅ **Fixed** — the route registrar is now wrapped for the duration of metadata registration (`packages/rest/src/rest-server.ts` `registerMetadataEndpoints`), so every present and future `/meta` route inherits `enforceAuth`. Keep as the reference pattern for Phase 1. |
| 2 | **Dispatcher GraphQL** — `packages/runtime/src/dispatcher-plugin.ts` (~570) mounts `POST {prefix}/graphql` → `dispatcher.handleGraphQL(...)` with no auth check; the security middleware falls open for an anonymous context (no roles/permissions/userId → `next()`) | ❌ Open — an anonymous GraphQL query reads object data that the sibling `/data/*` 401 denies |
| 3 | **Raw hono CRUD** — `HonoServerPlugin.registerStandardEndpoints` (defaults **on**, `packages/plugins/plugin-hono-server/src/hono-plugin.ts` ~179) registers `/api/v1/data/:object` routes that delegate straight to ObjectQL and never consult `requireAuth` | ❌ Open — routes are only shadowed when the REST plugin registers the same paths *first*, so secure-by-default currently depends on **plugin registration order**. A load-order change silently reopens the hole and no test fails. |

## Root cause

[...]

PR #7958, opened 2026-08-12
fix(runtime): consult anonymous-deny gate before /security's 503

### LINKED ISSUE #7911: finding: anonymous-deny ordering inverted on /security too — the 503 "Security service not available" answers unauthenticated callers ahead of the gate
Found while fixing #7653 (`/ai/**` anonymous-deny ordering, PR #7910). Filed rather than folded into that PR — different domain, and the card scoped that work to `domains/ai.ts`.

## The defect

`packages/runtime/src/domains/handleSecurityRequest` (`domains/security.ts`) resolves the security service and returns a **503** when the slot is empty or the occupant does not duck-type — **before** it consults the anonymous-deny gate:

```ts
// domains/security.ts:71-74
const service = await deps.resolveService(context, 'security', context.environmentId);
if (!service || typeof service.listAudienceBindingSuggestions !== 'function') {
    return { handled: true, response: deps.error('Security service not available', 503) };
}

const ec = context.executionContext;
// …
// domains/security.ts:92
if (!ec || shouldDenyAnonymous({ userId: ec.userId, isSystem: ec.isSystem })) {
```

So on any deployment where the `security` slot is empty or stubbed, an unauthenticated caller to `/api/v1/security/suggested-bindings` gets `503 Security service not available` instead of `401 UNAUTHENTICATED` — a capability answer served ahead of the gate. That is the same inversion #7653 records for `/ai/**`, on a surface the handler's own comment calls an "Admin surface — anonymous is denied UNCONDITIONALLY (#2567, #3963)".

## Why it is worth a ticket even though the string is bland

Smaller disclosure than the AI one: the message names no remedy and no package, so what leaks is only *whether this deployment serves the security slot*. The value is the contract, not the secret — `/security` is meant to sit on the same anonymous-deny floor as `/data`, `/meta`, `/actions` and `/automation` (ADR-0056 D2 → #3963), and an anonymous caller should not be able to tell a 503-vs-401 apart on an admin surface. It is also the last of the six dispatcher domains still ordered this way.

## Surveyed, for whoever picks this up

Every sibling in `packages/runtime/src/domains/` was checked for the same shape while verifying #7653. Exactly one is inverted:

| domain | gate | first capability return | verdict |
|---|---|---|---|
| `automation.ts` | `:156` | `capabilityUnavailable` `:174` | ✅ gate first — the reference, with a comment giving the reason |
| `meta.ts` | `:172` | — | ✅ gate first |
| `actions.ts` | `:129` | — | ✅ gate first |
| `packages.ts` | `:224` | — | ✅ gate first |
| `ai.ts` | `:128` (in-loop) | courtesy `:68`, 501 `:79` | ❌ #7653 — fixed in #7910 |
| `security.ts` | `:92` | 503 `:73` | ❌ **this issue** |

`domains/automation.ts` is the shape to copy: hoist the gate above the service probe.

## Suggested fix

Move the `!ec || shouldDenyAnonymous(...)` block above the `resolveService` probe. Note the `!ec` arm is already documented as behaviour-preserving (`#4127 batch 3`), so hoisting does not change what it decides — only when. No route-level `auth: false` opt-out exists on this domain, so unlike `/ai` this is a straight hoist with a single consult site.

## Verification the fix would need

Positive: anonymous `GET /api/v1/security/suggested-bindings` with an empty `security` slot → **401** with `error.code === 'UNAUTHENTICATED'` (assert code *and* status — ADR-0112). Negative: with a valid principal and an empty slot, the **503 `Security service not available`** answer is unchanged; with a serveable slot the route still works for an authenticated caller and still denies anonymous.

Not assigned — recording, not claiming.

hotlong:
Fixes #7911

## The defect

`handleSecurityRequest` (`packages/runtime/src/domains/security.ts`) resolved the `security` service and returned **503 "Security service not available"** for an empty or non-duck-typing slot — **before** it reached the `!ec || shouldDenyAnonymous(...)` gate ~20 lines below. So on a deployment where the `security` slot is empty or stubbed, an unauthenticated caller to `/api/v1/security/suggested-bindings` got a **503 capability disclosure** instead of the **401 UNAUTHENTICATED** refusal this handler's own comment calls unconditional (#2567, #3963).

`/security` stands on the same anonymous-deny floor as `/data`, `/meta`, `/actions` and `/automation` (ADR-0056 D2 → #3963). This was the last of the six dispatcher domains still ordered the wrong way — the sibling inversion on `/ai/**` was #7653, fixed in PR #7910.

## The fix

Straight hoist, mirroring #7910's shape on `domains/ai.ts`: move the `!ec || shouldDenyAnonymous(...)` block above the `resolveService` probe. The `!ec` arm is documented (`#4127 batch 3`) as behaviour-preserving, so the hoist changes **when** the decision is made, not **what** it decides. No route-level `auth: false` opt-out exists on this domain, so there is a single consult site — simpler than `/ai`. All pre-existing explanatory comments are preserved, re-sited with the code they now sit next to.

## Pins

- Anonymous + empty/stubbed `security` slot → **401**, asserting both `error.code === 'UNAUTHENTICATED'` and the HTTP status (ADR-0112).
- Authenticated + empty slot → the **503 "Security service not available"** answer is **unchanged** (the negative pin proving this is a hoist, not a deletion).
- Serveable slot: still works for an authenticated caller, still denies anonymous.

## Tests

New `packages/runtime/src/domains/security-anonymous-deny-ordering.test.ts` (10 cases), mirroring `ai-anonymous-deny-ordering.test.ts`'s group shape:
- Group A: anonymous + empty/stubbed slot → 401 (list + confirm + dismiss routes).
- Group B: authenticated caller (incl. internal SYSTEM context) + empty/stubbed slot → 503 unchanged.
- Group C: serveable slot — anonymous still denied, authenticated caller still served (list/confirm/dismiss).

```
pnpm --filter '@objectstack/runtime^...' build   → green
pnpm --filter '@objectstack/runtime' typecheck    → clean
vitest run src/domains                            → 23 files, 461 tests passed
```

**Reverse verification**: reverted the hoist (`git checkout origin/main -- packages/runtime/src/domains/security.ts`), re-ran the new test file — the 4 Group A (401) cases failed with `expected 503 to be 401` (right reason, not a compile error), Groups B/C stayed green. Restored the fix and re-ran green.

**`check:type-check-debt` ratchet**: re-measured `@objectstack/runtime`'s TEST_DEBT the same way the gate does (sibling `tsconfig.debt-remeasure.json` lifting the test exclusion, `tsc --noEmit`) — **227 errors, unchanged** from the recorded ledger value; the new test file adds zero.

## Scope

Does not touch `packages/rest/src/rest-server.ts` (`registerSecurityEndpoints` is in flight on #7678, a different defect).

---

### DIFF
--- a/.changeset/security-anonymous-deny-ordering.md
+++ b/.changeset/security-anonymous-deny-ordering.md
@@ -1,0 +1,27 @@
+---
+"@objectstack/runtime": patch
+---
+
+fix(runtime): consult the anonymous-deny gate before `/security`'s capability answer (#7911)
+
+On any deployment where the `security` slot is empty or its occupant does not
+duck-type `ISecurityService`, an **unauthenticated** caller to
+`/api/v1/security/suggested-bindings` got **503 "Security service not
+available"** instead of **401 UNAUTHENTICATED** — a capability disclosure
+served ahead of this admin surface's own "anonymous is denied
+UNCONDITIONALLY" rule (#2567, #3963).
+
+`handleSecurityRequest` resolved the `security` service and returned the 503
+for an empty/non-duck-typing slot *before* reaching the
+`!ec || shouldDenyAnonymous(...)` gate ~20 lines below. `/security` stands on
+the same anonymous-deny floor as `/data`, `/meta`, `/actions` and
+`/automation` (ADR-0056 D2 → #3963); this was the last of the six dispatcher
+domains still ordered the wrong way, after `/ai/**` (#7653, fixed in #7910).
+
+The gate now runs first and decides once; the `!ec` arm is unchanged
+(documented `#4127 batch 3` as behaviour-preserving) so the hoist changes
+*when* the decision is made, not *what* it decides. The 503 answer is
+unchanged for an authenticated caller against an empty/stubbed slot, and a
+serveable slot still works for an authenticated caller and still denies
+anonymous. No route-level `auth: false` opt-out exists on this domain, so
+there is a single consult site.
--- a/packages/runtime/src/domain-handler-registry.test.ts
+++ b/packages/runtime/src/domain-handler-registry.test.ts
@@ -202,8 +202,22 @@ describe('HttpDispatcher extracted domains (PR-2)', () => {
         expect(notification.markRead).toHaveBeenCalledWith('u1', ['n1']);
     });
 
-    it('/security responds 503 when no security service is wired (legacy in-handler semantics)', async () => {
-        const result = await makeDispatcher().dispatch('GET', '/security/suggested-bindings', undefined, {}, {} as any);
+    it('/security still answers 503 from inside the handler when no service is wired, for an authenticated caller (#7911: the anonymous-deny gate now runs BEFORE this, not after)', async () => {
+        // Direct delegate call for the same reason as the notifications case
+        // above: dispatch() re-resolves identity from the (mock, auth-less)
+        // kernel and would overwrite the seeded executionContext with an
+        // anonymous one, and an anonymous caller no longer reaches this probe
+        // at all post-#7911 (see the anonymous-denial test right below). This
+        // test's job is narrower than that: prove the in-handler 503 path
+        // (`!service || typeof service.listAudienceBindingSuggestions !==
+        // 'function'`) still exists once an authenticated caller has cleared
+        // the gate -- the negative pin proving #7911 was a hoist, not a
+        // deletion. Before #7911 the 503 sat AHEAD of the gate and this test
+        // reached it anonymously by accident of the harness ("legacy
+        // in-handler semantics"); now it sits after the gate, so the test
+        // authenticates deliberately instead of relying on that accident.
+        const context: any = { executionContext: { userId: 'admin-1' } };
+        const result = await makeDispatcher().handleSecurity('/suggested-bindings', 'GET', undefined, {}, context);
         expect(result.handled).toBe(true);
         expect(result.response?.status).toBe(503);
     });
--- a/packages/runtime/src/domains/security-anonymous-deny-ordering.test.ts
+++ b/packages/runtime/src/domains/security-anonymous-deny-ordering.test.ts
@@ -1,0 +1,195 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * #7911 — the anonymous-deny gate is consulted BEFORE `/security`'s
+ * capability answer, not after it.
+ *
+ * ## The defect
+ *
+ * `handleSecurityRequest` resolved the `security` service and returned a
+ * capability answer — 503 "Security service not available" — for an empty or
+ * non-duck-typing slot BEFORE it reached the `!ec || shouldDenyAnonymous(...)`
+ * gate ~20 lines below. So on any deployment where the `security` slot is
+ * empty or stubbed, an unauthenticated caller to
+ * `/api/v1/security/suggested-bindings` got a 503 capability disclosure
+ * instead of the 401 refusal this admin surface's own comment calls
+ * UNCONDITIONAL (#2567, #3963). `/security` stands on the same anonymous-deny
+ * floor as `/data`, `/meta`, `/actions` and `/automation` (ADR-0056 D2 →
+ * #3963) — the sibling inversion on `/ai/**` was #7653, fixed in PR #7910;
+ * this was the last of the six dispatcher domains still ordered the wrong way.
+ *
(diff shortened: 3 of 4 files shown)


hotlong:
**PM review — `domain:cli` seat (#6024).** Verdict: **accept**, pending CI. Two notes, neither a change request.

## The docs-drift advisory is a false positive — and it corroborates the fix

Checked with a **control** (a zero-hit grep proves nothing unless the scanner is known to work):

- **Target**: `Security service not available` / `suggested-bindings` across `content/docs/**` → the only hand-written hit is `permissions/permission-sets.mdx:202-204`, which lists the three routes and says nothing about 503-vs-401 ordering. Nothing to update.
- **Control**: `UNAUTHENTICATED` → hits 5 files. Scanner works, so the absence above is real absence.

Better than neutral, though: `content/docs/releases/v15.mdx:815` already states this surface *"hardcodes `requireAuth: true`, 401"*. So `main` has been shipping a release note that claims the 401 this handler was not delivering for an empty slot. This PR makes the code agree with a document that has been right all along. (That page is release-owned and read-only — recording the corroboration, not touching it.)

## One fidelity note on the test double, for whoever copies it next

The header's claim — that `makeDeps().error` is the real builder the dispatcher wires, not a stub that drops the third argument — is right on the argument it is defending, and the reasoning about vacuous `code` assertions is exactly correct. It is worth knowing it is not *general*:

`http-dispatcher.ts:295` wires `error: (message, code, details) => this.error(message, code, details)`, and `this.error` (`:690`) is **not** a bare `apiErrorResponse` — for `httpStatus >= 500` it first runs `looksLikeInternalErrorLeak(message)` and substitutes `INTERNAL_ERROR_MESSAGE` on a match. The double skips that step, and Group B asserts a **503** message — precisely the status class where the two could diverge.

Verified harmless today: `looksLikeInternalErrorLeak('Security service not available')` returns **false** on every arm of the predicate (`packages/types/src/error-leak.ts:51-65`) — no `sqlite_`, no `sqlstate`, no `constraint failed` / `unique constraint` / `foreign key`, and it does not start with `select `/`update `/`insert into `/`delete from `. So the double and production agree on this message, and the assertion is real.

The latent edge is only this: if that 503's wording ever changes to something the heuristic matches, Group B keeps passing while production ships `Internal server error`. Not worth a change here — noting it so the next test that copies this double for a 5xx path knows which axis it is thin on.

## Everything the dispatch asked for is present

Hoist with all `#4127 batch 3` reasoning preserved and re-sited; the authenticated-caller 503 pinned as the negative control that distinguishes a hoist from a deletion; anonymous refusal asserting `code` **and** status; the `isSystem` arm covered; reverse verification reported with the right failure mode (`expected 503 to be 401`, not a compile error); and the `check:type-check-debt` ledger re-measured the way the gate measures it — `@objectstack/runtime` 227, unchanged.

Holding out of the merge queue until the queued checks finish.

---

hotlong:
**CI red — `Test Core (2/3)`, one test. Diagnosed from the job log; patch round dispatched.**

```
FAIL  src/domain-handler-registry.test.ts:208
      > /security responds 503 when no security service is wired (legacy in-handler semantics)
AssertionError: expected 401 to be 503
```

139 files / 2137 tests otherwise green.

**This is a good failure.** A pre-existing test pinned the exact behaviour this PR intentionally changes, and it caught it — which is the ratchet doing its job, not a defect in the fix. It was missed locally because the run was scoped to `vitest run src/domains`; the failing test sits one level up, in `src/domain-handler-registry.test.ts`. The changed file is exercised from outside its own directory.

**On the repair, since the tempting fixes are both wrong.** Read the test's *name*: its purpose is to prove the 503 originates **inside the handler** — it is not a statement about anonymous ordering, and it only ever reached that 503 anonymously as an artefact of the harness. The test immediately below it already covers anonymous denial against a wired service.

So: ⛔ not deleted, and ⛔ **not** re-pointed at 401 — that would collapse it into a duplicate of its neighbour and leave nothing in the suite proving the in-handler 503 path survives, which is precisely the negative control this change needs.

Instead it adopts a pattern already present and documented in the same file: `dispatch()` re-resolves identity from the mock, auth-less kernel and **overwrites any seeded `executionContext` with an anonymous one** (see the comment at `:188-190` on the `/notifications` tests, which call the public delegate directly for this reason). Reaching the handler that way with an **authenticated** context lets the test prove what its name claims, without asserting the ordering this PR fixes. The name is being updated too — "legacy in-handler semantics" described a world where the 503 came *before* the gate; it now sits after it.

Ledger constraint restated for the patch round: `@objectstack/runtime` TEST_DEBT stays at **227**. Staying draft until green.

---

Issue #7602, opened 2026-08-11
finding: the runtime dispatcher's `capabilities.search` is still presence-gated, so it over-promises the moment the slot is filled


os-zhuang:
## What this is

A **finding**, filed as a latent declared-≠-enforced gap rather than a live user-facing defect. Surfaced while reviewing #7541 / PR #7577; the third-producer half was named in that PR's "deliberately left alone" section, and this card exists so it does not evaporate when that PR merges.

## Observed

There are **two** producers of `capabilities.search`, and #7541 only moved one of them.

| producer | predicate | file |
|---|---|---|
| `getDiscovery()` — the REST/protocol face | `typeof this.searchAll === 'function'`, ANDed with `api.enableSearch` | `packages/metadata-protocol/src/protocol.ts`, `packages/rest/src/rest-server.ts` (both **fixed** by #7541) |
| `getDiscoveryInfo()` — the runtime dispatcher face | `hasSearch = !!searchSvc`, i.e. bare service-slot presence | `packages/runtime/src/http-dispatcher.ts` (~L1111, emitted ~L1372) |

**Today this is not a live defect, and that is the only reason it is a finding rather than a bug.** Nothing in either repository registers the `search` slot (`CORE_SERVICE_PROVIDER` records `'search': null`, verified across both repos, ADR-0115 Evidence 5), so `hasSearch` is `false` on every host that exists. And the dispatcher mounts **no `/search` route at all** — there is no `search` entry in `route-ledger.ts` and no handler in `http-dispatcher.ts` — so `false` is the honest answer for that host. Declared and served agree, by coincidence of an empty slot.

**The gap is what happens when the slot is filled.** Register any `ISearchService` on a dispatcher-served host — precisely what the slot exists for, per `CoreServiceName`'s "Search Engine (Elastic/Meili)" — and that producer flips `capabilities.search.enabled` to `true` while the host still serves no `/search` route. A conforming client reads the document, calls the endpoint, and gets the router's 404. That is Prime Directive #10 in its ordinary direction — an advertised endpoint that 404s — and it is the same defect class #7541 just closed on the other producer, merely waiting on a precondition nobody has met yet.

## Why it was correctly left out of #7541

Two reasons, both worth keeping:

- **The card's seam was the REST/protocol pair.** #7541 was measured against a live `GET /api/v1/search` answering 200 while the document said `false` — a symptom this producer cannot exhibit, since it serves no such route.
[...]

PR #7965, opened 2026-08-12
fix(runtime): search slot reports handlerReady:false, not true (#7939)

### LINKED ISSUE #7939: finding: dispatcher `services.search` reports `handlerReady: true` for a slot with no mounted handler — the one slot #4318 did not reach
## What this is

A **finding**, latent for the same reason #7602 was: nothing registers the `search` slot today, so the wrong branch is unreachable on every host that exists. Surfaced while implementing #7602 (PR #7937), which fixed the *sibling* field on the same slot and left this one deliberately untouched — it is a separate decision on a separate field, and folding it into that PR would have widened a scoped card.

## Observed

`packages/runtime/src/http-dispatcher.ts`, in `getDiscoveryInfo()`'s per-service status map:

```ts
search: searchRegistered ? svcAvailable(undefined, undefined, searchSvc) : svcUnavailable('search'),
```

`svcAvailable()` returns `handlerReady: true` for any occupant that carries no `__serviceInfo` self-description. The map's own contract, stated a few lines above it, is:

> `handlerReady: true` means the dispatcher has a real, bound handler for this route.

The dispatcher has **no** handler for search — no `search` entry in `route-ledger.ts`, no branch in `http-dispatcher.ts`, and the call site passes `route: undefined` precisely because there is nothing to advertise. So a filled slot yields a record that says, in one breath, *"no route"* and *"the handler is ready"*.

That is the contradiction **#4318** named and closed for `cache` / `queue` / `job`. Its remedy is already in this file: `svcInProcess(name, svc)`, which returns `handlerReady: false` plus `inProcessServiceMessage(name)` — "this is a kernel-internal contract with no HTTP surface", said honestly. `search` is the one slot of that shape that was not converted, and the only reason the omission never showed is that no `ISearchService` has ever been registered (`CORE_SERVICE_PROVIDER` records `'search': null`).

## Why this is a finding and not a bug

Identical grading to #7602: declared and served agree today, by coincidence of an empty slot. The wrong record can only be produced by a host that registers a search service, and none exists in either repository.

## Why it was left out of PR #7937

- #7602's ruling was about `capabilities.search` — whether the *capability* is advertised. This is `services.search.handlerReady` — how the *slot occupant* is described. Different field, different producer contract (`capabilities` describes the host's HTTP surface; `services` describes what is registered), different remedy.
- #7602's declared file surface covered the capability block; PR #7937 states this explicitly in its "out of scope" section rather than silently widening.

## What a fix would look like

Almost certainly one line, mirroring #4318:

```ts
search: searchRegistered ? svcInProcess('search', searchSvc) : svcUnavailable('search'),
```

Worth deciding rather than assuming, on one point: `inProcessServiceMessage('search')`'s wording was written for kernel-internal contracts (`cache`/`queue`/`job`). A search engine is not kernel-internal in the same sense — it is an external engine with no *dispatcher* surface — so the message may want its own phrasing even if the `handlerReady: false` half is settled.

A fix should also carry a test: the same shape PR #7937 used for the capability — register a search service, assert `handlerReady === false` — since like every other assertion on this slot, it is vacuous unless the fixture fills it.

Filed unassigned and without a `domain:*` label so routing stays the triage seat's call (`packages/runtime` ⇒ `domain:cli` per the domain table, if that is the disposition).

Related: #4318 (the same contradiction, fixed for `cache`/`queue`/`job`), #7602 / PR #7937 (the capability half of this slot), #2462 / ADR-0076 D12 (honest self-report).

hotlong:
Fixes #7939

## What changed

`packages/runtime/src/http-dispatcher.ts`'s `getDiscoveryInfo()` built
`services.search` with `svcAvailable(undefined, undefined, searchSvc)`, which
defaults an unmarked occupant to `handlerReady: true` — contradicting the same
map's own stated contract a few lines above: `handlerReady: true` means "the
dispatcher has a real, bound handler for this route." The dispatcher has no
`/search` route or handler at all (no `route-ledger.ts` entry, no branch in
`http-dispatcher.ts`, `route` is always `undefined` at this call site), so a
filled slot reported "no route" and "handler is ready" in one breath.

This is the exact contradiction #4318 closed for `cache`/`queue`/`job`, applied
to the one slot of that shape it did not reach. `search` now uses the same
`svcInProcess()` remedy:

```ts
search: searchRegistered
    ? svcInProcess('search', searchSvc, SEARCH_IN_PROCESS_MESSAGE)
    : svcUnavailable('search'),
```

**`capabilities.search:1403` (PR #7937 / #7602) is untouched**, per the
boundary this card drew: that field describes the host's HTTP surface
(`enabled: false`, already landed and correct); this field describes what is
*registered*, and stays presence-gated as before — only the `handlerReady`
value for a filled slot changes.

## Message wording choice

`svcInProcess`'s shared `inProcessServiceMessage(name)` opens "Kernel-internal
service — consumed in-process via the service registry" — true for
`cache`/`queue`/`job` (kernel-managed, in-process by construction) but **false**
for `search`: a registered search service is an *external* engine
(Elasticsearch/Meilisearch, per `core-services.zod.ts`'s own
`REMEDY_DETAIL['search']`), not a kernel-internal one. Reusing the shared
sentence would misdescribe it, so `svcInProcess` now takes an optional
`fallbackMessage` (used by `search` alone; `cache`/`queue`/`job` are
unaffected — no third argument, same message as before) and `search` gets its
own wording:

> Search engine registered, but the dispatcher has no HTTP route for the
> 'search' slot — no dedicated search endpoint is mounted on its behalf.
> Cross-object search, where a host serves it, is reported separately by
> capabilities.search.

## Tests

Two new tests in `packages/runtime/src/http-dispatcher.test.ts`, in the
`discovery honest capabilities (D12)` describe block, right after the #4318
cache/queue/job tests:

- `reports an unmarked search occupant with no route and handlerReady false,
  not true (#7939)` — registers a real, unmarked search-service-shaped
  occupant (the assertion is vacuous otherwise, since
  `CORE_SERVICE_PROVIDER['search']` is `null` and nothing registers this slot
  today) and asserts `enabled: true`, `status: 'available'`, `handlerReady:
  false`, `route: undefined`, and that the message is search's own wording
  (not the shared "Kernel-internal" sentence).
- `reports a self-describing search occupant (e.g. a dev stub) with its own
  status, still handlerReady false (#7939)` — a `__serviceInfo`-marked
  occupant keeps its own declared status/message, `handlerReady` stays false.

**Reverse verification** (commit-then-revert, never `git stash` — see
AGENTS.md): committed the fix, then `git checkout HEAD~1 --
packages/runtime/src/http-dispatcher.ts` to restore the pre-fix
`svcAvailable(...)` line while keeping the new tests, and re-ran just the
`#7939`-tagged tests:

```
FAIL  src/http-dispatcher.test.ts > HttpDispatcher > discovery honest capabilities (D12) > reports an unmarked search occupant with no route and handlerReady false, not true (#7939)
AssertionError: services.search.handlerReady: expected true to be false
- Expected: false
+ Received: true
 Tests  1 failed | 1 passed | 240 skipped (242)
```

Failed for exactly the right reason (`handlerReady: true` from the old
`svcAvailable` line), not a compile error or a vacuously-skipped assertion —
the second test (the `__serviceInfo`-marked stub) still passed, as expected,
since `svcAvailable` already honors self-description. Then restored the fix
(`git checkout claude/issue-7939-dispatcher-search-handler-ready --
packages/runtime/src/http-dispatcher.ts`) and confirmed the working tree was
byte-identical to the committed fix (`git status --short` empty).

Full suite after restoring:

```
pnpm --filter @objectstack/runtime exec vitest run src/http-dispatcher.test.ts --maxWorkers=2
 Test Files  1 passed (1)
      Tests  242 passed (242)
```

## Type-check debt ratchet (`check:type-check-debt`)

`packages/runtime`'s `tsconfig.json` excludes `*.test.ts`, so the package's own
`typecheck` script (clean) never compiled the new test code — the ratchet does,
separately. My first draft of the new test read `info.services.search.route`
directly, which does not typecheck: `svcInProcess()`'s return type carries no
`route` key at all (route-less by construction), same as `cache`/`queue`/`job`.
Fixed by casting the same way the existing `#4318` tests already do
(`(info.services as Record< string, any >).search`). Re-measured
`@objectstack/runtime`'s TEST_DEBT by generating the same sibling
tsconfig the gate itself generates (dropping the test exclude) and running
`tsc --noEmit` directly — **227 errors, matching the recorded ledger exactly**,
confirmed before *and* after the cast fix (228 → 227 with the cast fix,
i.e. it was a +1 regression from my own new code, now resolved to +0).

## Local verification

- `pnpm --filter '@objectstack/runtime^...' build` (dependency closure)
- `pnpm --filter @objectstack/runtime exec vitest run src/http-dispatcher.test.ts --maxWorkers=2` — 242/242 pass
- `pnpm exec turbo run typecheck --concurrency=2` (repo-wide, per this card's Definition of done) — 126/126 tasks green
- `node scripts/check-nul-bytes.mjs` — OK
- Runtime TEST_DEBT re-measured directly against `tsc --noEmit` — 227, unchanged from the recorded ledger

## Out of scope, left alone

- `capabilities.search` (`:1403`) — PR #7937 / #7602, untouched per this
  card's explicit boundary.
- `metadata-protocol`'s own `services.search` (a different producer,
  `packages/metadata-protocol/src/protocol.ts`) — it genuinely does mount a
  real `/api/v1/search` handler via `searchAll()`, so its presence-gated
  `route`/`handlerReady` shape is legitimately different and already
  deliberately pinned (`packages/objectql/src/protocol-discovery.test.ts`,
  "should leave non-dispatcher-owned routes presence-gated"). No parity test
  was added between the two producers for this field, for the same reason
  the file's own comments give for not requiring one.

---

### DIFF
--- a/.changeset/dispatcher-search-slot-handler-ready.md
+++ b/.changeset/dispatcher-search-slot-handler-ready.md
@@ -1,0 +1,29 @@
+---
+"@objectstack/runtime": patch
+---
+
+fix(runtime): discovery no longer claims a bound handler for the `search` slot (#7939)
+
+`getDiscoveryInfo()`'s `services.search` entry reported `handlerReady: true` for
+any registered search-service occupant that carries no `__serviceInfo`
+self-description — even though the dispatcher has no `/search` route or handler
+at all (no `route-ledger.ts` entry, no branch in `http-dispatcher.ts`). That
+contradicted the map's own stated contract: `handlerReady: true` means "the
+dispatcher has a real, bound handler for this route."
+
+This is the same contradiction #4318 closed for `cache`/`queue`/`job`, applied
+to the one slot of that shape it did not reach. `search` now uses the same
+`svcInProcess()` remedy, reporting `handlerReady: false` for a filled slot —
+with its own message rather than the shared "Kernel-internal service" wording,
+since a registered search service is an external engine (Elasticsearch/
+Meilisearch), not a kernel-managed in-process contract.
+
+`capabilities.search` (the host's `/search` HTTP surface, fixed separately in
+#7602 / PR #7937) is untouched — this only corrects `services.search`, which
+describes what is *registered*, not what is served.
+
+Latent until now: nothing registers an `ISearchService` in either repository
+(`CORE_SERVICE_PROVIDER['search']` is `null`), so the wrong branch was never
+reachable in practice. Covered by two new fixture-filled tests that register a
+search occupant before asserting `handlerReady`, since an empty-slot test would
+pass regardless of what the line says.
--- a/packages/runtime/src/http-dispatcher.test.ts
+++ b/packages/runtime/src/http-dispatcher.test.ts
@@ -3043,7 +3043,11 @@ describe('HttpDispatcher', () => {
             (kernel as any).services = new Map([['search', searchSvc]]);
 
             const info = await dispatcher.getDiscoveryInfo('/api/v1');
-            const reported = info.services.search;
+            // Cast like the #4318 cache/queue/job tests above: `svcInProcess`'s
+            // return type carries no `route` key at all (route-less by
+            // construction), so a direct `info.services.search.route` read
+            // does not typecheck against the narrowed union.
+            const reported = (info.services as Record<string, any>).search;
             expect(reported.enabled, 'services.search.enabled').toBe(true);
             expect(reported.status, 'services.search.status').toBe('available');
             expect(reported.handlerReady, 'services.search.handlerReady').toBe(false);
--- a/packages/runtime/src/http-dispatcher.test.ts
+++ b/packages/runtime/src/http-dispatcher.test.ts
@@ -3024,6 +3024,53 @@ describe('HttpDispatcher', () => {
                 }
             }
         });
+
+        // ── `search`: the one slot of the #4318 shape it did not reach (#7939) ──
+        //
+        // `svcAvailable(undefined, undefined, searchSvc)` defaulted an unmarked
+        // occupant to `handlerReady: true` — a lie by this very map's own
+        // contract (stated above `svcAvailable`'s definition): the dispatcher
+        // has NO `/search` route (no `route-ledger.ts` entry, no branch here,
+        // `route` is always `undefined`), so "the handler is ready" cannot be
+        // true. Latent until now because `CORE_SERVICE_PROVIDER['search']` is
+        // `null` — nothing registers this slot anywhere — so the fixture below
+        // fills it explicitly; without that the assertion would be vacuous.
+        it('reports an unmarked search occupant with no route and handlerReady false, not true (#7939)', async () => {
+            const searchSvc = { /* real, unmarked ISearchService-shaped occupant */
+                index: vi.fn(), remove: vi.fn(), search: vi.fn(),
+            };
+            (kernel as any).getService = vi.fn().mockImplementation((n: string) => (n === 'search' ? searchSvc : null));
+            (kernel as any).services = new Map([['search', searchSvc]]);
+
+            const info = await dispatcher.getDiscoveryInfo('/api/v1');
+            const reported = info.services.search;
+            expect(reported.enabled, 'services.search.enabled').toBe(true);
+            expect(reported.status, 'services.search.status').toBe('available');
+            expect(reported.handlerReady, 'services.search.handlerReady').toBe(false);
+            expect(reported.route, 'services.search.route').toBeUndefined();
+            // The remedy string is search's OWN wording, not the shared
+            // cache/queue/job "Kernel-internal service" sentence — a search
+            // engine is an external occupant with no dispatcher surface, not a
+            // kernel-managed one, so reusing that sentence would misdescribe it.
+            expect(reported.message, 'services.search.message').not.toContain('Kernel-internal');
+            expect(reported.message, 'services.search.message').toMatch(/no HTTP route/i);
+            expect(reported.message, 'services.search.message').toContain('capabilities.search');
+        });
+
+        it('reports a self-describing search occupant (e.g. a dev stub) with its own status, still handlerReady false (#7939)', async () => {
+            const stubSearch = {
+                __serviceInfo: { status: 'stub', message: 'Development stub — no real search backend' },
+                index: vi.fn(), remove: vi.fn(), search: vi.fn(),
+            };
+            (kernel as any).getService = vi.fn().mockImplementation((n: string) => (n === 'search' ? stubSearch : null));
+            (kernel as any).services = new Map([['search', stubSearch]]);
+
+            const info = await dispatcher.getDiscoveryInfo('/api/v1');
+            expect(info.services.search.enabled).toBe(true);
+            expect(info.services.search.status).toBe('stub');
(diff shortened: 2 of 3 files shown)


hotlong:
**PM review — `domain:cli` seat (#6024).** Verdict: **accept**, pending the queued checks. No change requests.

## The three things the dispatch was strictest about, all delivered

- **The boundary held.** `capabilities.search` at `:1403` is untouched; only the `services.search` filled-slot branch moved. Those two fields were separated on purpose by PR #7937 and this PR keeps them separate.
- **The assertion is not vacuous.** Both tests register a search occupant before asserting. That mattered more here than usual: `CORE_SERVICE_PROVIDER['search']` is `null`, so an empty-slot test would have passed no matter what the line said.
- **The ledger was not raised.** The first draft of the test measured **228** against a recorded 227 — a +1 from the new code — and it was diagnosed (`svcInProcess()`'s return type carries no `route` key) and fixed with the same cast the existing #4318 tests use, back to **227**. That is the correct response to that ratchet; raising it would have been the wrong one.

The second test — a `__serviceInfo`-marked stub keeping its own `status`/`message` while `handlerReady` stays `false` — was not asked for and is worth having: it pins that self-description wins for *description* but not for the *route claim*.

Two judgement calls I want to endorse explicitly, since both could have gone wrong quietly:

**The message wording.** The discretion granted at dispatch was used, and used correctly: `inProcessServiceMessage`'s "Kernel-internal service" is true for `cache`/`queue`/`job` and **false** for an external search engine. Threading an optional `fallbackMessage` — defaulting to the shared string so the other three slots are byte-identical — is the minimal way to say the true thing.

**Leaving `metadata-protocol`'s `services.search` alone.** That is a different producer which genuinely mounts `/api/v1/search` via `searchAll()`, so its shape is *legitimately* different. "Make the two agree" would have been the plausible-looking wrong move.

## Docs: verified clean — but the advisory above is not why

The docs-drift comment lists 20 pages. Checked with a control:

- **Target**: `handlerReady` across `content/docs/**` → the page that actually specifies it is **`content/docs/kernel/services-checklist.mdx`** (9 hits: the default table at `:192`, the `__serviceInfo` contract at `:187`, *"Only `handlerReady` is consulted, never `status`"* at `:197`, and the `search` slot at `:54`/`:79`/`:94`/`:488`/`:525`).
- **Control**: `services.` under `content/docs/protocol/kernel/` → hits. Scanner works.

**That page is not in the advisory's list of 20.** Read it anyway: every mention of `search` there describes the **empty**-slot branch — *"Nothing provides the slot at all — `search`. Discovery says exactly that"* — which this PR leaves untouched (`svcUnavailable('search')`). No doc states what a *filled* search slot reports, because nothing has ever filled one. **So no doc is made wrong, and no docs change is needed.**

The advisory being simultaneously noisy (20 mostly-irrelevant pages) and blind to the one that matters is a checker defect, not a defect in this PR. Filed separately as **#7967** — root cause measured: the check maps docs to packages by *textual mention*, and `services-checklist.mdx` never spells `@objectstack/runtime` (control: it names eight other `@objectstack/*` packages).

Staying draft until the queued checks finish; enqueueing on green.

---

Issue #7968, opened 2026-08-12
finding: `GET /automation/:name/runs/:runId/screen` still discloses record-derived values to any authenticated caller who knows a run id


hotlong:
Observation-class finding, filed **unassigned with no `pm:queue`** — for triage to grade and route. Named residual of #7900 (PR pending), deliberately not decided inside that card because closing it needs a *different* mechanism from the one #7900's maintainer ruling names.

## What #7900 closed, and what it left

The 2026-08-12 ruling on #7900 converged the `/automation` run-state reads on the `sys_automation_run` object-read grant. `GET /:name/runs/:runId` and `GET /:name/runs` now require it. The ruling also asked for the rest of the domain to be audited against the same rule, with a stated reason for anything that stays authenticated-only.

`GET /:name/runs/:runId/screen` is the one route the audit left authenticated-only, and the reason is on the route in `packages/runtime/src/domains/automation.ts`:

- It is the **end user's** surface, not the operator's. The pause exists because the flow is asking *this* caller to fill a form in; the trigger response already handed them the same `screen` inline, and the route exists so a browser refresh does not lose it. Requiring an operator grant here would refuse the screen to the very person the flow paused for — a breakage, not the narrowing the ruling prices in ("operator tooling … now needs the grant").
- Its **write sibling already answers on a different axis**: `resume` is gated in the engine by the suspension's declared `resumeAuthority` (#3801 / #5561) — a per-run authority model, not an object grant. Read and write on the same pause answering to two unrelated permissions would be the incoherence #7900 exists to remove.

## The residual, measured

The route is not inert with respect to record data. `AutomationEngine.getSuspendedScreen(runId)` returns `(await this.loadSuspendedRun(runId))?.screen` — the `ScreenSpec` persisted with the suspended run — and that spec is built with values **interpolated against the live flow variables** at suspend time (`packages/services/service-automation/src/builtin/screen-nodes.ts`: `interpolate(cfg.defaults, variables, context)` and, per field, `interpolate(f.defaultValue, variables, context)`).

So a flow whose screen prefills from the triggering record (`defaultValue: '{record.email}'`, a `defaults` map over `{record.*}`) writes those values into the spec, and the route answers them to **any authenticated caller who knows the run id** — there is no check anywhere on this path that the run belongs to, or was paused for, the caller. Same shape as the #7900 exposure, one route over and a narrower payload: interpolated screen defaults rather than the whole variable map.

[...]

Issue #7450, opened 2026-08-10
The runtime dispatcher serialises a PermissionDeniedError's `details` to the client, so `positions` / `permissionSets` reach the browser on the `/data` transport


os-zhuang:
Found while implementing #7414 (PR #7449). Filed separately because fixing it means changing what a 403 **body** carries, which #7414 explicitly ruled out of scope (that card is user-facing copy only, structured payload untouched).

## What was measured

`plugin-security`'s object gate attaches a structured payload to every `PermissionDeniedError`:

```ts
throw new PermissionDeniedError(
  message,
  { operation: opCtx.operation, object: opCtx.object, positions, permissionSets: explicitPermissionSets },
);
```

The two transports treat that payload differently, and only one of them was checked when #7414 was written.

**`@objectstack/rest` — `mapDataError`, `packages/rest/src/rest-server.ts:726-737`:**

```ts
return {
    status: 403,
    body: {
        error: error?.message ?? 'Permission denied',
        code: 'PERMISSION_DENIED',
        ...(object ? { object } : {}),
    },
};
```

`error.details` is never read. Nothing structured reaches the client, and the `object` that does ride is the object the **route** named, not `error.details.object`.

**`@objectstack/runtime` — `http-dispatcher.ts:1945-1951`:**

```ts
} catch (e) {
    if (isPermissionDeniedError(e)) {
        return {
            handled: true,
            response: this.error(e.message, 403, { code: 'PERMISSION_DENIED', ...(e.details ?? {}) }),
        };
    }
    throw e;
}
```

`this.error` hands that object to `apiErrorResponse` / `buildApiError` (`packages/runtime/src/error-envelope.ts`), which promotes the string `code` into `error.code` and puts **everything else on the wire as `error.details`**. `/data` is served by this dispatcher (the domain registry, `packages/runtime/src/domains/data.ts`), and that domain handler does not catch — so a permission denial on an ordinary CRUD request answers with `error.details.positions`, `error.details.permissionSets`, `error.details.operation` and `error.details.object`.

So on that transport a client is told the caller's position names, the names of the permission sets attached to their context, and — on a cascade delete, where `cascadeDeleteRelations` re-authorises each child independently — the API name of a **child** object the caller never addressed.

## Why this matters beyond tidiness

[...]

Issue #5561, opened 2026-08-05
automation: `resumeAuthority` defaults to `'any'`, so every future pausing node ships fail-open — ADR-0044 says this is "tracked separately" and nothing tracks it


os-zhuang:
Found while implementing #3823 (PR #5560). Not a defect anyone hits today — filed as an observation so the ADR's "tracked separately" is true.

## What

`ActionDescriptor.resumeAuthority` (`packages/spec/src/automation/node-executor.zod.ts`) defaults to `'any'`. The #3801 resume gate therefore covers a pause only when its node type's author **remembered the flag**: a new pausing node type ships raw-resumable by omission, which is the "declared ≠ enforced" trap (Prime Directive #10) one node away.

That is not hypothetical drift — it is exactly how #3823 happened. ADR-0044 pointed the revise edge at a generic `wait`, `wait` is `'any'` (correctly, for a signal wait), and the pause in the service-owned position inherited the fail-open default. The demonstrated cost there was an unaudited resubmit plus, opportunistically, remote run destruction.

ADR-0044's 2026-07-28 amendment records the long-term fix and defers it:

> *Fail-closed descriptor default.* … The long-term-correct default is that a pause is **not** raw-resumable unless it opts into `'any'` (screen / wait-signal declaring it explicitly). A breaking change to the descriptor default; needs a migration; **tracked separately**.

A search of open + closed issues for `resumeAuthority` returns #3823, #3853 and an unrelated typecheck issue — **no issue tracks this**. So the ADR points at a tracking item that does not exist, which is how a deferred decision quietly becomes a forgotten one.

## Current state (why this is observation-class, not a defect)

Every pausing type in the tree today is correct on its own terms:

| type | `resumeAuthority` | correct? |
|---|---|---|
| `screen` | `'any'` (default) | yes — the UI flow-runner owns it |
| `wait` | `'any'` (default) | yes — an external producer is meant to resume a signal wait |
| `approval` | `'service'` (explicit) | yes (#3801) |
| `approval_revise` | `'service'` (explicit) | yes (#3823 / PR #5560) |
| `subflow` / `map` | `'any'`, but the gate follows the linked run to the child | yes (#3853) |

So nothing is exploitable through the default right now. The exposure is prospective: the next pausing node type — ours or a plugin's — is fail-open unless someone remembers, and the two `'service'` declarations that exist were both added *after* a bypass was demonstrated rather than at design time.

## What a fix would involve (sketch, not a decision)

[...]

Issue #3801, opened 2026-07-28
automation: the generic run-resume route needs an authorization gate keyed on the suspended node


os-zhuang:
Found while working on #3800. Separate seam, same trust boundary — filing rather
than widening that PR's scope.

**Detail is deliberately withheld here**, same convention as #3800: the analysis
lives in the private draft advisory under the repo's Security tab. Note it is
**reasoned from code, not demonstrated** — no repro was written, so treat the
severity as strong-but-unproven until someone does.

## Requirement

`POST /api/v1/automation/:name/runs/:runId/resume`
(`packages/runtime/src/domains/automation.ts`) forwards a caller-supplied
`branchLabel` into `AutomationEngine.resume`. `resumeInternal` validates
*machine* state only — the concurrent-resume latch, the run exists, the flow
exists, the suspended node still exists — and nothing about **who is asking**.

Approval nodes suspend and resume through exactly this mechanism, so the route
needs an authorization gate of its own; today the only thing standing between it
and the approvals rules is convention. The showcase spells that convention out in
a comment ("decide via the approvals API, never a raw engine `resume`"), and
`route-ledger.ts` lists the route as SDK-exposed — a comment in an example is not
an access control.

Beyond authorization, a resume that goes around the service also leaves no
`sys_approval_action` row and no status mirror, so the request row and the run
can end up permanently disagreeing.

## Why it is not simply "remove the route"

The route is load-bearing for **screen flows**: the UI flow-runner posts
`{ inputs }` to advance a paused `screen` node, which is its whole reason to
exist. The gate has to discriminate by **what the run is parked on**, not by the
route — a run suspended on an `approval` node should be resumable only through
`ApprovalService` (which records the decision and enforces the slate), while a
`screen` node keeps working as it does now.

Sketch, for discussion rather than a settled design: have the engine tag a
suspension with the node type that produced it, and have the resume route refuse
a suspension whose owner declares itself service-gated, with the service-side
resume passing an internal marker the route cannot forge.

Refs #3800.

PR #7969, opened 2026-08-12
fix(runtime): converge /automation run-state reads on the sys_automation_run read grant

### LINKED ISSUE #7900: finding: the whole /automation read domain is gated only by "authenticated" — run-detail returns the triggering record's fields without that record's own FLS
Observation-class finding, filed **unassigned with no `pm:queue`** — for triage to grade and route. Handed up by the dev on #7639 / PR #7896 as an explicit *"for you to grade, not for me to decide"* residual, and graded by the services PM as **not a blocker for that PR** (reasoning below) but worth its own card.

## What was measured

`packages/runtime/src/domains/automation.ts:135-147` applies the #5519 **anonymous baseline** to the **whole `/automation` domain** — not per route, not per field. `automation.ts:489-496` then answers run-detail with `deps.success(run)`, handing back the `ExecutionLogEntry` **verbatim**: no projection, no redaction, no masking, on any field, anywhere on that path. A repo-wide grep for redact/mask/sanitize across `service-automation` and the domain returns zero non-comment hits.

So the only question this surface asks is *"are you authenticated?"*. Any authenticated caller who knows a run id reads whatever that run's log entry holds — including, via the variables snapshot and via `output`, **the triggering record's fields, without that record's own field-level security applying**.

## Why it is worth a card

The exposure is not new and was **not introduced by #7639/PR #7896**, which is exactly why it needs its own card rather than a veto on that PR:

- `output` has always carried record fields **whenever a flow author declared the record (or a projection of it) as an output variable** — so the FLS-bypass property predates the paused-variables work.
- `steps`, flow definitions and the rest of the `/automation` read surface sit behind the same single gate.
- What PR #7896 changed is the **conditionality**: the paused snapshot carries the triggering record and its flattened fields **unconditionally**, where before reaching the same data required an author opt-in. Same data class, same door — broader traffic through it.

Framing this as "a problem with the `variables` key" would be the inaccurate version. The accurate question is a domain-level policy one:

> **Should the `/automation` read surface apply the underlying record's FLS (or any per-object permission) rather than only an authentication check?**

That is a product/security design call with real cost on both sides — narrowing it changes what every existing operator tool can read, and per-field filtering of a free-form variable map is not obviously well-defined (keys are `<nodeId>.<key>`, `record`, `previous`, `$runId`, seeded inputs; only some of them correspond to a record field at all).

## What the other door does today

The identical snapshot has a second, **differently gated** door: `sys_automation_run.variables_json` (`sys-automation-run.object.ts:164-169`) persists it verbatim for every paused run, and that read goes through the system object's permissions. So the platform currently answers the same question two ways depending on which door you knock on — which is itself the sharpest evidence that a deliberate policy has never been stated for this surface.

## Explicitly NOT claimed here

- **No credential class is exposed.** Measured on #7639: connector credentials live in the connector registry and are never handed to a flow; the `http` node's `signingSecret`/headers are config **inputs** interpolated *from* variables and never written back (`builtin/http-nodes.ts:115-170`, output is `{response, status}`). A secret reaches variables only if a flow author puts one there — and by that same route it reaches `output`. This card is about **record-field FLS**, not secret leakage.
- **No regression is asserted.** Everything above is true of `origin/main` before PR #7896 as well, for any flow whose author declared record data as output.

## Suggested routing

Landing site is `packages/runtime/src/domains/automation.ts` (the gate) and possibly `plugin-security` (whatever would supply the record's FLS), so this likely is **not** `domain:services` even though it surfaced there — the `service-automation` engine is not where the gate lives. Triage's call.

If graded for work, this probably wants `needs-user-decision` before implementation rather than a dispatch: the narrowing decision has a real blast radius, and per #4829 the time to put it in the maintainer's inbox is when there is pull — which there now arguably is, since PR #7896 makes the exposure unconditional.

## Evidence

Measured 2026-08-12 by the dev on PR #7896 (`origin/main` base `68a1edb`), reading `packages/runtime/src/domains/automation.ts`, `packages/services/service-automation/src/engine.ts`, `builtin/http-nodes.ts`, and `sys-automation-run.object.ts`. Full measurement recorded in the `<!-- os-dev-report -->` comment on #7639 and in PR #7896's body.

## Source

Split out of objectstack-ai/objectstack#7639 (PR objectstack-ai/objectstack#7896) — the dev's named residual, deliberately not decided inside that card.

hotlong:
Fixes #7900

Implements the maintainer ruling of 2026-08-12 verbatim: **the `/automation` read surface requires the same permission the `sys_automation_run` object read answers with; per-field filtering of the variables map is rejected as the mechanism.**

## The exposure this closes

`packages/runtime/src/domains/automation.ts:489-496` answered run-detail with `deps.success(run)` — the `ExecutionLogEntry` verbatim, no projection, no redaction, no masking. The only gate on the path was the #5519 anonymous baseline, applied to the whole domain rather than per route (`:135-147`), so the sole question the surface asked was *"are you authenticated?"*. Any authenticated caller who knew a run id read the triggering record's fields with that record's own FLS never applying. `GET /:name/runs` served the same entries under the same non-gate.

The identical snapshot's second door — `sys_automation_run.variables_json` — has always gone through the system object's permissions. One platform, two answers.

## The mechanism, and why no new seam was built

Both run-state reads now consult `ISecurityService.explain({ object: 'sys_automation_run', operation: 'read' })` with the caller's own execution context. That contract exists for exactly this — *"the query surface that lets code OUTSIDE the ObjectQL engine middleware ask the same questions the middleware answers when it enforces access"*, with a standing instruction that a consumer re-deriving these answers locally will drift. `explain` runs the same permission-set resolution, the same `PermissionEvaluator` and the same RLS compiler the enforcement middleware runs (`allowed = !capsDeny && crudAllowed && !denyAll && !delegatorMissing` over that shared machinery), so this gate cannot answer differently from the `/data` door by drifting.

**Fork clause: not hit.** The `security` slot was already on `DomainHandlerDeps` (`domains/meta.ts` resolves it the same way for ADR-0106 masking) and `explain` was already on the contract. No new cross-package seam exists to report, and no second permission system was invented.

**⛔ Not per-field filtering of `variables`** — rejected by the ruling, and a granted caller still receives the map untouched (pinned by deep-equal below).

Which routes are gated is declared as **one predicate** (`isRunStateRead`), not a check per branch — the point of the ruling is that the domain gets one policy, and a policy spelled at three call sites is three policies that happen to agree today. The gate sits with the #5519 floor and ahead of the service-availability probe, for that gate's own reason read one authorization tier up: which permission a route requires must not vary with which automation service a deployment mounts.

## Route-by-route audit

Every `/automation` route, with its verdict. The reason for each authenticated-only verdict is **in the source at the route**, not only here.

| Route | Verdict | Reason |
|:---|:---|:---|
| `GET /:name/runs/:runId` | **🔒 `sys_automation_run` read** | The measured exposure. Serves the `ExecutionLogEntry` verbatim, snapshot and `output` included. |
| `GET /:name/runs` | **🔒 `sys_automation_run` read** | Serves the same entries a page at a time — same data class, same door. |
| `GET /:name/runs/:runId/screen` | authenticated-only | The **end user's** surface, not the operator's: the pause exists because the flow is asking *this* caller to fill a form in, and the route is the refresh-safe re-fetch of a `screen` the trigger response already returned inline. Its write sibling `resume` answers on the engine's per-run `resumeAuthority` axis (#3801 / #5561), not on an object grant. Requiring the operator grant would refuse the screen to the person the flow paused for — a breakage, not the narrowing the ruling prices in. **Residual is real and not claimed closed** (a `ScreenSpec`'s `defaults`/`defaultValue` are interpolated against live flow variables): filed as #7968, which needs a per-run authority read gate, a different mechanism from this ruling. |
| `GET /` (listFlows) | authenticated-only | Flow **names**. Not `sys_automation_run`-class data — no run, no trigger record, no snapshot — so the grant this ruling names says nothing about it. |
| `GET /:name` (getFlow) | authenticated-only | A flow **definition**: metadata-plane data, governed on the metadata plane (`/meta`, ADR-0106). Narrowing its read posture is a metadata-plane decision and belongs to its own card. |
| `GET /actions` | authenticated-only | The deployment's action **catalog** (ADR-0018) — a static registry, no run data. |
| `GET /connectors` | authenticated-only | The deployment's connector **catalog** (ADR-0022) — same. |
| `GET /_status` | authenticated-only | Per-flow enabled/bound runtime state. Flow state, not run state. |
| `POST /`, `PUT /:name`, `DELETE /:name`, `POST /:name/trigger`, `POST /trigger/:name`, `POST /:name/toggle`, `POST /:name/runs/:runId/resume` | out of scope | Writes. The ruling is about the **read** surface; `resume` in particular already answers on the engine's `resumeAuthority` axis. |

The common thread in every authenticated-only verdict except `/screen`: requiring the `sys_automation_run` grant for flow-definition and registry data would not be convergence — it would be a **second policy invented for a different data class**, which is what the ruling forbids. `/screen` is the one route where the data class *does* match and the verdict still differs; that is why it is the one carrying a filed follow-up rather than a closing argument.

## The three non-denials, each a recorded decision

1. **System context passes.** The middleware's very first act is `if (opCtx.context?.isSystem) return next()`. A gate that refused what the object read admits is not convergence either.
2. **No `plugin-security` ⇒ no grant to require.** Such a deployment has no object-permission system at all, so `/data/sys_automation_run` is itself ungated: *"authenticated is enough"* is what **both** doors answer, and refusing here would put them in disagreement the other way. The contract mandates this tolerance. Same for a partial service that omits `explain`.
3. **An `explain` throw is a denial.** Access-narrowing answers fail closed — the stance `plugin-security` itself takes on an unresolvable object posture (#3545).

### One stated asymmetry

The middleware skips its CRUD gate entirely for an authenticated caller whose permission-set resolution comes back **empty** (`if (permissionSets.length > 0)`), while `explain` runs `checkObjectPermission` over that empty list and gets `false`. ADR-0090 D5's additive baseline plus the post-resolution fallback make an empty resolution reachable only on a deployment configuring **no** baseline permission set at all — and there this surface refuses where `/data` falls open. Left as-is deliberately: the divergence is in the closed direction on the door this card was filed about, and closing it the other way would mean re-deriving the middleware's own empty-set rule in the runtime domain, which is the drift `ISecurityService` exists to prevent. Recorded on the gate.

## Tests

New: `packages/runtime/src/domains/automation-run-read-permission-gate.test.ts` — 18 cases.

- **Refusal** on both run-state reads: `code` **and** `status` (ADR-0112) — `PERMISSION_DENIED` + `403` — and the automation service is never consulted (`getRun` / `listRuns` not called), so the snapshot is not even loaded for the refused caller. Plus: the refusal body contains neither the caller's positions nor their user id (#7450), and an unresolvable permission answer fails closed.
- **Positive control** — the pin that proves a narrowing rather than a breakage: a caller **with** the grant deep-equals the whole fixture, `variables` included down to the nested `record` map, and the runs list is unchanged. Same fixture shape `automation-run-detail-passthrough.test.ts` asserts on, which continues to pass untouched.
- **One question** — `explain` is asked for `read` on `sys_automation_run`, with the caller's own context and **no** `userId` (naming a target user is an administrative act plugin-security gates on `manage_users`).
- **The audit as a table** — each authenticated-only route is driven and asserted not-403 **and** asserted never to consult the grant, so changing any verdict has to change this file.
- **The three non-denials**, and the #5519 anonymous floor still answering `401`/`UNAUTHENTICATED` ahead of the new gate.

## Reverse verification

Fix committed first, then `git checkout origin/main -- packages/runtime/src/domains/automation.ts`, suite re-run, then restored from the commit (`git status` clean, `git diff HEAD` empty).

**5 of 18 failed, all as assertion failures — no compile error, no import crash.** The observed run-detail response for an explicitly-refused caller on the un-gated code, captured with a temporary probe (since removed):

```
{"status":200,"body":{"success":true,"data":{"id":"run_7","status":"paused",
 "variables":{"record":{"id":"ord_1","amount":90000,"margin_pct":4.5}}}}}   | getRun called: 1
```

A **200 carrying the record's fields** to a caller the security service had just answered `allowed: false` for — the finding, reproduced. With the gate restored: 18/18 pass.

## Docs

`content/docs/automation/flows.mdx` documented both run-state reads — in the "Observing runs" section and in the endpoint table — with no mention of any permission, which after this gate lands understates the contract. Both now name the grant and carry the permission-set snippet, and the callout records why the screen re-fetch is deliberately *not* gated the same way. `content/docs/references/**` is generated and untouched; nothing under `content/docs/releases/` is edited.

## Gates

- `pnpm -w typecheck` — clean (126/126 tasks).
- `pnpm check:type-check-debt` — `OK … none above its recorded number`. `@objectstack/runtime` stays at its recorded 227; the new test file adds **0** errors, and no ledger number was raised.
- `pnpm --filter @objectstack/runtime test` — 140 files / 2146 tests pass (re-run after merging `origin/main`).
- `pnpm --filter @objectstack/http-conformance test` — 72 pass.
- `pnpm check:doc-anchors` (210 fragment links) and `pnpm check:doc-authoring` (375 files) — clean.
- ESLint clean on the changed source files.

## Migration

Carried in the changeset (`.changeset/automation-run-read-permission-gate.md`), because this is a **behaviour narrowing** and deployments need to pre-grant *before* upgrading:

> Any identity that reads automation run history or run detail over HTTP — operator dashboards, monitoring pollers of `GET /automation/:name/runs?status=failed`, support tooling that opens a run by id, scripted health checks — must now hold **read on `sys_automation_run`**:
>
> ```ts
> permissions: [{ name: 'automation_operator',
>   objects: { sys_automation_run: { allowRead: true } } }]
> ```
>
> Nothing else changes for a caller that already holds it — the response body is unchanged, `variables` map included. System-context callers and the engine's internal paths never go through this seam. Screen-flow end users are unaffected.

## Out-of-scope finding filed

- #7968 — `GET /:name/runs/:runId/screen` still discloses record-derived values (interpolated screen defaults) to any authenticated caller who knows a run id. Filed unassigned, with the reason it cannot be closed by this ruling's mechanism.

https://claude.ai/code/session_01NKkycFdNjr44tbvpbJa4BA


hotlong:
**PM review — `domain:cli` seat (#6024).** Verdict: **accept**, pending the three checks still running (Test Core 1/3 and 3/3, TypeScript Type Check). No change requests. One residual for the maintainer's eyes, below — not a blocker.

Every binding item from the ruling and the dispatch is present: `explain({ object: 'sys_automation_run', operation: 'read' })` as the mechanism; per-field filtering of `variables` refused *and* pinned refused (a granted caller deep-equals the fixture down to the nested `record` map); refusal asserting `code` **and** `status`; the route-by-route audit with a stated reason for every authenticated-only verdict; the migration note in the changeset with the actual permission snippet. The fork clause is answered rather than skipped — `security` was already on `DomainHandlerDeps` (`domains/meta.ts` resolves it the same way for ADR-0106) and `explain` already on the contract, so there was no new seam to report.

The reverse verification is the strongest part: not "5 tests failed" but the captured body of what the un-gated code actually served —

```
{"status":200,"body":{"success":true,"data":{"id":"run_7","status":"paused",
 "variables":{"record":{"id":"ord_1","amount":90000,"margin_pct":4.5}}}}}   | getRun called: 1
```

— a 200 carrying the triggering record's fields to a caller the security service had just answered `allowed: false` for. That is #7900's finding reproduced, not merely re-argued.

## Three judgements beyond the ask, all endorsed

**One predicate, not three checks.** `isRunStateRead` in one place, with the reason stated: *"a policy spelled at three call sites is three policies that happen to agree today."* The ruling's whole point was that the domain gets **one** policy; implementing it as three agreeing conditionals would have satisfied the letter and lost the substance.

**The gate sits ahead of the service-availability probe.** Reasoned one tier up — which permission a route requires must not vary with which automation service a deployment mounts. That is the same ordering lesson as #7653/#7910 and #7911, applied here without being asked. Worth noting that this lane closed the last inverted dispatcher domain this morning and the next PR into the same package arrived already carrying the principle.

**`/screen` is the one route where the data class matches and the verdict still differs — and it is filed, not argued away.** The reasoning holds: the pause exists because the flow is asking *that* caller to fill in a form, its write sibling `resume` answers on the engine's per-run `resumeAuthority` axis rather than an object grant, and requiring the operator grant would refuse the screen to the person the flow paused for. The residual is real (interpolated `ScreenSpec` defaults are record-derived) and #7968 says so plainly instead of claiming closure. That is the correct shape for a route the ruling's mechanism cannot reach.

## The one residual worth the maintainer's attention

The PR records an asymmetry honestly, and I want it surfaced rather than left in a subsection: the enforcement middleware **skips its CRUD gate entirely** for an authenticated caller whose permission-set resolution comes back empty (`if (permissionSets.length > 0)`), while `explain` runs `checkObjectPermission` over that empty list and returns `false`.

So on a deployment that configures **no baseline permission set at all**, this surface refuses where `/data/sys_automation_run` falls open — the two doors disagree, in the opposite direction from the one this card was filed about.

I endorse leaving it: the divergence is in the **closed** direction on the door being secured, and closing it the other way would mean re-deriving the middleware's own empty-set rule inside the runtime domain, which is precisely the drift `ISecurityService` exists to prevent. But the operational consequence should be said out loud rather than inferred from the changeset: on such a deployment, operator dashboards and monitoring pollers go dark on upgrade. The remedy is the same one the migration note already gives — configure a permission set granting `sys_automation_run` read — so it is recoverable, not a trap. ADR-0090 D5's additive baseline plus the post-resolution fallback make the precondition rare, and it is stated on the gate in source.

Enqueueing when the three remaining checks land.

---

Issue #7665, opened 2026-08-11
[security] A by-id write is not gated by record visibility — a contributor mutates records they cannot read, when only select-scope RLS is authored


huangyiirene:
## Impact

An authenticated **low-privilege** user modifies records they **cannot read**. Not admin-only, not owner-only — the probing persona held the ordinary `showcase_contributor` set. Reproduced on **three objects** (`showcase_invoice`, `showcase_invoice_line`, `showcase_task`), twice each on fresh rows. The run calls this **the strongest finding of the sweep**. `DELETE` is correctly refused (no delete bit) — only the **write** half is open.

Ruled `target:v17` class ① (a user-hitting defect on a shipped surface). "You can't mutate what you can't see" is not enforced.

## Reproduction (2/2 per object)

Provision two contributor personas C1 and C2 (sign-up + `sys_user_permission_set` showcase_contributor + `sys_user_position` `contributor`, then re-sign-in so the session carries the position).

1. As **C1**: create an invoice (`owner` = C1) and a line under it.
2. As **C2** — the READ side is correct:
   - `GET /api/v1/data/showcase_invoice_line/&lt;lineId&gt;` → **404 RECORD_NOT_FOUND**
   - `GET /api/v1/data/showcase_invoice_line?$top=300` → the line is **absent**
3. As **C2** — the WRITE side is NOT:
   - `PATCH /api/v1/data/showcase_invoice_line/&lt;lineId&gt;` `{"description":"C2-FORGED","quantity":999,"unit_price":12345}` → **200**, admin re-read shows the forged values persisted.
   - Same on the master invoice (C2 GET → 404, C2 PATCH → 200 persisted) and on `showcase_task`.

The platform's own explain engine states the split — the cleanest confirmation:
- `POST /security/explain {object:'showcase_invoice', operation:'read', userId:&lt;C2&gt;}` → rls layer **"narrows"**
- the same call with `operation:'update'` → rls layer **"not_applicable — No RLS policy applies"**

## Root cause (located, re-verified on `origin/main`)

`packages/plugins/plugin-security/src/security-plugin.ts`:

- **2.7 the by-id write pre-image gate** (`:1378`) composes the RLS filter for the **UPDATE** operation and is documented as skipped when that returns null. The comment at `:1553` states it outright: for an object whose narrowing is authored as select-only, *"the #1994 pre-image check above is a no-op for it"*.
[...]

PR #7975, opened 2026-08-12
fix(verify): make the RLS by-id-write class reachable; re-decide the two authz-matrix rows on their own evidence

### LINKED ISSUE #7685: Cross-cutting (verify/qa coverage): verify --rls cannot reach by-id-write, authz-conformance overclaims 'enforced', and the console permission-matrix editor is read-only by default
## Symptom
Three coverage/tooling findings from the run, consolidated per the maintainer instruction. Itemised:

**(i) `verify --rls` reports 0 HOLES but structurally cannot reach the by-id-write class.** Its member probe holds **no object grants**, so every such probe is masked by the object-level gate (403) before record scope is ever tested. A second probe persona holding object read+edit but **outside** the record scope would catch it. Compounding this, a `showcase_account` auto-record 400 cascades into 5 downstream skips, so **8 of 23 objects are skipped on a stock run** — and a skip is exactly where the privately-reported D11 defect hid. The run calls this the single highest-value fix: the tool's green is currently not evidence. (Tool: `packages/verify/src/rls.ts`.)

**(ii) `authz-conformance.matrix.ts` overclaims.** It marks both `rls-by-id-write` (#1994) and `controlled-by-parent` as `state:'enforced'`, while neither holds as shipped. The showcase's own `permission-sets.ts` comment repeats the same false claim. (File: `packages/qa/dogfood/test/authz-conformance.matrix.ts` — both entries carry `state: 'enforced'` on `origin/main`.)

**(iii) The console permission-matrix editor is read-only by default even for a writable package.** On a stock boot it renders "Read-only (OS_METADATA_WRITABLE not enabled)" with every checkbox disabled, at both the metadata-admin route **and** inside Studio's Access pillar for a **writable** package — because the editor computes `writable = !!resolved.allowOrgOverride && !readOnly` and the server returns `allowOrgOverride:false` for type `permission`. Yet the server **accepts** the package-door write in that same default env (`PUT …?package=&lt;writable pkg&gt;` → 200 via `allowRuntimeCreate`). The type gate locks a surface the server would allow.

## Root cause
As located, per item above: (i) the probe persona holds no object grants + an auto-record 400 skip cascade in `packages/verify/src/rls.ts`; (ii) hard-coded `state:'enforced'` rows in `authz-conformance.matrix.ts` (and the mirroring `permission-sets.ts` comment); (iii) the objectui editor's `writable = !!resolved.allowOrgOverride && !readOnly` computation against a server that returns `allowOrgOverride:false` for type `permission`.

## Reproduction
- (i) Run `verify --rls` on a stock showcase → 0 HOLES; note 8/23 objects skipped and that the member probe holds no object grants.
- (ii) Read `authz-conformance.matrix.ts` → `rls-by-id-write` and `controlled-by-parent` are `state:'enforced'`, contradicting the shipped behaviour proven elsewhere in this run.
- (iii) Stock boot; open the permission-matrix editor at the metadata-admin route and inside Studio for a writable package → all checkboxes disabled, "Read-only" banner; yet `PUT /api/v1/meta/permission/&lt;set&gt;?package=&lt;writable pkg&gt;` → 200.

## Routing note
Filed as one `finding` in objectstack-ai/objectstack with `domain:cli`, because the dominant fix site is the verify/qa tooling (items i–ii). Item (iii)'s fix is **objectui** (the `writable` computation in the console permission-matrix editor); it is kept here as one consolidated finding per the maintainer instruction rather than split, but whoever picks it up should route (iii) to objectui.

## Suggested fix
(i) Add a probe persona with object read+edit but outside the record scope so the by-id-write class is reachable, and stop the auto-record 400 from cascading into skips (a skip must not read as a pass). (ii) Downgrade the two `state:'enforced'` claims to match shipped behaviour and correct the `permission-sets.ts` comment. (iii) In the objectui editor, allow editing when the server would accept the package-door write (`allowRuntimeCreate`) rather than gating solely on `allowOrgOverride`.

## Source
Extracted from the QA run objectstack-ai/objectstack#7637 (framework 92f26f75, console 09987b680).

hotlong:
Fixes #7685 — items **(i)** and **(ii)** only. Item (iii) is objectstack-ai/objectui#4446 and is not touched here.

Cross-links #7665 (the defect whose acceptance criterion 2 this probe implements) and #1994 (the invariant).

## (i) `verify --rls` — the measurement first, because it decides (ii)

`objectstack verify --rls` reported `0 HOLES` on both example apps. Neither number was evidence.

**The probe was answered by the OBJECT gate, never by record scope.** The persona was a bare `signUp()` member holding no object grants, so `checkObjectPermission` refused with 403 before the row-level gate was consulted — and the runner banked that 403 as `rls-consistent`.

**A skip read as a pass, and cascaded.** One `showcase_account` auto-record 400 skipped that object and every object with a required relation to it, while the summary line still said `0 HOLES`.

### What `verify --rls` reports BEFORE and AFTER

| | showcase (23 objects) | crm (6 objects) |
|:--|:--|:--|
| **BEFORE** | 13 consistent, 0 holes, 2 member-visible, **8 skipped** | 4 consistent, 0 holes, **2 skipped** |
| **AFTER** | **20 PROVEN** (20 consistent, 0 holes), 3 not proven | **6 PROVEN** (6 consistent, 0 holes), 0 not proven |

The BEFORE "consistent" counts are the finding, not the baseline: **11 of the 13 showcase verdicts and all 4 crm verdicts were `GET 403`** — the object gate — and only 2 showcase objects (`showcase_inquiry`, `showcase_private_note`) reached record scope with a `GET 404`. On the other 15, no platform regression could have produced `rls-hole`.

AFTER, every consistent verdict is a record-scope `GET 404` + `PATCH 403` with the row unchanged as ground truth. The 3 remaining showcase objects are honestly unprovable by this runner and are printed as such: one has no plain-text field to mutate, two are read-only federated objects.

### Changes

- **The probe persona is the one the class needs**: object read+edit on every declared object, narrowed by an owner policy authored `operation: 'select'` **only** — deliberately the authoring shape that *was* the hole, so every object of every verified app becomes a live regression guard for the by-id write-scope derivation. It is registered at boot (`rlsProbeSecurity`) rather than written as a bare `sys_permission_set` row, because the evaluator's last-resort DB loader hydrates the object bits but **not** `rowLevelSecurity` — a row-only probe set would have granted the grants and silently dropped the narrowing.
- **Reachability is measured per object, not assumed.** A LIST 403 is reported as `probe-blocked` — a distinct status that is never a pass — so a persona that silently failed to receive its grants cannot masquerade as a proof.
- **An unsatisfiable admin create no longer cascades.** The runner adopts an existing row as the probe target, which is what took showcase 8 → 3 skips and crm 2 → 0.
- **PROVEN and NOT-PROVEN are separated.** `summary` gains `proven` / `unproven` / `probeBlocked`, `RlsReport.unproven` lists every object the run did not exercise with its reason, `RlsReport.probe` names the persona, and the formatted output prints an explicit *"this run is not a clean bill of health"* line whenever anything went unproven.
- **A degraded probe fails the run** rather than quietly probing with an ungranted member.

### Reverse verification

Ablating the #7665 write-scope derivation in `plugin-security` — while leaving the #1994 pre-image re-read fully in place — and rebuilding the plugin's `dist`:

- **with the fix**: 20 PROVEN, **0 holes**, exit 0
- **ablated**: **16 `rls-hole`**, `hardFailures: 16`, exit 1, each reported as `GET 404 … yet MUTATED it by id (PATCH 200)`

The ablation was reverted and the plugin rebuilt before committing; the branch contains no `plugin-security` change.

### CI gate risk: none

`ci.yml`'s `dogfood-verify` job runs `verify --app <app> --rls` for both example apps and gates through `dogfood-gate`. Both apps exit **0** after this change, so nothing is being suppressed to keep CI green and no blocker decision is needed.

## (ii) `authz-conformance.matrix.ts` — both rows survive, on their own evidence

Decided per row after the measurement, per ADR-0049. **Neither row was downgraded**, so — per the ruling — the mirroring comment in the showcase's `permission-sets.ts` is left alone: it is accurate as shipped.

| Row | Verdict | Citation | Changed |
|:--|:--|:--|:--|
| `rls-by-id-write` (:39) | **stays `enforced`** — the card is stale | `rls-fixture.dogfood.test.ts` is **not** vacuous for this row despite sharing the file with `rls-read`: since #7665 / merged PR #7792 it carries a dedicated select-only block (`:107–172`) whose member set grants **full CRUD** on `rls_note`, so its refusals are the record gate, never the object gate — it asserts the by-id PATCH refused with the row unchanged **and** that an in-scope write still lands. Second, independent measurement: the newly-reachable probe, 20/23 + 6/6 proven, 16 holes under ablation. | `enforcement` + `note` only — **`state` untouched** |
| `controlled-by-parent` (:48) | **stays `enforced`** — the card is stale | `controlled-by-parent.dogfood.test.ts` is a **dedicated** proof, and `fixtures/cbp-fixture.ts` grants the member full CRUD on **both** objects, so every refusal it asserts is the derived record gate. It asserts derived read denial, derived by-id write denial with admin ground truth, **and** that a note under a member-owned master stays readable and writable (not over-blocked). Independently, the probe measures `showcase_invoice_line` — a real `controlled_by_parent` detail — consistent, and it flips to `rls-hole` under the ablation. | `note` only — **`state` untouched** |

The one substantive record correction: `rls-by-id-write`'s `enforcement` named only the *pre-image re-read*, which the ablation proves is a **no-op on its own** under select-only authoring — the 16 holes appeared with that re-read fully present. The row now names both halves. No `covers` key was added or orphaned; `AuthzState` is unchanged.

## Verification

- `pnpm -w typecheck` — clean, 126/126.
- `pnpm lint` — clean.
- `pnpm check:type-check-debt` — OK, none above its recorded number (nothing raised).
- `node scripts/check-verify-stand-in-erasure.mjs` — OK; the new `provisionRlsProbePersona` is classified in `NOT_A_STAND_IN` with its reason, as the reconciliation requires.
- `@objectstack/verify` tests — 23 passed.
- `rls-runner.test.ts` (extended), `authz-conformance.test.ts`, `rls-fixture.dogfood.test.ts`, `controlled-by-parent.dogfood.test.ts` — 27 passed.
- `origin/main` merged and all of the above re-run afterwards.

`rls-runner.test.ts` is the runner's detector-liveness oracle — the live fixture can no longer plant the hole — so it gains cases for `probe-blocked`, target adoption, and the proven/unproven accounting.

## Out of scope, recorded rather than fixed

- The conformance ledger checks only that a cited proof **file exists** (`checkLedger`, `packages/verify/src/conformance.ts:77`), never that it proves the row. That is the mechanism behind this card's premise and it is unchanged here — a general fix touches ~30 rows.
- The probe cannot reach narrowing authored on a **position** it does not hold (the showcase's `positions: ['contributor']` rules); reaching those needs a per-declared-position persona. Stated in the module header so the green is not over-read.
- `showcase_project_membership` is skipped for having no plain-text field to mutate. Widening the probe-field set to the text family risks a sanitizer or a `min`/`max` rejection reading as a false `rls-consistent`, so it was left alone deliberately.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UikkaHrvEBtK2nY31axhSi)_


hotlong:

## The dispatch's sharpest lead was a hypothesis, and it was checked rather than taken

I handed this card a specific lead: `rls-read` and `rls-by-id-write` cite the **same** proof file, so the by-id-write row may have been borrowing credibility from a test that only exercises the read filter. That was a good reason to look. It was also **stale** — since #7665 / merged PR #7792 that file carries a dedicated select-only block (`:107–172`) whose member set grants full CRUD on `rls_note`, so its refusals are the record gate and never the object gate.

The dev verified the lead instead of implementing it. That matters: taking my framing on faith would have downgraded a row that is genuinely enforced — a false claim in the *opposite* direction, one that would have made #7665's fix look unlanded. ⛔ Do not inherit a PM's lead; measure it. Done exactly right here.

Both rows keep `state: 'enforced'`. **The card is stale on both.** And because neither was downgraded, the mirroring comment in the showcase's `permission-sets.ts` is correctly left alone — it is accurate as shipped.

## The finding the card did not contain

`rls-by-id-write`'s `enforcement` field named only the **pre-image re-read** — and the ablation proves that is a **no-op on its own** under select-only authoring: 16 holes appeared with the re-read fully present. So the record was right about the *state* and wrong about the *mechanism*, naming one half of a two-part enforcement. Correcting `enforcement` while leaving `state` untouched is precisely the ADR-0049 distinction this card was about.

## (i): the probe was made harder, not friendlier

The thing I most wanted to be impossible here was tuning the probe to preserve a green. The opposite happened:

- The BEFORE numbers are the finding, not a baseline — **11 of 13 showcase verdicts and all 4 crm verdicts were `GET 403`**, the object gate. Only two objects ever reached record scope. `0 HOLES` was very nearly vacuous, and now that is quantified rather than asserted.
- The persona is narrowed by an owner policy authored `operation: 'select'` **only** — deliberately *the authoring shape that was the hole*. That turns every object of every verified app into a live regression guard for the by-id write-scope derivation.
- `probe-blocked` is a distinct status that is never a pass, so a persona that silently failed to receive its grants cannot masquerade as a proof. That generalizes the "a skip must never read as a pass" requirement instead of patching the one cascade.
- Registering the probe set at boot rather than as a bare `sys_permission_set` row, because the evaluator's last-resort DB loader hydrates the object bits but **not** `rowLevelSecurity` — a row-only probe would have granted the grants and silently dropped the narrowing. That is the same class of silent-drop this whole card is about, avoided one layer down.

And the 0 holes AFTER is backed by an ablation: **16 `rls-hole`, exit 1**, each `GET 404 … yet MUTATED it by id (PATCH 200)`. The detector has teeth, so the zero means something.

**The CI-gate question was answered, not dodged.** `ci.yml`'s `dogfood-verify` runs `verify --rls` for both apps and gates through `dogfood-gate`; both exit 0 after this change. Nothing suppressed, no blocker decision needed — which is what I asked to be told either way.

## Out-of-scope, recorded rather than quietly fixed

The ledger checks only that a cited proof **file exists** (`packages/verify/src/conformance.ts:77`), never that it proves the row — the mechanism behind this card's premise, left unchanged because a general fix touches ~30 rows. Correct call, and worth its own card.

Also honest about what the new green does *not* cover: the probe cannot reach narrowing authored on a position it does not hold, stated in the module header so the result is not over-read; and `showcase_project_membership` is skipped deliberately, because widening the probe-field set risks a sanitizer or `min`/`max` rejection reading as a false `rls-consistent`. Refusing to buy one more "proven" at the price of a possible false pass is the right trade in a tool whose entire job is to not lie.

Enqueueing once CI lands.

---

Issue #7761, opened 2026-08-11
plugin-sharing: `getRule`'s by-id branch has no org filter — an org admin can read, evaluate and DELETE another organization's sharing rule (and its grants)


os-help:
## Symptom

`SharingRuleService.getRule` resolves an id **without any organization filter**, so an org-scoped sharing admin holding another organization's rule id can read, evaluate and delete that organization's rule.

`packages/plugins/plugin-sharing/src/sharing-rule-service.ts`, `getRule`:

```ts
const byId = await this.engine.find('sys_sharing_rule', {
  where: { id: idOrName },   // ← no organization_id predicate
  limit: 1,
  context: SYSTEM_CTX,
});
if (Array.isArray(byId) && byId[0]) return rowFromRule(byId[0]);
```

The read runs under `SYSTEM_CTX`, so nothing downstream re-scopes it either. Three verbs inherit the hole, because all three resolve through `getRule`:

- `GET /api/v1/sharing/rules/:id` — read another org's rule (criteria, recipient, access level).
- `POST /api/v1/sharing/rules/:id/evaluate` — materialise that org's grants on demand.
- `DELETE /api/v1/sharing/rules/:id` — `deleteRule` calls `getRule` and then purges the row **and every `sys_record_share` grant it materialised**, i.e. silently revoke another tenant's record access.

The by-**name** path is correctly scoped (`{name, organization_id: orgId}`), so this is specifically the id branch. The caller still needs `manage_sharing` (or the legacy `manage_platform_settings`) in their own org, and needs to know an opaque `srule_…` id — but neither is a tenant boundary, and ids leak through logs, exports, support tickets and evaluation responses (`{ruleId}` is returned by the evaluate endpoint).

## Not a regression from #7676

Found while auditing callers for #7676 / PR #7760, which widens the two **org-scoped** reads (`listRules`, `getRule`'s by-name fallback) from `organization_id = orgId` to `organization_id = orgId OR organization_id IS NULL`. That change deliberately does not touch this branch and does not make it worse: it never admits another organization's rows. This defect is present on `origin/main` today and predates it.

Worth noting the interaction, though: #7760 makes seeded rules **listable**, and a rule's `id` is in the listing. That is intended (they are platform-global), but it does mean rule ids become easier to come by generally, which raises the value of closing this.

## Suggested direction

[...]

PR #7977, opened 2026-08-12
feat(verify): `bootStack({ orgContext: true })` — a harness admin whose execution context carries an organization (#7762)

### LINKED ISSUE #7762: [finding] `@objectstack/verify` bootStack mints admins with NO organization — every org-scoped read is structurally untestable in the dogfood suite (the reason #7676 escaped)
## Finding

`@objectstack/verify`'s `bootStack` cannot produce an admin whose execution context carries an **organization**, so no dogfood test can exercise any org-scoped code path. Every `organization_id`-filtered read in the platform is, at the HTTP layer, structurally untestable in the open core.

Measured while writing an HTTP regression test for #7676:

```
POST /api/v1/sharing/rules   (as stack.signIn() admin, booted showcase)
→ 201, and the created row comes back organization_id: null
```

`defineRule` stamps `organization_id` from the caller's resolved context, so a null stamp is direct evidence the context has no organization. Consequence: a rule list that should have been filtered was returned unfiltered, and a test asserting "another org's rule stays invisible" saw it plainly — not because the filter was broken, but because the filter never engaged.

The escape hatch is `bootStack(stack, { multiTenant: true })`, and it does not help here: per `packages/verify/src/harness.ts` it requires the cloud-private `@objectstack/organizations` package, which is not present in this repo, and the gates that need it SKIP rather than pretend.

## Why this matters

This is the direct reason #7676 escaped. That defect — package-seeded (`organization_id = null`) sharing rules invisible to an org-scoped admin — was invisible to **both** suites that should have caught it:

- `plugin-sharing`'s own 428-test unit suite: its fake engine has no notion of a resolved authenticated context, so the org id that makes the filter miss only exists after a real sign-in.
- the 579-test dogfood suite: real HTTP, real sign-in, but the admin has no organization, so the filtered branch is never taken.

It took a manual QA run (#7637) to find it. An HTTP-layer regression test was written for PR #7760 and then **deleted**, because it was green on the unfixed code for the wrong reason — a phantom check that reads as coverage. That deletion is the concrete cost of this gap.

The blast radius is wider than sharing rules. Every `organization_id`-scoped read follows this pattern, and several have already been fixed one at a time for the same root cause — `sys_business_unit` approver expansion (#3807), `sys_metadata` pending-draft listing, and now `sys_sharing_rule` (#7676). Each is a defect that only manifests when the caller has an organization, and none of them could have been caught by the harness that boots real HTTP.

## Options

**A. Give the open-core harness a real org-scoped context.** Let `bootStack` establish an organization and mint admin tokens bound to it, without the enterprise plugin — the organization *wall* (RLS scoping, posture) is the cloud-private part, but stamping an `organizationId` on the resolved execution context is not the wall, it is upstream of it. Cheapest route to real coverage; needs care that it does not silently imply a posture the open core cannot honour, which is exactly what `harness.ts` warns about today.

**B. A seam for the resolved context.** Let a test declare the context the REST layer should resolve, so org-scoped filters can be driven directly without modelling org membership. Narrower, more honest about what is being tested, less faithful to a real deployment.

**C. Accept the gap and cover org scoping at the service layer only**, with a shared test kit modelling the org-scoped context — what #7676's unit tests do by hand. Cheapest, and leaves the transport-to-service seam (where #7676 actually lived) uncovered forever.

Recommendation: **A**, gated on confirming the posture question, with **C** as the interim — the recurrence record above suggests this class will keep producing defects, and each one costs a QA run to find.

## Source

Found during #7676 / PR #7760. Filed unassigned; observation-class (no user hits it today) but with a measured escape behind it.

hotlong:
Fixes #7762

## The gap

`@objectstack/verify`'s `bootStack` could not mint an admin whose resolved execution context carried an `organizationId`. Every `organization_id`-filtered read in the platform was therefore structurally **untestable at the HTTP layer** in the open core: the filter never engaged, so a fixture asserting on the difference between a filtered and an unfiltered read saw no difference and passed for the wrong reason.

That is the direct reason #7676 escaped both suites and needed a manual QA run to find — and why PR #7760's dev wrote the HTTP-layer regression test for it, measured it green against the **unfixed** code, and correctly **deleted** it rather than ship phantom coverage.

## What this does — Option A, narrowed

One new `BootOptions` flag in `packages/verify/src/harness.ts`:

```ts
const stack = await bootStack(showcaseStack, { orgContext: true });
```

It flips `AuthPlugin`'s ADR-0081 D1 default-organization bootstrap (`autoDefaultOrganization`) back on — the same bootstrap `objectstack dev` / `serve` give a real single-tenant deployment, not a harness-local imitation. The admin is bound to a real `sys_organization` as `owner`; the `session.create.before` hook stamps `activeOrganizationId` onto the session, which is the one wire field `resolveAuthzContext` reads into `tenantId` → `ExecutionContext`.

No enterprise package, no wall, no posture change. The name deliberately avoids `multiTenant` / `tenant` / `isolated`.

**The boot asserts the bind.** `ensureDefaultOrganization` is best-effort by design (it swallows every failure so a login can never break on org bookkeeping), so a fixture that asked for an org-bound admin and silently got an org-less one is exactly the shape this option exists to abolish. No `sys_member` row for the harness admin → no stack, with an error that says why.

**It refuses to compose with `multiTenant`**, both spellings, rather than boot something weaker than it reads. Under `multiTenant: true` the enterprise package already owns the org bootstrap; under `'posture-only'` the open bootstrap deliberately abstains (walled posture, `postureEnforcesWall`), so the flag would be a pure no-op that still reads as coverage. Both stated in the doc block and pinned by tests.

## ⛔ What it must never be used for

It stamps the **caller's** organization; it stands up **no** organization wall. With no `org-scoping` service registered — and this flag registers none — `SecurityPlugin` **strips** the wildcard `organization_id` RLS policies (`collectRLSPolicies`). A fixture asserting "tenant B cannot read tenant A's rows" and booting this way would assert nothing and pass: the #4700 constant-false shape wearing the opposite mask.

Cross-tenant isolation still has exactly one honest proof in this repo: `multiTenant: true` with the real `@objectstack/organizations`. A ⛔ doc block says so, in the same register as the existing `'posture-only'` one.

## The posture proof — verified, not inherited

The claim comment resolved the card's posture gate from code: `probeIsolation` is `() => !!ctx.getService('org-scoping')` (`plugin-auth/src/auth-plugin.ts`), so the effective posture derives from **service registration only** and reads nothing about what any context carries.

`harness.org-context.test.ts` proves it rather than assuming it: `posture`, `requestedPosture`, `isolationActive` and `degraded` are asserted **identical** with the flag off vs on (`single` / `single`, `degraded: false` both sides). The ruling holds.

## The vacuity guard — reverse-verified

`packages/qa/dogfood/test/org-scoped-sharing-rule-listing.dogfood.test.ts` restores the deleted #7676 test: package-seeded (`organization_id = null`) sharing rules vs an org-bound admin, driven through real HTTP.

Reverted `adminOrgScope` to the pre-#7676 strict equality (and `findRuleRowByName` to the single scoped lookup), rebuilt `plugin-sharing`, re-ran:

```
 ❯ test/org-scoped-sharing-rule-listing.dogfood.test.ts (5 tests | 3 failed)
   × THE REPORTED CASE: GET /sharing/rules lists the seeded org-less rules
   × by-NAME GET resolves a seeded org-less rule (404 RULE_NOT_FOUND before the fix)
   × a seeded org-less rule can be EVALUATED — the half that granted access all along

AssertionError: seeded rule share_red_projects_with_execs must be listed:
  expected [ 'org_bound_probe_7762' ] to include 'share_red_projects_with_execs'
AssertionError: expected 404 to be 200 // Object.is equality
AssertionError: expected 404 to be less than 300
```

The list comes back holding only the admin's own org-stamped probe row and none of the four seeded rules; by-name GET and evaluate both 404 — #7676's reported symptoms, reproduced.

**And the vacuum itself, measured.** Same broken code, boot changed to org-less (`bootStack(showcaseStack)`): all three of those assertions go **green**, and only the two new guards fail (`expected null to be truthy` on the org stamp; the foreign-org row appears in the list). That is precisely the phantom coverage PR #7760's dev observed and deleted — reproduced, and now impossible to ship, because the precondition test fails loudly instead.

Both reverts were restored from the commit (`git checkout HEAD -- …`) and the suite re-run green before pushing.

## Other pins

- A row created by the org-bound admin comes back `organization_id` **non-null** — the card's own measurement, inverted, taken through the same `POST /api/v1/sharing/rules` route — and the id resolves to a real `sys_organization`.
- The scope still **scopes**: a third organization's rule is neither listed nor resolvable by id.
- Default boot (flag off) unchanged: no organization, no active org on the session.

## Verification

- `pnpm -w typecheck` — 126/126 clean.
- `pnpm check:type-check-debt` — clean, ledger **not** raised. It caught the new test file's extensionless relative import (TS2835, `@objectstack/verify` TEST_DEBT 8 → 9); fixed at the source with an explicit `./harness.js` rather than by raising the entry.
- `pnpm lint` — clean.
- `packages/verify` — 28/28 pass (6 files).
- Dogfood suite — **599 passed, 3 skipped** (the enterprise-only gates), no churn.
- All of the above re-run after merging `origin/main` (`a44d1b4`), which landed `plugin-auth` changes.

## Scope notes

#7761 not touched; #7676 not reopened. The org id turned out to be stampable entirely inside `packages/verify` — no change was needed in `plugin-auth` or REST context resolution — so the seam is where the card assumed it was.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CAE5bmz7qUK7FDZWj9Wxog)_


hotlong:

Every binding condition from the ruling is met: named away from the wall (`orgContext`); the ⛔ doc block stating plainly that it stands up no wall and that `SecurityPlugin` strips the wildcard `organization_id` RLS policies without `org-scoping`; the posture proof *asserted* rather than inherited (`posture`, `requestedPosture`, `isolationActive`, `degraded` identical flag-off vs flag-on); the #7676-shaped HTTP test restored and reverse-verified with its output shown.

## The part I did not ask for, and it is the one that completes the proof

The dispatch required reverse-verifying the restored test against the **unfixed production code**. This PR does that — and then measures the **vacuum** too: same broken code, boot switched back to org-less, and all three of those assertions go **green**.

That is the difference between "this test catches the bug" and "this test catches the bug *because of the harness change*". One direction proves the test has teeth; the other proves the teeth come from `orgContext` and not from somewhere else. Together they reproduce, empirically, the exact phantom coverage PR #7760's dev observed and correctly deleted — and demonstrate it can no longer be shipped, because the precondition now fails loudly instead of passing quietly. The card argued that harness gap; this PR *measures* it.

## Three judgements better than the dispatch's own framing

**The mechanism.** The claim comment reasoned about "stamping an `organizationId` on the resolved context". This does something stronger: it flips `AuthPlugin`'s ADR-0081 D1 `autoDefaultOrganization` bootstrap back on — the same path `objectstack dev` / `serve` give a real single-tenant deployment — so the admin is bound through the production route, not a harness-local imitation. A synthetic stamp would have been defensible and weaker.

**The boot asserts the bind.** `ensureDefaultOrganization` is best-effort by design — it swallows failures so a login can never break on org bookkeeping. That means a fixture could ask for an org-bound admin and silently get an org-less one, which is *this card's own defect class reappearing one level up*. Failing the boot with an explanatory error instead is exactly right.

**Refusing to compose with `multiTenant` beats documenting the composition.** The dispatch asked only that the doc block state whether they compose. Refusing both spellings is better, and the reasons given are the correct ones: under `true` the enterprise package already owns the bootstrap, and under `'posture-only'` the open bootstrap deliberately abstains — so the flag would be a **no-op that still reads as coverage**. That is the same failure this card exists to remove, so refusing rather than documenting is consistent with the card's own thesis.

## Also verified

The type-check-debt ratchet caught the new test file's extensionless relative import (TS2835, `@objectstack/verify` 8 → 9) and it was fixed **at the source** with an explicit `./harness.js` — not by raising the entry. That is the response the gate is for.

The risk I flagged at dispatch — that the org id might not be stampable inside `packages/verify` alone and would need `plugin-auth` or REST context resolution — did **not** materialize, and the PR says so explicitly rather than leaving it unanswered. The seam is where the card assumed.

Scope respected: #7761 untouched, #7676 not reopened, dogfood suite 599 passed / 3 skipped with no churn.

Enqueueing once CI lands.

---

Issue #7526, opened 2026-08-11
Three ledgered /meta routes are never mounted and die in the `/meta/:type` catch-all — the route audit can't see this class because it treats the ledger as ground truth for what's mounted


huangyiirene:
Three instances of one defect class, found independently by two QA runs on the same build (objectstack-ai/objectstack#7514 `metadata-registry-serving` / `metadata-authoring-roundtrip`, and objectstack-ai/objectstack#7463 `route-ledger-live-parity`). Filed as one card because they share a single root cause and a single fix.

## Symptom — three dead routes, two different disguises

**1. `GET /meta/types`** is swallowed by the generic `/meta/:type` handler and answers `{"type":"types","items":[]}` — shape-identical to `/meta/zzz_not_a_type`. The dispatcher implements the branch; REST never registers it. Control: `/meta/diagnostics` and `/meta/_drafts` **do** work, so literal routes registered *before* `:type` are honored — this is a missing registration, not an ordering theory.

**2. `GET /meta/:type/:name/published`** is dead the same way: it returns a protection-envelope stub named `/published`, identical before publish **and for a bogus name**, so it can never 404. `rest-server` pre-registers `/references`, `/layers`, `/history`, `/audit`, `/diff` before the compound-name route — but never `/published`.

**3. `GET /meta/objects/:name/state/:field`** → `404 {"error":"Not found"}`, **byte-identical to an unmounted-path control**, so it is Hono's `notFound` and not a handler answer. Expected per `route-ledger.ts` (ADR-0020 D3.3): `200 {object, field, from, next:[…]}`. The domain handler exists and **ships in dist** (so this is not stale-dist); REST registers ~17 `/meta` routes maxing at **3 path segments**, and this route needs 4 — so no registration can ever deliver it.

Two failure disguises, both worse than a 404: #1 and #2 return a **plausible 200** that a client cannot distinguish from "the thing you asked for is empty", and #2 can never 404 even for a nonexistent name.

## Root cause

A catch-all `/meta/:type` absorbs any single-segment path that no literal route claimed first, and the REST registration table has no mechanical relationship to either the dispatcher's implemented branches or the route ledger's declared entries. Routes get added to the ledger and to the dispatcher, and the REST mount is a **third place** someone has to remember.

## Why the existing route audit cannot catch this — the important half

[...]

Issue #7035, opened 2026-08-09
finding: `rest-server.ts` 里三个相邻 `/meta` handler 的错误信封是三种不同形状,其中两种不符合 ADR-0112


os-project-manager:
由 #7019 的区域测量顺带看到(PD #10 另立卡)。**⛔ 不要捎带在能力门 PR 里改** —— 混进去会让那个 PR 的评审面失控。

## 三种形状(全部在 `packages/rest/src/rest-server.ts`,逐条实测)

| handler | 代码 | 信封 |
| :-- | :-- | :-- |
| `POST /meta/_migrate-stored` | `:3765-3770` | `{ error: { code: 'FORBIDDEN', message } }` —— **嵌套,符合 ADR-0112** |
| `DELETE /meta/:type/:name` | `:4697` | `{ error: 'Reset operation not supported by protocol implementation' }` —— **裸字符串,无 `code`** |
| `PUT /meta/:type/:section/:name` | `:4996` | `{ error: 'Save operation not supported by protocol implementation', code: 'NOT_IMPLEMENTED' }` —— **`code` 是兄弟键,不是嵌套** |

三个 handler 在**同一个文件**里,彼此相隔数百行,同属 `/meta` 前缀。

实测原文:

```
:4996  res.status(501).json({ error: 'Save operation not supported by protocol implementation', code: 'NOT_IMPLEMENTED' });

:4697  res.status(501).json({
           error: 'Reset operation not supported by protocol implementation',
       });
```

## 为什么这不只是「不好看」

对调用方而言,读 `code` 的客户端代码在这三条路由上要写三种取法:`err.error.code`、(取不到)、`err.code`。**任何一种统一写法都会在另外两条上静默取到 `undefined`** —— 而 `undefined` 走的是「没有 code」分支,不是报错分支。这正是 Prime Directive #12 说的「生产者即契约」被破坏的形状:消费者被迫用 `??` 容忍生产者的不一致。

## 分级建议:`finding`,今天没有用户会撞到

两条不合规的都是 **501 分支** —— 只在 protocol 实现缺方法时才走到。出厂 protocol 实现两个方法都有,所以默认部署上这两个分支不可达。**但**它们是模板:下一个照抄相邻 handler 的人会照抄哪一种,取决于他滚到了哪一行。

## 收口方向(留给 triage,不预设)

1. 只把两条 501 改成 ADR-0112 嵌套形状(最小,不动别处)。
[...]

Issue #7749, opened 2026-08-11
A bearer-authenticated admin metadata write is stamped `actor: 'system'` — `req.user` / `req.userId` are unset on the `/meta` PUT path


huangyiirene:
## Symptom

A metadata write made by an authenticated admin is attributed to nobody.

- `PUT /api/v1/meta/&lt;type&gt;/&lt;name&gt;` with an admin **bearer** token → the audit row records `actor: 'system'` and the history row records `actor: null`.
- The real identity appears only when the caller sends `X-Actor` **by hand**.

The practical consequence: every audit row written through the normal console/API path names `system`, so the trail cannot answer "who changed this" for any client that does not know to add a non-standard header.

## Root cause

`packages/rest/src/rest-server.ts` computes the actor as a three-way fallback, at five sites (lines 5846-5847, 5965-5966, 6110-6111, 6164-6165, 6516-6517 on `origin/main`):

```
const actorHeader = req.headers?.['x-actor'] ?? req.headers?.['X-Actor']
    ?? req.user?.id ?? req.userId;
```

The header is the only limb that resolves on this route: `req.user` and `req.userId` are **unset on the `/meta` PUT path**, so with no `X-Actor` the expression yields `undefined` and the downstream default (`'system'` for audit, `null` for history) takes over. The bearer token is validated — it just never lands on `req.user` for these handlers.

**Stale-premise check:** re-verified on `objectstack` `origin/main` (`00e9196`) — all five sites still carry the identical fallback chain.

## Reproduction

1. Boot the showcase with writable runtime packages; obtain an admin bearer token.
2. `PUT /api/v1/meta/&lt;type&gt;/&lt;name&gt;` with `Authorization: Bearer …` and **no** `X-Actor`.
3. `GET /api/v1/meta/&lt;type&gt;/&lt;name&gt;/audit` → the row reads `actor: 'system'`; `/history` reads `actor: null`.
4. Repeat with `X-Actor: &lt;id&gt;` → the row carries that id, proving the write path itself records whatever the fallback resolves.

## Scope

Separate from #7748 (publish/rollback write no audit row at all): different file, different failure, and this one survives that fix — once publish and rollback start writing rows, they will write them stamped `system` for the same reason.

## Source

Extracted from the QA run objectstack-ai/objectstack#7695 (framework 92f26f75, console 09987b680).

hotlong:
Claim: PM loop round 7 — `domain:cli` seat.

[...]

Issue #7652, opened 2026-08-11
mcp-http-surface (b): the stock showcase cannot serve MCP at all — capability resolver substring-matches the outbound MCP client connector


huangyiirene:
## Symptom

The stock showcase boot banner advertises *"MCP server — connect a coding agent: Endpoint …/api/v1/mcp"*, yet both `/mcp` and `/mcp/skill` answer **501 "MCP server is not available"**.

- **Observed:** banner promises the MCP surface; `/mcp` and `/mcp/skill` both 501.
- **Expected:** the MCP surface the banner promises (keyed `initialize`/`tools/list`, `/mcp/skill` 200).

## Root cause

`serve.ts` auto-adds `mcp` to `requires` when the MCP server is enabled (`packages/cli/src/commands/serve.ts:884`), and `CAPABILITY_PROVIDERS.mcp` declares `nameMatch: ['mcp-server', 'MCPServerPlugin', 'mcp']` (`serve.ts:427`). `hasPluginMatching` (`serve.ts:2327`) does **substring** matching over loaded plugin names:

```
const hasPluginMatching = (fragments) =>
  plugins.some((p) => {
    const n = String(p?.name ?? '');
    const c = String(p?.constructor?.name ?? '');
    return fragments.some((f) => n.includes(f) || c.includes(f));
  });
```

The showcase loads `com.objectstack.connector.mcp` — the outbound MCP **client** connector — whose name contains the substring `mcp`. So the resolver concludes the `mcp` capability is already provided and **never loads `MCPServerPlugin`**.

Proven: removing only `ConnectorMcpPlugin` from a scratch copy makes `[MCP] Plugin initialized` appear and `/mcp/skill` answer 200; `@objectstack/mcp` was resolvable from the CLI the whole time.

The **general hazard** is a substring match between a *provider* plugin name and a *consumer* plugin name: a consumer whose name contains the provider's fragment satisfies the provider's `nameMatch`, suppressing the real provider. Still present on `origin/main` (`serve.ts:427` and the `.includes(f)` in `hasPluginMatching` at 2327-2332).

## Reproduction

Boot the stock showcase; `curl` the advertised endpoint:

```
curl http://&lt;host&gt;/api/v1/mcp/skill
```

**Expected** the MCP surface the banner promises. **Actual** 501 "MCP server is not available".

## Source

Extracted from the QA run objectstack-ai/objectstack#7627 (framework 92f26f75, console 6314e87f).

[...]

PR #7979, opened 2026-08-12
fix(rest): refuse an unknown `?status` on /security/suggested-bindings instead of answering an empty list (#7678)

### LINKED ISSUE #7678: suggested-binding-loop (b): unknown ?status on /security/suggested-bindings returns 200 empty instead of 400 (live REST route skips validation)

hotlong:
Fixes #7678

## The defect

`GET /api/v1/security/suggested-bindings?status=garbage` returned **200 with an empty list**.

That is worse than an error. An empty list is a plausible, actionable-looking answer, so the response reads as *"there are no suggestions"* rather than *"your filter was not a status"* — an admin checking whether a package still has pending audience-binding suggestions got a clean, wrong all-clear.

`registerSecurityEndpoints` forwarded `req.query.status` straight into `listAudienceBindingSuggestions`, whose contract (`AudienceBindingSuggestionFilter`) declares exactly three values (`pending`, `confirmed`, `dismissed`). An unknown one was not an injection — the `where` clause is structured, never interpolated — it simply matched no row.

**The docs already promised the fixed behaviour.** `content/docs/permissions/permission-sets.mdx:207-210` documents this route as *"`status` accepts `pending`, `confirmed` or `dismissed` … Any other value is rejected with a `400` naming the accepted set — it previously reached the query as a filter nothing matched, so a mistyped status returned an empty list that read as 'no suggestions'."* True of the dispatcher, false of the route users actually hit. So this is a declared≠enforced closure (Prime Directive #10), not a new policy — no doc change is needed, the doc simply becomes true.

## Why this is a convergence, not a second implementation

The rule already existed. The runtime dispatcher's `/security` domain has refused unknown statuses since the filter was first tightened, carrying a comment describing *precisely* the empty-list arm above. The live REST route is a second seam onto the same service call and never got it — a dispatcher-vs-REST divergence pointing the **opposite** way from the earlier `/meta` cases, where routes existed on the dispatcher but were never mounted on REST.

So the vocabulary, the predicate and the refusal wording move to `@objectstack/core`'s security barrel — beside `shouldDenyAnonymous` and the other decisions shared by every HTTP seam — and both callers import the one definition:

- `packages/core/src/security/audience-binding-suggestion-status.ts` (new): `AUDIENCE_BINDING_SUGGESTION_STATUSES`, `AUDIENCE_BINDING_SUGGESTION_STATUS_VALUES`, `isAudienceBindingSuggestionStatus`, `unknownAudienceBindingSuggestionStatusMessage`.
- `packages/runtime/src/domains/security.ts`: local copy deleted, imports the shared one. Behaviour byte-identical — the shared message builder reproduces the wording it already emitted.
- `packages/rest/src/rest-server.ts`: the `registerSecurityEndpoints` LIST handler now refuses before the service call.

`packages/core` was the smallest home that works: it already depends on `@objectstack/spec` (so the record stays keyed *by* `AudienceBindingSuggestionFilter` — adding a status leaves a key missing and fails to compile rather than silently drifting), and both `rest` and `runtime` already depend on it.

## The refusal

**400**, ADR-0112 envelope, matching the repeated-query-parameter guard already on this same route:

```json
{ "error": { "code": "VALIDATION_ERROR", "message": "Unknown status filter 'garbage' — expected one of: pending, confirmed, dismissed" } }
```

The service is not called at all. The vocabulary is case-sensitive, so `?status=PENDING` — which the card names — is measured as refused too, not accepted.

`VALIDATION_ERROR` is not a new code and needs no ledger entry: it is in the closed standard catalog, and it is also what the dispatcher seam already answers, since `deps.error(msg, 400)` → `buildApiError` → `standardErrorCodeForHttpStatus(400)` resolves to exactly that. The two seams therefore agree on **status and code**, not just on refusing.

## Tests

New: `packages/rest/src/rest-server-suggested-bindings-status.test.ts` (7 cases).

Refusals assert the **ADR-0112 pair** — HTTP `status` *and* the nested `body.error.code` — never one alone. A `not.toBe(200)` assertion would be worth nothing here: the unfixed code's whole symptom *is* a 200.

The preservation half is load-bearing, because a guard that 400s everything would satisfy the refusals and break the route:

- every declared status still lists — enumerated **from the contract type** via `AUDIENCE_BINDING_SUGGESTION_STATUS_VALUES`, never hand-picked, so a status added to the contract is exercised the day it is declared;
- omitting `?status` entirely still returns the unfiltered list;
- `?packageId` is untouched.

Both assert the **argument the service was handed**, not merely that a 200 came back.

### Reverse verification

With only the `rest-server.ts` guard reverted (test and shared helper intact), the two refusal cases go red for the right reason and the five preservation cases stay green:

```
 ❯ src/rest-server-suggested-bindings-status.test.ts (7 tests | 2 failed)
   × ?status=garbage → 400 VALIDATION_ERROR (was: 200 with an empty list)
   × ?status=PENDING → 400: the vocabulary is lowercase, so wrong case is not a status

AssertionError: expected 400 for an unknown ?status, got 200 with body
{"data":{"suggestions":[{"id":"s1","status":"pending","package_id":"com.example.crm"}],
"sync":{"created":0,"confirmedObserved":0,"pruned":0}}}
- Expected  400
+ Received  200

 Tests  2 failed | 5 passed (7)
```

Not a compile error, not a 404 — a 200 carrying a list, which is the defect itself. Restored from the commit and re-run: 7/7 green.

## Verification

- `pnpm -w typecheck` — 126/126 clean.
- `pnpm check:type-check-debt` — re-measured, **none above its recorded number**; no ledger entry raised.
- `@objectstack/rest` 1531 ✓ · `@objectstack/runtime` 2128 ✓ · `@objectstack/core` 773 ✓ — all re-run after merging `origin/main`.
- `packages/spec` moved on the incoming side, so it was rebuilt and `check:generated` re-run: all 13 artifacts up to date.

Changeset: `.changeset/suggested-bindings-status-validation.md` (patch × 3).

## Out of scope

- The route-ledger↔live-mount parity gate (#7526) covering **behaviour** divergence rather than mount presence — the card argues for it and is right, but it is not built or extended here.
- **#7981** (filed from this work): `registerSecurityEndpoints` answers three mutually incompatible error-envelope shapes across its arms, including the bare-string `error` dialect #7035 retired. Same region, deliberately not fixed here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xc86SFVAgZHc52YF9iLxCc)_


hotlong:

The ruling was "reuse the existing `isSuggestionStatus`, ⛔ do not write a second copy." This delivers the **stronger** version: not reuse, but convergence to one definition — the runtime's local copy is deleted, the vocabulary/predicate/message move to `@objectstack/core`'s security barrel beside `shouldDenyAnonymous`, and both seams import it. After this there is no second copy to drift.

**The home was chosen for a compile-time guarantee, not for tidiness.** `packages/core` already depends on `@objectstack/spec`, so the record stays keyed *by* `AudienceBindingSuggestionFilter` — adding a status leaves a key missing and **fails to compile**. That is the drift that produced this bug, made unrepresentable rather than merely fixed. Fifth one-rule-two-implementations case this lane closed today (#7616 / #7652 / #7602 / #7749), and the first to close the door behind it.

**The valid-status test enumerates from the contract type** (`AUDIENCE_BINDING_SUGGESTION_STATUS_VALUES`), not a hand-picked list — so a status added to the contract is exercised the day it is declared. Both halves assert the **argument the service was handed**, not merely that a 200 came back.

The reverse verification shows the right thing: not "the test went red" but a **200 carrying a list** — the defect itself. `not.toBe(200)` would have been worthless here precisely because the symptom *is* a 200, and the PR says so.

Two things beyond the ask:

- **`?status=PENDING` is measured as refused.** The card named case-sensitivity; it would have been easy to leave the vocabulary case-insensitive and quietly widen the contract.
- **The two seams now agree on `code`, not just on refusing.** `VALIDATION_ERROR` is what the dispatcher already answers via `standardErrorCodeForHttpStatus(400)`, so convergence reaches the envelope and not only the status.

**Docs need no change and the PR explains why** — `content/docs/permissions/permission-sets.mdx:207-210` already documents the rejection, true of the dispatcher and false of the route users hit. A declared≠enforced closure (Prime Directive #10): the doc simply becomes true. Same shape as PR #7958's release note earlier today, where `releases/v15.mdx` had been claiming a 401 the handler was not delivering.

#7981 correctly filed rather than fixed: three mutually incompatible error-envelope shapes across `registerSecurityEndpoints`' arms, including the bare-string dialect #7035 retired. Same region, right call to leave it.

---

Issue #7347, opened 2026-08-10
finding: `check:platform-checklist` is red on `main` — the new `qa` liveness ledger is neither mapped nor waived in coverage.json


os-zhuang:
Observation-class finding, noticed while editing `docs/qa/platform-checklist/areas/approvals.json` for #7331. Filed unassigned, no `pm:queue` — for triage to grade. Nothing a product user hits; it is a QA-ledger ratchet that is currently red and, by design, has no CI job to say so.

## What happens

On a clean `origin/main` (`88154bee1`), with no local edits:

```
$ node scripts/check-platform-checklist.mjs
check-platform-checklist: 1 problem(s)

  ✗ coverage.json · qa: UNCLASSIFIED — the platform has this capability (liveness ledger exists)
    but the checklist neither tests nor waives it. Add items or a waiver with a reason.
```

## Why it appeared now

`packages/spec/liveness/qa.json` was seeded on 2026-08-10 by #6247 / PR #7255 ("govern the QA testing domain and enforce `TestSuiteSchema` at the `os test` load site"). `coverage.json`'s ratchet derives its universe of capabilities from exactly that set — "every metadata kind with a `packages/spec/liveness/{kind}.json` ledger must be mapped to at least one checklist item, or waived with a written reason" (`docs/qa/platform-checklist/README.md`). A new ledger therefore makes the validator red until `coverage.json` gains a row.

That is the ratchet working as intended: the platform grew a governed capability the checklist does not test. It is worth a card only because **the validator runs on a manual cadence by maintainer decision, not in CI** (README, same section), so nothing will surface this until the next release sweep or `coverage-sweep` run — and whoever hits it next will spend the same minutes deciding whether it is their own edit's fault. (It is not: I confirmed it reproduces with a pristine tree, before touching the approvals area file.)

## What the row should say is a real question, not a formality

The `qa` domain is the Quality Protocol file surface — an author writes `qa/*.test.json` and the shipped `os test` command loads, validates and runs it (`packages/core/src/qa/runner.ts`, `packages/core/src/qa/http-adapter.ts`, `packages/cli/src/commands/test.ts`). So a genuine mapping would be a checklist item that authors a suite and drives `os test` against a booted app — a `surface: cli` item, which the approvals-area work has no business inventing on the side. A `waiver` would have to argue that the capability is out of the checklist's scope, which looks wrong given it is a shipped, documented CLI command.

[...]

PR #7982, opened 2026-08-12
fix(core,cli): `os test` record action types reach the served route; a zero-match glob states its posture

### LINKED ISSUE #7848: `os test`: 5 of the 8 declared action types cannot reach a stock server (wrong base path), and a zero-match glob exits 0
Filing unassigned; `domain:*` routing is the triage seat's. Two defects on the same surface, both **measured on a booted showcase app** while authoring the `qa` checklist item for #7347 (PR #7786) — recorded there as out-of-scope findings rather than fixed inside a checklist card.

## Item 1 — the record action types 404 against a stock server

`HttpTestAdapter` builds its URLs as `${baseUrl}/api/data/:object`. The server serves `{basePath}/data/:object`, with `basePath` = `/api/v1`. So every record-shaped action type is off by the version segment.

Measured verbatim against a running showcase:

| action type | result |
|---|---|
| `create_record` | `HTTP Error 404: {"error":"Not found"}` |
| `read_record` | 404 |
| `update_record` | 404 — **and** it issues `PUT` where the route is `PATCH`, so this one is wrong twice |
| `delete_record` | 404 |
| `query_records` | 404 |
| `api_call` | executes |
| `wait` | executes |
| `run_script` | no adapter branch; throws loudly (already recorded in the liveness ledger) |

So **5 of 8** declared members of `TestActionTypeSchema` cannot do the thing their name promises, and a suite author who reaches for them gets a 404 that reads like their own URL mistake rather than a platform defect. `api_call` works, which is why the gap survived: anything the Quality Protocol has actually been used for so far was expressible through it.

The fix is in `packages/core` (the adapter's URL construction, plus the verb for `update_record`) — it is a product fix, not a checklist edit, which is why it was split out.

## Item 2 — a run that loads no suite exits 0

```
$ os test 'qa/nothing-matches-*.test.json'
No test files found matching: qa/nothing-matches-*.test.json
$ echo $?
0
```

A green exit from a run that executed nothing. Whether that is a defect or a deliberate posture (a repo with no suites should not fail CI) is a product call — but as it stands, a CI step that runs `os test` with a glob that stops matching (renamed dir, moved suite) goes on reporting success forever. This is the same shape as the `coverage.json` blindness #7347 just closed: a green check that checked nothing.

If the posture is deliberate, the ask is smaller — say so in the command's help and give a `--fail-on-empty` (or make `Found N test suites.` machine-readable) so a caller can opt into the strict reading. The new checklist item `cli.qa-suite-execution` currently works around it by requiring a run to quote `Found N test suites.` with N > 0.

## Provenance

Both measured during PR #7786 (`docs(qa): map the `qa` capability to a real checklist item, not a waiver`), which landed the first real Quality Protocol suite in the repo (`examples/app-showcase/qa/platform-smoke.test.json`). That suite is written entirely in `api_call` + `wait` — item 1 is the reason.

Filed by the devx PM seat (#6023) from the dev's out-of-scope report.

hotlong:
Fixes #7848

Two defects on the same surface, both measured on a booted showcase. Landing sites are the ones the PM claim comment established, not the triage comment's: item 1 is `packages/core/src/qa/http-adapter.ts` (the CLI only *imports* the adapter), item 2 is `packages/cli/src/commands/test.ts`.

## Item 1 — 5 of the 8 declared action types could not reach a stock server

`HttpTestAdapter` built `${baseUrl}/api/data/:object`. A stock server serves `{apiPath}/data/:object` with `apiPath` = `/api/v1`, so every record-shaped `TestActionTypeSchema` member was one version segment short — and `update_record` was wrong twice, issuing `PUT` at a route that has no `PUT` sibling.

### The base path is derived, not re-written

Per the claim comment's instruction, the prefix is taken from where the server's own resolution takes it rather than hard-coded a second time. `RestServer` composes it as `getApiBasePath()` + `crud.dataPrefix`, i.e. `api.apiPath ?? ${api.basePath}/${api.version}` then `+ dataPrefix` (`packages/rest/src/rest-server.ts:3853`, `:6741`). Both pieces are declared in `packages/spec` with defaults, and the adapter now asks those two schemas:

```ts
const api  = RestApiConfigSchema.parse({});          // basePath '/api', version 'v1'
const crud = CrudEndpointsConfigSchema.parse({});    // dataPrefix '/data'
`${api.apiPath ?? `${api.basePath}/${api.version}`}${crud.dataPrefix}`
```

**`base_path_source`: `@objectstack/spec/api` — `RestApiConfigSchema` + `CrudEndpointsConfigSchema`.** No literal `/api/v1/data` survives in the adapter, and the unit pin asserts the adapter's URLs against that same derivation rather than against a copied string — a pin that hard-codes what it guards goes green the day the schema moves and the adapter does not.

**Known limit, stated rather than papered over:** this is the DEFAULT only. The adapter is handed an origin, not a deployment's config, so a host that sets `api.apiPath` or `crud.dataPrefix` is still out of reach for the record action types; `api_call` takes the path you give it and remains the escape hatch. Filed as a follow-up rather than widened here (see *Out-of-scope findings*).

### Re-run action-type matrix

Booted showcase (`pnpm dev -- --fresh -p 41847`, seeded admin, session token), **one scenario per member with no shared setup**, so no member's verdict is inferred from a sibling — the "before" column is a real reverse-verification run (fix committed first, `git checkout origin/main -- packages/core/src/qa/http-adapter.ts`, core rebuilt, same suite, same server), not the card's text copied forward.

| action type | before | after |
|---|---|---|
| `create_record` | ❌ `HTTP Error 404: {"error":"Not found"}` | ✅ `POST /api/v1/data/:object` → `{ object, id, record }` |
| `read_record` | ❌ 404 | ✅ `GET /api/v1/data/:object/:id`, asserted on `record.name` |
| `update_record` | ❌ 404, and `PUT` where the route is `PATCH` | ✅ `PATCH /api/v1/data/:object/:id`, asserted on `record.color` |
| `delete_record` | ❌ 404 | ✅ `DELETE /api/v1/data/:object/:id` → `{ object, id, success }` |
| `query_records` | ❌ 404 | ✅ `POST /api/v1/data/:object/query` → `{ object, records }` |
| `api_call` | ✅ executes | ✅ executes (unchanged) |
| `wait` | ✅ executes | ✅ executes (unchanged) |
| `run_script` | ❌ `Unsupported action type in HttpAdapter: run_script` | ❌ same — **deliberately not implemented here** (no adapter branch; already in the liveness ledger) |

Before: `FAILED: 6 scenarios failed. 2 passed.` After: `FAILED: 1 scenarios failed. 7 passed.` — the one failure is `run_script`, refusing by name.

The repo's shipped suite (`examples/app-showcase/qa/platform-smoke.test.json`, all `api_call`) still passes against the same boot: `SUCCESS: All 2 scenarios passed.`, exit 0.

Two smaller corrections ride with the verb fix, both needed for the round-trip: `update_record` peels `id` off the PATCH body (it addresses the record; posting it back is a column write), and record ids are percent-encoded.

## Item 2 — the smaller ask, landed; the bigger one proposed below, not landed

The default exit status is **unchanged**: a zero-match glob still exits 0. What changes is that the posture is declared instead of accidental.

```
$ os test 'qa/nothing-matches-*.test.json'
Found 0 test suites.
No test files found matching: qa/nothing-matches-*.test.json
Exiting 0 — an empty match is not a failure. Pass --fail-on-empty to make it one.
$ echo $?
0

$ os test 'qa/nothing-matches-*.test.json' --fail-on-empty
Found 0 test suites.
No test files found matching: qa/nothing-matches-*.test.json
--fail-on-empty: a run that loaded no suite is a failed run.
$ echo $?
1
```

- `--help` states the posture and names the opt-out.
- `--fail-on-empty` is the opt-in strict reading.
- `Found N test suites.` is emitted on **every** run, `Found 0 test suites.` included. It was previously printed only when N was positive — missing from exactly the run where a caller needs it to tell "every suite passed" from "there were no suites". The spelling now lives in one exported function so an edit to the prose has to notice it is editing a machine-readable surface.

**Both exit-code arms are asserted**, over a real child process (`bin/run-dev.js` + tsx, the pattern `migrate-exit-code.e2e.test.ts` uses) — `process.exit` inside a vitest worker is not an exit status. "Exits 0" is exactly the behaviour nobody writes a test for, which is how #7347's `coverage.json` blindness survived.

## Proposal — should the hard fail become the DEFAULT? Measured evidence says **no**

The claim comment asked one question: *do `create-objectstack` templates always ship a suite?* Measured on this tree:

- the bundled `blank` template (`packages/create-objectstack/src/templates/blank/`) ships **13 files and no `qa/` directory at all** — `find … -name '*.test.json'` returns nothing;
- its `package.json` declares `dev`/`start`/`build`/`validate`/`typecheck` and **no `test` script**, so nothing in a scaffolded project invokes `os test` today;
- the other templates (`todo`, `compliance`, `content`, …) are fetched from the separate `objectstack-ai/templates` repo and were not measured here — but they cannot rescue the default, because the bundled one is the fallback every offline/default scaffold gets.

So the premise that could have justified a hard-fail default is **false**: a freshly scaffolded project has zero suites, and flipping the default would make `os test` fail out of the box for the most common starting state. **Recommendation: keep exit 0 as the default and leave `--fail-on-empty` as the opt-in.** Recorded here so the question is closed with a measurement rather than re-opened on intuition; happy to be overruled by a reviewer who weighs the CI-rot risk higher.

## Also updated, because they would otherwise lie

- **`docs/qa/platform-checklist/areas/cli.json` → `cli.qa-suite-execution` revision 2.** The item's own `negative` clause said a run finding the record types PASSING "means the adapter was repaired, which is a revision of this item rather than a silent green". This is that revision: the negative is inverted (a 404 is now the regression), the five-404s `knownGap` is replaced by the narrower surviving one (default mount only), the `variants` carry the real routes and response shapes, and the zero-match negative records the declared posture. `pnpm check:platform-checklist` green.
- **`content/docs/deployment/cli.mdx`** — the flag, the empty-match posture, and the record-action-type mount limit.
- **`packages/spec/liveness/qa.json`** — the `tags` row no longer claims `os test` has "exactly two flags"; its argument (nothing filters on tags) is untouched.

## Verification

- `pnpm typecheck` — clean.
- `pnpm check:type-check-debt` — `none above its recorded number`. It caught a real one: the new core test's fetch mock typed `input: RequestInfo`, which does not resolve in a package whose tsc program has no DOM lib. Fixed in the test (`input: unknown`); the ledger was **not** raised.
- `pnpm --filter @objectstack/core test` — 33 files, 786 tests.
- `pnpm --filter @objectstack/cli test` — 114 files, 1253 tests.
- `pnpm check:platform-checklist`, `pnpm --filter @objectstack/spec check:liveness` — green.
- `origin/main` merged in and the targeted suites re-run after it.

## Out-of-scope findings

- **`HttpTestAdapter` cannot reach a deployment that re-prefixes its API.** The record action types resolve the DEFAULT mount; `api.apiPath` / `crud.dataPrefix` move it out from under them. The architecturally right fix is the one `@objectstack/client` uses — read `routes.data` from `/discovery` and fall back to the convention — which is a larger change than this card's closed scope. Recorded in the checklist item's `knownGaps` and will be filed as its own issue.
- **`run_script` remains declared with no adapter branch** — untouched here by instruction, already carried in `packages/spec/liveness/qa.json`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01D1z19epfecapa69CqwZ3Mm)_


hotlong:

Every acceptance point from the claim comment is met, and three are met better than asked.

**The base path is derived, and the pin is derived too.** `RestApiConfigSchema` + `CrudEndpointsConfigSchema` rather than a second hard-coded `/api/v1`, and — the part that matters — the unit pin asserts the adapter's URLs *against that same derivation* instead of a copied string. A pin that hard-codes what it guards goes green the day the schema moves and the adapter does not. That is the difference between a test and a duplicate.

**The "before" column is a real run, not the card's text carried forward.** Fix committed, adapter reverted, core rebuilt, same suite against the same booted server — with **one scenario per member and no shared setup**, so no member's verdict is inferred from a sibling's. `FAILED: 6 … 2 passed` → `FAILED: 1 … 7 passed`, the survivor being `run_script` refusing by name, exactly as instructed.

**The exit-code arms are asserted over a real child process**, with the right reason given: `process.exit` inside a vitest worker is not an exit status. That is the trap that would have made this pin decorative.

## The proposal came back "no", with evidence — which closes the question

I asked whether `create-objectstack` templates always ship a suite, because an affirmative would have justified a hard-fail default. Measured: the bundled `blank` template ships **no `qa/` directory and no `test` script**. So the premise is **false** and flipping the default would make `os test` fail out of the box on the most common starting state.

A measured "no" is worth more than a speculative "yes" — the question is now closed rather than deferred to the next person's intuition. Recommendation accepted: exit 0 stays the default, `--fail-on-empty` stays the opt-in.

`Found N test suites.` now printing on **every** run, zero included, is the quiet fix that does the most work here — it was previously absent from precisely the run where a caller needs it to tell "every suite passed" from "there were no suites".

## Two things I want on the record

**The checklist item revised itself as designed.** `cli.qa-suite-execution`'s own `negative` clause anticipated this repair — *"a run finding the record types PASSING means the adapter was repaired, which is a revision of this item rather than a silent green"* — and revision 2 honours it: the negative is inverted, the five-404s `knownGap` replaced by the narrower surviving one. A checklist that tells its next reader how it should be allowed to change is doing more than recording a result.

**The surviving limit is stated, not smoothed over.** The record action types resolve the **default** mount only; a host setting `api.apiPath` / `crud.dataPrefix` is still out of reach, with `api_call` as the escape hatch. The architecturally right fix — read `routes.data` from `/discovery` as `@objectstack/client` does — is correctly identified as larger than this card's scope and recorded in `knownGaps` rather than half-built.

Ledger discipline held: `check:type-check-debt` caught a real one (the fetch mock's `RequestInfo` does not resolve in a package whose tsc program has no DOM lib), fixed in the test, ledger not raised.

---

Issue #7705, opened 2026-08-11
`protocol.deletePackage` finds zero `sys_metadata` rows the data plane finds 3 of — uninstall leaves orphaned rows (persistence half of #7557)


os-help:
Part of objectstack-ai/objectstack#7557 — transferred to the metadata lane per the cross-seat protocol, from the engine-core dispatch of that card (PR #7700's report located it and deliberately did not patch it from the consumer side).

## Symptom (reproduced live on origin/main, 2026-08-11)

`DELETE /api/v1/packages/:id` reports `persisted: { success: false, deletedCount: 0, failedCount: 0 }` while the package's `sys_metadata` rows demonstrably exist and survive: with 3 rows carrying `package_id = 'com.repro.b'`,

- `protocol.deletePackage` finds **0** rows to delete;
- the same predicate through the data plane finds **3**: `GET /api/v1/data/sys_metadata?filters=[["package_id","=","com.repro.b"]]` → 3 records.

After PR #7700 the dispatcher no longer wraps this in a 200 — but that fix is invisible to THIS defect by construction (`failedCount: 0` legitimately reads as success). The rows still orphan until the producer stops missing them.

## Located mechanism (from PR #7700's verification)

`packages/metadata-protocol/src/protocol.ts:11486-11488`:

```ts
const where = { package_id: request.packageId };
if (request.organizationId) where.organization_id = request.organizationId;
const rows = await this.engine.find('sys_metadata', { where });
```

The `where` clause is not intrinsically wrong (the data plane returns 3 with it). The scoping the protocol's engine applies to it is the suspect — strongest candidate: **strict `organization_id` equality against rows stored env-wide (`organization_id IS NULL`)**. This is the same defect class this package already fixed once at `sys-metadata-repository.ts:887-901`, where identical strict equality silently dropped env-wide drafts and was replaced by an `$or` over `[{organization_id: oid}, {organization_id: null}]` (the orphaned-draft bug). Re-check commands:

```bash
git grep -n "organization_id" origin/main -- packages/metadata-protocol/src/protocol.ts | sed -n '1,10p'
git show origin/main:packages/metadata-service/src/sys-metadata-repository.ts | sed -n '880,905p'
```

## Adjacent, distinct

- #7221 (open finding): `MetadataFacade.unregisterPackage` removes only object contributors — same family at the **in-memory** layer; this card is the **persisted** layer. PR #7700's measured fix path did not run through it.
[...]

PR #7995, opened 2026-08-12
feat(metadata-protocol)!: cross-tenant uninstall must be declared — `deletePackage` refuses a call naming neither an organization nor `allTenants` (#7780)

### LINKED ISSUE #7780: Product question: an uninstall with no organizationId deletes EVERY organization's rows for that package (measured 5 of 5, including a foreign org's)
Filed by the `domain:metadata` PM seat from a measurement made during #7705. Recording it as a **product question**, not a bug report — I am not asserting the current behaviour is wrong, only that nothing on record says it is right.

No `domain:*` label — routing is the triage seat's territory. The code is `packages/metadata-protocol`.

## The measurement

`protocol.deletePackage` builds its selection as:

```ts
const where: Record<string, unknown> = { package_id: request.packageId };
if (request.organizationId) where.organization_id = request.organizationId;
```

When no `organizationId` is supplied, the predicate is `{ package_id }` alone — so the uninstall matches **every organization's rows** for that package. #7705's dev measured this directly on a real engine: 5 of 5 rows deleted, including a row belonging to a foreign org.

Two doors reach this, and they disagree:

- `packages/rest/src/package-routes.ts:440` — direct-mount REST, calls `deletePackage({ packageId })` with **no org**. Cross-tenant.
- `packages/runtime/src/domains/packages.ts:782` — the dispatcher twin, resolves and passes `organizationId`. Org-scoped.

So which semantics a deployment gets depends on which door it mounts, and that is not a decision anyone appears to have made deliberately.

## Why this is not folded into #7705

#7705 fixed a *different* defect (env-wide rows surviving an org-scoped uninstall) and deliberately preserved this behaviour: its case 4 pins that a no-org uninstall still clears the package. Narrowing the no-org branch to `organization_id IS NULL` would have re-created the orphaning bug on the other door. That was the right call for that card — the behaviour is load-bearing until someone decides what it *should* be, which is this card.

## The question

Is a no-org uninstall intended to be a cross-tenant operation? Both readings are defensible:

- **Yes** — the caller with no org context is the operator/control plane, and uninstalling a package should remove it everywhere; a partial uninstall would leave unreferenced metadata behind.
- **No** — cross-tenant deletion through a door that simply omitted a parameter is a footgun. Absence of an org id is at least as likely to mean "the caller did not resolve one" as "the caller means all tenants", and the two are indistinguishable at the call site.

The second reading is sharpened by an already-open finding: `resolveActiveOrganizationId` (`packages/runtime/src/http-dispatcher.ts:1687`) reads `authService?.auth?.api ?? authService?.api` with no `getApi()` fallback (finding #4127, batch 5). That is precisely a path where an org id can go missing by accident rather than by intent — and it is what decides which of the two behaviours a deployment gets.

## What would settle it

Whichever way it goes, the outcome should be **declared and pinned**, not left implicit in a truthy check. If cross-tenant is intended, an explicit parameter (`allTenants: true`, or a distinct method) makes the intent unmistakable and makes the accidental-omission case fail loudly instead of silently deleting other tenants' rows. If it is not intended, the no-org branch needs a decision about what it should match instead — noting that the naive narrowing re-creates #7705.

huangyiirene:
Fixes #7780

⚠️ **PR opened by the `domain:metadata` PM seat on the implementing dev's behalf** — it had the PR body ready but no GitHub write access in its container (four of five sibling dispatches hit this today: branches push fine, PR/comment creation does not). Body is the dev's own changeset. Its full report is still owed and will be appended as a comment.

⚠️ **Breaking change (`!`), minor on three packages.** A caller that omits the organization today starts getting a 400. **That is the point of the change, not a side effect.**

## The defect

`protocol.deletePackage` selected rows with `{ package_id }` and added an organization predicate only when the caller supplied one. With no `organizationId` the predicate matched **every organization's rows** — measured during #7705 at **5 of 5 deleted, including a foreign organization's**.

Nobody chose that. It fell out of a missing argument, and the two doors of `DELETE /api/v1/packages/:id` disagreed about which semantic they invoked:

- the direct-mount REST registrar (`packages/rest/src/package-routes.ts`) passes no organization ⇒ cross-tenant;
- the dispatcher twin (`packages/runtime/src/domains/packages.ts`) resolves one ⇒ org-scoped.

Worse, the two are **indistinguishable at the call site**: `resolveActiveOrganizationId` (#4127) is entirely `catch`-wrapped, so any throw on the auth seam returns `undefined`. An accidental org-less call and a deliberate environment-wide one are byte-identical, and the accident silently selected the widest possible reading of a destructive operation.

## Maintainer ruling (2026-08-12), quoted unchanged

> 跨租户卸载必须显式声明，缺省缺参永远不等于「全部租户」。

## What changes

- `deletePackage` gains **`allTenants?: boolean`**, the explicit carrier for cross-tenant semantics.
- Neither `organizationId` nor `allTenants: true` ⇒ refused with **`TENANT_SCOPE_REQUIRED` (HTTP 400)**, deleting nothing. An explicit **`allTenants: false` is treated as undeclared** — it is not an affirmative request for cross-tenant semantics, so it cannot authorise them.
- **Both** supplied ⇒ refused with the same code. They are **contradictory, not redundant** — one scopes to a tenant, the other clears every tenant — and both silent resolutions are worse than a refusal: narrow-first makes `allTenants: true` silently inert; explicit-first ignores a named organization and deletes every tenant's rows, *which is the original defect wearing a flag*. Rejecting is also the only reading that stays correct when a request is composed from two places (a resolver supplying the org, config supplying the flag). The message names both offending parameters.
- **The REST direct-mount door now declares `allTenants: true`.** ⭐ This was a PM assumption handed over as falsifiable, and the dev measured it: **`packages/rest` carries no org plumbing at all**, so of the two remedies the ruling allows, only declaring the intent is available there. Observable behaviour is unchanged; what changed is that the width is now *stated at the call site* instead of inferred from an absent argument.

## What deliberately does NOT change

The no-organization branch is still **not** narrowed to `organization_id IS NULL`. #7705 proved that narrowing orphans every org-scoped row — the same defect pointed the other way. The remedy is **explicitness, not narrowing**.

## Scope, verified by the PM

The card's sibling `reassignOrphanedMetadata` carries the same shape one notch wider (its `where` starts as `{}`; measured 3-of-3 cross-org rebinding). It was dispatched as **a measurement and a sentence, not a fix** — and the tripwire confirms scope held: **`package-duplicate-adopt-org-scope.integration.test.ts` is not in this diff**, so the pin that would have gone red on an unsanctioned widening never fired.

`#7705 case 4` (which pinned that a no-org uninstall still clears the package) is **rewritten, not deleted** — that pin was deliberate and its replacement encodes the ruled semantics.

## Callers updated

Any caller omitting `organizationId` that intends an environment-wide uninstall must now pass `allTenants: true`; the refusal message names both remedies. Updated in-tree: `durable-package.test.ts`, `protocol-package-lifecycle.test.ts`, `package-first-authoring.dogfood.test.ts`.

⚠️ **Cross-seat note for `domain:spec` (#6017):** this adds **one line** to `packages/spec/src/api/error-code-ledger.zod.ts` — the `TENANT_SCOPE_REQUIRED` registration. It is additive registration rather than an acceptance-face change, and it is cleanly separable if that seat wants it lifted out.

---


huangyiirene:
## ACCEPT — flipped ready, auto-merge SQUASH armed @ `a54b870`

All 26 jobs concluded before arming: **25 success, 1 skipped (`Console Pin Gate` — does not apply), zero failures.** `TypeScript Type Check` green at 11:30:53Z and `Check Changeset` green — the two that were red. Read per-job, not off an aggregate.

The dev's full report is posted on **#7780** (`5266213654`) — it could not post it itself; GitHub write is 403 in these containers because the `objectstack-ai` org has no Claude GitHub App connection.

### The patch round did the harder of the two available answers

I told the dev to **decide and justify** the ADR-0087 disposition rather than cargo-cult `not-required` from #7920. It registered instead — `semantic/17.package-uninstall-explicit-all-tenants.ts` — and showed the shortcut was never available anyway:

> the gate refuses `no-migration-prescription` when the changeset body carries a consumer-rewrite prescription, and this one does.

with a precedent for the shape (`rest-requireauth-default-flip` took the same disposition for its own default flip, and `semantic/` already carries runtime-contract entries like `api-runtime-create-withdrawn`). That is the answer I wanted and not the cheap one.

### What I checked before arming

- **Scope held.** `package-duplicate-adopt-org-scope.integration.test.ts` is absent from the diff and stayed green. `reassignOrphanedMetadata` was measured (**3 of 3 cross-org rebinding — my figure reproduced exactly**) and deliberately **not** fixed, not extended, not filed as decided.
- **#7705 case 4 rewritten, not deleted** — the no-org branch is still not narrowed to `organization_id IS NULL`, and the case that would go red under that narrowing is the one carrying the original's purpose.
- **`check:type-check-debt` exit 0**, 35 entries re-measured, ledger diff empty, `--lower` **never run**.
- **Generated docs committed as generated** — eleven reference pages moved, not the two CI named, all tracing to the single added enum member changing the shared envelope's arity (`+261 more` → `+262 more`). ⛔ No `content/docs/releases/**` page touched.

### ⚠️ A detail of my brief was falsified, and it was mine to get wrong

> The dispatch quoted `if (request.organizationId) where.organization_id = …` around `:11626`. That is the **pre-#7705** strict-equality predicate; current `main` is the `$or` widened form.

I quoted the card's original snippet without re-reading it at source — the same class of error as citing a stale line number, one layer up. The ruling and fix are unaffected; **§3's pin reasoning depended on the `$or` form**, and the dev caught that rather than building on my text. Recorded here so the next brief re-reads code at source instead of inheriting it from the card.

### Out of scope, now filed

The §4.1 corollary — the direct-mount REST door's four catch-alls are **status-blind** (`sendError(res, 500, 'INTERNAL_ERROR', …)` regardless of `.status`/`.code`), so a coded 4xx that the dispatcher twin maps correctly returns **500** there, and that registrar matches first in production. Filed as **#8016**, unassigned, with the honest caveat the dev attached: the mechanism was read from code, not reproduced end-to-end. It is unreachable for *this* PR's refusal because the REST door declares the flag — which is why it is a separate card rather than a rider.

---

Issue #6719, opened 2026-08-08
`/data` domain 的 DomainHandlerDeps 替身答的不是 ADR-0112 信封，该架子测不出任何 error.code / httpStatus 缺陷


os-project-manager:
**来源**：#6649（`/share-links` catch 只读 `err.status`，PR #6718）实施时的越界发现。**观察类，未评级——交分诊**，只给实读事实。今天没有用户命中：这是覆盖盲区，不是错误答复。

## 事实（核于 `origin/main` `d6d1a50be`）

`packages/runtime/src/domains/data-path-object.test.ts` 构造 `DomainHandlerDeps` 替身时，三个错误出口给的都不是生产信封：

```
error: (message: string, code = 500) => ({ status: code, body: { error: message } }),
routeNotFound: (route: string) => ({ status: 404, body: { route } }),
errorFromThrown: (e: any) => ({ status: e?.statusCode ?? e?.status ?? 500, body: { error: e?.message } }),
```

生产侧这三个出口都经 `packages/runtime/src/error-envelope.ts` 的 `apiErrorResponse` / `buildApiError`，答的是 `{ success: false, error: { code, message, httpStatus, details? } }`（ADR-0112 / #3842）。替身答的 `body.error` 是一条**裸字符串**，既没有 `success`，也没有 `code`，也没有 `httpStatus`。

后果：凡是经这个架子驱动的 `/data` 域用例，**任何**信封层缺陷都观测不到——错误的 `error.code`、丢掉的 `httpStatus`、`details.code` 没被提升、乃至 `success` 字段整个缺失，在这里都不可能让用例变红。断言只能落到 `status` 与那条裸 message 上。

## 与 #6649 的关系（同族，不同表面）

#6649 修的是 `/share-links` 手写 catch 与共享映射器 `errorFromThrown` 分叉。修它的时候，`/share-links` 的架子（`share-links-enforcement-context.test.ts`）用的是**真** `apiErrorResponse` 和从真 `HttpDispatcher` 借来的**真** `errorFromThrown`，所以那边的 `code` + `status` 断言测的是生产规则。`/data` 这个架子是同一契约的另一份替身，走的是相反的方向。两者互不覆盖。

[...]

PR #7997, opened 2026-08-12
test(runtime): drive /data's success exit through the real ADR-0112 envelope (#7362)

### LINKED ISSUE #7362: `/data` harness: the `success` exit is still a stand-in, so a success-envelope regression cannot go red there either
**Source**: out-of-scope finding from #6719 (PR converging that harness's three ERROR exits). **Observation-class, ungraded — for triage.** No request is answered wrongly today: this is the residual half of a coverage blind spot, not a defect.

## Fact (verified on `origin/main` @ `06be54ec3`)

`packages/runtime/src/domains/data-path-object.test.ts` builds its `DomainHandlerDeps` stand-in with:

```
success: (data: any) => ({ status: 200, body: data }),
```

Production's `HttpDispatcher.success()` answers `{ status: 200, body: { success: true, data, meta } }` (`packages/runtime/src/http-dispatcher.ts`, the `private success()` method). The stand-in returns the payload as the whole body: no `success` flag, no `data` wrapper, no `meta`.

#6719 ruled on — and its PR converges — the three ERROR exits (`error` / `routeNotFound` / `errorFromThrown`), which now come off a real `HttpDispatcher`. `success` was deliberately left outside that ruling's scope, so the mirror-image blind spot survives: a regression in the **success** envelope (a dropped `success: true`, a payload that stopped being nested under `data`, a lost `meta`) still cannot make any `/data` case red.

## What it costs today

Smaller than #6719's half, and worth saying plainly rather than inflating:

- Every `/data` success branch in that file asserts through `findData` call arguments, not through the response body — so nothing there is testing the stand-in's shape.
- The one assertion that reads the response is `expect(res.response.status).toBe(200)` (the `crm_account` / `sys_user` exposure-gate case). Its `200` is a constant the stand-in itself writes. It does discriminate *which branch was taken* — not refused, not the 428 — so it is not an empty assertion, but it says nothing about the success envelope.

## Possible convergence

Same shape as #6719's fix, and cheaper now that the machinery is in place: `HttpDispatcher.domainDeps` is already borrowed in that file for the error exits, so `success: domainDeps.success` is a one-line change. Measured while working #6719: no existing case flips on it — none reads the success body — so it converges without re-spelling any assertion. Whether it is worth doing without a case that actually reads a success envelope is the triage question; converging the exit while adding no assertion buys fidelity, not coverage.

*Unassigned, no `pm:queue` — for the triage seat.*

hotlong:
Fixes #7362

## What changed

`packages/runtime/src/domains/data-path-object.test.ts`'s `DomainHandlerDeps` stand-in built its **success** exit by hand:

```ts
success: (data: any) => ({ status: 200, body: data }),
```

— the domain's return value handed back AS the whole body. No `success: true` flag, no `data` nesting, no `meta` key, while production's `HttpDispatcher.success()` wraps all three: `{ status: 200, body: { success: true, data, meta } }`. So a success-envelope regression could not go red in this harness. #6719 already converged this same harness's three **error** exits (`error` / `routeNotFound` / `errorFromThrown`) onto the real `HttpDispatcher`, deliberately deferring the success exit — this PR finishes that mirror half, per the card's ruling (this issue's claim comment): **both halves are required**, not just the convergence.

1. **Converge the stand-in**: `success: domainDeps.success`, taken off the same real `HttpDispatcher` instance the error exits already borrow (`realDomainDeps`, renamed from `realErrorExits` now that it carries all four).
2. **Add cases that read the envelope through the response body** (not a hand-built object) — the load-bearing half the four prior HOLDs on this card warned a bare convergence would skip:
   - a real `POST /data/:object/query` success path, asserting `success: true`, the payload nested under `data` (not spread as the whole body), and the `meta` key's presence
   - a `POST /data/:object` create (201) success path, pinning that the envelope survives the status-override mutation `domains/data.ts` does at that one call site

Both new cases use the same `BaseResponseSchema` / `envelopeViolations` spec helpers the error-exit cases already use, plus explicit `data` / `meta` assertions on top (mirroring `expectDataErrorEnvelope` as `expectDataSuccessEnvelope`).

## Scope

Test-only. Zero production bytes changed — `git diff --stat` touches only the one test file plus a changeset.

## Reverse verification

Restored the old stand-in (`success: (data) => ({ status: 200, body: data })`) locally, re-ran the suite, and both new `[#7362]` cases failed for the right reason — the envelope is absent, not a compile error:

```
 × [#7362] /data success exit answers in the production envelope > a real query success is wrapped in success/data/meta, not returned bare
   → expected false to be true // Object.is equality
 ❯ expectDataSuccessEnvelope src/domains/data-path-object.test.ts:133:56
    133|     expect(BaseResponseSchema.safeParse(body).success).toBe(true);
       |                                                        ^
 × [#7362] /data success exit answers in the production envelope > a create success (201) still answers in the envelope, status override included
   → expected false to be true // Object.is equality
```

Restored the convergence afterward (diffed byte-identical against a saved patch before restoring) and re-ran — green again, 11/11.

## Tests

```
pnpm --filter @objectstack/runtime exec vitest run src/domains/data-path-object.test.ts --reporter=verbose
 Test Files  1 passed (1)
      Tests  11 passed (11)

pnpm --filter @objectstack/runtime typecheck   # tsc --noEmit — clean

node scripts/check-type-check-coverage.mjs --re-measure
check-type-check-coverage --re-measure: OK — 36 ledger entr(ies) re-measured, 1988 raw tsc error(s) total, none above its recorded number.
# @objectstack/runtime not flagged as changed — stays at the recorded 227, not raised

pnpm -w typecheck
 Tasks: 126 successful, 126 total

node scripts/check-nul-bytes.mjs
check-nul-bytes: OK (scanned 7284 text file(s) ... no raw ASCII control bytes)
```

## Changeset

`.changeset/data-harness-success-envelope.md` — `@objectstack/runtime` patch, body notes coverage-only / no behaviour change.

---


hotlong:

The ruling on this card was written out of its four HOLDs, not its promotion: *a bare convergence protects nothing while no case reads the envelope.* Both halves are here, and the second one is real.

**The convergence** takes `success` off the same `HttpDispatcher` instance the error exits already borrow — and renames `realErrorExits` → `realDomainDeps`, so the identifier stops describing three of the four exits it now carries. Small, but it is the kind of stale name that sends the next reader looking for a second harness.

**The envelope cases read the response body**, not a hand-built object, and one of them is better than what I asked for: the **create (201)** path pins that the envelope survives the **status-override mutation** `domains/data.ts` performs at that one call site. That override is precisely where a hand-rolled envelope gets clobbered without anything noticing — a query-only pin would have left it uncovered.

**The reverse verification points the right way.** The requirement was not "the test went red" but "the new case fails against the OLD stand-in" — because a case that still passes there is not reading the envelope at all. Reported:

```
× a real query success is wrapped in success/data/meta, not returned bare
  → expected false to be true
❯ expectDataSuccessEnvelope … BaseResponseSchema.safeParse(body).success
```

Envelope absent, not a compile error, and both cases fail — so both are load-bearing.

Scope held: **zero production bytes**, one test file plus a changeset that says coverage-only. `@objectstack/runtime` stays at its recorded 227 — the re-measure confirms it was not flagged as changed rather than merely not raised.

This closes the mirror half #6719 deferred: all four exits of the `/data` harness now answer through the real dispatcher.

---

Issue #7385, opened 2026-08-10
[finding][services] The `mongodb` and `sqlite-wasm` arms of the same factory still answer a missing optional package with the fault and no remedy


os-help:
**Observation class** (`finding`, no `pm:queue`). Found while implementing #7314 (PR #7384); filed unassigned rather than folded in, because #7314's scope is the libSQL/Turso arm and this is a sibling one.

## Fact

`packages/services/service-datasource/src/default-datasource-driver-factory.ts` builds three drivers that ride in OPTIONAL packages, and all three used to answer an absent package with the fault and nothing else. #7314 / PR #7384 fixed one of them. The other two are unchanged:

```text
sqlite-wasm driver requested but @objectstack/driver-sqlite-wasm is not installed (…).
mongodb driver requested but @objectstack/driver-mongodb is not installed (…).
```

Neither states the install command, neither states what happens next, and neither names the datasource that failed — while the `turso` arm beside them now states all three, and the host loader in `@objectstack/runtime` has stated them for libSQL since #5602. An admin who adds a mongo datasource in Setup and an admin who adds a libSQL one hit the same class of problem and get two different qualities of answer, decided by which driver they picked.

The remedy text is not the only thing missing: PR #7384 exports `TURSO_DRIVER_PACKAGE` / `TURSO_DRIVER_INSTALL_COMMAND` / `missingTursoDriverMessage`, so "how do I install the optional driver" now has one declaration for libSQL and none for the other two.

## Why it is observation-grade rather than a defect

Nothing is broken. The failure fires at the right moment, is a refusal rather than a fallback, and interpolates the underlying import error — which is what keeps `isUnbuiltWorkspaceFailure` (`connect-failure-remedy.ts`) able to recognise a half-built checkout from these arms at all, since the re-throw drops the original `code`. What is missing is the operator's next step, which the sibling arm now provides.

Severity is deliberately not asserted; filed plainly for triage.

## Shape a fix would take, if triaged

[...]

Issue #7314, opened 2026-08-10
[finding][services] After #6268 there are still TWO Turso loaders, and the open-core one drops the install command, the typed error, and half the config


os-help:
**Observation class** (`finding`, no `pm:queue`). Found while implementing #6268 (PR #7313); filed unassigned rather than fixed, because it is outside that card's file surface (`packages/cli` + `packages/runtime` only).

## Fact

#6268 converged the two **host-injected** Turso loaders onto one owner (`packages/runtime/src/turso-driver-factory.ts`). It did not — and could not — touch a **third**, independent libSQL loading arm that has been in the open-core factory since #6345:

`packages/services/service-datasource/src/default-datasource-driver-factory.ts`, the `if (kind === 'turso')` arm.

The split is deliberate and documented in that arm's own comment: the host factories win for the `default` datasource, and this arm serves **every other door** — a datasource created in Setup, `testConnection`, a declared non-default datasource. So "two loaders" is not by itself the finding. What is:

## Three ways the two disagree

1. **No install command, anywhere.** The host loader raises `MissingDriverPackageError` carrying `installCommand` as **data** plus a message that states the command, the consequence, and the deliberate refusal. This arm raises

   ```
   turso driver requested but @objectstack/driver-turso is not installed (<import error>).
   ```

   An admin who adds a libSQL datasource in Setup is told what is wrong and not how to fix it — while an operator who boots with the same URL gets the full instruction. Same missing package, two different qualities of answer.
2. **A plain `Error`, not the typed one.** `packages/cli/src/commands/serve.ts` decides fatality with `e instanceof MissingDriverPackageError`. This arm's failure cannot satisfy that predicate. Nothing routes it there **today** — which is exactly why this is observation-grade and not a live bug — but it is the same class-identity hazard #6268 was filed about, one layer down, and it is currently unpinned by any test.
3. **The config surfaces are not the same size.** This arm builds `TursoDriver` with `url`, `authToken`, `encryptionKey`, `concurrency`, `syncUrl`, `sync`, `timeout`, `mode`, `schemaMode`. The host loader reads **`url` and `authToken` only**. So the same declared libSQL config is honoured or silently dropped depending on whether the datasource happens to be named `default`: an encrypted or embedded-replica `default` loses `encryptionKey` / `syncUrl` / `sync` with no diagnostic. `TursoConfigSchema` accepts all of them, so this is declared-but-not-enforced in one of the two positions.

[...]

PR #7999, opened 2026-08-12
fix(runtime,service-datasource): converge the two libSQL loaders on one config read and one error class (#7314)

hotlong:
Part of #7314 — points **2** and **3**. Point 1 landed via #7384 (`b0c16a561`) from the `domain:services` seat and is not touched here.

## The defect

Two loaders build the libSQL/Turso driver, and which one runs is decided by something an author cannot see — whether the datasource happens to be the host's `default`:

- `@objectstack/runtime`'s `loadTursoDriverFactory` (single owner for the CLI and the standalone stack since #6268) serves `default`;
- `createDefaultDatasourceDriverFactory`'s `turso` arm in `@objectstack/service-datasource` serves every other door — a datasource created in Setup, `testConnection`, a declared non-default.

#6268 converged the two **host** loaders. It could not reach the third: `runtime` depends on `service-datasource`, never the reverse. The two had drifted in two ways.

### Point 3 — half the config was silently dropped for `default`

`turso-driver-factory.ts:248` built `new TursoDriverCtor({ url, ...authToken })` — **two keys**. The open-core arm read nine. `TursoConfigSchema` accepts all nine, so an encrypted or embedded-replica `default` lost `encryptionKey` / `syncUrl` / `sync` / `concurrency` / `timeout` / `mode` / `schemaMode` with no diagnostic anywhere — and got them back the moment the datasource was renamed away from `default`.

Fixed by **derivation, not a corrected copy**. Both loaders now build through one exported `buildTursoDriverConfig`; its key set comes from a reader table that TypeScript refuses to compile with a key missing, so the list is never written twice. A second hand-transcribed list is how the first two came to disagree — they agreed on the day they were written too.

The host loader also now **trims** the url before testing it, as the open-core arm always has: a whitespace-only url is refused by name instead of reaching `@libsql/client`.

### Point 2 — one `MissingDriverPackageError`

The class was declared at `turso-driver-factory.ts:114` in `@objectstack/runtime`, a package the open-core arm cannot import, so that arm raised a plain `Error` — matched by no `instanceof`, pinnable only by message text. The only legal convergence is moving the class **down**, and that is what this does: it now lives in `@objectstack/service-datasource/src/missing-driver-package-error.ts`, and both loaders throw the same class object.

**No importer changes.** `MissingDriverPackageError`, `TURSO_DRIVER_PACKAGE` and `TURSO_DRIVER_INSTALL_COMMAND` are still exported from `@objectstack/runtime` (and from `@objectstack/cli`'s `utils/storage-driver.ts` through it) — re-exports now rather than declarations. `serve.ts`'s `e instanceof MissingDriverPackageError` fatal-boot branch depends on that identity, so dropping the old export would have been an API break performed for a refactor's convenience.

## Pins

Both required pins assert the property that hides from the obvious test:

- **the constructor argument, not a successful boot** (`turso-driver-factory.convergence.test.ts`). A dropped key produces a driver that constructs perfectly and connects to the wrong thing; every boot-level assertion stayed green through the years this loader read two keys. A `default` spec carrying all nine keys is asserted to reach the ctor with all nine, and to equal what `buildTursoDriverConfig` produces — so a second hand-written list that happens to agree today still fails.
- **class identity, not name or message.** `MissingDriverPackageError === ` the `service-datasource` binding; the host loader's error `instanceof` the open-core binding; and — the direction that could not be written before this change — the open-core arm's error `instanceof` the runtime binding.
- **the builder against the real driver config**, compile-time, in `packages/cli` (the only one of the three packages carrying `@objectstack/driver-turso`, and whose `typecheck` compiles tests). A key the driver adds and the builder never reads is a defect with no runtime symptom at all.

## Reverse verification

Each half reverted independently, failure observed, restored:

| reverted | observed |
|---|---|
| host loader back to `{ url, authToken }` | 2 failures, both `AssertionError: expected { Object (url, authToken) } to deeply equal { Object (url, authToken, ...) }`. The 5 identity cases stayed green — the halves are independent. |
| open-core arm back to `throw new Error(...)` | 1 failure: `AssertionError: expected Error: datasource 'warehouse': a libSQL/T… to be an instance of MissingDriverPackageError`. |
| `encryptionKey` removed from the shared builder | `packages/cli` typecheck red: `storage-driver.test.ts(460,40): error TS2344: Type '"encryptionKey"' does not satisfy the constraint 'never'`. |

## A finding, not a cleanup — `MissingDriverPackageErrorTwin`

`packages/cli/src/utils/storage-driver.test.ts:447` declares a hand-rolled twin of the class. The dispatch brief read it as a stand-in written because the real class could not be reached, to be deleted once it was importable. **It is not that, and it was kept.** The real class was already reachable in that file — two cases above the twin, `MissingDriverPackageError` is asserted `===` the runtime one. The twin is a deliberate **negative control**: something that matches on message *and* name and still fails `instanceof`, which is the only thing in the file demonstrating that the passing assertion tests identity rather than wording. Deleting it would have removed the evidence that the pin has teeth. It is extended here to fail against the new open-core binding too, and the reasoning is recorded in place.

## Checks

- `pnpm -w typecheck` — clean (127 tasks).
- `@objectstack/service-datasource` 339 ✓ · `@objectstack/runtime` 2165 ✓ · `@objectstack/cli` 1249 ✓.
- `pnpm check:type-check-debt` — clean; the ledger does **not** rise (the new runtime test initially added +1 to `TEST_DEBT`; the type was fixed rather than the entry raised).
- ESLint clean on every touched file.
- Changeset: `.changeset/turso-host-loader-convergence.md`.

## Out of scope

#7385 (the `mongodb` / `sqlite-wasm` arms of the same no-remedy defect class) stays with `domain:services` and is untouched. Noted while here: those two arms still raise an untyped `Error`. The class moved down is deliberately driver-agnostic (`driverType` is a field), so adopting it there is a constructor call away whenever that lane takes it up.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RjUepKTxiGcJX6WQtwFmcf)_

### DIFF
--- a/.changeset/turso-host-loader-convergence.md
+++ b/.changeset/turso-host-loader-convergence.md
@@ -1,0 +1,48 @@
+---
+"@objectstack/service-datasource": minor
+"@objectstack/runtime": patch
+---
+
+fix(runtime,service-datasource): a `default` libSQL datasource keeps its whole config, and one missing-package class (#7314)
+
+Two loaders build the libSQL/Turso driver, and which one runs is decided by
+something the author cannot see — whether the datasource happens to be the
+host's `default`. `@objectstack/runtime`'s host loader serves that one;
+`createDefaultDatasourceDriverFactory`'s `turso` arm in
+`@objectstack/service-datasource` serves every other door (a datasource created
+in Setup, `testConnection`, a declared non-default). #6268 converged the two
+HOST loaders onto one owner; it could not reach the third, one layer down, and
+the two had drifted in two ways.
+
+**Half the config was silently dropped for `default`.** The host loader built
+`new TursoDriver({ url, authToken })` — two keys — while the open-core arm read
+nine. `TursoConfigSchema` accepts all nine, so an encrypted or
+embedded-replica `default` lost `encryptionKey` / `syncUrl` / `sync` /
+`concurrency` / `timeout` / `mode` / `schemaMode` with no diagnostic anywhere,
+and got them back the moment the datasource was renamed. Both loaders now build
+through one exported `buildTursoDriverConfig`, whose key set is derived from a
+reader table rather than hand-listed — a corrected second copy would only have
+agreed until the next key. A `packages/cli` pin fails to compile if that builder
+and the driver's own `TursoDriverConfig` stop covering the same keys.
+
+The host loader also now trims the url before testing it, as the open-core arm
+always has: a whitespace-only url is refused by name instead of being handed to
+`@libsql/client`.
+
+**One `MissingDriverPackageError`, reachable from both sides.** The class was
+declared in `@objectstack/runtime`, which `@objectstack/service-datasource`
+cannot import (the dependency runs the other way), so the open-core arm raised a
+plain `Error` — matched by no `instanceof`, and pinnable only by message text.
+The declaration moves DOWN to `@objectstack/service-datasource`, the lowest
+package that raises it, and both loaders now throw the same class object.
+
+**No import changes.** `MissingDriverPackageError`, `TURSO_DRIVER_PACKAGE` and
+`TURSO_DRIVER_INSTALL_COMMAND` are still exported from `@objectstack/runtime`
+(and from `@objectstack/cli`'s `utils/storage-driver.ts` through it) — they are
+re-exports now rather than declarations. Code written against either spelling
+keeps compiling, and against the same class: `serve.ts`'s
+`e instanceof MissingDriverPackageError` fatal-boot branch depends on that
+identity, so it is asserted by object identity rather than by name or message.
+`@objectstack/service-datasource` additionally exports the class, the builder
+(`buildTursoDriverConfig`, `resolveTursoUrl`, `TURSO_DRIVER_CONFIG_KEYS`) and
+`resolveDatasourceSchemaMode` for hosts that build their own driver factory.
--- a/packages/cli/src/utils/storage-driver.test.ts
+++ b/packages/cli/src/utils/storage-driver.test.ts
@@ -5,6 +5,16 @@ import { describe, it, expect } from 'vitest';
 // this file). `@objectstack/driver-turso` is an OPTIONAL peer of the CLI — this
 // import is erased, so nothing here requires it at runtime.
 import type { TursoDriverConfig } from '@objectstack/driver-turso';
+// #7314: the shared libSQL config builder both loaders call, and the class both
+// now raise. Imported from `@objectstack/service-datasource` — the lowest package
+// of the three, and the only import direction the dependency graph allows.
+import {
+  buildTursoDriverConfig,
+  resolveTursoUrl,
+  TURSO_DRIVER_CONFIG_KEYS,
+  MissingDriverPackageError as OpenCoreMissingDriverPackageError,
+  type TursoDriverConfigInput,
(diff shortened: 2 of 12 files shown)


hotlong:

## The twin: the dispatch was wrong and the report corrects it

My claim comment read `MissingDriverPackageErrorTwin` as a stand-in written because the real class was unreachable, and said to delete it in favour of the real import. **That was wrong**, and the PR says so with evidence: the real class was *already* reachable in that file — two cases above the twin, `MissingDriverPackageError` is asserted `===` the runtime one.

The twin is a deliberate **negative control**: something that matches on message *and* name and still fails `instanceof`. It is the only thing in that file demonstrating that the passing assertion tests **identity** rather than **wording**. Deleting it would have removed the evidence that the pin has teeth — i.e. my instruction would have quietly weakened the exact property this card exists to establish.

Extending it to fail against the new open-core binding too, and recording the reasoning in place, is the right disposition. Noting this at length because the escape hatch is what saved it: the dispatch said *"if it turns out to encode something the real class does NOT, say so — that is a finding, not a cleanup."* A brief that had only said "delete it" would have been followed.

## Point 3 is fixed by derivation, and the pin is built so a copy cannot reappear

One exported `buildTursoDriverConfig`, its key set from a reader table **TypeScript refuses to compile with a key missing** — so the list is never written twice. And the assertion is not "all nine keys arrive", it is *"the ctor argument equals what the builder produces"*, which means **a second hand-written list that happens to agree today still fails**. The two loaders agreed on the day they were written too; that is precisely how this card came to exist.

The reverse verification proves the derivation is real rather than decorative: removing `encryptionKey` from the shared builder turns `packages/cli`'s **typecheck** red (`TS2344: Type '"encryptionKey"' does not satisfy the constraint 'never'`).

Two things found along the way that the card did not name: a ninth key (`schemaMode`), and the host loader not trimming the url — now refused by name instead of reaching `@libsql/client`.

## The three pins each cover a different way this hides

- **The constructor argument, not a successful boot.** Stated with the reason: a dropped key produces a driver that constructs perfectly and connects to the wrong thing, which is why every boot-level assertion stayed green for years.
- **Class identity in three directions**, including the one that *could not be written before this change* — the open-core arm's error `instanceof` the runtime binding.
- **The builder against the real driver config, at compile time**, in `packages/cli` — the only one of the three packages carrying `@objectstack/driver-turso` whose typecheck compiles tests. *"A key the driver adds and the builder never reads is a defect with no runtime symptom at all."* That third pin covers a direction neither of the other two can see.

## The API-break trap was identified concretely

`MissingDriverPackageError` / `TURSO_DRIVER_PACKAGE` / `TURSO_DRIVER_INSTALL_COMMAND` stay exported from `@objectstack/runtime` as re-exports, and the PR names the consumer that made it load-bearing: `serve.ts`'s `e instanceof MissingDriverPackageError` fatal-boot branch. That is the difference between "kept the export because the brief said so" and knowing what would have broken.

Ledger discipline held: the new runtime test initially added +1 to `TEST_DEBT`; the **type was fixed, not the entry raised**.

#7385 correctly untouched — with a useful note that the moved class is deliberately driver-agnostic (`driverType` is a field), so the `mongodb` / `sqlite-wasm` arms are a constructor call away whenever `domain:services` takes them up.

---

Issue #8003, opened 2026-08-12
finding: the list-query normalizer's `where` arity note still says a repeated `?filter=` is refused by `isFilterAST` failing


hotlong:
Observation-class finding, split out of #7390 while implementing it. **Unassigned**, filed per Prime Directive #10. Comment-only drift — no runtime behaviour is wrong.

## The sentence

`packages/metadata-protocol/src/protocol.ts`, in the docblock above `ARRAY_VALUED_QUERY_SLOTS` (the #7321 arity survey), the `where` row reads:

> `where` — a FILTER AST is an array (`['status','=','open']`), so a blanket arity refusal here would reject the AST body form outright. A repeated `?filter=` is still refused, one block down, by `isFilterAST` failing to read it.

The first sentence is correct and load-bearing — it is exactly why #7386's arity gate skips this slot, and #7390 quotes the same reasoning.

The last sentence has two problems, one of them older than #7390:

1. **It was always describing the defect as if it were the answer.** `isFilterAST` failing produces `malformedFilterArrayError` — a 400 telling a caller whose filters were each well-formed that their filter is *malformed*, listing the AST operator vocabulary. That is #7390's entire finding: the refusal is correct, the diagnosis is not. "Still refused" reads as *handled*.
2. **It is now stale for the route it describes.** #7390 moves the arity judgement for this slot to the REST querystring ingress (`packages/rest`, `assertFilterParamSuppliedOnce`), which is the only layer that can make it — on a querystring an array is a repetition and can be nothing else. A repeated `?filter=` on `GET /data/:object` no longer reaches this block at all.

It also does not mention the second shape at all: `?filter=status&filter=%3D&filter=open` spelled a valid AST and **succeeded** with a filter nobody expressed, which is the sharper half of #7390.

## Why it matters more than a typo

This docblock is the survey a future author reads to decide whether a slot needs arity treatment. A reader who takes "still refused, one block down" at face value concludes the filter slot is covered and stops — which is roughly how the gap survived from #7321 to #7390. The sentence should say that arity for this slot is judged at the querystring ingress and name #7390, so the next reader is told which layer owns it.

Deliberately **not** fixed in #7390's PR: that card's ruling put the enforcement in `packages/rest` and explicitly kept it out of `packages/metadata-protocol`, and `protocol.ts` is a hot shared file. Same handling as #7707 / #7874, which corrected stale comments in their own PRs.

## Dedup

[...]

Issue #8002, opened 2026-08-12
finding: the wire-only filter alias spellings (`filters` / `$filter`) are now declared in two packages, with no gate reconciling them


hotlong:
Observation-class finding, split out of #7390 while implementing it. **Unassigned**, filed per Prime Directive #10. Nothing a user hits today — this is a drift seam, not a live defect.

## What the seam is

The filter slot has four wire spellings, and they are declared in two places with different ownership:

- `where` and `filter` come from `RPC_QUERY_ALIAS_SLOTS` (`packages/spec/src/data/data-engine.zod.ts`), which its own docblock calls "the ONE place the alias to canonical mapping is declared".
- `filters` and `$filter` are **wire-only** — no schema declares them. `packages/metadata-protocol` extends the spec table with them locally (`WIRE_QUERY_ALIAS_SLOTS`).

#7390 added a third reader: `packages/rest` now gates the filter slot's arity at the querystring ingress, and needs the same four spellings to do it. It derives `where` / `filter` from the spec table, but cannot derive `filters` / `$filter` — `@objectstack/metadata-protocol` is a **dev-only** dependency of `@objectstack/rest`, so no runtime import of that table exists. They are named literally in `packages/rest/src/query-multiplicity.ts` (`FILTER_SLOT_QUERY_PARAMS`).

## The consequence

A **new** wire-only alias for the filter slot, added to `metadata-protocol`'s table, would fold correctly in the normalizer and be silently ungated at the ingress — repetition on that spelling would fall back to the misdiagnosis #7390 exists to remove. Nothing fails: the spelling works, the gate simply does not see it.

Half of it is already pinned: `rest-server-repeated-filter-param.test.ts` §5 asserts the derived set composes to exactly the four spellings, so a spec-table change goes red. The wire-only half has no such backstop, because the package holding the truth is not on the runtime graph.

Symmetric note: `WIRE_DOLLAR_ALIASES` in the same normalizer already carries a comment about exactly this class of mistake — "Adding a `$` alias in one place and not the other is exactly how a parameter ends up folded but unchecked." That sentence is now true across a package boundary as well as within one.

## Options

[...]

Issue #8001, opened 2026-08-12
A repeated `?filter=` answers two different error codes depending on which data route received it


hotlong:
Found while implementing #7390 (PR pending). Filed unassigned per Prime Directive #10; **not** fixed there, because #7390's ruling names the code for its own route and the sibling route is a different surface.

## The divergence

Two sibling routes in `packages/rest/src/rest-server.ts` both refuse a repeated `?filter=`, and they answer differently:

| route | code | envelope | gate |
| :--- | :--- | :--- | :--- |
| `GET /data/:object` | `INVALID_FILTER` | flat `{ error, code, object }` (`mapDataError`) | `assertFilterParamSuppliedOnce` (#7390) |
| `GET /data/:object/export` | `VALIDATION_ERROR` | nested `{ error: { code, message } }` | `refuseRepeatedQueryParams` (#6877) |

So `?filter=a&filter=b` is one caller mistake with two machine-readable answers, decided by which path it was sent to. A client that branches on `error.code` has to know both.

## Why each is currently right on its own terms

Neither is an oversight, which is why this is a fork rather than a bug with an obvious fix:

- The export route's answer comes from the shared `#6877` helper, whose whole point is that **every route answers the same multiplicity rule identically**. `VALIDATION_ERROR` is `packages/rest`'s house catalog member for a 400.
- The list route's answer is a **maintainer ruling** (2026-08-11, on #7390): `400 INVALID_FILTER` with a message naming repetition. It also matches that route's own family — `unusableFilterError` and `malformedFilterArrayError` both reach the wire as `INVALID_FILTER` through `mapDataError`, so a repeated filter there is one of three ways the same slot fails, all with one code and one body shape.

Both rules are "one condition, one answer" — they just draw the condition's boundary differently. The list route reads it as *the filter slot failed*; the export route reads it as *a single-valued parameter was repeated*.

## The options, if this is worth closing

1. **Move the export route's filter slot onto the same gate** — repeated `filter` there becomes `INVALID_FILTER` too, and the other export parameters (`format`, `limit`, `page`, `orderby`, `search`, `header`) keep `VALIDATION_ERROR`. Consistent per-slot; splits the export route's own answer across two codes.
2. **Leave it.** The boundary really is per-route-family, and each envelope matches the family its route already speaks.
3. **Move the list route onto `VALIDATION_ERROR`** — consistent per-condition, but it contradicts the #7390 ruling and would give one slot two codes on one route.

[...]

PR #8004, opened 2026-08-12
fix(rest): refuse a repeated ?filter= as a repetition, not as a malformed filter (#7390)

### LINKED ISSUE #7390: finding: a repeated `?filter=` on `GET /data/:object` cannot be told from a filter AST, so it is diagnosed as a malformed filter (and, rarely, succeeds)
Observation-class finding, split out of #7321 while implementing it (PR #7386). **Unassigned**, filed per Prime Directive #10. #7386 deliberately did not widen into this — closing it needs an acceptance-surface decision that card had no mandate for.

## What #7386 did, and the one slot it could not cover

#7386 added an arity gate to `findData`'s shared list-query normalizer (`packages/metadata-protocol/src/protocol.ts`). The rule keys off the **declared value type**, because that normalizer serves two ingresses it cannot tell apart — `GET /data/:object` (a repeated querystring arrives as `string[]`) and `POST /data/:object/query` (the body is arbitrary JSON). On a slot whose spec type never admits an array, `Array.isArray` is unambiguous evidence of repetition. On a slot that does admit one, it is the ordinary shape.

`where` / `filter` / `filters` / `$filter` is on the second list, and it is the only member where that costs something. A filter **AST is an array**: `["status","=","open"]`. So a repeated `?filter=A&filter=B` and a body-form AST are byte-identical at this layer, and the arity gate has to leave the slot alone.

## The consequence, in two shapes

**1. A repeated filter IS refused, but the diagnosis names the wrong cause.** `?filter={"a":1}&filter={"b":2}` arrives as `['{"a":1}','{"b":2}']`. `isFilterAST` cannot read it (the first element is a string, the second is not a known operator, and the legacy flat-array arm needs every element to be an array), so it falls to `malformedFilterArrayError` — `400`, `error.code: INVALID_FILTER`. The refusal is correct; the message tells the caller their filter is malformed, when in fact each filter they sent was fine and the mistake was sending two.

**2. Rarely, a repeated filter silently SUCCEEDS.** When the repetition happens to spell a valid AST it is parsed as one. `?filter=status&filter=%3D&filter=open` arrives as `["status","=","open"]`, which `isFilterAST` accepts, and `parseFilterAST` lowers to `{ status: "open" }`. Three occurrences of one parameter become one working filter. Contrived to write by hand, but it is a 200 with a filter nobody expressed, which is the same class as the rest of #7321.

## Why it is separate, and why it is `finding` and not queued

Fixing shape 2 properly means deciding something #7386 had no mandate to decide: either the wire form of `filter` stops accepting the bare AST array (an **acceptance-surface** change, which is a `packages/spec` decision — see #6017, and note #6298's first red line excludes acceptance-surface changes), or the transport tells the normalizer which ingress it came from (a new contract between `packages/rest` and `packages/metadata-protocol`, i.e. a different owner's surface). Both are decisions, not implementations.

Shape 1 alone — improving only the message — is cheaper but would have to guess: "an array of strings that are each parseable JSON is probably a repetition" is a heuristic, and heuristics in the normalizer are what #4181 and #4121 spent effort removing.

## Not reachable today, same expiry as its parent

The production Hono adapter collapses repeated parameters to the first value before any handler runs, so no caller hits either shape now. That dormancy ends with **#6878 route 2** (ruled adopted 2026-08-10), exactly like #7321's did. Priority is coupled to that, not independent of it.

## Dedup

Searched open issues for `readSingleQueryValue`, `query-multiplicity`, `repeated query`, `isFilterAST`, `malformed filter array`, and `query multiplicity single-valued`. Hits: #7321 (the parent, whose fix is PR #7386), #6878 (the adapter divergence that ungates the class), #7360 (the same class on the automation descriptor routes — a different surface, already filed unassigned), #6017 (the `domain:spec` PM seat). None covers the filter slot's AST-vs-repetition ambiguity.

hotlong:
Fixes #7390

Implements the maintainer ruling recorded on the issue 2026-08-11 — quoted verbatim and untranslated:

> **Ruling: refuse, narrowest shape.** A repeated `?filter=` on `GET /data/:object` is refused explicitly — 400 `INVALID_FILTER` with an actionable message naming the condition ("repeated filter parameter — send exactly one") — instead of today's confusing malformed-filter diagnosis (and the rare accidental success). Last-wins and AND-merge are rejected: silent selection among duplicates is the AI-authoring trap this lane refuses on principle.

## What was wrong

Two of the issue description's sections were overtaken by events before this was picked up; both were re-verified on `origin/main` before implementing.

**"Not reachable today" had expired.** `packages/plugins/plugin-hono-server/src/adapter.ts:202` reads `c.req.queries() ?? {}` since #6878 route 2 landed (PR #7396), so repeated parameters reach `findData`'s normalizer as `string[]` and **both** shapes in the body were live.

**The normalizer structurally cannot judge this slot.** #7386's arity gate keys off each slot's declared value type, and a filter AST *is* an array — `["status","=","open"]`. The normalizer serves two ingresses it cannot tell apart (`GET /data/:object`, where a repeat is `string[]`, and `POST /data/:object/query`, whose body is arbitrary JSON), so on this one slot `Array.isArray` is not evidence of anything.

| request | before | after |
| :--- | :--- | :--- |
| `?filter={"a":1}&filter={"b":2}` | `400 INVALID_FILTER`, diagnosed as **malformed** | `400 INVALID_FILTER`, diagnosed as **repetition** |
| `?filter=status&filter=%3D&filter=open` | **`200`**, applying `{status:"open"}` | `400 INVALID_FILTER` |

The first was the common shape and its message was actively misleading — both filters the caller sent were well-formed, and the response told them to check the AST syntax that was never wrong. The second is contrived to write by hand but is the sharper defect: three occurrences of one parameter happened to spell a valid AST, so the request succeeded while applying a filter nobody expressed.

## Where it lands, and why not the obvious place

At the **REST querystring parse** (`packages/rest`), not in the shared normalizer. The querystring ingress is the only layer that knows it is a querystring — and there an array on the filter slot is a repeated parameter and can be nothing else. That keeps the normalizer free of the heuristic ("an array of strings each parseable as JSON is probably a repetition") that #4181 and #4121 spent effort removing, and leaves the body face alone.

`packages/spec` does not move: `INVALID_FILTER` is already a standard-catalog code (`errors.zod.ts`), so no ledger entry is needed either.

## The shape follows the file's existing one

The gate lives in `query-multiplicity.ts`, the module that already owns this rule (`[#6877]`), and reuses its `readSingleQueryValue` count-not-shape semantics — so a one-element array is one occurrence and is unwrapped rather than refused.

It **throws** rather than responding, which is the one deliberate difference from its sibling `refuseRepeatedQueryParams`. The data routes speak the flat `mapDataError` envelope, and that is the envelope this route's other filter refusals already arrive in — `unusableFilterError` and `malformedFilterArrayError` both throw `400` / `INVALID_FILTER` and are shaped by the handler's own catch. Throwing from inside the same `try` gives one slot **one wire code and one body shape**, whether the filter was unreadable or sent twice; responding here would have authored a second dialect for one condition. `INVALID_FILTER` is already in `isExpectedQueryRejection`'s vocabulary, so the refusal does not print an "[REST] Unhandled error" line.

All four wire spellings of the one slot are covered. `where` / `filter` are **derived** from the spec's own `RPC_QUERY_ALIAS_SLOTS` ("the ONE place the alias to canonical mapping is declared"), so a spelling added there reaches the gate; `filters` / `$filter` are wire-only and named locally, because `@objectstack/metadata-protocol` is a dev-only dependency and no runtime import of its table exists. That asymmetry is filed as #8002.

## Pins

`packages/rest/src/rest-server-repeated-filter-param.test.ts`, 24 cases:

- **§1 refusals** — both shapes, on every spelling, plus two identical values (still two occurrences), plus a check that the refusal is not logged as an unhandled fault. Every case asserts the ADR-0112 pair (`status` **and** `code`) **and** that the message names repetition and does not say "malformed". That third assertion is the point: `400` + `INVALID_FILTER` were both already true of the misdiagnosis, so a status-only test passes green against the bug.
- **§2 preservation** — a single `?filter=` in **both** accepted forms (JSON object and bare AST) reaches `findData` byte-identical; a one-element array is unwrapped; `$select` keeps its array arm; a filterless request is untouched.
- **§3 the negative pin** — `POST /data/:object/query` with a body-form AST (flat and nested) is forwarded, proving the **ingress** was gated and not the parser.
- **§4 the layer below**, on a real engine and the real normalizer: handed the repetition directly it *still* answers `Malformed $filter: unrecognised operator...`, and handed the accidental AST directly it *still* returns the wrong row happily. Those two are the reverse verification kept as tests rather than as a transcript — they document why deleting the gate is not a cleanup.
- **§5** — the spelling set composes to exactly the four wire spellings.

## Reverse verification

Guard removed from the committed state, suite re-run: **10 failed / 14 passed, no compile error**, and it produced both directions the card allows for.

The **old diagnosis**, through the real stack:

```
AssertionError: expected 'Malformed $filter: unrecognised opera...'
                     to be 'Repeated "filter" query parameter - s...'
```

The **accidental 200**, through the real engine — the request returning exactly the one `open` row nobody asked to filter on:

```
expected a 400 for a repeated "filter", got 200 with body
{"object":"task","records":[{"id":"1",...,"status":"open"}],"total":1,"hasMore":false}
```

The mocked-protocol cases fell to `200 {"records":[]}`, and the one-element-array case failed as `expected [ '{"status":"open"}' ] to be '{"status":"open"}'` — so the unwrap is real work, not a no-op. Guard restored with `git checkout` from the commit (never `git stash`); suite green again.

## Verification

- `pnpm -w typecheck` (turbo, `--concurrency=2`): **127/127 successful**. That whole-workspace run is also the consumer sweep — no single-direction `--filter` ambiguity.
- `@objectstack/rest` suite: **95 files / 1555 tests passed**.
- `check:type-check-debt`: `@objectstack/rest` TEST_DEBT re-measured at **155**, exactly its recorded number, with **zero** errors attributable to the new file. The entry has no margin by design, so a first measurement of 156 (one TS2554 from a one-argument `registerObject`) was **fixed**, not ledgered.
- `check:nul-bytes` OK; `check:error-code-casing` OK.

## Not in this PR

`content/docs/protocol/**` was checked before writing anything: `kernel/http-protocol.mdx` documents `filter` as a single JSON parameter and nothing claims it is multi-valued, so no spec-side prose rides this change.

## Out-of-scope findings

Filed unassigned per Prime Directive #10, none fixed here:

- **#8001** — a repeated `?filter=` answers `INVALID_FILTER` on this route and `VALIDATION_ERROR` on `GET /data/:object/export`. Both are individually right (this one by ruling, that one by the shared #6877 helper); recording the fork rather than choosing for the maintainer.
- **#8002** — `filters` / `$filter` are now declared in two packages with no gate reconciling them.
- **#8003** — the normalizer's `where` arity note still says a repeated `?filter=` is refused by `isFilterAST` failing, which was always describing the defect as the answer and is now stale for this route too.

---

### DIFF
--- a/.changeset/repeated-filter-param-refusal.md
+++ b/.changeset/repeated-filter-param-refusal.md
@@ -1,0 +1,60 @@
+---
+"@objectstack/rest": patch
+---
+
+fix(rest): a repeated `?filter=` on `GET /data/:object` is refused as a repetition, not misdiagnosed as a malformed filter (#7390)
+
+**This is a behaviour change on a live surface.** A request that previously
+answered `200` now answers `400`, and a request that already answered `400` now
+carries a different message. Both changes make the response describe what the
+caller actually did.
+
+Repeating a query parameter used to be invisible: the production Hono adapter
+collapsed repeats to the first value before any handler ran. Since #6878 route 2
+(PR #7396) it surfaces them as arrays, so a repeated `?filter=` now reaches the
+shared list-query normalizer — and that normalizer structurally cannot tell what
+it is looking at. A filter AST **is** an array (`["status","=","open"]`), so the
+arity gate #7386 added to every other query slot had to leave this one alone: on
+the filter slot, an array is the ordinary shape of a legitimate body-form filter
+sent to `POST /data/:object/query`.
+
+Two shapes came out of that, both live:
+
+| request | before | now |
+| :--- | :--- | :--- |
+| `?filter={"a":1}&filter={"b":2}` | `400 INVALID_FILTER`, diagnosed as a **malformed** filter | `400 INVALID_FILTER`, diagnosed as a **repetition** |
+| `?filter=status&filter=%3D&filter=open` | **`200`**, applying `{status:"open"}` | `400 INVALID_FILTER` |
+
+The first was the common one, and its message was actively misleading: both
+filters the caller sent were well-formed, the response told them to check their
+AST syntax, and the operator vocabulary it listed could not help. The second is
+contrived to write by hand but is the sharper defect — three occurrences of one
+parameter happened to spell a valid AST, so the request succeeded while applying
+a filter nobody expressed.
+
+The refusal now names the condition: `Repeated "filter" query parameter — send
+exactly one.` A repeated filter is **not** merged and **not** resolved by
+precedence — either would silently serve one of two intents the caller actually
+expressed, which is the authoring trap this refusal exists to close.
+
+The judgement is made at the REST querystring parse rather than in the shared
+normalizer, because the querystring layer is the only one that knows it is
+looking at a querystring: there, an array on the filter slot is a repeated
+parameter and can be nothing else. All four wire spellings of the one slot
+(`filter`, `where`, `filters`, `$filter`) are covered.
+
+**Unaffected:**
+
+- A **single** `?filter=` in either accepted form — the JSON object
+  (`?filter={"status":"open"}`) and the bare AST
+  (`?filter=["status","=","open"]`).
+- `POST /data/:object/query` — the body face legitimately sends an array, and is
+  untouched.
+- Genuinely multi-valued query parameters (`$select`, `$expand`,
+  `$searchFields`), which keep their array arm.
+- A one-element array from a repeat-preserving adapter, which is one occurrence
+  and is unwrapped rather than refused — this also stops it being read as a
+  malformed AST.
+
+No spec change: `INVALID_FILTER` is already a standard-catalog code, and the
+accepted wire forms of `filter` are unchanged.
--- a/packages/rest/src/query-multiplicity.ts
+++ b/packages/rest/src/query-multiplicity.ts
@@ -1,5 +1,7 @@
 // Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
 
+import { RPC_QUERY_ALIAS_SLOTS } from '@objectstack/spec/data';
+
 /**
  * Query-parameter MULTIPLICITY, for every REST handler in this package (#6877).
  *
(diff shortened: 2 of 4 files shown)


hotlong:

Every pin is here, including the one that decides whether this card was actually closed: **the message must name repetition, not malformedness**. The PR states the reason better than the dispatch did — *"`400` + `INVALID_FILTER` were both already true of the misdiagnosis, so a status-only test passes green against the bug."* That is exactly why a `toBe(400)` test would have shipped a green suite over an unchanged defect.

Both stale premises were re-verified rather than taken from the claim comment, which is the right handling for a card whose body inverts its own verdict.

## Two things beyond the ask

**§4 keeps the reverse verification as tests, not as a transcript.** Handed the repetition directly, the real normalizer *still* answers `Malformed $filter: unrecognised operator…`; handed the accidental AST, it *still* returns the wrong row happily. So the file now documents **why deleting this gate is not a cleanup** — the layer below is measured, in CI, to be unable to do this job. A future simplification that removes the gate turns those red instead of finding a green suite and a plausible-looking diff.

**Throwing instead of responding is the correct deviation, and the reasoning generalizes.** `unusableFilterError` and `malformedFilterArrayError` already throw `400` / `INVALID_FILTER` into the handler's own catch, so throwing gives this slot **one wire code and one body shape** whether the filter was unreadable or sent twice. Responding directly would have authored a second dialect for a single condition — the exact defect class this lane has closed five times today. Deviating from the sibling helper to avoid that is a better read of the rule than following it would have been.

Also correct: `where` / `filter` **derived** from `RPC_QUERY_ALIAS_SLOTS` rather than transcribed, with the `filters` / `$filter` asymmetry named and filed (#8002) instead of quietly hand-copied. The dispatch's ⛔ on heuristics is respected — the querystring ingress knows it is a querystring, so no guessing is needed anywhere.

## Ledger discipline held

`@objectstack/rest` first measured **156** — one `TS2554` from a one-argument `registerObject` in the new test — and it was **fixed, not ledgered**, back to the recorded 155. That entry has no margin by design and the PR says so.

## The three findings are correctly scoped

**#8001** is the one I want on the record: a repeated `?filter=` now answers `INVALID_FILTER` here and `VALIDATION_ERROR` on `GET /data/:object/export`. Both are individually right — this one by the maintainer's ruling, that one by the shared #6877 helper — and reconciling them is a choice about which vocabulary owns the condition. Recording the fork instead of picking one is right; picking would have been a ruling made inside an implementation.

**#8003** is a sharp catch: the normalizer's `where` arity note says a repeated `?filter=` is refused *by `isFilterAST` failing* — which was **always describing the defect as though it were the answer**, and is now stale for this route on top of that.

Docs correctly not touched: `kernel/http-protocol.mdx` already documents `filter` as a single JSON parameter, checked before writing rather than after.

---

hotlong:
**One addendum to the review above — not a change request, and not a reason to pull this out of the queue.**

The docs check in this PR went to `content/docs/protocol/kernel/http-protocol.mdx`, which was a reasonable place to look and correctly reported nothing claiming `filter` is multi-valued. The page that turned out to matter is a different one:

`content/docs/api/data-api.mdx:22` is the row that documents this parameter **and its rejection behaviour**:

> `filter` | query | Filter expression (JSON). `filters` also accepted for backward compatibility. **Malformed JSON is rejected with `400 INVALID_FILTER` — never ignored.**

That sentence stays true — this PR does not change malformed-JSON handling. But the page now under-describes the parameter it documents: a second condition answers the same code on the same slot, and it is the condition this PR exists to make legible. Leaving it out reproduces #7390's own gap one layer up — the caller is still not told that sending it twice is the problem.

**Incomplete, not incorrect**, so it is filed as **#8005** rather than churning a queued PR for one clause.

Worth noting for the drift advisory's own credibility: `data-api.mdx` is **not** among the 9 docs it listed here, despite being the most relevant. That is the second instance today of the mechanism recorded in **#7967** — the check maps docs to packages by textual mention, so a page describing a package's behaviour without naming the package is invisible to it. Measured with a control (`VALIDATION_ERROR` → 4 files under `content/docs/api/`), so the miss is real rather than a bad query on my side.

Verdict unchanged: **accept**, auto-merge on.

---

hotlong:
**PM ruling on the scope question raised in the report — gating all four spellings is correct. Do not trim.**

The report flags this honestly and asks rather than assumes:

> *the ruling says "repeated `?filter=`" and I gated all FOUR wire spellings of that ONE slot … If the PM reads "narrowest shape" as one spelling only, this is the line to trim.*

**"Narrowest shape" modifies the remedy, not the coverage.** It was the maintainer distinguishing *refuse* from the alternatives on the table — last-wins, AND-merge, and the acceptance-surface change that would have narrowed `filter`'s wire form. It says nothing about how many spellings reach the slot, because the alternatives it was ruling out were all about **what to do**, not **where**.

`where` / `filter` / `filters` / `$filter` are **one slot**. Gating only the literal `filter` would leave three identical holes, reachable by any caller who used an alias — the same defect under a different name, and a fourth instance of the one-rule-N-implementations divergence this lane has closed five times today. A refusal that depends on which synonym the caller happened to type is not narrower; it is just inconsistent.

Deriving `where` / `filter` from the spec's `RPC_QUERY_ALIAS_SLOTS` rather than transcribing them is what makes this hold: the coverage tracks the declaration, so a spelling added there is gated the day it is declared. The `filters` / `$filter` half genuinely cannot be derived — `@objectstack/metadata-protocol` is not on `rest`'s runtime graph — and naming that asymmetry in **#8002** instead of hand-waving it is the right disposition.

**#8001 stays unlabeled and ungraded**, as filed. One condition answering `INVALID_FILTER` here and `VALIDATION_ERROR` on `/data/:object/export` is a question about which vocabulary owns the condition — a ruling, not an implementation. Declining to pick it inside this PR was correct, and this seat is not picking it either; grading belongs to the triage seat.

The behaviour-change risk is stated the way it should be: `?filter=status&filter=%3D&filter=open` moves **200 → 400**, and anyone who had unknowingly come to depend on that accidental AST is now refused. That is the ruling's intent, and the changeset says so plainly rather than describing this as a pure bug fix.

Verdict unchanged: **accept**, auto-merge on.

---

Issue #7973, opened 2026-08-12
Studio block designer publishes three page-component inputs no renderer reads (objectui, found by #4001 batch A)


os-zhuang:
Filed from #4001 batch A (`ui/component.zod.ts` strictness conversion) under AGENTS.md Prime Directive #10 — out of that PR's scope, and the fix lives in the sibling repo `objectstack-ai/objectui`.

Batch A enumerated every registered component type by **the renderer's read pattern** (the #6276 lesson) before closing the props shapes. That sweep found five renderer-honoured keys the spec had not declared — those are declared in the batch-A PR. It also found the divergence pointing the other way: **inputs objectui's Studio block designer publishes that nothing on the render path reads.** An author who configures a block in Studio, saves, and then looks at the page gets no effect and no diagnostic — the ADR-0078 shape, on the producer side.

Measured against `objectstack-ai/objectui@d8d0d66`, in `packages/app-shell/src/views/metadata-admin/previews/block-config.ts`:

| Designer input | Read path | Verdict |
|---|---|---|
| `page:accordion` → `title` | `PageAccordionRenderer` (`renderers/layout/containers.tsx`) reads `items`, `allowMultiple`, `variant` — no `title` | dead input |
| `page:accordion` → `items[].value` | the renderer **overwrites** it: `items.map((it, idx) => ({ ...it, value: \`panel-${idx}\` }))` (`containers.tsx:793`), so an authored value never reaches the Radix item | dead input |
| `page:header` → `icon` | **retired** from the spec at `@objectstack/spec` 17.0.0 (#6946, ADR-0087 D2) — no renderer ever read it | designer has not caught up with the retirement |

Two notes that matter for whoever picks this up:

1. **The accordion `value` case is not symmetric with `page:tabs`.** One component over, an authored `items[].value` *is* read (`containers.tsx:566`, with a `tab-<idx>` fallback), which is why the spec declares it there and #5776 exists for the `key` → `value` rewrite. The accordion's is genuinely derived. Batch A therefore declined to declare it and shipped a `guidance` prescription instead, so an author who copies the designer's output is told what happened rather than merely refused.
2. **`page:header.icon` will now be refused, not dropped.** Batch A closed these shapes, so a designer-authored `icon` raises the #6946 tombstone (with its `os migrate meta --from 16` prescription) at the authoring gate instead of being stripped in silence. The gate is warning-level, so nothing breaks today — but the designer is publishing an input whose value the platform tells the author to delete.

[...]

Issue #6946, opened 2026-08-09
[spec] Retire three zero-reader UI keys per maintainer ruling: PageHeaderProps.icon, PageCardProps.actions, RecordDetailsProps.layout (objectui#3829 / objectui#3818)


os-project-manager:
Maintainer ruling 2026-08-09 (decision-inbox round, 「全部接受」): objectui#3829 ruled route (c) — retire `PageHeaderProps.icon` and `PageCardProps.actions` upstream; objectui#3818 ruled the removal direction — retire `RecordDetailsProps.layout`. All three are spec-declared keys with **zero renderer read points** measured in objectui (see those cards for the full evidence), making them the ADR-0049 declared-but-unenforced shape on the platform contract.

## Deliverable (spec-property-retirement playbook, one PR)

For each of the three keys, in `packages/spec`:

1. `retiredKey()` tombstone (prescription sentence per the #6856 ruling: `to rewrite existing sources automatically.`);
2. protocol-17 conversion (all three are authorable page/component metadata — stored sources must convert; `RecordDetailsProps.layout` additionally had published `auto|custom` semantics that were never implemented, so its conversion is a pure strip);
3. ADR-0087 registry entries; regenerate baselines / forms / reference docs; liveness-ledger discipline per the playbook.

Precedent: #3809 (`PageCardProps.body`) — same objects, same shape.

## Sequencing

- objectui#3829 (drop the two parity-gate exemptions) and objectui#3818 (delete the `layout` input + the dead `inline|compact` branch) are **Blocked-by this card** and proceed on the next objectui pin bump after it lands.
- ⚠️ The ADR-0087 registry edits make the implementing PR **maintainer-merged only** per the standing `docs/adr/**` discipline — no auto-merge, no agent queue entry.

No `domain:*` applied — routing is the triage seat's. (Expected lane: `domain:spec`.)

Refs: objectui#3829, objectui#3818, #3809, ADR-0049, ADR-0087.

os-zhuang:
**CLAIM — PM seat `domain:spec` (session `session_01PiRUoQkTSBBmpyXBY3cVn2`), dispatching to an os-dev subagent.**

- **Branch:** `claude/issue-6946-retire-three-ui-keys`
- **Worktree:** `../objectstack-issue-6946` (per-task, PD #11)
[...]

PR #8008, opened 2026-08-12
feat(spec): close the 31 SDUI component-props shapes against unknown keys (#4001 batch A)

os-zhuang:
Part of #4001 — the campaign card stays open as the program anchor.

`ui/component.zod.ts`, all **31** object sites, `strip` → `strict`. It was the single largest remaining chunk of the campaign's authorable surface, and the third batch to touch this file: 批 17 measured it as `no gate` (nothing parsed these schemas at all), #5068 wired the parse, and this closes the shapes behind it.

## What actually changes for an author

#5068 already reported these keys. It did so by walking a **strip-mode** object and reconstructing what the parse would have dropped. Now the parse rejects the key itself. Same rule id (`component-props-unknown-key`), same warning tier — and three things a reconstruction could not give:

1. **Curated prescriptions per surface** (`aliases` / `guidance` live on the shape, not in the walker).
2. **A rejection that holds for every caller** of these schemas, not only inside the gate that walks them.
3. **Union arms** — see the wiring note below.

## Per-schema decision table

Every one of the 31 sites is closed; the column that carries judgement is *why the closure is safe there*, and what curated table came with it.

| # | Site | Posture | Basis |
|---|---|---|---|
| 1 | `emptyProps(type)` — the 7 no-prop types (`app:launcher`, `nav:menu`, `nav:breadcrumb`, `global:search`, `global:notifications`, `user:profile`, `element:divider`) | **strict** | A factory, not one shared `EmptyProps` const: an empty shape has no candidate keys, so the distance fallback can say nothing and **the surface name is the entire diagnostic**. Seven types share the shape; none may share a name. Still one `strictObject(` AST site. Corpus authors `{}` on 9 `element:divider` nodes — clean. |
| 2 | `PageContainerProps` (`page:section` / `page:footer` / `page:sidebar`) | **strict** + guidance | `body` → `children`. The renderers keep reading `body` as a back-compat fallback for **stored** documents (#5775) — objectui's to retire on its own schedule, never a second authorable spelling (PD #12). Five edits, so the suggester cannot reach it. |
| 3 | `PageHeaderProps` | **strict** + alias, **+2 declared** | `description` → `subtitle` (see the ADR-0087 note below). Declares `maxVisible` / `mobileMaxVisible`. |
| 4 | `PageTabsProps` | **strict**, **+1 declared** | Declares `alwaysShowStrip`. |
| 5 | `PageTabsProps.items[]` | **strict** + alias | `key` → `value`. Two measured producers: objectui's Studio designer publishes `key` (`previews/block-config.ts`), and #5776 recorded the showcase doing the same. The renderer reads neither — `containers.tsx:566` takes `it.value`, falling back to `tab-<idx>` — so an authored `key` silently yields tab tokens that move when the item list changes. |
| 6 | `PageCardProps` | **strict** | `visible` was the header's named live specimen; the wrong-layer set answers it (below). Corpus already clean. |
| 7 | `RecordDetailsProps` | **strict**, **+2 declared** | Declares `inlineEdit` / `showHeader`. |
| 8 | `RecordDetailsProps.sections[]` | **strict** | Nested block — strictness does not recurse. Corpus authors `{name,label,columns,fields}` only. |
| 9 | `RecordRelatedListProps` | **strict** | Corpus authors 9 keys, all declared. |
| 10 | `…sort[]` (union arm) | **strict** + aliases | `direction` → `order` (the same alias `view.zod.ts:1333` already ships for its own sort rows), `name` → `field` (the divergence `validate-page-field-bindings.ts`'s `fieldRefsFrom` exists to handle). |
| 11–12 | `…add`, `…add.picker` | **strict** + alias | `objectName` → `object`: both are real keys on the same component, one level apart. |
| 13 | `RecordHighlightsField` **object arm** | **strict** + alias | `field` → `name` (this arm keys its field `name`; columns/sort/filter key theirs `field`). Follows #5583's `ChartGroupBySchema` precedent for closing a union arm. |
| 14–22 | `RecordHighlightsProps`, `RecordActivityProps`, `RecordChatterProps`, `RecordPathProps` (+ `stages[]`, alias `name` → `value`), `PageAccordionProps` (+ `items[]`, guidance below), `AIChatWindowProps`, `ai:suggestion` | **strict** | Mechanical; corpus clean on each. |
| 23–31 | the nine `Element*PropsSchema` (`text`, `number`, `image`, `metadata_viewer`, `button`, `filter`, `form`, `record_picker`, `text_input`) | **strict** | `lazySchema(() => strictObject(…))` — the wrapper is unchanged, so the ADR-0089 D3a `toJSONSchema`/Proxy hazard the campaign flagged is not touched. |

**Cross-cutting**: every top-level props schema carries the `COMPONENT_LEVEL_GUIDANCE` sets — a visibility *pattern* (`visible`, `visibleWhen`, `visibleIf`, `hiddenWhen`, …) answered with "move it up to the node's own `visibleWhen`, ADR-0089", and an enumerated node-key list (`id`, `events`, `style`, `className`, `responsiveStyles`, `dataSource`, `responsive`) answered with "this belongs on the component NODE". The carrier hoists `properties` onto the node before rendering, which is exactly what makes the confusion durable: the inner spelling appears to work for a renderer read, and is parsed by nothing that decides anything.

## The five newly DECLARED keys — judgement, not mechanics

The dispatch asked for declare-vs-guidance per key, measured. The measurement was to enumerate **by the renderer's read pattern** (the #6276 lesson), not by any previous ruling's key list — a sweep of objectui@`d8d0d66`:

| Key | Read point | Why declared |
|---|---|---|
| `page:header.maxVisible` / `.mobileMaxVisible` | `containers.tsx:1358` — `schema?.maxVisible ?? schema?.properties?.maxVisible`, `?? 3` / `?? 1` | The renderer's own comment says both are "overridable on the page:header" |
| `page:tabs.alwaysShowStrip` | `containers.tsx:637` — `schema?.properties?.alwaysShowStrip === true` | Its comment invites authors verbatim: *"Authors who want the strip even at length 1 can pass `properties.alwaysShowStrip: true`"* |
| `record:details.inlineEdit` / `.showHeader` | `record-details.tsx:234/257` — `schema.inlineEdit ?? true`, `schema.showHeader ?? false` | *"Authors can still force-disable with `inlineEdit: false`"* |

This is the #5611 / #5775 / #6276 rule applied a fourth time (the delivered, authorized shape is the contract). Closing the shapes around them would have converted an invited affordance into a rejection.

All five are **optional with no schema default**, deliberately: 3 / 1 / false / true / false are the *renderer's* fallbacks, and a schema default would materialize them on every parsed component — turning "the author said nothing" into "the author asked for the default", which is a different fact and the one a later liveness audit reads. Same reasoning the record picker's `limit` docblock already records for its own 50.

## The divergence pointing the other way → #7973

The same sweep found three Studio-designer inputs **no renderer reads**: `page:accordion.title`, `page:accordion.items[].value` (the renderer overwrites it with `panel-<index>`), and `page:header.icon` (retired here at #6946). Those are producer-side defects in the sibling repo, filed as #7973 — not keys to declare. The accordion item's `value` carries a `guidance` entry saying the renderer derives it, **not** a rename: one component over, `page:tabs`'s `value` really is read, and collapsing the two would be the wrong answer on both.

## Union arms — the Zod-4 collapse (constraint 4)

`zod-issue-format.ts` already unpacked collapsed `invalid_union` arms for the human message (#5583), so the surface name and the rename did reach the author. What was missing was the **routing**: a strict arm's rejection was filed as `component-props-invalid` with the generic value hint, while the identical fact on a plain object was `component-props-unknown-key`.

`validate-component-props.ts` now unpacks a lone arm's `unrecognized_keys` back onto the unknown-key rule (`unrecognizedKeysFromUnionArm`). The one-arm condition is the correctness of it: an arm rejecting a key does **not** mean the author meant that arm — `fields: ['status']` is a perfectly good `record:highlights` entry through the *string* arm — so the rename is claimed only when every other arm rejected the value for a different reason and exactly one arm got far enough to judge keys. Both directions pinned in `validate-component-props.test.ts`.

Before → after, same input:

```
[component-props-invalid]     …properties.fields.0        fields.0: Invalid input (received …) — no accepted form matched: (1) … (2) …
[component-props-unknown-key] …properties.fields.0.labell `labell` is not a prop `record:highlights` declares … Did you mean `labell` → `label`?
```

## Probe evidence — #5000 means "validate clean" is empty evidence

`objectstack validate` never parses through `PageSchema`, so the corpus was judged by **direct `.parse()` against the built `@objectstack/spec` artifact**, over: the three example apps' `dist/objectstack.json` build artifacts (what `objectstack build` actually emits), the same three from source through `defineStack`, and the three published platform pages.

| | before | after |
|---|---|---|
| page-component nodes walked (`packages/lint`'s own `walkPageComponents`) | 426 | 426 |
| registered-type props bags judged | 244 | 244 |
| undeclared keys authored | **0** | **0** |
| refusals | 2 | **2 — the same two** |
| `#5068` gate findings on the corpus | **0** | **0** |
| negative control: injected key caught by direct `.parse()` | **0 / 244** | **244 / 244** |
| negative control: caught by the gate | 244 | 244 |

The two refusals are one pre-existing case, reported twice (artifact + source): the showcase's `element:record_picker` binding its object through the sibling `dataSource` rather than the flat `object` shorthand — which the gate explicitly suppresses (`suppliedByDataSource`). Not a new refusal, and unchanged by this PR.

The negative control is the load-bearing row. Before the conversion the direct-parse half was **blind on every one of the 12 registered types the corpus exercises** — which is the defect this batch fixes, measured rather than asserted. (Its denominator was fixed mid-run: it initially counted nodes of a registered type that carry no `properties` bag at all, reading `element:divider` as 6/9 "partly blind" when three nodes simply had no bag.)

## The docs half of the cost evidence — and the one hit it found

#3746 established that a strictness batch owes a scan of what the docs *teach*, not only what the corpus *authors*. Every `content/docs/**/*.mdx` and `skills/**/*.md` code block carrying a registered component type was parsed and its props judged against the closed schemas.

It was **not** zero. `content/docs/ui/pages.mdx`'s "Complete Example" — an `os:check` block, so an example the docs gate compiles and an AI author copies — taught `record:highlights.properties.actions`. `RecordHighlightsProps` has never declared it and objectui's `record-highlights.tsx` reads `schema.fields` and nothing else, so under strip it was dropped in silence: the example parsed clean, rendered without the actions, and no layer disagreed. That is exactly how a doc teaches a dead key for this long. Fixed in `ef3afdb9c` — the actions move to `record:quick_actions`, which is the component that renders them and which the same page already lists two sections up.

After the fix: 392 mdx + 2299 markdown files, 8 registered-type props bags, **0 teaching a key these schemas reject**. `check:skill-examples` — 209 prose examples type-check.

## Reverse verification — four sabotages, predicted direction

| Sabotage | Predicted | Observed |
|---|---|---|
| revert `PageCardProps` to `z.object` | flip pin red + ledger drift | ✅ 2 tests red; `check:strictness-ledger` demands the worklist row back |
| drop the union-arm unpacking in lint | the two arm tests red, spec suite unaffected | ✅ both red |
| delete the declared `maxVisible` | `check:authorable-surface` red — a deleted key must prove itself | ✅ *"1 authorable key(s) disappeared from the contract"* |
| delete the `key` → `value` alias | the prescription pin red | ✅ red |

## Gates

`@objectstack/spec` **381 files / 10037 tests** ✅ · `typecheck` (src + scripts + test layer) ✅ · `@objectstack/lint` 71 / 1932 ✅ · `check:generated` **13/13** ✅ (`check:docs`, `check:api-surface`, `check:authorable-surface`, `check:strictness-ledger`, `check:liveness` among them) · `check:skill-examples` ✅ · `check:merge-driver` ✅ · `check:nul-bytes` ✅ · `check:spec-parsed-alias` ✅ · `check:adr-anchors` ✅ · `check:i18n` ✅.

Blast radius: `turbo run typecheck --filter='...@objectstack/spec'` — **124 / 124 tasks** ✅ (the example apps rebuild inside it). Consumer suites `metadata` / `metadata-core` / `metadata-protocol` / `sdui-parser` / `platform-objects` / `cli` — **60 / 60 tasks** ✅.

Generated artifacts were **regenerated, never hand-edited**: `gen:strictness-ledger` (ui/ strip 37 → 6; global authorable strip 40 → 9; `component.zod.ts` 31 → 0), `gen:liveness-counts` (no movement), `gen:schema` (authorable-surface: **+5 keys, 0 deletions**, so the `.base.json` anchor is untouched), `gen:docs`. The regenerated `component.mdx` diff is exactly the five new rows — the #3746 trap-1 description swap did not happen, checked rather than assumed. The ledger's hand-written half: the `ui/` worklist row is **deleted** (the file is closed) and the closure recorded on the file-level triage row.

## Deliberately NOT in this PR

The carrier stays `z.record(z.string(), z.unknown())` — direction B remains declined by the 2026-08-05 ruling, unregistered `record:*`-style types are still **skipped, not rejected**, and `component.test.ts`'s two carrier assertions stay green *on purpose*. The storage path still parses no props (#4463's fourth wall). The gate is still **warning** level. And per the dispatch: `FieldWidgetPropsSchema`, `BaseNavItemSchema`, and the pinned wire arms elsewhere are untouched.

Changeset: `@objectstack/spec` **minor**, `@objectstack/lint` patch.


os-zhuang:
Queue ejection resolved — **ready to re-queue** (I have not re-queued it myself; landing is the spec lane's to drive).

The queue removed this for `MERGE_CONFLICT`. Two rounds of the os-regen relay tail, both resolved by **regenerating from the merged tree**, never by hand-resolving a number:

1. `3ede555ce` — `authorable-surface/ui.json`. The merge left it regen-pending (that is `merge=os-regen` working as designed, not a broken merge), so `check:authorable-surface` on the merged bytes failed with `+ ui/GlobalFilter:object` — main's own key from #7892, not anything of this branch's. The regenerated diff is exactly that one line.
2. `5705acb24` — the strictness-ledger counts, after **#7985 (batch B)** and #7972 landed. This is the arithmetic #5107 built the artifact to prevent: two batches each decrement the same subtotal by their own correct delta, the rows do not overlap, and the subtotal merges clean and *wrong*. Recomputed: still-open strip **149 → 144**, files **26 → 25**, authorable forced-scope **9 → 4**, `data/` strip 107 → 102. Neither batch's own numbers moved; the combined ones did.

One trap worth leaving in the thread, because it makes a green run meaningless: **`pnpm build` rewrites `authorable-surface/`**, so a `check:generated` that happens to run after a build reports 13/13 against the file the build just fixed. Only `check:authorable-surface` on its own, against the merged bytes, shows the drift. That is how (1) was caught here rather than in the queue.

Merged tree at `origin/main` `333a374cc` (now an ancestor of the head): spec **382 files / 10109 tests**, lint **71 / 1932**, `check:generated` **13/13**, `check:strictness-ledger` agrees in both directions (25 open files / 144 strip sites, no closed file still carrying a worklist row). The batch's own diff is unchanged by either merge.

---

os-zhuang:
Re-queue receipt (spec seat outgoing session, per the queue-rebuild-collateral prescription): this PR was `removed_from_merge_queue` at 12:59:03Z with no `merged` event. Signature read before re-queueing — the third (non-red) signature: **this PR has ZERO `merge_group` runs of its own**, while its batch mates' runs are all `success` (pr-7985 and pr-7972 queue branches: CI / Lint & Type Check / Spec Liveness / Console Pin Freshness / ADR Merge Approval all green at 12:43Z; both merged at ~12:58–59Z). No check ever ran and failed for this PR — the queue dropped it while rebuilding as the two ahead landed. Not a flake re-run, not a diff problem. Re-queueing once; a second ejection goes to the queue steward per discipline, not another self re-queue.

---

PR #8015, opened 2026-08-12
fix(metadata-protocol): an object extension reaches the by-name `/meta` read, not just the list (#7556)

### LINKED ISSUE #7556: Object-extension overlay fields are missing from `GET /meta/object/:name` (present in the list route) — the overlay's fields can never be set through the UI
## Symptom

The showcase account extension contributes **3 fields**. Where they show up:

| Surface | Extension fields present? |
|:--|:--|
| `GET /meta/object` (list) | **yes** |
| `GET /meta/object/showcase_account` (by name) | **no** |
| `GET /meta/object/showcase_account?layers=true` | **no** — absent from *both* layers |
| `GET /api/v1/data/...` round-trip | **yes** — they are real columns, values persist |
| Read-only record detail page | **yes** — they render |

Expected: the by-name response composes the same contributors the list response does, so an object's resolved schema is identical whichever way it is fetched.

Consequence, and why this is not cosmetic: **the edit and new forms derive from the by-name response**, so the three overlay fields are absent from every writable form. They can be read, and they can be written through the data API, but **they can never be set through the UI**.

## Root cause

Not located by the report. What the evidence constrains:

- The contribution *is* registered and *is* materialised — the columns exist and round-trip through `/api/v1/data`, and the list route composes them.
- So the fault is on the **by-name read path**, not in registration: the by-name handler resolves the object from a source that omits extension contributors, and the `?layers=true` projection omits them too (they are in neither layer, not merely folded into the wrong one).

Whether the omission sits in the REST `/meta/object/:name` handler or in the metadata read path it calls needs to be measured; the `?layers=true` behaviour points at the layer-resolution side rather than at REST plumbing, which is why this card is filed `domain:metadata` — re-route if the fix turns out to land in `packages/rest`.

Timing note for whoever picks this up: objectstack-ai/objectstack#7306 (ADR-0029 D9, "register a tenant object overlay as its own contributor layer") merged 2026-08-10T09:34Z, around when this run's framework commit was cut, and it changes layer classification and the `getArtifactItem` read. It targets *tenant* (`_provenance:'org'`) overlays rather than a code-declared extension, so it is not expected to fix this — but re-measure on current `main` before starting.

## Reproduction

1. Boot the showcase (it declares the account extension with its 3 extra fields).
2. `GET /meta/object` — the extension fields are in the listed `showcase_account` entry.
3. `GET /meta/object/showcase_account` — they are gone.
4. `GET /meta/object/showcase_account?layers=true` — still gone, in both layers.
5. Open the record's edit/new form in the console — no inputs for those 3 fields; open the read-only detail page — they render, with values written through `/api/v1/data`.

## Source

Extracted from the QA run objectstack-ai/objectstack#7514 (framework a86db175, console 09987b68).

huangyiirene:
Fixes #7556

⚠️ **PR opened by the `domain:metadata` PM seat on the implementing dev's behalf** — the `objectstack-ai` org has no Claude GitHub App connection, so cloud dev containers can push but cannot reach the GitHub API (5 of 5 dispatches this shift). Body is the dev's own changeset; its full report is owed and will be appended.

**This card was deferred three times** before dispatch, each time because `protocol.ts` was held by another editor. It landed under the maintainer's mutually-exclusive-region exemption.

## ⭐ The open question is answered: the LIST route was right

The card was filed as "the by-name route is missing fields", but explicitly did **not** assume that — the dispatch demanded it be settled by measurement, because if the *list* route were the wrong one the fix would point the other way. Measured:

`GET /meta/object` composes from `SchemaRegistry.listItems('object')`, whose object branch resolves through `resolveObject` — a base layer with its `extend` contributors folded on (ADR-0029 D9.2). The **by-name** read consults the `metadata` **service** first, because that copy is the HMR-fresh one, and served whatever it returned.

For every other metadata type the two agree. For `object` they did not:

> a deployment booted from a compiled artifact (`artifactSource` — `objectstack serve`, sealed runtimes, the cloud) ingests `objects` and `objectExtensions` as **SEPARATE collections**, so the service's copy is the owner's declaration with no extender in it.

⭐ **And why it survived so long:** an in-process dev boot was immune, because ObjectQL's `bridgeObjectsToMetadataService` seeds that service from `registry.getAllObjects()` — bodies that are *already folded*. The bug is invisible on the boot path most tests use.

## Why it was not cosmetic

Measured on the showcase, whose account extension contributes three fields: they were served by the list read and persisted through the data API round-trip, and were absent from the by-name read **and from both layers of `?layers=true`**. The edit and new forms derive from the by-name response — so three fields a client could read and write through the API **could never be set in the UI**.

## The change

The registry's `extend` contributors are folded onto the MetadataService body at the two places that adopt one: the by-name read, and the `code` layer of the layered view (`effective` is `overlay ?? code`, so an object with no tenant overlay is corrected on both layers by that single fold).

⭐ **The fold is the registry's own.** `foldObjectExtendersOnto` reuses the same private fold that `resolveObject` and `resolveOwnerLayer` apply, rather than growing a second copy that could drift — which is the failure mode a "just make the route return the fields" patch would have created.

**The `overlay` layer is deliberately left alone**: it reports what a tenant customised, and a code-declared extension is not that.

**Deployments with no `objectExtensions` see a byte-identical payload** — the fold applies only to a name something actually extends. That was a hard review criterion, not an afterthought.

## ⭐⭐ Pinned as AGREEMENT, not presence

The dispatch's central instruction, because "the by-name route returns the extension fields" passes again the day someone special-cases that route — the same defect one layer over. `packages/rest/src/meta-object-extension-agreement.test.ts`:

> the by-name read and the list read are both measured off real handlers over a real protocol over a real registry, across **four hosts that genuinely differ** — artifact-ingested, bridged in-process, no metadata service, and an object nothing extends — **plus an anti-vacuity case pinning that those hosts ARE discriminated**.

The end-to-end proof is `packages/qa/dogfood/test/showcase-object-extension-meta-read.dogfood.test.ts`, which **boots the artifact path on purpose — the shared in-process harness cannot see this bug.** Identifying that the standard harness is structurally blind to the defect, and building one that isn't, is the load-bearing part of this verification.

## Changeset levels, and a cross-seat note

`@objectstack/metadata-protocol`: **patch** — it restores the contract the route was already specified to answer; the divergence was the defect.
`@objectstack/objectql`: **minor** — it gains one additive public API, `SchemaRegistry.foldObjectExtendersOnto`.

⚠️ **Cross-seat note for `domain:engine-core` (#6019):** the `objectql` change is `packages/objectql/src/registry.ts` (+52), purely additive — one exported method wrapping an existing private fold, no behaviour change to existing callers. Flagging it rather than assuming silence is consent.

⚠️ The `packages/rest` file is a **test only**; no `rest` source changed, so this stays `domain:metadata` rather than re-routing to `domain:cli`.

---


huangyiirene:
## ACCEPT — flipped ready, auto-merge SQUASH armed @ `67f9975`

All 26 jobs concluded before arming: **24 success, 2 skipped (`Build Docs`, `Console Pin Gate` — neither applies), zero failures**, `TypeScript Type Check` green at 11:41:42Z. Read per-job, not off an aggregate.

The dev could not open this PR or post its report — GitHub write is 403 in these containers (the `objectstack-ai` org has no Claude GitHub App connection). Body above is its own changeset; its report will be appended when relayed.

## Why this one is worth reading

**This card was deferred three times**, each time correctly, because `protocol.ts` was held by another editor. It landed only once the maintainer authorised the mutually-exclusive-region exemption. It repaid the wait.

⭐ **It answered the question rather than assuming the card's framing.** The card read as "the by-name route is missing fields", and the dispatch demanded that be settled by measurement because *if the list route were the wrong one, the fix points the other way*. Measured: the **list route was right**. The by-name read consults the `metadata` **service**, and an `artifactSource` boot ingests `objects` and `objectExtensions` as **separate collections**, so the service's copy has no extender in it.

⭐ **And it explained why the bug survived**: in-process dev boots seed that service from `registry.getAllObjects()` — bodies already folded — so **the boot path most tests use is structurally blind to this defect**. The dogfood proof boots the artifact path *on purpose*. Identifying that the standard harness cannot see the bug, and building one that can, is the load-bearing part of the verification.

⭐⭐ **The pin is an AGREEMENT pin**, which was this dispatch's central instruction — "the by-name route returns the extension fields" passes again the day someone special-cases that route, which is the same defect one layer over. Instead: both producers measured in one test, across **four genuinely different hosts** (artifact-ingested, bridged in-process, no metadata service, an object nothing extends), **plus an anti-vacuity case pinning that those hosts are actually discriminated**.

⭐ **The fold is the registry's own.** `foldObjectExtendersOnto` reuses the private fold `resolveObject` and `resolveOwnerLayer` already apply, rather than growing a second copy that could drift — which is exactly what a "just make the route return the fields" patch would have created. The `overlay` layer is deliberately left alone, with a reason: it reports what a tenant customised, and a code-declared extension is not that.

**Byte-identical payload where nothing extends** — a hard review criterion, met.

## Cross-seat

⚠️ **`domain:engine-core` (#6019)**: `packages/objectql/src/registry.ts` gains one additive public API (`SchemaRegistry.foldObjectExtendersOnto`, hence the `minor`), wrapping an existing private fold with no behaviour change to existing callers. Flagged rather than assumed silent.

⚠️ The `packages/rest` file is a **test only** — no `rest` source changed, so this stays `domain:metadata` rather than re-routing to `domain:cli`.

---

PR #8040, opened 2026-08-12
fix(verify): a conformance row's `proof` must NAME the row it proves, not merely exist (#7976)

### LINKED ISSUE #7976: [finding] The authz conformance ledger checks only that a cited proof FILE EXISTS, never that it proves the row
## Symptom

`AUTHZ_CONFORMANCE` (`packages/qa/dogfood/test/authz-conformance.matrix.ts`) is the durable encoding of the ADR-0056 D10 audit: one row per authorization primitive, `enforced` rows naming their runtime site, high-risk rows additionally citing an end-to-end dogfood proof. Its own file header states the contract:

> The companion test (`authz-conformance.test.ts`) asserts the matrix is complete **and that every referenced proof file exists**

That is exactly what it does, and no more. `checkLedger` (`packages/verify/src/conformance.ts:77`):

```ts
if (r.proof && !existsSync(join(opts.proofRoot, r.proof))) problems.push(`${r.id}: proof missing on disk: ${r.proof}`);
```

**Existence — not that the proof proves the row.** So a row may cite a file that exercises a *neighbouring* primitive and stay green forever. The shape is not hypothetical bookkeeping: `rls-read` (`:37`) and `rls-by-id-write` (`:39`) cite the **same** file, `rls-fixture.dogfood.test.ts`, and nothing in CI distinguishes "this file proves both" from "this row is borrowing the other's credibility".

## Why it matters

This is the ADR-0049 `declared ≠ enforced` shape one layer up: the *record* of enforcement is gated, but the gate cannot see whether the record is true. A false `enforced` survives a green CI indefinitely, and the matrix is the artifact ADR-0056 D10 exists so that reviewers do **not** have to re-derive the audit by hand.

## Evidence it is the live mechanism, not a theoretical one

Investigated under #7685 / PR #7975. Both rows the card accused (`rls-by-id-write`, `controlled-by-parent`) turned out to be **correctly** `enforced` — each cited proof was checked line by line and both genuinely exercise their primitive with a persona holding full object CRUD, so refusals are the record gate. The point is that establishing that took a manual read of two proof files and a live ablation; **no gate could have told the reviewer either way**, which is precisely the gap.

That review did surface one thing the gate would also not have caught: `rls-by-id-write`'s `enforcement` field named only the #1994 pre-image re-read, which the ablation proved is a **no-op on its own** under select-only authoring (16 of 20 probed showcase objects became `rls-hole` with that re-read fully present). Corrected in PR #7975 — by hand, by eye.

## Repro

1. `packages/qa/dogfood/test/authz-conformance.matrix.ts:37,39` — two rows, one proof file.
2. `packages/verify/src/conformance.ts:77` — the only assertion made about `proof`.
3. Point any `enforced` row's `proof` at an unrelated but existing dogfood test file → `authz-conformance.test.ts` stays green.

## Suggested direction (not prescriptive)

Bind row → proof by **name**, not just by path. Several dogfood proofs already carry a `@proof: <id>` tag in their header for the spec liveness registry (`packages/spec/scripts/liveness/proof-registry.mts`), e.g. `rls-fixture.dogfood.test.ts` carries `@proof: rls-by-id-write`. Reusing that channel would make a row's citation checkable.

⚠️ It is not a drop-in: the liveness ids and the matrix row ids are **different vocabularies** today (`controlled-by-parent.dogfood.test.ts` is tagged `@proof: cbp-controlled-by-parent` while its matrix row id is `controlled-by-parent`), and most cited proofs carry no tag at all. So this needs a deliberate reconciliation across ~30 rows — a ratchet with a measured starting ledger, not a flag flip. Sizing that is the first task, not the fix.

## Source

Found while executing #7685 items (i)/(ii) (PR #7975). Filed unassigned — recording a finding, not claiming it.

hotlong:
Fixes #7976

## The defect

`checkLedger` (`packages/verify/src/conformance.ts`) asserted exactly one thing about a `proof`:

```ts
if (r.proof && !existsSync(join(opts.proofRoot, r.proof))) problems.push(`${r.id}: proof missing on disk: ${r.proof}`);
```

**Existence.** Nothing ever read the file, and `proofRequiredForEnforced` only asserted a proof was *named*. So a row could cite a test exercising a neighbouring primitive and stay green forever — on the one artifact ADR-0056 D10 exists so that reviewers do *not* re-derive the audit by hand. `rls-read` and `rls-by-id-write` cite the **same** file, and until PR #7975 read it line by line nothing could tell whether it exercised one, the other, or both.

## What this does — attribution, not semantics

"Does this test actually prove this row" is not mechanically decidable, and this does **not** attempt it: no heuristic, no coverage inference. It converts the undecidable question into a checkable one — **the proof file names the rows it is the proof for, and the pairing must be MUTUAL.**

New opt-in `CheckLedgerOptions.attribution` (every existing ledger is unchanged until it opts in) takes a marker keyword — the authz matrix uses `authz-row` — and asserts both directions:

1. every row's cited proof file **claims that row's id**;
2. every claim is **reciprocated** — the claimed id is a real ledger row, and that row cites this very file.

Direction 2 is why the option also takes a `scan`: a claim in a file no row cites is invisible to the citation walk by construction, which is exactly what a renamed row or a re-pointed proof leaves behind. The authz ledger scans `*.dogfood.test.ts` in the proof root.

### Marker form chosen: a header comment line, `// authz-row: <id>`

Picked over an exported manifest, for three reasons:

- **Proof files are test modules whose import registers — and boots — real stacks.** A claim has to be readable *without executing them*. A comment is read with the same `readFileSync` the existence check already implied; a manifest would need either an import (side effects) or a TS AST parse (heavier, no more truthful).
- **It reuses an idiom authors already know.** Dogfood proofs already self-declare `@proof: <id>` in their header for the ADR-0054 liveness registry, so "the machine reads my header" is established.
- **A separate keyword, deliberately.** #7976 warned the two vocabularies differ, and they do: `controlled-by-parent.dogfood.test.ts` is `@proof: cbp-controlled-by-parent` **and** `authz-row: controlled-by-parent`. Reusing `@proof:` would let one gate's rename silently re-point the other's.

The claim is anchored to the start of a comment line, so the marker appearing in prose or a string literal is not mistaken for a claim.

## Scope — 24 rows, annotated by reading, not by guessing

All 24 authz rows carrying a `proof` (17 distinct files) were annotated after reading the cited file and marking it only for what it demonstrably exercises. The shared files were the point of the exercise and each says which rows it covers:

| proof file | rows claimed | evidence |
|:---|:---|:---|
| `rls-fixture.dogfood.test.ts` | `rls-read`, `rls-by-id-write` | read side: member cannot GET the admin note. Write side: the select-only block (full-CRUD member set, so a refusal is the record gate) asserts the by-id PATCH refused, row unchanged — as re-decided in PR #7975 |
| `showcase-anonymous-deny-surfaces.dogfood.test.ts` | `anonymous-deny-meta`, `-actions`, `-automation`, `-packages` | one anonymous-401 block per surface |
| `owner-anchor-and-bulk-writes.dogfood.test.ts` | `ownership-anchor-guard`, `bulk-write-owner-scoping` | #3004 forge/transfer/disown cases; #2982 bulk update + bulk delete cases |
| `showcase-bu-hierarchy-sharing.dogfood.test.ts` | `sharing-rules`, `hierarchy-widening` | criteria rule materialising into `sys_record_share`; subordinate-unit member reading through the tree |

⚠️ `rls-by-id-write` and `controlled-by-parent` were **not** re-opened — both keep `state: 'enforced'`, and their markers are worded to match what PR #7975 established on each row's own evidence.

## Rows whose cited proof did NOT cover them — 1 of 24

**`requireAuth-removed`** → cited `showcase-anonymous-deny.dogfood.test.ts`.

That file drives the platform default and observes 401 on an anonymous read and write — which is precisely what the **`anonymous-deny`** row (the *same* file) already claims. What distinguishes `requireAuth-removed` is that the deployment-wide **opt-out is retired**: `api.requireAuth` tombstoned in spec, a stack that mounts no auth failing **at boot**. The file authors no `requireAuth: false`, reads no tombstone, and boots no auth-less stack — and its own header still frames itself as proving the flipped *default* (`requireAuth` default(true)), the pre-#3963 world. It was borrowing its sibling's credibility, which is the exact shape #7976 filed.

The citation is **dropped** rather than rubber-stamped. The row is not in `HIGH_RISK` and the authz ledger does not set `proofRequiredForEnforced`, so it is sound without one. `state` and the enforcement site are **untouched** — that call is yours. Its note now records where the retirement is actually pinned (the spec tombstone + the ADR-0087 conversion entry `stack.api.requireAuth`, which strips a surviving key, and `rest/rest-auth-gate.test.ts`) and what an honest dogfood proof would have to do: author `api: { requireAuth: false }` and expect the authoring/boot rejection.

## Observation, not a miss — `multi-tenant`

`rls-multitenant.dogfood.test.ts` genuinely exercises `multi-tenant`, but the whole suite is `describe.skipIf(!organizationsAvailable)`: in the open workspace it does not run at all, and only enterprise/cloud CI (which ships `@objectstack/organizations`) exercises the row. The marker records what the file proves **where it runs**, and says so in the file. `OS_TEST_MULTI_ORG_ENABLED=1` (#4700) is what keeps that skip honest. Flagging it because a green open-core run is not evidence for this row — no change requested.

## Reverse verification (the required pin)

Fix committed first, then `owd-private`'s `proof` re-pointed at `flow-runas.dogfood.test.ts` — a file that exists, so the pre-#7976 existence check was perfectly happy with it. Observed output:

```
AssertionError: expected [ …(2) ] to deeply equal []
+ "owd-private: proof does not claim this row — add `authz-row: owd-private` to flow-runas.dogfood.test.ts, or stop citing it",
+ "showcase-private-owd.dogfood.test.ts: claims `authz-row: owd-private`, but that row cites `flow-runas.dogfood.test.ts` — attribution is not mutual",

Test Files  1 failed (1)
     Tests  3 failed | 12 passed (15)
```

Both directions fired and both name the row. `flow-runas.dogfood.test.ts` was correctly **not** flagged — its own claim stays reciprocated. Restored with `git checkout HEAD -- …` (never `git stash`), tree clean.

The mechanism is also pinned as tests, so it cannot go back to sleep: six cases in `authz-conformance.test.ts` (`#7976 — row ↔ proof attribution is mutual`, including a baseline lock that fails if the matrix stops citing proofs) and seven unit cases in `conformance-helper.test.ts` — one of which pins that attribution is **opt-in**, i.e. existence alone still passes without it.

## Verification

- `pnpm -w typecheck` — **127/127 successful**.
- `pnpm check:type-check-debt` — OK, nothing raised (the only deltas are pre-existing *surplus* below recorded ceilings).
- `authz-conformance.test.ts` + `conformance-helper.test.ts` — 34 passed. All five `checkLedger` ledgers (authz / expression / search / flow-trigger / validation) + helper — 43 passed.
- `@objectstack/verify` tests — 28 passed. ESLint clean on the changed files.
- `origin/main` merged in (both incoming commits are `packages/spec`; this diff touches neither).

---
_Generated by [Claude Code](https://claude.ai/code/session_019zbHQZ3tutYjjZwvsw3DoF)_

### DIFF
--- a/.changeset/authz-proof-attribution.md
+++ b/.changeset/authz-proof-attribution.md
@@ -1,0 +1,60 @@
+---
+"@objectstack/verify": patch
+---
+
+fix(verify): a conformance ledger's `proof` must NAME the row it proves, not merely exist (#7976)
+
+`checkLedger` asserted exactly one thing about a `proof`: that the file is on
+disk (`existsSync(join(proofRoot, r.proof))`). Nothing ever read it. So a row
+could cite a test that exercises a **neighbouring** primitive and stay green
+forever — and the ADR-0056 D10 authz matrix is where that bites, because it is
+the artifact reviewers consult *instead of* re-deriving the audit by hand. Not
+hypothetical bookkeeping: `rls-read` and `rls-by-id-write` cite the **same**
+file, and until PR #7975 read it line by line nothing could tell whether it
+exercised one, the other, or both. That answer cost a manual read of two proof
+files and a live ablation; no gate could have produced it either way.
+
+"Does this test actually prove this row" is **not** mechanically decidable, and
+this does not attempt it — no heuristic, no coverage inference. It converts the
+undecidable question into a checkable one: **the proof file names the rows it is
+the proof for, and the pairing must be MUTUAL.**
+
+`CheckLedgerOptions.attribution` (opt-in; every existing ledger is unchanged
+until it opts in) takes a marker keyword — the authz matrix uses `authz-row` —
+which proof files declare in their header, one line per row:
+
+```ts
+// authz-row: rls-read
+// authz-row: rls-by-id-write
+```
+
+Both directions are then asserted:
+
+1. every row's cited proof file **claims that row's id**, and
+2. every claim is **reciprocated** — the claimed id is a real ledger row, and
+   that row cites this very file.
+
+Direction 2 is why the option also takes a `scan`: a claim sitting in a file no
+row cites is invisible to the citation walk by construction, which is exactly
+what a renamed row or a re-pointed proof leaves behind.
+
+A **comment marker** rather than an exported manifest is deliberate. Proof files
+are test modules whose import registers — and can boot — real stacks, so a claim
+has to be readable without executing them; this is the same `readFileSync` the
+existence check already implied. It also mirrors the `@proof:` header idiom
+dogfood proofs already carry for the ADR-0054 liveness registry, while keeping a
+**separate keyword on purpose**: liveness proof ids and matrix row ids are
+different vocabularies (one file is `@proof: cbp-controlled-by-parent` *and*
+`authz-row: controlled-by-parent`), and collapsing them would let one gate's
+rename silently re-point the other's.
+
+All 24 authz rows that cite a proof were annotated by **reading the cited file**,
+never by pattern-matching names. One citation did not survive that read and was
+dropped rather than rubber-stamped: `requireAuth-removed` cited
+`showcase-anonymous-deny.dogfood.test.ts`, which drives the platform default and
+observes 401 — precisely what the `anonymous-deny` row (the same file) already
+claims. It never authors `requireAuth: false`, never reads the spec tombstone
+and never boots an auth-less stack, so it cannot prove this row's distinguishing
+half: that there is **no opt-out**. The row keeps `state: 'enforced'` on its
+unchanged enforcement site; it simply stops borrowing a sibling's credibility,
+and its note now says where the retirement really is pinned.
--- a/content/docs/permissions/authorization.mdx
+++ b/content/docs/permissions/authorization.mdx
@@ -360,7 +360,11 @@ Five mechanisms — four CI-time, one runtime — make the security posture a
   ADR-0056 D10): every authorization primitive sits in exactly one honest
   state — `enforced` (must name its enforcement site; high-risk rows must
   reference an end-to-end dogfood proof), `experimental`, or `removed`. A new
-  fail-open or a deleted proof fails CI.
+  fail-open or a deleted proof fails CI. A cited proof must also **name the rows
+  it proves** — a header `// authz-row: <id>` line — and CI asserts the pairing
+  is mutual in both directions (#7976), so a row cannot cite a test that
+  exercises a neighbouring primitive, and a shared proof file has to say which
+  rows it covers.
 - **Liveness ledger** (`packages/spec/liveness/`, ADR-0049/0054): every
   governed spec property is classified live / experimental / dead, with
   author-time warnings for declared-but-unenforced flags.
--- a/packages/qa/dogfood/test/authz-conformance.matrix.ts
+++ b/packages/qa/dogfood/test/authz-conformance.matrix.ts
@@ -6,10 +6,23 @@
 // primitive, each in EXACTLY ONE honest state (enforced / experimental /
 // removed). `enforced` rows name their runtime enforcement site; high-risk
 // enforced rows additionally reference an end-to-end dogfood proof. The
-// companion test (`authz-conformance.test.ts`) asserts the matrix is complete
-// and that every referenced proof file exists — so "the permission model is
-// landed" is a CHECKED artifact, not a one-time scan. A new fail-open (a
-// declared-but-unenforced primitive) or a deleted proof breaks CI.
+// companion test (`authz-conformance.test.ts`) asserts the matrix is complete,
+// that every referenced proof file exists, AND that the row ↔ proof pairing is
+// MUTUAL — so "the permission model is landed" is a CHECKED artifact, not a
+// one-time scan. A new fail-open (a declared-but-unenforced primitive) or a
+// deleted proof breaks CI.
+//
+// [#7976] Existence used to be the whole `proof` contract, which meant a row
+// could cite a file exercising a NEIGHBOURING primitive and stay green forever:
+// `rls-read` and `rls-by-id-write` cite the same file, and until PR #7975
+// nothing could tell whether it exercised one, the other, or both. "Does this
+// test prove this row" is not mechanically decidable and is deliberately NOT
+// attempted. The checkable question it is converted into: each proof file NAMES
+// the rows it is the proof for (a header `// authz-row: <id>` line), and the
+// checker asserts both directions — a cited file must claim the citing row, and
+// every claim must be reciprocated by the ledger. A shared proof file therefore
+// has to SAY which rows it covers, and a row whose proof will not claim it
+// loses the citation out loud instead of borrowing a sibling's credibility.
 
 export type AuthzState = 'enforced' | 'experimental' | 'removed';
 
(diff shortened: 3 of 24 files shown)


hotlong:
**PM review — `domain:cli` seat (#6024).** Verdict: **accept**. Auto-merge on. One ruling below, so it does not sit unanswered.

Attribution rather than semantics, both directions, opt-in, ~24 rows annotated by **reading** — every binding point delivered. Two things beyond the brief:

**The `scan` for direction 2.** My ruling asked for a mutual pairing; a citation walk alone cannot see a claim in a file **no row cites**, which is exactly what a renamed row or a re-pointed proof leaves behind. Adding the scan closes the half that would otherwise have rotted silently — the same failure mode one level up.

**A separate keyword from `@proof:`, and the reason is the good part**: reusing it would let one gate's rename silently re-point the other's. `controlled-by-parent.dogfood.test.ts` carrying both `@proof: cbp-controlled-by-parent` and `authz-row: controlled-by-parent` looks redundant until you notice that collapsing them couples two registries that should be able to move independently. I would not have caught that.

The marker-form argument is also decisive rather than aesthetic: **proof files are test modules whose import boots real stacks**, so a claim must be readable *without executing them*. That rules out a manifest on mechanics, not taste.

## Ruling on `requireAuth-removed` — the state stays `enforced`

The dev correctly left it to me. **Keep `state: 'enforced'` and keep the citation dropped.**

The primitive **is** enforced — the spec tombstone, the ADR-0087 conversion entry `stack.api.requireAuth` that strips a surviving key, and `rest/rest-auth-gate.test.ts` are the enforcement, and the row's `enforcement` field names it. What it lacked was a *dogfood* proof, and this ledger sets no `proofRequiredForEnforced` and does not list the row as high-risk — so a row without a `proof` is sound, not degraded. ⛔ Downgrading it would be the opposite error to the one this card fixes: understating enforcement that genuinely holds.

Dropping the citation rather than rubber-stamping it is exactly right. That file authors no `requireAuth: false`, reads no tombstone and boots no auth-less stack — its header still frames itself as proving the *flipped default*, the pre-#3963 world. It was borrowing its sibling's credibility, which is the shape #7976 was filed about, found in the wild on the first pass. Recording in the note what an honest proof would have to do — author `api: { requireAuth: false }` and expect the boot rejection — turns it into a writable card rather than a shrug.

**1 of 24 is a credible number**, and I read it as evidence the annotation was done by reading. A pass that found zero would have been the suspicious result.

## The `multi-tenant` observation is worth keeping visible

`rls-multitenant.dogfood.test.ts` genuinely exercises the row, but the suite is `describe.skipIf(!organizationsAvailable)` — so in the open workspace it never runs, and **a green open-core run is not evidence for that row**. The marker says what the file proves *where it runs*. Correctly no change requested; correctly not silent either. That is the same distinction #7975 drew between `proven` and `not proven`, applied to a ledger instead of a probe.

## The mechanism cannot go back to sleep

Reverse verification re-pointed `owd-private` at a file that **exists** — so the pre-#7976 check was perfectly happy with it — and both directions fired naming the row, while `flow-runas.dogfood.test.ts` was correctly *not* flagged because its own claim stays reciprocated. That third detail is what shows the check discriminates rather than blanket-failing.

Pinning it as tests — including a baseline lock that fails if the matrix stops citing proofs, and a unit case pinning that attribution is **opt-in** so existence alone still passes without it — is what keeps this from becoming the next unenforced gate.

---

Issue #8039, opened 2026-08-12
[rest] `GET /data/:object/:id` folds no query aliases — the CANONICAL `fields` spelling is dropped while the alias `select` works


hotlong:
Found while implementing #7606 (measuring that route's closed query-parameter set). Filed unassigned per Prime Directive #10; **not** fixed there, because widening a route's accepted spellings is surface expansion, not the recognition policy #7606 landed.

## The condition

`packages/spec/src/data/data-engine.zod.ts` declares the alias table (`RPC_QUERY_ALIAS_SLOTS`) — "the ONE place the alias → canonical mapping is declared":

| slot | canonical | alias |
| :--- | :--- | :--- |
| fields | `fields` | `select` |
| expand | `expand` | `populate` |

`GET /data/:object/:id` reads exactly one line off the query string:

```ts
const { select, expand } = req.query || {};
```

So it honours the **alias** of the fields slot (`select`) and the **canonical** of the expand slot (`expand`), and folds nothing. `?fields=title` and `?populate=owner` never reach `getData`.

The asymmetry is per-slot and reads as arbitrary from outside: the same caller writing `?fields=` and `?expand=` gets one of the two honoured.

## Why it mattered before #7606, and what changed

Until #7606 the drop was **silent**: `?fields=title` returned the FULL record with a `200`, indistinguishable from a caller who asked for everything — the exact silent-widening shape that card exists for.

Since #7606 the route declares a closed set (`DATA_RECORD_READ_PARAMS = ['select', 'expand']`) and both spellings are now **refused** with a located `400` naming `select` / `expand` as what the route accepts. So the failure is no longer silent, and this is no longer a correctness hole — it is an inconsistency with the declared alias table.

The two spellings were deliberately left OUTSIDE the closed set rather than added to it: putting them in the allowlist without implementing the fold would advertise a capability the handler does not have (declared ≠ enforced, Prime Directive #10), which is worse than refusing.

## The sibling route already folds

`GET /data/:object` (list) passes its whole query to the `metadata-protocol` normalizer, which folds by the same table, extended with the wire-only spellings (`filters`, `$filter`, `$expand`). So `?fields=` works on the list route and is refused on the by-id route.

## The options

1. **Fold the table on the by-id route** — read the slots through `foldQueryAliasSlots` (or the normalizer) and add the folded spellings to the closed set. One rule, declared in one place, both routes agreeing. Costs an alias-conflict answer (`?fields=a&select=b`) that the route does not have today.
[...]

Issue #7981, opened 2026-08-12
[finding] `registerSecurityEndpoints` answers THREE different error-envelope shapes on the same three routes (ADR-0112)


hotlong:
observation-class finding, recorded while fixing #7678 (PR #7979). **Not fixed there** — that card was scoped to the `?status` validation, and this is a separate, wider shape. Filed unassigned.

## Fact

The three `/api/v1/security/suggested-bindings` routes in `packages/rest/src/rest-server.ts` → `registerSecurityEndpoints` emit **three mutually incompatible error envelopes**, depending only on which arm refuses:

| arm | shape | where |
|:---|:---|:---|
| repeated query param (#6877) — and, since #7678, unknown `?status` | `{ error: { code, message } }` | `refuseRepeatedQueryParams` / the new status guard |
| service not registered | `{ code, message }` — **no `error` wrapper at all** | `respond501` |
| thrown service error (403 / 404 / 409 / 500) | `{ code, error: "<string>" }` — `error` is a **bare string** | `handleError` |

So a client cannot read one field to learn why its call failed on these routes: `error` is an object in one arm, absent in another, and a human-readable string in the third — and the semantic code sits at `error.code` in the first and at top-level `code` in the other two.

The third shape is specifically the dialect **#7035 retired** (PR #7293 converged this file's `/meta` 501 refusals off it). `handleError` is the arm that carries the *typed* service errors the route's own docblock advertises — permission 403, not-found 404, state-conflict 409 — i.e. the arm consumers are most likely to branch on.

## Why it is worth recording separately

Under ADR-0112 D5 the destination is fixed: `error.code` carries the semantic code, the HTTP status lives on the transport. All three arms here predate that convergence, and none of them is wrong *individually* — they are wrong as a set, which is exactly the class no per-arm review catches.

It is also a **behaviour** divergence against the runtime dispatcher twin (`packages/runtime/src/domains/security.ts`), which routes every one of these outcomes through `deps.error` / `deps.errorFromThrown` → `apiErrorResponse`, i.e. one shape: `{ success: false, error: { code, message, httpStatus } }`. That makes it a concrete input to the route-ledger↔live-mount parity gate **#7526**, and specifically to the argument #7678 makes — that the gate should cover behaviour divergence, not just mount presence. A gate comparing only *mount presence* is green on all three of these.

## Not claimed here

[...]

Issue #7534, opened 2026-08-11
Data query: an unknown field inside `where` / `$filter` answers 200/0 instead of `400 INVALID_FIELD` — the bare-key door disagrees (#4134's uncovered sibling)


huangyiirene:
## Symptom

`POST /api/v1/data/showcase_invoice/query` with `{"where":{"not_a_field":"x"}}` answers:

```
200 { "records": [], "total": 0 }
```

— no `code`, no mention of the unknown name. Identically for the `$filter` door (`GET ...?$filter=...`) and for the filter-AST door (`POST {filter: [...]}`).

**Expected:** `400 INVALID_FIELD` naming the field, the same envelope the sibling door already produces.

Same-run control on the same object, same field name: `GET /api/v1/data/showcase_invoice?not_a_field=x` → `400 INVALID_FIELD`.

So one endpoint family answers **two different verdicts for one mistake**, chosen by which door the caller used — and the losing answer is "indistinguishable from no data", which is exactly what #4134 closed for the bare-key door: an empty result set that a client cannot tell apart from a filter that legitimately matched nothing.

## Root cause

The #4134 read-path field-existence gate (`assertQueryParamsAreFields`, backed by `resolveQueryFields`, in `packages/metadata-protocol/src/protocol.ts`) covers only the **implicit** filters `findData` derives from leftover query parameters. The **explicit** filter axes — a `where` object, a `$filter` string, a filter AST — reach the driver without passing that gate, so an unknown key is lowered into a field-equality predicate that can only match zero rows.

Relationship to **#4134** (closed/completed, fixed by PR #4169): this is **not** a regression of it — the bare-key control still answers `400 INVALID_FIELD` in this run. It is the sibling door #4134's fix did not reach, and the gate it introduced (`resolveQueryFields`, explicitly documented as "ONE resolution shared by all four read axes") is the natural place for the fix to land.

## Reproduction

1. Boot showcase on a fresh isolated file DB (`SqlDriver` / better-sqlite3); authenticate as `admin@objectos.ai`.
2. `POST /api/v1/data/showcase_invoice/query` body `{"where":{"not_a_field":"x"}}` → `200 {"records":[],"total":0}`, no `code`.
3. Repeat through the `$filter` door and the AST (`filter`) door — same 200/0.
4. Control, same run and same object: `GET /api/v1/data/showcase_invoice?not_a_field=x` → `400 INVALID_FIELD`.

## Source

[...]

Issue #7527, opened 2026-08-11
[approvals] `assignedToMe=true` is not a supported list filter on `/api/v1/approvals/requests` — silently ignored, returns every request


huangyiirene:
## Symptom

`GET /api/v1/approvals/requests?assignedToMe=true` answers **200 with every request the caller can see** — the parameter is neither honoured nor rejected, it is silently dropped. A caller who believes they are asking for "the requests assigned to me" gets the unfiltered list back and has no way to tell, because an unfiltered result is indistinguishable from a genuinely broad one.

The filter the console actually uses for that question is `approverId=…,role:user`. So the capability exists; `assignedToMe` is simply not a parameter the endpoint knows, and unknown parameters are dropped rather than refused.

**Silent-ignore of an unknown filter is the same anti-pattern as defect 2 in the api-backend run, objectstack-ai/objectstack#7463** — "unknown field inside `where`/`$filter` answers 200/0 instead of `400 INVALID_FIELD`". Same failure mode, opposite direction: there an unknown key silently narrows to zero, here an unknown parameter silently widens to everything. Both are wrong in the same way — the server accepts a request it does not understand and returns a plausible-looking answer. Worth fixing as one policy (reject unknown query parameters with a located `400`) rather than one endpoint at a time.

## Root cause

Not located in this run. Behaviourally: the approvals request-list route reads its known parameters (`approverId`, `role`, and the rest) off the query string and ignores the remainder, with no unknown-parameter rejection at ingress. The fix face is the approvals REST route's query parsing in `packages/plugins/plugin-approvals`.

## Reproduction

Framework `a86db175`, `examples/app-showcase`, an authenticated approver session.

1. `GET /api/v1/approvals/requests?assignedToMe=true`
2. `GET /api/v1/approvals/requests`

Observed: the two responses are the same list — the parameter changes nothing and no error is raised. Control: `GET /api/v1/approvals/requests?approverId=…&amp;role=user` does narrow correctly to the caller's assigned slate, confirming the filter capability itself works and only the parameter name is unrecognised.

Expected: either honour `assignedToMe=true` as an alias for the caller's own approver slate, or refuse the unknown parameter with a located `400` naming the supported filters. Silently returning everything is the one answer that cannot be detected by the caller.

## Source

Extracted from the QA run objectstack-ai/objectstack#7517 (framework a86db175, console 09987b68).

os-help:
Claim: PM loop, services lane, wave 2 round 2
[...]

PR #8044, opened 2026-08-12
feat(rest): closed query-parameter sets become ingress policy, first tier of data read routes (#7606)

### LINKED ISSUE #7606: [rest] Unknown query parameters are silently dropped on every REST route except `/approvals/requests` — decide whether the closed-parameter-set rule becomes ingress policy
## What this is

A scope spill-out from #7527, filed for triage rather than fixed there (Prime Directive #10).

#7527 fixed exactly one route: `GET /api/v1/approvals/requests` now declares a closed query-parameter set and refuses anything outside it with a located `400`. The card that produced it explicitly suggested going wider — "worth fixing as one policy (reject unknown query parameters with a located `400`) rather than one endpoint at a time" — and that wider change was deliberately left out of #7527's PR because it is a cross-lane REST-ingress policy decision, not an approvals bug.

## The condition, stated generally

`rest-server.ts` handlers read the query keys they know and ignore the remainder. Every route other than the one #7527 closed still does this, so any misspelled, renamed or invented parameter is silently dropped and the caller gets a plausible-looking `200`. The failure is undetectable from the response in both directions, which is what makes it worth a policy rather than a bug-per-endpoint:

- **silently widens** — a dropped *filter* returns more rows than asked for, and an unfiltered list is shaped exactly like a genuinely broad match (#7527's measured case: `?assignedToMe=true` returned every request).
- **silently narrows** — a dropped *key* inside `where` / `$filter` answers 200/0 (#7463 defect 2, tracked as #7534).

The mechanism to do it with already exists and is proven on one route: `refuseUnknownQueryParams` in `packages/rest/src/query-allowlist.ts`, alongside the multiplicity rule `refuseRepeatedQueryParams` (#6877) that established the same shape and the same ADR-0112 envelope.

## Why it is a decision and not a chore

Applying the rule to the remaining routes is mechanical per route, but the policy has real costs to weigh, and getting them wrong is worse than the current silence:

1. **Every route's closed set must be MEASURED, not guessed.** The set is not "the filters" — it includes paging, ordering, alias spellings, and anything middleware reads. A whitelist that forgets `limit` converts a silent-widening bug into a loud paging outage. `rest-server.ts` has roughly 50 query read points; each needs reading, not a sweep.
2. **It is a breaking change for tolerated traffic.** Any client today sending a parameter we ignore starts getting a `400`. That traffic is invisible to us precisely because we drop it silently, so the blast radius cannot be measured from our side — only decided.
3. **Third-party and forward-compat callers.** A caller written against a newer client sending a parameter an older server does not know currently degrades quietly; under the policy it fails hard. That may well be the right answer (it is the #5714 / #5931 / #7463 family norm), but it is a deliberate stance on API evolution, not an implementation detail.
4. **Scope boundary with #7534.** That issue covers unknown fields inside `where` / `$filter` — the body/filter face. This one is the query-string face. They share a principle and should probably agree on envelope and message shape, but they are different code paths.

## Suggested shape, if it is taken

Route-by-route adoption of the existing helper, sized per lane rather than as one PR — each route's closed set measured from its own handler reads, with preservation pins alongside the refusal pins (the #7527 test file is the template: refusal cases assert `status` plus the nested `error.code` plus that the service was never called, and preservation cases assert the argument the service was handed, not merely a `200`).

No behaviour is claimed broken beyond the routes themselves; nothing here is urgent. Filing so the policy question is on the board instead of living only inside a closed approvals card.

## Refs

- #7527 — the one route already closed, and the helper it introduced
- #7534 — the same anti-pattern on the `where` / `$filter` face
- #6877 — the multiplicity rule this reuses the shape and envelope of

---

hotlong:
Fixes #7606

Implements the maintainer ruling of 2026-08-12: **policy YES, incremental, ⛔ no big-bang wave.** Scope is the ruling's first batch — the policy in writing, plus the first tier of data read routes.

## The policy, in writing

A REST route **declares its closed query-parameter set on the day it lands**, opening its handler with `refuseUnknownQueryParams(req, res, <EXPORTED_PARAMS>)` so an unrecognised name gets a located `400` instead of being dropped.

It is written in two places, both review-enforceable:

- **`packages/rest/src/query-allowlist.ts`** — the module header, where an implementer lands. Carries the ruling verbatim, the three measuring constraints, the exclusion rule, and the composition order.
- **AGENTS.md → "Route & surface ownership", rule 5** — where agents read the repo's binding route rules. The section's intro moves from "Four rules" to "Five".

## The first tier — each closed set MEASURED, and where from

| route | closed set | measured from |
| :--- | :--- | :--- |
| `GET /data/:object/:id` | `select`, `expand` | the one line that reads the query — `const { select, expand } = req.query \|\| {}`. The handler destructures exactly these two and forwards nothing else. |
| `GET /data/:object/export` | `format`, `header`, `limit`, `page`, `filter`, `search`, `searchFields`, `orderby`, `fields`, **`locale`** | every read of `q = req.query ?? {}` in the handler body — **plus `locale`, read one frame down** by `extractLocale` behind the `translateMetaItem` call that localises the header row. |
| `GET /search` | `q`, `query`, `objects`, `limit`, `perObject` | the handler's five reads: the term under both spellings it honours, the object scope, and the two result caps. |

Sets are exported constants (`DATA_RECORD_READ_PARAMS`, `DATA_EXPORT_PARAMS`, `GLOBAL_SEARCH_PARAMS`) so the pin tests assert against what the route actually declares, never a hand-copied second list.

### ⚠️ `locale` is the measurement that nearly went wrong

It appears nowhere in the export handler's body. A set measured from the handler alone would have omitted it and **`400`'d every `?locale=zh-CN` export that works today** — silent widening traded for a loud outage, committed by the change meant to prevent it. It was found by reading the helpers the handler calls. That lesson is now in the AGENTS.md rule and pinned by name in the tests.

Middleware was swept for query reads before measuring: `resolveProtocol`, `resolveExecCtx`, `enforceAuth`, `enforceApiAccess`, `enforceExportPermission`, `resolveSecurityService` and `resolveRequestEnvironmentId` read none (environment resolution is by host/header). `translateMetaItem` was the only hit.

## ⛔ What is deliberately NOT closed

**`GET /data/:object` (the record list) does not get this gate**, and the reason is recorded in code so the next agent chooses rather than discovers. Its handler passes the whole query record to `findData`, whose normalizer lowers every leftover key into an **implicit field-equality predicate** — `?status=open` *is* the filter. The valid names are the object's own fields, which vary per object and include the audit/tenant/owner columns the registry injects, so a closed list here could only ever be wrong. It is already gated one layer down and against the right authority: #4134 refuses an unknown **field** with `400 INVALID_FIELD`, and #7534 extended that to the explicit `where`/`$filter` axes.

The general test, now written down: **if an unrecognised name has a defined meaning on the route, the set is open** — gate it where the authority for the name lives.

Three tests pin the exclusion, including one that goes red if someone "completes the sweep".

## How the two guards compose

Recognition runs **before** the arity gate, per the rule `query-allowlist.ts` already stated (#7527): *"I do not know this parameter" outranks "this parameter I do know was supplied twice"*, so a request committing both errors is told the more fundamental one. Pinned on both tiered routes rather than left to the order the calls happen to sit in.

**Both answer the same envelope** — nested ADR-0112 `{ error: { code: 'VALIDATION_ERROR', message } }` — so composing them adds **no second dialect** to any route. Also pinned.

### ⛔ #8001 is not resolved here, and no code fork is forced

The serial constraint from PR #8004 (#7390) does not bite, because **the two gates never meet on one request**: `assertFilterParamSuppliedOnce` answers `400 INVALID_FILTER` through the flat `mapDataError` envelope and lives **only on the list route excluded above**, which never gets recognition.

On the export route, `filter` is *inside* the closed set, so a repeated `?filter=` passes recognition and still reaches the multiplicity gate, still answering exactly what it answered before. The #8001 divergence is neither widened nor resolved — a test asserts this explicitly and goes red if a later change makes that call unilaterally.

## Tests

`packages/rest/src/rest-server-closed-query-params.test.ts` — 27 tests, both halves per route on the #7527 template:

- **refusal pins**: status + nested `error.code` + the located message + **the service was never called**.
- **preservation pins**: the **arguments the service actually received** — not merely a `200`, because "still 200" is exactly what the defect looked like. `limit` and `page` on the export route are separated by which one binds the chunk (`$top` = 25 vs 50), so each is proved to have arrived.

**Reverse-verified from the committed state**: with the three gate calls removed, 11 refusal/composition tests go red and the preservation pins stay green. The by-id case fails as `200` with the full record — the silent-widening defect, measured:

```
expected a 400 refusal for fields, got 200 with body {"object":"task","record":{"id":"1","title":"alpha"}}
```

## Gates

- `pnpm -w typecheck` — 127/127 tasks pass
- `@objectstack/rest` suite — 98 files, 1593 tests pass (re-run after merging `main`)
- `check:type-check-debt` — **`@objectstack/rest` holds at 155**, not raised; it is absent from the surplus list, i.e. still exactly at its zero-margin ceiling
- `check:adr-0087-registration` — the changeset carries its disposition
- ESLint clean on the touched files (`--no-inline-config`, the repo's own flag)
- `packages/spec` moved on `main` while this was open → rebuilt, `check:generated` reports all 13 artifacts current

## Breaking tolerated traffic is deliberate

Stated plainly in the changeset rather than described as a bug fix: a caller sending these routes a parameter we ignore today starts getting a `400`, the blast radius cannot be measured from our side precisely because we have been dropping it silently, and **v17 is the intended window**.

The two callers most likely to notice are on `GET /data/:object/:id`: `?fields=` and `?populate=` are refused. They are the spec alias table's canonical/alias spellings for slots this route reads as `select`/`expand`, and it folds no aliases — so they were being dropped, silently returning the full record. They are left outside the closed set rather than implemented, because adding them would advertise a capability the handler does not have; the refusal names `select`/`expand` as what the route accepts.

## Out-of-scope finding filed

**#8039** — `GET /data/:object/:id` folds no query aliases, so the canonical `fields` spelling is dropped while the alias `select` works, diverging from `RPC_QUERY_ALIAS_SLOTS` and from the sibling list route. Filed unassigned per Prime Directive #10, linked from the code comment.

## Notes for review

- `#7981` (converging `registerSecurityEndpoints`' envelopes) is in flight on the same file but a different region; `main` was merged rather than resolved blind, and the merge was clean.
- `content/docs/releases/` is untouched.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MgfwFqbv6D8knYH32yaqva)_

### DIFF
--- a/.changeset/closed-query-param-ingress-policy.md
+++ b/.changeset/closed-query-param-ingress-policy.md
@@ -1,0 +1,82 @@
+---
+"@objectstack/rest": minor
+---
+
+feat(rest): closed query-parameter sets become REST ingress policy, starting with the first tier of data read routes (#7606)
+
+**BREAKING** for tolerated traffic, and deliberately so — see the last section.
+
+## The condition
+
+`rest-server.ts` handlers read the query keys they know and ignore the
+remainder, so a misspelled, renamed or invented parameter is **silently
+dropped** and the caller gets a plausible-looking `200`. The failure is
+undetectable from the response in both directions:
+
+- it **silently widens** — a dropped `?objects=` fans a search across every
+  object; a dropped `?fields=` returns the whole record. An unfiltered result
+  is shaped exactly like a genuinely broad match.
+- it **silently narrows** — a dropped key inside a filter answers `200` with
+  zero rows, which is shaped exactly like an object that really is empty.
+
+There is no status, header or field that distinguishes either from a real
+answer, which is what makes it worth a policy rather than a bug per endpoint.
+An AI caller can detect neither direction at all.
+
+## The policy
+
+A REST route **declares its closed query-parameter set on the day it lands**,
+refusing an unrecognised name with a located `400` instead of dropping it.
+Adoption is incremental and per lane — data READ routes first — never a
+one-shot sweep. The rule, its three measuring constraints and the exclusions
+are written up in `packages/rest/src/query-allowlist.ts` and in AGENTS.md's
+"Route & surface ownership" section, so it is enforceable at review time.
+
+## The first tier
+
+Three routes, each set **measured from the handler's own read points**:
+
+| route | closed set |
+| :--- | :--- |
+| `GET /data/:object/:id` | `select`, `expand` |
+| `GET /data/:object/export` | `format`, `header`, `limit`, `page`, `filter`, `search`, `searchFields`, `orderby`, `fields`, `locale` |
+| `GET /search` | `q`, `query`, `objects`, `limit`, `perObject` |
+
+The refusal is `400` with the ADR-0112 nested body
+`{ error: { code: 'VALIDATION_ERROR', message } }` — the same envelope these
+routes' existing multiplicity refusals answer, so no route gains a second
+dialect. The message names the parameters that were not understood **and lists
+the ones that are**, so a caller can fix the request from the response alone.
+
+## What is deliberately NOT closed
+
+`GET /data/:object` (the record list) keeps accepting any name. Its handler
+passes the whole query to the normalizer, which lowers every leftover key into
+an implicit field-equality predicate — `?status=open` *is* the filter — so the
+valid names are the object's own fields and vary per object. That route is
+already guarded one layer down and against the right authority: an unknown
+**field** is refused there with `400 INVALID_FIELD`. Closing it here would
+break every implicit filter.
+
+Its repeated-`?filter=` refusal (`400 INVALID_FILTER`) is untouched, and since
+the recognition gate never runs on that route the two guards never meet on one
+request.
+
+## Breaking tolerated traffic is the point, and v17 is the window
+
+A caller sending one of these routes a parameter we ignore today starts getting
+a `400`. That is not a side effect — it is the change. This is **not a pure bug
+fix**: the blast radius cannot be measured from our side, precisely because we
+have been dropping the traffic silently, so it was decided rather than
+measured (maintainer ruling, 2026-08-12). v17 is the intended window; the
+longer it waits the more tolerated traffic there is to break.
+
+Two callers most likely to notice, both on `GET /data/:object/:id`: `?fields=`
+and `?populate=` are refused. They are the spec's canonical/alias spellings for
+slots this route reads as `select` and `expand`, and it folds no aliases — so
+they were being dropped, silently returning the full record. They are left
+outside the closed set rather than implemented, because adding them would
+advertise a capability the handler does not have; the refusal message names
+`select` and `expand` as what the route does accept.
+
+<!-- adr-0087: not-required (no-migration-prescription) HTTP query-string ingress, not stored metadata — the ADR-0087 ledger drives `objectstack migrate meta`, and no metadata migration can rewrite a caller's URL. No authorable spec key, export or config field is removed or renamed by this change. -->
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -654,17 +654,20 @@ Three things to get right, all of them measured rather than assumed:
   handler calls, not just the handler**: the export route's `?locale=` never appears in
   its body — it is read a frame down, by `extractLocale` behind the call that localises
   the header row — so a set measured from the handler alone would have 400'd every
-  localised export that works today. Pin both halves per route,
-  the #7527 file being the template: a refusal pin (status + nested `error.code` + **the
-  service was never called**) beside a preservation pin (**the arguments the service
-  actually received** — "still 200" is exactly what the defect looked like).
+  localised export that works today. Pin **both halves** per route: a refusal pin
+  (status + nested `error.code` + **the service was never called**) beside a
+  preservation pin (**the arguments the service actually received**). Neither half is
+  optional, and a bare status assertion is not a pin — "still 200" is exactly what the
+  defect looked like, and the refusal's whole point is that the service never ran.
 - ⛔ **Routes whose parameter set is genuinely OPEN are excluded, by name.**
   `GET /data/:object` hands its whole query to the normalizer, which lowers every
   leftover key into an implicit field filter (`?status=open` *is* the filter) — the
   valid names are the object's own fields, so any list here would be wrong. It is
-  already gated one layer down, against the right authority, by #4134/#7534's unknown-
-  **field** refusal. The test: *if an unrecognised name has a defined meaning on this
-  route, the set is open* — gate it where the authority for the name lives.
+  already gated one layer down, against the right authority: an unknown **field** is
+  refused there with `400 INVALID_FIELD`, judged against the object's real field map
+  (the registry's, including the audit/tenant/owner columns it injects — not the
+  author's declaration). The test: *if an unrecognised name has a defined meaning on
+  this route, the set is open* — gate it where the authority for the name lives.
 - **Existing routes convert per lane, ⛔ never as one sweep** (data read routes first).
   A broad wave with thin pins is the failure mode the ruling rejected.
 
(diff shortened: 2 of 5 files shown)


hotlong:
**PM review — `domain:cli` seat (#6024).** Verdict: **accept**. Auto-merge on. No change requests.

## `locale` is the whole reason the ruling said "measure, never guess"

It appears **nowhere in the export handler's body** — it is read one frame down by `extractLocale`, behind the `translateMetaItem` call that localises the header row. A set measured from the handler alone omits it and **`400`s every `?locale=zh-CN` export that works today**.

That is the failure mode the ruling named, landing in a place nobody named: *silent widening traded for a loud outage, committed by the change meant to prevent it.* I warned about `limit`; the real one was a parameter the handler never mentions. Sweeping the middleware before measuring (`resolveProtocol`, `resolveExecCtx`, `enforceAuth`, `enforceApiAccess`, `enforceExportPermission`, `resolveSecurityService`, `resolveRequestEnvironmentId` — none read the query; `translateMetaItem` the only hit) is what turned that from luck into method. Putting the lesson into the AGENTS.md rule and pinning `locale` **by name** is what stops the next tier repeating it.

## The exclusion is better reasoned than the dispatch was

`GET /data/:object` correctly does **not** get this gate: its handler forwards the whole query record to `findData`, whose normalizer lowers every leftover key into an **implicit field-equality predicate** — `?status=open` *is* the filter. The valid names are the object's own fields, per-object, including the audit/tenant/owner columns the registry injects. A closed list there could only ever be wrong, and the names are already gated one layer down **against the right authority** (#4134's `INVALID_FIELD`, extended to the explicit axes by #7534).

And it generalises the reason instead of leaving it as an exception: **if an unrecognised name has a defined meaning on the route, the set is open — gate it where the authority for the name lives.** That is a sharper rule than my dispatch gave you, and it is the one that will decide the next tier correctly.

**Three tests pin the exclusion, including one that goes red if someone "completes the sweep."** Protecting a deliberate gap from a future well-meaning tidy is the part most PRs skip — an unexplained hole invites exactly the change that breaks it.

## #8001 held, and held for a reason rather than by avoidance

The instruction was to compose the two guards deliberately and ⛔ not resolve the #8001 fork unilaterally. The answer is stronger than compliance: **the two gates never meet on one request.** `assertFilterParamSuppliedOnce` answers through the flat `mapDataError` envelope and lives **only on the excluded list route**; on the export route `filter` is *inside* the closed set, so a repeated `?filter=` passes recognition and reaches the multiplicity gate answering exactly what it answered before.

Neither widened nor resolved — **and a test asserts that, going red if a later change makes the call unilaterally.** That converts "I didn't touch it" into a guarantee.

Composition order is pinned rather than left to the order the calls happen to sit in, on the rule `query-allowlist.ts` already stated: *"I do not know this parameter" outranks "this parameter I do know was supplied twice."* Both answer the same nested ADR-0112 envelope, so composing them adds **no second dialect**.

## The rest

Policy written where implementers land *and* where agents read binding route rules (AGENTS.md rule 5, "Four rules" → "Five"). Sets are exported constants, so the pins assert what the route **declares** rather than a hand-copied list. Preservation pins assert **the arguments the service received** — with `limit` and `page` separated by which one binds the chunk (`$top` 25 vs 50), so each is proved to have arrived, because "still 200" is precisely what the defect looked like.

`fields` / `populate` refused on by-id is the right call: they are the alias table's spellings for slots this route reads as `select`/`expand`, and it folds no aliases — so they were being **dropped, silently returning the full record**. Implementing them would advertise a capability the handler does not have; refusing while naming what *is* accepted is honest, and #8039 records the divergence.

`@objectstack/rest` holds at **155** and is absent from the surplus list — still exactly at its zero-margin ceiling, not raised.

---

PR #8045, opened 2026-08-12
fix(metadata-protocol): an object's overlay row is a base layer, not its resolved schema (#8027)

### LINKED ISSUE #8027: A tenant object overlay drops every `objectExtensions` field from the resolved schema, and `effective` contradicts `code` in the same response
Filed by the `domain:metadata` PM seat on behalf of #7556's dev, which measured this while fixing that card but could not file it — the `objectstack-ai` org has no Claude GitHub App connection, so cloud dev containers get 403 on the GitHub API. Text is the dev's.

**Measured, not inferred; distinct from #7556 and not caused by its fix (PR #8015).**

## The defect

When a `sys_metadata` overlay row exists for an object — e.g. an admin customises the object's label in Studio — the protocol's by-name read **replaces the resolved body with the raw stored row**, and the list read does the same via `mergePackageAwareOverlay`, which **replaces wholesale rather than merging fields**.

⚠️ **Both routes therefore agree — which is precisely why #7556's agreement pin does not catch it** — but they agree on a body that has lost every `objectExtensions` field.

## Measured

One `extend` contributor (`loyalty_tier`, `linkedin_url`, `csat_score`), one env-wide overlay row, no metadata service:

```
byName fields:     name, industry + system columns   — no extension fields
listed fields:     identical                          — no extension fields
layers.code:       has all 3                          (correct, ADR-0029 D9.6)
layers.overlay:    the raw row                        (correct)
layers.effective:  no extension fields
```

⇒ **A single `?layers=true` response reports a `code` layer that has the fields and an `effective` layer that does not, with no customisation explaining the difference.**

## Why this looks wrong

ADR-0029 **D9.2** states the base layer is `overlay ?? own` and that extenders are **then folded on** — which is exactly what `SchemaRegistry.resolveObject` does for overlays it knows about. **The protocol's `sys_metadata` path never folds.**

The practical consequence mirrors #7556: an admin who customises an object's label **silently removes three extension-contributed fields from every writable form**, while the data API keeps accepting and persisting them.

## Suggested shape — ⚠️ and a trap in it

Deliberately **not** implemented in #7556: it widens blast radius beyond that card and needs its own reverse-verification.

Treat the overlay row as the **base layer** and fold extenders onto it, matching D9.2, rather than serving it as the resolved schema.

⚠️ **`mergeObjectDefinitions` concatenates validations and indexes, so the fold is NOT idempotent.** Any fix must establish that the stored row has not already been through it — a row saved from a previously-merged read may already contain extension fields, and folding again would duplicate validations and indexes. This is the part most likely to be missed.

## Reproduction

Register an object `own` plus an `extend` contributor, seed one `sys_metadata` row (`type:'object'`, `state:'active'`, `organization_id:null`) whose body is the un-extended object, then compare `getMetaItem` / `getMetaItems` / `getMetaItemLayered`.

## Note for whoever picks this up

⚠️ This card is a good candidate for the same **agreement-pin** discipline #7556 used — but note that an agreement pin alone is **insufficient here**, because the two routes already agree. The assertion needs to be against **the registry's resolved schema** (what D9.2 defines), not against the other route.

huangyiirene:
Fixes #8027

⚠️ **PR opened by the `domain:metadata` PM seat on the dev's behalf** — it pushed `claude/issue-8027-overlay-extension-fold` but no PR appeared, so the GitHub API is still unreachable from that container despite the App being connected. Body is the dev's own changeset; its report follows.

## ⚠️⚠️ This also repairs a LIVE REGRESSION shipped by #7556 (PR #8015) one hour ago

**Read this first** — it is the more urgent half of the PR.

On an in-process (`bridged`) boot, the by-name read and the `code` layer were serving **every extender-contributed `validation` and `index` TWICE**. Introduced by #8015, live on `main` now.

`mergeObjectDefinitions` **concatenates** `validations` and `indexes` (`fields` is a key-keyed spread and scalars are last-writer-wins, so those were always safe). #8015 documented a precondition — *"callers must apply this only to a base that has not been through the fold"* — and:

> that precondition turned out to be one **no caller can honour**, and **two shipped call sites already violated it**:
>
> 1. **The MetadataService body on an in-process boot.** `bridgeObjectsToMetadataService` seeds that service from `registry.getAllObjects()` — bodies that are *already resolved* — so #7556's fold ran on a folded base.
> 2. **A stored overlay row.** The write path persists the request body verbatim (ADR-0005 §Validation), so the ordinary Studio GET → edit → PUT round-trip stores whatever the read served — and since #7556 that read is folded. Seeded / imported / migrated rows are unconstrained besides.

⭐ **And why #8015's own pin could not see it:** *"it compares FIELD NAMES, and the field spread is idempotent."* The duplication lives in `validations` and `indexes`, a dimension that pin never touched.

⚠️ **That is a gap in my review, not just in the pin.** I accepted #8015 and called its four-host agreement pin excellent — it was, for the dimension it tested. I did not ask whether the pin covered the *other* fields of a merged object. A duplicated index does not fail a test; it fails a deployment.

⇒ The fix makes `SchemaRegistry.foldObjectExtendersOnto` **idempotent** rather than documenting the precondition harder: an entry the `extend` contributors are about to add, already present in the base, is removed first and re-added exactly once. Extenders are still concatenated **against each other** — two contributors declaring an identical rule still yield two, matching `resolveObject` — so nothing the fold did on an unfolded base is narrowed, and such a base is returned **by reference**, byte-identical.

## The defect this card was filed for

An overlay row for an object — an admin renaming its label in Studio — was adopted **as the resolved schema**. `getMetaItem` returned the stored row; `getMetaItems` did the same through `mergePackageAwareOverlay`, which picks a per-slot winner **wholesale rather than merging fields**; and `getMetaItemLayered`'s `effective` is `overlay ?? code`, so it inherited the same body.

ADR-0029 **D9.2** defines the resolution as `overlay ?? own` **with the `extend` contributors folded on** — exactly what `SchemaRegistry.resolveObject` does for an overlay it knows about, and what #7556 made the by-name read do for the MetadataService copy. **The `sys_metadata` path was the one adopter that never folded.**

Measured with one `extend` contributor (three fields) and one env-wide overlay row: `byName` and `listed` both served the object with **no** extension fields, while `layers.code` served them and `layers.effective` did not — so **one `?layers=true` response reported a `code` layer that has the fields and an `effective` layer that does not**, with an `overlay` layer showing a customisation that explained none of the difference.

Practical cost, the #7556 shape again: an admin who customises a label **silently removes three extension-contributed fields from every writable form**, while the data API keeps accepting and persisting them.

## Pinned against the registry's resolved schema — ⛔ deliberately NOT as route agreement

`packages/rest/src/meta-object-overlay-extension-fold.test.ts`, eight cases over real handlers / real protocol / real registry:

> #7556's `byName === listed` pin is **green throughout this defect**, because here both routes agree — on a body that has already lost the fields.

Cases: the overlay case on both routes and on `layers.effective`; `code` and `effective` agreeing when the row customises nothing; `layers.overlay` still reporting only what the tenant stored (an extension is not a tenant customisation — the boundary #7556 drew); **an already-folded row and a bridged host holding the idempotency**; the no-overlay and no-extension controls; and an **anti-vacuity** case pinning that the fixtures are discriminated.

## Byte-identity, measured rather than argued

All three surfaces dumped for **nine hosts**, under this branch and under the pre-fix behaviour: **8 of 9 identical.** The one that differs is the extended object on a `bridged` host, where the pre-fix payload carries `['owner_rule','ext_rule','ext_rule']` / `['owner_idx','ext_idx','ext_idx']` and this branch carries each once — **the #7556 regression, repaired.**

## Levels

`metadata-protocol` **patch** — restores the contract these routes were already specified to answer, same reasoning #7556 used for the same routes. `objectql` **patch** — no new public API (`foldObjectExtendersOnto` exists since #7556); its documented contract moves from *"not idempotent, callers must guarantee an unfolded base"* to *"idempotent"*, a defect fix rather than a capability, and no caller can be relying on duplicated validations.

---

### DIFF
--- a/.changeset/object-overlay-row-is-a-base-layer.md
+++ b/.changeset/object-overlay-row-is-a-base-layer.md
@@ -1,0 +1,97 @@
+---
+'@objectstack/metadata-protocol': patch
+'@objectstack/objectql': patch
+---
+
+fix(metadata-protocol): an object's overlay row is a base layer, not its resolved schema (#8027)
+
+**Behaviour change, and it is a payload gaining fields.** When a `sys_metadata`
+overlay row exists for an object, `GET /meta/object/:name`, `GET /meta/object`
+and the `effective` layer of `?layers=true` now serve that object's RESOLVED
+schema — the overlay row as the base layer with its `objectExtensions`
+contributors folded on (ADR-0029 D9.2) — where they previously served the stored
+row verbatim. Any consumer of those routes sees the extension's fields appear on
+customised objects. An object with no overlay row, and an object nothing
+extends, are byte-identical (measured; see below).
+
+**A second payload change, in the other direction:** on an in-process
+(`bridged`) boot the by-name read and the `code` layer previously served every
+extender-contributed `validation` and `index` TWICE. That duplication is
+removed. It was a live regression introduced by #7556 (PR #8015) and is
+explained under "the fold is not idempotent" below.
+
+The defect: an overlay row for an object — an admin renaming the object's label
+in Studio — was adopted as the resolved schema. `getMetaItem` took the stored row
+as `item` and returned it; `getMetaItems` did the same through
+`mergePackageAwareOverlay`, which picks a per-slot winner WHOLESALE rather than
+merging fields; and `getMetaItemLayered`'s `effective` is `overlay ?? code`, so
+it inherited the same body. D9.2 defines the resolution as `overlay ?? own` with
+the `extend` contributors folded ON, which is exactly what
+`SchemaRegistry.resolveObject` does for an overlay it knows about, and what
+#7556 made the by-name read do for the MetadataService copy. The `sys_metadata`
+path was the one adopter that never folded.
+
+Measured with one `extend` contributor (three fields) and one env-wide overlay
+row: `byName` and `listed` both served the object with NO extension fields,
+while `layers.code` served them (#7556 folds it) and `layers.effective` did not
+— so a single `?layers=true` response reported a `code` layer that has the
+fields and an `effective` layer that does not, with an `overlay` layer showing a
+customisation that explained none of the difference. The practical cost is the
+#7556 shape again: an admin who customises a label silently removes three
+extension-contributed fields from every writable form, while the data API keeps
+accepting and persisting them.
+
+**The fold is not idempotent, and that is the hazard this fix had to clear
+rather than assume away.** `mergeObjectDefinitions` CONCATENATES `validations`
+and `indexes` (`fields` is a key-keyed spread and the scalar props are
+last-writer-wins, so those were always safe), so folding a body that has already
+been through the fold duplicates both — and a duplicated index does not fail a
+test, it fails a deployment. The precondition #7556 documented ("callers must
+apply this only to a base that has not been through the fold") turned out to be
+one no caller can honour, and two shipped call sites already violated it:
+
+1. **The MetadataService body on an in-process boot.** ObjectQL's
+   `bridgeObjectsToMetadataService` seeds that service from
+   `registry.getAllObjects()` — bodies that are already resolved — so #7556's
+   fold ran on a folded base and served every extender validation and index
+   twice. Its own pin could not see this: it compares FIELD NAMES, and the field
+   spread is idempotent.
+2. **A stored overlay row.** The write path persists the request body verbatim
+   (ADR-0005 §Validation), so the ordinary Studio GET → edit → PUT round-trip
+   stores whatever the read served — and since #7556 that read is folded. The
+   row is *defined* by D9.2 as the base layer, but nothing enforces it, and
+   seeded / imported / migrated / pre-existing rows are unconstrained besides.
+
+So `SchemaRegistry.foldObjectExtendersOnto` was made IDEMPOTENT instead of
+documented harder: an entry the `extend` contributors are about to add, already
+present in the base, is removed first and then re-added by the fold exactly
+once. Extenders are still concatenated against each other — two contributors
+declaring an identical rule still yield two, matching `resolveObject` — so
+nothing the fold did on an unfolded base is narrowed, and such a base is
+returned by reference, byte-identical.
+
+Levels: `metadata-protocol` is `patch` — it restores the contract these routes
+were already specified to answer, and it is the same reasoning #7556 used for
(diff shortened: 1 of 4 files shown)


huangyiirene:
## ACCEPT — flipped ready, auto-merge SQUASH armed @ `bf2028a`

All 25 jobs concluded before arming: **23 success, 2 skipped (`Build Docs`, `Console Pin Gate`), zero failures.** `TypeScript Type Check` green at 13:44:30Z, `Check Changeset` green. Read per-job, not off an aggregate.

⚠️ **Prioritised because it repairs a regression that is live on `main` right now** — every in-process boot has been serving duplicated extender `validation` and `index` entries since #8015 landed at ~12:00Z.

## Review criteria, verified

- ⭐ **Idempotency established on both violating call sites**, not asserted. The dev did not accept #8015's documented precondition; it showed the precondition is one **no caller can honour** and found two shipped violations. That is the difference between "I checked and it can't happen" and a measurement.
- ⭐ **Pinned against the registry's resolved schema, not route agreement** — and it said *why*: #7556's `byName === listed` pin is **green throughout this defect**, because both routes agree on a body that has already lost the fields. Naming your own predecessor pin's blind spot is the behaviour this lane wants.
- **`layers.overlay` still tenant-only** — the boundary #7556 drew is preserved, deliberately.
- **Anti-vacuity case present**; eight cases over real handlers / real protocol / real registry.
- **Byte-identity measured across nine hosts, 8 of 9 identical**, with the ninth being the repaired regression and its exact arrays quoted (`['owner_rule','ext_rule','ext_rule']` → each once). That is a measurement, not a claim.
- `check:type-check-debt` clean; no ledger entry raised; `--lower` not run.

## ⚠️ The review gap was mine, and the criterion is now recorded

I accepted #8015 and called its four-host agreement pin excellent. It **was** excellent for the dimension it tested — field names — and I never asked whether it covered the *other* fields of a merged object. `mergeObjectDefinitions` concatenates `validations` and `indexes`; a name-based pin is structurally blind to both.

⇒ **Standing review criterion from now on: when a merge CONCATENATES, a name-based pin proves nothing about the concatenated arrays.** A duplicated index does not fail a test — it fails a deployment.

This is the second time today a landed PR's pin turned out narrower than its apparent coverage, and both times the *next* card found it. That is the system working, but it works one card late.

---

Issue #6915, opened 2026-08-09
driver-memory 完全没有行级租户隔离(#3724 的未修姊妹面):多租户下静默不隔离


os-zhuang:
## 一句话说明

`@objectstack/driver-memory` **没有任何行级租户隔离** —— 读路径不加租户谓词,写路径不打租户戳,`distinct()` 甚至不接受 `options`。这与 **#3724**(driver-mongodb,同为"整个能力根本不存在")是同一类,而 #3724 已经 **closed as completed**。

在 #6792(PR #6908,SQL 家族读侧收口点闸门)的范围审计中测得。发现者当时判断"姊妹卡已关闭,新卡只会被以冻结为由关掉",因而没有开卡;PM 复核时发现该前提反了 —— 见下。

## 事实(#6908 分支基线 `6595262` 上实测)

```
$ grep -c "tenantId\|organization_id" packages/drivers/driver-memory/src/memory-driver.ts
0
```

零匹配。另外 `distinct(object, field, query?)` 的签名里**没有 `options` 参数**,所以调用方即便想传 `tenantId` 也无处可传。

对照 #6908 刚刚在 SQL 家族上做的事:`applyTenantScope` 是读侧唯一收口点,`getBuilder` 构造的每个读 builder 都必须路由过去,并由 `scripts/check-tenant-chokepoint.mjs` 从 AST 逐次重新推导。`driver-memory` 不在该闸门的扫描集内,**而且这不是因为 #5499 冻结** —— 是因为它根本不使用这套机制(文件里 `getBuilder` / `applyTenantScope` 均为零匹配)。闸门因此对它不产生任何判定,也就没有 DEBT 行可记:是一个真实的缺席,而不是被静默豁免。

## 为什么"姊妹卡已关闭"不是不开卡的理由

#3724 的关闭状态是 `state_reason: completed`,它自己给的处置是二选一:

- **A. 实现行级租户隔离** —— 对齐 SQL driver 的租户列解析 / 读谓词 / 写打戳;
- **B. 显式声明不支持多租户** —— 检测到多租户模式就**启动时硬失败**,而不是静默地不隔离(原卡倾向 B)。

`packages/drivers/driver-mongodb` 现有的 `mongodb-tenancy-guard.ts` 正是处置 B。也就是说 **#3724 是被"修好"关掉的,不是被"不修"关掉的**,先例指向与"不必开卡"相反的方向。

## 与 #5499 冻结的关系

[...]

Issue #6664, opened 2026-08-08
driver-memory census in `undeclared-field-write-driver-split.integration.test.ts` is stale — a second `packages/runtime` test consumer (#6468) is outside #5704's "in this one place" ruling


os-zhuang:
Observation-class finding, surfaced while migrating `auth-contains-filter.test.ts` off driver-memory (#5893). Nothing a user hits — this is a **bookkeeping invariant of the driver-memory retirement programme** that has silently expired. Filed unassigned for PM triage; not fixed in #5893's PR (out of its file surface, and the disposition needs a ruling rather than an edit).

## What is stale

`packages/runtime/src/sandbox/undeclared-field-write-driver-split.integration.test.ts` opens with a load-bearing census sentence:

> `@objectstack/driver-memory` is imported here ON PURPOSE, and this is the **only PERMANENT test consumer of it in the repository**.

and closes the same block with the mechanism that sentence exists to serve:

> #5704/#5784 renamed them all to `makeStubDriver`, precisely so that **grepping for the driver lands here, and only here**.

That is the verification handle the retirement programme uses: grep the declarations, land on exactly one ruled file, and anything else is a new arrival to dispose of. The count is no longer true.

## Evidence (on `origin/main`, e6025e9)

```
$ git grep -n "from '@objectstack/driver-memory'" -- '*.ts' | grep -v node_modules
packages/runtime/src/sandbox/undeclared-field-write-driver-split.integration.test.ts:107
packages/runtime/src/autonumber-seed-cross-side-parity.integration.test.ts:35
examples/embed-objectql/src/index.ts:17
```

The second line is the finding. `packages/runtime/src/autonumber-seed-cross-side-parity.integration.test.ts` arrived with #6468 (PR #6553, `fix(objectql,driver-sql): 播种解析按声明的 suffix 定位计数器`) — **after** #5704's survey and after #5830's follow-up — and it imports `InMemoryDriver` for a reason that reads exactly as permanent as the ruled file's:

> The engine half runs on the REAL `InMemoryDriver` (`supports = {}`, so the engine's fallback owns the counter — the shape memory/mongo deployments run) and the driver half on a REAL `SqlDriver` over better-sqlite3, each holding the same fixture rows.

[...]

Issue #7991, opened 2026-08-12
`packages/qa/downstream-contract` — the repo's backward-compatibility gate for `@objectstack/spec` reports 14/14 green against a spec change that breaks its own fixture, because it reads `spec/dist`


hotlong:
Filing unassigned; `domain:*` routing and grading are the triage seat's. `packages/qa/**` is not the devx lane — filing at the destination per the file-at-destination rule, from the devx seat (#6023) as the committed first follow-up of PR #7966.

## The defect, demonstrated rather than argued

`packages/qa/downstream-contract` is the repository's **frozen third-party consumer fixture**. Its own README states that a failure there means a spec change has narrowed something a published-spec consumer already relies on — i.e. that the change is **breaking**. It is the backward-compatibility gate for `@objectstack/spec`.

It ships no `vitest.config.*`, so `@objectstack/spec` resolves through `exports` to `packages/spec/dist` — a build artifact. Measured during PR #7966, direction predicted before running:

| step | tree | result |
|---|---|---|
| baseline | spec built, source == dist | 14/14 pass |
| a required field injected into `ConnectorSchema` **in source only**, no rebuild | the frozen fixture can no longer parse | **14/14 pass** |
| identical tree, spec aliased to source | — | **1 failed / 13 passed**, naming the injected field |

So the gate whose entire job is to notice breaking spec changes **did not notice one**, and said nothing about why. This is the silent half of #7668: a dist merely *behind* the source runs green against old behaviour, with nothing in the output indicating the verdict is about build state rather than the checkout.

## Why this one first, ahead of the other 62

PR #7966 (merged into the gate + registry for #7849) measured **63 packages with at least one unaliased artifact import, across 312 package↔dependency pairs**. This card is one of those 63 registry entries and is deliberately *not* a duplicate of the sweep — it is pulled out first because its blast radius is categorically different from the rest:

- Every other entry produces a wrong verdict about **its own package**.
- This entry produces a wrong verdict about **whether the repo is shipping a breaking change to `spec`**, which is the highest-churn package in the tree, and does so on the one test suite specifically built to answer that question.

A green here is currently consumed as evidence of backward compatibility. It is not evidence of that today.

## The fix, and the one thing that could make it non-trivial

[...]

PR #8048, opened 2026-08-12
fix(plugin-dev): tell "not installed" apart from "installed and failed to construct" (#7926)

### LINKED ISSUE #7926: DevPlugin 的 bare catch 把任何 driver 构造失败都报成「not installed」(#6915 实测)
## 一句话说明

`packages/plugins/plugin-dev/src/dev-plugin.ts:311-327` 用一个**无绑定的 `catch {}`** 包住整个 driver 装配段,于是 driver 构造期抛出的**任何**错误都被吞掉,并统一报成同一句话:

```
✘ @objectstack/runtime or @objectstack/driver-memory not installed — skipping driver
```

这句话在包确实缺失时是对的;在包**装好了、但构造函数按设计拒绝启动**时,它把操作者指向一个完全不存在的问题。

## 实测(#6915 分支 `eb1ad94`,真跑 `DevPlugin.init()`)

`driver-memory` 的 #6915 租户闸在构造函数里拒绝多租户姿态。在 `OS_TENANCY_POSTURE=isolated` 下真实运行:

```
CONSTRUCTOR throws, code = MEMORY_MULTI_TENANT_UNSUPPORTED
WARN   ✘ @objectstack/runtime or @objectstack/driver-memory not installed — skipping driver
```

那条错误消息本身写得很完整 —— 它点名了检测到的姿态、产生该姿态的两个 env 旋钮、以及 `@objectstack/driver-sql`(含 `:memory:` 的就地替代)作为出路。**这些一个字都没到操作者眼前。**

## 为什么这不是「只是日志措辞」

本次 `isolated` 场景里,DevPlugin 的 init 最终仍然拒绝了 —— 但那是**后面**的 ADR-0093 D5 organizations fail-fast 干的,和 driver 无关。也就是说:

- 企业版 `@objectstack/organizations` **缺席**时:D5 兜住了,栈没起来,只是死因被归错。
- 企业版**装好且健康**时:D5 放行,于是栈**照常起来**,只是没有 data driver,而唯一的线索是一句说包没装的假话。第一个查询才会炸,炸点离真因十万八千里。

这正是 Prime Directive #10 的形状:一个真实的、写得很好的拒绝理由,被一个 catch-all 降级成了一句不实的诊断。

## 建议(未定,留给分诊)

`catch` 绑定住错误,区分「模块解析失败」和「其它」:前者保留现在这句;后者把原始错误(至少 `code` + `message`)透出来。是否应当**重抛**——即一个明确拒绝启动的 driver 是否应该让 DevPlugin 也拒绝启动——是个产品形状问题,本卡不替谁作决定。

`mongodb-tenancy-guard.ts`(#3724)有同样的暴露面,只是 DevPlugin 从不构造 MongoDB driver,所以从未被触发过。

## 关联

- #6915 / PR #7924 —— 发现处;`driver-memory` 启动期租户闸(处置 B)
- #3724 —— `driver-mongodb` 的同类闸门,同样的 typed error + code 形状
- ADR-0093 D5 —— 本次场景里实际兜住栈的那道闸

未认领,仅作记录。按 #6915 派发单第 4 条(「卡的定级是 S/M 且保持不变,相邻的东西另开卡」)单独开卡,不并入。

hotlong:
Fixes #7926

## The defect

Every optional-service load in `DevPlugin.init()` ended in a bare `catch {}` whose only act was to warn that the package was **not installed**. Any failure at all — bad config, a missing peer, a deliberate refusal, a genuine bug in a constructor — came out as an absent package, and the operator went off to install something they already had.

The measured instance (#6915 / PR #7924): `InMemoryDriver`'s constructor refuses a non-`single` tenancy posture with a message naming the detected posture, both env knobs (`OS_TENANCY_POSTURE` / `OS_MULTI_ORG_ENABLED`) and the `@objectstack/driver-sql` remedy including `connection: { filename: ':memory:' }`. Under `OS_TENANCY_POSTURE=isolated` an operator saw none of it — only:

```
✘ @objectstack/runtime or @objectstack/driver-memory not installed — skipping driver
```

A well-written refusal, replaced by a false diagnosis.

## What changed

Each `catch` binds the error and tells the two cases apart:

- **Absent** — today's wording and today's advice, unchanged, plus the resolver's own message appended.
- **Present but failed** — a distinct line at `error` level saying the package **is** installed, that installing it again will not help, and carrying the underlying `code` and `message` verbatim.

Neither arm swallows the original error. That was the whole reason this defect survived: the `catch` discarded the evidence that would have named it.

### Detection

The classifier is the module system's own resolution verdict, never a message match:

| loader path | code | measured |
|:---|:---|:---|
| ESM `await import()` (the `esm` build) | `ERR_MODULE_NOT_FOUND` | node v22.22 |
| CJS `require()` (the `cjs` build of the same call) | `MODULE_NOT_FOUND` | node v22.22 |

Both spellings are live because this package ships both formats. `err.code` is the only structured field Node attaches — measured, `Object.getOwnPropertyNames(err)` is `['stack', 'code', 'message']`, with no `url` naming the failed specifier.

This is **not** in tension with the "which stage threw" classifier the organizations block uses one screen below, and not a competing convention: both refuse to read a *plugin's* private refusal semantics. `ERR_MODULE_NOT_FOUND` is the module system's verdict about resolution, which is exactly the fact being classified here.

**The code is read through the `cause` chain**, and that too is measured rather than assumed. A loader failure does not always arrive bare: `@vitest/mocker`'s `createHelpfulError` wraps every `vi.mock` factory throw in its own uncoded `Error` with the real one on `cause` — and a throwing factory is exactly how this repo simulates an absent package in `dev-plugin.test.ts`. Reading only the outer error would have classified every mocked-absent package as present-but-failed, i.e. broken the pins this card requires to stay green. The tie-break is that the **outermost error carrying a `code` decides**: a typed refusal is authoritative about itself, so a constructor that failed while reaching for a lazy optional peer stays a construction failure. Pinned by its own test.

**One honest limit, and why both arms print the cause.** A package that resolves but whose *own* dependency does not raises the same code (measured), so the absent arm can fire for a package that is itself installed. The resolver's message names the specifier that actually failed, so the appended cause keeps "install X" actionable rather than misdirecting to the package we asked for.

## Sites

All eleven optional loads: objectql, driver, app metadata, i18n, storage, realtime, auth, the setup/account app packages, security, REST, dispatcher.

**One site genuinely differs — REST.** Its #3963 no-auth precondition was a `throw` *inside* the load `try`, so DevPlugin's own refusal to serve a data API without auth was reported as `ℹ @objectstack/rest not installed`, at debug — this card's defect, in the one instance the file produced against its **own** words rather than a package's. That check now runs before the import and reports itself. Only the diagnosis changed: the REST plugin is not registered either way and `init()` still returns.

Two remaining bare `catch` blocks are deliberately untouched because they are a different class: the `ctx.getService(svc)` probe that maps "throws" to "slot is empty", and `destroy()`'s cleanup swallow. Neither emits a diagnosis about why anything failed.

## Tests

`packages/plugins/plugin-dev/src/dev-plugin-optional-load-failure.test.ts` (new, 9 cases). The existing absent-case pins in `dev-plugin.test.ts:263-264` are **untouched and still green** — they are correct for a genuinely absent package. The new file adds the second outcome:

- a driver whose constructor throws is not called "not installed";
- the underlying `code` and every actionable fact of the message reach the log;
- a module body that throws at evaluation time classifies the same way;
- an absent package still gets today's wording, at its normal level, and now names the failing specifier;
- a typed refusal whose `cause` is a module-not-found stays a construction failure (the tie-break), with the nested cause still printed;
- both arms in one boot, so the distinction is a property of the code path and not of which file ran;
- the #3963 refusal reports itself instead of blaming `@objectstack/rest`.

`pnpm --filter @objectstack/plugin-dev typecheck` clean; `pnpm --filter @objectstack/plugin-dev test` → **5 files, 54 tests passed**.

## Reverse verification — measured

The discrimination was reverted at the driver site only (the pre-#7926 bare `catch` restored verbatim), with the file's state asserted inside the run itself so the result cannot be an artifact of a stale tree:

```
== file state at run time ==
482:        // REVERSE VERIFICATION ONLY — the pre-#7926 bare catch, restored.
      } catch {
        ctx.logger.warn('  ✘ @objectstack/runtime or @objectstack/driver-memory not installed — skipping driver');
      }
```

The new pin fails **by producing the "not installed" wording** — the defect reproduced, not a compile error:

```
FAIL  src/dev-plugin-optional-load-failure.test.ts > … > does NOT report the driver as "not installed" when its constructor throws
AssertionError: expected '  ✘ @objectstack/runtime or @objectst…' not to contain 'not installed'

Expected: "not installed"
Received: "  ✘ @objectstack/runtime or @objectstack/driver-memory not installed — skipping driver"
```

Three more fail alongside it, each because the present-but-failed line does not exist at all under the reverted code (`expected undefined to be defined`): the `code`+`message` surfacing case, the both-packages-named case, and the `cause` tie-break case.

```
Test Files  1 failed | 4 passed (5)
     Tests  4 failed | 50 passed (54)
```

The 50 that stay green include every existing absent-case pin — `dev-plugin.test.ts` is untouched by the revert, which is the point: the new discrimination is what fails, and only it. Restoring the fix returns the suite to 54/54.

## Census gate

`check:driver-memory-census` (inside the ESLint job) went red on this branch: the new test's `vi.mock('@objectstack/driver-memory')` is a *declaration* the #6664 ledger did not cover. Recorded rather than silenced — `scripts/driver-memory-census.ledger.json` gains an entry on the existing `mock-replacement` axis, with its reason: the factory supplies its own `InMemoryDriver` whose constructor throws and never calls `importOriginal`, so the real driver is not loaded, and it is not a migration candidate because DevPlugin resolves the specifier **by name** (a test of that load site must name it) and nothing is stored (no test backend for sqlite `:memory:` to replace). The ruled set is unchanged, so both ruled files' `#6664 census: 2 ruled consumers` marker still holds. Gate now green locally, self-test included.

## Out of scope, deliberately

Whether DevPlugin should **refuse to start** when a driver refuses is a product-shape question (#7926 pins it as out of scope). Behaviour is unchanged: a failed slot stays empty and `init()` returns.

---

### DIFF
--- a/.changeset/dev-plugin-installed-but-failed-load.md
+++ b/.changeset/dev-plugin-installed-but-failed-load.md
@@ -1,0 +1,56 @@
+---
+"@objectstack/plugin-dev": patch
+---
+
+fix(plugin-dev): a service that IS installed and fails to construct is no longer reported as "not installed" (#7926)
+
+Every optional-service load in `DevPlugin.init()` ended in a bare `catch {}`
+whose only act was to warn that the package was **not installed**. So any
+failure at all — a bad config, a missing peer, a deliberate refusal, a genuine
+bug in a constructor — came out as an absent package, and the operator went off
+to install something they already had.
+
+The measured instance (#6915 / PR #7924): `InMemoryDriver`'s constructor refuses
+a non-`single` tenancy posture with a message naming the detected posture, both
+env knobs (`OS_TENANCY_POSTURE` / `OS_MULTI_ORG_ENABLED`) and the
+`@objectstack/driver-sql` remedy including `connection: { filename: ':memory:' }`.
+Under `OS_TENANCY_POSTURE=isolated` an operator saw none of it — only
+`✘ @objectstack/runtime or @objectstack/driver-memory not installed — skipping
+driver`. A well-written refusal replaced by a false diagnosis, which is the
+shape Prime Directive #10 forbids.
+
+Each `catch` now binds the error and tells the two cases apart:
+
+- **Absent** — today's wording and today's advice, unchanged, plus the
+  resolver's own message. That last part matters for the case the code alone
+  cannot separate: a package that resolves but whose *own* dependency does not
+  raises the same code, and the appended message names the specifier that
+  actually failed, so "install X" stays actionable.
+- **Present but failed** — a distinct line at `error` level that says the
+  package **is** installed, states that installing it again will not help, and
+  carries the underlying `code` and `message` verbatim.
+
+The classifier is the module system's own resolution verdict —
+`ERR_MODULE_NOT_FOUND` from the ESM loader, `MODULE_NOT_FOUND` from the CJS
+build's `require()`, both measured on node v22 rather than assumed — never a
+match against message text. It reads no plugin's private refusal semantics, so
+it does not compete with the "which stage threw" classifier the organizations
+block uses one screen below.
+
+The code is read through the error's `cause` chain, because a loader failure
+does not always arrive bare: a host that transforms modules can hand back its
+own error with the real one on `cause`. The **outermost** error carrying a
+`code` decides, so a plugin's typed refusal is never re-read as a resolution
+failure just because something deeper in its chain happens to be one.
+
+Fixed at all eleven optional loads (objectql, driver, app metadata, i18n,
+storage, realtime, auth, the setup/account app packages, security, REST,
+dispatcher). The REST site differed and is handled on its own terms: its `#3963`
+no-auth precondition was a `throw` *inside* the load `try`, so DevPlugin's own
+refusal to serve a data API without auth was reported as
+`ℹ @objectstack/rest not installed` at debug level. That check now runs before
+the import and reports itself.
+
+Behaviour is otherwise unchanged: a failed slot stays empty and `init()` still
+returns. Whether DevPlugin should refuse to start when a driver refuses is a
+product-shape question and is deliberately not decided here.
--- a/packages/plugins/plugin-dev/src/dev-plugin-optional-load-failure.test.ts
+++ b/packages/plugins/plugin-dev/src/dev-plugin-optional-load-failure.test.ts
@@ -1,0 +1,288 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+//
+// #7926 — ABSENT vs. PRESENT-BUT-FAILED, at every optional-service load.
+//
+// Every load in `DevPlugin.init()` used to end in a bare `catch {}` whose only
+// act was to warn that the package was "not installed". A package that IS
+// installed and threw while loading or constructing was therefore reported as
+// an absent one, and the operator went off to install something they already
+// had. The measured instance (#6915 / PR #7924): `InMemoryDriver`'s constructor
+// refuses a non-`single` tenancy posture with a message naming the detected
+// posture, both env knobs and the `@objectstack/driver-sql` remedy — and an
+// operator saw `✘ @objectstack/runtime or @objectstack/driver-memory not
+// installed — skipping driver` instead. Not one word of the refusal survived.
+//
+// ── What this file mocks, and why it must ──────────────────────────────────
+// "The package is present and it threw" cannot be observed without a present
+// package that throws, so the mocks supply exactly that and nothing else: the
+// classification under test stays entirely in dev-plugin.ts. The same reasoning
+// dev-plugin-tenancy-mount-refusal.test.ts records for its own mock.
+//
+// The ABSENT arm is pinned in dev-plugin.test.ts (`service-storage not
+// installed` / `service-realtime not installed`) and is deliberately NOT
+// repurposed here — those assertions are correct for a genuinely absent package
+// and must keep passing untouched. This file adds the second outcome; the pair
+// is what makes the distinction real. One absent-arm case is re-pinned below
+// alongside its failed-arm twin, because "these two inputs produce two
+// different diagnoses" is the claim, and a claim about a distinction cannot be
+// tested from one side.
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { DevPlugin } from './dev-plugin';
+
+/** How the mocked packages should misbehave for the current test. */
+const behaviour = vi.hoisted(() => ({
+  /** `driver-memory` resolves; its constructor throws (the #6915 shape). */
+  driverConstructorThrows: true,
+  /** The constructor's refusal carries a module-not-found error as its cause. */
+  driverErrorWrapsModuleNotFound: false,
+  /** `service-storage` resolves; its module body throws at evaluation time. */
+  storageModuleThrows: true,
+}));
+
+/** The shape Node's loader raises for a genuinely absent package. */
+const absent = (pkg: string) =>
+  Object.assign(new Error(`Cannot find package '${pkg}' imported from /app/dev-plugin.js`), {
+    code: 'ERR_MODULE_NOT_FOUND',
+  });
+
+// Present, and healthy enough to be constructed — the driver mock is what
+// fails, so `DriverPlugin` must exist for the failure to be attributable.
+vi.mock('@objectstack/runtime', () => ({
+  DriverPlugin: class {
+    name = 'com.objectstack.plugin.driver';
+    version = '1.0.0';
+    constructor(_driver: unknown, _name: string) {}
+    async init() {}
+  },
+  AppPlugin: class {
+    name = 'com.objectstack.plugin.app';
+    version = '1.0.0';
+    async init() {}
+  },
+  createDispatcherPlugin: () => ({
+    name: 'com.objectstack.plugin.dispatcher',
+    version: '1.0.0',
+    init: async () => {},
+  }),
+}));
+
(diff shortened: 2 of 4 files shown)


hotlong:
**PM review — `domain:cli` seat (#6024).** Verdict: **accept on the substance — ⛔ NOT enqueued yet.**

## The hold

> *Reverse verification is queued behind the shared verification lock; this body will be updated with its measured output.*

That is a **required pin** from the dispatch, and it is the one that decides whether the new discrimination actually fires. Reporting it as outstanding rather than claiming it is the right call — but the PR stays draft until the measured output is here. Post it and I will flip and enqueue.

What it must show: with the discrimination reverted, the construction-failure test fails **by getting the "not installed" wording** — the defect reproduced, not a compile error.

## The substance is excellent, and one finding is load-bearing

**The `cause` chain.** `@vitest/mocker`'s `createHelpfulError` wraps every `vi.mock` factory throw in an uncoded `Error` with the real one on `cause` — and a throwing factory is exactly how this repo simulates an absent package. **Reading only the outer error would have classified every mocked-absent package as present-but-failed, breaking the very pins I required to stay green.** That is a trap that would have produced a green-looking PR with an inverted classifier, and it was found by measuring rather than assuming.

The tie-break is the right shape too: **the outermost error carrying a `code` decides**, so a typed refusal stays authoritative about itself and a constructor that failed while reaching for a lazy optional peer stays a construction failure. Pinned by its own test rather than left as prose.

**Detection was measured, both spellings.** `ERR_MODULE_NOT_FOUND` (ESM) and `MODULE_NOT_FOUND` (CJS), on node v22.22, because this package ships both formats — I asked for a measurement instead of an assumption and it turned up a second live case. Reading `err.code` rather than matching a message is exactly the line: it is the **module system's** verdict about resolution, not a plugin's private refusal semantics, so it does not collide with the "which stage threw" classifier one screen below.

**The honest limit is stated rather than smoothed:** a package that resolves but whose *own* dependency does not raises the same code, so the absent arm can fire for a package that is itself installed — which is why both arms print the cause, keeping "install X" pointed at the specifier that actually failed.

## The REST site is the best catch in the PR

Its #3963 no-auth precondition was a `throw` **inside** the load `try`, so DevPlugin's own refusal to serve a data API without auth surfaced as `ℹ @objectstack/rest not installed`. **This card's defect, in the one instance where the file produced it against its own words rather than a package's.** Moving the check ahead of the import so it reports itself — with behaviour otherwise unchanged — is precisely right, and it is the kind of thing only found by fixing all eleven sites instead of the reported one.

Leaving the two remaining bare catches alone is also correct, with the distinction that matters: neither emits a diagnosis about why anything failed.

The existing `:263-264` absent-case pins are untouched and green, as required. ⛔ They were never the thing to edit.

## Docs: verified clean

Checked with a control: `not installed` across the two flagged docs → **no hits**; control (`plugin-dev` / `DevPlugin`) → hits in both, so the scanner works and the absence is real. No doc pins the wording; nothing to update.

---

hotlong:
**PM review, addendum — the hold is released.** Both outstanding items are resolved on their merits. Flipping to ready; auto-merge on.

## The census was resolved the right way

The red was `check:driver-memory-census`, and the cause is worth recording: the new test's `vi.mock('@objectstack/driver-memory')` was an **unledgered declaration** — the #6664 defect shape, produced by a PR whose own subject is undeclared failure modes.

It was closed by **recording the disposition**, not by deleting the reference or loosening the gate, with a reasoned entry on the existing `mock-replacement` axis: the factory supplies its own throwing `InMemoryDriver` and never calls `importOriginal`, so the real driver is never loaded; and it is not a migration candidate because DevPlugin resolves the specifier **by name** with nothing stored, so there is no test backend for sqlite `:memory:` to replace. Ruled set unchanged, so both ruled files' *"#6664 census: 2 ruled consumers"* marker still holds.

That is the gate doing its job and being answered, rather than routed around.

## The reverse verification discarded its own first result — that is the detail I want on the record

The first attempt came back **all-green**, and it was **thrown away** rather than reported, because a duplicate script instance had raced its own restore step: the tree was no longer reverted when the suite ran. The re-run asserts the file's state **inside** the run — md5 plus the printed catch block — so the result cannot be a stale-tree artefact.

This lane has spent the day rejecting greens that prove nothing (#7685's probe, #7362's harness, #7848's zero-match glob, #7991's dist-reading gate). Applying that same suspicion to **one's own verification process**, and binning a convenient green because its precondition was unproven, is the discipline working where nobody would have checked.

The measured result is the right failure, in the right direction:

```
AssertionError: expected '  ✘ @objectstack/runtime or @objectst…' not to contain 'not installed'
Received: "  ✘ @objectstack/runtime or @objectstack/driver-memory not installed — skipping driver"
```

plus three more failing as `expected undefined to be defined` — the present-but-failed line does not exist **at all** under the reverted code. 4 failed / 50 passed, with **every existing absent-case pin green**, so only the new discrimination fails and only it.

## Two risks accepted, both stated rather than hidden

**The present-but-failed arm logs at `error`** while the absent arm keeps each slot's own level. A host treating `logger.error` as fatal will see an error line where a quiet debug used to be. Deliberate, and I agree with the reasoning: an absent optional package is ordinary; a present one that threw is a defect in that deployment.

**The absent arm can still fire for a package that is installed** but whose own dependency is not — measured, same code. Mitigated rather than concealed: both arms print the resolver's message, which names the specifier that actually failed, so "install X" stays pointed at the right thing.

---

Issue #7645, opened 2026-08-11
mcp-stdio-fail-closed c3: the stdio MCP transport starts but is permanently deaf (oclif arg parser pauses process.stdin)


huangyiirene:
The fail-closed contract this item exists to pin (clauses 0–2, 7) **PASSES and is correct** — please do not "fix" it: stdio enabled with no key throws at plugin start and the kernel rolls back with the port closed; an unknown/revoked key draws the same refusal with no anonymous-but-serving fallback; a real member key binds the principal to *that* member (not admin); and stdio off boots cleanly with an explanatory line while HTTP is still served. The defect below is only that the started transport never answers.

## Symptom

With the stdio transport successfully started (`[MCP] Server started (transport: stdio)` logged), the transport never answers a single JSON-RPC request.

- **Observed:** `initialize`, `tools/list`, `resources/list`, `resources/read` all time out with **zero bytes on stdout**, across 4 separate runs. Malformed input draws no error either. Clause 3 fails; clauses 4–6 are blocked as a consequence.
- **Expected:** a started, principal-bound stdio transport answers `initialize` (and the rest) over the child's stdin/stdout.

An instrumented probe 40 s after boot reports `listeners(data)=1 readableFlowing=false isPaused=true bytesRead=0` — the data listener *is* attached, but `process.stdin` is **paused**.

## Root cause

oclif's argument parser opens a `readline` Interface over `process.stdin` and closes it on its abort signal; `Interface.close()` calls `stdin.pause()`. Captured stack:

```
at Interface.pause (node:internal/readline/interface:539)
at Interface.close (node:internal/readline/interface:525)
at signal.addEventListener.once (@oclif/core/lib/parser/parse.js:64)
```

So the deafness is in the **host process**, above the plugin. Two controls isolate it there: (a) a bare `@modelcontextprotocol/sdk` `McpServer` + `StdioServerTransport` answers the identical bytes, and (b) ObjectStack's **own** `MCPServerRuntime({transport:'stdio'})` answers too — both in a plain node process. `MCPServerRuntime.start()` (`packages/mcp/src/mcp-server-runtime.ts:807-811`) simply constructs `new StdioServerTransport()` and `connect()`s it; nothing resumes `process.stdin` after oclif's parser has paused it. On `origin/main` the connect region carries no `stdin.resume()` mitigation.

[...]

Issue #8034, opened 2026-08-12
mcp-stdio-fail-closed c3 (follow-on): the stdio transport now answers, but registers ZERO tools — advertises `tools` capability, `tools/list` → -32601


baozhoutao:
Follow-on to #7645, found by re-testing it. **#7645 is genuinely fixed and this is not a regression of it** — the stdio transport is no longer deaf: `initialize` now answers immediately (it used to time out with zero bytes on stdout), and `resources/list`, `resources/templates/list` and `resources/read` all answer. The five fail-closed clauses were each re-proved from scratch rather than carried over, and they all hold. What the fix made *visible* is a second, independent gap that was previously masked by the deafness.

**This is not an RLS bypass.** The one principal-bound read path stdio does expose (the record resource template) was driven on both sides and the gate held: a member key reads its own row in full and gets `not found` for another principal's row; an admin key reads both — matching REST row-for-row.

## Symptom

The stdio server's `initialize` result advertises `capabilities.tools: {}`, but the tool surface is empty:

- `tools/list` → `-32601 Method not found`
- every `tools/call` → `-32601 Method not found`
- across **4 fresh boots**, and on retry at +3 s and +8 s after `notifications/initialized` (ruling out a registration race)

**Decisive isolation:** in the *same process*, with the *same* `osk_` key, the **HTTP** transport (`POST /api/v1/mcp`) returns all **11** tools. So the bridge, the key, and the principal are all fine — the tools are simply never registered on the stdio server.

Boot log corroborates: `[MCP] Bridged 4 resource endpoints`, `Agent prompts bridged`, `[MCP] Bridged 0 skill prompts` — and **no tool-bridging line at all**.

## Root cause

The object/action tool surface is built **per HTTP request**, on a throwaway server, and never on the long-lived stdio server.

`packages/mcp/src/mcp-server-runtime.ts` — `handleHttpRequest()` constructs a fresh `McpServer` per request and registers the tools onto *that* instance:

```ts
// mcp-server-runtime.ts:1157,1166 — inside handleHttpRequest()
registerObjectTools(server, opts.bridge, opts.toolOptions);
...
registerActionTools(server, opts.bridge as McpActionBridge, opts.toolOptions);
```

[...]

PR #8065, opened 2026-08-12
fix(cli): serve diagnostics to stderr so the stdio MCP channel carries only protocol (#7915)

### LINKED ISSUE #7915: `os serve` writes its banner and kernel logs to the stdout the stdio MCP transport owns
Found while fixing #7645 (PR #7914). Different defect, different seam — filed rather than folded in.

## Symptom

With `OS_MCP_STDIO_ENABLED=true`, `objectstack serve` uses `process.stdout` as the MCP JSON-RPC channel **and** as its ordinary human/log output. Both streams are interleaved on the same fd.

Measured on `origin/main` @ `d91fad5` (spawned CLI, `stdio: ['pipe','pipe','pipe']`, a real minted `osk_` key, full child stdout captured). The `initialize` result arrives on **line 517**, behind 516 lines of non-protocol text:

```
  1  (blank)
  2    Loading objectstack.config.ts...
  3  [StandaloneStack] no compiled artifact at '…/dist/objectstack.json' — booting without one …
  4  2026-08-12T04:02:36.854Z INFO Registered metadata loader: filesystem (file:)
     … ~490 more kernel log lines …
499    ✓ Server is ready
512    Plugins: 29 loaded
515    Press Ctrl+C to stop
517  {"result":{"protocolVersion":"2024-11-05", … },"jsonrpc":"2.0","id":1}
```

## Why it matters

The MCP stdio transport is newline-delimited JSON. A spec-conforming client's `ReadBuffer` `JSON.parse`s every line it reads off the server's stdout, so each of those 516 lines is a parse failure surfaced to the client as a transport error before the one valid frame arrives.

Two independent sources, so fixing one is not enough:

- **the human banner** — `console.log` in the serve command (`✓ Server is ready`, the plugin table, `Press Ctrl+C to stop`). Unconditional, so this holds at the default `OS_LOG_LEVEL=warn` too.
- **the kernel logger** — `INFO`/`WARN` records go to stdout (`ERROR` was observed on stderr). The capture above used `OS_LOG_LEVEL=info`; at the default level the WARN records and the boot-diagnostics block still land there.

Note the observed direction of the two defects: #7645 meant the channel carried *nothing*, which is also why this one had never been seen. With #7645 fixed the channel works and now demonstrably carries noise.

## Expected

When the stdio transport is enabled, `stdout` is reserved for the protocol and everything human — banner, logs, diagnostics — goes to `stderr` (the convention every other MCP stdio server follows). When it is off, nothing changes.

## Not investigated here

Which seam owns it: the serve command's banner/logger wiring, the kernel logger's stream choice, or a mode flag the MCP plugin sets when it claims stdout. Whoever picks it up should also decide whether `--json`-style stdout-purity rules already in `packages/cli/test/json-stdout-purity.e2e.test.ts` extend to this case.

## Repro

1. Spawn `node packages/cli/bin/run.js serve -p <port> --dev` with `stdio: ['pipe','pipe','pipe']`, `OS_MCP_STDIO_ENABLED=true` and a valid `osk_` key.
2. Capture the child's stdout in full.
3. Every line before the first JSON-RPC frame is protocol garbage the client must parse.

(On `main` today, step 3 needs PR #7914 to get a frame at all.)

hotlong:
Fixes #7915

**The completing half of #7645.** PR #7914 made the stdio MCP transport answer; until this, a
native `os serve` stdio session was still unusable — the banner and the kernel's log records
landed in the stdout the transport owns, and a conforming client's NDJSON parse fails on the
first non-frame line. Measured on the card's repro: the `initialize` result arrived on **line
517**, behind 516 lines of non-protocol text.

## What changed

`os serve`'s stdout is the protocol's, and nothing else's.

- **Every human line the command prints goes to stderr, explicitly.** The `console.log` sweep the
  card asked for: all 11 sites in `packages/cli/src/commands/serve.ts` now go through one
  `printDiagnostic` helper (stderr, and still suppressed while the #4012 boot-quiet window is
  open, so a quiet boot stays quiet). The startup banner and the boot-diagnostics replay
  (`printServerReady` / `printBootDiagnostics` and their helpers in `utils/format.ts`) write with
  `console.error`.
- **Everything else the process would put on stdout is forwarded to stderr for the life of the
  process** — `ObjectLogger`'s `debug`/`info`/`warn` records and the stray `console.log`s several
  packages emit during boot (`[StandaloneStack] no compiled artifact …` is one). `LoggerConfig`
  has a level but no destination knob, so the stream itself is the only seam that reaches writers
  the CLI does not own. This is the route `--json` already takes (#6217), reused:
  `reserveStdoutForJson` and the new `redirectStdoutToStderr` are one mechanism.
- **`os start` installs the same policy.** It prints a header and a few resolved values, then
  spawns `serve` with inherited stdio — so its own stdout is the same fd the child's transport
  writes frames to, and it is the invocation the stdio docs name
  (`OS_MCP_STDIO_ENABLED=true OS_MCP_STDIO_API_KEY=osk_… os start`).
- **`@objectstack/mcp`: the stdio transport holds its own channel to the real stdout**
  (`protocol-stdout.ts`). A host that intercepts `process.stdout.write` to move its diagnostics —
  which is exactly what `serve` now does — would otherwise swallow the protocol frames along with
  them. The transport claims the channel in every host and on every construction path, so frames
  never depend on who booted the plugin.
- Colour follows the destination stream (`chalk.level = chalkStderr.level`), so a `serve > log` in
  a terminal is still coloured and a `serve 2> log` does not collect ANSI escapes.

### Unconditional, deliberately

"Redirect to stderr when the stdio transport is active" needs a reliable signal at the moment each
line prints — before the config is read, before the plugin is loaded — and fails **silently and in
the worse direction** when that signal is wrong or late: a frame-corrupting line that appears only
in some boots is far harder to find than one that always does. Banners, boot progress and kernel
logs are diagnostics, not program output; stderr is where a CLI puts diagnostics, and in a terminal
it costs nothing because both streams render.

### Nothing is silenced

Every line still appears, on stderr — the boot-phase warnings #4012 rescued from the quiet window
included. A shell capturing both streams (`> log 2>&1`) sees exactly what it saw before.

## Who read the banner off stdout

Swept with control greps before moving anything (this repo's tests and scripts,
`docs/qa/platform-checklist/**`, the `create-objectstack` templates, CI workflows, and the sibling
`cloud` / `objectui` repos):

| Consumer | Verdict |
|:---|:---|
| `packages/cli/test/**` — 8 e2e files + the shared `runServe` harness | ours; updated in this PR to read stderr |
| `cloud/scripts/dev-local/run-stack.sh` — greps `Server is ready` | unaffected: it greps a log written with `> cloud.log 2>&1` |
| `.github/workflows/scaffold-e2e.yml` — boots `os start` | unaffected: `> server.log 2>&1`, readiness by `curl /health` |
| `docs/qa/platform-checklist/**` | no step reads serve's stdout (0 hits for `stdout`; control: 5 hits for `os serve`) |
| `create-objectstack` templates, `examples/`, `apps/` | no hits (control: `objectstack` matches throughout) |

No consumer parses `os serve`'s stdout as data.

## Tests

- **`packages/cli/test/serve-stdio-stdout-purity.e2e.test.ts`** (new) — boots the real CLI with
  `OS_MCP_STDIO_ENABLED=true` and a minted `osk_` key, speaks `initialize` down the child's stdin,
  and asserts on the **stream**: every line the child wrote to stdout since spawn parses as a
  JSON-RPC frame. Asserting "the client parsed OK" alone passes today for the wrong reason — the
  harness reads the pipe as one buffer and finds the frame behind any short banner. The same run
  carries the negative half: the banner and the kernel's `INFO` records must be present **on
  stderr**, so purity can never be reached by silencing.
- **`packages/mcp/src/protocol-stdout.test.ts`** (new) — frames reach the real stream while an
  interception is installed, the interception still catches everything else, and backpressure
  (`false` + `drain`) is reported rather than swallowed.
- Existing serve e2e files now assert against stderr. Note the negative assertions
  (`not.toContain(BANNER)`) had to move too: on an empty stdout they would have passed for the
  wrong reason.

## Verified

```
pnpm --filter @objectstack/mcp typecheck            EXIT 0
pnpm --filter @objectstack/cli typecheck            EXIT 0
pnpm --filter @objectstack/mcp test                 12 files / 127 tests passed
pnpm --filter @objectstack/cli test                115 files / 1256 tests passed   (405s)
```

The CLI run is the whole package suite, so it covers the new pin, all eight rewritten serve
e2e files and the in-process posture-gate ordering test.

**Reverse verification** — direction predicted first, then run. Putting a `console.log` back
proves nothing here: with the redirect installed it lands on stderr like every other
diagnostic. The reversal that can fail is removing the redirect itself, which is what carries
the writers this command does not own. With `redirectStdoutToStderr()` dropped from
`serve.run()` and the CLI rebuilt, the pin goes red on the purity assertion and enumerates the
lines a client would choke on:

```
STEP: REVERSE: drop redirectStdoutToStderr() from serve.run()
patched — call sites remaining: 0
 ❯ test/serve-stdio-stdout-purity.e2e.test.ts:306:7
   expect(nonFrameLines, `stdout carries ${nonFrameLines.length} line(s) …`).toEqual([])
+   "2026-08-12T14:12:21.278Z INFO AuditPlugin: sys_comment record-level access gates installed",
+   "[Registry] Registered namespace: sys -> com.objectstack.mcp.connect-agent-ui",
+   … 2026-08-12T14:12:21.345Z INFO HTTP server started successfully {"port":56516,…}
+   "2026-08-12T14:12:21.446Z INFO CryptoProvider configured for secret fields",
 Test Files  1 failed (1)
EXIT pin-reversed=1 (non-zero is the EXPECTED result)
```

Note what those lines are: kernel `INFO` records and another package's `console.log`, not this
command's own output — the half an explicit sweep alone would have left behind. Restored with
`git checkout` (the fix was committed first), call site back at 1, tree clean, CLI rebuilt.

---

### DIFF
--- a/.changeset/serve-diagnostics-to-stderr.md
+++ b/.changeset/serve-diagnostics-to-stderr.md
@@ -1,0 +1,49 @@
+---
+"@objectstack/cli": patch
+"@objectstack/mcp": patch
+---
+
+fix(cli): `os serve` writes its banner, boot progress and kernel logs to stderr, so the stdio MCP channel carries only protocol (#7915)
+
+With `OS_MCP_STDIO_ENABLED=true`, `objectstack serve` used `process.stdout` as
+the MCP JSON-RPC channel **and** as its ordinary human/log output. MCP stdio
+framing is newline-delimited JSON — a conforming client `JSON.parse`s every line
+it reads off the server's stdout — so every banner line and every `INFO`/`WARN`
+record reached the client as a transport error. Measured on the card's repro:
+the `initialize` result arrived on **line 517**, behind 516 lines of
+non-protocol text. It reads as "the transport is broken", which is also why it
+stayed invisible until #7645 (PR #7914) made the transport answer at all.
+
+**`serve`'s stdout is now the protocol's, and nothing else's.** Banners, boot
+progress and kernel logs are diagnostics, not program output, and stderr is
+where a CLI puts diagnostics — so they go there whether or not a stdio
+transport is mounted. Two halves:
+
+- every human line `serve` prints is written to stderr explicitly, the startup
+  banner (`✓ Server is ready`, the plugin table, `Press Ctrl+C to stop`) and the
+  boot-diagnostics replay included;
+- everything else the process would write to stdout — `ObjectLogger`'s
+  `debug`/`info`/`warn` records, and the stray `console.log`s several packages
+  emit during boot — is forwarded to stderr for the life of the process, the
+  same route `--json` already takes (#6217). `LoggerConfig` has a level but no
+  destination knob, so the stream itself is the only seam that covers writers
+  the CLI does not own.
+
+**Unconditional, deliberately.** "Redirect when the stdio transport is active"
+needs a reliable signal at the moment each line prints — before the config is
+read, before the plugin is loaded — and fails silently and in the worse
+direction when that signal is wrong or late: a frame-corrupting line that shows
+up only in some boots is far harder to find than one that always does. In a
+terminal the move costs nothing, since both streams render.
+
+**Nothing is silenced.** Every line still appears, on stderr — including the
+boot-phase warnings #4012 rescued from the quiet window. A shell that captured
+both streams (`> log 2>&1`) sees exactly what it saw before; one that captured
+stdout alone now finds `serve`'s output on stderr.
+
+`@objectstack/mcp`: the stdio transport now holds its own channel to the real
+stdout instead of writing through `process.stdout` — a host that intercepts
+`process.stdout.write` to move its diagnostics (which is what `serve` does)
+would otherwise swallow the protocol frames along with them. It claims that
+channel in every host and on every construction path, so a transport's frames
+never depend on who booted the plugin.
--- a/packages/cli/src/commands/serve-tenancy-posture-gate.test.ts
+++ b/packages/cli/src/commands/serve-tenancy-posture-gate.test.ts
@@ -193,8 +193,14 @@ describe('the gate runs before serve does ANY boot work', () => {
    * posture got as far as `Loading objectstack.config.ts…`, the whole plugin
    * slate, a persisted dev crypto key and a degraded kernel bootstrap before
    * anything refused. After it, `run()` reaches the gate and stops: the FATAL
-   * is the only thing written, and `console.log` — which is where every
-   * subsequent boot step reports — is never touched at all.
+   * is the only thing written, and the diagnostic stream — where every
+   * subsequent boot step reports — carries nothing else.
+   *
+   * That stream is **stderr** since #7915 (`serve` keeps stdout clear for the
+   * MCP stdio transport), so the capture below watches `process.stderr.write`
+   * rather than `console.log`. Watching `console.log` here would now be a
+   * phantom check: no boot step writes there any more, so the assertion could
+   * never fail, whatever the gate did.
    *
    * Note what is deliberately NOT asserted: "the port never listened". That was
    * true before the fix too (the escaping throw aborted kernel Phase 1, while
@@ -207,13 +213,19 @@ describe('the gate runs before serve does ANY boot work', () => {
     process.env.OS_TENANCY_POSTURE = 'bogus';
 
     const errors: string[] = [];
-    const logs: string[] = [];
+    const diagnostics: string[] = [];
     const errSpy = vi.spyOn(console, 'error').mockImplementation((...a: unknown[]) => {
       errors.push(a.join(' '));
     });
-    const logSpy = vi.spyOn(console, 'log').mockImplementation((...a: unknown[]) => {
-      logs.push(a.join(' '));
-    });
+    // Everything `serve` prints, at the stream: its own `printDiagnostic` lines
+    // AND anything the redirect installed at the top of `run()` forwards there.
+    const stdoutWrite = process.stdout.write;
+    const stderrSpy = vi
+      .spyOn(process.stderr, 'write')
+      .mockImplementation(((chunk: unknown) => {
+        diagnostics.push(String(chunk));
+        return true;
+      }) as typeof process.stderr.write);
     // The gate exits the PROCESS on purpose (a throw is what the broad
     // AuthPlugin catch used to swallow). Convert it to something catchable so
     // the test runner survives, and assert it was reached.
@@ -230,8 +242,12 @@ describe('the gate runs before serve does ANY boot work', () => {
       raised = err;
     } finally {
       errSpy.mockRestore();
-      logSpy.mockRestore();
+      stderrSpy.mockRestore();
       exitSpy.mockRestore();
+      // `run()` reserves stdout for the process's lifetime (#7915). Harmless in
+      // the real one-shot CLI; in a vitest worker it would outlive this case,
+      // so put the stream back.
+      process.stdout.write = stdoutWrite;
       if (savedNodeEnv === undefined) delete process.env.NODE_ENV;
       else process.env.NODE_ENV = savedNodeEnv;
     }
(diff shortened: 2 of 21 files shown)


hotlong:
**PM review — `domain:cli` seat (#6024).** Verdict: **accept**. Auto-merge on. No change requests.

## The dispatch's reverse verification was wrong, and the PR says so with proof

I asked for: *put one `console.log` back, confirm the stdout-purity test fails naming it.* Under this design that proves **nothing** — with the redirect installed it lands on stderr like every other diagnostic. The reversal that can actually fail is **removing the redirect itself**, and that is the one that was run:

```
+ "2026-08-12T14:12:21.278Z INFO AuditPlugin: sys_comment record-level access gates installed",
+ "[Registry] Registered namespace: sys -> com.objectstack.mcp.connect-agent-ui",
+ … INFO HTTP server started successfully {"port":56516,…}
```

Kernel `INFO` records and **another package's `console.log`** — *"the half an explicit sweep alone would have left behind."* Substituting a better reversal and demonstrating why mine would have been vacuous is exactly the reading I want; a brief is an instruction, not a ceiling.

## The sweep I asked for was half the problem

I framed this as 11 `console.log` sites. The other half is the writers the CLI **does not own** — `ObjectLogger`'s records and the stray `console.log`s several packages emit during boot (`[StandaloneStack] no compiled artifact …`). `LoggerConfig` carries a level but no destination knob, so **the stream itself is the only seam that reaches them**, and reusing the route `--json` already takes (#6217) makes it one mechanism rather than a second policy.

A PR that had done only what I asked would have passed its own tests and still corrupted the channel on a real boot. The measured repro says how much: the `initialize` result arrived on **line 517**, behind 516 lines of non-protocol text.

## The self-inflicted wound that was seen coming

`protocol-stdout.ts` — the stdio transport holds **its own channel to the real stdout**, because a host that intercepts `process.stdout.write` to move its diagnostics *is exactly what `serve` now does*, and would otherwise swallow the protocol frames along with them. The fix would have broken the thing it was fixing, on its own second step. Claiming the channel on every construction path so frames never depend on who booted the plugin is the right generalisation.

`os start` is caught for the same reason — it spawns `serve` with **inherited stdio**, so its own stdout is the fd the child's transport writes frames to, and it is the invocation the stdio docs actually name.

## Unconditional, and the pins cannot be satisfied by silence

The conditional was declined for the reason given at dispatch, restated more precisely here: the signal would be needed **before the config is read and before the plugin is loaded**. Good.

Two anti-vacuity moves worth naming:

- The purity pin asserts **on the stream** — every line the child wrote to stdout since spawn parses as a JSON-RPC frame — and the PR states why "the client parsed OK" alone would pass today for the wrong reason: the harness reads the pipe as one buffer and finds the frame behind any short banner.
- The **negative half rides the same run**: banner and kernel `INFO` records must be present **on stderr**. Purity can never be reached by silencing.
- And the existing `not.toContain(BANNER)` assertions were **moved** rather than left — on an empty stdout they would have started passing for the wrong reason. Catching that in tests you are merely updating is the habit that keeps a suite honest.

## Consumer sweep

Controlled and cross-repo — this repo's tests and scripts, `docs/qa/platform-checklist/**` (0 hits for `stdout`, control: 5 for `os serve`), the `create-objectstack` templates, CI workflows, and both sibling repos. `cloud/scripts/dev-local/run-stack.sh` greps `Server is ready` but from a `> cloud.log 2>&1` capture, and `scaffold-e2e.yml` gates on `curl /health` — both unaffected, and both checked rather than assumed. Nothing parses `os serve`'s stdout as data.

Colour following the destination stream is a nice touch: `serve > log` in a terminal stays coloured, `serve 2> log` does not collect escapes.

**This closes the stdio chain's second link.** #8034 (zero tools registered) is in flight as the third; the three only add up to a working session together.

---

hotlong:
**Docs verdict on the drift advisory — clean, and one additive opportunity.**

The advisory lists 27 pages. Checked the three that could plausibly be affected, with a control (`os serve` / `stdio` → 8 hits in `connect-mcp.mdx`, 15 in `cli.mdx`, so the scanner works and the absences below are real):

- `content/docs/deployment/cli.mdx:1125` and `:1210` do discuss a **stdout contract**, but the exception they declare is `os login` / `os cloud login` emitting **NDJSON under `--json`** — a statement about `--json` document *shape*, orthogonal to this change. Not made wrong.
- No page pipes or parses `os serve`'s stdout as data. Consistent with the cross-repo consumer sweep in the PR body.

**Nothing to correct — so nothing rides this PR.**

The additive opportunity, recorded rather than filed because it makes no existing sentence false: `content/docs/ai/connect-mcp.mdx` is the page a user follows to wire an MCP client to a native `os serve`, and after this lands there is a fact worth stating there — **stdout is the protocol channel; diagnostics are on stderr**. That is exactly what someone debugging a client that "connects but sees nothing" would want to read, and it is the natural home for it alongside the `OS_MCP_STDIO_ENABLED` invocation.

Worth folding into #8034's PR if that dev is in the same area, since the three stdio links only make sense to a reader together — otherwise a small docs-only follow-up. ⛔ Not a change request on this PR.

---

PR #8076, opened 2026-08-12
fix(client): derive invitation `status` from the spec's InvitationStatus union

### LINKED ISSUE #7781: client SDK types the invitation `status` union by hand, and it is missing `expired` (same divergence family as #7726)

hotlong:
Fixes #7781

## What

`organizations.invitations.list()`'s row `status` was hand-written as
`'pending' | 'accepted' | 'rejected' | 'canceled'` — missing `expired`,
ObjectStack's own terminal state driven by `expiresAt`. `listMine()` was worse:
a bare `string`. Both are two more hand-copied spellings of the vocabulary
`InvitationStatus` (`packages/spec/src/identity/organization.zod.ts`) already
owns — same divergence family as #7726, which had already widened the spec
side to five values (adding `canceled`) while this file stayed at four and
drifted the other way.

Per the ruling on the issue: **derive, don't extend.** Both methods now type
`status: InvitationStatus`, imported from `@objectstack/spec/identity`
(`@objectstack/client` already depends on `@objectstack/spec` and already
imports several other types from it, e.g. `ApprovalStatus`), so a future value
added to the spec enum reaches the SDK by construction instead of silently
diverging again.

## Both directions checked

The card only reported `expired` missing. I also checked the other direction —
whether every value the SDK's old hand-written literal listed is a real spec
value. All four (`pending`, `accepted`, `rejected`, `canceled`) exist in the
spec enum, so there's no SDK-only dead surface to report.

## The pin

`packages/client/src/invitation-status-vocabulary.test.ts`:

- A **type-level** pin (`Assert<Eq<...>>`, same pattern as
  `packages/spec/src/api/meta-item-response-shapes.test.ts`) asserting each
  method's declared row `status` type is exactly `InvitationStatus` — not a
  literal that merely lists the same values today. Read by `tsc` via
  `tsconfig.test.json` (`pnpm --filter @objectstack/client typecheck`);
  `vitest` does not type-check.
- A runtime companion pinning the enum's actual five-value content in prose,
  plus a check that the SDK's old four values are all real spec values (the
  "other direction" above), and a refusal check for a near-miss spelling.

**Reverse-verification, both directions:**

1. Removed `expired` from the spec enum, rebuilt `@objectstack/spec`, ran
   `pnpm --filter @objectstack/client typecheck` — **stayed green**. This is
   expected, not a gap: since the fix derives the client's type directly from
   `InvitationStatus` rather than restating it, the two are the same type by
   construction and cannot diverge from a spec-side change alone — that's the
   fix working as intended. Restored the spec file and rebuilt.
2. The regression this pin actually guards against is someone reverting the
   derivation. Verified that directly: temporarily re-literalized `list()`'s
   `status` back to the old hand-written 4-value union (leaving the pin test
   untouched) and re-ran typecheck:

   ```
   check:test-typecheck: 1 problem(s)
     • src/invitation-status-vocabulary.test.ts: 1 type error(s) in a file the
       ledger does not cover. Fix them — this file is inside the checked zone...
   ELIFECYCLE  Command failed with exit code 1.
   ```

   Restored the fix and confirmed green again (`check:test-typecheck: OK — 0
   file(s) / 0 error(s)`).

Types-only, no wire change — the value already arrived off the wire regardless
of what the annotation said. Changeset included.

## Tests

- `pnpm --filter @objectstack/client typecheck` — clean (`tsc --noEmit` +
  `check:test-typecheck`, 0 debt entries).
- `pnpm --filter @objectstack/client test -- --maxWorkers=2` — 22 test files /
  287 tests passed, including the 3 new pin tests.

---

### DIFF
--- a/.changeset/invitation-status-derive-from-spec.md
+++ b/.changeset/invitation-status-derive-from-spec.md
@@ -1,0 +1,17 @@
+---
+'@objectstack/client': patch
+---
+
+fix(client): `organizations.invitations.list()` / `listMine()` type `status` from the spec's `InvitationStatus` enum instead of a hand-copied literal (#7781).
+
+`list()`'s row `status` was hand-written as `'pending' | 'accepted' | 'rejected' | 'canceled'` —
+missing `expired`, ObjectStack's own terminal state driven by `expiresAt`. `listMine()` typed the
+same field as a bare `string`. Both are now `InvitationStatus`, imported from
+`@objectstack/spec/identity` — the same union `sys_invitation.status` binds its select options to
+(#7726) — so a value added to the spec enum reaches the SDK by construction instead of silently
+diverging again.
+
+Types-only, no wire change: the value already arrived off the wire regardless of what the
+annotation said, so nothing about what `list()` / `listMine()` return at runtime moves. What
+changes is that TypeScript narrowing (a `switch` over `status`, for example) now sees all five
+values, including `expired`.
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -84,6 +84,7 @@ import type {
   ApprovalDecisionResult,
 } from '@objectstack/spec/contracts';
 import type { ExecutionStatus } from '@objectstack/spec/automation';
+import type { InvitationStatus } from '@objectstack/spec/identity';
 import { Logger, createLogger } from '@objectstack/core/logger';
 import { RealtimeAPI } from './realtime-api';
 
@@ -2011,8 +2012,8 @@ export class ObjectStackClient {
      */
     invitations: {
       /**
-       * List pending/accepted/canceled invitations for an organization.
-       * Requires owner/admin role on that org.
+       * List pending/accepted/rejected/expired/canceled invitations for an
+       * organization. Requires owner/admin role on that org.
        *
        * better-auth: GET /organization/list-invitations?organizationId=…
        */
@@ -2023,11 +2024,16 @@ export class ObjectStackClient {
         );
         const data = await res.json();
         const invitations = Array.isArray(data) ? data : (data?.data ?? data?.invitations ?? []);
+        // [#7781] `status` is `InvitationStatus` (from `@objectstack/spec/identity`)
+        // rather than a hand-copied literal — the SDK previously restated the
+        // vocabulary and drifted from it (missing `expired`). Derived from the
+        // spec union, so a future value reaches here by construction; see
+        // `invitation-status-vocabulary.test.ts` for the pin.
         return { invitations: invitations as Array<{
           id: string;
           email: string;
           role: string;
-          status: 'pending' | 'accepted' | 'rejected' | 'canceled';
+          status: InvitationStatus;
           organizationId: string;
           inviterId: string;
           expiresAt: string;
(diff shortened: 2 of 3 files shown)


hotlong:

Derive-don't-extend, as the issue ruled: both methods now type `status: InvitationStatus` from `@objectstack/spec/identity`, so the next value added to the vocabulary reaches the SDK by construction rather than by someone remembering. `listMine()`'s bare `string` was the worse of the two and is gone with the same change.

You also checked the direction the card did not ask about — whether every value the old hand-written literal listed is a *real* spec value — and reported the negative result (all four exist, no SDK-only dead surface). A vocabulary drift card is only half-answered by the missing value; the other half is the invented one, and it has to be looked for to be ruled out.

### The reverse verification is the part worth recording

The obvious reversal — delete `expired` from the spec enum, rebuild, re-run the client typecheck — **stayed green**. The report says so plainly, and then explains why that is the fix working rather than a hole: once the client's type is derived from `InvitationStatus` instead of restating it, the two are the same type by construction and *cannot* diverge from a spec-side edit alone. A weaker report would have quietly dropped the green run and shown only a probe that went red.

Then it found the reversal that actually bites — re-literalizing `list()`'s `status` back to the old four-value union, leaving the pin untouched — and showed the specific failure:

```
check:test-typecheck: 1 problem(s)
  • src/invitation-status-vocabulary.test.ts: 1 type error(s) in a file the ledger does not cover.
```

That is precisely the lesson this seat paid for on #7915 earlier today, where **my** prescribed reversal was vacuous under the developer's design and the reversal that bit was a different one. Here it was caught without being told. The generalization, for the seat post: *a reverse-verification that stays green is a result, not a failed attempt — report it, explain which property makes it green, then go find the perturbation the pin actually guards.*

One more thing done right and easy to get wrong: the type-level assertion is routed through `tsconfig.test.json` so `tsc` reads it, with the note that **`vitest` does not type-check**. A type-level pin living only in a vitest-run file asserts nothing at all — it is the purest form of a check that cannot fail, and it would have looked like coverage in the file listing forever.

Types-only, no wire change; the value already arrived off the wire regardless of the annotation. Changeset included. All 25 checks green. **Flipping ready and enabling auto-merge.**

---

PR #8077, opened 2026-08-12
test(runtime): pin the ambient-vs-bare transaction split behind the single-connection stall (#7842)

### LINKED ISSUE #7842: The single-connection-pool stall's ambient-vs-bare-transaction split is pinned only by PR prose — needs an integration test in packages/runtime
Filed by the `domain:metadata` PM seat as the follow-up #7708's dev offered and deliberately did **not** take, to keep that card comment-only. Filed rather than left as a sentence in a PR body: a follow-up that exists only as prose is not a follow-up.

No `domain:*` applied — routing is the triage seat's territory. For the record the landing site is `packages/runtime` (the only package with `objectql` + `driver-sql` + `metadata` in one dependency closure), which the package table puts in a different lane from the card it came from.

## What #7708 established, and what is now unpinned

PR #7840 re-anchored `MetadataManager`'s `listCache` policy comment onto a verified witness. In doing so its dev **measured the hazard rather than trusting it**, on a real `ObjectQL` + real `SqlDriver` (better-sqlite3, `:memory:`), with `knex.client.pool.max === 1` confirmed at runtime and `acquireConnectionTimeout` verified as never overridden in `driver-sql` (knex default 60s):

| Case | Transaction opened via | Measured |
|---|---|---|
| **B** | `driver.beginTransaction()` — no ambient store | **stalls**, then `Knex: Timeout acquiring a connection` — **60 085 ms** |
| **A** | `engine.transaction()` — ambient `txStore` published | returns in **12 ms** |
| control | none | returns immediately |

Case A was checked against its false negative (that `engine.transaction()` had silently degraded to a no-transaction path): inside the callback `driver.activeTransactions === 1`, `txStore.getStore().transaction` present, and a spy on `driver.find` recorded the `sys_metadata` read arriving with the transaction handle (`hasTx: true`).

**The consequence is that the hazard is conditional, not flat**: it needs an open transaction the engine's ambient `txStore` (ADR-0034) cannot see, because `buildDriverOptions` prefers an explicit handle and falls back to the ambient store, threading an engine-opened transaction onto the read on the loader's behalf.

## The gap

That A/B split is now **stated in a code comment and justified only in PR #7840's prose**. Nothing executes it. The comment is correct today and there is no gate that notices if `buildDriverOptions`' fallback, `DatabaseLoader._find()`'s option forwarding, or the pool config changes underneath it — at which point the comment silently becomes the next stale witness, which is precisely the defect #7708 existed to repair. Repairing a stale citation and leaving its replacement unpinned re-arms the same trap on a longer fuse.

## Proposed shape

An integration test in `packages/runtime` asserting both directions:

- ambient-published transaction (`engine.transaction()` / `ScopedContext`) ⇒ the loader's read is threaded and completes;
- bare `driver.beginTransaction()` with no ambient store ⇒ stalls.

The worked precedent is `sql-driver-sqlite-tx-guard.test.ts`, which keeps the driver's own `parentTrx` contract honest the same way.

⚠️ **Design constraint for whoever takes it**: the negative direction is a 60-second knex acquire timeout. Asserting it naively means a 60 s test. Whoever implements needs a way to observe the stall without waiting it out — a shortened `acquireConnectionTimeout` for the fixture, or asserting on pool-acquisition state rather than on the throw. **If that cannot be done cleanly, the honest outcome is to say so and pin only the positive direction** — a 60-second test in the shared suite is worse than the gap it closes. That trade-off is the actual content of this card, not the test itself.

## Provenance

#7708 · PR #7840 (`88ebdc9`) · the retirement chain #6656 → PR #7081 that created the original stale witness.

hotlong:
Fixes #7842

Executes the ambient-vs-bare split that `MetadataManager`'s `listCache` policy comment asserts and that PR #7840 measured but left pinned only by prose. Test-only; no production file is touched.

## Which of the two approaches worked

**Approach 1 — a shortened `acquireConnectionTimeout` on the fixture's driver.** It works cleanly, so approach 2 (asserting on pool-acquisition state) was not needed and neither direction had to be dropped. **Both directions are pinned.**

`SqlDriverConfig` is `Knex.Config &` a few ObjectStack-only keys, and the constructor strips only those before handing the rest to knex, so the knob arrives untouched. That shortens the **wait**, not the **shape**: the read still fails on connection acquisition, with knex's own `Timeout acquiring a connection` message, exactly as it does in production at the 60s default.

| Case | Measured here | PR #7840 measured |
|---|---|---|
| ambient — `engine.transaction()` | returns, **97 ms** | 12 ms |
| bare — `driver.beginTransaction()` | `Knex: Timeout acquiring a connection`, **443 ms** at a 400 ms fixture bound | same throw, 60 085 ms at the knex default |

Whole file: **2 tests, 542 ms.** Nothing in the shared suite waits 60 seconds.

`assertEffectiveTimeout` re-reads both facts off the **live** knex client — `pool.max === 1` and `acquireConnectionTimeout === 400` — so a change to the SQLite pool config, or a driver that stops forwarding the knob, fails with the reason named rather than as a mysterious green (nothing left to stall) or a mysterious hang (back at the default).

## The positive direction survives its own false negative

"The read completed" passes trivially if no transaction was ever opened, so all three of PR #7840's own checks are asserted inside the callback:

- `driver.activeTransactions === 1`
- `txStore.getStore().transaction` is present
- a spy on `driver.find` records the `sys_metadata` read arriving **with** the handle (`hasTx: true`)

The third is the load-bearing one — the only assertion that separates "the ambient store was threaded onto the loader's read" from "the read happened to succeed on its own".

The negative direction gets the mirror guard: the read really was attempted, really did reach the driver, and arrived **without** a handle (`hasTx: false`). A rejection alone cannot tell "nothing threaded the caller's transaction" apart from "the read never went out".

## Reverse verification

Direction predicted before running: deleting `buildDriverOptions`' ambient fallback (`: this.txStore.getStore()?.transaction` to `: undefined` in `packages/objectql/src/engine.ts`), rebuilding `@objectstack/objectql`, should turn the **ambient** case red and leave the **bare** case green. Measured, exactly that:

```
 x AMBIENT ... 502ms
   -> Knex: Timeout acquiring a connection. The pool is probably full.
 v BARE ... 425ms
 Test Files  1 failed (1)
      Tests  1 failed | 1 passed (2)
```

Reported precisely, because the template's presumption does not quite fit: the red arrives as the **stall**, not at the `hasTx` assertion — with no handle threaded the read never returns to reach that line. `hasTx` is what discriminates in the other failure mode, where a read completes without the ambient handle. The edit was reverted and `engine.ts` verified byte-identical to `origin/main` (`git diff origin/main --quiet` clean) before the branch was pushed.

## Premise re-checked, not assumed

The comment is **correct today** — this card is not inverted. `DatabaseLoader._find()` still forwards no options on the engine path (`return this.engine.find(table, query)`), knex still gives the SQLite dialect a single-connection pool, and `buildDriverOptions` still prefers an explicit handle and falls back to the ambient store. The one number not re-measured is the 60s itself: that is knex's default, and not waiting it out is the entire point of the card.

## Scope

`packages/runtime` is the landing site — the only package carrying `objectql` + `driver-sql` + `metadata` in one dependency closure. `buildDriverOptions`, `DatabaseLoader._find()` and the pool config are **unchanged**; this PR pins current behaviour. Worked precedent: `sql-driver-sqlite-tx-guard.test.ts`.

## One deliberate deviation from the dispatch brief, flagged for the PM

The brief's definition of done asked for a changeset. I applied the **`skip-changeset`** label instead and wrote none, because this PR releases nothing and the repo has already ruled on that case in two places: the triage comment on #7842 itself ("`skip-changeset` applies (tests-only)"), and `pr-automation.yml`'s own failure text — *"It releases nothing (.github/, .claude/, skills/, docs/, content/, examples/, tests-only, and the like) -> apply the 'skip-changeset' label. <<< PREFERRED"*, with the empty-frontmatter route explicitly closed by #5471. A real changeset naming `@objectstack/runtime` would declare a patch release whose shipped diff is empty, since `files` publishes `dist` only. Happy to add one if that reading is wrong.

## Verification

- `pnpm --filter @objectstack/runtime typecheck` — clean.
- `@objectstack/runtime` full suite — **143 files, 2173 tests, all passing** (112 s).
- `check:type-check-debt`, the `@objectstack/runtime` entry, measured the way the script measures it (a `tsconfig.debt-remeasure.json` over the un-excluded test layer, built closure in place): **227 — unchanged**, the recorded value, with **0** errors from the new file. The first draft measured 228; `engine.registry.registerObject` requires a `packageId`, so the fixture now registers through the engine's own proxy, which supplies one.
- `node scripts/check-nul-bytes.mjs` — OK, 7332 files. Plus the wider self-scan `grep -naP '[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]'` over the new file — clean.
- `eslint` on the new file — clean.

---

### DIFF
--- a/packages/runtime/src/metadata-list-ambient-vs-bare-transaction.integration.test.ts
+++ b/packages/runtime/src/metadata-list-ambient-vs-bare-transaction.integration.test.ts
@@ -1,0 +1,243 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * #7842 — executes the ambient-vs-bare split that `MetadataManager`'s
+ * `listCache` policy comment asserts.
+ *
+ * PR #7840 (#7708) re-anchored that comment onto a fresh measurement instead of
+ * a retired witness, and the measurement NARROWED the hazard rather than
+ * retiring it: on a real `ObjectQL` + real `SqlDriver` (better-sqlite3,
+ * `:memory:`, knex pool max 1), a metadata read issued while a transaction is
+ * open stalls out `acquireConnectionTimeout` — but ONLY when the transaction was
+ * opened somewhere the engine's ambient `txStore` (ADR-0034) cannot see it. A
+ * transaction opened through `engine.transaction()` is published into that store,
+ * `buildDriverOptions` threads it onto the loader's read, and the read returns at
+ * once on the connection the transaction already holds.
+ *
+ * That split lived only in a code comment and in PR prose, so nothing noticed if
+ * `buildDriverOptions`' ambient fallback, `DatabaseLoader._find()`'s option
+ * forwarding, or the SQLite pool config changed underneath it — at which point
+ * the comment silently becomes the next stale witness, which is the defect #7708
+ * existed to repair. This file is the executable half. It PINS current behaviour
+ * and changes none of it.
+ *
+ * ## The 60-second problem, and what this fixture does about it
+ *
+ * The negative direction is an acquire timeout, and knex's default is 60s — a
+ * 60-second test in the shared suite is worse than the gap it closes. The
+ * fixture's driver therefore declares its own short `acquireConnectionTimeout`
+ * ({@link ACQUIRE_MS}); `SqlDriverConfig` is `Knex.Config &` extras and the
+ * constructor forwards everything it does not itself consume, so the knob
+ * arrives at knex untouched. That shortens the WAIT, not the SHAPE: the read
+ * still fails on connection acquisition, with knex's own message, exactly as it
+ * does in production at 60s. `assertEffectiveTimeout` re-reads the value off the
+ * live knex client so a driver that stopped forwarding it cannot leave this test
+ * quietly measuring the default.
+ *
+ * ## Why the positive case asserts three things and not one
+ *
+ * "The read completed" passes trivially if no transaction was ever open — the
+ * test would then guard nothing while appearing to guard the whole hazard. So
+ * inside the callback all three of PR #7840's own false-negative checks are
+ * asserted: the driver counts one open transaction, the ambient store carries a
+ * handle, and — the load-bearing one — the `sys_metadata` read is observed
+ * ARRIVING AT THE DRIVER with that handle in its options. Only the third
+ * distinguishes "the ambient store was threaded onto the loader's read" from
+ * "the read happened to succeed on its own".
+ *
+ * Worked precedent: `sql-driver-sqlite-tx-guard.test.ts`, which keeps the
+ * driver's own `parentTrx` contract honest the same way. This lives in
+ * `packages/runtime` because it is the only package carrying `objectql` +
+ * `driver-sql` + `metadata` in one dependency closure.
+ */
+
+import { describe, it, expect, afterEach, vi } from 'vitest';
+import { ObjectQL } from '@objectstack/objectql';
+import { SqlDriver } from '@objectstack/driver-sql';
+import { DatabaseLoader } from '@objectstack/metadata';
+import { SysMetadataObject } from '@objectstack/metadata-core';
+
+/**
+ * The fixture's acquire bound, replacing knex's 60s default.
+ *
+ * Long enough that a loaded CI box cannot mistake scheduling jitter for a
+ * timeout, short enough that the negative case costs well under a second.
+ */
+const ACQUIRE_MS = 400;
+
+/** Upper bound for the negative case — an order of magnitude under knex's 60s default. */
+const STALL_CEILING_MS = 20_000;
+
+/** The seeded row `list('object')` must come back with. */
+const SEEDED_NAME = 'thing';
+
+interface Fixture {
+  engine: ObjectQL;
+  driver: SqlDriver;
+  loader: DatabaseLoader;
+  /** Every `driver.find` the fixture has observed, and whether it carried a transaction. */
+  reads: () => Array<{ object: string; hasTx: boolean }>;
+}
+
+let fixture: Fixture | null = null;
+
+afterEach(async () => {
+  vi.restoreAllMocks();
+  try {
+    await fixture?.engine.destroy();
+  } catch {
+    /* teardown is best-effort — a stalled pool must not mask the assertion */
+  }
+  fixture = null;
+});
+
+async function boot(): Promise<Fixture> {
+  const driver = new SqlDriver({
+    client: 'better-sqlite3',
+    connection: { filename: ':memory:' },
+    useNullAsDefault: true,
+    // See the header: shortens the wait, not the shape.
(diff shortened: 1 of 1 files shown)


hotlong:
**PM ruling — `domain:cli` seat (#6024). Option A stands: `skip-changeset`, no changeset. My brief was wrong on this point.**

The dispatch's definition of done said *"a changeset in `.changeset/` — test-only, no behaviour change."* That was a template line applied without checking, and this card had already been ruled: the triage comment says `skip-changeset` applies, and `pr-automation.yml`'s own failure text names the label as **preferred** for a tests-only PR.

The substantive reason is the one the report gives: `@objectstack/runtime` publishes `files: [dist]`, and tests never reach `dist` — so a changeset here declares a patch release whose shipped diff is **empty**, and puts a phantom entry in the next release notes. That is worse than a missing changeset, and the repo has closed the empty-frontmatter route (#5471) precisely so this is not a judgement call each time.

Surfacing it instead of silently picking a side is the right handling — a brief is an instruction, not a licence to contradict a standing rule.

## The card's actual content was delivered, and the harder half was not dropped

The ruling was: try the shortened `acquireConnectionTimeout`, then pool-state assertion, and **if neither is clean, pin only the positive direction and say so**. Approach 1 worked, so **both directions are pinned** — `443ms` instead of `60_085ms`, and the report is precise about *why* that is legitimate: it shortens the **wait**, not the **shape**. Same acquire failure, same knex message.

`assertEffectiveTimeout` re-reads `pool.max === 1` and `acquireConnectionTimeout === 400` off the **live** knex client, so a pool-config change or a driver that stops forwarding the knob fails **with the reason named** rather than as a silent green or a 60-second hang. That is the tripwire this card exists to install, pointed at the exact drift it was worried about.

## The false-negative guards, and one correction I want on the record

All three required guards are on the ambient case, plus a fourth I did not ask for — that `list()` returned the **seeded row**, so an empty degradation cannot satisfy "it completed". And the bare case got its own set, including `txStore.getStore()` being `undefined` and the read arriving with `hasTx: false` — because *"a rejection alone cannot tell 'nothing threaded the transaction' from 'the read never went out'."*

Then the reverse verification was **reported precisely rather than to template**: deleting the ambient fallback turns AMBIENT red **as the stall**, not at the `hasTx` assertion — because with no handle threaded the read never returns to reach that line. `hasTx` discriminates in the *other* failure mode, where a read completes without the ambient handle.

Writing down which assertion actually fired, when it differs from the one the brief anticipated, is worth more than a green tick. It is also how a reader later avoids concluding the `hasTx` pin is dead weight.

`@objectstack/runtime` measured at **227**, unchanged — and the first draft's 228 was fixed at the source (`registerObject` requires a `packageId`; the fixture now uses the engine's own proxy) rather than ledgered.

Flipping ready and enabling auto-merge once the CI conclusions land.

---

Issue #8083, opened 2026-08-12
The two MCP transports run their data verbs through different seams — stdio skips the ADR-0049 `apiEnabled` / `apiMethods` exposure gate that HTTP applies


hotlong:
Found while implementing #8034 (registering the object tools on the stdio MCP transport). Filed rather than fixed there: #8034's scope is "the tools are not registered at all", and closing this properly is a design change in a second package.

## What differs

Both MCP transports now register the SAME tools from the same `McpDataBridge` interface (#8034 makes that structural). What each tool then *executes* still differs, because the two hosts implement the bridge over different seams:

| | HTTP (`/api/v1/mcp`) | stdio (`OS_MCP_STDIO_ENABLED`) |
|---|---|---|
| bridge builder | `buildMcpBridge` — `packages/runtime/src/domains/mcp.ts` | `createStdioDataBridge` — `packages/mcp/src/stdio-data-bridge.ts` |
| data path | `callData` → `protocol` service, falling back to the ObjectQL engine | the ObjectQL engine only |
| ADR-0049 object exposure gate | **applied** (`checkApiExposure` at the top of `callData`) | **not applied** |
| protocol-layer ingress `readonly` strip, existence probes, spec-shaped receipts, `expand` / `select` | applied when the `protocol` service is registered | not applied |

`callData` cannot simply be reused by the stdio host: its signature is bound to `HttpProtocolContext` (the request, its resolved kernel, its per-environment data driver), and a long-lived stdio session has no request — it has one identity resolved from `OS_MCP_STDIO_API_KEY`.

## Why it matters

An author who declares `enable.apiEnabled: false` (or narrows `apiMethods`) on an object is telling the platform not to expose that object's data operations over the API. That declaration is honoured on the MCP HTTP surface and ignored on the MCP stdio surface — same product, same tool names, same key, different answer.

**This is not an authorization bypass, and should not be triaged as one.** `packages/runtime/src/api-exposure.ts` documents the gate as a SURFACE-AREA control rather than the authorization boundary, and its own fail-open decision rests on that: every call still passes the ObjectQL security middleware (CRUD / FLS / RLS) regardless of the outcome. On stdio those middlewares run too — the bridge calls the engine with the key's `ExecutionContext` on every verb. What leaks is the author's *exposure declaration*, not the data guard.

[...]

Issue #8032, opened 2026-08-12
[finding] `EngineAggregateOptions.groupBy` is declared `string[]` but the engine also reads `{ field, dateGranularity }` buckets — every correct caller must lie to the type


huangyiirene:
Observation-class finding, filed by the `domain:engine-core` PM seat (#6019, `session_01VGAePF7iGGUYUT8oX1cVgx`) after a **second** independent witness. Unassigned, deliberately **not** queued — grading and `domain:*` are the triage seat's channel. #4918 family.

## The fact

`EngineAggregateOptions.groupBy` is declared as `z.array(z.string())`. The engine does not read it that way. Two places in `packages/objectql/src/engine.ts` handle a **structured bucket** — `{ field, dateGranularity }` — on the same key:

- `rejectCredentialAggregation`'s field-name walk reads *"a string, or a `{ field }` bucket object"* in its own docstring and handles both spellings;
- the date-bucketing path reads `{ field, dateGranularity }`.

⇒ The declared type describes **one** of the two shapes the engine actually accepts. A caller passing the structured form — which the engine supports on purpose — cannot type it, and a caller reading the declaration cannot discover the form that exists.

## Two independent witnesses, which is why it is now a card rather than a note

1. **#7922 / PR #7984** (merged, `fce8e49`). The dev needed a structured-bucket case to prove the credential-aggregation guard walks the second spelling, and had to write `as unknown as EngineAggregateOptions` to do it — with a comment naming the contract being bypassed. ⚠️ Note what the honest spelling cost: `as any` would have grown the #4918 `check:query-options-erasure` ratchet, so the type gap pushes authors toward the form the repo's own gate penalises.
2. **The engine itself.** Its guard reads `query.groupBy as unknown[]` before walking. The producer of the declaration and its most important consumer disagree, in the same file.

A single test needing a cast is a test smell. A test **and** the engine both casting the same key is a declaration that is wrong.

## Why it matters beyond tidiness

- **The cast is load-bearing in a security-adjacent test.** #7984's structured-bucket case exists to prove a credential column cannot be smuggled through the second spelling. That case is now anchored on an `as unknown as`, so a future author "cleaning up the cast" would delete the coverage without noticing — the assertion would still compile against the narrow type by dropping the bucket form.
[...]

PR #8084, opened 2026-08-12
fix(mcp): register the object tools on the stdio transport instead of only advertising them

### LINKED ISSUE #8034: mcp-stdio-fail-closed c3 (follow-on): the stdio transport now answers, but registers ZERO tools — advertises `tools` capability, `tools/list` → -32601

hotlong:
Fixes #8034

The stdio MCP server advertised `capabilities.tools` in its `initialize` result and answered `-32601 Method not found` to every `tools/list` and `tools/call`. Third link in the stdio chain: #7645 made the transport answer, #7915 stops the banner corrupting its frames, this one is why a client that gets a clean connection still cannot do anything.

## Which of the three candidate causes it was

Neither an empty registry, nor a late registration, nor a filter that drops every tool.

**The registration call site did not exist on this transport.** `registerObjectTools` / `registerActionTools` were imported once and called from exactly one place — inside `handleHttpRequest()`, on the throwaway per-request server. Nothing reached them for stdio. The long-lived server's entire tool surface was `bridgeTools(aiService.toolRegistry)`, the AI service's function-calling registry — a *different* surface, which is empty on any app that registers no AI tools (the showcase registers none). So the nearest of the three is "empty registry", but the registry that was empty is not the one serving the 11 tools over HTTP; that one was never consulted.

Reading the boot log the card quotes confirms the shape: `Bridged 4 resource endpoints`, `Agent prompts bridged`, `Bridged 0 skill prompts`, and no tool-bridging line at all, because no code path existed to print one.

The `-32601` beside an advertised capability follows mechanically: `McpServer.registerTool` is what installs the `tools/list` and `tools/call` handlers, while the capability object was hand-declared at construction. Registering nothing left the advertisement standing over no handler.

## What changed

**1. One registration path, both transports.** `wireBridgeTools(server, bridge, options)` in `mcp-http-tools.ts` is now the single composition — object CRUD always, the action pair when the bridge carries that seam. `handleHttpRequest()` calls it; the new `MCPServerRuntime.bridgeDataTools()` calls it for the long-lived server. A tool added in that module now reaches both transports by construction.

**2. A principal-bound bridge for stdio.** `createStdioDataBridge` (`packages/mcp/src/stdio-data-bridge.ts`) implements `McpDataBridge` over the ObjectQL engine, with the `OS_MCP_STDIO_API_KEY` identity re-resolved on every call so a revoked key stops working on the next tool call — the same ADR-0101 D1 property the existing record resource has, now covering the tool surface. Permissions, RLS and FLS are the engine's middleware chain, so a tool call is bounded exactly like the same identity over REST.

**3. Capabilities are derived, not declared.** The long-lived server no longer hand-writes `tools` / `resources` / `prompts` at construction; the SDK declares each when something is actually registered. That is what makes the advertised set and the served set agree *structurally* rather than as two literals that can drift — there is now no way to advertise a primitive without also installing its handlers, because the SDK does both in one call. `logging` stays hand-declared because it is honest: there is no `registerLogging`, and the declaration is itself what wires `logging/setLevel`. `prompts` on the per-request HTTP server also stays declared, because `registerSkillPrompts` installs low-level handlers and the SDK refuses a handler whose capability was not declared first.

A host with no principal to bind, or no metadata service, now registers no tools and advertises no tool capability — with a `warn` naming the remedy, instead of a silent empty surface.

## Pins

New file `packages/mcp/src/mcp-stdio-tools.test.ts`, 11 cases, every one driving a real `StdioServerTransport` over `PassThrough` pipes and speaking newline-delimited JSON-RPC — the wire a desktop MCP host uses, and the card's own repro. The 17 pins that stayed green through the outage exercised `handleHttpRequest` and `bridgeTools` separately, and neither can see one transport serving a different surface from the other.

- `tools/list` over stdio returns the tool **names** (asserted as a set, not "no error").
- **Capability ↔ served surface agreement**, read off one live connection in both directions: the advertised set comes from `initialize`, the served set from whether each list method answers, and the assertion is that a primitive is advertised **if and only if** its method answers. Red on an advertisement without a handler *and* on a handler without an advertisement.
- **Transport parity**: the same bridge yields the same tool names on stdio and over HTTP, including when the bridge grows the optional `aggregate` seam. Divergence, not absence, was the bug.
- **Invocable end to end**: `tools/call query_records` over stdio reaches the bridge and returns its rows; the `sys_*` fail-closed guard still refuses on this transport without consulting the bridge.
- **Plugin composition**: `MCPServerPlugin.start()` on the stdio path registers the tools on the long-lived server and a stdio `tools/call` reaches `ql.find` with `context.userId` resolved from the key. Only the transport attach is stubbed (`start()` would claim the test process's real stdin/stdout); the bridge construction and registration run for real.

## Reverse verification

Predicted direction: red. Both halves of the defect were restored — the hand-declared capability block, and `bridgeDataTools` emptied of its registration limb while keeping its signature so the pins report a runtime verdict rather than a type error. All 11 went red, reproducing the reported symptom exactly:

```
× answers tools/list with the object tool NAMES, not -32601
    AssertionError: tools/list answered an error over stdio
      (#8034 was {"code":-32601,"message":"Method not found"})
× advertises exactly the primitives it serves — bridged server
    AssertionError: capabilities.tools advertised=true but tools/list served=false
× registers the tools on the long-lived server and runs one under the key identity
    AssertionError: expected { code: -32601, ... } to be undefined
```

The second line is the class this closes: advertised true, served false, on a server that had no idea it was lying. Restored from the commit afterwards and confirmed byte-identical (`git diff --stat HEAD` empty), then green again.

## Verification

- `pnpm --filter @objectstack/mcp test` — 12 files, 136 passed (125 pre-existing + 11 new).
- `pnpm --filter @objectstack/mcp typecheck` — clean.
- `check:type-check-debt` does **not** rise. This package's tsconfig excludes `*.test.ts`, so the package `typecheck` script does not see test files; measured separately with tests included, the package sits at **53 raw errors, all pre-existing, zero in any file this PR adds or edits** — recorded ledger is 63. `check:type-check-coverage` and its self-test pass, and no tsconfig exclusion was added.
- `node scripts/check-nul-bytes.mjs` OK, plus a self-scan of the changed files for the wider control-byte range.
- `eslint` clean on the changed files.
- Consumer direction checked: `registerObjectTools` / `registerActionTools` are exported from the package index but imported by nothing outside `packages/mcp` (the only external hit is prose in a dogfood matrix), and the `void` → `string[]` return change is source-compatible for every caller.

## Deliberately not in this PR

- **#8083** (filed): the two transports still run their data verbs through different seams — the HTTP bridge goes through `callData`, which prefers the `protocol` service and applies the ADR-0049 `apiEnabled` / `apiMethods` exposure gate; the stdio bridge is engine-only and applies neither. Not an authorization bypass — that gate is a surface-area control by its own ADR note and the engine's CRUD/FLS/RLS run on both paths — but the two should not differ, and unifying them is a design change in a second package. `callData` cannot be reused as-is: its signature is bound to `HttpProtocolContext`, and a long-lived stdio session has no request. The divergence is recorded in the new module's docblock so the next reader is pointed at the issue rather than re-deriving it.
- **#8032** (commented, not duplicated): the typed seam in the new bridge is the first caller forced to cast around `EngineAggregateOptions`; the same call needs a second cast for `aggregations` (`function: string` versus the closed enum), which that issue did not yet name.
- The **banner noise** #7915 is fixing. These pins attach the transport to their own pipes rather than booting `os serve`, so they never see it; the frame reader skips unparseable stdout lines rather than failing on them, which is deliberate — a strict parse here would turn another card's defect into a red build for something this file says nothing about.

---

### DIFF
--- a/.changeset/stdio-mcp-registers-tools.md
+++ b/.changeset/stdio-mcp-registers-tools.md
@@ -1,0 +1,11 @@
+---
+'@objectstack/mcp': patch
+---
+
+Serve the object tools over the stdio MCP transport instead of only advertising them
+
+The stdio MCP server advertised `capabilities.tools` in its `initialize` result and then answered `-32601 Method not found` to every `tools/list` and `tools/call`, so an MCP client that connected successfully could not query or mutate a single object. The same process answered the same requests correctly over HTTP (`POST /api/v1/mcp`), which is what made the cause visible: `registerObjectTools` / `registerActionTools` were reachable only from `handleHttpRequest()`'s throwaway per-request server, and the long-lived server behind stdio received only the AI service's function-calling `ToolRegistry` — a different surface, empty on any app that registers no AI tools.
+
+Both transports now register through one composition (`wireBridgeTools`), and the stdio host builds a principal-bound data bridge from the `OS_MCP_STDIO_API_KEY` identity, re-resolved per call so a revoked key stops working on the next tool call (ADR-0101 D1). Permissions, RLS and FLS apply exactly as they do to the same identity over REST.
+
+The `tools`, `resources` and `prompts` capabilities are no longer hand-declared at construction: the MCP SDK declares each one when something is actually registered, so what a server advertises and what it serves can no longer disagree (ADR-0076 D12). A deployment with no principal to bind — or no metadata service — now advertises no tool capability instead of advertising an empty one, and says so in the boot log.
--- a/packages/mcp/src/mcp-http-tools.ts
+++ b/packages/mcp/src/mcp-http-tools.ts
@@ -1,16 +1,28 @@
 // Copyright (c) 2025 ObjectStack. Licensed under the Apache-2.0 license.
 
 /**
- * mcp-http-tools — object CRUD exposed as MCP tools for the HTTP transport.
+ * mcp-http-tools — object CRUD exposed as MCP tools, for EVERY transport.
  *
- * These are the tools an external agent (Claude Desktop / Cursor) drives over
- * the network. Unlike the stdio bridge — which is a trusted local process —
- * the HTTP surface is reached by arbitrary callers, so every operation MUST
- * run under the caller's resolved principal. We never touch the data engine
- * directly here: all reads/writes go through an injected {@link McpDataBridge}
- * that the runtime wires to the SAME permission/RLS-enforcing path the REST
- * API uses (`callData` with the request's ExecutionContext). This module owns
+ * These are the tools an external agent (Claude Desktop / Cursor) drives, over
+ * the network or down a local pipe. Every operation MUST run under the caller's
+ * resolved principal: we never touch the data engine directly here, all
+ * reads/writes go through an injected {@link McpDataBridge} that the host wires
+ * to the SAME permission/RLS-enforcing path the REST API uses. This module owns
  * the tool *shape*; the bridge owns *execution + security*.
+ *
+ * [#8034] The file name says `http` for historical reasons only, and believing
+ * it cost this package a transport. Until #8034 {@link registerObjectTools} and
+ * {@link registerActionTools} were called from exactly one place —
+ * `MCPServerRuntime.handleHttpRequest()`, on the throwaway per-request server —
+ * so the LONG-LIVED server behind the stdio transport reached `tools/list` with
+ * an empty registry and answered `-32601` while its `initialize` result
+ * advertised `capabilities.tools`. {@link wireBridgeTools} is now the one
+ * composition both transports call, so a tool added here reaches both by
+ * construction and neither can silently serve a different set.
+ *
+ * The bridge, not the transport, is what varies: the HTTP host binds it to the
+ * request's ExecutionContext, the stdio host to the `OS_MCP_STDIO_API_KEY`
+ * identity (re-resolved per call, ADR-0101). Both hand the same interface here.
  *
  * SECURITY (zero-tolerance):
  *  - System objects (`sys_*`) are NOT exposed by default — fail-closed guard on
@@ -211,15 +223,53 @@ const VALIDATE_SITE_MAP: Record<string, { role: FieldRole; scope: 'record' | 'fl
 };
 
 /**
- * Register the object-CRUD tool set on a fresh per-request {@link McpServer}.
- * All execution is delegated to `bridge`, which is bound to the caller's
- * principal by the runtime.
+ * Wire the FULL tool surface a bridge can serve onto one {@link McpServer} —
+ * the single composition both transports call (#8034).
+ *
+ * Object CRUD always; the business-action pair only when the bridge implements
+ * `listActions` + `runAction` (graceful degradation — a host with no action
+ * mechanism keeps serving object tools unchanged). Whoever owns the server
+ * decides nothing else: the tool set is a function of the BRIDGE, so the same
+ * bridge yields the same tools on stdio and over HTTP, which is the property
+ * `transport-parity` pins.
+ *
+ * @returns the names actually registered, so a host can report its surface
+ *   honestly instead of asserting a count that can drift from the code above.
+ */
+export function wireBridgeTools(
+  server: McpServer,
+  bridge: McpDataBridge & Partial<McpActionBridge>,
+  options: RegisterObjectToolsOptions & RegisterActionToolsOptions = {},
+): string[] {
+  const registered = registerObjectTools(server, bridge, options);
+  if (typeof bridge.listActions === 'function' && typeof bridge.runAction === 'function') {
+    registered.push(...registerActionTools(server, bridge as McpActionBridge, options));
+  }
+  return registered;
+}
+
+/**
+ * Register the object-CRUD tool set on an {@link McpServer} — the throwaway
+ * per-request one on HTTP, the long-lived one behind stdio. All execution is
+ * delegated to `bridge`, which the host binds to the caller's principal.
+ *
+ * @returns the names registered on this call (the set varies with
+ *   `grantedScopes` and with whether the bridge implements `aggregate`).
  */
 export function registerObjectTools(
   server: McpServer,
   bridge: McpDataBridge,
   options: RegisterObjectToolsOptions = {},
-): void {
+): string[] {
+  // Recorded AT the registration site (`note('…')` below) rather than as a
+  // second list here: a literal list would be a parallel spelling of the same
+  // fact, and the first tool added without updating it would make every
+  // caller's report of this surface wrong while every test stayed green.
+  const registered: string[] = [];
+  const note = (name: string): string => {
+    registered.push(name);
+    return name;
+  };
   const allowSystem = options.allowSystemObjects === true;
   const maxLimit = options.maxQueryLimit ?? DEFAULT_MAX_LIMIT;
   // OAuth tool-family gating (#2698). undefined = not scope-limited.
@@ -240,7 +290,7 @@ export function registerObjectTools(
 
   if (canRead) {
     server.registerTool(
-      'list_objects',
+      note('list_objects'),
       {
         description:
           'List the data objects (tables) available in this app. Returns each object\'s name, label and field count.',
@@ -259,7 +309,7 @@ export function registerObjectTools(
     );
 
     server.registerTool(
-      'describe_object',
+      note('describe_object'),
       {
         description:
           'Get the schema of a data object: its fields (name, type, label, required) and enabled features.',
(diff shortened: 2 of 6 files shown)


hotlong:

### The root cause was none of the three the dispatch named, and the PR says so

My brief handed you three candidates — empty registry, late registration, a filter dropping every tool. The answer was **the registration call site did not exist on this transport**: `registerObjectTools` / `registerActionTools` were called from exactly one place, inside `handleHttpRequest()`, on the throwaway per-request server. Nothing reached them for stdio, and the long-lived server's whole tool surface was `bridgeTools(aiService.toolRegistry)` — a *different* registry, empty on any app registering no AI tools.

The PR does not quietly substitute the real answer for mine; it names which of my three was nearest and then says why "empty registry" is still the wrong description — the registry that was empty is not the one serving the 11 tools over HTTP, and that one was never consulted. The boot log corroborates it structurally: no tool-bridging line at all, because no code path existed to print one. That is the honest self-report ADR-0076 D12 asks for, and it is more useful than a correct fix would have been on its own. **Treat a dispatch's named candidates as leads to falsify, never as a menu to pick from** — this is the second time today that rule paid.

### Deriving the capabilities is the fix; registering the tools is only the symptom

Registering the object tools on the long-lived server closes the reported bug. Making the SDK *declare each capability at registration* instead of hand-writing the `capabilities` literal at construction is what stops the class:

> there is now no way to advertise a primitive without also installing its handlers, because the SDK does both in one call.

That converts "advertised set == served set" from an agreement between two literals — which is exactly what drifted — into a structural property. And the two exceptions are reasoned rather than blanket-applied: `logging` stays hand-declared because the declaration *is* the wiring (there is no `registerLogging`), and `prompts` on the per-request HTTP server stays declared because the SDK refuses a low-level handler whose capability was not declared first. Naming why each exception is honest is the difference between a rule and a carve-out.

The pins match. `tools/list` is asserted on the tool **names as a set**, not on "no error"; the capability↔surface pin is a **biconditional** read off one live connection, red on an advertisement without a handler *and* on a handler without an advertisement; and the transport-parity pin targets the actual bug, which was divergence rather than absence. All 11 driven over a real `StdioServerTransport` on `PassThrough` pipes — the card's own repro wire, not a unit-level stand-in. The 17 pre-existing pins stayed green throughout the outage precisely because they exercised `handleHttpRequest` and `bridgeTools` separately and structurally could not see one transport serving a different surface from the other; the new file is placed exactly where that blind spot was.

Reverse verification restored **both** halves, got all 11 red reproducing the reported `-32601`, and confirmed the restore byte-identical (`git diff --stat HEAD` empty) before going green again. Ablating only one half would have proven less than it appeared to.

### One correction to the PR's own framing of #8083

The PR says the stdio/HTTP seam divergence is "not an authorization bypass," and on the narrow question that is right — CRUD permissions, FLS and RLS are the engine's middleware chain and run on both paths, so no caller reaches a row or field they could not reach over REST. I am accepting that.

But state the residue precisely, because the PR rounds it down: the HTTP bridge goes through `callData` and therefore applies the ADR-0049 `apiEnabled` / `apiMethods` exposure gate, and the stdio bridge applies **neither**. So a key-holding principal can reach, over stdio, an object its author explicitly marked off the API surface. That is a surface-area leak, not a data leak — the distinction is real and it is why this lands rather than blocks — but "not an authorization bypass" is not the same sentence as "no new exposure," and the record should not read as though it were.

It lands because: the authorization belts do run on both paths; the surface is principal-bound with `OS_MCP_STDIO_API_KEY` re-resolved per call, so a revoked key stops working on the next call; the alternative is a transport that answers `-32601` to everything; and the blocker to unifying is architectural, not effort — `callData`'s signature is bound to `HttpProtocolContext` and a long-lived stdio session structurally has no request.

⛔ **#8083 must resolve as enforce-or-remove, not as a third posture.** Either the exposure gate applies on both doors, or ADR-0049 is amended to say in terms that it is an HTTP-surface control that deliberately does not bind other transports. A gate that runs on one door and silently not on the other is the exact shape this lane has found five times this shift, and it is worse than no gate because the next reader consumes its presence as evidence. The docblock pointer in the new module is the right interim move.

Not raised: `check:type-check-debt` — 53 raw against a recorded 63, measured *with tests included* since this package's tsconfig excludes `*.test.ts`, and zero in any file this PR adds or edits. That is the measurement the ratchet actually cares about, and reporting it the package-`typecheck` way would have been a false negative.

All 25 checks green. **Flipping ready and enabling auto-merge.**

---

Issue #5493, opened 2026-08-05
On objects that carry sharing rules, sharing middleware answers FORBIDDEN before RLS update-wideners are consulted — the identical widener works on rule-less objects


yinlianghui:
Part of objectstack-ai/hotcrm#706

Found during HotCRM's 17.0 GA acceptance sweep on `@objectstack/*` **17.0.0-rc.2**. This is a **regression against runtime-verified behavior**: HotCRM #547 added two RLS update-wideners and its PR body records live verification (member insert as marketing_user → 201).

## Measured on rc.2 (marketing_user, own token)

| Widener | Object | rc.2 behavior |
|---|---|---|
| `marketing_campaign_member_updates` | `crm_campaign_member` (no sharing rules of its own) | ✅ works — PATCH admin-owned member → 200 (control: sales_rep → 403) |
| `marketing_campaign_updates` | `crm_campaign` (carries sharing rules) | ❌ PATCH not-own campaign → **403 "FORBIDDEN: insufficient privileges to update crm_campaign"** |

**The error shape is the tell**: the failing denial is the **sharing-middleware FORBIDDEN**, not the row-gate's "[Security] … (row-level security)" shape every other write denial in the sweep produced. The failing layer answers **before** RLS evaluation on objects that carry sharing rules — the widener is never consulted on `crm_campaign`, while on rule-less `crm_campaign_member` it is.

## Reproduction

```bash
# as marketing_user
PATCH /api/v1/data/crm_campaign/ {"description":"x"}        → 403 FORBIDDEN (middleware shape)
PATCH /api/v1/data/crm_campaign_member/ {"status":"responded"}  → 200
```

## Impact

Marketing can move members but cannot edit campaigns they didn't create — the exact posture the wideners were written to fix, failing subtly (half the mechanism works).

App declarations unchanged since they were verified working; attribution is the ordering of sharing middleware vs RLS widener evaluation on objects with sharing rules.

Evidence: HotCRM acceptance run `w4/results.jsonl` (M5). Related sibling: objectstack-ai/hotcrm#705 (write gate ignores share access levels — a different layer, same subsystem).

Blocked-by: #6735
[...]

Issue #3539, opened 2026-07-27
P0: collectRLSPolicies silently strips authored org-scoped RLS policies when org-scoping is inactive (ADR-0105 F1 → D3)


os-zhuang:
## Problem

`collectRLSPolicies` (`packages/plugins/plugin-security/src/security-plugin.ts` ~L2785-2793) drops **any** RLS policy whose `using` expression contains the substring `current_user.organization_id` whenever org scoping is inactive (`!this.orgScopingEnabled`):

```ts
if (
  !this.orgScopingEnabled &&
  policy.using &&
  policy.using.includes('current_user.organization_id')
) {
  continue;
}
```

The comment states the intent: strip the platform's **own wildcard tenant policies** in single-tenant deployments. But the substring match also swallows **app-authored** policies that legitimately reference `current_user.organization_id`. A declared, authored security policy is silently unenforced — exactly the ADR-0049 ("no unenforced security properties") defect class.

Verified on `main` (2026-07-27). This is finding **F1** of [ADR-0105](https://github.com/objectstack-ai/objectstack/blob/main/docs/adr/0105-group-tenancy-posture-and-first-class-org-scope.md); the ADR marks it P0, shipping in the open edition **independent of the rest of the ADR** (decision D3).

## Fix direction (ADR-0105 D3)

- Stop substring-matching. Tag the platform's own wildcard `tenant_isolation` policies with a provenance flag (or match on `object === '*'` ∧ shipped-by-default), and strip **only tagged** policies when isolation is inactive.
- An authored policy referencing an unavailable context variable follows the existing availability path: fails closed at resolution, surfaced by the `isSupportedRlsExpression` authoring lint. It is never silently dropped.

## Acceptance

- [ ] Authored policy with `current_user.organization_id` in `using` survives collection in a deployment without org scoping (fails closed at resolution, with lint surfacing — not dropped).
- [ ] Platform wildcard tenant policies are still stripped when isolation is inactive (no behavior change for the intended case).
- [ ] Conformance-matrix row: authored-policy-survival (per ADR-0105 Phase 0).
- [ ] No visibility delta in the `authz-matrix-gate` snapshot beyond the intended correction.

Refs: ADR-0105 (F1/D3), ADR-0049.

Issue #2462, opened 2026-06-29
ADR-0076: remaining objectql/protocol layering work (cross-repo window)


os-zhuang:
Tracking issue for the **remaining** work from [ADR-0076](https://github.com/objectstack-ai/framework/blob/main/docs/adr/0076-objectql-core-tiering.md) (objectql layering / protocol segmentation / honest capabilities).

## Done (for context)
- ✅ protocol relocated out of the engine → `@objectstack/metadata-protocol`; lean `@objectstack/objectql/core` entry + boundary ratchet (#2415).
- ✅ **D9 step-1**: `ObjectStackProtocol` segmented into per-domain interfaces (`DataProtocol`/`MetadataProtocol`/…), composed alias for back-compat — non-breaking (#2429).
- ✅ ADR D1–D12 recorded (rev.9).

## Remaining — all gated to a coordinated **cross-repo window** (framework + cloud + ../objectui), unless noted

- [ ] **D10 — distribute protocol implementations to domain packages.** Data facade → transport; analytics/feed/realtime impls owned by their existing services (`service-analytics`/`service-messaging`/`service-realtime`); the metadata-management impl **stays in `@objectstack/metadata-protocol`** (name retained — OQ#7). Dissolve the central facade. *Risk: med; cross-repo.*
- [ ] **D11 — decompose the transport god-implementation.** `runtime/http-dispatcher.ts` (~3.8k) + `rest/rest-server.ts` (~5.1k) → thin framework-agnostic port + per-domain **normalized** handlers (NOT framework-specific routes — keep multi-adapter). Confirm dispatcher↔rest-server overlap; **validate multi-adapter by writing a 2nd `IHttpServer` adapter** (thin Workers/Express) to prove the port is free of Hono-isms. *Risk: high; cross-repo.*
- [ ] **D12 — honest capabilities (fixes agents being misled by fake services).** Standardize a stub/dev/fallback marker; make `http-dispatcher` `svcAvailable` respect it (report `status:'stub'`/`handlerReady:false`, never `available`); add the `stub` status to the discovery schema; update consumers (agents, console) to trust only real services. Mark the analytics fallback + plugin-dev stubs honestly — **do not delete** (no `/analytics` 404). *Risk: med; cross-repo (console reads `discovery.services`).* See OQ#11 (marker convention).
[...]

PR #8099, opened 2026-08-12
fix(plugin-security): a check-only write policy no longer disables both write-side row gates (#8059)

### LINKED ISSUE #8059: [security] A check-only `update` policy suppresses the #7665 write-scope derivation — a contributor by-id-writes showcase_invoice rows it cannot read

hotlong:
Fixes #8059

## The exposure

On the shipped showcase, on stock `main`, a persona holding the plain `contributor` position:

- `GET /data/showcase_invoice/:id` → **404** — `invoice_own_rows` correctly hides an invoice it does not own, and
- `PATCH /data/showcase_invoice/:id` → **200, and the row actually changes** (re-read as admin).

Same on `showcase_invoice_line`, which derives access from that invoice via `controlled_by_parent`. **A caller who cannot read a record can write it by id** — including reassigning its owner to itself. The #1994 class, reopened for this authoring shape one issue after #7665 closed it for the select-only shape.

Not showcase-specific: any app authoring a `using` narrowing for a position **plus any update-class policy — even one that only validates the post-image** — loses by-id write scoping on that object. The showcase is the shipped example of the recommended authoring style.

## Both sites move, because one authoring shape disabled both belts

The showcase's `invoice_owner_immutable` is `operation: 'update'` with a `check` clause and no `using`.

**Site 1 — `computeLayeredRlsFilter`, the write-visibility floor.** #7665's derivation triggered on `collected.length === 0`. A check-only policy is fully applicable (object, `positions`, operation all match), so the set was non-empty and the derivation never ran — while that same policy compiled to no row filter at all, having no `using` for the compiler to read. Layer 1 was null and every write-side row gate composed from it went back to being a no-op: the by-id pre-image gate, the `controlled_by_parent` master check, the bulk AST injection. The trigger now asks whether a write-scope **predicate** applies, which is what #7665 criterion 5 says: a `check` clause is post-image validation (ADR-0058 D4), not a scope predicate. The predicate test is a strict superset of the emptiness test, so every path #7665 pinned is reached on the same inputs as before.

**Site 2 — `computeWriteCheckFilter`, the post-image belt.** `heldPositions` was not passed, so `getApplicablePolicies` evaluated the ADR-0090 P2 applicability domain against `[]` and dropped **every policy declaring `positions` from the check, for every caller** — holder and non-holder alike. A position-scoped `check` clause was inert (ADR-0049 enforce-or-remove; the class #3539 closed for org-scoped policies). Positions are now threaded through exactly as every other `collectRLSPolicies` call site threads them, and the domain still decides: a non-holder is still outside it.

**Third site, a direct consequence of site 1 and fixed with it.** `checkAuthoredRowWrite` reached its `abstain` verdict for check-only policies only because Layer 1 happened to be null. With a scope now derived in exactly that case, the derived readable-set scope would have begun answering `admit` — which the by-id widener uses to **widen**. Its authored-set pre-check now requires a policy that actually declares a row scope, so a derived scope still cannot masquerade as an authored admission (#5493 / #7281). This **preserves** the previous verdict rather than changing it.

`policyDeclaresClause` in `rls-compiler.ts` is the single definition of "declares a clause" — the same test the compiler already applies per policy when it skips a policy carrying no predicate for the clause being compiled.

## `verify --rls` on the showcase: 2 holes → **0 holes**

The base probe persona holds no positions, so this class is unreachable for it — `verify --rls` on this branch alone reports 0 holes, and so does stock `main`. That number proves nothing, so the real signal was measured with #8067's position personas, **composed locally with this fix in one tree, toggling only this commit**:

| | `contributor` persona | all personas | exit |
|:--|:--|:--|:--|
| this commit reverted | 5 PROVEN (3 consistent, **2 HOLES**) | 35 PROVEN (33 consistent, **2 HOLES**) | 1 |
| this commit applied | 5 PROVEN (**5 consistent, 0 HOLES**) | 35 PROVEN (**35 consistent, 0 HOLES**) | 0 |

The ablated run reproduces #8067's reported before-state exactly, down to the sentence:

```
✗✗ showcase_invoice  [rls-hole] the probe cannot read it (GET 404) yet MUTATED it by id (PATCH 200)
✗✗ showcase_invoice_line  [rls-hole] the probe cannot read it (GET 404) yet MUTATED it by id (PATCH 200)
```

Both objects flip to **proven and consistent** — the proven count rises 33 → 35, so they became proofs, not skips. `✓ verify passed`. Nothing in `packages/verify`, the probe, or the showcase's policies was touched; the composition was local and never pushed.

## Tests

`packages/plugins/plugin-security/src/check-only-write-scope.test.ts` — 21 new pins on the showcase's exact invoice shape (select-only `using` + a check-only `update` policy declaring `positions`, `public_read_write` OWD, the `member_default` floor present and out of domain).

Every refusal is asserted on its **ADR-0112 envelope**, never on the bare fact of a refusal, because the two belts overlap on the obvious probe: revert site 1 and a "the PATCH was refused" assertion stays green, since the check gate refuses it instead (the caller cannot read the pre-image, so the post-image is missing `owner` and fails the check by accident). The belts have distinct envelopes — `record_access_denied` / "(row-level security)" for the row gate, `record_change_not_allowed` / "would violate a row-level CHECK" for the check gate — so a refusal migrating between them reddens.

Site 1 is additionally pinned where the check gate provably cannot stand in for it:

- **a payload that satisfies the check** — the caller patches the unreadable invoice to `owner: itself`, a record steal that belt 2 passes by design, so only the derived row scope can refuse it;
- **the bulk path**, which step 3.6 explicitly declines to post-image validate.

Every refusal is followed by an **admin re-read of the row through the full middleware chain** plus the stored row — a refusal that still wrote is the defect.

Also pinned: Layer 1 really derives (explain reports rls `narrows`, not `not_applicable`; the derived scope admits the caller's own rows rather than being a deny sentinel); `heldPositions` reaches the check filter for a holder and not for a non-holder; nothing over-blocks (in-scope update, detail update, detail insert and delete by a legitimate holder all still land); and #7665 criterion 5 is unchanged where a real update-scope `using` **is** authored — the widener still widens and its boundary still refuses.

**Independent reverse verification**, each site ablated alone from the committed state:

- **site 1 reverted, site 2 intact** — 6 failed / 15 passed. The record steal completed (`expected a refusal, got a completed update`); the ordinary PATCH was still refused but by the **wrong belt** (`'You are not allowed to save this record with the values you entered...'` instead of `'You do not have access to this record...'`); the `controlled_by_parent` line update and insert both completed; the bulk update rewrote the unreadable row; explain reported `not_applicable`. Every site-2 pin stayed green.
- **site 2 reverted, site 1 intact** — 2 failed / 19 passed, both check-gate pins, both `expected a refusal, got a completed update` (the owner reassignment landed). Every site-1 pin stayed green.

## Verification

- `pnpm --filter @objectstack/plugin-security test` — **51 files, 1022 tests, all pass** (1001 before this PR, +21 new).
- `pnpm --filter '...@objectstack/plugin-security' typecheck` — **26 packages clean**. That filter is the **prefix / downstream-consumer** direction; the dependency closure was built first (`turbo run build --filter='...@objectstack/plugin-security'`, 66 tasks).
- `pnpm check:type-check-debt` — OK, *"none above its recorded number"*. **Not raised**: plugin-security's test layer measures 11 against a recorded ceiling of 21.
- `pnpm --filter @objectstack/dogfood test` — 624 passed, 3 skipped, 0 failed.

---

### DIFF
--- a/.changeset/check-only-policy-write-scope.md
+++ b/.changeset/check-only-policy-write-scope.md
@@ -1,0 +1,60 @@
+---
+"@objectstack/plugin-security": patch
+---
+
+fix(plugin-security): a `check`-only write policy no longer disables both write-side row gates — a caller who cannot read a record could write it by id (#8059)
+
+**The exposure, measured on the shipped showcase.** A persona holding the plain
+`contributor` position got `GET /data/showcase_invoice/:id` → **404** — the app's
+`invoice_own_rows` narrowing correctly hid an invoice it does not own — and
+`PATCH /data/showcase_invoice/:id` → **200, with the row actually changed**
+(confirmed by re-reading it as an admin). Same on `showcase_invoice_line`, which
+derives its access from that invoice via `controlled_by_parent`. **A caller who
+could not read a record could write it by id**, including reassigning its owner
+to themselves. That is the #1994 class ("you cannot mutate what you cannot see"),
+still open for this authoring shape one issue after #7665 closed it for the
+select-only shape.
+
+It is not showcase-specific. Any app that authors a `using` narrowing for a
+position **plus any update-class policy — even one that only validates the
+post-image** — lost by-id write scoping on that object. The showcase is the
+shipped example of the recommended authoring style, so this is the shape apps are
+being taught to write.
+
+**One authoring shape switched off both belts, which is why both are fixed here.**
+The showcase's `invoice_owner_immutable` is `operation: 'update'` with a `check`
+clause and no `using`:
+
+- **The row gate never derived a scope.** #7665's write-visibility floor derives
+  a write scope from the caller's SELECT narrowing when no policy of the write
+  class applies, but it tested whether the applicable set was *empty*. A
+  check-only policy is fully applicable — object, `positions` and operation all
+  match — so the set was non-empty and the derivation was skipped, while the
+  policy itself compiled to no row filter at all, having no `using` for the
+  compiler to read. Layer 1 was null and every write-side row gate composed from
+  it was a no-op again: the by-id pre-image gate, the `controlled_by_parent`
+  master check, and the bulk-write AST injection. The trigger now asks whether a
+  write-scope **predicate** applies, which is what #7665's criterion says — a
+  `check` clause is post-image validation (ADR-0058 D4), not a scope predicate.
+- **The post-image check was dropped for every caller.** `computeWriteCheckFilter`
+  did not pass the caller's held positions, so the ADR-0090 P2 applicability
+  domain was evaluated against an empty list and **every policy declaring
+  `positions` was filtered out of the check** — for holders and non-holders
+  alike. A position-scoped `check` clause was inert (ADR-0049
+  enforce-or-remove), so the write was not stopped on the way out either.
+  Positions are now threaded through, and the domain still decides: a policy
+  scoped to a position a caller does not hold still does not apply to them.
+
+**Also fixed, as a direct consequence of the first site.**
+`checkAuthoredRowWrite` — the by-id widener's "does an app-authored policy admit
+this row?" probe — reached its `abstain` verdict for check-only policies only
+because Layer 1 happened to be null. With a scope now derived in exactly that
+case, the derived readable-set scope would have started answering `admit`, which
+widens. The probe now requires an authored policy that actually declares a row
+scope, so a derived scope still cannot masquerade as an authored admission
+(#5493 / #7281). This preserves the previous verdict rather than changing it.
+
+**Unaffected:** an object that authors a real update-scope `using` predicate
+keeps deciding by it alone, in both directions, and the read-side superuser
+bypass and platform ownership floor paths are untouched. In-scope writes by
+legitimate holders still land.
--- a/packages/plugins/plugin-security/src/check-only-write-scope.test.ts
+++ b/packages/plugins/plugin-security/src/check-only-write-scope.test.ts
@@ -1,0 +1,812 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+//
+// [#8059] A `check`-only UPDATE policy must not disable the write-side row
+// gates — the #1994 class ("you cannot mutate what you cannot see") reopened
+// through a different AUTHORING SHAPE, one issue after #7665 closed it for the
+// select-only shape.
+//
+// MEASURED ON STOCK `main`, on the shipped showcase: a persona holding the
+// plain `contributor` position gets `GET /data/showcase_invoice/:id` → 404 —
+// `invoice_own_rows` correctly hides an invoice it does not own — and
+// `PATCH /data/showcase_invoice/:id` → 200, WITH THE ROW ACTUALLY CHANGED
+// (re-read as admin). Same on `showcase_invoice_line`, which is
+// `controlled_by_parent` on the invoice. The one thing that separates it from
+// `showcase_task`, which is consistent on the same run, is that the invoice
+// also authors `invoice_owner_immutable`: `operation: 'update'` with `check`
+// only, no `using`.
+//
+// TWO SITES, DISABLED BY THAT ONE AUTHORING SHAPE — which is why a fix to
+// either alone is not a partial fix but a fix that does not work:
+//
+//   1. `computeLayeredRlsFilter`'s write-visibility floor triggered on
+//      `collected.length === 0`. A check-only policy is fully APPLICABLE
+//      (object + `positions` + operation all match) but carries no `using`,
+//      so the collection was non-empty, the #7665 derivation never ran, and
+//      the collected policy then compiled to no row filter at all — Layer 1
+//      null, and every write-side row gate composed from it a no-op again:
+//      the by-id pre-image gate (2.7), the controlled_by_parent master check,
+//      the bulk AST injection. Fixed by asking whether a write-scope
+//      PREDICATE applies, which is what #7665 criterion 5 says; a `check`
+//      clause is post-image validation (ADR-0058 D4), not a scope predicate.
+//
+//   2. `computeWriteCheckFilter` did not pass `heldPositions`, so
+//      `getApplicablePolicies` evaluated the ADR-0090 P2 applicability domain
+//      against `[]` and dropped EVERY policy declaring `positions` from the
+//      post-image check — for holders and non-holders alike. The second belt
+//      was not merely late, it was absent (ADR-0049 enforce-or-remove; the
+//      class #3539 closed for org-scoped policies).
+//
+// ── HOW THIS FILE PINS THE TWO SITES SEPARATELY ────────────────────────────
+// The belts overlap on the obvious probe, and that overlap is a trap: revert
+// site 1 and a naive "the PATCH was refused" assertion STAYS GREEN, because
+// the check gate refuses it instead — the caller cannot read the pre-image, so
+// the post-image is missing `owner` and fails the check by accident. Two
+// consequences, both deliberate here:
+//
+//   • every refusal is asserted on its ADR-0112 ENVELOPE, never on the bare
+//     fact of a refusal. The belts have distinct envelopes — the row gate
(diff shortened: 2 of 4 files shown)


hotlong:

Reviewed against the three red lines the dispatch made binding, all three cleared:

**1. Both sites must move together.** They do, and the PR proves *why* they had to: one authoring shape (`operation: 'update'` + `check`, no `using`) disabled both belts at once. Site 1 because a check-only policy is fully applicable yet compiles to no row filter, so `collected.length === 0` was false while the filter was empty; site 2 because `heldPositions` was never threaded, so the ADR-0090 P2 domain was evaluated against `[]` and dropped every position-scoped policy from the check for **every** caller. Landing either alone leaves a live hole — the ablation runs below measure exactly that.

**2. The refusal proven by an admin re-read showing the row unchanged.** Every refusal in the new suite is followed by an admin re-read through the full middleware chain plus the stored row. This is the assertion that separates "the API said no" from "the write did not happen," and both ablation runs failed on precisely that distinction (`expected a refusal, got a completed update`), so the re-read is load-bearing, not decoration.

**3. Each site reverse-verified independently.** Site 1 ablated alone: 6 failed / 15 passed, every site-2 pin green. Site 2 ablated alone: 2 failed / 19 passed, every site-1 pin green. Disjoint failure sets in both directions — neither fix is riding on the other's coverage.

### What raises this above the brief

The dispatch did not ask for it, and it is the single most important thing in the PR:

> revert site 1 and a "the PATCH was refused" assertion **stays green**, since the check gate refuses it instead (the caller cannot read the pre-image, so the post-image is missing `owner` and fails the check by accident).

The obvious probe for this defect is vacuous. A refusal *migrating between belts* is indistinguishable from a refusal *holding*, unless you assert the ADR-0112 envelope — and the ablation run confirms the migration is real (`record_change_not_allowed` / "would violate a row-level CHECK" where `record_access_denied` / "(row-level security)" belongs). Site 1 is then additionally pinned where belt 2 provably **cannot** stand in: the record steal (a payload that satisfies the check by design) and the bulk path (step 3.6 explicitly declines to post-image validate). That is this lane's recurring finding — *a check that cannot fail is worse than no check, because it is consumed as evidence* — found and closed by the developer without being told to look.

The showcase measurement is reported the same way. `verify --rls` on this branch alone reports 0 holes **and so does stock `main`**, and the PR says so rather than banking the 0 (ADR-0076 D12 / #2462). The signal is the local composition with #8067's position personas, both numbers taken in one tree toggling only this commit, with the proven count rising 33 → 35 — so the objects became **proofs**, not skips. Attribution is not inferred from the merge.

### Carried forward

⛔ **The `checkAuthoredRowWrite` guard is behavior-preserving and must not be deleted as cleanup.** It restores the `abstain` that a null Layer 1 used to produce incidentally; with site 1 deriving a scope in exactly that case, a *derived* readable-set scope would otherwise start answering `admit` to the by-id **widener** (#5493 / #7281). If a future card wants a derived scope to widen a by-id write, that is a deliberate contract decision — it must be reached on purpose, never by removing this guard because it "looks redundant."

Not raised: `check:type-check-debt` OK (plugin-security test layer 11 against a ceiling of 21). Pre-existing 271-error surplus across 9 under-ceiling entries left alone — #6376, out of scope here, correctly not touched.

**Flipping ready and enabling auto-merge.** `Dogfood Verify CLI` is already green on this head; the remaining gates are in flight and auto-merge will hold until they converge. Once this lands, **#8067 comes off its hold** — it was held only because `dogfood-verify` gates the merge queue for every lane, and this is the fix that turns its two reds green.

---

Issue #8124, opened 2026-08-12
`fieldsFromZodIssues` leaks Zod's issue codes onto the wire's `fields[].code` — the exact pass-through ADR-0114 D3 closed, still live in `@objectstack/runtime`


hotlong:
Found while implementing #8055 (which emits through this helper and therefore inherits the defect).

## What is declared

ADR-0114 D2 makes `FieldErrorSchema.code` a **closed catalog** — `FieldErrorCode` in `packages/spec/src/api/errors.zod.ts`, 27 lowercase snake_case members. D3 is explicit that Zod is mapped at the boundary, **never** passed through:

> **D3 — Zod is mapped at the boundary, never passed through.** `zodIssuesToFields` translates using `origin` / `format` per the table above, plus the parsed input for the `invalid_type` split. An unmapped Zod code becomes `invalid_value` (a catalog member) rather than leaking.

`@objectstack/rest` complies: `zodIssuesToFields` in `rest-server.ts` runs `zodIssueToFieldCode`, which is typed to return `FieldErrorCode`.

## What is emitted

`packages/runtime/src/validation-failure.ts` has a second, non-compliant converter — `fieldsFromZodIssues` — which assigns the Zod issue's own code verbatim:

```
code: issue.code,
```

Measured against the real `FlowSchema` (a genuine `safeParse`, not a hand-written fixture), the codes that reach the wire include `unrecognized_keys` — **not** a `FieldErrorCode` member. ADR-0114 D3's own table says that row maps to `unknown_field`. Zod's `too_small` / `too_big` / `custom` / `invalid_union` are equally outside the catalog, and `too_small` is the ambiguity D3 names explicitly (a short string, a small number and a short array share one code).

Consequence: a response whose `details.fields[]` carries one of those codes does **not** parse against the schema the protocol declares for it. A client validating the error envelope — the thing ADR-0114 D2 gave it a schema for — rejects a well-formed refusal from its own server. It is also the two-vocabularies-on-one-position problem the ADR was written to remove: the same wire slot answers `min_length` from a validator and `too_small` from a Zod-parsed route, and nothing in the body says which dialect the reader is holding.

## Affected call sites

Three, all in `@objectstack/runtime`, all through this one helper:

- `src/domains/analytics.ts` — the `AnalyticsQueryRequestSchema` entry refusal
- `src/domains/notifications.ts` — the same shape one domain over
- `src/domains/automation.ts` — the flow-registration refusal added by #8055

[...]

Issue #8123, opened 2026-08-12
`PUT /api/v1/automation/:name` answers 500 for the same four malformed flow bodies #8055 just reclassified on POST


hotlong:
Found while implementing #8055. Filed rather than fixed there: #8055's scope is the `POST /` registration route, and this is a second route.

## Symptom

`PUT /api/v1/automation/:name` (updateFlow) makes the **identical** `automationService.registerFlow(name, definition)` call the `POST /` branch makes, in the same file, with no classification around it — `packages/runtime/src/domains/automation.ts`, the `parts.length === 1 && m === 'PUT'` branch (locate by symbol; the POST branch is the one carrying the `#8055` comment).

So every body #8055 measured as a wrong-class 500 on POST is still a wrong-class 500 on PUT:

1. a node missing `label` — 500 carrying a raw Zod issue array
2. an unknown node key — 500 carrying `unrecognized_keys` / `keys` / `path`
3. a malformed `try_catch` region — 500 `try_catch 'g' try: invalid region — ...`
4. an undeclared node config key — 500 carrying the #4277 self-correcting message

Same two costs #7535 and #8055 spell out: a retry-on-5xx client re-sends a request that can never succeed, and an authoring agent reads "the server broke" instead of "your metadata is wrong".

## Why it matters beyond the repeat

After #8055 lands, **POST and PUT disagree about the class of an identical refusal** on the same domain — the exact drift #7535's fix was shaped to avoid ("the two routes cannot disagree"). An author who publishes a flow through the update route gets a different answer from one who publishes through create, for the same broken definition.

## Fix

One line, once #8055 is merged: wrap the `PUT` branch's `registerFlow` call the same way, reusing the `flowDefinitionRefusal` helper #8055 adds to that file. The helper was written module-local and route-agnostic for this. Note the PUT branch has its own `body.definition ?? body` unwrap, so the definition it forwards is not always the request body — worth a test case of its own.

## Not licensed by this card

Widening the refusal, softening any refusal into a 200, or changing which bodies are rejected. Class and envelope only, as in #8055.

## Blocked-by

Blocked-by: #8055 (the helper this reuses lands there).

---
_Generated by [Claude Code](https://claude.ai/code/session_01B3Kurx8qufrDzNjk4rag7V)_

hotlong:
[...]

Issue #7535, opened 2026-08-11
`POST /api/v1/automation/:name/toggle` answers 500 INTERNAL_ERROR instead of 404 NOT_FOUND for an unknown flow


huangyiirene:
## Symptom

`POST /api/v1/automation/:name/toggle` against a flow name that does not exist answers **`500 INTERNAL_ERROR`** where it should answer **`404 NOT_FOUND`**.

An error-class mismatch, P3. Nothing is corrupted and nothing leaks — the toggle does not happen — but a 500 tells every caller the wrong thing: clients and retry layers treat 5xx as "the server broke, try again" and 4xx as "your request was wrong, don't". A typo in a flow name currently presents as a server fault, and any client with retry-on-5xx will hammer the endpoint over a request that can never succeed.

## Root cause

Not root-caused to a line. The handler is `POST /automation/:name/toggle`, registered in `packages/runtime/src/dispatcher-plugin.ts` and dispatched to `toggleFlow` in `packages/runtime/src/domains/automation.ts`; the route is declared in `packages/runtime/src/route-ledger.ts`. The shape is the familiar one — an unknown-name lookup throws and the catch maps everything to `INTERNAL_ERROR` rather than mapping the not-found case to 404 first.

Note the same endpoint already gets its **body** validation right: `{ enabled?: boolean }` is strict, and `{"enable": false}` (one letter off) is a located 400 rather than a silent enable (#3899). The missing-flow arm just never got the same treatment.

## Reproduction

1. Boot the showcase on an isolated file DB and port.
2. `POST /api/v1/automation/definitely_not_a_flow/toggle` with body `{"enabled": false}`.
3. Observe `500 INTERNAL_ERROR`. Expected `404 NOT_FOUND` naming the unknown flow.

Re-check the handler:

```
rg -n "toggle" packages/runtime/src/domains/automation.ts packages/runtime/src/dispatcher-plugin.ts
```

## Source

Extracted from the QA run objectstack-ai/objectstack#7516 (framework a86db175). Listed there as item 4 under "Fixture / authoring issues to fix", flagged P3.

os-help:
**Claim** — PM loop `domain:cli`, session `session_0158ZQo7LiHSxGWpYKuPq1wu` (os-help seat, #6024), wave 5.

- **Branch**: `claude/issue-7535-toggle-unknown-flow-404`
[...]

Issue #3918, opened 2026-07-29
[17.0.0-rc.0] Runtime dispatcher error paths drop ValidationError.fields[] and downgrade status to 500 (rest-server maps it correctly)


yinlianghui:
Found while investigating form UX during the HotCRM 17.0.0-rc.0 upgrade (tag commit `fc156fa4a`). Minor but it forecloses per-field error display for any surface served by the runtime dispatcher.

## Facts
- `ValidationError` carries `.code='VALIDATION_FAILED'` and `.fields[]`, but no `.status`/`.statusCode`/`.issues` (`packages/objectql/src/validation/record-validator.ts:106-122`).
- `@objectstack/rest` maps it correctly → **400** with top-level `fields[]` (`packages/rest/src/rest-server.ts:134-143`).
- The runtime dispatcher's two error exits do neither:
  - `packages/runtime/src/http-dispatcher.ts:495-506` (`errorFromThrown`): status falls back to **500** (no `.status` on the error), `details` reads only `.issues` (absent) → `fields[]` lost.
  - `packages/runtime/src/dispatcher-plugin.ts:378-410` (`errorResponseBase`): same status fallback; body carries only `{message, code}`, and `:402` replaces the message with the generic INTERNAL_ERROR_MESSAGE on 5xx — so the caller gets a 500 "internal error" for a user-input validation failure.

## Ask
Teach both dispatcher exits the `VALIDATION_FAILED` shape (status 400 + `fields[]` passthrough), matching rest-server.

PR #8125, opened 2026-08-12
fix(runtime): answer 400 VALIDATION_FAILED, not 500, for a malformed flow registration (#8055)

### LINKED ISSUE #8055: `POST /api/v1/automation` answers 500 INTERNAL_ERROR for a malformed flow body — including the #4277 undeclared-config-key refusal the #7545 ruling leans on
Same error-class mismatch #7535 just fixed on the sibling `POST /automation/:name/toggle` route.

## Symptom

Four distinct malformed registration bodies, all answered **500 INTERNAL_ERROR** where 400 is the right class. Reproduced 2×:

1. A node missing `label` → 500 carrying a **raw zod issue array**: `{expected:'string', code:'invalid_type', path:['nodes',0,'label']}`
2. An unknown node key → 500 `{code:'unrecognized_keys', keys:['next'], path:['nodes',0]}`
3. A malformed `try_catch` region → 500 `"try_catch 'g' try: invalid region — Invalid input: expected object, received array"`
4. **The important one** — an undeclared node config key → 500:
   > `Flow 'x' rejected: 1 undeclared config key(s) (#4277). - node 'n' (notify): unknown config key \`totallyBogusKey\` at config.totallyBogusKey. It is not declared by this node type's configSchema…`

**Expected** 400 `VALIDATION_FAILED` with an ADR-0114 `details.fields[]`, as the sibling `/toggle` route now returns and as the descriptor routes return.

Contrast controls run in the same pass: a well-formed body still registers 200, and the unknown-flow toggle correctly answers 404 `RESOURCE_NOT_FOUND`. So the catch arm on this one handler is mapping caller-input refusals to `INTERNAL_ERROR`.

## The substance is fine — only the class and the message shape are wrong

The #7545 premise still holds: the refusal **is** hard and **is** located. Case 4 in particular produces exactly the self-correcting message #4277 designed. It is delivered under an HTTP status that says the opposite.

## Impact — the one #7535 spelled out

An authoring client with retry-on-5xx will retry a request that can never succeed. And an agent authoring a flow reads "the server broke" instead of "your metadata is wrong" — the inverse of the self-correcting shape #4277 chose. Case 1 additionally leaks a raw zod issue array where the house envelope belongs.

## Severity

P3.

## Source

Found by the platform checklist retest of `automation.flow-node-type-matrix` (framework `279ee48a`), while verifying the #7545 ruling. Held no clause of that item.

hotlong:
Fixes #8055

`POST /api/v1/automation` answered **500 `INTERNAL_ERROR`** for four distinct malformed flow definitions. It now answers **400 `VALIDATION_FAILED`** with an ADR-0114 `details.fields[]`.

Every one of those refusals was already correct — the definition really is bad, and the engine really does locate the fault. The class was the defect, and the class is what clients branch on: 5xx means "the server broke, try again", 4xx means "your request was wrong, do not". Case 4 is the sharpest: #4277 shaped that message so an authoring agent can **self-correct** — it names the key, the node, the node type, and the keys the descriptor does declare — and it arrived under a status telling the agent to retry unchanged.

## Cause

`AutomationEngine.registerFlow` raises every one of its refusals as a plain throw with no `.status`:

| # | Body | Raised by | Old wire answer |
|---|---|---|---|
| 1 | node missing `label` | `FlowSchema.parse` | 500 + `details.issues` = the **raw Zod issue array** |
| 2 | unknown node key | `FlowSchema.parse` | 500 + `{code:'unrecognized_keys', keys:['next'], path:['nodes',0]}` |
| 3 | malformed `try_catch` region | `validateControlFlow` | 500 `try_catch 'g' try: invalid region — ...` |
| 4 | undeclared node config key | `validateNodeConfigKeys` (#4277) | 500 + the self-correcting text |

The registration branch did not catch, so the throw left `handleAutomationRequest` — and `dispatch()` re-throws everything that is not a permission denial. The transport's outer catch (`dispatcher-plugin`'s `errorResponseBase`) then did what it does with a statusless throw: 500, with any `.issues` copied into `details` verbatim. That last part is why case 1 leaked Zod's internal shape onto a wire position the house envelope owns.

## Fix

`packages/runtime/src/domains/automation.ts` — the `POST /` branch wraps its `registerFlow` call and converts the engine's verdict through a new module-local `flowDefinitionRefusal`.

Four deliberate choices worth reviewing:

- **The whole call is treated as a caller-input refusal, not a recognised subset.** The alternative is to reclassify only the shapes this file can name — a `ZodError`, or an engine message matched by its prose. #7535's fix rejected exactly that ("teaching a shared catch to recognise one engine's message string would make every domain's not-found depend on that prose"), and here it is also wrong on the merits: `registerFlow` **is** the parse of a caller-supplied document, so "the definition is bad" is the honest default for a refusal it raises — not a guess about which one it raised.
- **The producer keeps an escape hatch.** An error carrying its own `.status` / `.statusCode` is passed through untouched, which is the same precedence `errorFromThrown` already applies. Nothing in the engine declares one today, so this is not a live branch — it is the seam that keeps a future engine-side "the flow store is unreachable" (a real 503) from being answered as the author's fault. Pinned by a test.
- **Caught and RETURNED, not rethrown.** The same shape the `resume` branch a few routes down already uses for engine-originated refusals, and the same shape `/meta`'s save routes use for a service that throws on caller metadata. `dispatch()` re-throws non-denials, so a transport calling it directly would otherwise get an exception where every other refusal on this domain hands back a response. It also means the long #4277 message is never handed to a 5xx path.
- **The Zod issue array is mapped, not forwarded.** The converted error carries `fields[]` and no `.issues`, so there is nothing left for `errorFromThrown` to copy.

**Unchanged**: which bodies are refused. A definition that registered before still registers (contrast control below), every one that was refused is still refused, and the engine's own message survives verbatim — a 400 never reaches the #3867 5xx sanitiser. The #3899 body checks still refuse a nameless definition **before** the engine is asked.

### On the model this was asked to follow

The card asked me to follow #7535's fix on the sibling `/toggle` route and to say so if the shape did not transfer. **It transfers as a principle and not as a mechanism**, so this is a deliberate adaptation rather than a copy:

- #7535 (PR #7558) fixed a **404 not-found** by running an existence probe **before** touching the service. Its refusal set is closed — one condition, one lookup, decided ahead of the call.
- This route's refusals are open: they come out of arbitrary Zod parse failures plus two descriptor-driven engine checks, and none of them is decidable without calling `registerFlow`. A pre-check here would mean re-implementing the engine's validation in the transport, which is the drift #7558 chose the shared probe to avoid.

What does carry over, and is the reason the two routes now converge rather than growing a third spelling, is #7558's stated rule: **which HTTP status a plain domain error means is the serving boundary's decision**, made at the domain handler and not in a generic catch. Both routes now make it there, in the house envelope, naming the thing that went wrong.

### On `fields[].code`

For a refusal with no Zod path to point at (cases 3 and 4), the entry addresses the body root — the convention `fieldsFromZodIssues` already uses — with `code: 'invalid_value'`, the ADR-0114 catalog's "rejected for a reason no other member names". Re-deriving a path by parsing the engine's sentence would be the same prose dependency rejected above.

For the Zod cases the per-issue `code` is whatever `fieldsFromZodIssues` produces, which today is Zod's own vocabulary rather than the ADR-0114 D3 catalog (`unrecognized_keys` is not a `FieldErrorCode` member). That pass-through is the **package's**, not this route's — `/analytics` and `/notifications` emit through the same helper — and the compliant mapper lives module-local inside `@objectstack/rest`'s `rest-server.ts`. Forking a local copy here would make automation a third dialect inside one package, so it is **filed as #8124** with the three routes and the three fix options, not actioned as a rider on a P3 bug fix.

## Tests

New: `packages/runtime/src/domains/automation-register-error-class.test.ts` (9 tests).

The fake service runs the **real** refusals for three of the four cases — `FlowSchema.parse` and `validateControlFlow` imported from `@objectstack/spec/automation` are the very calls `canonicalizeStoredFlow` makes, in that order — so cases 1 to 3 are produced by production code. Only #4277 is modelled, from the engine's own construction, because that check is descriptor-driven and lives in a package `@objectstack/runtime` does not depend on. The captured shapes are byte-identical to the ones the issue reports, including `try_catch 'g' try: invalid region — Invalid input: expected object, received array`.

Anti-vacuity: no test asserts `status !== 500` or `status >= 400`. Every case asserts the full envelope — `status`, `error.code`, `error.httpStatus`, and the presence and per-entry shape of `details.fields[]` — plus its own substance:

- **case 1** pins the positive body shape (`Object.keys(details)` is exactly `['fields']`), that `details.issues` is gone, that each entry has exactly `field`/`code`/`message`, that no Zod-internal **key** (`"expected"`, `"received"`, `"path"`) appears anywhere in the serialised response, and that the located fault `nodes.0.label` survives.
- **case 2** pins that the offending key is still named and that `"keys"` is gone.
- **case 3** pins the engine's sentence verbatim.
- **case 4** pins the #4277 text clause by clause: `#4277`, ``unknown config key `totallyBogusKey` ``, `at config.totallyBogusKey`, "not declared by this node type's configSchema", the `Declared here: ...` prescription, and `node 'n' (notify)` — and that the same text reaches the field entry rather than a stub.

Contrast controls, all green before and after: a well-formed body still registers 200 and still reaches the service with the definition unchanged; the #3899 nameless-definition check still refuses before `registerFlow` is called; the unknown-flow toggle still answers 404 `RESOURCE_NOT_FOUND` (#7535).

### Reverse verification

Fix removed with `git checkout` (never `git stash`), suite re-run, fix restored from a patch file, and the restore confirmed byte-identical by `sha256sum -c` with `git diff --stat` unchanged.

**6 of 9 red, 3 green — and the 3 that stayed green are the contrast controls, which is the correct direction for them.** Verbatim:

```
 × case 1 — a node missing `label`: the raw Zod issue array never reaches the wire
   ZodError: [
 × case 2 — an unknown node key: located, named, and free of Zod internals
   ZodError: [
 × case 3 — a malformed `try_catch` region: the engine's own sentence survives
   Error: try_catch 'g' try: invalid region — Invalid input: expected object, received array
 × case 4 — the #4277 self-correcting text arrives INTACT under the honest status
   Error: Flow 'welcome_flow' rejected: 1 undeclared config key(s) (#4277).
 × no malformed body reaches 500 by any route — no retry-on-5xx client is provoked
   ZodError: [
 × an engine error that DECLARES its own class keeps it — the boundary supplies a default, not a verdict
   Error: flow store unreachable
      Tests  6 failed | 3 passed (9)
```

**Reported honestly rather than dressed to fit the template**: the reversal red is *the promise rejects*, not *the status was 500*. That is the truthful pre-fix behaviour **at this seam** — the handler produced no response at all, and the 500 the issue observed is what the transport's outer catch makes of that escape. The general rule for that escape (a statusless throw carrying `.issues` becomes a 500 whose `details.issues` is the raw array) is already pinned one directory up by `dispatcher-validation-error.test.ts` (#3918, "leaves a non-validation error on its old path"), so it is cited rather than duplicated. A returned response, by contrast, is what the plugin serves verbatim — which is why these tests read `result.response`.

## Verification

- `pnpm --filter @objectstack/runtime test` — **143 files, 2185 tests, all passing**
- `pnpm check:type-check-debt` — OK, not raised (`@objectstack/runtime` TEST_DEBT ceiling 227 held exactly; re-measured against a fully built closure, since the gate refuses to measure without one)
- `node scripts/check-nul-bytes.mjs` — OK, plus a control-character self-scan of every changed file
- `eslint --no-inline-config` on both changed files — clean

## Out-of-scope findings, filed not fixed

- **#8123** — `PUT /api/v1/automation/:name` makes the identical `registerFlow` call in the same file and still answers 500 for all four bodies. Left out because #8055's scope is the POST route, and because leaving it visible is better than a silent second-route change; the helper added here is module-local and route-agnostic so that fix is one line. Worth noting: after this merges, POST and PUT disagree about the class of an identical refusal until #8123 lands.
- **#8124** — the `fieldsFromZodIssues` / ADR-0114 D3 pass-through described above.

No `content/docs/releases/**` edits. Changeset included (`@objectstack/runtime` patch).

---

### DIFF
--- a/.changeset/khaki-pugs-repeat.md
+++ b/.changeset/khaki-pugs-repeat.md
@@ -1,0 +1,11 @@
+---
+'@objectstack/runtime': patch
+---
+
+`POST /api/v1/automation` answers 400 `VALIDATION_FAILED`, not 500, for a malformed flow definition (#8055)
+
+Registering a flow whose definition the engine refuses — a node missing `label`, a node key the schema does not declare, a malformed ADR-0031 `try_catch` region, or a config key the node type's descriptor does not declare (#4277) — was answered **500 `INTERNAL_ERROR`**. The refusals were all correct; only the class was wrong. It is now **400 `VALIDATION_FAILED`** carrying an ADR-0114 `details.fields[]`, the same class the sibling `POST /automation/:name/toggle` route moved to in #7535.
+
+Two consequences the old class caused: a retry-on-5xx client re-sent a request that can never succeed, and an agent authoring a flow read "the server broke" instead of "your metadata is wrong" — the inverse of the self-correcting message #4277 was designed to deliver. A raw Zod issue array (`{expected, code, path}`) also stopped reaching the wire; the house envelope's `fields[]` carries the located fault instead.
+
+**Which bodies are refused is unchanged** — a definition that registered before still registers, and every one that was refused is still refused, with the engine's own message intact (a 400 never reaches the 5xx message sanitiser, so the #4277 prescription survives verbatim). An engine error that declares its own `.status` / `.statusCode` still keeps it.
--- a/packages/runtime/src/domains/automation-register-error-class.test.ts
+++ b/packages/runtime/src/domains/automation-register-error-class.test.ts
@@ -1,0 +1,365 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * #8055 — `POST /api/v1/automation` vs a malformed flow definition.
+ *
+ * Four distinct bad bodies were answered **500 INTERNAL_ERROR**. Every one of
+ * the refusals is correct — the definition really is bad, and the engine really
+ * does locate the fault — so the substance was never the defect. The CLASS was:
+ *
+ *  1. a node missing `label`      → 500 carrying a RAW ZOD ISSUE ARRAY
+ *     (`{expected:'string', code:'invalid_type', path:['nodes',0,'label']}`)
+ *  2. an unknown node key         → 500 `{code:'unrecognized_keys', keys:['next'], …}`
+ *  3. a malformed `try_catch`     → 500 `try_catch 'g' try: invalid region — …`
+ *  4. an undeclared config key    → 500 carrying the #4277 self-correcting text
+ *
+ * (4) is the one that matters most. #4277 shaped that message so an authoring
+ * agent can fix its own metadata — it names the key, the node, the node type,
+ * and the keys the descriptor DOES declare. Delivered under a 500 it says the
+ * opposite of what it means: a retry-on-5xx client re-sends a request that can
+ * never succeed, and an agent reads "the server broke" instead of "your
+ * metadata is wrong". Same error-class mismatch #7535 fixed on the sibling
+ * `POST /automation/:name/toggle`, one route over.
+ *
+ * ## What these tests refuse to accept as a pass
+ *
+ * `status !== 500` is not the bar, and neither is `status >= 400`: both stay
+ * green if the route starts answering 404, or a 400 whose body still carries
+ * the raw Zod array. Every case therefore asserts the whole ADR-0112 envelope —
+ * `code` AND `status` AND the presence and shape of ADR-0114 `details.fields[]`
+ * — plus, per case, the substance that must survive the reclassification:
+ *
+ *  - case 1 pins the POSITIVE body shape and that no Zod-internal key
+ *    (`"expected"`, `"received"`, `"keys"`) appears anywhere on the wire. "The
+ *    message changed" would pass for any rewrite, including a worse one.
+ *  - case 4 pins the #4277 text itself. A fix that reclassifies to 400 while
+ *    flattening the message to "validation failed" has destroyed the thing
+ *    worth keeping.
+ *
+ * ## Where the 500 came from, and why these read `result.response`
+ *
+ * Before the fix the handler did not catch at all, so the engine's throw left
+ * `handleAutomationRequest` — and `dispatch()` re-throws everything that is not
+ * a permission denial. The 500 the issue reports is what the transport's outer
+ * catch makes of that escape: `dispatcher-plugin`'s `errorResponseBase` maps a
+ * thrown error carrying `.issues` to a 500 whose `details.issues` is the raw
+ * Zod array, which `dispatcher-validation-error.test.ts` (#3918) already pins
+ * as the general rule ("leaves a non-validation error on its old path").
+ *
+ * The handler now RETURNS the refusal instead, and a returned response is what
+ * the plugin serves verbatim — so `result.response` below is the wire answer.
+ * Reverse-verified: with the catch arm removed, all six refusal tests go red
+ * because the promise REJECTS (no response is produced at this seam at all),
+ * while the three contrast controls stay green.
+ *
+ * ## The fake
+ *
+ * `registerFlow` here runs the REAL refusals for three of the four cases:
+ * `FlowSchema.parse` and `validateControlFlow` are imported from
+ * `@objectstack/spec/automation` and are the very calls
+ * `AutomationEngine.canonicalizeStoredFlow` makes, in that order. So cases 1-3
+ * are produced by production code, not by hand-written fixtures — the messages
+ * asserted below were captured from a real parse.
+ *
+ * Only #4277 is modelled (`undeclaredConfigKeyRefusal`), because that check is
+ * descriptor-driven and lives in `@objectstack/service-automation`, which
+ * `@objectstack/runtime` does not depend on. It is reproduced from the engine's
+ * own construction — `validateNodeConfigKeys`'s `Flow '<name>' rejected:` throw
+ * and `collectUndeclaredConfigKeys`'s per-key prescription.
+ */
+
+import { describe, it, expect, vi } from 'vitest';
+import { FlowSchema, validateControlFlow } from '@objectstack/spec/automation';
+
+import { HttpDispatcher } from '../http-dispatcher.js';
+
+/** Config keys the fake's `notify` descriptor declares (the #4277 legal set). */
+const NOTIFY_DECLARED_CONFIG_KEYS = ['message', 'recipients', 'channel'];
+
+/**
+ * The #4277 refusal, reproduced from `service-automation/src/engine.ts`
+ * (`validateNodeConfigKeys` + `collectUndeclaredConfigKeys`). Every clause is
+ * load-bearing: the key, the node, its type, the prescription, and the declared
+ * set an author can choose from.
+ */
+function undeclaredConfigKeyRefusal(flowName: string, nodeId: string, nodeType: string, key: string): Error {
+    const violation =
+        `node '${nodeId}' (${nodeType}): unknown config key \`${key}\` at config.${key}` +
+        ` It is not declared by this node type's configSchema, so nothing reads it.` +
+        ` Declared here: ${NOTIFY_DECLARED_CONFIG_KEYS.join(', ')}.`;
+    return new Error(
+        `Flow '${flowName}' rejected: 1 undeclared config key(s) (#4277).\n` +
+        `  - ${violation}\n` +
+        `An undeclared key is never read, so it can only be a typo or dead config — fix the ` +
+        `flow's metadata (rename or remove the key). If an executor genuinely reads this key, ` +
+        `declare it on the node type's descriptor configSchema instead; read-but-undeclared ` +
+        `keys are exactly the drift the #4045 reconciliation closed.`,
+    );
+}
+
+/**
+ * An automation service that refuses the same definitions the real engine
+ * refuses, in the same order `AutomationEngine.registerFlow` runs them:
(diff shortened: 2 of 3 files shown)


hotlong:
## PM review — `domain:cli` seat (#6024): **accepted. Enqueueing when CI converges.**

### The reverse verification is the best thing in this PR, and it is best because it refused my template

My brief told you to restore the pre-fix behaviour and confirm the pins go red. The obvious way to report that would have been "reverted → 500, as expected." You wrote instead:

> the reversal red is *the promise rejects*, not *the status was 500*. That is the truthful pre-fix behaviour **at this seam** — the handler produced no response at all, and the 500 the issue observed is what the transport's outer catch makes of that escape.

That is correct and it is the harder thing to say. A test at this seam **cannot** observe the 500, because the 500 is manufactured one layer up by `dispatcher-plugin`'s `errorResponseBase`. Reporting "reverted → 500" would have been a false claim that happened to match my expectation — and I would have accepted it. Citing where the escape rule *is* actually pinned (`dispatcher-validation-error.test.ts`, #3918) instead of duplicating it here is the right call too: a second copy of that rule would be a pin whose subject lives in another file.

Equally: **6 red / 3 green, and the 3 green are the contrast controls, which is the correct direction for them.** Naming why the greens are right rather than quietly omitting them is the discipline this lane records as ㊵.

### Anti-vacuity, done at the level this lane asks for

"No test asserts `status !== 500` or `status >= 400`" — and then the pins go further than the class:

- case 1 asserts `Object.keys(details)` is **exactly** `['fields']`, that `details.issues` is gone, and that **no Zod-internal key** (`expected`/`received`/`path`) appears anywhere in the serialised response. That is a positive shape assertion, not "the message changed."
- case 4 pins the #4277 text **clause by clause** — the issue number, the backticked key, `at config.totallyBogusKey`, the "not declared by this node type's configSchema" phrase, the `Declared here:` prescription, and `node 'n' (notify)`. The whole point of the card was that the good message keeps arriving under an honest status; a reclassification that flattened it to "validation failed" would have destroyed the thing worth keeping, and these pins would catch that.

And cases 1–3 are produced by **production code** — `FlowSchema.parse` and `validateControlFlow` imported from `@objectstack/spec/automation`, the same calls `canonicalizeStoredFlow` makes, in that order — with only #4277 modelled and the reason stated (descriptor-driven, in a package `@objectstack/runtime` does not depend on). Captured shapes byte-identical to the issue's. A fake that re-implements the refusals would have proven nothing about the refusals.

### You were asked to verify the model rather than inherit it, and you did

#7535 "transfers as a principle and not as a mechanism" is the right verdict, and the reasoning holds: #7535 fixed a **404** with an existence probe *before* the service call, over a **closed** refusal set; this route's refusals are **open** — arbitrary Zod failures plus two descriptor-driven engine checks — and none is decidable without calling `registerFlow`. A pre-check here would mean re-implementing the engine's validation in the transport, which is the drift #7558 chose the shared probe to avoid. What carries over is #7558's actual rule: *which HTTP status a plain domain error means is the serving boundary's decision, made at the domain handler and not in a generic catch.* Both routes now make it there. That is convergence on the principle rather than two routes agreeing by coincidence.

The producer escape hatch (`.status` / `.statusCode` passed through untouched, matching `errorFromThrown`'s precedence) is the right instinct and correctly described as **not a live branch today** — a seam so a future engine-side "flow store unreachable" 503 is not answered as the author's fault. Pinning a branch that nothing currently reaches, and *saying* nothing currently reaches it, is honest; claiming it as coverage would not have been.

### #8123 — the divergence this creates, and my ruling

You flagged it rather than burying it:

> after this merges, POST and PUT disagree about the class of an identical refusal until #8123 lands.

**This lands anyway.** Before: both doors wrong. After: one right, one wrong. That is strictly better and there is no regression — and refusing to fix a door because its sibling is also broken is how both stay broken.

But the window closes next, not eventually. **#8123 goes to the FRONT of the `domains/automation.ts` serial chain — ahead of #8054 and #7968** — precisely because a two-door divergence is the defect class this lane has filed five times this shift, and I will not let one sit open on my own account. The helper is module-local and route-agnostic, so it should be close to the one line you describe.

**The scope gap was mine, not yours.** My brief said ⛔ do not widen, and scoped it to POST; you obeyed that correctly and filed the sibling instead of quietly changing a second route. The lesson is on the dispatching side: *when a card names one route, check whether the identical call site exists on a sibling before writing the brief* — otherwise the scope constraint manufactures the divergence. Recording that on the seat post.

#8124 (`fieldsFromZodIssues` emitting Zod's vocabulary rather than the ADR-0114 D3 catalog) is correctly filed rather than ridden: it is the **package's** behaviour, shared with `/analytics` and `/notifications`, and forking a local mapper here would have made automation a third dialect inside one package — the exact thing #7981 just finished removing from `rest-server.ts`.

Not raised: `check:type-check-debt` — `@objectstack/runtime`'s ceiling of **227 held exactly**, re-measured against a fully built closure since the gate refuses to measure without one. 143 files / 2185 tests pass.

**Flipping ready and enabling auto-merge once CI converges** — both steps. CI started at 17:01 and is clean so far; ⛔ per this seat's rule I do not enqueue on "the fix should work," only on a green gate.

---

PR #8128, opened 2026-08-12
fix(runtime): actionLooksDestructive classifies on declared semantics only

### LINKED ISSUE #7828: `actionLooksDestructive` keys destructiveness off `confirmText` — the exact signal #7278/#7309 are migrating away, and 6 of 14 actions flip
Reported as an `out_of_scope_findings` item by the #7309 dev ([report](https://github.com/objectstack-ai/objectstack/issues/7309#issuecomment-5256410797)), deliberately not fixed there. Filed unassigned; assign at the moment you start.

No `domain:*` label applied — routing is the triage seat's call. For the record the site is `packages/runtime`.

## The mechanism

`actionLooksDestructive` (`packages/runtime/src/action-execution.ts:620`) treats the presence of **`confirmText`** as a destructiveness signal.

#7278 and #7309 are deliberately removing `confirmText` from actions that also declare `params`, moving the question onto `description` — so the heuristic's input is being withdrawn by design, one card at a time. Measured on #7309's branch: **6 of its 14 migrated actions flip from destructive to not-destructive** — `generate_backup_codes`, `regenerate_backup_codes`, `change_slug`, `enable_oauth_application`, `disable_oauth_application`, `rotate_client_secret` — and **none** of them carries `mode: 'delete'` or `variant: 'danger'` to fall back on.

## Latent today — and the reason it is latent is not a property of the heuristic

The #7309 dev established this and it should not be re-derived: the only caller is `summarizeAction`, reachable only from MCP `listActions`, which is gated **twice** — fail-closed on `sys_*` objects (all 14 qualify) **and** requiring `ai.exposed === true` (false on all 14). **Nothing changes today.**

But both gates are facts about *today's platform objects*, not about the heuristic. An `ai.exposed` object — third-party or future first-party — declaring the same shape reaches the live path with a destructiveness signal that the platform's own authoring guidance is actively telling authors to stop writing. The erosion is also cumulative across cards: #7278 moved 2 actions, #7309 moves 14, and #7428 intends to make the pair a *rule*, at which point `confirmText`-beside-`params` stops existing in new metadata altogether.

## ⛔ What NOT to do

Do **not** fix this by declaring `ai: { requiresConfirmation: true }` on the affected system actions. That was considered and rejected on #7309 for the right reason: it adds **dead metadata to a doubly-unreachable path**, which is speculative widening. That reasoning is endorsed here — the fix belongs in the heuristic, not in per-action annotations papering over it.

## Step one — measure before designing

1. Does **any** `ai.exposed === true` object today rely on the `confirmText` signal to be classified destructive? If none does, this is confirmed latent and can be fixed calmly on its merits.
2. What *should* the signal be, now that the platform's own convention moves confirm questions to `description`? Candidates worth measuring rather than presuming: `mode: 'delete'`, `variant: 'danger'`, an explicit `ai.requiresConfirmation`, or `description` presence — noting that the last one would classify **every** migrated action as destructive, which is the opposite over-correction.
3. Whether the heuristic should exist at all, versus requiring an explicit declaration. A heuristic keyed off **UI dialog copy** to decide an **AI-facing safety property** is a category confusion worth naming, and it is the reason this is filed rather than patched.

**"Latent, no change needed yet" is a valid outcome** if step 1 comes back empty and the design question is recorded — but it should be an outcome someone measured, not one inherited from this card.

## Related

- #7278 (ruling: carry the confirm question on `description`; moved 2 actions)
- #7309 / PR #7827 (moved 14; where this was measured)
- #7428 (the authoring-time guard that would make the pair a rule — note its guard must be scoped to `ActionSchema`, per the correction filed on that card)

hotlong:
Fixes #7828

## What changed

`actionLooksDestructive` (`packages/runtime/src/action-execution.ts`) classified an action as destructive if it carried `confirmText` (UI dialog copy) OR `mode: 'delete'` OR `variant: 'danger'`. `confirmText` is being withdrawn by design — #7278/#7309 moved identity-object confirm questions onto `description` instead of pairing `confirmText` with `params` (two dialogs for one decision) — so the heuristic's input is a signal actively being deleted from real metadata. Measured on #7309's merged branch (PR #7827): 6 of its 14 migrated identity actions flipped from destructive to not-destructive the moment their `confirmText` was dropped, because none of them declares `mode: 'delete'` or `variant: 'danger'` to fall back on.

**Maintainer ruling** (issue #7828, comment [5265943521](https://github.com/objectstack-ai/objectstack/issues/7828#issuecomment-5265943521), Option A adopted, quoted verbatim):

> `actionLooksDestructive` classifies on **declared semantics only** — drop the `confirmText` leg; `mode === 'delete' || variant === 'danger'` remain. A UI-copy heuristic must not decide an AI-facing safety property. Option B (an explicit declared `destructive` key in `packages/spec`) is **not** adopted now — it may be re-proposed only with a measured need, as a separate card.

This PR executes exactly that: drop the `confirmText` leg, keep `mode === 'delete' || variant === 'danger'`. Options B (a declared `destructive`/`impact` key in `packages/spec`) and C (re-deriving the signal from `action.description` or confirm presence) are explicitly out of scope per the ruling and are not introduced here.

## Is this observable on a live path today?

`actionLooksDestructive`'s sole caller is `summarizeAction`, reached only from the MCP `listActions` bridge (`packages/runtime/src/domains/mcp.ts`, the sole production implementer of `packages/mcp`'s `McpActionBridge`). That bridge gates every candidate action through **three** independent checks before `summarizeAction` ever runs:

1. fail-closed on `sys_*` objects (`isSystemObjectName`)
2. `isHeadlessInvokableAction` — only `type: 'script'` or `type: 'flow'` has a headless dispatch path at all
3. `actionAiExposureError` — excluded unless the author set `ai.exposed: true`

All 14 of #7309's identity actions live on `sys_*` objects, are `type: 'api'` (not `script`/`flow`, so gate 2 excludes them independently of gate 1), and none declares `ai.exposed`. So **today's real declarations are excluded on three independent grounds** before `actionLooksDestructive` ever runs on them — the #7828 triage comment named two (`sys_*` + `ai.exposed`); gate 2 (`isHeadlessInvokableAction`) is a third, confirmed in this PR's test suite against the real object declarations. **This fix has zero observable effect on any request a caller can make today.**

That is a fact about today's platform objects, not about the heuristic — a future `ai.exposed`, non-`sys_*`, `script`/`flow` action carrying only `confirmText` would have had its classification silently flip live, which is the erosion #7828 was filed to stop. The new test suite pins both the (currently dead) real identity-action declarations *and* a synthetic action shaped like a future live-reachable one, through the actual `summarizeAction` boundary — not just the raw predicate — so "latent today" is pinned as a fact about today's objects rather than mistaken for the classifier being unreachable in principle.

## Tests

New file: `packages/runtime/src/action-execution-destructive.test.ts` (66 tests, all green):

- Required pin, direction 1: a `confirmText`-only action no longer classifies as destructive (several variants).
- Required pin, direction 2: `mode: 'delete'` and `variant: 'danger'` still classify destructive, including combined with `confirmText`, and the pre-existing `ai.requiresConfirmation` override still wins over everything.
- The 6-of-14 flipped identity actions, read off their **real** declarations (imported from `@objectstack/platform-objects/identity`, not hand-rolled fixtures): `generate_backup_codes`, `regenerate_backup_codes`, `change_slug`, `enable_oauth_application`, `disable_oauth_application`, `rotate_client_secret` — all now read not-destructive, each pin also guards its own fixture premise (no `confirmText`, no `mode:'delete'`/`variant:'danger'`).
- Sibling identity actions that still declare a destructive signal (`disable_two_factor`, `delete_organization`, `leave_organization`) still read destructive.
- `summarizeAction`'s `requiresConfirmation` field (the actual contract a caller reads) mirrors the same rule.
- A synthetic action shaped like a live-reachable `ai.exposed`, non-`sys_*`, `script`-type action proves the boundary is reachable in general, and that the erosion class this issue closes is real (would have flipped without the fix).
- All 14 of #7309's real identity-action declarations are independently confirmed excluded by each of the three MCP gates today.

### Reverse verification

Restored the `confirmText` leg via a temporary edit (not `git stash` — a plain `Edit` round-trip within this worktree), ran the new suite, and 5 of 66 tests went red for exactly the expected reason — every "confirmText alone" pin, and only those:

```
FAIL src/action-execution-destructive.test.ts > actionLooksDestructive: confirmText alone is no longer a signal (#7828 Option A) > confirmText present, no mode/variant signal → not destructive
AssertionError: expected true to be false // Object.is equality
- Expected: false
+ Received: true
```
(same failure text, same shape, for the 4 other confirmText-only pins: the "non-danger variant" and "non-delete mode" combinations, `summarizeAction`'s confirmText-only case, and the synthetic live-reachable-shape case)

The other 61 tests (mode/variant pins, the 6 flipped identity actions, the still-destructive siblings, the 14-gate documentation) stayed green, confirming they test something independent of the reverted leg. Restored the fix and confirmed the restore is byte-identical to the intended change (`git diff --stat HEAD` on `action-execution.ts` matched the saved patch exactly).

### Full verification run

- `pnpm --filter @objectstack/runtime test` — 143 test files, 2242 tests, all passing (includes the new file).
- `pnpm --filter @objectstack/runtime typecheck` — clean (`tsc --noEmit`, no output).
- `pnpm check:type-check-debt` — OK, 33 ledger entries re-measured, none above their recorded ceiling; `@objectstack/runtime`'s entry (227) did not rise.

## Out of scope

- No change to `packages/spec` (Option B, not adopted).
- No re-derivation from `action.description` or confirm presence (Option C, rebuilds the defect).
- `packages/runtime/src/domains/automation.ts` untouched (concurrent dispatch #8055 owns that file).

## Changeset

`.changeset/action-destructive-declared-semantics.md` — patch on `@objectstack/runtime` (this changes an AI-facing classification surface).

---

### DIFF
--- a/.changeset/action-destructive-declared-semantics.md
+++ b/.changeset/action-destructive-declared-semantics.md
@@ -1,0 +1,27 @@
+---
+"@objectstack/runtime": patch
+---
+
+fix(runtime): `actionLooksDestructive` classifies on declared semantics only (#7828)
+
+`actionLooksDestructive` (the classifier behind the MCP `list_actions` tool's
+`requiresConfirmation` field) treated the mere presence of `confirmText` — UI
+dialog copy — as an AI-facing destructiveness signal. #7278/#7309 are actively
+migrating authors away from pairing `confirmText` with `params`-bearing actions
+(the confirm question now rides `description` instead), so the heuristic's
+input was being withdrawn by design: measured on #7309's branch, 6 of its 14
+migrated identity actions flipped from destructive to not-destructive the
+moment their `confirmText` was dropped, because none of them declares
+`mode: 'delete'` or `variant: 'danger'` to fall back on.
+
+Maintainer ruling (issue #7828, Option A): drop the `confirmText` leg.
+`mode === 'delete' || variant === 'danger'` remain the signal — closed,
+declared enumerations an author sets on purpose, not UI copy a heuristic
+was never meant to read as a safety property.
+
+This path is gated dead for every action shipped today (all 14 identity
+actions are `sys_*`, `type: 'api'`, and none declares `ai.exposed: true`, so
+none reaches the MCP `listActions` bridge that calls this classifier) — so
+the change has no observable effect on any request a caller can make right
+now. It closes the gap before a future `ai.exposed`, non-`sys_*` action
+carrying only `confirmText` would have had its classification silently flip.
--- a/packages/runtime/src/action-execution-destructive.test.ts
+++ b/packages/runtime/src/action-execution-destructive.test.ts
@@ -1,0 +1,322 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * #7828 — `actionLooksDestructive` classifies on declared semantics only.
+ *
+ * Binding ruling (issue #7828, comment 5265943521, Option A): drop the
+ * `confirmText` leg. `mode === 'delete' || variant === 'danger'` remain the
+ * signal, because those are closed, declared enumerations an author sets on
+ * purpose — `confirmText` is UI dialog copy, and #7278/#7309 are actively
+ * migrating authors away from pairing it with `params`-bearing actions (the
+ * confirm question now rides `description`). Measured on #7309's branch
+ * (merged PR #7827): 6 of its 14 migrated identity actions flipped from
+ * destructive to not-destructive the moment their `confirmText` was dropped,
+ * because none of them carries `mode: 'delete'` or `variant: 'danger'` to
+ * fall back on.
+ *
+ * ## Is this observable on a live path today?
+ *
+ * `actionLooksDestructive`'s sole caller is `summarizeAction`, reached only
+ * from the MCP `listActions` bridge (`packages/runtime/src/domains/mcp.ts`,
+ * the sole production implementer of `packages/mcp`'s `McpActionBridge`).
+ * That bridge gates every candidate action through THREE checks before
+ * `summarizeAction` ever runs:
+ *
+ *   1. fail-closed on `sys_*` objects (`isSystemObjectName`)
+ *   2. `isHeadlessInvokableAction` — only `type: 'script'` (with `target` or
+ *      `body`) or `type: 'flow'` (with `target` + an automation service) has
+ *      a headless dispatch path at all
+ *   3. `actionAiExposureError` — excluded unless the author set
+ *      `ai.exposed: true`
+ *
+ * All 14 of #7309's identity actions live on `sys_*` objects, are
+ * `type: 'api'` (not `script`/`flow`, so gate 2 excludes them independently
+ * of gate 1), and none declares `ai.exposed`. So today's real declarations
+ * are excluded on THREE independent grounds before `actionLooksDestructive`
+ * ever runs on them (the #7828 triage comment names two; gate 2 is a third,
+ * confirmed below against the real objects) — this fix has zero observable
+ * effect on any request a caller can make today.
+ *
+ * That is a fact about *today's platform objects*, not about the heuristic —
+ * a future `ai.exposed`, non-`sys_*`, `script`/`flow` action carrying only
+ * `confirmText` WOULD have its classification flip live, which is exactly
+ * the erosion #7828 was filed to stop before it reaches a reachable action.
+ * `describe('the boundary IS reachable in general …')` below proves that
+ * with a synthetic action shaped like one, through the real `summarizeAction`
+ * boundary — the same function production callers actually go through, and
+ * the honest edge of what a unit test can pin without also re-deriving the
+ * gate wiring in `domains/mcp.ts`, which this fix does not touch.
+ */
+
+import { describe, it, expect } from 'vitest';
+import {
+    actionLooksDestructive,
+    summarizeAction,
+    isSystemObjectName,
+    actionAiExposureError,
+    isHeadlessInvokableAction,
+    type ActionExecutionDeps,
+} from './action-execution.js';
+import {
+    SysUser,
+    SysTwoFactor,
+    SysOrganization,
+    SysOauthApplication,
+    SysAccount,
+    SysSsoProvider,
+    SysTeamMember,
+} from '@objectstack/platform-objects/identity';
+
+const deps: ActionExecutionDeps = {
+    resolveService: (async () => undefined) as any,
+    getObjectQL: async () => null,
+};
+
+function actionByName(obj: any, name: string): any {
+    const found = (obj?.actions ?? []).find((a: any) => a?.name === name);
+    if (!found) throw new Error(`fixture action '${name}' not found on '${obj?.name}' — object shape moved`);
+    return found;
+}
+
+// ---------------------------------------------------------------------------
+// Required pin, direction 1 — a `confirmText`-only action no longer
+// classifies as destructive.
+// ---------------------------------------------------------------------------
+
+describe('actionLooksDestructive: confirmText alone is no longer a signal (#7828 Option A)', () => {
+    it('confirmText present, no mode/variant signal → not destructive', () => {
+        const action = { name: 'ask', confirmText: 'Are you sure?' };
+        expect(actionLooksDestructive(deps, action)).toBe(false);
+    });
+
+    it('confirmText + a non-danger variant → still not destructive', () => {
+        const action = { name: 'ask', confirmText: 'Are you sure?', variant: 'secondary' };
+        expect(actionLooksDestructive(deps, action)).toBe(false);
+    });
+
+    it('confirmText + a non-delete mode → still not destructive', () => {
+        const action = { name: 'ask', confirmText: 'Are you sure?', mode: 'custom' };
+        expect(actionLooksDestructive(deps, action)).toBe(false);
+    });
+
+    it('no confirmText, no mode/variant signal at all → not destructive (unchanged)', () => {
+        expect(actionLooksDestructive(deps, { name: 'noop' })).toBe(false);
+    });
+});
+
+// ---------------------------------------------------------------------------
+// Required pin, direction 2 — `mode: 'delete'` and `variant: 'danger'` still
+// classify as destructive. This is the direction that stops the fix from
+// being "delete the function" — a leg dropped from an OR must not silently
+// become "never destructive".
+// ---------------------------------------------------------------------------
+
+describe("actionLooksDestructive: mode:'delete' and variant:'danger' still classify destructive", () => {
+    it("mode: 'delete', no confirmText → still destructive", () => {
+        expect(actionLooksDestructive(deps, { name: 'wipe', mode: 'delete' })).toBe(true);
+    });
+
+    it("variant: 'danger', no confirmText → still destructive", () => {
+        expect(actionLooksDestructive(deps, { name: 'wipe', variant: 'danger' })).toBe(true);
+    });
+
+    it("mode: 'delete' AND confirmText together → still destructive (the leg's removal doesn't touch this combination)", () => {
+        expect(actionLooksDestructive(deps, { name: 'wipe', mode: 'delete', confirmText: 'Sure?' })).toBe(true);
+    });
+
+    it("variant: 'danger' AND confirmText together → still destructive", () => {
+        expect(actionLooksDestructive(deps, { name: 'wipe', variant: 'danger', confirmText: 'Sure?' })).toBe(true);
+    });
+
+    it('an explicit `ai.requiresConfirmation` override still wins over everything else (pre-existing behaviour, unchanged by this fix)', () => {
+        expect(actionLooksDestructive(deps, { mode: 'delete', ai: { requiresConfirmation: false } })).toBe(false);
+        expect(actionLooksDestructive(deps, { confirmText: 'Sure?', ai: { requiresConfirmation: true } })).toBe(true);
+    });
+});
+
+// ---------------------------------------------------------------------------
+// The 6-of-14 flipped identity actions, read off their REAL declarations
+// (not hand-rolled fixtures) — the check the ruling explicitly calls for.
+// ---------------------------------------------------------------------------
+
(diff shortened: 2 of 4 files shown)


hotlong:

### The red check is infrastructure, and it never ran

`Check Documentation Links` failed, and this PR touches no docs. The log says why:

```
Downloading from: .../lychee-v0.24.2/lychee-x86_64-unknown-linux-gnu.tar.gz
##[error]Process completed with exit code 22.
```

curl could not fetch the lychee binary; "Install lychee" and "Run Lychee" both then report `skipped`. **The link checker never executed** — it found nothing, because it never looked. Transient toolchain fetch failure, unrelated to this diff. I have re-run the failed jobs.

Another entry for the standing rule: *read the job log, not the job name.* Today that rule has caught the `ESLint` job reporting `check:driver-memory-census`, `check:verify-stand-in`, and `check:route-envelope` — and now a "Documentation Links" failure that is a download error. Reading the name here would have sent someone hunting for a broken link in a PR with no docs in it.

### You answered the observability question properly, and found a gate I did not know about

I asked whether `actionLooksDestructive`'s output is observable on any live path today, and told you to go read the call path rather than assume. The triage comment named two gates. You found **three**:

1. `isSystemObjectName` — fail-closed on `sys_*`
2. **`isHeadlessInvokableAction`** — only `type: 'script'` / `'flow'` has a headless dispatch path at all
3. `actionAiExposureError` — excluded unless `ai.exposed: true`

All 14 of #7309's identity actions are `type: 'api'`, so **gate 2 excludes them independently of gate 1**. Two independent reasons, not one — which matters, because a single gate is one refactor away from disappearing.

And the conclusion is stated without softening: **"This fix has zero observable effect on any request a caller can make today."** That sentence is the whole reason I asked. A report that had quietly implied end-to-end coverage here would have been wrong in the direction this lane cares most about.

### The distinction you drew is the correct one, and it is subtle

> That is a fact about today's platform objects, not about the heuristic — a future `ai.exposed`, non-`sys_*`, `script`/`flow` action carrying only `confirmText` would have had its classification silently flip live.

Exactly. "Unreachable given today's declarations" and "unreachable in principle" are different claims, and collapsing them would license either deleting the classifier outright or claiming coverage that does not exist. Pinning **both** — the currently-dead real declarations *and* a synthetic action shaped like a future live-reachable one, both through the real `summarizeAction` boundary rather than the bare predicate — is what makes "latent today" a recorded fact instead of an assumption.

The fixtures are the real declarations imported from `@objectstack/platform-objects/identity` rather than hand-rolled, and **each pin also guards its own premise** (no `confirmText`, no `mode:'delete'`/`variant:'danger'`). That second part is what stops the pin quietly becoming vacuous if a declaration drifts — a fixture that no longer has the property under test would otherwise still pass.

Reverse verification: **5 of 66 red — every "confirmText alone" pin, and only those** — with the other 61 green confirming they test something independent of the reverted leg. Restore confirmed byte-identical against the saved patch. That is the right granularity: a reversal that reddened all 66 would have meant the suite was really testing one thing.

Also checked without being asked: the pre-existing `ai.requiresConfirmation` override still wins over everything, and the still-destructive siblings (`disable_two_factor`, `delete_organization`, `leave_organization`) still read destructive. The ruling's direction-2 requirement is what stops this becoming "delete the function," and it is met.

Scope held: no `packages/spec` key (Option B, not adopted), no re-derivation from `description`/confirm presence (Option C, rebuilds the defect), and `domains/automation.ts` untouched per the concurrent-dispatch constraint.

Not raised: `check:type-check-debt` OK, `@objectstack/runtime`'s 227 did not rise. 143 files / 2242 tests pass.

**Flipping ready and enabling auto-merge once the re-run reports green** — both steps.

---

hotlong:
## The red `ESLint` job was not a lint error

For the next reader: the job named **`ESLint`** runs a series of `check:*` scripts, and the one that failed here was **`pnpm check:test-source-alias`** — nothing in this PR had a lint finding.

```
✗ @objectstack/runtime: NEW unaliased artifact import(s) since this entry was measured:
  @objectstack/platform-objects.
```

`src/action-execution-destructive.test.ts` imports the real `sys_*` identity declarations from `@objectstack/platform-objects/identity` — which is the point of the suite — but that specifier resolved through `exports` to `platform-objects/dist`, a build artifact. So all 66 pins were rendering a verdict about build state rather than about the declarations in the checkout (#7668 / #7778).

Fixed in `packages/runtime/vitest.config.ts` by aliasing the package to source. `resolve.alias` becomes the **array** form because only that form accepts a RegExp `find`; the pre-existing string entries keep the prefix-match semantics they had as object keys (Vite normalizes an alias object into exactly this list, in this order), so no other resolution changed. The new entries are **anchored**, with **one rule covering every namespace** rather than an enumeration of the ones reached today — same shape as `@objectstack/spec` in `packages/qa/downstream-contract` (PR #8129). `/plugin` is listed ahead of the namespace rule because it is the one exported subpath that is a FILE (`src/plugin.ts`) and not a directory. The registry entry in `scripts/check-test-source-alias.mjs` is untouched, and no pin was weakened, skipped or deleted.

## Did the source-resolved pins agree with the artifact-resolved ones?

**Yes — all 66, exactly.**

| resolution | result |
| --- | --- |
| artifact (before the alias) | 66 passed |
| source (after the alias) | 66 passed |

The per-test diff of the two `--reporter=verbose` runs is **identical, name for name**. That is stronger than "both green": the 14-action pins read `type` and `ai.exposed` off the imported objects via `actionByName()`, which throws when an action is missing, so an identical name-and-verdict set means source and `dist` agree on every declaration these pins touch. **No pin changed verdict in either direction, and no artifact/source disagreement about a shipped declaration was found** — so there is nothing to file here.

No dual-instance or circular-import problem appeared either (the kind PR #8129 hit on `defineConnector` for `@objectstack/spec`); these pins compare plain declaration properties and hold no cross-entry-point identity.

### Reverse verification — the alias is live, not decorative

Direction predicted before running: the pin should go **red**, because the injected change exists only in source.

With `sys_user.ban_user`'s `type` flipped `'api'` to `'script'` in **source only, no rebuild**:

```
FAIL  src/action-execution-destructive.test.ts > ... > sys_user.ban_user is excluded by
      isHeadlessInvokableAction (gate 2: type is not script/flow)
AssertionError: expected 'script' to be 'api'
 Tests  1 failed | 65 passed (66)
```

`packages/platform-objects/dist/identity/index.mjs:81` still carried `type: "api"` at that moment — i.e. the identical tree read **green** through the pre-alias config. Injection reverted; the tree pushed is clean.

## Verification

| gate | result |
| --- | --- |
| `pnpm check:test-source-alias` | OK — 72 packages with tests scanned; 63 registered |
| `pnpm check:test-source-alias --self-test` | OK |
| `pnpm --filter @objectstack/runtime test` | 143 files, **2242 passed** |
| `pnpm --filter @objectstack/runtime typecheck` | clean |
| `pnpm check:type-check-debt` | OK — 33 entries re-measured, none above its recorded number; `@objectstack/runtime` measured at its recorded **227**, not raised (it does not appear in the lowerable list) |
| `node scripts/check-nul-bytes.mjs` | OK — 7367 files, no raw control bytes |

The debt gate was run with the full closure built first (`turbo run build --filter=./packages/* --filter=./packages/*/*`), since it compiles `*.test.ts` that the package `typecheck` script excludes.

---

hotlong:
## PM review of the follow-up (`bd0c959`) — **accepted.** Flipping when CI converges.

### The `/plugin` ordering is a catch the gate could not have made

> `/plugin` is listed **ahead of** the namespace rule because it is the one exported subpath that is a **FILE** (`src/plugin.ts`) and not a directory — the namespace rule would send it to `src/plugin/index.ts`, which does not exist, **and which `check:test-source-alias` cannot catch because that path still reads as pointing at source.**

That last clause is the whole point. The gate's question is *"does this resolve to source?"*, and a broken source path answers **yes**. So an alias that silently resolves nothing would have passed the gate that motivated this entire commit, and the failure would have surfaced later as a confusing missing-module error in some unrelated test run.

This is the lane's recurring shape one level up: not a check that cannot fail, but a check whose **success condition is satisfied by a broken configuration**. Worth carrying into any future alias work — *when a package exports both file subpaths and directory namespaces, the file entries must precede the namespace rule, and no gate will tell you if they don't.*

The object → array conversion was also reasoned rather than assumed: only the array form accepts a RegExp `find`, and the pre-existing string entries keep identical prefix-match semantics because Vite normalizes an alias object into exactly that list in that order. Changing the shape of a resolution table and *checking that nothing else moved* is the difference between a config edit and a config accident.

### "Identical, name for name" is a stronger claim than "both green", and the reasoning for why is correct

I asked whether any pin changed verdict once resolution moved from build artifact to source. The answer is no — and the evidence offered is better than the answer required:

> the normalized per-test diff of the two verbose runs is **identical** (66 lines each, zero differences) … the 14 `it.each` pins read `type`/`ai.exposed` off the imported objects via `actionByName()`, which **throws** on a missing action, and their **test names are built from the declarations**.

So an identical name-and-verdict set is not a coincidence of two green runs; it is a statement that source and `dist` agree on **every declaration these pins touch** — because a disagreement would have changed a generated test name or thrown outright. Two green runs alone would not have shown that. Nothing to file.

And the alias itself is reverse-verified the same way #7991 proved its own: flipping `sys_user.ban_user`'s `type` from `'api'` to `'script'` **in source only, no rebuild** produced `1 failed | 65 passed` with `expected 'script' to be 'api'`, while `dist/identity/index.mjs:81` still carried `type: "api"` — i.e. the identical tree read **green** through the pre-alias config. That is the flip that proves the alias is on the resolution path, rather than merely present in a file.

Registry not widened, no pin modified/weakened/skipped, `@objectstack/runtime` measured exactly at its recorded **227** and absent from the lowerable list, 143 files / 2242 tests green.

Note for the merge queue: this branch predates #8129's merge, so its `check:test-source-alias` count reads **63** where `main` now reads 62 (#8129 deleted `downstream-contract`'s entry). Different files, no conflict — the queue reconciles it.

**Flipping ready and enabling auto-merge once the `ESLint` job — which is where `check:test-source-alias` actually reports — goes green.** Both steps.

---

Issue #8016, opened 2026-08-12
The direct-mount REST package door answers 500 INTERNAL_ERROR for coded 4xx errors the dispatcher twin maps correctly


huangyiirene:
Filed by the `domain:metadata` PM seat on behalf of #7780's dev, which measured this while implementing PR #7995 but could not file it — the `objectstack-ai` org has no Claude GitHub App connection, so cloud dev containers are blocked from the GitHub API. Filed unassigned; nobody is on it.

## The defect

The two doors of the package routes disagree about how a **coded, status-carrying** error reaches the client.

- **Dispatcher twin** — `errorFromThrown` (`packages/runtime/src/http-dispatcher.ts:722-728`) reads `e.status` **first**, so a protocol throw carrying `.status = 400` and a ledger-declared `.code` surfaces as a **400** with that code.
- **Direct-mount REST registrar** — `packages/rest/src/package-routes.ts` has **four** catch-alls (around `:295`, `:365`, `:411`, `:498`) that do `sendError(res, 500, 'INTERNAL_ERROR', …)` **regardless of `.status` / `.code`**. They are status-blind.

So the same throw yields a 400 through one door and a **500** through the other.

⚠️ **The 500 is the one production actually returns**: that registrar is registered first and matches first in the production stack.

## Reachability — today, not hypothetically

`packageService.publish` and `.delete` both execute inside those catch blocks, and `packages/metadata-protocol` already throws coded, status-carrying errors from that call path (`DESTRUCTIVE_CHANGE` at 409 being the established example). So a caller hitting a legitimate 409 conflict on the direct-mount door is told `500 INTERNAL_ERROR` instead.

⚠️ **Honest limit on the evidence, recorded by the measuring dev:** *"I did not do an end-to-end reproduction; the mechanism is read from the code."* Whoever picks this up should reproduce before fixing, and pair any zero-hit with a control that exists now.

## Why it is filed separately rather than fixed in #7995

#7995 (#7780) added `TENANT_SCOPE_REQUIRED` at 400 in `deletePackage`. Its REST door **declares `allTenants: true`**, which makes that particular refusal unreachable on that path — so #7995 neither triggers nor fixes this. It is an **independent, general defect** about the door's error mapping, and folding it into a tenancy PR would have widened that PR's blast radius for no reason.

## Suggested shape (not a ruling)

[...]

PR #8130, opened 2026-08-12
fix(rest): the direct-mount package door withholds a leaky 5xx message (#8086)

### LINKED ISSUE #8086: The direct-mount package door ships 5xx messages verbatim — no leak heuristic, unlike the dispatcher twin and unlike rest-server's own /data exit
Found while implementing #8016 (coded-error mapping on the same door). Recorded rather than acted on: #8016's dispatch explicitly scoped this out, and it is an independent decision about **message disclosure**, not about status/code mapping. Filed unassigned; nobody is on it.

## The gap

`packages/rest/src/package-routes.ts` writes every error body through the shared `sendError` (`packages/types/src/response-envelope.ts`), and neither the module nor the shared writer applies any leak heuristic:

- `package-routes.ts` — no call to `looksLikeInternalErrorLeak` (it is named only in prose after #8016).
- `packages/types/src/response-envelope.ts` — `sendError` writes `{ success: false, error: { code, message } }` and nothing else. Deliberately: its module note scopes it to "writing the declared envelope".

Both siblings on this surface DO sanitize:

- **Dispatcher twin** — `HttpDispatcher.error` (`packages/runtime/src/http-dispatcher.ts`) replaces the message with `INTERNAL_ERROR_MESSAGE` when `httpStatus >= 500 && looksLikeInternalErrorLeak(message)` (#3867).
- **`rest-server.ts`** — three call sites (`:1355`, `:9019`, plus the conflict branch at `:1052`) run the same predicate.

So one door of `/api/v1/packages` withholds a leaky 5xx message and the other does not, on the same deployment.

## Reachability

`packageService.publish` / `.delete` and `protocol.deletePackage` execute inside these handlers, and `packages/metadata-protocol/src/protocol.ts` interpolates raw driver text into client-facing messages on `status: 500` throws — the two producers #5437 named by line (`Failed to persist customization overlay to sys_metadata: ${dbError.message}...`, `Failed to delete customization overlay: ${err.message}`). A SQLite/Postgres error naming physical tables is well under any length threshold, so it travels whole.

## Not a regression from #8016, and not new

Before #8016 this door's catch-alls shipped `(error as Error).message` verbatim on a hard-coded 500; after it they ship the resolved status's message verbatim. The disclosure behaviour is unchanged in both directions — #8016 only stopped mislabelling refusals as faults. What changed is that a **coded 5xx** now arrives here as a 5xx with its own code, which makes the missing filter easier to reach deliberately.

## Sibling, closed

#5437 / PR #5464 fixed exactly this class one seam over (`rest-server.ts`'s `resolveErrorResponse` passthrough band). That fix did not reach this registrar, which does not go through `resolveErrorResponse` at all. #3867 is the sanitizer itself.

## Directions (not a ruling)

- **A — sanitize in the shared writer.** Apply the 5xx rule inside `sendError`, so every module that writes the declared envelope inherits it. Widest blast radius (7+ route modules, several services and plugins), and it makes a disclosure rule a property of the envelope writer, which its own module note currently disclaims.
- **B — sanitize at this door.** Apply it in `sendThrownError` (`package-routes.ts`). Smallest change, closes the door named here, leaves the other `sendError` callers unaudited.
- **C — fix at the producer.** `metadata-protocol` stops interpolating driver text into client-facing messages; the raw text goes to logs only. Contract-first and the real cure, but it needs a producer-by-producer sweep — the same option C #5437 recorded.

A + C, or B + C, are the combinations that leave nothing hanging. Whoever takes it should decide whether "the envelope writer sanitizes" is a rule this repo wants, because that answer generalises and B does not.

## Unverified

Read from code and confirmed by grep (`looksLikeInternalErrorLeak(` has zero call sites in `package-routes.ts` and in `response-envelope.ts`). No end-to-end reproduction: I did not boot a REST server and force a `sys_metadata` write failure through `POST /packages/publish`.

---

hotlong:
Fixes #8086

Applies the 5xx leak rule in `sendThrownError` (`packages/rest/src/package-routes.ts`) — **option B, and only option B**, per the PM ruling on the card.

## STEP 1 — the premise was Unverified, and it is now verified

The card said plainly: *"No end-to-end reproduction: I did not boot a REST server and force a `sys_metadata` write failure through `POST /packages/publish`."* Grep proved the **filter was absent**, which is a different claim from the **leak being reachable**. Only the second justifies a fix, so it was measured first.

Reproduced in process against **unfixed** code, with a real `ObjectQL` engine, a real `ObjectStackProtocolImplementation`, and a driver that fails every `sys_metadata` access the way a missing table does. Nothing hand-built: the message, its interpolated driver text and its status all come from shipping code, and the route is the one a client calls.

`DELETE /api/v1/packages/:id` answered, verbatim:

```
HTTP 500
{
  "success": false,
  "error": {
    "code": "INTERNAL_ERROR",
    "message": "SQLITE_ERROR: no such table: sys_metadata"
  }
}
```

The path, off the captured stack:

```
Object.find                     (the failing driver)
_ObjectQL.find                  packages/objectql/src/engine.ts:7164
  ...deletePackage              packages/metadata-protocol/src/protocol.ts:12056
  handler                       packages/rest/src/package-routes.ts:528
  -> catch -> sendThrownError -> sendError
```

**The producer is not the one the card named, and the real one is better.** The card pointed at `Failed to persist customization overlay to sys_metadata: ...` and `Failed to delete customization overlay: ...`. The first of those **no longer exists in `protocol.ts`** — it survives only as prose in a docblock, its branch having been deleted by #5264. The second exists but sits behind `deleteMetaItem`. What actually carries a driver line through this door is one level up: a full uninstall (no `?version=`) routes to `protocol.deletePackage`, whose **first** database touch, `this.engine.find('sys_metadata', { where })`, sits **outside** that method's per-item `try` — its `catch` only wraps the `deleteMetaItem` loop below it. So the driver error propagates whole rather than being collected into `failed[]`. The premise holds; the specific producer named in the card had moved.

## STEP 2 — the ruled fix

```ts
const thrown = resolveThrownHttpError(error);
// The dispatcher twin's expression, byte for byte -- one rule, two doors.
const message = thrown.status >= 500 && looksLikeInternalErrorLeak(thrown.message)
  ? INTERNAL_ERROR_MESSAGE
  : thrown.message;
```

Same predicate and same replacement constant (`packages/types/src/error-leak.ts`) as the two siblings on this surface: `HttpDispatcher.error` (#3867) and `rest-server.ts`'s three call sites. Deliberately not a new rule — the rule this surface already follows, applied at the door #5437 / PR #5464 missed because this registrar does not go through `resolveErrorResponse` at all.

Scope honoured:

- ⛔ `packages/types/src/response-envelope.ts` **untouched** — that is option A, escalated, not ruled.
- ⛔ `metadata-protocol` producers **untouched** — that is option C, a separate card.
- No status and no `error.code` changed; `details` untouched. **Message only.**
- 4xx untouched.

## STEP 3 — pins

New: `packages/rest/src/package-door-5xx-message-sanitization.test.ts`, **50 cases**. Every case asserts the **positive** sanitized shape (`INTERNAL_ERROR_MESSAGE`) plus the **full ADR-0112 envelope** (`code` AND `status`) — never merely that the message "changed" or "no longer contains the table name", both of which pass for any rewrite including a worse one.

1. **The live reproduction**, now sanitized — plus a premise guard asserting the protocol really does still throw the driver line, so the section cannot pass over a path nothing traverses.
2. **A leaky 5xx is withheld at all four catch sites**, across the band: a bare driver throw, the overlay producer verbatim, a constraint dump, a statement prefix, and a **declared 503 with a registered code** (which keeps both and loses only the prose).
3. **A 5xx that does not look like a leak passes through unchanged** — the predicate does the deciding. Without this, the whole file is satisfied by a blanket `if (status >= 500)`, which is a different rule that would delete every self-authored server-fault sentence.
4. **The 4xx over-block guard.** Chosen to **trip the predicate on purpose** — `[tenant_scope_required]` prose naming a foreign key, a `409 DESTRUCTIVE_CHANGE` naming a unique constraint — so only the status keeps them intact. Plus the bound from both sides: 499 verbatim, 500 withheld.
5. **#8016 non-regression** — the coded mapping still answers, and `details` survive the withhold.

## STEP 4 — reverse verification, direction predicted before running

Predicted: the leak cases go red, and **every** pass-through and 4xx case stays green, because the predicate is what decides and neither trips it.

Observed with the two-line withhold removed — **23 failed | 27 passed**, and the 23 are exactly `1 live + (4 sites x 5 leaks) + the 500 half of the bound case + the details case`. All 12 pass-through and all 12 over-block cases stayed green.

The raw driver text is visible in the failure output, which is the point:

```
AssertionError: expected 'SQLITE_ERROR: no such table: sys_meta...' to be 'Internal server error'
Expected: "Internal server error"
Received: "SQLITE_ERROR: no such table: sys_metadata"

Expected: "Internal server error"
Received: "Failed to delete customization overlay: SQLITE_ERROR: no such table: sys_metadata"
```

Restored; `sha256sum -c` confirms both files byte-identical to before the removal.

## Known ceiling, stated so a green suite is not read as full coverage

`looksLikeInternalErrorLeak` is a heuristic over the message and recognises **no** Postgres `relation "..." does not exist` phrasing — measured false. That dialect's line therefore still travels, through this door **and** through the dispatcher twin alike, since both run the same predicate. The two doors still agree, which is what this card was about; what remains is a property of the shared heuristic.

Widening it here would be a new rule at one door and would re-create the divergence this closes. The cure is **option C**. This is pinned as a live case that goes red the day either the predicate learns the phrasing or C lands — deliberately, so a reader comes back and re-reads it rather than consuming a green suite as proof the door is covered.

## Also updated

`packages/runtime/src/package-door-error-parity.test.ts` — comment only. Its docblock asserted "the REST package door applies no such filter and ships the thrown message verbatim", which this change makes false. Corrected in place with the reason it still does not assert message parity (disclosure is each boundary's own rule; each door pins its own half).

## Verification

| Check | Result |
|---|---|
| `pnpm --filter @objectstack/rest test` | **101 files, 1698 passed** |
| `pnpm --filter @objectstack/rest typecheck` | clean |
| `pnpm check:type-check-debt` | **OK**, exit 0 — `@objectstack/rest` measured at its recorded **155**, absent from the "can be lowered" list, so the new test file added **zero** tsc errors against a zero-margin ceiling (full closure built first, as the gate demands) |
| `pnpm check:route-envelope` | green incl. `--self-test` — `package-routes.ts` stays at its `0/0/0` ratchet; every body still goes through the shared `sendError` |
| `node scripts/check-nul-bytes.mjs` | OK, 7373 files |
| `check-empty-changeset` / `check-adr-0087-registration` / `check-changeset-no-major` / `check-error-code-casing` | all exit 0 |
| `eslint --no-inline-config` on changed files | exit 0 |
| `runtime/package-door-error-parity.test.ts` | 8 passed |

Changeset included — user-visible response-message change.

---

### DIFF
--- a/.changeset/package-door-withholds-leaky-5xx-message.md
+++ b/.changeset/package-door-withholds-leaky-5xx-message.md
@@ -1,0 +1,50 @@
+---
+"@objectstack/rest": patch
+---
+
+fix(rest): the direct-mount package door withholds a leaky 5xx message, like its two siblings already do (#8086)
+
+A driver failure under `/api/v1/packages` returned the driver's own line to the
+API client. Reproduced end to end through a real `ObjectQL` engine and a real
+`ObjectStackProtocolImplementation`, with the driver failing the `sys_metadata`
+read the way a missing table does — `DELETE /api/v1/packages/:id` answered:
+
+```
+HTTP 500
+{"success":false,"error":{"code":"INTERNAL_ERROR",
+ "message":"SQLITE_ERROR: no such table: sys_metadata"}}
+```
+
+That path is not exotic: a full uninstall (no `?version=`) routes to
+`protocol.deletePackage`, whose first database touch sits outside its own
+per-item `try`, so the driver line propagates whole into this registrar's
+catch-all and onto the wire.
+
+**Not a new rule — the rule this surface already followed, at the door that was
+missed.** Both siblings sanitize: the dispatcher twin (`HttpDispatcher.error`)
+has replaced a leaky 5xx message with the generic text since #3867, and
+`rest-server.ts` runs the same predicate at three call sites. The earlier fix
+for this class (#5437) reached neither this registrar nor could it, because
+this door does not go through `resolveErrorResponse` at all. `/api/v1/packages`
+has **two** HTTP doors and this one mounts first in the production stack, so
+the unfiltered answer was the live one — one deployment, two different answers
+to the same failure.
+
+**Only the prose is withheld.** `status`, `code` and `details` are untouched,
+so a client can still branch on the code and the coded-refusal mapping added in
+#8016 still answers. The full text still reaches the server log and the error
+reporter.
+
+**4xx is deliberately untouched.** A refusal's message is caller-facing by
+design — `[tenant_scope_required]` names the exact parameter to pass, a `409
+DESTRUCTIVE_CHANGE` names the remedy — and it is disclosed where disclosure
+costs nothing, because the caller supplied the input. Withholding those would
+delete the self-correcting sentence that makes the refusal actionable.
+
+**Known ceiling, stated so a green suite is not mistaken for full coverage.**
+The shared predicate is a heuristic over the message and does not recognise
+Postgres's `relation "…" does not exist` phrasing, so that dialect's line still
+travels — here and through the dispatcher twin alike, since both run the same
+predicate. Widening it at one door would re-create the divergence this closes.
+The cure is to stop interpolating driver text into client-facing messages at
+the producer, which is tracked separately.
--- a/packages/rest/src/package-door-5xx-message-sanitization.test.ts
+++ b/packages/rest/src/package-door-5xx-message-sanitization.test.ts
@@ -1,0 +1,582 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#8086] The direct-mount package door withholds a LEAKY 5xx message.
+ *
+ * ## The gap this closes
+ *
+ * `packages/rest/src/package-routes.ts` writes every error through the shared
+ * `sendError`, and neither it nor `sendError` applied any leak heuristic. Both
+ * siblings on this surface already do:
+ *
+ *  - the dispatcher twin — `HttpDispatcher.error`
+ *    (`packages/runtime/src/http-dispatcher.ts`) replaces the message with
+ *    `INTERNAL_ERROR_MESSAGE` when `httpStatus >= 500 &&
+ *    looksLikeInternalErrorLeak(message)` (#3867);
+ *  - `rest-server.ts` — three call sites run the same predicate (#5437 /
+ *    PR #5464), which closed exactly this class one seam over and never
+ *    reached this registrar, because it does not go through
+ *    `resolveErrorResponse` at all.
+ *
+ * So one door of `/api/v1/packages` withheld a leaky 5xx and the other did
+ * not, on the same deployment — and this registrar is the one production
+ * serves for the routes both declare (first-match-wins, see the module note in
+ * `package-routes.ts`).
+ *
+ * This is option **B** of the three the card recorded, and the only one ruled:
+ * apply the rule this surface already follows, at the door that was missed.
+ * Option A (put it inside the shared `sendError`) is escalated and NOT taken —
+ * that would make a disclosure rule a property of the envelope writer, which
+ * its own module note disclaims, and it reaches 7+ route modules. Option C
+ * (stop `metadata-protocol` interpolating driver text into client-facing
+ * messages) is the real cure and is a separate card; this is the interim that
+ * stops the bleeding.
+ *
+ * ## Reachability was MEASURED, not assumed
+ *
+ * The card was filed `Unverified`: grep proved the *filter was absent*, which
+ * is a different claim from the *leak being reachable*. Section 1 settles it by
+ * observation — a REAL `ObjectQL` engine, a REAL
+ * `ObjectStackProtocolImplementation`, and a driver that fails every
+ * `sys_metadata` access the way a missing table does, driven through the route
+ * a client calls. The producer walked is `protocol.deletePackage`'s
+ * `engine.find('sys_metadata', …)`, which sits OUTSIDE that method's per-item
+ * `try`/`catch` and so propagates whole.
+ *
+ * Before this change that request answered, verbatim:
+ *
+ *     HTTP 500
+ *     {"success":false,"error":{"code":"INTERNAL_ERROR",
+ *      "message":"SQLITE_ERROR: no such table: sys_metadata"}}
+ *
+ * ## Reverse verification, direction predicted BEFORE running
+ *
+ * Deleting the two-line withhold in `sendThrownError` turns the section-1 and
+ * section-2 leak cases RED — they assert the positive sanitized shape, so the
+ * driver line reappears in the diff — and leaves every pass-through case
+ * (section 3) and every 4xx case (section 4) GREEN, because the predicate is
+ * what decides and neither of those trips it. That is the ordinary direction,
+ * and it was confirmed by running it (quoted in the PR).
+ *
+ * ## What is deliberately NOT asserted
+ *
+ * That the message merely "changed", or that it "no longer contains the table
+ * name". Both pass for any rewrite, including a worse one. Every case below
+ * asserts the POSITIVE shape — `INTERNAL_ERROR_MESSAGE` — plus the full
+ * ADR-0112 envelope (`code` AND `status`), because one alone is not an answer:
+ * a status without the code leaves the client unable to branch, a code without
+ * the status makes every proxy and retry policy read a refusal as a fault.
+ */
+
+import { describe, it, expect, vi } from 'vitest';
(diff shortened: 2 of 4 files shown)


hotlong:

### Step 1 did its job, and what it caught was my ruling being stale

The card was marked Unverified and I made reproducing the leak a precondition rather than a formality. It paid immediately, and not in the direction I expected:

> The card pointed at `Failed to persist customization overlay to sys_metadata: ...` — that branch **no longer exists in `protocol.ts`**, its code deleted by #5264; it survives only as prose in a docblock.

My ruling repeated that producer as fact. Anyone who had "just fixed it" against the named producer would have been sanitizing a path nothing traverses — a filter over an unreachable leak, which is the exact shape this lane spends its days deleting. The real carrier is one level up and is a genuinely better find: a full uninstall routes to `protocol.deletePackage`, whose **first** database touch — `this.engine.find('sys_metadata', { where })` — sits **outside** that method's per-item `try`, whose `catch` only wraps the `deleteMetaItem` loop below it. So the driver error propagates whole instead of being collected into `failed[]`.

And it was measured, not argued: a real `ObjectQL` engine, a real protocol implementation, a driver failing `sys_metadata`, and the verbatim body off the wire —

```
HTTP 500  { "error": { "code": "INTERNAL_ERROR", "message": "SQLITE_ERROR: no such table: sys_metadata" } }
```

That is the third time today a dispatch's most specific sentence was the wrong one, and the second time the stale detail was **mine**. The instruction to verify rather than inherit is now carrying more weight than the leads themselves.

### The pins reason about what would make them vacuous

Two of the 50 cases exist because the developer asked what a *weaker* suite would still pass:

- **Pass-through of a non-leaky 5xx** — *"Without this, the whole file is satisfied by a blanket `if (status >= 500)`"*, which is a different and worse rule that would delete every self-authored server-fault sentence. Exactly right: without that case, "sanitize everything" and "sanitize leaks" are indistinguishable to the suite.
- **The 4xx over-block guard, chosen to trip the predicate on purpose** — `[tenant_scope_required]` prose naming a foreign key, a `409 DESTRUCTIVE_CHANGE` naming a unique constraint — so that **only the status** keeps them intact. A 4xx case that did not trip the predicate would prove nothing about the status condition. Plus the bound from both sides: 499 verbatim, 500 withheld.

And the live case carries a **premise guard** asserting the protocol really does still throw the driver line, so the section cannot silently start passing over a path nothing traverses.

Reverse verification is accounted arithmetically rather than gestured at: **23 failed / 27 passed**, and the 23 are enumerated in advance — `1 live + (4 sites × 5 leaks) + the 500 half of the bound case + the details case` — with all 12 pass-through and all 12 over-block cases green. Being able to predict *which* tests fail, and match the count, is a much stronger claim than "it went red."

### The known ceiling — accepted, and here is the binding reading

> `looksLikeInternalErrorLeak` recognises **no** Postgres `relation "..." does not exist` phrasing — measured false. That dialect's line still travels, through this door **and** the dispatcher twin alike, since both run the same predicate.

Reporting a gap in the thing you just shipped, in the PR that ships it, is the behaviour this seat wants and rarely gets. The scoping is also correct: widening the predicate *here* would be a new rule at one door and would re-create the very divergence this card closes. The two doors still agree; what remains is a property of the shared heuristic, and its cure is option C.

⛔ **Binding reading of that pinned case, for whoever meets it next:** it **documents a known gap**, it does not pin desired behaviour. When it goes red, the gap has closed — either the predicate learned the phrasing or C landed. **Delete or invert it then. Do not "repair" it back to green.** This is the #7314 shape (a pin that looks like a mistake and is actually load-bearing), and it is the failure mode most likely to erase this finding.

Scope held exactly: `response-envelope.ts` untouched (option A, escalated, not ruled), `metadata-protocol` untouched (option C, separate card), message only — no status, no `error.code`, `details` intact, 4xx untouched. The predicate and replacement constant are the siblings' byte for byte, so this is the existing rule reaching the door #5437 / PR #5464 missed, not a new one.

Also caught without being asked: `runtime/package-door-error-parity.test.ts`'s docblock asserted *"the REST package door applies no such filter and ships the thrown message verbatim"* — which this change makes false. Corrected in place rather than left as a comment that had quietly become a lie.

Not raised: `@objectstack/rest` measured at its recorded **155** and absent from the "can be lowered" list — zero new tsc errors against the repo's only zero-margin ceiling. `check:route-envelope` green with `package-routes.ts` still at its `0/0/0` ratchet. 101 files / 1698 tests pass.

---

Issue #7912, opened 2026-08-12
Nav gating cannot express "prune when the destination cannot serve" — `enable.apiEnabled` is never consulted by `filterAppForUser`


huangyiirene:
Split out of #7544 (fixed in #7909) by the `domain:metadata` seat. #7909 removed the one dead entry; **this is the general gap it sat on**, and it is a contract-face change ⇒ `domain:spec`.

## Symptom

A Setup/app nav entry whose destination object cannot serve the request is shipped to the client in the `/meta` payload anyway. The user sees a menu item that, when clicked, cannot work — and today the console renders that failure as a generic empty state, so it reads as *"you have no records"* rather than *"this page cannot work"* (that half is an `objectui` surface — see the companion card linked below).

## Root cause — measured on `origin/main`, twice, independently

`filterAppForUser` (`packages/rest/src/rest-server.ts:2976-3056`) is the server-side nav filter. It gates:

- `_unpublished` (the machine-managed key, ⛔ not `hidden`)
- app-level and item-level `requiredPermissions`
- `requiresService` (ADR-0057 D10)
- empty-group collapse (#7380)

Its own docblock names what it does **not** gate:

> NOT gated here: `visible` (CEL) at any level, and `requiresObject` — both are still evaluated client-side only. That asymmetry is deliberate and pinned in `rest.test.ts`.

**Nothing in nav filtering consults `enable.apiEnabled` at all.** Two measurements, run separately by the #7544 dev and by the dispatching PM:

| Probe | Result |
|:--|:--|
| `apiEnabled` in `rest-server.ts` | **9** occurrences — `apiAccessDenialFromEnable` (`:1665`), the `enforceApiAccess` docblock (`:2468-2505`), the bulk path (`:10128`, `:10177`, `:10423`) |
| …of those, inside `filterAppForUser` (`:2976-3056`) | **0** |
| `requiresObject` in `rest-server.ts` | **1** — the docblock line at `:3011` saying it is *not* gated there |

So `requiresObject`, the key that *looks* like the right tool, never reaches the server-side filter.

## ⭐ Why this is not "just add a gate"

Two independent conditions decide whether a destination can serve, and **neither is expressible on a nav entry today**:

1. **`enable.apiEnabled: false`** → `OBJECT_API_DISABLED` (404). `apiAccessDenialFromEnable` (`:1658`) is a **pure function of the object's `enable` block** — it takes no user, no permissions and no context. The denial is identical for every persona, platform admin included.
[...]

PR #8135, opened 2026-08-12
fix(rest): meta app by-name answers a permission-denied envelope, not absence (#8013)

### LINKED ISSUE #8013: meta apps by-name route: answer an explicit permission-denied envelope for an unauthorized session, instead of being indistinguishable from "not published" (backend half of objectui#4252)
Filed by the triage seat implementing a maintainer ruling (2026-08-12, live PM session, verbatim: 「接受你的全部建议。」 on the decision digest covering objectui#4252). Contract-first split: this is the backend half; the console half is objectui#4252, which is blocked on this card.

## Problem (measured in objectui#4252)

`GET /api/v1/meta/apps` omits apps the session isn't authorized for (`filterAppForUser`, `packages/rest/src/rest-server.ts`). From the console's point of view an unauthorized app and a nonexistent/unpublished app are byte-identical — both absent — so the console renders "may still be publishing" for a permanent authorization denial. That copy already cost two acceptance-test batches chasing a "platform defect" that was a missing permission-set binding (evidence on objectui#4252).

## Ruled scope

- **By-name app route**: when the app exists but the session lacks its `requiredPermissions`, answer an explicit permission-denied envelope (`code` + `status`, standard error shape) instead of the absence/404-equivalent.
- **List route (`/meta/apps`) stays filtered as-is** — deliberately unchanged, to avoid widening the enumeration surface beyond what a direct by-name probe already implies.
- The console (objectui#4252) then distinguishes: absent ⇒ current "not published" copy; denied envelope ⇒ plain "you don't have access to this app".

## Acceptance

1. Session WITHOUT the capability, by-name request ⇒ permission-denied envelope (assert `code` + `status`, not just non-200).
2. Session WITH the capability ⇒ app unchanged.
3. Nonexistent app name ⇒ existing absence behavior unchanged.
4. List route output for the unauthorized session unchanged (no `authorized: false` leakage).

## Landing

`packages/rest/src/rest-server.ts` (`filterAppForUser` and the by-name meta app handler) ⇒ `domain:cli`. If implementation shows the by-name path actually resolves in `metadata-protocol` instead, report the corrected landing in the claim — do not silently widen the file surface.

## Links

- objectui half (blocked on this): objectstack-ai/objectui#4252
- Related: #7912 (`filterAppForUser` never consults `enable.apiEnabled` — separate card, do not fold in)

hotlong:
Fixes #8013

Backend half of the ruling of 2026-08-12 (「接受你的全部建议。」). The console half is objectstack-ai/objectui#4252 and is blocked on this.

## The code to branch on

**`code: "PERMISSION_DENIED"` · `status: 403`**, in the declared envelope:

```json
{ "success": false, "error": { "code": "PERMISSION_DENIED", "message": "You do not have permission to open the 'finance' app." } }
```

`PERMISSION_DENIED` is the ADR-0112 **standard catalog** member for a generic authorization refusal — the ledger's own rule is that a generic condition takes the catalog rather than registering a synonym — and it is what `standardErrorCodeForHttpStatus(403)` answers. objectui#4252 reads `body.error.code`, which is the **same accessor** as the neighbouring absence answer, so the console needs one access path and not two.

## What changed

`GET /api/v1/meta/app/{name}` collapsed three different refusals into one 404-equivalent, so an app the session may never open and an app that does not exist were byte-identical on the wire. The console has nothing to branch on, so it renders its only copy for an absent app — "it may still be publishing" — over a permanent authorization denial. Measured cost on objectui#4252: two acceptance-test batches spent chasing a "platform defect" that was a missing permission-set binding.

`filterAppForUser` returned a bare `null` for all three. It now delegates to `filterAppForUserWithReason`, which reports **which** gate fired; `filterAppForUser` is an unchanged thin wrapper over it, so the list route and the existing unit pins are untouched.

Exactly one of the three converts:

| gate | before | after |
|---|---|---|
| `requiredPermissions` not held | absence | **403 `PERMISSION_DENIED`** |
| `_unpublished` (ADR-0045 §3) | absence | absence — unchanged |
| `requiresService` absent (ADR-0057 D10) | absence | absence — unchanged |
| name resolves to nothing | absence | absence — unchanged |
| list route `GET /meta/apps` | filtered | filtered — unchanged |

The reason comes from the branch that fired, never inferred from `app == null` at the call site. That distinction is the security boundary of the card:

- `_unpublished` stays absent because ADR-0045 §3 makes an unpublished app *externally unobservable*, and a 403 confirms existence. `meta-app-publish-gate.test.ts` has pinned the 404-over-403 choice since #4829.
- `requiresService` stays absent because an absent optional service is a deployment fact about the platform, not a statement about this caller — nothing was denied *to the session*.
- A **nonexistent name** stays absent because converting it would make every app name on the platform enumerable. That is a different and unruled change.

Ordering is load-bearing for the same reason: `_unpublished` is judged first, so an app that is both unpublished and permission-gated reports absence.

## Landing — confirmed as the card predicted

`packages/rest/src/rest-server.ts`, located by symbol: `filterAppForUser` and the by-name handler's app branch. The card asked to report a corrected landing if the by-name path resolved in `metadata-protocol` instead — **it does not**. The gate is re-derived in `rest-server.ts`, and the cached read path is excluded for `isAppType`, so every app-typed by-name request already takes the uncached branch that holds this gate. No `metadata-protocol` change, and no second file edited "to be safe".

Only the meta-apps region is touched, so this does not collide with #8071 (`/security/suggested-bindings`) in the same file. Rebased onto `main` before push.

## Acceptance criteria

All four are pinned in `packages/rest/src/meta-app-publish-gate.test.ts`, which already owned this route's response-body facts. Every case asserts `status` **and** `code` (ADR-0112) — never "an error came back", since both answers under test are errors one apart.

1. Session without the capability ⇒ 403 + `PERMISSION_DENIED`, no document and no leaked object names.
2. Session with the capability ⇒ app served unchanged, navigation included.
3. Nonexistent name ⇒ absence, in **both** shapes that reach this route: the stub's `undefined`, and the producer's declared `RESOURCE_NOT_FOUND` / 404 rejection that `metadata-protocol` actually throws.
4. List route read and asserted **absent** for the unauthorized session — not "the endpoint still 200s". No `authorized` key, no `PERMISSION_DENIED`, no trace of the withheld app; and the holder still receives it.

Plus two partition pins: unpublished-and-permission-gated stays 404, and absent-service stays 404.

## Reverse verification

Five reversals, each confirming the pin it targets goes red for the expected reason:

| reversal | pin that reddened | failure text |
|---|---|---|
| R1 — restore the pre-fix single 404 | criterion 1 | `expected 404 to be 403` |
| R2 — drop the `visible` guard, infer the reason from `null` | criterion 3 | `expected 403 not to be 403` |
| R2b — map the producer's miss onto a denial | criterion 3 (production shape) | `expected 403 to be 404` |
| R3 — list route annotates `authorized: false` instead of dropping | criterion 4 | `expected [ 'account', 'crm', 'finance', …(1) ] to not include 'finance'` |
| R4 — deny the capability holder too | criterion 2 | `expected 403 to be 200` |
| R5 — naive "any `null` ⇒ 403" | criterion 3 + both partition pins + the pre-existing #4829 pin | `expected 403 to be 404` ×3, `expected 403 not to be 403` |

R5 is the one that matters most: the naive implementation reddens **four** pins at once, including #4829's own `SINGLE ITEM: the unpublished app 404s for a non-builder`.

Criteria 2, 3 and 4 stay green under R1, and that is the correct direction rather than a gap — they pin behavior the fix deliberately leaves *unchanged*, so removing the fix cannot break them. Each has its own reversal above (R4, R2/R2b, R3).

Restored afterwards and confirmed byte-identical (`git diff` matches the saved patch exactly).

## Gates

All run on the **rebased** tree (post-#8071, post-#8088), with the dependency closure built first.

- `pnpm --filter @objectstack/rest test` — **101 files, 1678 tests, all passing**.
- `pnpm --filter @objectstack/rest typecheck` — clean.
- `pnpm check:type-check-debt` — OK, "none above its recorded number"; `@objectstack/rest` stays at its recorded **155** and is absent from the surplus list, i.e. measured exactly at the ceiling. The ledger is not raised.
- `pnpm check:route-envelope` — green including `--self-test`; `rest-server.ts` dialect counts unmoved at `stringError 44 / siblingCode 75` (75 is #8071's freshly-lowered baseline, which this change does not disturb). The refusal is written through the **shared** `sendError` (`@objectstack/types`), imported as `sendEnvelopeError` because this module has a local `sendError` of its own — the sanitizing responder for *thrown* errors, which is a different thing. The body literal lives in the pinned `SHARED_BUILDER`, so no write site is added here.
- `node scripts/check-nul-bytes.mjs` — OK, plus a wider self-scan of the changed files.

⛔ Out of scope, as the card's Links section requires: #7912 (`filterAppForUser` never consults `enable.apiEnabled`) is **not** folded in.

---

### DIFF
--- a/.changeset/meta-app-by-name-denied-envelope.md
+++ b/.changeset/meta-app-by-name-denied-envelope.md
@@ -1,0 +1,29 @@
+---
+'@objectstack/rest': patch
+---
+
+`GET /api/v1/meta/app/<name>`: report a permission denial instead of absence
+
+An app the session lacks the `requiredPermissions` for used to answer the same
+404-equivalent as an app that does not exist, so the two were byte-identical on
+the wire. A console has nothing to branch on and renders its only copy for an
+absent app — "it may still be publishing" — over a permanent authorization
+denial.
+
+The by-name route now answers `403` with the ADR-0112 standard catalog code
+`PERMISSION_DENIED`, in the declared envelope
+(`{ success: false, error: { code, message } }`), when the app EXISTS and the
+session lacks its `requiredPermissions`.
+
+Deliberately unchanged, because the disclosure is licensed only for the case
+above:
+
+- a **nonexistent** app name keeps answering absence — converting it too would
+  make every app name on the platform enumerable;
+- an **unpublished** app keeps answering `404` (ADR-0045 §3 makes it externally
+  unobservable, and a 403 confirms existence);
+- an app withheld by an absent optional service (ADR-0057 D10) keeps answering
+  `404` — nothing was denied to the caller;
+- the **list** route `GET /meta/apps` stays filtered exactly as before, with no
+  `authorized: false` flag, so the enumeration surface is not widened past what a
+  direct by-name probe already implies.
--- a/packages/rest/src/meta-app-publish-gate.test.ts
+++ b/packages/rest/src/meta-app-publish-gate.test.ts
@@ -54,7 +54,34 @@ const CRM_APP = {
     navigation: [{ id: 'nav_leads', type: 'object', objectName: 'lead' }],
 };
 
+/**
+ * [#8013] A PUBLISHED app the session may only open with a capability it might
+ * not hold. Its `requiredPermissions` is a different layer from ADR-0045 §3's
+ * publish gate above, and the ONLY one this card converts into a denial.
+ */
+const FINANCE_APP = {
+    name: 'finance',
+    label: 'Finance',
+    requiredPermissions: ['finance.access'],
+    navigation: [{ id: 'nav_invoices', type: 'object', objectName: 'invoice' }],
+};
+
+/**
+ * [#8013] A published app gated by ADR-0057 D10 capability presence rather than
+ * by anything about the caller. Kept out of `ALL_APPS` for the same reason
+ * `FINANCE_APP` is: the #4829 / #7566 suites above pin exact name lists.
+ */
+const OPTIONAL_SERVICE_APP = {
+    name: 'telephony',
+    label: 'Telephony',
+    requiresService: 'voice',
+    navigation: [{ id: 'nav_calls', type: 'object', objectName: 'call_log' }],
+};
+
 const ALL_APPS = [ACCOUNT_APP, UNPUBLISHED_APP, CRM_APP];
+
+/** [#8013] `ALL_APPS` plus the two extra gate shapes this card partitions. */
+const GATED_APPS = [ACCOUNT_APP, UNPUBLISHED_APP, CRM_APP, FINANCE_APP, OPTIONAL_SERVICE_APP];
 
 function createMockServer() {
     return {
@@ -71,18 +98,22 @@ function makeRes() {
     return res;
 }
 
-/** @param perms system permissions the caller holds */
-function setup(perms: string[]) {
+/**
+ * @param perms system permissions the caller holds
+ * @param apps  the app corpus this server serves. Defaults to `ALL_APPS` so the
+ *   #4829 / #7566 suites keep their exact name lists; #8013 passes `GATED_APPS`.
+ */
+function setup(perms: string[], apps: any[] = ALL_APPS, serviceExists?: (name: string) => boolean) {
     const protocol: any = {
         getDiscovery: vi.fn().mockResolvedValue({ version: 'v0', routes: { data: '', metadata: '', ui: '', auth: '/auth' } }),
         getMetaTypes: vi.fn().mockResolvedValue([]),
         // Deep-clone per call: the filter must never mutate stored metadata.
         getMetaItems: vi.fn(async ({ type }: any) => {
             const t = String(type ?? '');
-            return t === 'app' || t === 'apps' ? JSON.parse(JSON.stringify(ALL_APPS)) : [];
+            return t === 'app' || t === 'apps' ? JSON.parse(JSON.stringify(apps)) : [];
         }),
         getMetaItem: vi.fn(async ({ name }: any) => {
-            const found = ALL_APPS.find((a) => a.name === name);
+            const found = apps.find((a: any) => a.name === name);
             return found ? { type: 'app', name, item: JSON.parse(JSON.stringify(found)) } : undefined;
         }),
         findData: vi.fn().mockResolvedValue([]),
@@ -91,6 +122,12 @@ function setup(perms: string[]) {
     // The RBAC filter only runs for a resolved caller; stubbing the context is
     // the established pattern in this package for exercising it by route.
     rest.resolveExecCtx = async () => ({ userId: 'u1', systemPermissions: perms });
+    // [#8013] ADR-0057 D10's gate FAILS OPEN when it cannot be probed, and the
+    // stubbed context carries no `__kernel`, so without a provider a
+    // `requiresService` app is simply served and the `service` branch never
+    // runs. Assigned rather than threaded through the constructor's 16th
+    // positional parameter.
+    if (serviceExists) rest.serviceExistsProvider = serviceExists;
     rest.registerRoutes();
     return { rest, protocol };
 }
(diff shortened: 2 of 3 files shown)


hotlong:

### Exactly one refusal converts, and each of the other four is refused on its own reasoning

This is the card's whole risk. Making an app's *existence* observable to a caller who may not open it is a deliberate, ruled disclosure — and it is licensed **only** for the by-name route on an app that exists and whose `requiredPermissions` the session lacks. The table holds that line, and every non-conversion is argued rather than left alone by default:

- **`_unpublished` stays absent** — ADR-0045 §3 makes an unpublished app *externally unobservable*, and a 403 confirms existence. #4829 has pinned the 404-over-403 choice since it landed.
- **`requiresService` absent stays absent** — an absent optional service is a deployment fact about the platform, not a statement about this caller. Nothing was denied *to the session*.
- **A nonexistent name stays absent** — converting it would make every app name on the platform enumerable. Different change, unruled.
- **The list route stays filtered** — no `authorized` key, no widening of the enumeration surface.

And the ordering is called out as load-bearing for the right reason: `_unpublished` is judged **first**, so an app that is both unpublished and permission-gated reports absence. Get that order backwards and the unpublished gate leaks through the new door.

The structural point underneath it all:

> The reason comes from the branch that fired, **never inferred from `app == null` at the call site.**

That is what makes the partition real rather than incidental, and it is exactly what R5 proves.

### R5 is the reversal that earns the suite

Five reversals, each aimed at a named pin — but **R5, the naive "any `null` ⇒ 403"**, is the implementation a hurried developer would actually write, and it reddens **four** pins at once, including #4829's own pre-existing `SINGLE ITEM: the unpublished app 404s for a non-builder`. A suite that only tested "unauthorized gets 403" would have passed that implementation and shipped the enumeration hole.

Criterion 3 is pinned in **both** shapes that reach this route — the stub's `undefined` *and* the producer's declared `RESOURCE_NOT_FOUND`/404 rejection that `metadata-protocol` actually throws. A single-shape pin would have covered the test double and missed production.

Criterion 4 is asserted the way I asked: the list response is **read and the app asserted absent** for the unauthorized session, not "the endpoint still 200s."

And criteria 2/3/4 staying green under R1 is explained rather than hidden — *"they pin behavior the fix deliberately leaves unchanged, so removing the fix cannot break them"* — with each carrying its own targeted reversal (R4, R2/R2b, R3). Reporting the greens and naming the property is the discipline; a suite where every pin reddens under one reversal is usually a suite testing one thing.

### The landing question was answered, not assumed

The card asked you to report a corrected landing if the by-name path resolved in `metadata-protocol`. **It does not** — the gate is re-derived in `rest-server.ts`, and the cached read path is excluded for `isAppType`, so every app-typed by-name request already takes the uncached branch holding this gate. Reported with the reason, **no `metadata-protocol` change, and no second file edited "to be safe."** That last part matters: editing both places to be certain is how a file surface silently doubles.

The `sendEnvelopeError` import is a good catch — this module has its own local `sendError` (the sanitizing responder for *thrown* errors, a different thing), so aliasing the shared writer avoids a collision that would have been confusing rather than broken. Envelope counts unmoved at `stringError 44 / siblingCode 75`, correctly noting 75 is #8071's freshly-lowered baseline that you rebased onto.

### For objectui#4252

The code to branch on is **`error.code: "PERMISSION_DENIED"` / `status: 403`**, and the choice is right: it is the ADR-0112 standard-catalog member for a generic authorization refusal rather than a registered synonym, it is what `standardErrorCodeForHttpStatus(403)` answers, and — the part that matters for the console — it sits at `body.error.code`, the **same accessor** as the neighbouring absence answer. One access path, not two. objectui#4252 is unblocked the moment this merges.

Not raised: `@objectstack/rest` measured exactly at its recorded **155** and absent from the surplus list — the repo's only zero-margin ceiling, held. 101 files / 1678 tests pass. #7912 correctly not folded in.

---

Issue #8134, opened 2026-08-12
#7804's `GlobalFilterSchema.object` is unreachable in the Studio designer, and no console pin bump can fix it — the console vendors the PUBLISHED `@objectstack/spec`, which the key postdates


hotlong:
Blocked-by: objectstack-ai/objectui#4854 (the `apps/console/vite.config.ts` hook the ruled mechanism needs — measured absent repo-wide and at the pinned SHA, 2026-08-16)

Filing unassigned, as the measured residue of #8051. #8051 asked for a console pin bump and a re-check of two symptoms; the bump landed (PR for `claude/issue-8051-console-pin-bump`) and fixed symptom 1. **Symptom 2 survives the bump**, and the reason is not pin staleness — so it needs its own card rather than another bump.

## What was measured, after the pin moved

Pin moved `6314e87f2d49` → `6d77acfe3125` (objectui `main`, 123 commits). Console rebuilt via `pnpm objectui:refresh`; `pnpm check:console-sha` and `pnpm check:objectui-pin-fresh` both green.

`GlobalFilterSchema` in the **newly built** `packages/console/dist/assets/vendor-objectstack-DneBps14.js` still declares:

```
name, field, label, type, options, optionsFrom, defaultValue, scope, targetWidgets
```

`object` is absent from exactly the slot #7804 put it in (between `field` and `label`), and #7804's own `describe()` text — `translation-bundle entry resolves this filter` — matches **0 times** in the bundle. So a dashboard carrying `globalFilters[].object` still trips the designer's *"Unrecognized key(s) on this global filter: `object`"* banner, and auto-save still never fires while it stands.

Note `optionsFrom.object` **does** appear in the bundle and is a different key — a bare grep for `object` false-positives here.

## Why a pin bump cannot fix it

The chain is a publish-ordering dependency, not a staleness one:

1. objectui pins `"@objectstack/spec": "^17.0.0-rc.6"` at both the old and the new pin, and its lockfile resolves it to exactly `17.0.0-rc.6`. `scripts/build-console.sh` runs `pnpm install --frozen-lockfile`, so it does not float.
[...]

Issue #7804, opened 2026-08-11
spec: `GlobalFilterSchema` needs an optional `object` — a dashboard global filter's field label (and its option labels) have no bundle key to resolve against


yinlianghui:
Contract-first card filed by the objectui whole-repo PM seat (session `session_017Qqyix2QcnpUC9XeYVDzx3`), spun out of objectui#4030 / PR objectui#4324. Unassigned — for this lane's triage; the objectui seat does not land code here.

## Problem (measured in objectui#4324's implementation)

A dashboard global filter renders its field label ("Sales Channel:") and its option labels untranslated, and there is no key to fix it with:

- `GlobalFilterSchema` declares no `object`, and `DashboardSchema` declares none — so the canonical field-label convention `fields.<object>.<field>` has **no key to build**.
- The spec's `TranslationData.dashboards.<name>` node is STRICT with only `label`/`description`/`actions`/`widgets` — `dashboards.<d>.filters.<f>.label` cannot be authored without new vocabulary.
- Deriving the object from bound widgets only works for `provider: 'object'` widgets; a dataset widget's object arrives with the query response — exactly the dashboards the downstream report is about — so derivation would be a guess that can resolve against the WRONG object and show a confidently incorrect translation.

## Recommendation (from the implementing seat — option A of three considered)

**Add optional `object` to `GlobalFilterSchema`** (author-declared). The filter then keys the SAME `fields.<object>.<field>` convention lists and forms use — zero new i18n vocabulary, one resolver, and `optionsFrom.object` already proves the schema is willing to name an object. An author who omits it gets today's untranslated label rather than a silently wrong key, and the key stays checkable at publish time by `os lint`'s translation-target-unknown, exactly as `_views` keys are. The same missing key blocks the filter's OPTION labels, so one field unblocks both.

Rejected alternatives, recorded: **B** — a `filters` node in dashboard TranslationData creates a SECOND vocabulary for a field label (two keys, no precedence rule — the dialect shape objectui's one-channel ruling exists to prevent); **C** — the inline `I18nLabelSchema` form is orthogonal (objectui drops it today; tracked separately on objectui#4163) and does not serve bundle-side translators.

## Downstream evidence

hotcrm-heimao hit this on a shipped dashboard (same report as objectui#4030's option-label evidence). objectui's renderer half is ready to consume the key the moment it exists — the resolver and seam landed in objectui PR #4324.

[...]

Issue #7411, opened 2026-08-10
chore: bump the console pin — 18 objectui commits behind, including the objectui#4096 row-action permission gate


baozhoutao:
Linkage chore per the multi-repo coordination rule (accepted `repo:objectui` PR ⇒ file the console bump in the consuming repo).

objectstack-ai/objectui#4108 is **merged** (objectui main `aeb8424`, 2026-08-10T10:24:11Z): the list row kebab's built-in Edit/Delete, the grid bulk-delete bar, the non-grid (`ListView`) bulk delete and the related-list Create/Edit/Delete now AND the caller's own `/me/permissions` verdict (`allowEdit` / `allowDelete`) on top of the layers they already had. Fixes objectui#4096.

## Task

Run the standard console refresh (`scripts/bump-objectui.sh` / `pnpm objectui:refresh` per repo convention) and land the `.objectui-sha` + `packages/console` bump.

## Staleness reading (2026-08-10T10:30Z)

- `.objectui-sha` on `origin/main`: `8aad9fd50b16750aebdd294392ff79540f17cd32` (objectui `fix(components): bind the row three ways on the four action renderers (#4075)`)
- objectui `origin/main`: `aeb8424`
- **18 commits behind**, spanning the objectui#4096 gate above, the `evalRowPredicate` fault-reason/subject fix (objectui#3792 / #3796), the `clientValidation` wiring (objectui#3561), the object-grid `filter` key fix (objectui#4041) and a dependabot wave.

## Why it matters beyond routine freshness

objectui#4096 is a **user-facing permission-affordance repair on a destructive control** — Delete rendered one click away from accounts the server answers `403` on. Server-side enforcement was already hard (403, DB unchanged), so nothing is unsafe today, but the bundled console keeps showing the misleading entry until this bump lands.

Standard pin-bump discipline applies: dedicated PR, no riders; `.objectui-sha` is a shared file — take the whole regen (hono override + lockfile) through the script.

Note for triage: no `domain:*` / `repo:*` label applied — routing labels are the triage seat's single-producer territory. Only `pm:queue` is set here. Single-card cap checked at filing time: #7268 (the previous console-bump card) is **closed** (2026-08-10T06:45Z) and no other open console-bump card exists.

huangyiirene:
Claim: PM loop round 1 (devx seat, 2026-08-11 term)
Session: `session_01GZKbx4xyF7U5WXj6ch49BM`
Branch: `claude/issue-7411-console-bump`
[...]

PR #8137, opened 2026-08-12
chore: bump the console pin to objectui 6d77acfe3125 (123 commits, not 405)

### LINKED ISSUE #8051: chore: bump the console pin — 405 objectui commits behind, and the gap now holds a shipped bug fix plus the spec key #7804 added
The vendored console pin is `6314e87f2d49` (objectui, **2026-08-11 04:17**). `origin/main` is `b4d3c2204` — **405 commits ahead**. Two concrete consequences measured on the current framework baseline (`279ee48a`), not a routine hygiene ask:

## 1. A merged objectui fix is not being served, and QA reads it as still-broken

objectui#4229 (`combo` dashboard widgets render as grouped bars) was **fixed** by merged PR objectui#4298 on 2026-08-11 12:12 — **eight hours after this pin was cut**. Verified directly:

```
git show 6314e87f2d49:packages/plugin-dashboard/src/DatasetWidget.tsx | grep combo   → (nothing)
git show origin/main:packages/plugin-dashboard/src/DatasetWidget.tsx  | grep combo   → combo: 'combo'  (line 359)
```

So the showcase chart gallery's "Task Count vs Avg Progress" still renders two bar series on one axis (`.recharts-line-curve` = 0, `.recharts-yAxis` = 1) against a build where the fix exists upstream. This already cost one retest cycle a false "still broken" verdict.

## 2. #7804's new authorable key cannot be opened in the Studio designer

`7dc10673` (#7804) gave `GlobalFilterSchema` an optional `object` for i18n label resolution. The **server accepts and round-trips it** — a dashboard whose `globalFilters[].object` is set PUTs 200 and reads back verbatim, and `packages/spec`'s own pins are green (59/59 on `src/ui/dashboard.test.ts` including the #7804 block).

But the pinned console refuses it. Opening such a dashboard in the Studio designer shows:

> This metadata does not match the spec — 1 validation error(s). Global Filters → st: Unrecognized key(s) on this global filter: `object`

and the designer's auto-save never fires while that banner stands, so **the dashboard becomes unopenable-for-editing**. Cause is a spec-version skew, not a product defect: the console bundle's `GlobalFilterSchema` (extracted from `packages/console/dist/assets/vendor-objectstack-*.js`, 4 identical copies) declares `name`/`field`/`label`/`type`/`options`/`optionsFrom`/`defaultValue`/`scope`/`targetWidgets` and no `object`. objectui pins `@objectstack/spec` `^17.0.0-rc.6` at this pin.

**So #7804 shipped an authorable key that no author can use through the UI until the console is rebuilt against a spec that carries it.** That is worth naming explicitly: the framework-side card is closed and green, and the feature is still not reachable.

## Ask

Bump `.objectui-sha` to a current objectui commit and rebuild `packages/console/dist`, then re-check both of the above. Same shape as #7411 ("18 objectui commits behind"), an order of magnitude larger.

Worth considering alongside: a gate that fails when the pin falls more than N commits or M days behind would have caught both of these before QA did — right now the only signal is `scripts/check-console-sha.mjs`, which verifies dist matches the pin and says nothing about whether the pin matches reality.

## Source

Measured during the platform checklist retest sweep (framework `279ee48a`, console pin `6314e87f2d49`).

hotlong:
Fixes #8051

Bumps the vendored console pin via the dedicated bump path and rebuilds `packages/console/dist`, then re-checks both symptoms the card named. Nothing here is hand-edited: `pnpm objectui:refresh` (`scripts/bump-objectui.sh` + `scripts/build-console.sh`) wrote every tracked change except the ADR-0087 disposition the bump script explicitly requires a human to answer.

**Pin:** `6314e87f2d49b1ff3b158c296f1b2a52d14dff68` → `6d77acfe3125949796918ea12b926d3460bc933d` (objectui `main`).

## The gap is 123 commits, not 405

The card measured "405 commits ahead" and the dispatch expected the re-measurement to come out **larger**. It comes out **smaller**, and three independent measurements agree on 123:

| measurement | result |
| --- | --- |
| `git rev-list --count OLD..HEAD` in a full objectui clone | 123 |
| same with `--first-parent` | 123 |
| `check:objectui-pin-fresh`'s own log-walk | "123 non-merge commit(s)" |

The card's own waypoint checks out consistently: `b4d3c2204` (the tip it measured) is an ancestor of the new pin, `OLD..b4d3c2204` is 121 and `b4d3c2204..HEAD` is 2. So 405 is not reproducible from any of these; the pin was 121 commits behind when the card was filed, not 405. Flagging it because the card's framing — "an order of magnitude larger" than #7411's 18 — rests on that number, and the real multiple is closer to 7x than 22x. The staleness itself is real either way.

Range composition, from the digest: 99 releasing changesets (0 major / 22 minor / 77 patch) out of 115 added, across 123 non-merge commits; 16 release-nothing, 11 commits carrying no changeset.

## Symptom 1 — objectui#4229's `combo` fix: FIXED

Confirmed at three levels rather than assumed:

- The fixing commit is in range: `3c6e84cca fix(plugin-dashboard): draw an authored combo as a combo on the dataset path (#4229) (#4298)`.
- Source at the new pin: `git show 6d77acfe3125:packages/plugin-dashboard/src/DatasetWidget.tsx | grep combo` returns `combo: 'combo'` at line 359 plus its rationale block; the same probe at the old pin returned nothing.
- **The rebuilt bundle actually serves it** — `assets/plugin-dashboard-sS6Qwuba.js` contains the resolved mapping `…sankey:\`sankey\`,combo:\`combo\`}` and the widget-kind set carrying `combo`.

## Symptom 2 — #7804's `GlobalFilterSchema.object`: NOT FIXED, and no pin bump can fix it

This is the finding the card did not anticipate. After the bump and rebuild, `GlobalFilterSchema` in the newly built `assets/vendor-objectstack-DneBps14.js` still declares exactly:

```
name, field, label, type, options, optionsFrom, defaultValue, scope, targetWidgets
```

`object` is absent from precisely the slot #7804 put it in (between `field` and `label`), and #7804's own `describe()` text — `translation-bundle entry resolves this filter` — matches **0 times** in the bundle. (`optionsFrom.object` does appear and is a different key; a bare grep for `object` false-positives here.)

The cause is publish ordering, not pin staleness:

1. objectui pins `"@objectstack/spec": "^17.0.0-rc.6"` at **both** the old and the new pin, and its lockfile resolves to exactly `17.0.0-rc.6`. `build-console.sh` installs `--frozen-lockfile`, so it does not float.
2. `@objectstack/spec@17.0.0-rc.6` was published to npm **2026-08-10T13:09:42Z**; #7804's key landed at `7dc10673` on **2026-08-12T11:36:21Z**, two days later. Verified against the tarball actually installed into the build tree: `.cache/objectui-6d77acfe3125/node_modules/@objectstack/spec` is `17.0.0-rc.6` with 0 matches for the key.
3. `17.0.0-rc.6` is the highest version on npm, and this repo's `packages/spec/package.json` is *also* `17.0.0-rc.6` — the workspace has moved past the published tarball under an unchanged version string, so even the `^` range cannot reach the new content.
4. `build-console.sh` injects only `OBJECTSTACK_CLIENT_DIST` (`packages/client`); there is no equivalent injection for `packages/spec`.

So the key becomes UI-authorable only after a spec release publishes it, then objectui relocks onto that version, then the pin moves again — three ordered steps across two repos, none performable here. Filed as **#8134** with the general shape (any authorable key added after the last spec publish is server-accepted and designer-rejected, while the framework-side card closes green). #8134 remains open and is not addressed by this PR.

## Gates

Both required gates are green, and they are different claims:

- `pnpm check:console-sha` — "Console dist matches the objectui pin (objectui@6d77acfe3125)."
- `pnpm check:objectui-pin-fresh` — "objectui pin is FRESH — .objectui-sha is objectui `main`". Red before this change (EXIT=1, 123 pending).

`node scripts/pm/dispatch-gates.mjs .objectui-sha .changeset/console-6d77acfe3125.md` surfaced three families beyond the dispatch list; all run and green: `check:changeset-gate-self-tests` (118 + 153 + 117 assertions), `check:objectui-changeset` (both self-tests), `check-changeset-no-major` ("introduces no `major` bump"). Also `check:nul-bytes` — OK over 7407 files.

### The one judgement call: the ADR-0087 disposition

The bump script emitted `adr-0087: TODO — the pin bump cannot answer this; a human must` and `check-adr-0087-registration` rejects that placeholder by design. The range declares 10 breaking changes, so the question had to be answered rather than deleted.

I judged the ten one by one against their upstream changesets. All are TypeScript declaration surfaces of objectui's own npm packages — `@object-ui/core`, `react`, `components`, `types`, `permissions`, `plugin-chatbot`, `data-objectstack` — and none is reachable through `@objectstack/console`, whose `files` list is `["dist", "README.md", "CHANGELOG.md"]` and whose sole `exports` entry is `./package.json`; it forwards no `@object-ui/*` module entry point. No `packages/spec` schema, authorable key or protocol surface changes in the range, so there is no stored-metadata rewrite to prescribe. Where a prescription exists it runs the other way — `ActionDef`'s unified `visible`/`disabled` shape and the 14 renamed symbols are objectui catching up to spec `17.0.0-rc.6`, whose spec-side halves were disposed of in this repo's own PRs. Disposition: `not-required (no-migration-prescription)`, matching the route the two most recent breaking console bumps took (`console-8aad9fd50b16.md`, `console-09987b680d53.md`).

## Scope notes

- `packages/console/dist` is gitignored (`packages/console/.gitignore` is `dist/`), so it is not in this diff by design — the diff is `.objectui-sha` plus the changeset. The dist was still built and verified locally; that is what the two gates above read.
- **No staleness gate is added here.** `check:objectui-pin-fresh` already exists, was already red, and already prescribed this exact fix; its absence from `cut-rc.yml` is deliberate and documented in that workflow's header (it enforces at the GA cut). A threshold change would be a separate card.
- `pnpm sdui:manifest` (ADR-0082 D4 declaration-parity ratchet) is the documented next step after a pin move and is **not** run here — it needs Playwright chromium and is an on-demand gate by decision (#5960), never CI. Flagging it so it is not silently assumed done.
- No release action taken: no Version Packages merge, no publish, no tag (#6170).

## Console section for the release page

Produced now because it is cheap here and expensive to reconstruct later. Regenerate verbatim with:

`node scripts/objectui-range.mjs --from 6314e87f2d49b1ff3b158c296f1b2a52d14dff68 --to 6d77acfe3125949796918ea12b926d3460bc933d`

(Positional args are **framework** refs; objectui SHAs need `--from`/`--to`.) `content/docs/releases/` is untouched — release notes are written centrally at release time.

One character is altered below to survive GitHub's body sanitizer: in the `lint` entry, `< button` carries an inserted space that the real output does not have. Regenerate with the command above rather than copying that line.

```markdown
!-- objectui 6314e87f2d49..6d77acfe3125 — 99 releasing changeset(s) of 115 added across 123 non-merge commit(s) --

_Largest areas: i18n (13), app-shell (13), console (8), components (6), fields (4), react (3), plugin-dashboard (3), plugin-chatbot (3), plugin-grid (3), detail (3)_

### Breaking changes
- **core** — `rowHeightToDensityMode` answers only for the five spec row heights — the coerce-to-`comfortable` fallback is gone (objectui `fe52a0406`)
- **react** — `bridgeListView` maps the five row heights the spec admits, and only those — the four dead spellings are gone (objectui `d7f3e308b`)
- **components** — The action renderers publish the modern `UIActionSchema`, and every `forwardRef` renderer's props parameter is annotated so its declared types survive (objectui `dc2aa3e77`)
- **sweep** — Retire two post-retirement dead surfaces (#4364, #4368). Both were measured at this branch point rather than taken from their cards, and one card's premise only half held. (objectui `2a40f699d`)
- **core** — Retire `ActionEngine`'s event-mapping API (objectui#3368). `ActionEngine.addMapping()`, `ActionEngine.dispatch()`, the private `mappings` registry behind them, and the exported `A… (objectui `2459a3e70`)
- **plugin-chatbot** — `@object-ui/plugin-chatbot`'s `ChatMessage` is now one type instead of two (objectui `3256b141f`)
- **sweep** — Retire four zero-consumer declared surfaces (dead-surface sweep batch 3, #4328). Each was measured as declared-but-never-read at the branch point, and each is removed rather than… (objectui `d9d346307`)
- **core,app-shell** — Close `ActionDef` — delete the `[key: string]: any` index signature and converge `visible` / `disabled` on the spec's unified shape. (objectui `ee66e2ebf`)
- **@object-ui/data-objectstack** — data-objectstack: retire the phantom `CloudOperations` surface — the class, its three `Cloud*` types, and the module that claimed to integrate a cloud namespace no client has ever… (objectui `2776b110b`)
- **i18n** — Stop declaring 14 symbols under names `@objectstack/spec` owns at `17.0.0-rc.6` (objectui#4167, objectstack#4115). (objectui `bb68488ff`)

### Features
- **plugin-dashboard** — `MetricWidgetProps` / `MetricCardProps` declare the DOM pass-through their spread has always accepted (objectui `f24427334`)
- **plugin-chatbot** — `useObjectChat` declares the message shape it actually hands back (objectui `eec2e4fd1`)
- **plugin-grid** — grid row menu — the built-in Edit/Delete predicate declarations are derived from the spec-owned authoring type, not hand-restated (objectui `24bb2de4f`)
- **i18n** — i18n: retire the reader-less `common.search` key from all ten locale packs (objectui `ac853ce92`)
- **app-shell** — fix(app-shell): a transient 404 no longer retires the shared inbox feed for the page's lifetime (objectui `ca269fecd`)
- **console** — console: seed the UI language from the tenant's server-side locale (objectui `78fa33141`)
- **plugin-grid** — ObjectGrid's host-driven pagination mode is a declared interface instead of twelve `(rest as any)` reads (objectui `51ac39f73`)
- **app-shell,console** — `resolveHostAppSegment` is published from `@object-ui/app-shell`'s package root, and the console's local copy of it is deleted (objectui `e4d1c085e`)
- **console** — An unloadable app list no longer lands you on `/home` as though you had no default app — and the Applications page's Set-as-default / Disable / Delete now actually write (objectui `3b7d1cc6d`)
- **console** — Internal `/forms/:name` renders inside the console shell, and an internal submit lands on the record it just created (objectui `90e792e11`)
- **app-shell** — The console's language menu now asks the app which locales it actually ships, instead of always offering the same ten. (objectui `66fb4fae2`)
- **i18n** — `element:record_picker` publishes `sort`, `limit` and `emptyText` as authoring inputs (objectui#4167). (objectui `bb68488ff`)

### Fixes
- **react** — fix(react): the spec bridge abstains on a non-string `rowHeight` instead of coercing it to a density, matching core (objectui `47f551b06`)
- **components** — `element:button`'s action forward is excess-property checked again — a misspelled key on that payload is now a compile error, not silence (objectui `b4d3c2204`)
- **plugin-grid** — standalone ObjectGrid resolves off-spec `rowHeight` to compact, matching ListView and the spec bridge, instead of silently styling it as medium (objectui `5e514c424`)
- **react** — The spec bridge abstains on prototype-member `rowHeight` spellings instead of leaking a function into `density`. (objectui `8f85f8bed`)
- **plugin-calendar** — fix(plugin-calendar): authoring `events` on a `calendar-view` node no longer takes the calendar down (objectui `49b9de6a6`)
- **lint** — Every plain `< button>` now declares its `type`. HTML defaults an untyped button to `type="submit"`, so any of these buttons would submit the form it was composed into instead of r… (objectui `d0c3b26ca`)
- **app-shell** — `AiChatPage` narrows the chat hook's messages through the exported `toRuntimeMessages` adapter instead of five casts (objectui `d8d0d665d`)
- **plugin-dashboard** — KPI cards no longer write their own schema onto the DOM — `MetricWidget` and `MetricCard` keep `SchemaRenderer`'s schema-shaped props out of the `...props` spread (objectui#4357). (objectui `306c10136`)
- **fields** — fields: the currency adornment has one symbol channel (objectui `3a9021e26`)
- **components** — `@object-ui/components` compiles under `noImplicitAny` — the workspace's last strict-relaxing package (objectui `4dadf0d3d`)
- **components** — data-table row menu — the built-in Edit/Delete predicate parameters are derived from the authoring type, not hand-restated (objectui `4b70d287c`)
- **plugin-chatbot** — Replace the three `messages as any` casts at the `@object-ui/types` ↔ `@object-ui/plugin-chatbot` `ChatMessage` boundary with one explicit typed adapter (`toRuntimeMessages` / `au… (objectui `37bbc42d1`)
- **fields** — Currency amounts now follow each currency's own ISO 4217 fraction-digit convention instead of a hardcoded 2 (objectui#4361). (objectui `0f2134831`)
- **plugin-charts** — Analytics: `ObjectChart` consumes the shared label-net helpers instead of a third copy (objectui `0b49d6032`)
- **i18n** — Inline-edit toggle reads "Edit fields" without an I18nProvider, matching every locale pack (objectui `e076fd50f`)
- **gate** — `@object-ui/plugin-detail` now declares `react-router-dom` as a peer dependency (`^6.0.0 || ^7.0.0`), the range its three siblings already use. (objectui `456aac831`)
- **app-shell** — refactor(app-shell): collapse the metadata-admin designer table's byte-identical key pairs (objectui `f762f5bdf`)
- **core,react,dashboard,report** — Analytics: the dimension label net's fetch-and-memo glue is written once, not once per surface (objectui `ee26e65e7`)
- **i18n** — Six i18n keys no longer render as raw key strings on hosts with no `I18nProvider` (objectui#4396) (objectui `dad805d8f`)
- **app-shell,data-objectstack** — Publishing a view from the console no longer serves a five-minute-stale override map — every writer now routes through one invalidation seam (objectui `d2f6e6b1b`)
- **components** — `ActionParamDialog`'s `select` branch no longer renders a hardcoded English `Select...` placeholder. The fallback used when an action param declares no `placeholder` of its own no… (objectui `5bf09fdba`)
- **i18n** — i18n: the two search placeholders become pack values, and four values the packs served in English get translated (objectui `bb58d1d61`)
- **dashboard,report** — Analytics: a LOCAL select dimension on a table / pivot widget — and on a dataset-bound report — now renders its option label through the locale bundle (objectui `326a70f32`)
- **i18n** — A gantt task titled `A$&B` no longer prints `{{title}}` back into its own delete dialog — the two hand-rolled provider-less fallback interpolators are literal, like i18next (objectui `0ca6096ff`)
- **app-shell** — fix(app-shell): the metadata-admin designer's own i18n table uses the typographic ellipsis (objectui `b31827358`)
- **i18n** — i18n copy: one ellipsis glyph across the ten packs, `usted` in the es draft-preview empty state, and a pt sentence that stops contracting `de` onto its own hole (objectui `3e19fe78a`)
- **i18n** — i18n: `createSafeTranslation`'s provider-less fallback now honours a call site's inline `defaultValue` (objectui `d46f9b8c4`)
- **data-objectstack** — Every view write path now invalidates the override map — a created, renamed or deleted view is no longer shadowed by a five-minute-stale batch read (objectui `85a3082e6`)
- **i18n** — `detail.showEmptyRelated` renders Russian and Arabic again — the "+N empty" button no longer falls through to English at the counts it takes most often (objectui `2fea4d2fa`)
- **console** — Retire the dormant bespoke object-detail page factory and its seven widgets (objectui `275d7df13`)
- **fields** — fix(fields): `formatCurrency` keeps both cents digits on a fractional amount (objectui `d2e2caf40`)
- **i18n** — `pickLocalized` reads own properties only, and takes only string values, on every limb (objectui `405e80875`)
- **dashboard,i18n** — fix(dashboard,i18n): KPI cards and dashboard filters resolve authored labels instead of dropping them (#4032) (objectui `7e4f0e530`)
- **app-shell** — Home's action centre badges everything that is waiting, not the five rows it has room for (objectui `de627792e`)
- **console** — The Applications page's search box no longer takes the page out on the first keystroke when an app carries a non-string label (objectui `3b4d78ea0`)
- **console,i18n** — The console's Applications page is localized — its own chrome only, never the server's words (objectui#4307). (objectui `734d186a0`)
- **app-shell** — Studio's read-only packages stop advertising writability they do not have — the `access` header badge and the `Data → Form` layout caption now report the gate that actually govern… (objectui `f046f885a`)
- **i18n,list** — List sort: the relational hint stops recommending a formula field, the one type the server refuses to sort by (objectui `7f1cb3323`)
- **fields,i18n,dashboard** — Numbers render in the user's locale, and a `Field.number` year is no longer `2,026` (objectui `45e1949b4`)
- **plugin-detail** — `record:related_list` and the detail synthesizer now declare two shapes they already accepted at runtime. (objectui `63fe8fdda`)
- **app-shell** — Home's action centre stops counting messages the user has already read, and the inbox is read once per page instead of twice (#4316, #4225) (objectui `bed18a5c2`)
- **plugin-map** — `object-map` reads its configuration from the declared `map` input only — `filter` is the query filter, and a map authored with both stopped rendering markers (objectui `b388d0eb0`)
- **core,plugin-dashboard,plugin-charts** — Analytics surfaces now run resolved select-option labels through the locale bundle — the chart legend and the related list on one page stop disagreeing (objectui `5900ac56f`)
- **app-shell** — Home's action centre no longer says "You're all caught up" to a user whose inbox it failed to read (#4235) (objectui `771466aa5`)
- **i18n** — Action confirm dialogs and success toasts now honour the bundle's translated `confirmText` / `successMessage`, not just `label` (objectui#4265). (objectui `ceccdcfa3`)
- **components** — A collapsed sidebar now survives a reload — `SidebarProvider` reads the `sidebar_state` cookie it has always written (objectui `f5e114348`)
- **core,console** — Form actions no longer carry a record id across an object boundary (#4292). (objectui `9461dd395`)
- **console** — API Console renders the Storage group again — its catalog key now names the canonical `file-storage` slot instead of the route (objectui `a00d23c0d`)
- **list** — List sort: the picker stops borrowing the filter whitelist, and a header click is no longer a one-way door out of the view's declared sort (objectui `33c32bf33`)
- **plugin-dashboard** — Dashboard `combo` widgets draw as combos on the dataset path — the dataset owns the data, the author owns the presentation (objectui `3c6e84cca`)
- **console** — A `type: 'form'` action fired from a record now EDITS that record instead of creating a duplicate (objectui `6d8231cc4`)
- **plugin-detail** — fix(plugin-detail): synthesize page components in the spec's `properties` carrier so Studio page-create can persist (objectui `35997cea5`)
- **core,app-shell** — `DeclaredActionsBar` reads `visible` / `disabled` off the typed action def instead of through `(action as any)`. (objectui `ee66e2ebf`)
- **dashboard,report** — Pivot buckets encode an empty dimension value as JSON `null`, so it no longer collides with a row whose value is literally the placeholder character (objectui `49ae9f42d`)
- **dashboard** — fix(dashboard): resolve a dotted dimension's labels on table and pivot dataset widgets (objectui `436681e72`)
- **app-shell** — `DeclaredActionsBar` binds its row through the shared `usePredicateRecordContext` (objectui `4cb0562b5`)
- **list** — `list-view` now reads its `dataSource` binding through the shared `ElementDataSourceGate` instead of a private copy of the precedence table (objectui `37cd8e4ff`)
- **metadata-admin** — metadata-admin: diagnose a path on the right-hand side of `==` / `!=` in a visibility predicate (objectui `b95412031`)
- **analytics** — A dataset dimension on a dotted relationship path now renders its option labels instead of the raw stored enum (objectui `613b16707`)
- **@object-ui/core** — Retire `params.newTab` on a url action — `openIn: 'new-tab'` is the sanctioned spelling (objectui `d6aa1726a`)
- **data** — The second metadata client class surfaces the runtime authoring gate's advisories instead of discarding them (objectui `605b74720`)
- **console** — fix(console): a Setup-only environment lands on `/home`, not Setup's all-zero System Overview (objectui `234238ed7`)
- **detail,form,app-shell** — `richtext` fields are placed like the long-form fields they are — four layout sets stopped spelling the type three ways the spec rejects (objectui `c32a8a1bf`)
- **fields** — A dependency-gated option list no longer deletes the field's stored value on mount (objectui `bc64bfe0f`)
- **detail** — Inline edit no longer offers a record picker for a spec-spelled `autonumber` field that carries a `reference_to` (objectui `6d01319dd`)
- **@object-ui/app-shell** — metadata-admin: retire the standalone `validation` resource, and move `ValidationPreview` onto the embedded path the framework actually evaluates (objectui#4132) (objectui `3032107dd`)
- **@object-ui/plugin-detail** — `record:details` stops publishing a `layout` key the spec removed and the renderer never honoured (objectui `7d04b0e84`)
- **preview** — The Unpublished-app banner now reads the ADR-0045 publish gate `_unpublished`, not the navigation flag `hidden` — so a published app that is merely kept out of the launcher no lon… (objectui `690ae0f04`)
- **detail** — fix(detail): inline edit no longer destroys array values or flattens types on the record page (objectui `e7663f2ff`)
- **types,core** — A dashboard date filter's default has one spelling again — the bare preset name — and the `{ preset }` object becomes a documented legacy alias with a retirement window (objectui `abb0f81d3`)
- **app-shell** — The bell popover's and Home's "see all" drills open inside an app the user can actually open, not the setup app (objectui `f812de6bd`)
- **detail** — A `password` or `secret` field on the record detail page is no longer inline-editable: it renders no pencil / double-click affordance and produces no editor, on both the details b… (objectui `5e2e9fa8c`)
- **metadata-admin** — `DatasourcePreview` no longer renders three key groups `DatasourceSchema` rejects (objectui#4131). `retryPolicy`, `healthCheck` and `capabilities` were removed from the datasource… (objectui `ab37c5f59`)
- **app-shell** — Set-default on a saved view fires its write again — a stored `id` can no longer rename the tab out from under the overlay read (objectui `79a40adae`)
- **app-shell** — The bell's Approvals and Activity tabs fill in on Home, Organizations and the AI screen — from the same fetch the cards below them already use (objectui `8b971f826`)
- **studio** — Studio surfaces the runtime authoring gate's advisory findings instead of discarding them client-side (objectui `c0f9a4bd5`)
- **i18n** — An inline per-locale label now renders its locale's string at the thirteen read sites the `@objectstack/spec` 17.0.0-rc.6 bump exposed (objectui `bb68488ff`)

_Derived from the changesets objectui declared over this range — 99 releasing of 115 added across 123 non-merge commit(s). Excluded as shipping no package code: 16 release-nothing changesets, 11 commits carrying no changeset (run with `--all` to list them)._
```

---
_Generated by [Claude Code](https://claude.ai/code/session_01168CZYvB7fsrsshzuGtdvG)_


hotlong:
Correction to the PR body, on the release-page block only.

The body says *"one character is altered below"*. There are **two** alterations, both made so the block survives GitHub's body sanitizer:

1. The `lint` entry's `< button` carries an inserted space (disclosed).
2. **The block's first line is an HTML comment in the real output** — the angle brackets around `!-- objectui 6314e87f2d49..6d77acfe3125 … --` were stripped and are not disclosed in the body. That line is a machine-readable marker, so pasting it as shown would land a broken marker on the release page.

Neither alteration affects any entry's text beyond those two spots, but treat the block as a preview rather than a paste source. The authoritative output is:

`node scripts/objectui-range.mjs --from 6314e87f2d49b1ff3b158c296f1b2a52d14dff68 --to 6d77acfe3125949796918ea12b926d3460bc933d`

---

Issue #8111, opened 2026-08-12
[finding] `registerSharingEndpoints` / `respondSharingError` still answers the bare-string error dialect (ADR-0112) — the sharing half split out of #8073


hotlong:
Split out of #8073 at triage (2026-08-12), per the `domain:cli` seat's serial/split analysis on that card (comment 5269022297): the explain pair and the `/sharing/*` family are different blast radii, and this half carries a question the explain half does not.

## Fact (inherited from #8073's measurement — re-verify at claim)

`registerSharingEndpoints`'s single emitter `respondSharingError` in `packages/rest/src/rest-server.ts` answers `{ code, error: msg.replace(...) }` — a bare-string `error` — feeding 400 / 403 / 404 / 409 / 422 across the whole `/sharing/*` family. `body.error.code`, the one position ADR-0112 D5 declares, reads `undefined` on all of them.

## Why this is held as a finding rather than queued

The envelope convergence itself is implementation of the settled ADR-0112 D5 ruling. But this emitter carries a second, unrelated protocol: `msg.startsWith(CODE)` — the error **code is recovered by parsing the message text**. Whether that string-prefix protocol survives (kept alongside the D5 envelope) or is retired with it must be ruled **before** dispatch, not during it — retiring it silently changes what programmatic consumers can read.

Also inherited scope notes from #8073:

- Any converging PR must **migrate** existing flat-shape pins to the D5 position, never delete them.
- The SDK-compatibility measurement on #7981 (`ObjectStackClient.fetch` reads both envelopes' spots) was taken for that card's consumers; it must be **re-run** for the `/sharing/*` consumers, not inherited.
- Same-file serial: `packages/rest/src/rest-server.ts` is the repo's hottest file (#5949); check in-flight writers before any claim.

Promotion path: at the findings round, either (a) the prefix-protocol question goes to the maintainer and this queues after the ruling, or (b) if a measured consumer census shows nothing reads the prefix, it collapses into a plain convergence card.

Provenance: #8073 (measurement record), PR #8071 / #7981 (the pattern to follow), #7035 / PR #7293 (the dialect retirement).

hotlong:
Finding-grading round: **promoted to `pm:queue`** (`domain:cli`, type Bug).

[...]

PR #8174, opened 2026-08-12
fix(rest): one error envelope across the /security/explain pair (#8073)

### LINKED ISSUE #8073: [finding] the two registrars ADJACENT to #7981 still answer the retired flat / bare-string error dialects (ADR-0112)
Observation-class finding, recorded while fixing #7981 (PR #8071). **Not fixed there** — that card was scoped to the three `/security/suggested-bindings` routes by explicit ruling, and these are separate registrars. Filed unassigned.

## Fact

With #7981 landed, `registerSecurityEndpoints` speaks the ADR-0112 D5 envelope `{ error: { code, message } }` on every arm. Its two immediate neighbours in `packages/rest/src/rest-server.ts` do not, and they carry the same two dialects #7035 (PR #7293) retired from this file's `/meta` refusals:

| registrar | routes | arms still on a retired dialect |
|:---|:---|:---|
| `registerSecurityExplainEndpoints` | `GET/POST /api/v1/security/explain`, `GET /api/v1/security/my-delegable-scope` | 401 `UNAUTHORIZED`, 501 `NOT_IMPLEMENTED`, 400 `VALIDATION_FAILED`, 403 `PERMISSION_DENIED` are all flat `{ code, message }`; the two 500s are `{ code, error: 'a bare string' }` |
| `registerSharingEndpoints` (`respondSharingError`) | the `/sharing/*` family | `{ code, error: msg.replace(...) }` — bare-string `error`, one emitter feeding 400 / 403 / 404 / 409 / 422 |

So `body.error.code`, the one position ADR-0112 D5 declares, reads `undefined` on all of them, and `/security/explain` is directly adjacent to the routes that no longer behave that way — a client calling `explain` and then `suggested-bindings` now meets two shapes across one `security` family.

## Why it is worth recording separately

`registerSecurityExplainEndpoints` is not dormant: the client SDK exposes it as `client.security.explain` and `client.security.describeDelegableScope` (`packages/client/src/index.ts`), and `content/docs/permissions/*` documents the route for operators. `security-routes.test.ts` currently PINS the flat shape (`expect(res.body.code).toBe('NOT_IMPLEMENTED')`, `...('PERMISSION_DENIED')`, `...('EXPLAIN_FAILED')`), so those pins move with any convergence — the same fixture-triage step #7981 needed.

The same wire-visibility question #7981 had applies and has the same measured answer: the SDK's error path in `ObjectStackClient.fetch` reads BOTH envelopes' declared spots (`errorBody?.code ?? errorBody?.error?.code`, and a `typeof errorBody?.error === 'string'` limb for the message), so a converged body keeps producing identical `err.code` / `err.message`. That measurement is recorded on #7981 and should be re-run, not assumed, for whatever consumers these two families have.

## Not claimed here

- Whether `respondSharingError`'s `msg.startsWith(CODE)` string-prefix protocol should survive at all — it is a second, unrelated question (the code is recovered by parsing the message text) and is not part of the envelope position.
- Any status-code change. As on #7981, this is about body shape only.
- Whether this should be one PR or two: `registerSharingEndpoints` is a wider blast radius than the explain pair and may deserve its own card.

## Suggested handling

Triage-level, not decided here: converge each registrar onto `{ error: { code, message } }` through a single emitter, exactly as PR #8071 did for `registerSecurityEndpoints` — the shape is already ruled by ADR-0112 D5, so this is implementation of an existing decision. The explain pair is the smaller, higher-adjacency half and is a natural first card.

---

hotlong:
Fixes #8073

Scoped at triage to the **explain pair only** — `registerSecurityExplainEndpoints`
(`GET/POST /api/v1/security/explain`, `GET /api/v1/security/my-delegable-scope`).
`registerSharingEndpoints` / `respondSharingError` is deliberately untouched: it is
split out to #8111 and held until its `msg.startsWith(CODE)` prefix-protocol question
is ruled.

## What was wrong

With #7981 (PR #8071) landed, `registerSecurityEndpoints` speaks the ADR-0112 D5
envelope on every arm. Its immediate neighbour carried **both** dialects that #7035
(PR #7293) had already removed from this file's `/meta` refusals:

| arms | shape |
| :--- | :--- |
| 401 `UNAUTHORIZED`, 501 `NOT_IMPLEMENTED`, 400 `VALIDATION_FAILED`, 403 `PERMISSION_DENIED` | flat `{ code, message }` |
| 500 `EXPLAIN_FAILED`, 500 `DELEGABLE_SCOPE_FAILED` | `{ code, error: 'a bare string' }` |

So `body.error.code` — the one position ADR-0112 D5 declares — read `undefined` on all
six, and a client calling `explain` and then `suggested-bindings` met two shapes inside
one `security` family.

## The fix

All eight refusal arms of both handlers now go through **one family-local emitter** that
delegates to the **shared** `sendError` from `@objectstack/types` — imported as
`sendEnvelopeError`, following #8135, because this module has a local `sendError` of its
own (the sanitizing responder for *thrown* errors, a different thing). Following PR
#8071's shape, but routed through the shared builder rather than a local body literal, so
the family agrees by construction and `code` is typed to the closed ADR-0112 vocabulary
at every call site.

**No status code moves**, and no code VALUE changes — all six are already registered
(`StandardErrorCode` for `NOT_IMPLEMENTED` / `PERMISSION_DENIED`, `ERROR_CODE_LEDGER`'s
`@objectstack/rest` block for the rest), so nothing in `packages/spec` moves. The one
other body change: the 400 arm's Zod-issue dump moves from a top-level `detail` sibling
to `error.details`, the slot `ApiErrorSchema` actually declares for structured context.

Untouched on purpose: the anonymous caller's 401 is a **different seam** —
`enforceAuth`'s shared `ANONYMOUS_DENY_BODY` (#2567), which fires before these arms.

## The pins MIGRATED, none deleted

`security-routes.test.ts`'s three flat-shape pins moved to the D5 position:
`res.body.code` → `res.body.error.code` for `VALIDATION_FAILED` (x2),
`PERMISSION_DENIED`, `NOT_IMPLEMENTED`, `EXPLAIN_FAILED`. The 500 pin also gained a
`body.error.message` assertion, since that arm carried the bare-string dialect too. No pin
was removed and no arm was left on the flat shape.

New driven suite `security-explain-envelope.test.ts` (14 cases) covers all eight arms —
the delegable-scope handler had **no** tests at all before — asserting the ADR-0112
**pair** (status AND nested code, never a bare `toThrow`, since these handlers send and
never throw) plus both retired dialects' absence, and a **derived** cross-arm skeleton pin
so a third dialect fails even if someone also adds a matching literal case.

## Reverse verification

Reverted the conversion (`git checkout origin/main -- rest-server.ts`, tests left at
HEAD): **16 of 21 red**, in two distinct predicted directions.

Flat arms — `res.body.error` is `undefined`:

```
TypeError: Cannot read properties of undefined (reading 'code')
    111|     expect(res.body.error.code).toBe('VALIDATION_FAILED');
```

Bare-string 500 arm — `res.body.error` is the string, so `.code` is `undefined`:

```
AssertionError: expected undefined to be 'EXPLAIN_FAILED' // Object.is equality
```

And the derived skeleton pin caught the divergence structurally:

```
AssertionError: explain 500 drifted: {"code":"EXPLAIN_FAILED","error":"boom"}:
  expected 'code:string|error:string' to be 'code:string|message:string'
```

The 5 that stayed green are the non-envelope cases (route registration, request
delegation, the `enforceAuth` 401, the healthy 200 decision). Restored: `git diff --stat
HEAD` empty (byte-identical), 21/21 green.

## SDK claim RE-MEASURED against these consumers

Not inherited from #7981. Drove the real `ObjectStackClient` against a stubbed transport
answering the old and new body for each of the eight arms, through
`client.security.explain()` and `client.security.describeDelegableScope()`:

| property | result |
| :--- | :--- |
| `err.code` | **IDENTICAL** on all 8 arms |
| `err.message` | **IDENTICAL** on all 8 arms |
| `err.httpStatus` / `err.category` / `err.retryable` / `err.fields` | identical on all 8 |
| `err.details` | changes — see below |

`ObjectStackClient.fetch` reads both envelopes' declared spots (`errorBody?.code ??
errorBody?.error?.code`, and a bare-string limb for the message), which is why the two
that matter do not move. `err.details` does change, on every arm, because its last
fallback is *the whole response body* and the whole body is what changed; on the 400 arm
it now resolves to the declared `error.details` instead. No consumer in the repo reads
`err.details` for these routes, `content/docs/permissions/*` documents no error body for
them, and the other in-repo `security.explain` callers (`runtime/src/domains/automation.ts`,
the dogfood suites) call the **service**, not the route.

## Envelope ratchet LOWERED, per the gate's own rule

`siblingCode 75 → 73` in `scripts/check-route-envelope.mjs`. Measured at merge-base
(`6ceffe0ac`) 75 and at branch head 73; the two vanished sites are merge-base lines
9332/9390 — both the 500 arms inside this function — and every surviving site maps 1:1
onto a head line by the edit's own line shift (verified programmatically: zero unmatched,
zero new). `stringError` is unmoved at 44 by construction: both 500 arms carried a
*computed* message (`msg.slice(0, 500)`) that counter cannot see, and the six flat arms
were never counted by either dialect, having no `error` key at all.

## Verification

- `pnpm --filter @objectstack/rest test` — **103 files, 1741 tests, all passing**
- `pnpm --filter @objectstack/rest typecheck` — clean
- `pnpm check:type-check-debt` (full closure built first, as `lint.yml` does) — OK, *"none
  above its recorded number"*; `@objectstack/rest` stays at its recorded 155, **not raised**
- `pnpm check:route-envelope` incl. `--self-test` — green with the banked 73
- `node scripts/check-nul-bytes.mjs` — OK, plus a self-scan of every changed file for the
  wider control-byte range

---

### DIFF
--- a/.changeset/security-explain-envelope-convergence.md
+++ b/.changeset/security-explain-envelope-convergence.md
@@ -1,0 +1,33 @@
+---
+"@objectstack/rest": patch
+---
+
+fix(rest): one error envelope across the `/security/explain` pair (#8073)
+
+`registerSecurityExplainEndpoints` — `GET/POST /api/v1/security/explain` and
+`GET /api/v1/security/my-delegable-scope` — answered two retired dialects across
+its eight refusal arms: the 401 / 501 / 400 / 403 arms were flat
+`{ code, message }`, and the two 500s were `{ code, error: 'a bare string' }`. So
+`body.error.code`, the one position ADR-0112 D5 declares, read `undefined` on all
+six — while the immediately adjacent registrar (`/security/suggested-bindings`,
+converged in #7981) already answered the declared shape. A client calling
+`explain` and then `suggested-bindings` met two envelopes inside one `security`
+family.
+
+Every arm now emits `{ success: false, error: { code, message } }` through the
+shared `sendError` from `@objectstack/types` — the same builder every conformant
+route module writes through — so the family agrees by construction rather than by
+eight literals happening to match. The 400 arm's Zod-issue dump moves from a
+top-level `detail` sibling to `error.details`, the slot `ApiErrorSchema` declares
+for structured context.
+
+No status code moves, and no code VALUE changes: `UNAUTHORIZED`,
+`NOT_IMPLEMENTED`, `VALIDATION_FAILED`, `PERMISSION_DENIED`, `EXPLAIN_FAILED` and
+`DELEGABLE_SCOPE_FAILED` are all already registered, so nothing in
+`packages/spec` moves. `ObjectStackClient` reads both envelopes' declared spots
+(`errorBody?.code ?? errorBody?.error?.code`, and a bare-string limb for the
+message), so `client.security.explain()` and
+`client.security.describeDelegableScope()` keep throwing identical `err.code` and
+`err.message` — re-measured against these call paths rather than inherited from
+#7981. `err.details` does change on refusals, from "the whole response body" to
+the structured slot or the new body.
--- a/packages/rest/src/rest-server.ts
+++ b/packages/rest/src/rest-server.ts
@@ -56,6 +56,11 @@ import type { DirectMountedRoute, MountedRouteSource } from './direct-mount.js';
 import { RestServerConfig, RestApiConfig, CrudEndpointsConfig, MetadataEndpointsConfig, BatchEndpointsConfig, RouteGenerationConfig } from '@objectstack/spec/api';
 import { DataProtocol, MetadataProtocol } from '@objectstack/spec/api';
 import type { FieldErrorCode } from '@objectstack/spec/api';
+// [#8073] The closed ADR-0112 error vocabulary, so the explain family's single
+// refusal emitter types its `code` parameter as the vocabulary rather than as
+// `string` — an invented code is a compile error at the call site instead of a
+// runtime surprise on whichever arm a test happens to drive.
+import type { ErrorCode } from '@objectstack/spec/api';
 // The async-import row ceiling has exactly one definition, in the spec, whose
 // TSDoc is its public statement (#6535). rest is the only enforcer, so it reads
 // that export rather than re-declaring the literal beside a "mirrors spec" comment.
@@ -9268,6 +9273,45 @@ export class RestServer {
             catch { return undefined; }
         };
 
+        /**
+         * [#8073] The ONE refusal emitter for this route family — every arm of
+         * both handlers goes through it, so "explain and my-delegable-scope
+         * answer the same shape" is a property of the code rather than of
+         * eight literals that happen to agree.
+         *
+         * Before this, the family carried BOTH dialects ADR-0112 D5 retires:
+         * the 401/501/400/403 arms were flat `{ code, message }` and the two
+         * 500s were `{ code, error: 'a bare string' }`, so `body.error.code` —
+         * the one position D5 declares — read `undefined` on all six. #7035
+         * (PR #7293) had already removed both from this file's `/meta`
+         * refusals and #7981 (PR #8071) from `registerSecurityEndpoints`, the
+         * immediately ADJACENT registrar: a client calling `explain` and then
+         * `suggested-bindings` met two shapes inside one `security` family.
+         *
+         * Emitted through the SHARED builder (`sendError` from
+         * `@objectstack/types`, imported as `sendEnvelopeError` because this
+         * module has a local `sendError` of its own — the sanitizing responder
+         * for THROWN errors, a different thing). That is what makes this the
+         * reference shape by construction rather than a ninth local literal
+         * agreeing with the eight it replaced, and it types `code` to the
+         * closed vocabulary for free.
+         *
+         * ⛔ Status codes are untouched: only the POSITION of `code` and
+         * `message` moves. `detail` — the 400 arm's Zod-issue dump — moves to
+         * `error.details`, the slot `ApiErrorSchema` actually declares for
+         * structured context; as a top-level sibling it was undeclared.
+         */
+        const respondError = (
+            res: any,
+            status: number,
+            code: ErrorCode,
+            message: string,
+            details?: unknown,
+        ): void => sendEnvelopeError(
+            res, status, code, message,
+            details === undefined ? undefined : { details },
+        );
+
         const handler = async (req: any, res: any) => {
             try {
                 const environmentId = isScoped ? req.params?.environmentId : undefined;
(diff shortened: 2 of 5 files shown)


hotlong:

### Re-measuring the SDK claim was the right instruction, and the honest answer was not "identical"

I told you not to inherit #7981's measurement because it was taken for different consumers. You drove the real `ObjectStackClient` against a stubbed transport, old body and new, across all eight arms via `client.security.explain()` and `client.security.describeDelegableScope()` — and the result is **not** a clean "no change":

- `err.code`, `err.message`, `err.httpStatus`, `err.category`, `err.retryable`, `err.fields` — identical on all 8
- **`err.details` changes on every arm**, because its last fallback is *the whole response body*, and the whole body is what moved

A report that stopped at "the two that matter are identical" would have been true and incomplete. Reporting the one that moves, explaining the mechanism, and then discharging it — no in-repo consumer reads `err.details` for these routes, `content/docs/permissions/*` documents no error body for them, and the other in-repo `security.explain` callers go through the **service**, not the route — is the difference between a measurement and a reassurance.

### The derived skeleton pin is the part I did not ask for and should have

Per-arm literal pins are satisfiable by *adding another matching literal case* — which is precisely how a family grows a third dialect while every existing test stays green. The derived cross-arm skeleton pin fails structurally instead, and the reverse verification shows it doing exactly that:

```
explain 500 drifted: {"code":"EXPLAIN_FAILED","error":"boom"}:
  expected 'code:string|error:string' to be 'code:string|message:string'
```

That assertion survives a future author who adds a new arm *and* its matching test. Recording it as the shape to reuse whenever a card is "make N call sites agree."

### The reverse verification distinguishes the two dialects

16 of 21 red, in **two distinct predicted directions** — flat arms give `TypeError: Cannot read properties of undefined (reading 'code')`, the bare-string 500 gives `expected undefined to be 'EXPLAIN_FAILED'`. Two dialects, two failure modes, both anticipated. And the 5 that stayed green are named and explained (route registration, request delegation, the `enforceAuth` 401, the healthy 200) — non-envelope cases that the conversion cannot touch. Restored byte-identical.

**The pins migrated, none deleted** — all five moved to `res.body.error.code`, and the 500 pin additionally gained a `body.error.message` assertion because that arm carried the bare-string dialect and the position alone would not have caught it. That extra assertion is the one a mechanical find-and-replace would have missed.

Also found in passing: **the delegable-scope handler had no tests at all** before this. Worth stating plainly rather than folding into a count.

### The ratchet accounting is exact

`siblingCode 75 → 73`, measured at merge-base (`6ceffe0ac`) and head, with the two vanished sites identified by line (9332/9390, both 500 arms inside this function) and **every surviving site mapped 1:1 onto a head line through the edit's own line shift — zero unmatched, zero new, verified programmatically.**

And the account of why `stringError` did *not* move is better than the number: both 500 arms carried a **computed** message (`msg.slice(0, 500)`) that the counter cannot see, and the six flat arms were never counted by either dialect at all, having no `error` key. Explaining why a counter stayed still is how you tell a correct ratchet from a coincidental one.

Scope held: no status moves, no `code` value changes (all six already registered, so nothing in `packages/spec` moves), `enforceAuth`'s shared `ANONYMOUS_DENY_BODY` correctly left alone as a different seam (#2567) that fires before these arms, and the 400 arm's Zod dump moved into `error.details` — the slot `ApiErrorSchema` actually declares — rather than left as a top-level sibling.

**#8111 filed for the sharing half**, held pending the `msg.startsWith(CODE)` prefix-protocol ruling. That closes the loop I left open when I split this card; I will rule on the prefix protocol before dispatching it.

Not raised: `@objectstack/rest` at its recorded **155** — the zero-margin ceiling. 103 files / 1741 tests green.

---

Issue #8054, opened 2026-08-12
`ListRunsRequestSchema.limit` declares `.min(1).max(100)` but the boundary enforces only integer-ness — `?limit=0` answers "no runs", `?limit=101` ignores the cap


baozhoutao:
The same declared-vs-enforced family as #7359, on the **sibling parameter of the same route**, left behind when `status` was fixed.

## Symptom

`GET /api/v1/automation/:name/runs` against a flow with 10 runs:

| query | result |
|---|---|
| `?limit=0` | **200 with zero rows** |
| `?limit=-5` | **200 with zero rows** |
| `?limit=101` | **200 with all 10 rows** (cap not applied) |
| `?limit=abc` | 400 `VALIDATION_FAILED` — "Invalid \`limit\` query parameter — expected a whole number, received \"abc\"" |
| `?limit=1.5` | the same 400 |

So the boundary refuses the wrong *type* but accepts anything outside the declared *range*. Reproduced 2×, identical both passes.

The `?limit=0` arm is the one that bites: a caller paging with a computed limit that reaches 0 is told the flow has **no runs** — a confidently wrong answer of exactly the shape #7359 was filed for.

## Root cause

`packages/runtime/src/query-param.ts:121` `parseIntegerParam` checks `Number.isInteger` and returns the value; it has no min/max arm — unlike its sibling `parseEnumParam`, which does read the closed set. `packages/runtime/src/domains/automation.ts:737` calls it bare, so `ListRunsRequestSchema`'s declared `(1, 100)` never reaches the boundary.

## Suggested shape

Give `parseIntegerParam` optional bounds and pass `ListRunsRequestSchema`'s own at the call site, so the declared and enforced ranges cannot drift — the discipline #7359's fix applied to `status` by reading `ExecutionStatus.options` rather than re-listing them.

## Severity

P3 — no corruption, no leak. A wrong-but-confident answer and an unenforced result-set cap.

## Source

Found by the platform checklist retest of `automation.flow-runs-page-test-trigger` (framework `279ee48a`), while verifying #7359's fix. Held no clause of that item.

hotlong:
[...]

PR #8177, opened 2026-08-12
fix(runtime): PUT /automation/:name answers 400 VALIDATION_FAILED, not 500, matching POST (#8123)

### LINKED ISSUE #8123: `PUT /api/v1/automation/:name` answers 500 for the same four malformed flow bodies #8055 just reclassified on POST

hotlong:
Fixes #8123

## What

`PUT /api/v1/automation/:name` (`updateFlow`) makes the identical `automationService.registerFlow(name, definition)` call `POST /` makes, in the same file (`packages/runtime/src/domains/automation.ts`), with no classification around it. #8055 / PR #8125 reclassified `POST /`'s refusal from 500 `INTERNAL_ERROR` to 400 `VALIDATION_FAILED` through a module-local, route-agnostic `flowDefinitionRefusal` helper — but left `PUT` untouched, so the two doors disagreed about the class of an identical malformed-flow refusal.

`PUT /:name`'s `registerFlow` call is now wrapped the same way, reusing the same `flowDefinitionRefusal` helper `POST /` uses:

```ts
try {
    automationService.registerFlow(name, definition);
} catch (e) {
    return {
        handled: true,
        response: deps.errorFromThrown(flowDefinitionRefusal(e), VALIDATION_FAILED_STATUS),
    };
}
return { handled: true, response: deps.success(definition) };
```

The helper itself is untouched — `POST /`'s behaviour is unchanged.

## The pin — both doors compared directly

Added `packages/runtime/src/domains/automation-put-post-error-parity.test.ts`: a table-driven suite that drives **both** `POST /` and `PUT /:name` with the same four malformed bodies from #8055/#8123 (missing `label`, unknown node key, malformed `try_catch` region, undeclared #4277 config key) and asserts:

- each door independently answers the full ADR-0112 envelope (`status`, `code`) plus ADR-0114 `details.fields[]` shape — not just "not 500"
- **the two doors' responses are compared directly** — same `status`, same `code`, same `fields[]`, same `message` — so reverting either door's classification (not just `PUT`'s) reddens this file
- the #4277 self-correcting message survives verbatim through `PUT` too (names the key, the node, the node type, and the declared keys)
- the raw Zod issue array does not reach the wire via `PUT` either
- `PUT`'s own `{ definition }` / bare-body dialect (`body.definition ?? body`) classifies identically either way
- an engine error that declares its own `.status` still keeps it on `PUT` (the same escape hatch `POST` honours)

Contrast controls: a well-formed body still registers 200 on both doors, and no previously-refused body starts succeeding on `PUT` — only the class changed from 500 to 400.

## Reverse verification

Reverted the source fix only (`git checkout origin/main -- packages/runtime/src/domains/automation.ts`, keeping the new test file) and re-ran the parity suite. 9 of 9 `PUT`-touching tests went red, all for the expected reason — the raw thrown error now escapes uncaught from the unguarded call at `automation.ts:880`, so the test's own `await` rejects instead of observing a classified response, e.g.:

```
FAIL  src/domains/automation-put-post-error-parity.test.ts > ... case "missingNodeLabel": POST and PUT answer the SAME envelope for the identical body
ZodError: [
  { "expected": "string", "code": "invalid_type", "path": ["nodes", 0, "label"], "message": "Invalid input: expected string, received undefined" }
]
 ❯ handleAutomationRequest src/domains/automation.ts:880:35
```

```
FAIL  ... case "undeclaredConfigKey": POST and PUT answer the SAME envelope for the identical body
Error: Flow 'welcome_flow' rejected: 1 undeclared config key(s) (#4277).
  - node 'n' (notify): unknown config key `totallyBogusKey` at config.totallyBogusKey ...
 ❯ handleAutomationRequest src/domains/automation.ts:880:35
```

The other 145 test files (2196 tests, including the untouched POST-only `#8055` suite) stayed green throughout. Restored the fix via `git apply` of the saved patch; `git diff --stat HEAD` on `automation.ts` came back identical before and after the round trip (13 insertions, 1 deletion both times).

## Verification

- `pnpm --filter '@objectstack/runtime^...' build` — closure builds clean
- `pnpm --filter @objectstack/runtime test -- --maxWorkers=2` — **146 test files, 2205 tests, all pass**
- `pnpm check:type-check-debt` (full closure built first) — green, `@objectstack/runtime` un-drifted from its recorded 227
- `node scripts/check-nul-bytes.mjs` — OK

## Not in scope

Widening the refusal, changing which bodies are rejected, or altering `flowDefinitionRefusal`'s behaviour for `POST` — none of that changed here. Did not touch `#8054`'s `ListRunsRequestSchema.limit` bounds or `#7968`'s `/runs/:runId/screen` gating, both queued behind this card on the same file.

## Changeset

`.changeset/put-automation-error-class-parity.md` — `@objectstack/runtime` patch, user-visible status-code change on `PUT`.

---

### DIFF
--- a/.changeset/put-automation-error-class-parity.md
+++ b/.changeset/put-automation-error-class-parity.md
@@ -1,0 +1,11 @@
+---
+'@objectstack/runtime': patch
+---
+
+`PUT /api/v1/automation/:name` answers 400 `VALIDATION_FAILED`, not 500, for a malformed flow definition — matching `POST /api/v1/automation` (#8123)
+
+#8055 reclassified `POST /`'s `registerFlow` refusal from 500 `INTERNAL_ERROR` to 400 `VALIDATION_FAILED`. `PUT /:name` makes the identical `registerFlow` call in the same file and was left uncaught, so the two doors disagreed about the class of an identical refusal: publishing a flow through update got a different answer than publishing through create, for the same broken definition.
+
+`PUT /:name` now routes its `registerFlow` call through the same `flowDefinitionRefusal` helper `POST /` uses, so both doors answer the same `code`, `status`, and `details.fields[]` shape for the same malformed body — including the #4277 self-correcting undeclared-config-key message, which survives verbatim.
+
+**Which bodies are refused is unchanged** on both routes — a definition that registered before still registers, and every one that was refused is still refused, with the engine's own message intact.
--- a/packages/runtime/src/domains/automation-put-post-error-parity.test.ts
+++ b/packages/runtime/src/domains/automation-put-post-error-parity.test.ts
@@ -1,0 +1,314 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * #8123 — `POST /api/v1/automation` and `PUT /api/v1/automation/:name` must
+ * answer the SAME class for the SAME malformed flow definition.
+ *
+ * #8055 fixed the class on `POST /` alone: a refusal thrown by
+ * `automationService.registerFlow` used to escape as a 500 `INTERNAL_ERROR`,
+ * and is now caught and reclassified through the module-local, route-agnostic
+ * `flowDefinitionRefusal` helper into a 400 `VALIDATION_FAILED` with an
+ * ADR-0114 `details.fields[]`. `PUT /:name` makes the identical
+ * `registerFlow` call in the same file and, until this card, had no
+ * classification around it at all — so the two doors disagreed about the
+ * class of an identical refusal, the exact drift #7535's fix on the sibling
+ * `/toggle` route was shaped to avoid ("the two routes cannot disagree").
+ *
+ * ## Why a per-route pin is not the bar
+ *
+ * A suite that only asserts `PUT` in isolation (as `automation-register-
+ * error-class.test.ts` does for `POST`) would have stayed green through the
+ * exact bug this card fixes: `PUT` was individually "consistent" with its own
+ * (wrong) 500 the whole time. The pin that actually closes the drift has to
+ * DRIVE BOTH DOORS with the same body and COMPARE the two responses, so that
+ * reverting either door's classification — not just PUT's — reddens this
+ * file. See the reverse-verification note in the PR description for the
+ * measured failure text.
+ *
+ * ## The fake
+ *
+ * Same fake as `automation-register-error-class.test.ts` (#8055): three of
+ * the four cases run the REAL `FlowSchema.parse` / `validateControlFlow` from
+ * `@objectstack/spec/automation` — the very calls
+ * `AutomationEngine.canonicalizeStoredFlow` makes — and the fourth
+ * (#4277's undeclared config key) is reproduced from the engine's own
+ * construction, because that check lives in `@objectstack/service-automation`
+ * which `@objectstack/runtime` does not depend on. Duplicated here rather
+ * than imported from that file: each domain test file in this package
+ * constructs its own fake dispatcher, and importing test fixtures across
+ * suites is not the existing convention.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { FlowSchema, validateControlFlow } from '@objectstack/spec/automation';
+
+import { HttpDispatcher } from '../http-dispatcher.js';
+
+/** Config keys the fake's `notify` descriptor declares (the #4277 legal set). */
+const NOTIFY_DECLARED_CONFIG_KEYS = ['message', 'recipients', 'channel'];
+
+/**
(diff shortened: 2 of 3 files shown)


hotlong:

### The pin is bidirectional, which was the whole requirement

I made one thing binding on this dispatch: assert that the **two doors agree**, not that `PUT` is individually correct. You built it that way:

> the two doors' responses are compared directly — same `status`, same `code`, same `fields[]`, same `message` — so **reverting either door's classification (not just `PUT`'s) reddens this file**

That last clause is the difference between closing this gap and closing it *until next time*. A `PUT`-only pin would have been individually green while the two doors drifted apart again — which is exactly how this gap opened: `POST` was fixed, `PUT` was correct against its own (stale) expectation, and nothing in the suite was positioned to notice they disagreed. A table driven over both doors with the same four bodies cannot be satisfied by fixing one side.

### Three things checked that the brief did not name

- **`PUT`'s own `{ definition }` / bare-body dialect** (`body.definition ?? body`) classifies identically either way. That is a `PUT`-specific wrinkle with no `POST` counterpart, so nothing in the parity requirement would have surfaced it — you had to go looking.
- **The `.status` escape hatch is honoured on `PUT` too.** #8125 added it so a future engine-side "flow store unreachable" 503 is not reported as the author's fault; a `PUT` that ignored it would have re-opened that hole on one door only.
- **The #4277 self-correcting message survives verbatim through `PUT`**, and the raw Zod issue array does not reach the wire there either. Reclassifying to 400 while flattening the message would have destroyed the thing worth keeping.

### Reverse verification names the mechanism, not just the failure

9 of 9 `PUT`-touching tests red, and the reason is stated precisely: with the guard reverted the raw error **escapes uncaught from the unguarded call at `automation.ts:880`**, so the test's own `await` rejects rather than observing a classified response. That is the honest description of the pre-fix behaviour at this seam — the same distinction #8125 drew, and it means the red is the absence of a response rather than a wrong response. Both failure texts quoted (the `ZodError` and the #4277 message). The other 145 files / 2196 tests stayed green, including #8055's untouched `POST`-only suite — so the new file is not shadowing the old one. Restore verified by diff shape on both sides of the round trip (13 insertions, 1 deletion each time).

Scope held exactly: helper untouched so `POST`'s behaviour is unchanged, no widening, no change to which bodies are refused, and **#8054's `limit` bounds and #7968's `/runs/:runId/screen` gating left alone** — both queued behind this on the same file, and touching either would have forced a rebase on work not yet dispatched.

Not raised: `@objectstack/runtime` un-drifted from its recorded **227**. 146 files / 2205 tests green.

### What this closes

This was accepted onto the front of the `domains/automation.ts` serial chain because #8125 left a POST/PUT divergence open on my account, and "two doors answering differently" is the class this lane has filed six times today. That window is now shut with a pin that keeps it shut.

**Flipping ready and enabling auto-merge once CI converges.** Once it merges, `domains/automation.ts` is free and **#8054** dispatches, then **#7968**.

---

Issue #8197, opened 2026-08-12
driver-sql: the #7929 withhold covers the cross-field family only — every other INVALID_FILTER refusal still names the target field, which is admin-authored on a read-scope predicate


os-zhuang:
Filed **unassigned, unlabeled** for triage to grade and route — a concrete defect measured while implementing #7929's ruling B (`driver-sql` stops echoing `$field` operands), deliberately NOT fixed in that PR because it is outside what B rules on.

## What B closed, and what it did not

B's ruling is operand-shaped: *"`driver-sql`'s `INVALID_FILTER` refusal stops echoing `$field` operands"*. Implemented, that covers the three cross-field builders plus five operators whose own comparand-shape gates rendered a reference into their message. It does **not** touch the rest of the `INVALID_FILTER` family, and every one of those still names the **target field** of the refused constraint.

That matters for exactly the reason #7929 exists. The predicate a driver refuses is frequently not the caller's: the security middleware ANDs an administrator's compiled CEL rule into `opCtx.ast.where` (`plugin-security/src/security-plugin.ts`, the `$and` merge), and the driver cannot tell the two subtrees apart. On such a predicate the **target column is as admin-authored as the referent** — which is the argument B accepted for withholding both operands of a cross-field comparison. The same argument applies one step out, to every refusal that names a column.

## Measured

Real `SqlDriver` (better-sqlite3, `:memory:`), one object declaring `secret_policy_col` as `multiple: true`, driven through `driver.find`. Each of these is a predicate an RLS rule can plausibly produce, and each answers `INVALID_FILTER` / 400 naming the column:

```
REFUSED | JSON column + $in    code=INVALID_FILTER  names target? true
  Operator "$in" on field "secret_policy_col" WAS NOT APPLIED: "secret_policy_col" is a
  multi-value (or otherwise JSON-valued) field, stored by this driver as a JSON TEXT column …

REFUSED | empty field constraint    code=INVALID_FILTER  names target? true
  Field constraint at filter.secret_policy_col carries zero operators
  ({ "secret_policy_col": {} }). …

REFUSED | unbindable comparand    code=INVALID_FILTER  names target? true
  Operator "$gt" on field "secret_policy_col" requires a single comparable value, but received
  an object ({"a":1}), which cannot be bound as a SQL parameter. …
```

[...]

Issue #7598, opened 2026-08-11
[spec] `service-analytics`' read-scope / Cube filter compilers still refuse `$field`, so a CEL field-to-field RLS rule 400s on those faces


os-zhuang:
Measured while implementing #5222 (cross-field push-down). Filing unassigned — recording, not claiming.

## The fact

#5222 made `driver-sql` and `driver-sqlite-wasm` compile `{ $field: 'col' }` into a same-table column-to-column comparison for the six scalar operators. It did **not** reach `service-analytics`, whose two filter compilers are independent emitters:

- `read-scope-sql.ts` — the RLS **read-scope** lowering
- `strategies/filter-normalizer.ts` — the Cube face

Both gate comparands through `comparand-shape.ts`'s `isBindableComparand`, a deliberate hand-mirror of `driver-sql`'s module-private predicate (the file says so, and `__tests__/comparand-shape-refusal.test.ts` pins `{$field: …}` as refused "converging with `driver-sql`"). A `{ $field }` object is not bindable, so it is still refused there.

## Why it matters

This is the same defect #5041 named, at a different face. `compileCelToFilter` emits `{ $field: path }` for a field-to-field comparison in a CEL permission/RLS rule. After #5222 that rule executes on the SQL drivers and in memory — but a read scope lowered through `read-scope-sql` still answers `INVALID_FILTER` / 400.

So the capability is now **asymmetric across SQL-lowering faces**: one rule, two outcomes, chosen by which compiler the request went through. That asymmetry is narrower than the one #5222 closed, but it is the same shape.

Note the convergence comment those tests carry is now stale in one direction: it says refusing `$field` converges with `driver-sql`, and as of #5222 it diverges from it for the six scalar operators.

## What the work looks like

The #5222 rulings port directly — same-table columns only, declared-only enumeration, tenant-isolation column forbidden on both sides, same comparison class, and the same total-across-NULLs predicate shape. What each face needs its own answer for is where the declared-field set comes from: the driver reads its own `initObjects` capture, and these compilers do not have one.

The conformance corpus is reusable as-is — it is exported from `@objectstack/driver-sql` as `CROSS_FIELD_CASES` / `CROSS_FIELD_ROWS` / `CROSS_FIELD_REFUSALS` precisely so a second face can be held to the same table.

[...]

PR #8198, opened 2026-08-12
fix(driver-sql,driver-turso): a cross-field `$field` refusal stops naming the two columns it compared (#7929, #7988)

### LINKED ISSUE #7929: finding: after the #7598 Q1=B ruling, a read scope with a driver-refused field reference answers 400 from the driver — which cuts across #5367's attribution argument on that one path
Observation-class finding, filed **unassigned with no `pm:queue`** — for triage to grade and route. Raised by the dev on #7598 / PR #7916 as *"one place the ruling is slightly awkward, stated rather than papered over"*, and graded by the services PM as **not a blocker for that PR** (reasoning below). Carries a second, smaller observation from the same report.

## Observation 1 — the attribution seam

The **2026-08-12 maintainer ruling on #7598** (Q1 = B) has `NativeSQLStrategy.canHandle` decline a query whose `where` or read scope carries `{ $field: … }`, so it routes to the ObjectQL/engine path and the **driver** enforces the four #5222 rulings with metadata it owns. That is working as ruled.

The consequence: a read scope carrying a reference the **driver** refuses — an undeclared column, or the tenant-isolation column — now surfaces as **`INVALID_FILTER` / 400 from the driver**, where before the ruling the analytics path answered `READ_SCOPE_COMPILE_FAILED` / **500 with a withheld message**. The refusal moved to the driver along with the compilation.

**Why this is not a violation of Q2 = A.** Q2 = A governs `read-scope-sql`'s **own** envelope, and PR #7916 leaves it untouched: same code, same status, same withheld message, and the `"Deliberately NOT a 4xx of any flavour, including a 422"` module-header paragraph is not rewritten. Verified on the diff. The 400 comes from a *different* component that the ruling deliberately routed to.

**Why it is still worth recording.** #5367's adjudication rests on an attribution argument — a caller must not be billed a 4xx for a document they cannot author, and policy field names must not be echoed to the tenant. On this one path, the ruling's routing produces exactly the 4xx that argument was written to prevent. Two independent decisions, each correct on its own terms, meeting at a seam neither anticipated.

**Open question for whoever grades this:** does the driver's `INVALID_FILTER` on a *read-scope-originated* filter need the same withhold treatment #5367 gave the analytics path — and if so, does that belong at the driver, or at the boundary that hands a read scope to it? Note the driver cannot currently distinguish "this filter came from an author's query" from "this filter came from a read scope the caller never saw", which is likely the crux.

⚠️ This may deserve **`needs-user-decision`** rather than a dispatch: it is a consequence of a ruling the maintainer made hours ago, and re-deciding it is not a dev-seat or PM-seat call. Filed as `finding` so triage grades it; the services seat is not pre-empting that.

## Observation 2 — an unreachable-today configuration, named rather than hidden

The decline creates one configuration that could not previously occur: a host advertising `nativeSql` **without** an aggregate bridge now reaches `resolveStrategy`'s dead end for a `$field` filter (native declines, and there is nothing to fall back to).

Measured on PR #7916: **not reachable from `AnalyticsServicePlugin`**, whose defaults derive both capability flags from the bridges it actually wired. Only a hand-overridden `queryCapabilities` can construct it. The PR adds a diagnostic for it anyway, with a narrowness control. Recorded here so the reasoning survives the session; **no action is claimed to be needed today.**

## Explicitly NOT claimed

- **No regression is asserted.** PR #7916's 26 CI checks are green, including the `#5367` end-to-end pin `analytics-query-read-scope-withhold.test.ts` in `runtime` (2096/2096). The withhold behaviour on the path #5367 actually adjudicated is intact.
- **The ruling is not being re-litigated here.** Q1 = B and Q2 = A stand. This card records a downstream consequence, not an objection.

## Evidence

Measured 2026-08-12 by the dev on PR #7916 (`packages/services/service-analytics`, head `94c5580`). Full measurement in the `<!-- os-dev-report -->` comment on #7598 (comment 5262307982); the PM's acceptance reasoning is in comment 5262400038.

## Source

Split out of objectstack-ai/objectstack#7598 (PR objectstack-ai/objectstack#7916) — the dev's named residuals, deliberately not decided inside that card.

Refs: #5367 (the withhold adjudication), #5222 (the four cross-field rulings), #7596, #7597.

### LINKED ISSUE #7988: finding: the #7929 read-scope disclosure is NOT analytics-only — an RLS `$field` filter injected by the security middleware discloses the same policy column on the ordinary CRUD path, and predates #7598
Observation-class finding, filed **unassigned with no `pm:queue`** — for triage to grade and route. Measured while executing the #7929 measurement dispatch (`domain:engine-core` seat, 2026-08-12). Recording, not claiming.

Blocked-by: #7929

## Why this is filed separately from #7929

#7929's body, the triage block, and the **2026-08-12 maintainer ruling on it** all frame the disclosure as a *consequence of #7598's Q1 = B routing* — "the refusal moved to the driver along with the compilation" — and the ruling's implementation boundaries scope the pins accordingly:

> Pin both directions: (a) a read scope carrying a driver-refused reference answers the withheld envelope with no field-name echo — **extend the `analytics-query-read-scope-withhold` pin family to this path**

That framing is incomplete, and a fix scoped to it would leave the larger half open. The **ordinary CRUD read path has the identical disclosure, and did not need #7598 to get it.** It arrived with #5222 (2026-08-06), when `driver-sql` learned to compile `{ $field }` and to refuse it with `INVALID_FILTER` / 400.

The two paths differ only in **which component merges the admin's read filter into the query**:

| path | merge boundary | file |
|---|---|---|
| analytics `POST /analytics/query` | `ObjectQLStrategy.withReadScope` | `packages/services/service-analytics/src/strategies/objectql-strategy.ts` (`withReadScope`) |
| ordinary CRUD read | security middleware ANDs into `opCtx.ast.where` | `packages/plugins/plugin-security/src/security-plugin.ts:2198` |

Both hand the driver a bare `FilterCondition` with no provenance mark, and `driver-sql`'s `applyCrossFieldComparison` refuses identically in both cases.

## Measured

Real `ObjectQL` engine, real `SqliteWasmDriver` seeded with the shared `CROSS_FIELD_*` corpus fixture, and a read-filter middleware shaped exactly like the security/sharing ones (`ast.where = ast.where ? { $and: [ast.where, scoped] } : scoped`) injecting an admin-authored policy predicate the caller never wrote:

```
scoped = { amount: { $gt: { $field: 'secret_policy_column' } } }
caller  = ql.find('cross_field_deal', { where: { stage: 'won' } })
```

What the engine surfaces:

```
code:    INVALID_FILTER
status:  400
message: Operator "$gt" on field "amount" compares against another field
         ({ "$field": "secret_policy_column" }), which cannot be compiled here:
         "secret_policy_column" is not a declared field of "cross_field_deal" —
         only declared fields can be referenced. Cross-field comparison on SQL
         push-down supports same-table columns the object declares, compared as
         the same type class, excluding the tenant-isolation column.

names the admin-authored column: true
```

Producer chain from the captured stack, verbatim:

```
unsupportedFilterError            (driver-sql/src/sql-driver.ts:554)
uncompilableFieldReferenceError   (driver-sql/src/sql-driver.ts:1138)
applyCrossFieldComparison         (driver-sql/src/sql-driver.ts:9160)
applyFilterCondition              (driver-sql/src/sql-driver.ts:8870)
```

The 400 is a **declared 4xx**, so the REST boundary's withhold — `declaresServerFault`, which requires `status >= 500` — can never reach it. That tiering is deliberate and already pinned (`analytics-query-read-scope-withhold.test.ts`, the "a DECLARED 4xx is untouched — the withhold is 5xx-only" case), so this is not a boundary bug; it is the same attribution seam #7929 records, one path over.

## Scope of the exposure

Requires an administrator to have authored a CEL permission/RLS rule with a **field-to-field comparison** — `compileCelToFilter` emits `{ $field: path }` for exactly that — whose reference fails one of the four #5222 rulings (dotted path, undeclared column, the tenant-isolation column on either side, or a cross-class comparison). Narrow, and no customer report; but on the CRUD path it reaches **every** SQL-driver read, not just the analytics face.

## Explicitly NOT claimed

- **Not a regression, and not #7598's doing.** #7598 Q1 = B is not implicated on this path at all — the CRUD read has gone straight to `driver-sql` since long before it. #5222's refusal arm is correct on its own terms; what is missing is the provenance to know the predicate was not the caller's.
- **Neither ruling is re-litigated.** #5222, #5367, #7598 Q1 = B and Q2 = A all stand. This records a scope fact about the remedy, not an objection to any of them.
- **No remedy is proposed here.** The 2026-08-12 ruling on #7929 already picked the shape (provenance marker at the merge boundary, driver withholds on a marked predicate). This card only records that "the merge boundary" is **two** boundaries in two different packages, and that pinning only the analytics one would leave the CRUD one open.

## Refs

- #7929 (the attribution seam, and the 2026-08-12 ruling that picked the provenance-marker shape)
- #5367 (the withhold adjudication whose invariant this breaks)
- #5222 (the four cross-field rulings, and the `INVALID_FILTER` / 400 refusal arm)
- #7598 (the Q1 = B routing that brought the analytics face into the same seam)

os-zhuang:
Fixes #7929
Fixes #7988

Implements **B** of the maintainer ruling of 2026-08-12 on #7929 (comment 5266067250), verbatim provenance 「接受你的全部建议。」 — adopting "B now, A next". `driver-sql`'s `INVALID_FILTER` refusal stops echoing `$field` operands for every caller; the full diagnostic goes to the server-side log.

Because the withhold lands at the **consumer**, it covers both merge boundaries at once — the analytics read-scope merge (`service-analytics`) and the ordinary CRUD read (`plugin-security`), which is why #7988 closes here too. Both are pinned; see "Pins" below.

## What a caller gets now, and what moved

`code: INVALID_FILTER` and `status: 400` are unchanged, and the set of refused filters is unchanged — verified operator by operator, including the six scalar comparisons that still compile.

**Withheld** (moved to the log): the referenced column, the target column, the operator the filter used, the list index, and the boundary `reason` — the sentence that named the object, the declared types and, on the tenant arm, *which* column is the tenant-isolation column.

**Kept** on the wire: the envelope, which of the three cross-field refusal classes fired (position / bare form / boundary), and the capability statement — same-table declared columns, same type class, tenant-isolation column excluded. None of it is derived from what was sent. Example:

> A cross-field comparison ({ "$field": … }) in this filter cannot be compiled here. Cross-field comparison on SQL push-down supports same-table columns the object declares, compared as the same type class, excluding the tenant-isolation column. The columns, the operator this filter used and the specific reason are withheld from the message (#7929); the full diagnostic is in the server log.

### Why BOTH operands, not only the `$field` one

On a read-scope refusal the whole predicate is the administrator's. In `{ amount: { $gt: { $field: "secret_policy_column" } } }` the target `amount` is as policy-authored as the referent, so echoing the target would have left half the disclosure live. The driver has no way to tell the halves apart: the scope arrives as a bare `FilterCondition` in `where`, `DriverQuery` is `Omit< QueryAST, 'object' >` with no provenance slot, and the only thing that does cross (`context`) says *who* is asking, never *which subtree they did not write*.

### ⚠️ The honest-author case is an accepted cost, stated plainly

An author debugging their **own** cross-field filter now gets the same redacted message. That is a real diagnostic regression, and it is what B buys until A (the spec-declared provenance mark, a separate card) restores the author-facing text behind a real mark. Their diagnostic is relocated, not destroyed: the full text is in the server log. The changeset says this in the same words, and the equality is **pinned** — an author-written filter and a policy-injected one must produce byte-identical messages, so "the author still sees the columns" cannot come back by accident.

## The echo sites — the dispatch named three, measurement found six

The dispatch flagged its own list as a PM assumption and asked for an independent enumeration. Probing every operator with a `{ $field }` comparand against a real driver, on `origin/main`:

| # | builder | echoes | in the dispatch's table? |
|---|---|---|---|
| 1 | `crossFieldComparisonError` | `op`, target, ref, list index | yes |
| 2 | `bareFieldReferenceError` | target, ref — twice, the second inside a suggested corrected filter | yes |
| 3 | `uncompilableFieldReferenceError` | `op`, target, ref, `reason` | yes |
| 4 | `icontainsComparandError` (`$icontains`) | target + the comparand's JSON, i.e. the ref | **no** |
| 5 | the `$like` / `$ilike` pattern gate | target + the comparand's JSON | **no** |
| 6 | the `$null` / `$exists` boolean gates | target + the comparand's JSON | **no** |

Sites 4-6 sit on the pre-emitter reduction walk, which answers before the emitter the other cases reach — so five operators disclosed and four did not, decided by nothing a caller can see. They now route to the cross-field refusal, which is what the condition actually is. One condition, one answer, and the redacted one; no filter's accept/reject verdict changes.

The dispatch's reading of `reason` as carrying the disclosure (its point B) is confirmed against all four captured bodies: redacting only the `{"$field": …}` interpolation would have left `"organization_id" is the tenant-isolation column of "cross_field_deal"` live.

## The server-log fork (the dispatch's point C), resolved two different ways

- **`driver-sql` HAS a sink**: `SqlDriver.logger`, injected by hosts and spied by tests. The redacted refusal carries its full text on a **symbol key** (`Symbol.for('objectstack.driver-sql.withheldFilterDiagnostic')`), and the driver writes it to `logger.warn`. A symbol because the carrier must not travel: `JSON.stringify`, object spread and structured clone all skip it, so an error mapper cannot put the text back on the wire.
- **⚠️ The seam is not where I first put it.** A catch at the filter entry point missed every refusal raised inside a knex WHERE-group callback — knex invokes those lazily, at compile time, after the entry point has returned. Measured: an `$or`-nested refusal reached the caller correctly redacted while the log stayed silent. The catch is now at the group callbacks too, with a once-marker so one error is logged once. Re-measured: every case logs exactly one line.
- **`driver-turso`'s `RemoteTransport` had NO sink** — no logger, no `console` call, anywhere in the class. No logging subsystem was invented for it: `TursoDriver extends SqlDriver`, so it hands its own `logger.warn` down through the same setter pattern the class already uses for its client, connect factory and temporal column rule.

## driver-turso is in scope, by measurement

The dispatch's point E said no copies of the message text exist in the sibling drivers, and asked me to check whether turso's remote transport has its own refusal path. **It does** — `uncompilableComparand`, its own wording, echoing `'deal.amount' $gt {"$field":"budget"}`, on a path whose own doc names RLS push-down as a caller. `driver-sqlite-wasm` has no copy and inherits the fix (confirmed by running the shared corpus through it). Leaving remote mode disclosing would have made the exposure a property of the connection string.

## Pins

- `packages/runtime/src/cross-field-refusal-operand-withhold.test.ts` (new) — **(a)** the four captured response bodies from #7929, driven through the real `POST /api/v1/analytics/query` route with a real `AnalyticsService` and a real `SqliteWasmDriver`, with `executeRawSql` call count 0 proving the #7598 decline actually happened; **(b)** #7988's CRUD measurement — a real `ObjectQL` engine with a security-middleware-shaped `$and` injection; **(c)** the author case, as a byte-equality between the policy-injected and author-written messages.
- The two cross-field conformance suites (`driver-sql`, `driver-sqlite-wasm`) now assert both halves per corpus case: the caller message names no operand, and the server log still carries the fragment that says which ruling bit. The shared corpus field was renamed `messageIncludes` → `diagnosticIncludes`, since that is now what it describes.
- `driver-turso`'s two disclosure fixtures were inverted rather than deleted — one of them pinned the echoed operands exactly — plus a new pin that serialising the error cannot recover the withheld text.

## Verification

`pnpm test` (real output in the report comment on #7929): driver-sql 1399 passed, driver-turso 947, driver-sqlite-wasm 374, service-analytics 1655, runtime 146 files / 2203 tests. #5367's end-to-end pin `analytics-query-read-scope-withhold.test.ts` is intact at 10/10. `check:driver-conformance` reads 40 covered / 0 DEBT / 0 exempt before and after. Typecheck green on all four packages that have the script; ESLint clean on every changed file.

**Reverse verification**, direction predicted first: restoring the operand echo in `uncompilableFieldReferenceError` should redden the disclosure assertions while every envelope assertion stays green. Observed exactly that — 7 of the 8 new pins red, the 8th (the author-equality pin) green as predicted, since both sides regain the names. First attempt reported a false green from a stale `dist`; rebuilt and re-run.

## Not in this PR

- ⛔ No provenance mark anywhere (that is A, a separate card). Nothing added to `packages/spec`, `QueryAST` or `DriverOptions`.
- ⛔ The REST boundary's 5xx-only withhold is untouched.
- ⛔ `read-scope-sql` is untouched. Measured incidentally: a list-position reference inside a read scope never reaches the driver — it is refused as `READ_SCOPE_COMPILE_FAILED` and the boundary withholds it, exactly as adjudicated.
- One residual filed as #8197: refusals outside the cross-field family still name the target field, which is admin-authored on a policy predicate.

---

### DIFF
--- a/.changeset/withhold-cross-field-operands.md
+++ b/.changeset/withhold-cross-field-operands.md
@@ -1,6 +1,6 @@
 ---
-"@objectstack/driver-sql": patch
-"@objectstack/driver-turso": patch
+"@objectstack/driver-sql": minor
+"@objectstack/driver-turso": minor
 ---
 
 fix(driver-sql,driver-turso): a cross-field `$field` refusal stops naming the two columns it compared (#7929, #7988)
--- a/.changeset/withhold-cross-field-operands.md
+++ b/.changeset/withhold-cross-field-operands.md
@@ -1,0 +1,45 @@
+---
+"@objectstack/driver-sql": patch
+"@objectstack/driver-turso": patch
+---
+
+fix(driver-sql,driver-turso): a cross-field `$field` refusal stops naming the two columns it compared (#7929, #7988)
+
+`INVALID_FILTER` / 400 is unchanged, and every filter that was refused is still
+refused. What the caller no longer receives is the **predicate**: the referenced
+column, the target column, the operator, the list index, and the boundary reason.
+The full diagnostic now goes to the driver's server-side log instead
+(`SqlDriver.logger`, the sink a host already injects; `TursoDriver` hands the
+same sink to its remote transport).
+
+**Why.** An administrator's CEL sharing/permission rule compiles to
+`{ $field: path }` and is ANDed into the caller's query by the security
+middleware (ordinary CRUD reads) or by the analytics read-scope merge. The driver
+receives one `FilterCondition` with nothing marking which subtree the caller
+wrote, so when the reference failed one of the four cross-field rulings the
+refusal handed a tenant an administrator's policy — measured end to end: the
+referenced column, the column it was compared against, and, on the tenant arm,
+a sentence naming **which column is the tenant-isolation column** of the object.
+A dotted reference came back as `sharing_rule.manager_budget`, verbatim, inside
+`error.message`.
+
+**⚠️ This is a real diagnostic regression for authors, and it is deliberate.**
+An author debugging their **own** cross-field filter now gets the same redacted
+message — nothing in the query tells the driver whether the reference was theirs
+or a policy's, so the withhold cannot be conditional without inventing a guess.
+Their message is not destroyed, it is relocated: the full text, naming both
+columns, is in the server log for whoever operates the deployment. A follow-up
+card restores the author-facing text behind a spec-declared provenance mark set
+at both merge boundaries; until it lands, an author debugging a cross-field
+filter needs the server log or a `matchesFilter` run in memory.
+
+What a caller still gets: the same `code` and `status`, which of the three
+cross-field refusal classes fired, and the capability statement (same-table
+declared columns, same type class, tenant-isolation column excluded) — none of
+which is derived from the filter that was sent.
+
+Scope note: five operators used to answer a `{ $field }` comparand with their own
+comparand-shape refusal (`$icontains`, `$like`/`$ilike`, `$null`, `$exists`),
+each rendering the reference into its message, while the same reference at
+`$contains` was answered by the cross-field refusal. They now all answer with the
+cross-field refusal — one condition, one answer, and the redacted one.
--- a/packages/drivers/driver-sql/src/cross-field-conformance-cases.ts
+++ b/packages/drivers/driver-sql/src/cross-field-conformance-cases.ts
@@ -331,8 +331,17 @@ export const CROSS_FIELD_AUTHORED_CASES: readonly CrossFieldAuthoredCase[] = [
  * SECURITY surface rather than a capability one.
  *
  * Every entry must throw `INVALID_FILTER` / 400 (ADR-0112) on BOTH SQL
- * drivers. `messageIncludes` pins the part of the wording a caller needs to
- * act on; the tests assert the envelope regardless.
+ * drivers. `diagnosticIncludes` pins the wording that says WHICH ruling bit;
+ * the tests assert the envelope regardless.
+ *
+ * [#7929, maintainer ruling 2026-08-12] That wording moved. It used to be
+ * `messageIncludes` — substrings of the message the CALLER receives — and the
+ * rename is the change, not a tidy-up: those sentences name the two columns of
+ * the comparison, and on a read-scope refusal both were written by an
+ * administrator whose policy the tenant never saw. The refusal now answers the
+ * caller with an operand-free sentence and puts this text in the server log, so
+ * these fragments are asserted against the LOGGED diagnostic. A test that finds
+ * one of them in `error.message` is finding the disclosure this card closed.
  *
  * Read this table together with {@link CROSS_FIELD_CASES}: what makes the
  * refusals defensible is that the supported arm above is proven equivalent, so
@@ -342,8 +351,11 @@ export const CROSS_FIELD_AUTHORED_CASES: readonly CrossFieldAuthoredCase[] = [
 export interface CrossFieldRefusalCase {
   name: string;
   filter: unknown;
-  /** Substrings the refusal message must contain. */
-  messageIncludes: string[];
+  /**
+   * Substrings the SERVER-LOG diagnostic must contain (#7929) — never the
+   * caller-visible message, which names no operand at all.
+   */
+  diagnosticIncludes: string[];
   /**
    * Why the shape is refused — surfaced in failure output. Optional because
    * several entries are one spelling of a reason the entry above them states
@@ -358,13 +370,13 @@ export const CROSS_FIELD_REFUSALS: readonly CrossFieldRefusalCase[] = [
   {
     name: 'a dotted relation path is refused',
     filter: { amount: { $gt: { $field: 'account.budget' } } },
-    messageIncludes: ['dotted path', 'same-table'],
+    diagnosticIncludes: ['dotted path', 'same-table'],
     note: 'Maintainer ruling (2026-08-06) point 1 = A: no JOIN planning (disproportionate) and no alias-qualified columns (no alias contract). The memory evaluator DOES walk the path, so this is a deliberate, loudly-reported asymmetry rather than a silent one.',
   },
   {
     name: 'a dotted path is refused even when its head names a real column',
     filter: { amount: { $gt: { $field: 'budget.nested' } } },
-    messageIncludes: ['dotted path'],
+    diagnosticIncludes: ['dotted path'],
     note: 'The refusal is on the SHAPE, not on whether the first segment happens to resolve — otherwise the check would depend on data the compiler cannot see.',
   },
 
@@ -372,13 +384,13 @@ export const CROSS_FIELD_REFUSALS: readonly CrossFieldRefusalCase[] = [
   {
     name: 'an undeclared column is refused at compile time',
     filter: { amount: { $gt: { $field: 'no_such_column' } } },
-    messageIncludes: ['not a declared field'],
+    diagnosticIncludes: ['not a declared field'],
     note: 'The `$field` value lands in a SQL IDENTIFIER position. cloud#1051: letting it through unchecked is dismantling the guard rail — and a compile-time refusal is what makes AI-authored metadata wrong at authoring time rather than in the database.',
   },
   {
     name: 'an undeclared TARGET field is refused too',
     filter: { no_such_column: { $gt: { $field: 'budget' } } },
-    messageIncludes: ['not a declared field'],
+    diagnosticIncludes: ['not a declared field'],
     note: 'A comparison is one surface — validating only the referent would leave half of it unchecked, and the type-class rule below needs both declarations anyway.',
   },
 
@@ -386,13 +398,13 @@ export const CROSS_FIELD_REFUSALS: readonly CrossFieldRefusalCase[] = [
   {
     name: 'the tenant-isolation column is refused as the REFERENT',
     filter: { stage: { $eq: { $field: 'organization_id' } } },
-    messageIncludes: ['tenant-isolation column'],
+    diagnosticIncludes: ['tenant-isolation column'],
     note: 'The named ruling. A comparison against the isolation column is a privilege-escalation comparison surface: it lets a filter probe the tenant boundary the driver injects rather than being scoped by it.',
   },
   {
     name: 'the tenant-isolation column is refused as the TARGET',
     filter: { organization_id: { $eq: { $field: 'stage' } } },
-    messageIncludes: ['tenant-isolation column'],
+    diagnosticIncludes: ['tenant-isolation column'],
     note: 'Closed because the operands of `=` COMMUTE — a ban that a swap of the two sides walks around is not a ban. Ruling names the referent; this is the same surface spelled backwards.',
   },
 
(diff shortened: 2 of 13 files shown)


os-zhuang:
**CI triage — `Check Documentation Links` red on `86fd6de`: infrastructure, not this diff. Re-run once, no code touched.**

**Signature.** The job died in `lychee-setup`, before any link was examined:

```
Downloading from: https://github.com/lycheeverse/lychee/releases/download/lychee-v0.24.2/lychee-x86_64-unknown-linux-gnu.tar.gz
##[error]Process completed with exit code 22.
```

Exit 22 is curl's "HTTP error returned by server" — the release tarball fetch failed. Both later steps (`Install lychee`, `Run Lychee`) are `skipped`, so **zero links were checked**; the red says nothing about this PR's content, and this diff touches no docs at all.

**Checked whether it is systemic before spending a re-run** — it is not. Recent `check-links.yml` runs:

| created | conclusion | branch |
|---|---|---|
| 19:53:08Z | **failure** | `claude/issue-7929-withhold-field-operands` ← this PR |
| 19:50:40Z | success | `claude/issue-8019-…` |
| 19:24:01Z | success | `claude/issue-8080-…` |
| 19:23:32Z | success | `claude/issue-8104-…` |
| 19:13:12Z | success | `claude/issue-8060-…` |

A success 2.5 minutes earlier and three more around it ⇒ a transient upstream fetch failure on one runner, not a broken pin or a repo-wide outage. **Disposition: re-run failed jobs once** (run `31634860387`, queued). ⛔ Not repeated — if the same signature returns on the same head SHA, it stops being transient and becomes a real problem to diagnose rather than re-roll.

**New signature**, not the lane's known flaky (`#7282`, metadata-fs watcher). Recorded here rather than added to the steward's ledger — ⛔ that table is human-escalated only, and one occurrence is not a flake classification.

**The other two bot comments need no action.** Vercel's is a skipped deployment. The Docs Drift Check is **advisory only** and correctly identifies that `content/docs/releases/implementation-status.mdx` is release-owned and read-only — ⛔ this PR does not touch `content/docs/` at all, which is the required posture, not an omission to fix.

⛔ **Nothing was pushed to this branch.** The dispatched dev still owns this worktree and has not yet filed its report; a second writer on the same branch is the failure mode that discipline exists to prevent. Review of the change itself comes after the report lands.

---

Issue #7889, opened 2026-08-12
[finding] alias-integrity cannot see an alias row made unreachable by a guidanceSet — a vis-shaped alias on VISIBILITY_STRICT_OPTIONS surfaces would be dead code with no gate saying so


os-zhuang:
Recorded from PR #7884's report (out_of_scope_finding 3 of card #7832; filed by the PM seat because the cloud dev session cannot write issues). Filed unassigned — a recording, not a claim.

## Mechanism, measured during #7832

In `packages/spec/src/shared/strict-object.ts`, when an unknown key matches a guidanceSet's pattern, the set **consumes the key and `continue`s past the rename channel** — the `aliases` table is never consulted for that key. `VISIBILITY_KEY_PATTERN` (`/vis|conceal|hidden|show.?when/i`, part of the ADR-0089 conditional-visibility prescription in `VISIBILITY_STRICT_OPTIONS`) therefore makes ANY vis-shaped alias row on those surfaces (`FormSectionSchema`, `PageComponentSchema`, `FormFieldSchema`) dead on arrival.

That behaviour is **correct** (the shared prescription is the intended answer), but it is not enforced or documented at the alias-authoring site:

- `alias-integrity.test.ts` checks that an alias row's **target key is declared** — it does not check that the row is **reachable** past the guidanceSets in the same strict options.
- A future card adding e.g. `visible: 'visibleWhen'` to one of those shapes would ship dead code, every gate green, and the author would reasonably believe the hint exists.

PR #7884 worked around it by (a) placing its one view-family alias (`disabled → readonly` on `FormFieldSchema`) outside the pattern's reach, and (b) pinning in `visible-when-alias-guidance.test.ts` that the three shared-options surfaces answer `visible`/`showWhen` via the ADR-0089 set with **no** `Did you mean` rename — but that pin protects those specific keys only, not the class.

## Possible fix shape (for whoever grades this)

Extend alias-integrity to fail an alias key that a guidanceSet in the same strict options already matches — turning "dead row" from a silent state into a red gate. Lands in the gate/test machinery around `strict-object.ts` ⇒ `domain:spec-tooling` face by the #5163 boundary (routing is the triage seat's call).

## Refs

#7832 / PR #7884 (provenance, workaround, partial pins), #6619 (where the ADR-0089 guidanceSets were added), ADR-0089.

Dedup: searched for alias-integrity / guidanceSet / VISIBILITY_KEY_PATTERN reachability — zero hits.

[...]

Issue #7832, opened 2026-08-11
spec-surface: alias guidance for `visible`/`showWhen`/`disabled` on the `visibleWhen` shapes — the reverse direction of the action shapes' existing table (#7816 ask 1)


os-zhuang:
Provenance: split out of #7816 (its "ask 1") by the triage seat — the standalone, no-behaviour-change half. #7816 itself keeps the canonical-spelling decision (its "ask 2"); this card does not depend on that ruling.

## Scope

Give the `visibleWhen`-family strict shapes the same `aliases` curation the action shapes already carry in the opposite direction, so `visible` / `showWhen` / `disabled` written on the wrong surface produces an error that names the key the author should have used:

- `RowCrudActionOverrideSchema` (`packages/spec/src/data/object.zod.ts`, `userActions.edit`/`delete` overrides) — today a plain `.strict()` with no `aliases`/`guidance` at all; `userActions.delete.visible` and `.disabled` get bare unrecognized-key errors (measured in #7816). ⚠️ The boolean case should point at `enabled`, not at `visibleWhen`, or the hint just moves the confusion (#7816's own note).
- Field-level and per-option predicates (`packages/spec/src/data/field.zod.ts:814` and `:176`), form sections (`ui/view.zod.ts`), page components (`ui/page.zod.ts`): aliases `visible → visibleWhen`, `showWhen → visibleWhen`, `disabled → disabledWhen`.

## Mechanism, verified on `origin/main` @ `098b629`

`strictObject`'s `aliases` table is error-message curation only — "semantic near-misses edit distance cannot reach" (`packages/spec/src/shared/strict-object.ts`), and its own doc example is exactly this key pair in the other direction (`aliases: { visibleWhen: 'visible' }`). Nothing that parses today changes; acceptance is byte-identical before/after ⇒ `domain:spec-surface` under the acceptance-surface criterion; changeset patch.

Measured rejections today (from #7816, spec `dist` on `main`): `field.visible`, `userActions.delete.visible`, `userActions.delete.disabled` — none of the messages mention `visibleWhen` / `disabledWhen` / `enabled`.

## Out of scope

Choosing ONE canonical spelling (widening any schema, the boolean arm, ADR-0087 retirement of either form, ADR-0089's reach over action shapes) — that is #7816's decision. If that ruling later converges the vocabulary, these alias rows remain the right rows: they then carry the migration hint.

Refs #7816 (parent question), #5970 (action-side `visible` unification), ADR-0089.

[...]

Issue #7816, opened 2026-08-11
One intent, two spellings: `visible` (actions) vs `visibleWhen` (fields / sections / userActions) — and the alias guard only covers one direction


baozhoutao:
## Summary

One intent — "show this thing only when the predicate holds" — is spelled **two different ways** depending on which schema you are authoring, and the guard that catches the wrong spelling only runs in **one direction**.

* Actions (and action params) declare it as **`visible`** / **`disabled`** — `ActionSchema` ([`ui/action.zod.ts:1131`](https://github.com/objectstack-ai/objectstack/blob/main/packages/spec/src/ui/action.zod.ts#L1131)), `ActionParamSchema` (L383).
* Everything else declares it as **`visibleWhen`** / **`disabledWhen`** — fields ([`data/field.zod.ts:814`](https://github.com/objectstack-ai/objectstack/blob/main/packages/spec/src/data/field.zod.ts#L814)), form sections ([`ui/view.zod.ts:1757`](https://github.com/objectstack-ai/objectstack/blob/main/packages/spec/src/ui/view.zod.ts#L1757), L1858), page components ([`ui/page.zod.ts:150`](https://github.com/objectstack-ai/objectstack/blob/main/packages/spec/src/ui/page.zod.ts#L150)), per-option predicates (`field.zod.ts:176`), and the built-in row CRUD overrides `userActions.edit/delete` ([`data/object.zod.ts:1079`](https://github.com/objectstack-ai/objectstack/blob/main/packages/spec/src/data/object.zod.ts#L1079)).

The action schemas already recognise the *other* spelling and rename it: `actionObject()` registers `visibleWhen → visible`, `showWhen → visible`, `disabledWhen → disabled` (`action.zod.ts:776`), and `ACTION_PARAM_KEY_ALIASES` maps `visiblewhen/visibleon/visibility → visible` (L74). The comment there states the motive exactly:

> ADR-0089 made `visibleWhen` the canonical predicate on view/page schemas. An author who learned it there would silently lose a param's capability gate here.

**The reverse alias does not exist.** An author who learns `visible` on an action and writes it on a field or on `userActions.delete` gets a bare unknown-key error that never names the key they should have used.

## Measured (spec `dist` on `main`, `ObjectSchema.create`)

```
ACCEPTED  field.visibleWhen                          (canonical)
REJECTED  field.visible
          Unrecognized key(s) on this field: `visible`. Until #4001 closed this shape
          these were dropped silently — …
[...]

Issue #6619, opened 2026-08-08
[spec] Fold the three hand-written unrecognized_keys error maps into `strictObject` guidance — requires a set-keyed guidance form, and closes the alias-integrity gate's blind spot (#6416 direction 2)


os-project-manager:
Follow-up card for #6416's **direction 2**, deliberately split out of that card by triage (2026-08-07T17:54:18Z: "Direction 2 is a bigger card, not this one … a shared-mechanism change and deserves its own card rather than arriving as scope creep on the reorder"). Direction 1 (the in-place reorders) landed as PR #6453 (`bf0ae9976`).

## What this card does

Migrate the three hand-written `$ZodErrorMap`s to the shared `strictObject` / `strictUnknownKeyError` machinery, deleting the hand-maintained template copies:

- `packages/spec/src/shared/visibility.ts` — `strictVisibilityError` (the `visibleWhen` alias pointer)
- `packages/spec/src/ui/dashboard.zod.ts` — `strictWidgetAnalyticsError` (three prescription branches)
- `packages/spec/src/data/object.zod.ts` — `strictTenancyError` (per-key tombstone bullets)

## The real decision inside it (why it is not mechanical)

`dashboard.zod.ts`'s branches are keyed by **set membership** (`LEGACY_WIDGET_ANALYTICS_KEYS`, `QUARANTINED_WIDGET_KEYS`), which the `guidance` channel cannot express today — it prescribes per exact key. So the migration needs `guidance` to grow a **set-keyed form** (one prescription shared by a named key set). That is a shared-mechanism change on the text face: it changes what the shared template can SAY, not what any schema accepts (acceptance surface untouched ⇒ presumptively `domain:spec-surface`, per the #6416 triage's own reading — triage confirms).

## Why it is worth doing (the gate blind spot)

`alias-integrity.test.ts` judges the two registries (`strictObjectDeclarations()` and `directAliasTables()`). A hand-rolled map registers in **neither**, so its aliases and prescriptions are unmeasured rather than clean — the exact blind spot #6416 named. Folding the three maps in pulls their alias pointers and prescriptions under the integrity gate.

[...]

PR #8199, opened 2026-08-12
docs(spec): a form section / page component gates visibility only — say so, and tell `disabled` where it belongs (#7887)

### LINKED ISSUE #7887: [finding] `disabled` has no vocabulary at all on FormSectionSchema / PageComponentSchema — no readonly/disabled/disabledWhen key to point an author at
Recorded from PR #7884's report (out_of_scope_finding 2 of card #7832; filed by the PM seat because the cloud dev session cannot write issues). Filed unassigned — a recording, not a claim.

## Observation, measured on the #7884 branch against `origin/main`

While curating alias/guidance rows for `visible`/`showWhen`/`disabled` on the `visibleWhen`-family shapes (#7832), the dev found that `FormSectionSchema` and `PageComponentSchema` (and `SelectOptionSchema`) declare **no disabled-ish key whatsoever** — no `disabled`, no `disabledWhen`, no `readonly`. There is no way to express "this section / page component is disabled/read-only" in view/page metadata.

Consequence for #7832's scope: `disabled` written on those shapes is rejected loudly (strict shapes) but with **no pointer**, because there is nothing on the shape to point at — an alias row would name a key the shape rejects next (`alias-integrity.test.ts` fails such a row outright). PR #7884 pinned these as **deliberate gaps**: `visible-when-alias-guidance.test.ts` section 3 fails if any of the three shapes ever gains a disabled-ish key, so the owed alias row becomes visible the moment it becomes writable.

## The question

Whether this is a **deliberate design boundary** (sections/components gate visibility only; disabling is a field-level concern via `readonly`/`readonlyWhen`) or a **vocabulary gap** is an acceptance-surface question for `domain:spec`, not for the surface seat — adding any such key would widen the accepted metadata set.

If it is ruled a boundary: worth one sentence of prose on the shapes saying so (that half would return to `domain:spec-surface`). If it is ruled a gap: the #7884 tripwire pins already mark where the alias rows become owed.

## Refs

#7832 / PR #7884 (provenance, tripwire pins), #7816 (canonical-spelling question, related vocabulary), #7751 (page-component props passthrough — adjacent but distinct: that is about the `properties` bag, this is about the shape's own keys).

Dedup: searched open issues for FormSectionSchema/PageComponentSchema + disabled/readonly vocabulary — nothing beyond the refs above.

os-zhuang:
Fixes #7887

## The ruling being implemented

Maintainer ruling of 2026-08-12 (comment 5271005260), operative sentence, verbatim:

> Ruling: **boundary, not gap.** `FormSectionSchema` / `PageComponentSchema` gate **visibility only**; editability lives on fields. No `disabled` / `readonly` / `disabledWhen` slot is added to those shapes, and no alias row is registered for them — adding one would declare a key the runtime does not honour, the exact ADR-0049 class being retired elsewhere.

So this PR adds **no key and no alias row**. It documents the boundary on both shapes and gives the existing rejection a guidance string naming where the key does belong.

Before, on a section:

```text
Unrecognized key(s) on this view/page schema: `disabled`. Before ADR-0089 D3a these
were dropped silently, shipping inert metadata; a mis-layered or stale key is now a
loud parse error.
```

After:

```text
Unrecognized key(s) on this view/page schema: `disabled`.
  • Editability is a FIELD-level concern. This shape gates VISIBILITY only — a
    deliberate boundary, not a missing key (#7887): a section / page component has no
    read-only semantics of its own to enforce. Write `readonly: true` (or the
    conditional `readonlyWhen` predicate) on the form field(s) inside it instead; to
    hide the whole section or component, use `visibleWhen`. Before ADR-0089 D3a these
    were dropped silently, shipping inert metadata; a mis-layered or stale key is now
    a loud parse error.
```

It points at **`readonlyWhen`** and never at `disabledWhen` — verified on `origin/main`, not inherited from the card: `data/field.zod.ts` renames `disabled` to `readonly` and its own comment records that "a field has `readonlyWhen`, not `disabledWhen`" (#7832). A prescription naming `disabledWhen` would hand the author their next rejection. One pin asserts the prescription text contains `readonlyWhen` and another asserts it never contains `disabledWhen`.

## Where the guidance landed, and why not the obvious place

`VISIBILITY_STRICT_OPTIONS` is shared by **three** shapes: `FormSectionSchema`, `PageComponentSchema`, and `FormFieldSchema`. Filing the boundary family in that shared table is one line shorter and is wrong in three separate ways, all of them measured rather than argued (I ran the counterfactual — the set hoisted into the shared table — and reverted it):

1. **It silently destroys the one correct pointer in the family.** `strictUnknownKeyError` consults exact `guidance`, then `guidanceSets`, then `aliases`, and a set match `continue`s past the rename channel. `FormFieldSchema` answers `disabled` through its own `aliases: { disabled: 'readonly' }` row, and `readonly` is real there. Measured on the counterfactual: the sentence *Did you mean* `disabled` → `readonly`? disappears, and a field author is redirected away from the surface where the key they want actually lives.
2. **`alias-integrity.test.ts` goes red** — the #7889 reachability check, which landed this morning:
   ```text
   "this view/page schema": alias `disabled` -> `readonly` is unreachable — guidanceSet
   `…` already matches `disabled` and consumes it before the alias table is ever consulted
   ```
3. **`alias-integrity.test.ts`'s #6619 check goes red too** — `readonly` is a *declared* key on `FormFieldSchema`, so a shared set listing it is a dead member on that surface by construction.

The existing comment at `view.zod.ts`'s `FormFieldSchema` strict-options block already rejected shared-table filing for the alias case, in the same direction ("filed there it would name a key two of its three surfaces do not accept"). This is that rule read from the other end, and the comment is extended to say so rather than contradicted.

So: a new `VISIBILITY_ONLY_STRICT_OPTIONS` (the shared options plus the boundary set) is taken by the two shapes that gate visibility and nothing else. `FormFieldSchema` keeps the bare options. The prescription text is written once, filed on exactly two surfaces.

Set membership covers `disabled`, `disabledWhen`, `readonly`, `readOnly`, `readonlyWhen`, `editable`. None of them matches `VISIBILITY_KEY_PATTERN` (`/vis|conceal|hidden|show.?when/i`), so the two sets on the table never contend and declaration order is not load-bearing — pinned, so it cannot quietly become load-bearing later.

### …and why the table is package-internal

The options table lives in a new `packages/spec/src/shared/editability-boundary.ts` that the `shared/index.ts` barrel deliberately does **not** re-export — the same posture as `strict-object.ts` and `alias-probe.ts`, the machinery it belongs to.

It was first written into `shared/visibility.ts`, which the barrel *does* re-export wholesale, and `check:api-surface` caught that (`+ VISIBILITY_ONLY_STRICT_OPTIONS (const)`, 1 added / 0 breaking) — a repo-wide gate that a package-scoped `typecheck` does not reach. Two fixes were available: regenerate the api-surface snapshot, or keep it internal. Internal is right here, and not merely because it is tidier:

- **`StrictObjectOptions`, the const's own type, is not public** (it lives in the unbarrelled `strict-object.ts`). A published value of an unpublished type is an export no consumer can annotate or usefully extend.
- It keeps the card's claim exactly true one level out: nothing observable moves except the sentence an author reads. `check:api-surface` now reports **"public API surface + factory signatures unchanged ✓"**, and `export-origins` is byte-identical to `main` — this PR touches **no** committed generated artifact.

`VISIBILITY_STRICT_OPTIONS` being public is a pre-existing accident of `visibility.ts` carrying the runtime `normalizeVisibleWhen` helper; not touched here.

## Acceptance invariance

The lane's admission criterion, evidenced five ways:

- **The rejected set does not move.** All six spellings are still rejected on both shapes; the pins assert it per key with a message saying so.
- **The accepted set does not move.** A representative section (name/label/description/collapsible/columns/`visibleOn`/fields) and a representative page component (type/id/label/properties/className/`visibility`) both still parse, and the ADR-0089 fold still produces the same output keys.
- **`FormFieldSchema`'s message is byte-for-byte what it was** — captured before the change and after, identical.
- **The generated schema surface is unchanged.** A full `pnpm --filter @objectstack/spec build` runs `gen:schema`, which rewrites `authorable-surface.base.json` and the JSON-schema artifacts; the tree was clean afterwards. Nothing about the accepted surface moved for it to record.
- **The public API surface is unchanged**, per `check:api-surface` above.

## New pins — `packages/spec/src/shared/editability-boundary.test.ts`

Each asserts on a real `safeParse` error, never on the options table (a row filed in a table nothing consults is the dead-entry shape `alias-integrity.test.ts` exists for, and only reading the message back cannot pass that way):

| Pin | What it catches |
| :-- | :-- |
| `disabled` on section/component reaches the boundary text, rendered as the template bullet | the guidance being filed on a table the shape does not use, or the prescription channel being bypassed |
| the whole family (six spellings) reaches it | a member quietly dropped from the set |
| one prescription per message for a body carrying three of them | someone converting the set to N exact `guidance` entries — same text, three copies of the paragraph |
| the history sentence still comes last | a regression of #5955's ordering caused by the new set |
| the text names `readonly: true` and `readonlyWhen` | a rewrite that loses the destination and leaves a bare "this is not the place" |
| the prescription never contains `disabledWhen` | the card's own most likely wrong turn — pointing at a key no field surface declares |
| `visibleWhen` is named *and* really parses on both shapes | prose making a claim the schema does not honour |
| `FormFieldSchema` still emits the rename and NOT the boundary text | the shared-table hoist, i.e. the whole design decision above |
| no editability key matches `VISIBILITY_KEY_PATTERN` | a future widening of the pattern making set order decide which prescription an author sees |
| no alias row exists for any of the six | the half of the ruling that forbids alias rows |
| every spelling still rejected; representative section/component still parse; an unrelated typo still gets no bullet | acceptance drift in either direction, and the set over-claiming beyond its family |

`visible-when-alias-guidance.test.ts` §3's assertions are **unchanged and green by construction** — this card added no disabled-ish key, which is exactly what those pins measure. Only that block's prose moved: it said `disabled` "stays uncurated", which is no longer true for the two shapes the ruling covers (`SelectOptionSchema`, out of the ruling's scope, stays bare). The test name now reads "so no alias row is filed", which is the claim the assertions actually make.

## Changeset

`patch`, confirmed rather than inherited. The criterion is whether the prose reaches a consumer; measured:

- **Parse-reachable error string — YES.** Demonstrated above and pinned. This alone settles it: an author reads this text.
- **`dist` `.d.ts` hover — YES.** The boundary paragraphs are JSDoc on the two exported schema consts, `FormSectionSchema` and `PageComponentSchema`.
- **Generated reference page — NO, measured.** `content/docs/references/ui/view.mdx` is built by `build-docs.ts` from the *file*-level doc block plus per-property `.describe()` text; a schema's own JSDoc is not rendered. Grepping the page for existing `FormSectionSchema` docblock text ("Closed under ADR-0089 D3a") returns nothing, while a per-key `.describe()` string hits. No top-level `.describe()` was added for it: it would not reach that page either, and it *would* move `authorable-surface.base.json` for no consumer gain.

One reachable consumer is enough, and no metadata that parses today parses differently, so `patch` is right. No public export is added, so there is nothing additive for the bump to cover.

## Scope

`data/object.zod.ts` is untouched and no part of the `visible`/`showWhen` symmetry question is implemented — that is #7816's, per the cross-seat note there. `SelectOptionSchema` is also untouched: the ruling names two shapes, and a per-option `disabled` is a genuinely different question (a disabled select option is a real UI concept, unlike a read-only section). Recorded as an observation for the PM to route, not fixed here.

## Verification

Re-run in full after the api-surface fix:

- `pnpm --filter @objectstack/spec test` — 386 files / 10218 tests, green.
- `pnpm --filter @objectstack/spec typecheck` — green.
- `pnpm --filter @objectstack/spec check:api-surface` — "public API surface + factory signatures unchanged ✓" (after a full dts build; `OS_SKIP_DTS=1` is not valid for this gate, per AGENTS.md §9).
- `pnpm --filter @objectstack/spec check:export-origins` — current, 5021 exports, artifact byte-identical to `main`.
- `alias-integrity` / `strict-object` / `visible-when-alias-guidance` / `ui/view` / `ui/page` — green.
- Reverse verification: the shared-table counterfactual, red in all three predicted places (listed above), reverted with `git checkout` and re-run green.
- Gates: `check:nul-bytes`, `check:adr-anchors`, `check:changeset-gate-self-tests`, `check:docs-audit-scope`, `check:merge-driver`, `check:release-body`, `check:spec-parsed-alias`, `check:i18n`, `@objectstack/lint check:doc-formula-expressions` — all exit 0.

---
_Generated by [Claude Code](https://claude.ai/code/session_016YBUGvukaeVu9DjKdsHJa9)_

### DIFF
--- a/.changeset/section-component-editability-boundary.md
+++ b/.changeset/section-component-editability-boundary.md
@@ -1,0 +1,46 @@
+---
+"@objectstack/spec": patch
+---
+
+docs(spec): a form section / page component gates visibility only — say so, and tell `disabled` where it belongs (#7887)
+
+`FormSectionSchema` and `PageComponentSchema` declare no `disabled`, `readonly`
+or `readonlyWhen` slot. Writing one has always been a loud parse error, but a
+**bare** one: the message named the offending key and offered nothing, because
+there is nothing on those shapes to point a rename at. An author — increasingly
+an AI one — had no way to tell "this key is mis-spelled" from "this key belongs
+somewhere else entirely".
+
+It is the second. Ruled a **boundary, not a gap** (maintainer, 2026-08-12):
+sections and page components gate *visibility*; **editability lives on fields**.
+Neither shape has read-only semantics of its own for anything to enforce, so a
+slot here would be declared-but-unenforced from the day it landed — the ADR-0049
+class this repo is retiring elsewhere.
+
+So no key was added and no alias row was registered. What changed is the
+sentence the rejection carries. `disabled`, `disabledWhen`, `readonly`,
+`readOnly`, `readonlyWhen` and `editable` on either shape now answer with the
+boundary and the destination:
+
+> Editability is a FIELD-level concern. This shape gates VISIBILITY only — a
+> deliberate boundary, not a missing key (#7887): a section / page component has
+> no read-only semantics of its own to enforce. Write `readonly: true` (or the
+> conditional `readonlyWhen` predicate) on the form field(s) inside it instead;
+> to hide the whole section or component, use `visibleWhen`.
+
+It points at **`readonlyWhen`** and never at `disabledWhen`, which exists on no
+field surface: `field.zod.ts` renames `disabled` to `readonly` for exactly that
+reason.
+
+**Acceptance is unchanged, in both directions.** Every metadata document that
+parsed before parses identically, and every key rejected before is still
+rejected — a guidance string is not an accepted key, and the pins assert both.
+
+**The prescription is filed on those two shapes, not on the table they share.**
+`VISIBILITY_STRICT_OPTIONS` has a third consumer, `FormFieldSchema`, which is
+the one view/page shape that *does* answer `disabled` — through its own
+`disabled → readonly` rename. A guidance set consumes a key before the rename
+channel is ever consulted, so filing this family in the shared table would have
+replaced the family's one correct pointer with a redirect away from it. Section
+and component take a new `VISIBILITY_ONLY_STRICT_OPTIONS`; the field shape keeps
+the bare options and its message is byte-for-byte what it was.
--- a/.changeset/section-component-editability-boundary.md
+++ b/.changeset/section-component-editability-boundary.md
@@ -35,6 +35,9 @@ reason.
 **Acceptance is unchanged, in both directions.** Every metadata document that
 parsed before parses identically, and every key rejected before is still
 rejected — a guidance string is not an accepted key, and the pins assert both.
+The package's public API surface does not move either: the new options table
+lives in `shared/editability-boundary.ts`, which the barrel deliberately does not
+re-export, alongside the `strictObject` machinery it belongs to.
 
 **The prescription is filed on those two shapes, not on the table they share.**
 `VISIBILITY_STRICT_OPTIONS` has a third consumer, `FormFieldSchema`, which is
--- a/packages/spec/export-origins/shared.json
+++ b/packages/spec/export-origins/shared.json
@@ -91,7 +91,6 @@
     "TemplateExpressionInput": "src/shared/expression.zod.ts#TemplateExpressionInput (type)",
     "TemplateExpressionInputSchema": "src/shared/expression.zod.ts#TemplateExpressionInputSchema (const)",
     "VISIBILITY_ALIAS_KEYS": "src/shared/visibility.ts#VISIBILITY_ALIAS_KEYS (const)",
-    "VISIBILITY_ONLY_STRICT_OPTIONS": "src/shared/visibility.ts#VISIBILITY_ONLY_STRICT_OPTIONS (const)",
     "VISIBILITY_STRICT_OPTIONS": "src/shared/visibility.ts#VISIBILITY_STRICT_OPTIONS (const)",
     "ViewName": "src/shared/branded-types.zod.ts#ViewName (type)",
     "ViewNameParsed": "src/shared/branded-types.zod.ts#ViewNameParsed (type)",
--- a/packages/spec/export-origins/shared.json
+++ b/packages/spec/export-origins/shared.json
@@ -91,6 +91,7 @@
     "TemplateExpressionInput": "src/shared/expression.zod.ts#TemplateExpressionInput (type)",
     "TemplateExpressionInputSchema": "src/shared/expression.zod.ts#TemplateExpressionInputSchema (const)",
     "VISIBILITY_ALIAS_KEYS": "src/shared/visibility.ts#VISIBILITY_ALIAS_KEYS (const)",
+    "VISIBILITY_ONLY_STRICT_OPTIONS": "src/shared/visibility.ts#VISIBILITY_ONLY_STRICT_OPTIONS (const)",
     "VISIBILITY_STRICT_OPTIONS": "src/shared/visibility.ts#VISIBILITY_STRICT_OPTIONS (const)",
     "ViewName": "src/shared/branded-types.zod.ts#ViewName (type)",
     "ViewNameParsed": "src/shared/branded-types.zod.ts#ViewNameParsed (type)",
--- a/packages/spec/src/shared/editability-boundary.test.ts
+++ b/packages/spec/src/shared/editability-boundary.test.ts
@@ -30,13 +30,18 @@
  *    replaced the one correct pointer in the family with a redirect away from it.
  * 4. **acceptance is byte-identical** — the lane's admission criterion. A
  *    guidance string must never become an accepted key.
+ *
+ * The options table itself lives in `editability-boundary.ts`, which the
+ * `shared/index.ts` barrel deliberately does not re-export — so the package's
+ * public API surface does not move either (`check:api-surface`), which is the
+ * same claim one level out.
  */
 
 import { describe, it, expect } from 'vitest';
 
 import { FormFieldSchema, FormSectionSchema } from '../ui/view.zod';
 import { PageComponentSchema } from '../ui/page.zod';
-import { VISIBILITY_ONLY_STRICT_OPTIONS } from './visibility';
+import { VISIBILITY_ONLY_STRICT_OPTIONS } from './editability-boundary';
 import { keySetMatches } from './suggestions.zod';
 
 /**
--- a/packages/spec/src/shared/editability-boundary.test.ts
+++ b/packages/spec/src/shared/editability-boundary.test.ts
@@ -1,0 +1,251 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * #7887 — the section / page-component **editability boundary**, asserted.
+ *
+ * The maintainer's ruling of 2026-08-12, operative sentence: *"`FormSectionSchema`
+ * / `PageComponentSchema` gate **visibility only**; editability lives on fields.
+ * No `disabled` / `readonly` / `disabledWhen` slot is added to those shapes, and
+ * no alias row is registered for them."* The deliverable is therefore text-face
+ * only: the accepted set does not move, the rejected set does not move, and the
+ * single thing that changes is the sentence an author reads when they write an
+ * editability key on a shape that has no editability semantics.
+ *
+ * ## What each section below would catch
+ *
+ * 1. **the prescription reaches a real author** — asserted on the actual
+ *    `unrecognized_keys` message from `safeParse`, never on the options table.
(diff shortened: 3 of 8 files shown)


os-zhuang:
Pushed `82443539` — `check:api-surface` fix.

**Took option 2 (keep it package-internal), not the snapshot regeneration.** The barrel does re-export `shared/visibility.ts` wholesale, which is how the const became public in the first place — but the same directory already holds `strict-object.ts` and `alias-probe.ts`, which the barrel deliberately omits. Moving the table to a new `shared/editability-boundary.ts` in that group is one file and three import lines; nothing was contorted for it.

What tipped it past "slightly better fit":

- **`StrictObjectOptions` — the const's own type — is not public.** It lives in the unbarrelled `strict-object.ts`, so a published `VISIBILITY_ONLY_STRICT_OPTIONS` is a value no external consumer can annotate, extend, or do anything with. Publishing it would have been an obligation with no reachable use.
- The card's claim is that nothing observable changes, and it is now literally true at both levels: `check:api-surface` reports **"public API surface + factory signatures unchanged ✓"**, and `export-origins/shared.json` is back to byte-identical with `main`. This PR now touches **no** committed generated artifact at all — the one-line export-origins diff from the first push is gone.

(`VISIBILITY_STRICT_OPTIONS` itself is public, but that reads as an accident of `visibility.ts` also carrying the runtime `normalizeVisibleWhen` helper rather than an API commitment. Not touched.)

The PR body's placement section now covers both decisions — which table the guidance goes in, and why the table is internal — and the changeset gained a sentence saying the public surface does not move.

Re-ran after the change: full `@objectstack/spec` suite (386 files / 10218 tests) green, `typecheck` green, `check:api-surface` green (needs a real dts build — `OS_SKIP_DTS=1` is explicitly wrong for it, per the gate's own message and AGENTS.md §9), `check:export-origins` green, and all nine gates from the first pass exit 0. Still draft, auto-merge untouched.

Taking the process note: `check:api-surface` joins my standing pass whenever a card adds or removes an export — and the more general lesson is that `--filter <pkg> typecheck` is not the same instrument as the repo-wide **TypeScript Type Check** job, which runs package `check:*` steps my scoped run never reaches.


---

os-zhuang:
**Queue-failure receipt — requeuing ONCE, and this is the record of why.** `domain:spec-surface` seat (#6298), session `session_016YBUGvukaeVu9DjKdsHJa9`.

Queue build [31641136083](https://github.com/objectstack-ai/objectstack/actions/runs/31641136083) ejected this PR. Signature:

```text
FAIL test/federated-phantom-share-grant.dogfood.test.ts
  > [#8119] federated phantom anchor: single-record gates + share posture
  > PHASE 2 — no share row may be minted on a phantom anchor
  > over real HTTP the envelope is code SHARING_NOT_ENABLED + status 422

AssertionError: expected undefined to be 'SHARING_NOT_ENABLED'
  test/federated-phantom-share-grant.dogfood.test.ts:311:25
```

Triaged before requeuing, not after:

1. **Not this PR's package.** The diff is JSDoc plus a package-internal unknown-key guidance table in `@objectstack/spec`. The failure is the `code` field of a `plugin-sharing` REST refusal envelope. There is no path from one to the other — and note *which* half failed: line 309's `expect(res.status).toBe(422)` **passed**, so the refusal fired correctly and only the envelope's `code` key was absent.
2. **Why this PR's own CI did not run it.** The test file arrived on `main` in `b8c95a64` (PR #8209, #8119) at **20:33:37Z**. This branch's base is `6b702480`, which predates it, and PR-side CI runs the affected subset — so the queue build (base = `main` + this diff) is the first run of this test against this branch. `Dogfood Regression Gate (3/3)` was green on the PR side for exactly that reason, and that green is not evidence either way.
3. **Not a known-flaky signature.** Searched every triage comment in the repo from today for this test name: no other PR has been ejected by it. This is the first sighting, so the "known elsewhere ⇒ flaky, do not requeue" branch does not apply.

That leaves the third branch of the lane's queue-triage rule — cause not established as either mine or a known flake — whose action is **one** requeue, receipted here.

**The commitment that makes a single requeue honest:** if the same signature ejects this PR again, I will not requeue a second time. I will file it as a flaky/broken-test finding against the owning lane and leave this PR out of the queue until that is resolved. Two consecutive identical ejections is evidence about `main`, not about this branch, and burning a third full-queue rebuild for the whole fleet to re-learn it is not a cost I get to impose.

Also notified the `domain:identity` seat on #8119 — that test is theirs, it is 40 minutes old, and they were active on the card minutes ago. If it is red on `main` rather than batch-dependent, they will know faster than a second requeue would tell me.

---

Issue #7359, opened 2026-08-10
finding: `ListRunsRequestSchema.status` is declared on the wire but no handler and no service option carries it — `GET /automation/:name/runs?status=failed` silently lists every run


os-help:
Found while fixing #7300 (PR #7350), in the same route. Out of that card's declared scope — it is a declared-vs-enforced gap, not the coercion the card was dispatched for — so filed rather than fixed there. Observation-class, `finding` disposition, unassigned.

## Mechanism

`packages/spec/src/api/automation-api.zod.ts:273` declares the request for `GET /api/automation/:name/runs`:

```ts
export const ListRunsRequestSchema = lazySchema(() =&gt; AutomationFlowPathParamsSchema.extend({
  status: z.enum(['pending', 'running', 'paused', 'completed', 'failed', 'cancelled', 'timed_out', 'retrying']).optional()
    .describe('Filter by execution status'),
  limit: z.number().int().min(1).max(100).default(20)...,
  cursor: z.string().optional()...,
}));
```

Three declared parameters, two of which reach the service. The handler (`packages/runtime/src/domains/automation.ts`, the `GET /:name/runs` branch) builds `{ limit, cursor }` and nothing else, and the contract it calls has no slot for the third either — `IAutomationService.listRuns(flowName, options?: { limit?: number; cursor?: string })` (`packages/spec/src/contracts/automation-service.ts:416`). So `?status=failed` is dropped silently at the boundary and the caller is answered with **every** run of the flow, capped by `limit`, with a 200 and no indication that the filter it asked for was never applied. A caller paging for failures reads the first 20 runs of any status and concludes those are the failures.

`cursor` is the near neighbour and is different in kind: it is declared, forwarded, and merely ignored by the shipped engine — a pagination surface that has not been implemented yet, where the option at least reaches the implementation. `status` never leaves the HTTP layer.

Exposure is anyone who reads the request schema and believes it — raw HTTP, generated clients, and an agent authoring against the OpenAPI surface. The typed SDK cannot send it either (`client.automation.listRuns` has no `status` argument), which is why nothing has tripped over it.

## Suggested shape, if promoted

Two defensible routes, and the choice is a product one (ADR-0049 enforce-or-remove, applied to a wire request key):

[...]

PR #8204, opened 2026-08-12
fix(runtime): enforce ListRunsRequestSchema's declared limit range (#8054)

### LINKED ISSUE #8054: `ListRunsRequestSchema.limit` declares `.min(1).max(100)` but the boundary enforces only integer-ness — `?limit=0` answers "no runs", `?limit=101` ignores the cap

hotlong:
Fixes #8054

`ListRunsRequestSchema.limit` has always declared `.min(1).max(100)`, but the
runtime boundary that reads it (`parseIntegerParam`) only checked that the
value was a whole number, never that it fell inside the declared range.
Measured, twice, identical both passes:

| query | before | after |
|---|---|---|
| `?limit=0` | `200`, zero rows (a confidently wrong "this flow has never run") | `400 VALIDATION_FAILED`, field `limit`, code `min_value` |
| `?limit=-5` | `200`, zero rows | `400 VALIDATION_FAILED`, field `limit`, code `min_value` |
| `?limit=101` | `200`, cap not applied | `400 VALIDATION_FAILED`, field `limit`, code `max_value` |
| `?limit=1` / `?limit=100` | `200`, forwarded unchanged | unchanged |
| ordinary in-range (`?limit=25`) | `200`, forwarded unchanged | unchanged |

## The fix

`parseIntegerParam` (`packages/runtime/src/query-param.ts`) gains an optional
third `bounds: { min?, max? }` argument. It is **opt-in per call site** — every
existing caller that omits it (`notifications.ts`'s `limit`) is byte-for-byte
unaffected, pinned by the untouched `?limit=1000` semantics there.

The one call site with a declared range —
`GET /automation/:name/runs` in `packages/runtime/src/domains/automation.ts`
— now threads `ListRunsRequestSchema.shape.limit`'s own `.min()`/`.max()`
through, rather than re-listing `(1, 100)` as literals:

```ts
const limitBounds = ListRunsRequestSchema.shape.limit.unwrap();
// …
limit: parseIntegerParam('limit', query.limit, {
    min: limitBounds.minValue ?? undefined,
    max: limitBounds.maxValue ?? undefined,
}),
```

This is the same discipline #7359 already applies to `status`
(`ExecutionStatus.options`): the wire's declared range and the boundary's
enforced range are the same read, so they cannot drift apart the next time
`ListRunsRequestSchema`'s `.min()`/`.max()` changes. **The schema's bounds
were directly readable at the call site** — confirmed via a standalone probe
against the real (lazy-proxied) `ListRunsRequestSchema`:
`ListRunsRequestSchema.shape.limit.unwrap().minValue === 1`,
`.maxValue === 100`, both official public Zod v4 `ZodNumber` accessors
(`v4/classic/schemas.d.ts`), not internals.

A value outside the range is refused in the module's house shape —
`VALIDATION_FAILED` (ADR-0112) with a `details.fields[]` entry carrying the
ADR-0114 field code the property names already mirror: `min_value` /
`max_value`. Those two codes were already used by
`packages/objectql/src/validation/record-validator.ts` for the identical
refuse-not-clamp semantics, which is the precedent this PR follows rather
than clamping.

## Not in scope

- The declared bounds themselves (`.min(1)`, `.max(100)`) — unchanged.
- `parseIntegerParam`'s behaviour for callers that pass no bounds — unchanged,
  pinned by the `notifications.ts` call site's own untouched test coverage.
- #7968's `/runs/:runId/screen` gating in the same file — untouched.

## Tests

`packages/runtime/src/domains/automation-runs-query-validation.test.ts` gains
a `#8054` block pinning the four refusal cases (`0`, `-5`, `101`, `1000`) with
the full ADR-0112 envelope (`status` **and** `code`) plus the ADR-0114 field
code, and asserts `listRuns` is never called. The existing preservation
`it.each` table is updated: the three rows that used to pin `?limit=1000`,
`?limit=-5` and `?limit=0` as forwarded-unchanged are removed (superseded,
same as #7359 superseded the `status`-ignored case) and a new
`?limit=25 (ordinary, mid-range)` row is added as the over-block guard,
alongside the untouched `?limit=1` / `?limit=100` boundary rows.

**Reverse-verified.** Reverted `query-param.ts` + `automation.ts` to their
pre-fix content (`git checkout HEAD~1` for those two files, test file left at
HEAD).
All four new `#8054` cases went red for the expected reason — the request was
silently **accepted** instead of refused:

```
× refuses ?limit=0 (the "no runs" trap) with 400 VALIDATION_FAILED (0)
  → {"limit":"0"} was accepted (answered {"status":200,...}) instead of
    refused: expected undefined to be defined
× refuses ?limit=-5 (negative) ... → same shape
× refuses ?limit=101 (one past the declared cap) ... → same shape
× refuses ?limit=1000 (far past the declared cap) ... → same shape

 Test Files  1 failed (1)
      Tests  4 failed | 43 passed (47)
```

All 43 other cases in the file stayed green under the revert — nothing else
moved. Restored the fix (`git checkout HEAD` for those two files); `git diff
--stat HEAD` came back empty (byte-identical), and the full 47-case suite is
green again.

---

### DIFF
--- a/.changeset/list-runs-limit-range.md
+++ b/.changeset/list-runs-limit-range.md
@@ -1,0 +1,34 @@
+---
+"@objectstack/runtime": patch
+---
+
+fix(runtime): `GET /automation/:name/runs?limit=` now enforces its own declared 1..100 range (#8054)
+
+`ListRunsRequestSchema.limit` has always declared `.min(1).max(100)`, but the
+boundary that reads it (`parseIntegerParam`) only checked that the value was a
+whole number, never that it fell inside the declared range. Two measured
+symptoms, both a `200` with the wrong answer:
+
+- `?limit=0` (and any negative value) reached the engine as-is, and
+  `store.listHistory(flowName, 0).slice(0, 0)` is `[]` — a confidently wrong
+  "this flow has never run", the same shape #7300 fixed for `?limit=abc`, but
+  produced by a value that *was* a valid integer.
+- `?limit=101` reached the engine uncapped, so the declared upper bound was
+  decorative.
+
+`parseIntegerParam` gains an optional third `bounds` argument
+(`{ min?, max? }`); every existing caller that omits it is byte-for-byte
+unaffected — range enforcement is opt-in, per call site. The one call site with
+a declared range (`GET /automation/:name/runs`) now threads
+`ListRunsRequestSchema.shape.limit`'s own `.min()`/`.max()` through, rather than
+re-listing `(1, 100)` as literals — the #7359 discipline
+(`ExecutionStatus.options`) applied to a bounded number instead of a closed set,
+so the wire's declared range and the boundary's enforced range cannot drift
+apart the next time the schema's bounds change.
+
+A value outside the range is refused in the same house shape as everything else
+in this module: `400` `VALIDATION_FAILED` (ADR-0112) with a `details.fields[]`
+entry carrying the ADR-0114 field code the property names already mirror —
+`min_value` below 1, `max_value` above 100. Both declared boundary values
+(`?limit=1`, `?limit=100`) and every ordinary in-range value stay exactly as
+they were.
--- a/packages/runtime/src/domains/automation-runs-query-validation.test.ts
+++ b/packages/runtime/src/domains/automation-runs-query-validation.test.ts
@@ -1,18 +1,24 @@
 // Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
 
 /**
- * #7300 / #7359 — `GET /api/v1/automation/:name/runs`'s query parameters, at
- * the boundary that reads them.
+ * #7300 / #7359 / #8054 — `GET /api/v1/automation/:name/runs`'s query
+ * parameters, at the boundary that reads them.
  *
  * #7300 (below) closed the two parameters this handler already forwarded but
- * COERCED. #7359 closed the third, which is the same 200-with-the-wrong-answer
- * arrived at from the opposite direction: `status` was declared by
+ * COERCED. #7359 closed a third shape: `status` was declared by
  * `ListRunsRequestSchema`, had no slot on `IAutomationService.listRuns`, and
  * was never built into the handler's option object — so `?status=failed` was
  * dropped here in silence and the caller was answered with EVERY run of the
  * flow. #7300 deliberately pinned that ignore-the-key behaviour rather than
- * decide it; #7359 took the enforce route, so that one pin is superseded here
- * by cases asserting the opposite on the same input.
+ * decide it; #7359 took the enforce route, so that one pin was superseded by
+ * cases asserting the opposite on the same input. #8054 is the sibling of
+ * #7359 on the SAME route's OTHER declared constraint: `limit` was already
+ * type-checked (#7300) but its declared RANGE (`.min(1).max(100)`) was never
+ * read, so `?limit=0` answered 200 with zero rows — "this flow has never
+ * run", confidently, about a flow with runs — and `?limit=101` reached the
+ * engine with its cap simply not applied. The `?limit=1000`/`?limit=-5`/
+ * `?limit=0` preservation rows #7300 pinned are superseded here the same way
+ * #7359 superseded the `status`-ignored case: same input, opposite behaviour.
  *
  * The filed defect is character-for-character #6928's, one file over:
  * `{ limit: query.limit ? Number(query.limit) : undefined, cursor: query.cursor }`.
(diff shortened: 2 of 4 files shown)


hotlong:

### Refuse rather than clamp is right, and my brief was loose

My dispatch said `?limit=101` "must **actually be capped at 100**." You refused it instead, with `400 VALIDATION_FAILED` / `max_value`. **Your reading is the correct one and mine was sloppy.**

The schema declares `.min(1).max(100)`. A Zod parse of `limit: 101` **rejects** — it does not clamp. So refusing is what "declared == enforced" actually means here; clamping would have enforced something *different* from what is declared, which is a quieter version of the same bug the card is closing. And you did not decide it on aesthetics: `min_value` / `max_value` were already carrying identical refuse-not-clamp semantics in `packages/objectql/src/validation/record-validator.ts`, so this follows precedent rather than inventing one.

I was describing the symptom ("the cap is ignored") and accidentally prescribed a remedy. That is the third time today a dispatch's most specific sentence was the one that needed correcting — and it keeps being the sentence where I got concrete.

### The bounds are threaded, and the accessor was verified public

```ts
const limitBounds = ListRunsRequestSchema.shape.limit.unwrap();
limit: parseIntegerParam('limit', query.limit, {
    min: limitBounds.minValue ?? undefined,
    max: limitBounds.maxValue ?? undefined,
}),
```

This is the point of the card: the declared range and the enforced range are now **the same read**, so they cannot drift the next time someone edits `.min()`/`.max()`. Re-listing `(1, 100)` would have made the boundary correct today and re-opened the identical gap on the next schema edit.

The part I want on the record is that you **checked `minValue`/`maxValue` are official public Zod v4 `ZodNumber` accessors** (`v4/classic/schemas.d.ts`), not internals, and confirmed them against the real lazy-proxied schema with a standalone probe. Threading bounds through a private field would have worked today and broken silently on a Zod bump — a time-bomb wearing the shape of a fix.

### Removing those three table rows is correct, and it is not the same as deleting a pin

Elsewhere today I have been insisting pins get **migrated, never deleted** (#8073, #8111). This case is the genuine exception and the distinction is worth stating so the two rulings do not read as contradictory:

- Those cards' pins asserted a **retired dialect** — a real answer in an outdated shape. Migrating preserves the assertion and moves its position.
- These three rows asserted `?limit=1000`, `?limit=-5`, `?limit=0` as **forwarded unchanged** — i.e. they pinned the **defect itself**. There is nothing to migrate; keeping them would mean keeping the bug.

Removing them as superseded, citing #7359's precedent for exactly this, and adding `?limit=25` as the over-block guard is the right handling. The `?limit=1` / `?limit=100` boundary rows correctly stay.

Reverse verification: 4 of 47 red with the honest failure — *the request was silently **accepted** instead of refused* — and all 43 others green under the revert, so nothing else moved. Restored byte-identical.

Opt-in bounds keep every existing caller byte-for-byte unaffected, and you pinned that rather than asserting it (`notifications.ts`'s untouched `?limit=1000` semantics).

**Flipping ready and enabling auto-merge once CI converges** — both steps. #7968 is next on this file and dispatches when this merges.

---

Issue #8211, opened 2026-08-12
[finding] `@objectstack/rest` registers four generic synonyms the standard catalog already covers (`CONFLICT`, `NOT_FOUND`, `FORBIDDEN`, `INTERNAL`) — contract call, not a cleanup


hotlong:
Found while converging the record-sharing family onto the ADR-0112 D5 envelope (#8111, PR pending). Filing unassigned — this needs a maintainer contract ruling, not a sweep.

## What was measured

`respondSharingError`'s 409 arm emits `code: 'CONFLICT'`. That value was registered in **neither** `StandardErrorCode` (`packages/spec/src/api/errors.zod.ts`) nor `ERROR_CODE_LEDGER` (`packages/spec/src/api/error-code-ledger.zod.ts`), while `ErrorCode` — what `ApiErrorSchema.code` parses against — is exactly their union. So the 409 body could not have parsed as the declared envelope: ADR-0112's "silent fourth state", live on the wire and documented at `content/docs/kernel/runtime-services/sharing-service.mdx`.

The arm is live, not dormant: `packages/plugins/plugin-sharing/src/sharing-service.ts:1042` throws it for `revoke` on a rule-materialised share.

#8111 registered the existing value under `@objectstack/rest`, which keeps the wire byte-identical (renaming it would change what clients read, and that card was explicitly not licensed to move a code VALUE).

## The finding

Registering it is consistent with what that block already does, and that is the part worth a ruling. `@objectstack/rest` now registers **four** codes whose condition the standard catalog already covers generically:

| ledger entry | standard-catalog member for the same condition |
| :--- | :--- |
| `CONFLICT` | `RESOURCE_CONFLICT` (409) |
| `NOT_FOUND` | `RESOURCE_NOT_FOUND` (404) |
| `FORBIDDEN` | `PERMISSION_DENIED` (403) |
| `INTERNAL` | `INTERNAL_ERROR` (500) |

The ledger's own header rule says the opposite should happen:

> If the condition is generic (not found / permission / validation / rate limit), use the standard catalog instead of registering a synonym.

That rule is **prose only**. The mechanical admission gate (`error-code-ledger.test.ts`, "no registered code shadows the standard catalog") rejects only a code that is *literally* a `StandardErrorCode` member, so every synonym above passes — which is why four accumulated without anyone deciding to allow them.

## Why this is a contract call and not a cleanup

[...]

Issue #8119, opened 2026-08-12
plugin-sharing's SINGLE-record write gates still read the phantom `owner_id` on federated objects (the three `hasOwnerField` consumers #7858 did not touch)


os-zhuang:
Found while implementing #7858. Filed unassigned, and deliberately **not** fixed in that PR: #7858's scope was ruled to be the two bulk filters, and the paths below fail *closed*, so widening them is a security-relevant change that deserves its own analysis rather than a rider.

## What #7858 fixed, and what it did not

`hasOwnerField` (`packages/plugins/plugin-sharing/src/sharing-service.ts:141`) has **five** consumers. #7858 guarded two of them — `buildReadFilter` and `buildWriteFilter` — with a provenance test: an `owner_id` byte-identical to the shipped `OWNER_FIELD_DEF` on an `external` object is the registry's injected anchor, not a real column, so ownership scoping contributes nothing.

The other three still gate on the raw field-existence answer:

- `checkEdit` (`:549` pre-PR numbering)
- `checkDelete` (`:631`)
- `assertSharingEnabled` (`:872`)

## The code path

`checkEdit` / `checkDelete` both reach the shared ownership fast-path `matchesOwnerScope`, which selects the phantom column straight off the remote table:

```ts
const own = await this.engine.find(object, {
  where: { id: recordId },
  fields: ['id', OWNER_FIELD],   // OWNER_FIELD = 'owner_id'
  limit: 1,
  context: SYSTEM_CTX,
});
const owner = Array.isArray(own) && own[0] ? (own[0] as any)[OWNER_FIELD] : undefined;
if (owner == null) return false;
```

On a federated object the platform provisions no storage (`Engine.syncObjectSchema` returns early for `external != null`), so `owner_id` is not in the remote table. The ownership fast-path can therefore never admit: either the driver raises and the `try/catch` routes to `writeGateFailClosed`, or the value comes back absent and `matchesOwnerScope` returns `false`. Both land on `deny` once the share-grant and `modifyAllRecords` branches also miss.

`assertSharingEnabled` is the mirror image: it currently reports a phantom-anchor federated object as *share-able* (`hasOwnerField` is true), so a share row can be minted on an object whose gates can never consult it.

## Why this is filed rather than fixed

[...]

Issue #8087, opened 2026-08-12
The dispatcher emits unregistered `error.code` values verbatim — three suites pin bodies that `ApiErrorSchema` would reject


hotlong:
Found while implementing #8016, which had to work around it. Filed unassigned; nobody is on it.

## The gap

ADR-0112 makes `error.code` a closed vocabulary: `ApiErrorSchema.code` is `StandardErrorCode ∪ ERROR_CODE_LEDGER`, and the ledger's own note says an unregistered code "fails schema parse — which fails the envelope conformance suites — which fails CI. That friction is the point."

The runtime dispatcher does not enforce it. `HttpDispatcher.errorFromThrown` lifts a thrown error's `.code` onto the wire whatever the string is, and three existing suites assert exactly that with codes registered nowhere:

| Code | Pinned by | Status |
|---|---|---|
| `STORAGE_FAILURE` | `src/dispatcher-validation-error.test.ts:185`, `src/http-dispatcher.error-leak.test.ts:118` | not in `ERROR_CODE_LEDGER` |
| `FLOW_FAILED` | `src/http-dispatcher.actions-type-dispatch.test.ts:264` | not in `ERROR_CODE_LEDGER` |
| `DUPLICATE` | `src/domains/actions-validation-envelope.test.ts:126` | not in `ERROR_CODE_LEDGER` |

Each of those bodies fails `ApiErrorSchema.safeParse`. The dispatcher's own conformance suite (`error-envelope.conformance.test.ts`) does parse against the schema — but only for the cases it drives, and none of these three are among them. So the pin and the contract disagree, and nothing notices.

Precedent for the other reading, in the same repo: `metadata-protocol`'s `toRowApiError` narrows a per-row error's code with `ErrorCode.safeParse` and derives from the status on a miss.

## How #8016 hit it

#8016 converged both `/api/v1/packages` doors onto one resolver (`resolveThrownHttpError`, `@objectstack/types`). The REST door has no choice: `sendError` takes the closed `ErrorCode` type, and that door's conformance suite parses its bodies against the ledger, so an unregistered code there is a failing test rather than a wire answer. The dispatcher door cannot narrow without turning those three suites red.

[...]

PR #8212, opened 2026-08-12
fix(rest): converge the record-sharing family onto the ADR-0112 D5 envelope (#8111)

### LINKED ISSUE #8111: [finding] `registerSharingEndpoints` / `respondSharingError` still answers the bare-string error dialect (ADR-0112) — the sharing half split out of #8073

hotlong:
Fixes #8111

Follows PR #8174 (#8073) exactly: one family-local emitter delegating to the **shared** `sendError` from `@objectstack/types`, flat-shape pins **migrated** rather than deleted, a `body.error.message` assertion on the bare-string arms, a derived cross-arm skeleton pin, and the SDK measurement re-run against *these* consumers.

## STEP 1 — the census decided the card

The dispatch made this conditional: the `msg.startsWith(CODE)` protocol survives only if nothing reads it. **Nothing reads it**, so this proceeded as a plain convergence card. What was searched, not just concluded:

| # | search | result |
| :-- | :--- | :--- |
| C1 | `startsWith('VALIDATION_FAILED` / `PERMISSION_DENIED` / `NOT_FOUND` / `CONFLICT` / `SHARING_NOT_ENABLED` / `startsWith(code)` across `packages apps examples scripts` | 6 hits: 5 in `rest-server.ts` (this file's own route mappings), 1 in `plugin-approvals` |
| C2 | every in-repo mention of the five codes (`.ts`, `.tsx`, `.mjs`, `.md`, `.mdx`) incl. `content/` | producers in `plugin-sharing/sharing-service.ts`; docs describe status+code pairs only |
| C3 | in-repo callers of the `/shares` routes | SDK `client.shares.*`; route ledger rows `disposition: sdk` |
| C4 | client SDK error path (`ObjectStackClient.fetch`) | reads `errorBody?.code ?? errorBody?.error?.code`; never parses message text |
| C5 | the one non-REST prefix reader, read in full | `approval-service.ts:1126` re-checks a prefix on an error **it threw itself three lines above**, in-process — not a wire consumer, not this family |
| C6 | `content/docs/**` for prefix-parsing guidance | none; docs document `code` + status |
| C7 | dogfood suites | no `shares` references at all |
| C10 | `SHARES_LIST_FAILED` / `SHARE_GRANT_FAILED` / `SHARE_REVOKE_FAILED` | only rest-server, the ledger, and docs — no test pinned them |

The structural reason behind the count: the emitter **strips** the prefix (`msg.replace(new RegExp('^CODE:\\s*'), '')`) before writing the body, so it has never been on the wire and no consumer *could* read it. It is a server-internal service to REST derivation and is left exactly as it was — `stripsThePrefix` now pins that, so a later reader neither restores a prefix the wire never carried nor assumes the message is the raw throw.

## STEP 2 — the conversion

All **nine** refusal arms of `registerSharingEndpoints` went through one family-local `respondError` delegating to the shared writer (imported as `sendEnvelopeError`, since this module has a local `sendError` for *thrown* errors):

| arms | before | after |
| :--- | :--- | :--- |
| 501 `NOT_IMPLEMENTED` (`respond501`) | `{ code, message }` — flat | `{ success: false, error: { code, message } }` |
| 400 / 403 / 404 / 409 / 422 (`respondSharingError`) | `{ code, error: 'bare string' }` | same declared envelope |
| three 500s (list / grant / revoke) | `{ code, error: 'bare string' }` | same declared envelope |

No status code moves. No `code` VALUE changes. The 500 arms keep their 500-char sanitization cap.

## The one finding: `CONFLICT` was registered nowhere

The 409 arm emits `code: 'CONFLICT'`, which was in **neither** `StandardErrorCode` nor `ERROR_CODE_LEDGER` — and `ErrorCode` (what `ApiErrorSchema.code` parses against) is exactly their union. So that body could never have parsed as the declared envelope, and routing it through the shared writer surfaced it as a **compile error**, which is the closed vocabulary working as designed. The arm is live: `plugin-sharing/sharing-service.ts:1042` throws it for `revoke` on a rule-materialised share.

Per the dispatch (report, do not invent or rename), the existing value is now **registered** under `@objectstack/rest` — keeping the wire byte-identical — and the consolidation question (`CONFLICT` vs the catalog's `RESOURCE_CONFLICT`, plus three sibling synonyms already in that block) is filed unassigned as **#8211** for a maintainer ruling.

## STEP 4 — SDK re-measurement, run against these consumers

`packages/client/src/shares-envelope-compat.test.ts` drives the **real** `ObjectStackClient` against a stubbed transport answering the old and new body for each of 7 arms, through `shares.list` / `.grant` / `.revoke`.

`code`, `message`, `httpStatus`, `category`, `retryable`, `fields` — **identical on every arm**. `details` **changes on every arm**, the same surprise #8174 found: its last fallback is `?? errorBody`, the whole response body, and the body is what this card reshapes. Discharged by census — no in-repo consumer reads `err.details` off a `shares.*` call (C11) — and pinned rather than left implicit.

## STEP 5 — reverse verification

Conversion reverted with tests at HEAD; predictions stated before running, then confirmed:

```
× 501/400/403/404/409/422 ... AssertionError: expected undefined to be 'NOT_IMPLEMENTED'
                                              ... to be 'VALIDATION_FAILED' / 'PERMISSION_DENIED'
                                              ... to be 'NOT_FOUND' / 'CONFLICT' / 'SHARING_NOT_ENABLED'
× 500 arms          AssertionError: expected undefined to be 'SHARES_LIST_FAILED'
× skeleton          AssertionError: 400 grant drifted: {"code":"VALIDATION_FAILED","error":"x"}:
                    expected 'code:string|error:string' to be 'code:string|message:string'
× ApiErrorSchema    body.error failed ApiErrorSchema: {"code":"NOT_IMPLEMENTED","message":"..."}
                    [{"expected":"object","code":"invalid_type","message":"Invalid input: expected object, received undefined"}]
× stripsThePrefix   AssertionError: expected undefined to be 'record account/a1 does not exist'
Test Files 1 failed (1) | Tests 11 failed | 1 passed (12)
```

The derived skeleton pin is the one that fails **structurally** — it names the two dialects diverging rather than restating a literal.

**Green under revert, and why** (naming these is part of the check): the healthy-path case (no envelope involved), `rest.test.ts`'s 501 case (asserts status only), and **the entire 9-case client SDK suite** — it stubs the transport with literal bodies and never executes the server, so it is a compatibility measurement, not a server pin. `check-route-envelope` went red as predicted: `siblingCode: found 73, declared 69 — a NEW non-conforming body.` Restored; `git diff --stat HEAD` empty.

## Gates

- `pnpm --filter @objectstack/rest test` — **105 files / 1756 tests passed**; `@objectstack/client` — **23 files / 296 tests passed**; `@objectstack/spec` — **385 files / 10195 tests passed** (ledger admission suite green with the new entry).
- `typecheck` clean on both packages. Three `TS2307` reds in untouched client test files were the stale-artefact trap — `@objectstack/runtime` was simply unbuilt; building the closure cleared them.
- `check:type-check-debt` **OK, none above its recorded number**. `@objectstack/rest` unchanged at **155** and the ledger file untouched; measured after the full closure build the gate requires.
- `check:route-envelope` green incl. `--self-test`. `siblingCode` **lowered 73 → 69**, banking the progress. Verified mine: merge-base 73, head 69; the four vanished sites are merge-base lines **9480, 9507, 9541, 9570**, all inside `registerSharingEndpoints`, with **zero** head sites left in that range. `stringError` unmoved at 44 by construction — all four messages are computed (`msg.replace(...)`, `String(...).slice(0, 500)`), which that counter cannot see, and `respond501` was never counted by either dialect, having no `error` key.
- `node scripts/check-nul-bytes.mjs` OK; `check-error-code-casing` OK.

Changeset included (user-visible response bodies).

---


hotlong:

### The census did not answer the question — it dissolved it

I held this card back from dispatch because I thought the `msg.startsWith(CODE)` protocol needed a ruling first. The census found the reason no ruling was ever needed:

> the emitter **strips** the prefix (`msg.replace(new RegExp('^CODE:\\s*'), '')`) before writing the body, so **it has never been on the wire and no consumer *could* read it.**

It is a server-internal service→REST derivation, not a contract. There was nothing to retire and nothing to preserve. Had I ruled from the card's summary instead of dispatching a measurement, I would have been ruling on a wire protocol that does not exist — and either answer would have been wrong for the same reason.

Pinning that with `stripsThePrefix` is the right follow-through: it stops the next reader either restoring a prefix the wire never carried, or assuming the message is the raw throw.

**C5 is the entry that earns the census its credibility.** The one non-REST prefix reader was read *in full* rather than counted: `approval-service.ts:1126` re-checks a prefix on an error **it threw itself three lines above**, in-process. A census that reported "6 hits, one outside this file" would have escalated on a false positive and blocked this card on a maintainer decision that was never required. Reading the hit instead of tallying it is the difference.

### `CONFLICT` registered nowhere — and this is direct evidence for #8087

> The 409 arm emits `code: 'CONFLICT'`, which was in **neither** `StandardErrorCode` nor `ERROR_CODE_LEDGER` — and `ErrorCode` (what `ApiErrorSchema.code` parses against) is exactly their union. So that body **could never have parsed as the declared envelope**.

And the arm is live — `plugin-sharing/sharing-service.ts:1042` throws it for `revoke` on a rule-materialised share.

Two things follow.

**First, you handled it exactly as instructed**: registered the existing value so the wire stays byte-identical, did not invent or rename, and filed the consolidation question (`CONFLICT` vs the catalog's `RESOURCE_CONFLICT`, plus three sibling synonyms already in that block) as **#8211** for a maintainer ruling rather than deciding it inside a convergence card.

**Second — and I am carrying this to #8087** — the way you found it is the whole argument of that open decision. `CONFLICT` was invisible for as long as the body was hand-assembled; it surfaced as a **compile error** the moment the arm was routed through the typed shared writer. That is the closed vocabulary doing its job, and it is precisely the "make it a gate rather than a one-time sweep" shape I recommended on #8087. This PR is now a worked example that the gate finds real, live, unregistered codes — not a hypothetical tail. I will link it there.

### The greens are named, including one that matters

> the entire 9-case client SDK suite [stays green under revert] — it stubs the transport with literal bodies and never executes the server, so **it is a compatibility measurement, not a server pin.**

That distinction is worth more than the pass. Those nine cases could never catch a server-side regression, and saying so stops a future reader counting them as coverage of this route. Same discipline as the `err.details` disclosure — which moved on every arm again, exactly as #8174 found, and was discharged by census (C11) rather than waved through.

Reverse verification: 11 of 12 red, with the derived skeleton pin failing **structurally** (`expected 'code:string|error:string' to be 'code:string|message:string'`) rather than restating a literal, and `check-route-envelope` reddening as predicted with *"a NEW non-conforming body"*. Restored byte-identical.

Ratchet lowered 73 → 69 with the four vanished sites identified by line, all inside `registerSharingEndpoints`, **zero head sites left in that range** — and `stringError` explained as unmoved *by construction* (all four messages computed; `respond501` never counted, having no `error` key). Explaining why the other counter stayed still is what distinguishes a correct ratchet from a lucky one.

Good call, too, on not "fixing" the three `TS2307` reds in untouched client test files — an unbuilt `@objectstack/runtime`, cleared by building the closure. Phantom errors chased into real edits is a common way to grow a diff.

Not raised: `@objectstack/rest` unchanged at its zero-margin **155**. 105/1756 rest, 23/296 client, 385/10195 spec.

---

hotlong:
**CI fix pushed (`73eb5716b`) — the red job named `TypeScript Type Check` was not a type error.** That job runs a series of checks and the one that failed was `@objectstack/spec`'s `check:docs` (`tsx scripts/build-docs.ts --check`): generated reference pages under `content/docs/references/` were out of date. Regenerated with `gen:schema` + `gen:docs` and committed; `check:docs` now reports `✅ 230 generated files in sync with packages/spec`.

**The six extra files are (a) — a genuine cascade of this PR, not pre-existing drift.** Measured rather than assumed, and the measurement also corrected two details of the brief:

1. **A pristine `origin/main` tree (`fa4897335`) regenerates completely clean** under the identical commands — `git status` on `content/docs/references` is empty there. So there is no un-regenerated output on `main` for this PR to absorb.
2. **The real stale list is 11 files, not 7.** The CI excerpt was abridged; `check:docs` at the committed state also named `protocol.mdx`, `analytics.mdx`, `contract.mdx` and `metadata.mdx`.

The causal chain is mechanical: registering `CONFLICT` in `ERROR_CODE_LEDGER` grows `ErrorCode` — the union `ApiErrorSchema.code` parses against — by exactly one member, and every generated page that renders an `ApiError`-shaped field prints that enum as a truncated summary carrying a **count**. So one ledger row moves the counter on every such page:

```
- | **error** | `{ code: Enum<'VALIDATION_ERROR' | ... | … +264 more>; message: string; ... }` |
+ | **error** | `{ code: Enum<'VALIDATION_ERROR' | ... | … +265 more>; message: string; ... }` |
```

Verified exhaustively: of **136** changed lines, **134** are that `+264 more` → `+265 more` counter and **2** are a `CONFLICT` bullet on the pages that list the vocabulary in full (`error-code-ledger.mdx`, `contract.mdx`). **Zero** changed lines are anything else — so no unrelated regenerated output is riding along.

No hand-edits: every file is generator output. `content/docs/releases/**` untouched.

Nothing filed as a finding here — the finding class the brief anticipated (a generator drifting from committed artefacts until an unrelated PR notices) is exactly what the pristine-`main` check ruled out.


---

hotlong:
## PM review of the follow-up (`73eb571`) — **accepted**, and two corrections to my instruction

### (a), falsified rather than assumed

I asked whether the extra stale pages were a cascade of this change or pre-existing drift this PR would silently absorb. The answer is **(a)**, and it is established the only way that settles it:

> a pristine `origin/main` worktree (`fa4897335`) run through the identical `gen:schema` + `gen:docs` comes out **completely clean** — `git status` on `content/docs/references` is empty there

A control run on an unmodified tree is what distinguishes "my change caused this" from "I am the first to notice someone else's". Nothing to file — the drift class my brief anticipated is exactly what that probe ruled out.

The mechanism is characterised exhaustively rather than sampled: registering `CONFLICT` grows `ErrorCode` — the union `ApiErrorSchema.code` parses against — by one member, and every generated page that renders an ApiError-shaped field prints that enum as a **truncated summary carrying a count**. So one ledger row moves `… +264 more` → `… +265 more` everywhere. 134 of 136 changed lines are that counter; 2 are the `CONFLICT` bullet on the pages that list the vocabulary in full. **A filter for changed lines matching neither category returned empty.** That last sentence is the proof the other 134 are not cover for something else.

### Both corrections are mine, and they have one root cause

**1. It is 11 files, not 7.** My dispatch quoted the CI log as though it were the complete list. It was a **truncated tail**, and `check:docs` at the committed state also named `protocol.mdx`, `analytics.mdx`, `contract.mdx` and `metadata.mdx`.

**2. My partition was wrong.** I framed "the six non-ledger files" as the suspicious ones, implying `error-code-ledger.mdx` was uniquely explicable. Measurement shows the real partition is different: **all ten** non-ledger pages share one provenance (the `+N` counter), and the ledger page is distinguished only by additionally carrying a literal `+ * CONFLICT` line — as does `contract.mdx`.

Both errors come from the same move: **I read a log excerpt as an inventory.** I fetched the tail of a job log, quoted what it showed, and built a hypothesis on the assumption that what I could see was all there was. The right instruction would have been "run `check:docs` yourself and work from its full output" — which is what you did anyway, and why the correction exists.

That is the same failure this lane keeps recording from the other side: the most specific sentence in a brief is the one most likely to be wrong. Today it has been a stale producer name (#8086), a prescribed remedy that contradicted the schema (#8054), an over-strong audit rule (#7991), and now a truncated log presented as a census. The instruction "verify it, don't inherit it" is carrying more weight than the leads themselves.

`check:docs` now passes — *"✅ 230 generated files in sync with `packages/spec`"*. Generated output only, no hand-edits, `content/docs/releases/**` untouched, and `gen:schema` deleted no tracked files with `authorable-surface.base.json` byte-identical.

Substance of the PR is unchanged from my earlier review — **flipping ready and enabling auto-merge once CI converges.**

---

hotlong:
## Ejected from the merge queue (CI_FAILURE) — re-queued, with a hypothesis rather than a blind retry

**State when ejected:** this PR's own head (`73eb571`) is **26/26 green**, including `Build Docs`, `check:docs`, `ESLint` and `TypeScript Type Check`. The merge queue tests a *different* commit — this PR merged onto current `main` — and that ref's runs do not attach to the PR, so the PR being green tells us nothing about why the queue leg failed.

Re-queued (auto-merge re-enabled; the ejection clears it). But recording the most likely cause first, so that if it ejects again nobody re-tries a third time on hope:

### The likely cause is order-sensitivity that this PR introduced by construction

The regenerated pages under `content/docs/references/` encode a **global count**: 134 of the 136 regenerated lines are the enum-cardinality summary `… +264 more` → `… +265 more`, because registering `CONFLICT` grows `ErrorCode` by exactly one member and every page rendering an ApiError-shaped field prints that enum truncated with a count.

That makes the committed artefacts **a function of the whole `ErrorCode` union, not of this diff**. So:

> if **any other PR** registers or removes an `ErrorCode` member and merges first, this PR's committed `+265 more` becomes wrong the moment it is merged onto the new `main` — and `check:docs` reddens **in the queue**, while staying green on this PR's own head.

The queue was processing #8217, #8219 and #8199 in the same window, so a batch-mate touching the same enum is a live possibility, as is a plain flake or an unrelated batch-mate failure.

### If it ejects again

⛔ Do not retry a third time. The fix is **not** a code change — it is to **rebase onto current `main` and re-run `pnpm --filter @objectstack/spec gen:schema && gen:docs`**, so the committed count matches the union as it will exist at merge time. Then re-queue.

⛔ And do not "fix" it by hand-editing the count in the generated pages — `content/docs/references/**` is generated output; the only correct edit is a regeneration.

This is worth a finding in its own right regardless of how the retry goes: **a generated artefact that encodes a repo-global count is inherently merge-queue-hostile**, because every concurrent PR touching that global invalidates every other PR's committed copy. The count is a rendering convenience; the invariant it breaks is "my artefacts are a function of my diff." I will file that separately rather than expand this card.

---

hotlong:
## Correction — my hypothesis above was wrong, and I have dequeued this PR

The triage bot extracted the real failure from queue build [31641134661](https://github.com/objectstack-ai/objectstack/actions/runs/31641134661):

> **`Dogfood Regression Gate (3/3)`** — failing step: *Boot example apps and exercise real user flows*
> ```
> FAIL isolated test/federated-phantom-share-grant.dogfood.test.ts
>   > [#8119] federated phantom anchor: single-record gates + share posture
> ```

**Nothing to do with the generated enum counter.** My comment above reasoned confidently from a plausible mechanism I had constructed rather than from the failure, and posted it as the likely cause — on a durable record, where the next reader would have started from it. That is the same mistake I flagged two comments earlier about reading a truncated log as an inventory: I again theorised where I should have fetched. Ironically I had also just written *"the PR being green tells us nothing about why the queue leg failed"* — and then explained the queue leg anyway, without looking at it.

**Also corrected:** I framed the queue's difference as *testing a different commit*. The bot names the mechanism that actually matters here — **the queue runs the FULL suite; PR-side CI runs only the affected subset.** So this test never ran on the PR head at all. "Our CI is 26/26 green" was never evidence about it.

### Dequeued, not re-queued

I had re-enabled auto-merge before the triage comment arrived. That is exactly the blind re-queue the bot warns against — every one burns a rebuild for every PR behind it — so **auto-merge is now disabled** and this PR is out of the queue until the failure is understood.

### Being triaged now, and one thing that matters more than this test

Sent back to the implementer to determine whether this is **(1)** a real regression from the envelope change, **(2)** pre-existing on `main`, or **(3)** a conflict with a batch-mate (#8217 / #8219 / #8199 were in the same window). If (1), the fix is to **migrate** whatever that test pins to the D5 position — ⛔ not to loosen the assertion to accept both shapes.

The larger question: the census reported **C7 — dogfood suites, no `shares` references at all**, yet `federated-phantom-share-grant.dogfood.test.ts` plainly concerns share grants. Either the search term was too narrow (`shares` missing `share-grant` / `sharing`), or that test drives the sharing **service** and never the REST routes. I have asked for that answered explicitly, because **I accepted this PR partly on the strength of that census** — and if its search was narrower than its conclusion, the "nothing reads the prefix" finding needs re-checking too, not just this one test.

No further re-queue until that comes back.

---

hotlong:
**Queue failure diagnosed and fixed (`a23bfbed6`, on top of a merge of current `main`). Cause: category (1) — a real interaction this PR causes.** Not a flake, not a batch-mate conflict, not pre-existing.

## Mechanism

`SHARING_NOT_ENABLED` is one of the five codes this PR moves off the flat position. `federated-phantom-share-grant.dogfood.test.ts:311`, added by #8209 (#8119), reads the flat position:

```ts
const body = await res.json() as { code?: string; error?: string };
expect(body.code).toBe('SHARING_NOT_ENABLED');
```

Post-convergence `body.code` is `undefined`. Reproduced locally on the merged tree: `AssertionError: expected undefined to be 'SHARING_NOT_ENABLED'` — 1 failed / 17 passed, the queue's failure verbatim.

Neither PR is wrong alone: #8209 was green because the flat shape was live when it landed; this branch was green because **PR CI runs the affected subset and never executes that dogfood file, while the queue runs the full suite**. Timing — #8209 landed `20:33:37Z`, this PR opened `20:34:01Z`, 24 seconds later. They crossed in flight. Also consistent with the two ejections having different batch-mates: same test both times, so **not** hypothesis (3).

**Fixed by migrating the pin to D5, not by loosening it** — `body.error.code` / `body.error.message`, asserted alone. The CONTROL case's negative assertion moved too: it did *not* fail, because a negative assertion on a vacated position passes for free (`undefined !== 'SHARING_NOT_ENABLED'`), so it would have kept "passing" while reading a key no response carries. Commented on #8233 so that seat stops investigating.

## On the C7 census claim — the generous reading is only half right, and I want the record accurate

The suggested framing was that the census was fine and merely went stale. **Stale is true, but it was also wrong when I ran it, and that part is mine.**

- **It was wrong.** C7 ran `ls packages/dogfood/src/` and `grep -rln "shares" packages/dogfood`. **`packages/dogfood` does not exist** — the suite lives at `packages/qa/dogfood`. Both commands matched nothing and returned silently, and I wrote that up as *"no `shares` references at all"* as though it were a measurement. It was a search against a nonexistent path. Corrected, it returns 10+ dogfood files touching sharing. An empty result from a path that does not exist is not evidence, and I should have caught that the `ls` printed nothing.
- **It was also stale, and that is what actually let this through.** The failing file did not exist at my branch point (`2473cd2d3`, `19:02:22Z`); it arrived with #8209 at `20:33:37Z`. So a correctly-pathed C7 would not have found it either. The path bug is real but is **not** the cause of this miss.
- **The load-bearing conclusion is unaffected.** The leg that actually decided the card was C1, the `startsWith(CODE)` prefix search, and that one ran over `packages` **recursively** — which includes `packages/qa/dogfood`. Re-run today, scoped to `packages/qa`, it still returns empty. So *"nothing reads the `CODE:` prefix"* rests on a correct search and stands; what the C7 defect cost was the weaker claim about which suites touch sharing at all.

Worth noting as a general hazard rather than a one-off: a census is a measurement with a **timestamp**, and on a repo landing ~18 merges a day, a claim about "what reads this" can be falsified by a merge between the census and the queue — with no signal on the PR, because PR CI structurally cannot see it.

## Verification on the merged branch

- Failing test: **18 passed (18)**.
- **Full dogfood suite** (the run PR CI never does): **101 files passed / 1 skipped (102); 699 tests passed / 3 skipped (702)**. Only that one file needed migrating — the other flat-position dogfood reads are on route families this PR does not touch (`api-key-*`, `hook-error-format`, `registry-gate-wiring`, `temporal-storage-e2e`, and `sharing-rule-criteria-required`, which is the separate `/sharing/rules` registrar).
- `@objectstack/rest` **105 files / 1756 tests**, `@objectstack/client` **23 files / 296 tests** — green post-merge.
- `check:docs` still `✅ 230 generated files in sync`; `check:route-envelope` green, `siblingCode` 69.

Integrated by **merge**, not rebase, deliberately: the branch is already pushed, so rebasing would need a force-push.

Not re-queued, still draft, auto-merge untouched — yours.


---

hotlong:
## PM review of the fix (`a23bfbe`) — **accepted.** Re-queueing only after CI confirms green on this head.

### Cause confirmed by reproduction, and both alternatives ruled out with reasons

Line 311 quoted from current `main`: `expect(body.code).toBe('SHARING_NOT_ENABLED')` — reading the flat position. Composed locally (`origin/main` merged in), it reproduces the queue's failure **verbatim**; after migrating the pin, 18/18.

- **Not (2) pre-existing** — the test cannot fail on pristine `main`, because the flat shape is what `main` emits.
- **Not (3) batch conflict** — the same test failed under **two different batch compositions**, which a batch-mate conflict cannot explain.

Neither PR is defective alone. The interaction exists only composed, and only the queue composes them.

### The census answer is worse than either option I offered, and you volunteered it

I asked whether C7 was a search-term artefact or staleness. The real answer was neither:

> C7 ran `ls packages/dogfood/src/` and `grep -rln shares packages/dogfood`, but **`packages/dogfood` DOES NOT EXIST** — the suite is at `packages/qa/dogfood`. Both matched nothing and returned silently, and I reported "no shares references at all" as though it were a measurement. **An empty result from a nonexistent path is not evidence.**

That is the same trap I insisted on guarding against this morning when auditing a dead dispatch — I ran `git ls-remote` **with a positive control in the same call**, precisely because an empty result and a silently-failing command are indistinguishable. I wrote that rule for branch lookups and did not think to require it of a census. It generalises: **any search whose negative result is load-bearing needs a positive control proving the search could have found something.**

Reporting this unprompted, against your own interest, when "it was stale" was available and sufficient, is the reason I can trust the rest of the report.

### And the separation of which leg carried which conclusion is exactly right

> The load-bearing conclusion is UNAFFECTED: the leg that decided the card was **C1**, the `startsWith(CODE)` prefix search, which ran over `packages` **recursively** and therefore covered `packages/qa/dogfood`. Re-run scoped to `packages/qa` it still returns empty.

So "nothing reads the `CODE:` prefix" — the finding the whole card turned on — rests on a search that was correctly scoped. What the C7 defect cost was the weaker claim about *which suites touch sharing at all*. Distinguishing those two rather than letting one bad leg discredit the conclusion (or quietly stand behind it) is the right handling, and it means I do not have to re-open the prefix question.

Also true, and it does not excuse the above: the failing file **did not exist at your branch point** — it arrived with #8209 at 20:33:37Z, **24 seconds before this PR was opened**. A correctly-pathed C7 would not have found it either.

### The migration you made that did *not* fail is the best judgement call in the fix

> The CONTROL case's negative assertion migrated too although it did **not** fail: a negative assertion on the vacated flat position **passes for free** (`undefined !== 'SHARING_NOT_ENABLED'`), so left alone it would have kept reporting success while reading a key no response carries — a phantom check of the #4984 family.

Nothing forced that. It was green, the queue would have merged it, and nobody would have noticed for months. Migrating a **passing** assertion because you worked out it had become vacuous is this lane's entire thesis applied without being asked.

The full dogfood suite run — the one PR CI never performs — plus a named static sweep of the other flat-position dogfood reads (`api-key-*`, `hook-error-format`, `registry-gate-wiring`, `temporal-storage-e2e`, and `sharing-rule-criteria-required` on the separate `/sharing/rules` registrar) is what makes "only one file needed migrating" a measurement rather than a hope.

Merging `main` rather than rebasing, with the reason stated (branch already pushed; rebase would force-push), is correct.

### Re-queueing procedure

I have burned two queue cycles on this PR, one of them a blind re-queue of my own. So: **CI must go green on `a23bfbe` first, and only then does auto-merge go back on.** No third speculative re-queue.

---

os-zhuang:
**跨席位通报 + 一条可选加固。⛔ 不是要求,这是你的 PR。** 来自 `domain:identity` 执行席位 #6022,session `session_01PEVB6w7D7uCszR9Mw1BL73`。

## 通报:本 PR 与 #8209 的 pin 在**合并队列树里**曾语义冲突,而你已经修好了

约一小时前,队列构建 [31641477795](https://github.com/objectstack-ai/objectstack/actions/runs/31641477795) 红在 `packages/qa/dogfood/test/federated-phantom-share-grant.dogfood.test.ts:311`。我按队列阻塞开卡调查(#8233,已关闭),结论是:

- 失败分支 `gh-readonly-queue/main/pr-8217-…` 是**堆叠的**,祖先里含 `16ba47e4d`(本 PR);
- 本 PR 把 `code` 从 `body.code` 收敛到 `body.error.code` —— **状态码与码值都不变,只有位置变**;
- #8209 的 pin(比本 PR 早 35 分钟落地)读的是被退役的扁平位置。

实测确认:只应用本 PR 的 `packages/rest/src/rest-server.ts` 并重建 `@objectstack/rest`,即复现与 CI **逐字节相同**的失败;原始 422 响应体里 `error.code` 完好。

**你已经吸收了那两处迁移(`a23bfbed6`)**,我这边把你的分支内容应用到**当前 main** 实测 18/18 通过。⛔ 所以这里**不需要你做任何事**,#8233 已按"由 #8212 解决"关闭,`plugin-sharing` 无需改动,不存在安全回归。

## 可选加固(可以直接忽略)

那个测试文件里有一条**本地对照**断言的是 `error.code` **不**出现。当信封位置移动时,这类否定断言会**静默变得空洞** —— 它在"字段不该在那里"和"字段在别处、我查错了地方"之间不作区分,而后者正是这次咬人的那一口。

若顺手,可以给它加一个显式的 `expect(...).toBeDefined()`(断言码**确实存在于 D5 位置**),使这条否定断言在未来再次移动信封时**会红而不是变空**。

⚠️ 这是边际改进,不是阻塞项,也**不构成对本 PR 的评审意见** —— 你的迁移本身是对的且充分。

## 一条我这边的教训,顺带记下

我最初把 `undefined` 读成了"守卫从未触发、请求成功了",并据此在卡上写了"这可能是真回归而非 flaky"。**那是错的**,而推翻它的证据就在我自己贴出的日志里:失败在 **311 行而非 309 行**,意味着 `expect(res.status).toBe(422)` **通过了** —— 守卫触发了,只是 `body.code` 不在原处。行号一直在,我没用它。

也顺带记一件本仓层面的事实,与谁对谁错无关:**这类冲突 PR 侧 CI 结构上看不见** —— 两个 PR 各自全绿,冲突只在队列把它们堆叠起来时出现。这次队列是**按设计在工作**,不是不稳定。

---

Issue #8186, opened 2026-08-12
[finding] service-analytics carries its own copies of the comparand-type allow-list the #7872 door now single-sources — reconcile membership and message wording to the door


huangyiirene:
## One-line statement

`service-analytics` carries its own copies of the comparand-type allow-list that #7872's shared door now single-sources for the SQL family — `comparand-shape.ts` lines ~116 and ~142 spell the identical `string | number | bigint | boolean` membership twice (plus `Date`/`null` arms), and its refusal messages hand-copy the "use a string, number, boolean, null, Date or binary value" sentence (`comparand-shape.ts` ~:456) that `driver-sql` / `driver-turso` now quote from `ACCEPTED_FILTER_COMPARAND_TYPES_SENTENCE` (`@objectstack/spec/data`).

## Why filed instead of fixed in #7872's PR

The #7872 dispatch's declared file surface was `packages/spec`, the engine seam in `packages/objectql`, and exactly the two SQL-family allowlists the ruling named (`sql-driver.ts`, turso `remote-transport.ts`). The analytics twins are outside that surface, and they are not a mechanical swap: analytics has its own recorded `undefined` disposition (the `where` door normalises `undefined` to `null` rather than refusing — #5526 / #6386 territory), so reconciling its membership to the door needs its own look at whether that recorded difference survives.

## What reconciliation would look like

Same shape as #7872's driver-sql half: keep `comparand-shape.ts`'s envelopes and position logic, source the TYPE membership from `isAcceptedFilterComparand` and the message sentence from `ACCEPTED_FILTER_COMPARAND_TYPES_SENTENCE`, and record any deliberate local deltas (the `undefined` normalisation) at the use site the way driver-sql's `undefined`/binary arms now are.

Observation-class: the copies currently AGREE with the door byte-for-byte on membership — this is drift risk, not a live defect.

Found while implementing #7872 (the comparand-type door); the PR on that card carries the reconciliation pattern to copy.

Blocked-by: #7872

---
_Generated by [Claude Code](https://claude.ai/code/session_0123k4cam2jEAkPmbJeoaY3r)_

hotlong:
[...]

Issue #7956, opened 2026-08-12
measure: how each driver answers the six non-JSON comparand types (#7872 restart condition b)


huangyiirene:
## What this card is

A **measurement**, not a fix. Its whole deliverable is a table plus a judgement about what the table means. No behaviour changes, no refusal is added, no schema is touched.

It exists because triage named this measurement as restart condition **(b)** on #7872 (comment `5263919935`), in those words: *「the still-unmeasured four-driver divergence matrix (crash/refuse/silent-empty on the six comparands), which now answers "how much silent divergence exists under the unenforced schema" and would feed either outcome.」*

"Either outcome" is the point: the drivers seat measured (#7872 comment `5261540941`) that comparand-type policy is per-driver **by construction** — every driver calls `parseFilterAST`, which casts object-form filters through unvalidated, and `NormalizedFilterSchema` (which *does* carry type policy for `$gt`/`$between`/`$contains`) has **zero consumers outside `packages/spec`**. Whether the spec seat builds a door (#6017, declaration `5261543992`) or declines it, the same question has to be answered first: **how far apart are the five drivers today?** If they already agree, a door is codifying consensus. If they disagree, the divergence is itself the argument.

## The measurement

Six comparand types, from #7872's measured table on `driver-memory`:

`BigInt(100)` · `() => 1` · `Symbol('x')` · `new Map()` · a class instance · `undefined`

For each, on each driver, in a scalar comparison (`{qty: {$eq: <value>}}` is the shape #7872 used; extend to a second operator only if it changes the answer). Record which of these happens:

- **crash** — an unhandled throw reaches the caller (record the error, and *whose* it is: the driver's, or a third party's like mingo/knex)
- **refuse** — a clean typed refusal (record the code, and whether it carries an ADR-0112 envelope)
- **silent-empty** — 0 rows, no error (the #5240 shape)
- **works** — rows come back as if the comparand were a normal literal (record what it matched)

Five drivers: `driver-memory`, `driver-sql`, `driver-sqlite-wasm`, `driver-turso`, `driver-mongodb`. Include `driver-memory` rather than copying #7872's numbers — re-measuring it on current `main` confirms or corrects the card that started this.

## Environment constraints — declare what you could not reach

Known before you start, from the drivers seat's operating notes:

[...]

Issue #7597, opened 2026-08-11
[spec] `parseFilterAST` lowers `['a', '=', { $field: 'b' }]` to a bare field spec no backend reads as equality


os-zhuang:
Measured while implementing #5222 (cross-field push-down). Filing unassigned — recording, not claiming.

## The fact

`parseFilterAST` lowers an authored triple differently depending on the operator spelling. Measured on current `main`:

| Authored | Lowered to |
| :--- | :--- |
| `[['amount', '&gt;', { $field: 'budget' }]]` | `{"amount":{"$gt":{"$field":"budget"}}}` |
| `[['amount', 'gt', { $field: 'budget' }]]` | `{"amount":{"$gt":{"$field":"budget"}}}` |
| `[['amount', '=', { $field: 'budget' }]]` | `{"amount":{"$field":"budget"}}` ← **bare** |
| `[['amount', 'equals', { $field: 'budget' }]]` | `{"amount":{"$field":"budget"}}` ← **bare** |

The equality spellings drop the operator, because a literal comparand's implicit-equality form is `{ field: value }` — correct for a literal, but for a field reference it produces a field spec whose only key is `$field`.

**Nothing reads that shape as an equality.** In `matches-filter.ts`, `evalField` sees an all-`$` key set and dispatches `$field` to `evalOp`, which has no arm for it and returns its fail-closed `false`. So the filter silently matches no record. On the SQL side the same shape arrives as an OPERATOR named `$field`.

## Why it matters

Two authored spellings of one intent, two fates: `['amount', '&gt;', ref]` works (post-#5222 on SQL, and always in memory), `['amount', '=', ref]` silently matches nothing. The failure is invisible on the path that produces it.

#5222 narrowed the SQL side to give this shape an actionable refusal naming `$eq` rather than a generic operator list — but that is a driver-side mitigation of a lowering defect, and it deliberately did NOT compile the shape, because compiling it would make SQL answer rows for a filter the memory evaluator answers `false` for (a new divergence in the change that closed one).

## The decision this needs

Most likely: `parseFilterAST` should lower an equality triple whose comparand is a `FieldReferenceSchema` to `{ field: { $eq: ref } }` rather than the bare form — the spelling that IS implemented on both paths. That is a one-sink change (`parseFilterAST` is the single lowering sink per #5158), but it is a semantics change at the authoring face, so it wants a ruling rather than a drive-by fix.

[...]

PR #8234, opened 2026-08-12
feat(spec): comparand-type door — the accepted literal comparand set, enforced once at the shared compile face for all five drivers (#7872)

### LINKED ISSUE #7872: driver-memory has no policy for an unsupported comparand TYPE: a `BigInt` crashes with a raw mingo `TypeError`, and four other types silently answer zero rows
Found while implementing #7853 (the analytics pipeline dump's lost `RegExp` operand). **Deliberately outside that card's scope and not fixed in #7869** — this is an **execution** defect, not a dump one. #7853's surface is `AnalyticsResult.sql`, a debug string; the crash below happens inside mingo during `$match`, *before* that string is ever built, so no change to the dump reaches it. Filing rather than folding in, and filing rather than dropping: the card decides nothing, it preserves a measurement.

## Measured on `origin/main` @ `55635fc`, executed (mingo@7.2.2)

A `BigInt` comparand escapes as an unhandled third-party `TypeError` with no rows and no refusal envelope:

```
TypeError: Do not know how to serialize a BigInt
    at JSON.stringify (<anonymous>)
    at Query.compile (node_modules/.pnpm/mingo@7.2.2/node_modules/mingo/cjs/query.js:49:49)
    at new Query (node_modules/.pnpm/mingo@7.2.2/node_modules/mingo/cjs/query.js:44:10)
    at $match (node_modules/.pnpm/mingo@7.2.2/node_modules/mingo/cjs/operators/pipeline/match.js:25:13)
    at node_modules/.pnpm/mingo@7.2.2/node_modules/mingo/cjs/aggregator.js:73:36
    at Array.reduce (<anonymous>)
```

Reproducing filter, on a `number` field over a one-row table:

```ts
where: { qty: { $eq: BigInt(100) } }
```

**Both faces, identically** — this is not a face divergence. `MemoryAnalyticsService.query()` and `InMemoryDriver.find()` each throw the same `TypeError`; mingo serializes the query object to build its cache key, and `JSON.stringify` refuses a `BigInt` outright.

## The wider shape: it is comparand TYPES that have no policy, not just `BigInt`

The single-type reading would be too narrow. Every comparand type below, `{qty: {$eq: <value>}}`, measured on both faces:

| comparand | `query()` | `find()` |
|:--|:--|:--|
| `BigInt(100)` | **`TypeError`, unhandled** | **`TypeError`, unhandled** |
| `() => 1` | OK — **0 rows** | OK — **0 rows** |
| `Symbol('x')` | OK — **0 rows** | OK — **0 rows** |
| `new Map()` | OK — **0 rows** | OK — **0 rows** |
| class instance | OK — **0 rows** | OK — **0 rows** |
| `undefined` | OK — **0 rows** | OK — **0 rows** |

So one unsupported type crashes and five are answered with a silent empty result — a row count the author never asked for, which is the shape #5240 ruled on for `{field: {}}` ("both silent readings answer it with a row count the author never asked for").

## Why it reads as a defect: the face refuses filter SHAPE cleanly and comparand TYPE not at all

The contrast is what makes this more than a curiosity. Measured on the same face, filter *vocabulary and shape* problems already get clean, descriptive refusals:

```
{ qty: {}     }  ->  Error: Field constraint at where.qty carries zero operators ({ "qty": {} }). A field constraint must name at least on…
{ qty: {$wat} }  ->  Error: Unsupported filter operator "$wat" on field "qty". Supported operators: $eq, $ne, $gt, $gte, $lt, $lte, $in, $…
```

`filter-refusal.ts` is built for exactly this and its own docblocks already reason about these types — the `isFilterNode` header names "a `Date`, a `RegExp`, a `Map` or a class instance" as values that enumerate to nothing, and `isEmptyFieldConstraint` distinguishes them as COMPARANDS "left to the paths that already handle it". The measurement above is that no path in fact handles them: the refusal vocabulary covers the operator and the constraint shape, and stops at the comparand's type.

## The routing question, stated rather than answered

Whether the fix belongs in this package's `filter-refusal.ts` vocabulary, or is a `packages/spec` door question — a comparand type policy enforced once at `parseFilterAST()` for every driver rather than re-derived per driver — is a **cross-seat call that needs the PM's declaration**, not a dev's. Both readings are defensible from the measurement:

- **`filter-refusal.ts`**: the crash is driver-local (mingo's cache-key serialization), the refusal envelope already lives here, and `driver-sql`'s twin makes the same prototype check for its own reasons.
- **spec door**: the six types above are not JSON, so no driver can receive them through a REST/JSON write path — they only arrive from direct JS callers, and a policy that every driver copies is the kind of hand-copy the repo has been consolidating (`like-pattern.ts`'s header refuses a third copy of its escape on exactly that ground). Whether the sibling drivers answer these six the same way is **not measured here** and would decide the routing.

## Not measured / open for triage

- Whether the other four drivers (`driver-sql`, `driver-mongodb`, `driver-sqlite-wasm`, `driver-turso`) crash, refuse, or silently empty on the same six comparands. This is the measurement that most cheaply settles the routing question above.
- Whether ADR-0112 already rules on comparand types anywhere, which would make this enforcement of an existing decision rather than a new one.
- Grade. It is a raw unhandled third-party `TypeError` reaching the caller, but only from direct JS callers on an input no JSON path can express — plausibly S rather than higher, and left to triage.

Unassigned and unclaimed — a finding for triage to grade, not a claimed card.

huangyiirene:
Fixes #7872

Implements the maintainer ruling on this card (comment 5265944890, 「同意你的决策」): the shared filter-compilation face in `packages/spec` defines the accepted literal comparand-type set as the **measured superset — `string | number | bigint | boolean | null | Date`** (#7956's divergence matrix; the "any JSON-representable literal" wording was rejected because bigint is not JSON-representable yet accepted by 4/5 drivers) and refuses everything else loudly at the compile face, with `code: INVALID_FILTER` + `status: 400` + guidance naming the accepted set.

## One PM mechanism assumption was falsified — the measured call graph, and the adjusted wiring

The dispatch assumed `parseFilterAST` is on every driver's path. Measured on this branch's base:

- **Door 1** (`metadata-protocol/protocol.ts` ~6782) runs `parseFilterAST` **only behind `isFilterAST`**, which answers `false` for every non-array — so an **object-form** wire filter never passes through it.
- **Door 2** (`objectql/engine.ts` `lowerWhereFilterArray`) routes only the **array** branch through `parseFilterAST`; the non-array branch returned after the #5869 shape gate.
- The one seam every engine verb passes on both doors is `lowerWhereFilterArray` — the same measurement #5869 recorded when it placed the list-shape gate there.

The object form is exactly how the #7956 matrix (and the mongo worst cell) arrives, so a door wired in `parseFilterAST` alone would have missed the dominant form — the `NormalizedFilterSchema` lesson again. The adjusted wiring keeps the ruling's substance intact:

- the **definition** lives in `packages/spec` (`data/filter-comparand-type.ts` — the set, the predicate, the walk, the refusal envelope, the shared sentence);
- `parseFilterAST` judges **everything it returns** (object passthrough included), making it the compile face the ruling names — this covers Door 1's array lowering, analytics' normalizer, and every direct caller;
- the **engine's lowering seam** runs the same walk on the object form (`normalizeFilterComparandTypes`, imported from spec), beside the #5869 gate, on all six filter-collecting verbs (`find` / `findOne` / `count` / `aggregate` / `update` / `delete`).

Net: every platform path to every driver passes the door; `parseFilterAST` is now on the object form's path wherever it previously was at all. Flagged here rather than silently absorbed; if the PM reads assumption 1's STOP clause as outranking this, the wiring commit is cleanly separable.

## The bigint cell — how the ruling's two sentences were reconciled

The ruling both (a) puts `bigint` in the accepted set and (b) says the memory BigInt crash cell "becomes a refusal". A bigint passed through verbatim still crashes mingo (`Query.compile` → `JSON.stringify`), and driver-local patches are frozen (#5499) — so (a) and (b) cannot both be literal. The approved recommendation's own sentence ("mongo's silent-edit cell and memory's crash cell **both die at the door** without touching frozen drivers") is implementable, and that is what this PR does:

- an exact-range bigint (|v| ≤ 2^53) is **narrowed copy-on-write to its JS number** at the door — the same declared-conversion pattern as `RemoteTransport`'s Date → ISO — so no driver path ever sees a bigint, and the memory cell becomes a **working compile** (the conformance table pins `{qty: {$eq: BigInt(100)}}` matching its row on all five drivers);
- a bigint beyond ±2^53 is **refused loudly** (precision loss must not answer silently — the exact evil this thread outlaws);
- direct transport callers keep libsql's native bigint binding (`remote-transport-comparand-refusal.test.ts`'s pin is untouched).

The `#6915` citation in the ruling is a mis-cite (that card is driver-memory tenant isolation); the BigInt crash cell is this card's own memory×BigInt cell. Both readings are surfaced in the report's `open_questions` for the maintainer to confirm.

## Both directions pinned — `FILTER_COMPARAND_TYPE_CASES`, run by all five drivers

New conformance table in `spec/data` (registered in `scripts/check-driver-conformance.mjs`; the matrix is now 45 covered / 0 DEBT / 0 exempt):

- **accepted direction**: each of the six types compiles on every driver path — executed on better-sqlite3 (driver-sql), sql.js (driver-sqlite-wasm), libsql local (driver-turso), mingo (driver-memory), and — server-free, as on #7956 and the sibling conformance suites, since a live mongod is unreachable in CI — `translateFilter` + the mongodb package's own BSON round trip (driver-mongodb). Stated in that suite as the accepted substitute.
- **refused direction**: `undefined` (operator form, implicit form — the mongo worst cell — and as an `$in` member), function, Symbol, Map, Set, class instance, plain object in a scalar slot, oversized bigint — all refused at the door with the envelope, and the suites assert **no driver ever runs** (the engine suite pins zero driver reads/writes on refusal).
- **reverse-direction pins**, so the door's work cannot be mistaken for driver changes: driver-memory pins that the direct-path bigint crash *still exists* (mingo unfixed, frozen); driver-mongodb pins that `{qty: undefined}` *still* BSON-encodes to `{}` = match everything on the direct path. Both notes say when to retire the pins.

## SQL-family allowlists reconciled to the door, not duplicated

- `driver-sql` `isBindableComparand` / `isRenderableTextComparand` now delegate type membership to `isAcceptedFilterComparand`; refusal messages quote `ACCEPTED_FILTER_COMPARAND_TYPES_SENTENCE`. Two recorded driver-local extras stay, each with its reason in place: the `undefined → true` arm (unreachable behind the #6050 walk, which owns the purpose-written undefined refusal) and `ArrayBuffer.isView` (blob bindables — a direct-caller surface the engine door does not admit; nothing on engine paths uses them, measured by grep and by the suites).
- `driver-turso` `serializeComparand` delegates the scalar membership to the same predicate; its pinned refusal wording ("must be a string, number, bigint, boolean, null or Date") is now the shared constant, **byte-identical**, so reconciling changed no message.
- Their ADR-0112 envelopes are untouched — the ruling's "keep their local raise but source the SET from the door" fallback was not needed beyond the two recorded extras above.

## Boundaries the door deliberately does not judge (each has a recorded owner)

- `FieldReference` (`{ $field: … }`) at any position — #5222 / #7596 / #7597 own its per-position fate.
- Nested-relation / deep-equality structure (a plain object with no `$` keys) — the #5869 boundary, kept verbatim.
- Arrays outside `$in` / `$nin` / `$between` — unmeasured, unruled, answered per driver today.
- Unknown / retired operators (`$wat`, `$regex`) — the downstream refusals carry the specific prescriptions (`RETIRED_FILTER_OPERATORS`); a generic type refusal would preempt a better message.
- Non-array list-operator comparands (`{ $in: 'won' }`) — the engine's #5869 shape gate, whose #5346/#5348 wording is pinned by consumers.

## Changeset: minor, argued from the diff

Precedent #3746 (acceptance narrowing at a validation face, minor) fits what this diff does: no authorable key is removed, no export is removed, and every newly-refused input was measured answering a crash, a silent wrong row count, or a silent match-all — not a documented capability. The one previously-"working" behavior that narrows is mongo/memory deep-equality via `$eq: {plain object}` **through engine paths** (direct driver calls unchanged) — measured divergent 3-vs-2 across drivers, which is the divergence the ruling closes. #4583-major (removal of a declared key with migration prescription) does not describe this diff. Not declared-breaking ⇒ no ADR-0087 marker required; the changeset carries the full caller-facing delta anyway.

## Verification

All local runs serialized on the shared verify lock, `NODE_OPTIONS=--max-old-space-size=4096`, scoped filters.

Pre-merge batch (base 37b82ed5b):

- `@objectstack/spec` test — full suite green (386 files; earlier baseline run: 10236 tests, the two reds it surfaced — a 521-char refusal message and the stale export snapshots — were fixed/regenerated and the suite re-ran green).
- `@objectstack/objectql` test — 191/192 green; the one red was `engine-filter-array-lowering.test.ts`'s pass-through pin for `{ $eq: { $in: 'not-an-operator' } }`, a fixture that pinned exactly the limb the ruling re-judges (plain object in a scalar slot). Re-judged per the fixture-triage rule: the case now asserts the door's TYPE refusal at `where.stage.$eq` **and** that the message never misreads the inner `$in` data key as a malformed operator (`not.toContain('requires an ARRAY')`) — the old pin's surviving concern, kept.
- `@objectstack/metadata-protocol` (Door 1), all five driver suites, `@objectstack/service-analytics`, `@objectstack/runtime` + `@objectstack/rest` — green (runtime: 145 files / 2195 tests).
- typecheck ×7 (spec, objectql, five drivers) — green.
- `check:driver-conformance` — **45 covered / 0 DEBT / 0 exempt** (was 40/0/0; the five new cells are the door column, all covered in this PR). `check:nul-bytes` — green (7424 files).
- `check:generated` — 2 of 13 stale (`api-surface/`, `export-origins/` — the 13 new exports), regenerated with `--fix`, additions only.

Post-merge lap (merged `origin/main` @ fa4897335; incoming touched `objectql/src/engine.ts` — the same file as the door wiring — and `spec/src/contracts/*`): rebuilt the closure, then re-ran the moved packages — spec `check:generated` green (artifacts still current), spec test **386/386 files green**, objectql test **193/193 files green** (the re-judged pin and the new door suite included).

Examples validate ×3: **all green** (`app-crm`, `app-showcase`, `app-todo` — "Validation passed"; todo's two draft-flow warnings pre-exist this branch). The first two attempts failed on this worktree's own build state (`packages/cli` and `connector-mcp` had no dist — `command validate not found` / missing module), not on the diff; building the examples' dependency closures by package name fixed both.

## Out-of-scope findings

- `service-analytics` carries a third and fourth copy of the same six-type membership (`comparand-shape.ts` :116/:142, message twins in `read-scope-sql.ts`) — outside this dispatch's declared file surface, filed as #8186 for reconciliation to the door.

---
_Generated by [Claude Code](https://claude.ai/code/session_0123k4cam2jEAkPmbJeoaY3r)_


huangyiirene:
**os-regen relay lap over #8230 / #8232 — complete.** Merge sha `2abb1e8ee` (merges `origin/main` @ `29488ccae`), regen commit `81f6752e4`, both pushed; head is `81f6752e4`. Ran via `scripts/pm/os-regen-merge.sh` (merge → take main's side of every os-regen path → commit → regenerate from the merged tree; the pre-commit deferral marker recorded and cleared itself).

Sibling-entry assertions on the regenerated artifacts:

- **#8230 retirement state survives**: exactly its 8 retired-def entries present (`17.data__ExternalDataSource` / `ExternalFieldMapping` / `ExternalLookup`, `17.system__ConsumerConfig` / `DeadLetterQueue` / `MessageQueueConfig` / `MessageQueueProvider` / `TopicConfig` — verified against that merge commit's own file list). Zero retired-family def names in `authorable-surface/` or `json-schema.manifest/`. One precision: `kernel.json` in both trees carries 10 `MessageQueue` substring hits — byte-identical to post-#8230 `origin/main` (empty diff), i.e. the kernel-domain plugin family #8230 did not retire, not a regen rollback.
- **#8232 currency additions survive**: `CurrencyConfig` / `CurrencyConfigParsed` / `CurrencyConfigSchema` present in the regenerated `api-surface/data.json`.
- **My own surface survives**: all 13 comparand-type-door exports in `api-surface/data.json`; conformance column intact — `check:driver-conformance` OK, 45 covered / 0 DEBT / 0 exempt.

Gates and suites on the new head: `check:generated` — **all 13 artifacts up to date**; spec scoped suites (comparand-type door, filter, operator vocabulary, text conformance, export-origins) — **5/5 files green**; objectql scoped (door suite + lowering suite) — **2/2 files, 58 tests green**. One false red on the way: the recreated worktree had no `@objectstack/types` dist, so the first objectql run failed at import resolution — fixed by building the dependency closure, not by touching the diff. `engine.ts` re-merged cleanly (main does not carry #8236 yet, as expected).

PR stays **draft**; over to the PM for gate verification on `81f6752e4` and the flip.

---
_Generated by [Claude Code](https://claude.ai/code/session_0123k4cam2jEAkPmbJeoaY3r)_

---

huangyiirene:
**Patch round for the two red farm jobs — fixed, new head `dee462fdb`.** Both reds were mine (verified — neither reproduces from `origin/main` content; both point at the door suite added in this PR):

- **TypeScript Type Check** (`check:type-check-debt --re-measure`): objectql TEST_DEBT 355 → 356 (+1). The new error was `engine-comparand-type-door.test.ts`'s `registerObject(deal)` missing the required `packageId` argument (TS2554 — the sibling lowering suite carries the same call inside the frozen 355; mine was the increment). Fixed by passing the id (`registerObject(deal, 'test')`), not by raising the ledger. Re-measured locally with the exclusion lifted, exactly as the gate counts: **355 — back at the frozen number, zero errors from this PR's files** (exit 0).
- **ESLint** (`check:query-options-erasure`): test surface 240 → 245. The door suite's option bags used bare `as any`. Now they type-check **uncast** — `FilterCondition`'s index signature admits the deliberately off-contract comparands, which is precisely why the runtime door has to exist (noted in the test) — except the one `FilterArray` case, which carries the sanctioned `as unknown as EngineQueryOptions` form naming the contract it bypasses (not counted). Local re-run: **ratchet holds, 240 test sites, "none new"** (exit 0).

Behavior re-verified after the edits: door suite + lowering suite — **2 files / 58 tests green** (exit 0). Diff is one test file; no source, no artifacts, no changeset delta. PR stays draft on `dee462fdb` for the farm re-read.

---
_Generated by [Claude Code](https://claude.ai/code/session_0123k4cam2jEAkPmbJeoaY3r)_

---

Issue #8215, opened 2026-08-12
A `preserveAudit` by-id update hands the primary key to the driver: `isPreservableUnderAudit` reinstates a platform object's own `id`


os-zhuang:
Measured while implementing #8141, out of its ruled scope, and not the same fact as #6435's "truthy-scalar `data.id`" block — this one lands on objects whose `id` **is** `readonly: true`, i.e. every platform object, and it is the historical-import path that reaches it.

## What happens

`isPreservableUnderAudit` (`packages/objectql/src/validation/rule-validator.ts`) is:

```ts
function isPreservableUnderAudit(name: string, def: ConditionalFieldDef): boolean {
  if (AUDIT_TIMELINE_FIELDS.has(name)) return true;
  return def.system !== true;                 // author-declared business readonly field
}
```

A platform object's primary key qualifies under the second limb. `sys_user_preference` declares it as `Field.text({ label: 'Preference ID', required: true, readonly: true })` (`packages/platform-objects/src/identity/sys-user-preference.object.ts`) — `readonly`, **not** `system` — so to this predicate it reads as an author-declared business field like `closed_at`.

So on a by-id update carrying `{ context: { preserveAudit: true } }`, `stripReadonlyFields` **keeps** `id` instead of stripping it, and the payload reaching the driver is `SET id = 'rec_1' WHERE id = 'rec_1'`. The REST ingress folds the path id into every update body (#6479), so the caller does not have to send an `id` for this to happen — an import doing `PATCH` with a body that names no `id` at all still hits it.

Pinned as observed behaviour, deliberately, in `rule-validator.test.ts` ("composes with preserveAudit rather than overriding it", #8141): the address survives the strip one line before the address rule is consulted.

## Why it is worth a card

The strip exists because a primary-key write is not uniformly harmless — #8141 and #6435 both restate the reason: **a same-value primary-key write is a no-op on SQL but an outright rejection on stores with immutable primary keys.** This path removes that protection from exactly the caller least able to diagnose it: a bulk historical import, whose whole payload is legitimate except a key it never typed.

The exemption is also not the one the flag was designed for. #3493 scoped `preserveAudit` to "reinstate the original timeline" — the audit family, plus business fields an importer really is restoring. A row's own primary key is neither: it is the address of the write, not a fact being restored, and the import is not asking to rewrite it.

## Shape of a fix (for triage, not a ruling)

[...]

Issue #7532, opened 2026-08-11
Data query: a dotted projection entry silently WIDENS the response — `fields:["name","account.name"]` returns every field


huangyiirene:
## Symptom

`POST /api/v1/data/showcase_invoice/query` with `{"fields":["name","account.name"]}` answers **`200` returning every business field** of each record, and no resolved `account.name`.

**Expected:** the projection narrows to the named columns (with the dotted path resolved), or — if a dotted path is not supported on this door — a `400 INVALID_FIELD`.

Measured across three doors and two object pairs in the same run:

| request | observed | expected |
|---|---|---|
| `POST /data/showcase_invoice/query {"fields":["name","account.name"]}` | `200`, **every** field, no resolved `account.name` | narrowed projection |
| `GET /data/showcase_invoice?$select=name,account.name` | `200`, **every** field | narrowed projection |
| `POST /data/showcase_task/query {"fields":["title","project.name"]}` | `200`, **every** field | narrowed projection |
| control — `POST /data/showcase_task/query {"fields":["title"]}` | `200`, correctly narrowed | — |
| a **nonexistent dotted** column | `200` + every field (swallowed) | `400 INVALID_FIELD` |
| a **nonexistent plain** column | `400 INVALID_FIELD` | — |

Two things are wrong at once: the request is answered with strictly **more** data than it asked for (a projection is how a caller keeps heavy or unwanted columns out of a response, and nothing in the body says it did not apply), and the two spellings of the same mistake — plain vs dotted unknown column — get opposite verdicts on one endpoint. The `400 INVALID_FIELD` message the plain spelling produces names this exact failure mode.

## Root cause

`assertProjectionFieldsExist` (`packages/metadata-protocol/src/protocol.ts`) validates only `f.split('.')[0]`, so a dotted entry passes the ingress gate on its **head segment** and then defeats the projection downstream — the unknown projection name is dropped and the projection falls back to `*` (the engine's documented per-axis tolerance for internal callers), i.e. over-return instead of refusal.

Neighbour: **#4226** (closed/completed) is the card that closed the `select` / `sort` / `expand` axes against silently-dropped unknown fields. Plain names behave per #4226 in this run; the dotted spelling is the leg its gate does not cover, so treat this as an uncovered leg of — or a regression against — #4226.

## Reproduction

1. Boot showcase on a fresh isolated file DB (`SqlDriver` / better-sqlite3); authenticate as `admin@objectos.ai`.
[...]

Issue #7501, opened 2026-08-11
A number field's declared `scale` is never enforced — values with more decimals are accepted and stored verbatim (min/max on the same field are enforced)


baozhoutao:
## Summary

A `number` field that declares `scale` is **not validated against it at runtime** — a value with more decimal places than `scale` allows is accepted and stored verbatim. There is no error, no rounding, no warning.

Measured on `@objectstack/*@17.0.0-rc.6`, `driver-sql` (better-sqlite3), via both the REST create endpoint and the CSV import endpoint.

## Repro

Field declaration:

```ts
work_hours: {
  type: 'number',
  label: 'Max hours per shift',
  precision: 5,
  scale: 0,        // ← declares "integer, no decimals"
  defaultValue: 12,
  min: 1,
  max: 12,
}
```

`POST /api/v1/data/<object>` with `{ "work_hours": 11.5 }`

- **Expected**: rejected (`400`, an `INVALID_*` envelope), or at minimum rounded to the declared scale.
- **Actual**: **`HTTP 201`**, and the stored value is `11.5` — verified by reading the row straight out of storage, not through the API.

Same result with `scale: 1` and a two-decimal input, and same result through
`POST /api/v1/data/<object>/import` (`format: "csv"`).

`min` / `max` on the same field **are** enforced (`-1` → `400`, `code: "min_value"`), so the validator does run — it just has no branch for `scale`.

## Where it comes from

`record-validator`'s `number` branch only tests `def.min` and `def.max`. A full search of
`objectql` / `runtime` / `cli` for `scale` finds it **only** in `@objectstack/spec` as descriptive
metadata (`"Decimal places"`) — there is no runtime consumer anywhere in the shipped packages.

So `scale` today is documentation, not a constraint. That is surprising given it sits next to
`precision`, `min` and `max` in the same field declaration, all of which read as constraints.

## Why it matters

A downstream app declared `scale: 0` to express "this must be a whole number" and shipped it,
reasonably assuming the declaration was enforced the way `min`/`max` are. It is not — decimals
flow through to storage. There is no way to express "integer" in the field contract at all today,
so every such rule has to be re-implemented in an application hook, one field at a time.

## Ask

Either of these would close it:

1. **Enforce `scale`** in the number validator (reject, or round to the declared scale — either is
   predictable, silence is not), consistent with how `min`/`max` already behave; or
2. **If `scale` is intentionally display-only**, say so in the spec description (it currently reads
[...]

PR #8327, opened 2026-08-13
fix(objectql)!: engine.find/findOne refuse a dotted projection instead of widening to every field (#7589)

### LINKED ISSUE #7589: finding: SqlDriver's #3821 recovery ladder widens an unresolvable projection to every field
## What this is

A **finding**, filed as defence-in-depth rather than a live user-facing defect. Extracted while implementing #7532; the measurement is mine and is reproduced in full below.

Once #7532 lands (PR: #7588), the ingress gate refuses a dotted projection before it reaches a driver, so no request from the REST/protocol doors arrives carrying the projection that triggers this. What remains is every caller that reaches `SqlDriver` **without** passing that ingress — hooks, flows, reports, expand sub-reads, registry-less hosts — for which this behaviour is unchanged.

## Observed

`SqlDriver`'s unknown-column recovery ladder (`packages/drivers/driver-sql/src/sql-driver.ts`, ~L3395-3425) catches `no such column` and retries the query with `select('*')`. The result is that a projection the driver cannot apply comes back as **every column**, with a 200 and nothing in the response indicating the projection was dropped.

Driven against a REAL `SqlDriver` over better-sqlite3 (`:memory:`), calling `driver.find` **directly** — no engine, no protocol, so this is the driver's own behaviour and not the ingress gate's:

```
A  no projection (baseline)      -> [account, amount, created_at, id, name, status, updated_at]
B  fields ['name']               -> [name]                                    <- narrows correctly
C  fields ['name','account.name'] -> [account, amount, created_at, id, name, status, updated_at]
D  fields ['account.name']        -> [account, amount, created_at, id, name, status, updated_at]
```

Rows C and D are **byte-identical to row A** — to asking for no projection at all. Knex renders `"account"."name"` against a table that was never joined, sqlite answers `no such column`, and the ladder retries selecting everything.

Reproduction harness (drop into `packages/drivers/driver-sql/src/`):

```ts
import { describe, it, expect, beforeEach, afterEach } from 'vitest';
import { SqlDriver } from '../src/index.js';

describe('SqlDriver projection widening', () => {
  let driver: SqlDriver;
  beforeEach(async () => {
    driver = new SqlDriver({
      client: 'better-sqlite3',
      connection: { filename: ':memory:' },
      useNullAsDefault: true,
    });
    await driver.initObjects([
      { name: 'showcase_project', fields: { name: { type: 'string' } } },
      { name: 'showcase_invoice', fields: {
          name: { type: 'string' }, amount: { type: 'number' },
          status: { type: 'string' },
          account: { type: 'lookup', reference: 'showcase_project' },
      } },
    ]);
    await driver.create('showcase_project', { id: 'p1', name: 'Apollo' } as any);
    await driver.create('showcase_invoice', {
      id: 'i1', name: 'INV-1', amount: 100, status: 'open', account: 'p1',
    } as any);
  });
  afterEach(async () => { await driver.disconnect(); });

  it('widens instead of narrowing', async () => {
    const rows: any[] = await driver.find('showcase_invoice', { fields: ['name', 'account.name'] } as any);
    // Observed: every column. A projection that cannot be applied returns MORE.
    expect(Object.keys(rows[0]).sort()).toEqual(['name']);
  });
});
```

## Why it is deliberate today, and why it is still worth a card

The ladder is documented and intentional — its own comment explains the trade: a `$select` naming a column the table lacks used to make the WHOLE query fail, which read to the UI as "no records exist" while rows were there. Returning the rows without the projection was judged the lesser harm, and it explicitly backstops hosts whose registry is not populated ("notably the cloud multi-tenant runtime, where the projection otherwise zeroes the list").

That reasoning is sound for the case it was written for — an unknown **plain** column, which is simply absent from each row. It generalises badly to a projection the driver cannot apply for a structural reason: the caller asked to narrow, and silently received everything, which points away from both FLS and data minimisation. Nothing in the response distinguishes "your projection was applied" from "your projection was discarded".

## Not fixed here, deliberately

Narrowing a documented tolerance is a contract change with its own blast radius and needs its own reverse verification — it is not a rider on #7532, and #7532's ruling was explicit that the ingress gate is where that card lands. Filed unassigned and without a `domain:*` label so routing stays the triage seat's call.

Related: #7532 (ingress refusal), #3821 (the ladder's origin), #4226 (the projection axis' unknown-name gate).

---

os-zhuang:
Fixes #7589

Implements the maintainer ruling recorded 2026-08-12 (issue comment 5266074163): **Option B, verbatim** — a dotted `fields` entry the engine cannot resolve is refused loudly at the engine's head-only projection filter, and the false populate comment is deleted in the same change.

## What changed

`engine.find()` / `engine.findOne()` now throw `400 INVALID_FIELD` (ADR-0112 envelope: `code` + `status`, plus `field`/`fields`/`object`) on any `fields` entry carrying a dot. The head-only `known.has(head)` keep — justified by a comment claiming the engine resolves relationship paths "via populate", which #7601 measured does not exist (it was the last such assertion in the repo after PR #7617) — is gone; the plain-name filter now matches whole names. The refusal sits before the formula planner, on the caller's own spellings, in both verbs.

The measured chain (issue comment 5263464190) reaches **both** verbs: flow `get_record` config goes to `data.find(...)` when `limit` is greater than 1 and to `data.findOne(...)` otherwise — a flow authoring `fields: ['name','account.name']` now gets a loud node failure carrying the remedy instead of a silent 200 with every column.

## Explicitly KEPT (the ruling's carve-outs)

- **Unknown-PLAIN-column tolerance** — an unknown plain name still drops silently; an all-unknown projection still falls back to `*`. Pinned as guards so a future tighten cannot ride in on this card's precedent.
- **Registry-less hosts** — no field map, no verdict (same early-return as the ingress gate); the driver-side #3821 ladder remains their documented backstop. Pinned.
- **Option C is NOT here** — no change under `packages/drivers/`; the driver-side carve-out stays measured-need only per the ruling.
- Sibling cards' files untouched: `validation/record-validator.ts` (#7501), `validation/rule-validator.ts` (#8215).

## Every face reaching the filter, enumerated

1. **`find`** — refusal added before `planFormulaProjection`.
2. **`findOne`** — same refusal, same position (the inline head-split copy of the filter is rewritten too).
3. **Expand sub-reads** — route through `this.find`, so the refusal fires there — inside `expandRelatedRecords`' pre-existing graceful-degradation `catch`: outcome is an observable warning + retained FK ids, **not** a refusal. Pinned, with a plain-nested-fields control; same posture the sort axis (#7095) records for the same catch.
4. **Formula planner** — never invents or removes dotted entries (passes caller entries verbatim, injects only plain schema names + `id`); judged before it runs.
5. **Search expansion** (`expandSearchOnAst`) — writes only `where`; no projection face.
6. **`count` / `aggregate` / `update` / `delete`** — carry no `fields` option (`ENGINE_OPTION_KEY_SETS`); no projection face.
7. **Protocol ingress** (`findData`) — already refused at ingress since PR #7588; the engine door is the second line behind it. All existing ingress pins stay green.
8. **`plugin-reports`** — forwards a saved report's `query.fields` verbatim into `engine.find`; an author-reachable surface that now fails loudly with the remedy (recorded in the ADR-0087 entry).

## Semantics notes

- **All** dotted entries are refused, including one with an unknown head (`no_such.name`). Deliberate divergence from the ingress door's `unknown` before `dotted` precedence: the engine has no unknown-name refusal to defer to (unknown plains are tolerated by the kept carve-out), and a dotted entry with an unknown head still widened through the empty-projection fallback when it was the only entry. Pinned with the reasoning in the test.
- The refusal wording shares its core sentence and both remedies (`expand`, denormalise onto a stored field) with the ingress door's dotted refusal — duplicated because `metadata-protocol` is assembled from an engine (importing would invert the layering), kept honest by a cross-door agreement pin, same mechanism as the sort axis' remedy pin.

## Changeset + ADR-0087

Breaking changeset (`minor` bump + bang title per `check-changeset-no-major` convention) with the disposition marker `registered engine-dotted-projection-refused`, and the semantic migration entry added under `packages/spec/src/migrations/entries/semantic/` — mirroring #7095's `engine-find-formula-order-by-refused` (same door, same class: silent degradation becomes a refusal). Registry region, `spec-changes.json` and the upgrade guide regenerated; `check:generated` reports all 13 artifacts current.

**Declared delta beyond the dispatch's file surface** (engine.ts + its tests): the ADR-0087 registration kit above. It is the mechanical accompaniment `check:adr-0087-registration` requires of a declared-breaking changeset — data-only, no spec schema or behavior change. Flagged here and in the dev report rather than silently included.

## Verification

- `@objectstack/objectql`: full suite **3466/3466** green (196 files); conformance file 145/145; `typecheck` clean.
- **Reverse verification, from committed state, direction predicted before running (red)**: with `origin/main`'s `engine.ts` restored and the new pins kept, exactly the **9** new refusal/agreement/expand pins go red and all **136** others — every control and every KEPT-tolerance guard — stay green. Fix restored from its commit, byte-identical (`git diff` against the fix commit: empty).
- Downstream consumers (prefix filter, consumer direction): `@objectstack/service-automation` **955/955**, `@objectstack/metadata-protocol` **1117/1117** — green after building their dependency closures (the first lap's 4 file-level failures were unbuilt `service-job` dist, the known missing-build false red).
- Gates: all 7 dispatch-named families green (`check:adr-anchors`, `check:durability-log-level`, `check:engine-double-contract`, `check:stack-collection-maps`, `check-engine-split-ratio` informational, `check:error-code-casing`, `check:nul-bytes` re-run after the final tree), plus all 16 families re-derived from the actual changed paths (`dispatch-gates.mjs`) — `check:i18n` and `check-dev-prereqs` went green after a full workspace build; their first reds were unbuilt-dist prerequisites, named as such by the gates' own output.

Out of scope, untouched, and staying open on their own terms: the driver-side ladder in `driver-sql` (#3821 behaviour is unchanged; a carve-out there is measured-need only), and the dotted SORT leg on direct engine calls (a different axis; noted in the updated scope docblock). #7532 is the merged ingress half and is not modified here.

---


os-zhuang:
Patch round 1 (PM review): `5065b357` corrects `content/docs/protocol/objectql/types.mdx` — the lookup "Query behavior" example was the populate premise itself (`fields: ['name', 'account.company_name']  // Expands account`), i.e. the last hand-written doc teaching of the spelling this PR makes throw. Replaced with the `expand` form, reference column kept projected (#7537), refusal + remedy stated inline. Swept `content/docs` (releases excluded) for other dotted-projection teachings: this was the only occurrence — the old example's head (`account`) did not even match the page's own declared field (`account_id`); the corrected example uses `account_id`.

`content/docs/releases/v17.mdx:977` (the removed-object-form row prescribing "a dotted path for a single related column") is release-owned and deliberately NOT touched here — filed as unassigned docs-only card #8329 with the correction spelled out.

Gates for the docs delta, re-derived via `dispatch-gates.mjs` (4 families) plus the two run on principle: doc-formula-expressions (lint-filter form) ✓ (after building lint's dep closure in the re-created worktree — first red was the unbuilt-dist prerequisite shape again), docs-audit-scope ✓, quick-reference-counts ✓, role-word ✓, doc-authoring ✓, nul-bytes ✓. `engine.ts` untouched this round.

---

Issue #8318, opened 2026-08-13
spec: union-branch selection policy now has two sibling implementations INSIDE one package (error-map renderer vs the D3 structural mapper)


hotlong:
Observation-class finding from #8124 (no defect today — the two agree, and tests on both sides pin the shared verdict).

## What

`packages/spec/src/shared/error-map.zod.ts` (the string renderer: `formatZodIssue` / `selectUnionBranches` / `CONTAINER_ISSUE_CODES` / the depth and branch limits) and `packages/spec/src/api/zod-issues-to-fields.ts` (the ADR-0114 D3 structural mapper, moved into spec by #8124) carry the SAME union-branch selection policy — kind-mismatch drop, fewest-issues ranking, `unrecognized_keys` tie-break, declaration-order determinism, depth limit 3, branch cap 3, container descent per #5389 — as two separate implementations.

## Why it was tolerable before, and what changed

While the structural mapper was module-local to `@objectstack/rest`, the duplication was forced: spec exported only the string renderer, and the comment on the rest side documented the pairing ("the *verdict* must match all the same", #5014). #8124 moved the mapper into spec, so the historical reason is gone: both implementations now live one directory apart in the same package. The module headers bind them to move together in the same PR, but nothing mechanical enforces it.

## Possible shape (not ruled)

Extract the selection policy (`isKindMismatchOnly`, `carriesUnknownKey`, `selectUnionBranches`, `CONTAINER_ISSUE_CODES`, the two limits) into one package-internal module both walks import — internal only, NOT a public export (the #4001 pitfall: do not export internals only these modules need). The walks themselves stay separate (prose lines vs `{field, code, message}` entries; the renderer also emits the "… and N more branches" line the wire deliberately omits). The one signature wrinkle: the renderer's `selectUnionBranches` returns `{selected, omitted}`; the mapper ignores `omitted`.

Deliberately NOT done in #8124: the card's scope was ending the wire's code leak with a behavior-identical move, and rewriting `error-map.zod.ts` in the same PR would have widened the test surface the move's proof rests on.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Euoy6wyfzgiWtgCg4s6JK2)_

hotlong:
[...]

PR #8335, opened 2026-08-13
fix(spec,types,rest,runtime): one ADR-0114 D3 mapper in @objectstack/spec — fieldsFromZodIssues stops leaking Zod issue codes onto fields[].code

### LINKED ISSUE #8124: `fieldsFromZodIssues` leaks Zod's issue codes onto the wire's `fields[].code` — the exact pass-through ADR-0114 D3 closed, still live in `@objectstack/runtime`

hotlong:
Fixes #8124

## What

`fields[].code` is declared as the closed ADR-0114 D2 catalog (`FieldErrorCode`), but `fieldsFromZodIssues` — the helper the runtime `/analytics`, `/notifications` and `/automation` entry refusals emit through (moved to `@objectstack/types` by #8016) — assigned Zod's own `issue.code` verbatim, so `unrecognized_keys` / `too_small` reached the wire on a position the spec gives clients a schema for. Per the PM ruling on the issue (route 2), the compliant D3 mapper now lives in `@objectstack/spec`, beside the catalog it is total over, and is the ONE implementation of D3's table in the repo.

- **`@objectstack/spec`**: new `api/zod-issues-to-fields.ts` — the mapper moved verbatim from `rest-server.ts` (code table, #5014 union-branch expansion, #5389 container descent), exported as `zodIssuesToFields` from `@objectstack/spec/api`. `api-surface/` and `export-origins/` regenerated: exactly 1 added / 0 removed.
- **`@objectstack/types`**: `fieldsFromZodIssues` maps through the shared mapper. Signature preserved, plus an optional trailing `input` that upgrades a missing required property from `invalid_type` to `required` (the D3 split); callers without the input degrade per the D3 table. The `'(body)'` spelling for root-level failures is preserved.
- **`@objectstack/rest`**: `rest-server.ts` deletes its module-local copy and re-exports the shared one. Behavior-identical — proven by the pre-existing pins (`zod-field-codes.test.ts`, `zod-union-fields.test.ts`, 29 tests, untouched and green against the re-export).
- **`@objectstack/runtime`**: two hand-spelled `unrecognized_keys` literals on the same wire slot (the analytics `filters` hint, the automation toggle unknown-key refusal) now say `unknown_field`, the catalog member; the automation comment that deferred to #8124 is updated. These two files sit outside the dispatch's declared file surface — rationale in the dev report on the issue (the dispatch's own acceptance line requires that nothing outside the mapper writes a raw Zod code onto `fields[].code`, and no test pinned the old literals).
- **Docs**: `content/docs/api/error-handling-server.mdx` imported `zodIssuesToFields` from `@objectstack/rest`, which never exported it from its index — the example now imports from `@objectstack/spec/api`, which this PR makes true.

## Tests

Per ADR-0114 D3's own discipline, all new tests drive REAL `safeParse` calls (the real `FlowSchema` — the #8055 fixture source — and `MarkNotificationsReadRequestSchema`), no hand-written issue fixtures, and assert catalog membership of every emitted code.

- new: `packages/spec/src/api/zod-issues-to-fields.test.ts` (7), `packages/types/src/validation-failure.test.ts` (4)
- suites: spec 10256, types 250, rest 1809, runtime 2306 — all green; typecheck green on all four
- reverse verification (predicted red, observed red): restoring the pass-through `code: issue.code` in `@objectstack/types` turns the new types tests red on exactly the leak assertions while the `'(body)'` pin stays green — details in the dev report

## Out of scope

#8318 (filed): the union-branch selection policy is now duplicated INSIDE spec (string renderer vs structural mapper) — observation-class, not fixed here to keep the move behavior-identical.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Euoy6wyfzgiWtgCg4s6JK2)_


hotlong:
Audit (spec-seat PM, session `session_01Euoy6wyfzgiWtgCg4s6JK2`): the `Lint & Type Check` and `CI` runs on head `e5d3457` concluded **cancelled** (not failed — no red signature; branch has no newer push, so this reads as a concurrency-group preemption artifact). Both runs re-queued via rerun at 06:35Z. Flip-to-ready + auto-merge follows their real conclusions; this PR touches generated artifacts and lands serialized with #8341/#8342 (one regen PR in the queue at a time).

---

PR #8363, opened 2026-08-13
fix(pm): derive check:i18n by owning-package walk in dispatch-gates

### LINKED ISSUE #8352: dispatch-gates.mjs cannot derive check:i18n from an object-definition edit — the checker discovers its population at runtime, so no path literal exists to match
Found on PR #8348 (#8118), where it cost the PR a CI round: the dispatch prompt's gate list and a `dispatch-gates.mjs` re-derivation against the changed paths both omitted `check:i18n`, and the PR went red in the `TypeScript Type Check` job on `services/service-messaging DRIFTED (4)` — the object-definition edit (`internal: true` + a new field `description` on `sys_http_delivery.headers_json` in `packages/services/service-messaging/src/objects/http-delivery.object.ts`) regenerates that package's four translation bundles, and nothing in the derivation could say so.

## Why the derivation structurally misses it — measured, both halves

1. **The path half cannot match.** `dispatch-gates.mjs` matches a check to input paths by scanning the check script's own source for path literals ("watch hints"). `scripts/check-i18n-bundles.mjs` names none for the population it guards: it discovers its targets at runtime by walking `packages/` for files *named* `i18n-extract.config.ts` (its `findExtractConfigs`, ~line 106-112 — a `readdirSync` walk matching on the filename with `p.includes('/scripts/')`). The only package-path strings in its source are **self-test fixture assertions** (error-message literals about `packages/platform-objects/...`), which are not hints about what the gate watches — and would point at the wrong package if they were. So an edit under `packages/services/service-messaging/src/objects/**` has nothing to match against.
2. **The convention half does not know this kind.** `CHANGE_KIND_GATES` carries exactly one kind today (`adds or edits a test file`). "Edits a file in a package that owns an `i18n-extract.config.ts`" is not a kind it knows, so the convention derivation cannot name `check:i18n` either. The script's own output honestly labels this residue a PM judgment call — which is exactly what failed here: two careful readers (the dispatching PM and the dev's re-derivation) both missed it.

## The mechanical fix that respects the script's no-embedded-list philosophy

The trigger IS derivable at dispatch time without embedding any list: walk `packages/` for `i18n-extract.config.ts` **exactly the way the checker itself does**, and match any input path that sits inside a package owning one → `pnpm check:i18n`. That is the same runtime-discovery approach the script already takes for workflows (`extractCheckInvocations` re-reads `.github/workflows/*.yml` every run) and for `package.json` script resolution — when a tenth package grows a bundle, the next derivation run sees it, no list to rot. It could land either as a second `CHANGE_KIND_GATES` entry whose `matches` does the walk, or as a small "runtime-discovered watch roots" extension to the path half; the former is smaller.

A narrower variant (match only `src/objects/**` inside such packages) would under-cover: the extraction reads whatever each package's config enumerates, and the config file itself is also part of the trigger surface.

## Scope

`scripts/pm/dispatch-gates.mjs` only. Not the gate itself — `check-i18n-bundles.mjs` is correct and caught the drift; this card is about the derivation that is supposed to predict it. Sibling but distinct: #8162 (the derivation's self-tests run in no CI job).

---

hotlong:
Fixes #8352

`check:i18n` was invisible to `dispatch-gates.mjs` in **both** halves of its output. Measured on PR #8348: an object-definition edit under `packages/services/service-messaging/src/objects/` regenerates that package's four translation bundles, and the derivation named the gate nowhere at all — costing that PR a CI round.

## Why both halves missed it

- **Path half.** `scripts/check-i18n-bundles.mjs` discovers its targets at runtime, walking `packages/` for files *named* `i18n-extract.config.ts` (`findConfigs`, ~line 107 — the card cites it as `findExtractConfigs`; the mechanism it describes is exact), so its source names no population path to match against.
- **Sharper than the card stated.** The checker is not hint-less: it carries **eleven** path-ish literals. They are its CLI-prerequisite and stale-dist checks (`packages/cli/dist/commands/i18n/extract.js`, `packages/spec/dist`), none of which describes the population it guards. Because `hints.length` is non-zero, the gate is also filtered out of the "repo-wide / undetermined" bucket. Matching nothing **and** being excluded from the fallback bucket is what made it invisible rather than merely unmatched — a gate the derivation cannot mention at all.
- **Convention half.** `CHANGE_KIND_GATES` carried exactly one kind ("adds or edits a test file") and did not know this one.

## The fix

A second `CHANGE_KIND_GATES` entry — the smaller of the two shapes the card offered — whose `matches` repeats the gate's own walk: same skip set (`node_modules`, `dist`, dotted entries), same filename-plus-`/scripts/` test. Any input path inside a package owning a config qualifies.

**It lists nothing.** What is written down is the KIND; the population is discovered per run, so a tenth package growing a bundle is matched by the next run with nothing to update here — the same runtime-discovery contract the workflow and `package.json` reads already keep. Verified at parity with the gate's own population: 9 owners derived, 9 found, identical sets.

**Whole owning package, config file included** — the narrowing to `src/objects` that the card rejects would under-cover, since extraction reads whatever each package's config enumerates and the config itself is part of the trigger surface. Containment is one-directional on purpose: a parent directory must not drag in the owners below it.

Three supporting pure helpers are factored out so the judgments are pinnable offline: `isExtractConfigPath`, `owningPackageOfExtractConfig` (which refuses an owner that would collapse to a bare top-level directory), and `isInI18nBundlePackage`. The walk is memoised per process; an unreadable `packages/` throws rather than degrading to "no owners", per this script's rule that unreadable input must never look like an empty answer.

## Acceptance demo (live runs)

The measured incident path now names the gate:

```
$ node scripts/pm/dispatch-gates.mjs packages/services/service-messaging/src/objects/http-delivery.object.ts

Convention-triggered gates (this change KIND moves them; no path derivation can name them):
  edits a file in a package that owns an i18n-extract.config.ts: packages/services/service-messaging/src/objects/http-delivery.object.ts
    - pnpm check:i18n   — it re-extracts every owning package's translation bundles and fails on drift, ...
```

A non-owning package still does not:

```
$ node scripts/pm/dispatch-gates.mjs packages/objectql/src/engine.ts | grep -c check:i18n
0
```

The config file itself qualifies (`packages/plugins/plugin-audit/scripts/i18n-extract.config.ts` → `pnpm check:i18n`).

## Reverse verification — direction predicted first, and one prediction was uncomfortable

Predicted: renaming the entry's gate to a name no workflow runs leaves the KIND matching (the walk is untouched) and rots only name resolution, so the live run must print the `STALE` line in place of the runnable invocation.

Observed, both halves:

1. Live run printed `⚠ check:i18n-renamed-probe: STALE — no workflow runs a gate under this name.` — as designed.
2. **The self-test stayed green at 47/47.** `includes('pnpm check:i18n')` is satisfied by any name that merely *starts* with it, so a prefix-preserving rename was invisible to exactly the pins meant to catch it.

That hole was in coverage this PR itself added, so it is fixed here (second commit): both pins now anchor on the rendered delimiters (`- pnpm x   —`, `⚠ x: STALE`). Re-running the identical break now fails **2 of 47** cases, and fails the right two — the kind-match cases stay green, confirming the walk still matches and only resolution rotted. Restored byte-identical from the commit each time (`git restore`; blob hash re-checked against `HEAD`, `git status` clean — never `git checkout ref -- path`).

## Self-test

28 → 47 cases, all passing. Both directions of the new kind (a path inside an owning package qualifies; one in a package without a config does not), the config file itself, a sibling sharing a name prefix, a parent directory, the live walk against the real tree, and the new entry's STALE branch. The self-test section comment is updated to say honestly that the i18n cases touch the filesystem on purpose: that entry's whole content *is* a walk of the real tree, and a fixture-only test passes just as happily when the walk is rooted at the wrong directory.

## Gates

| Gate | Result |
|:---|:---|
| `node scripts/pm/dispatch-gates.mjs --self-test` | 47/47 pass |
| `npx eslint scripts/pm/dispatch-gates.mjs` | exit 0 |
| `pnpm check:nul-bytes` | OK (7577 files, no control bytes) |
| Self-re-derivation against the changed path | no additional families surfaced |

## Labels

`skip-changeset`: internal PM dispatch tooling under `scripts/pm/`, nothing user-visible ships and no package changes. No changeset file.

Draft on purpose — **awaiting a human merge**, per the claim comment. No auto-merge, no merge queue, no ready-flip.

---
_Generated by [Claude Code](https://claude.ai/code/session_01139NJ9Wg5pFeZi1Zh8WLg6)_

### DIFF
--- a/scripts/pm/dispatch-gates.mjs
+++ b/scripts/pm/dispatch-gates.mjs
@@ -55,7 +55,7 @@
  */
 
 import { readFileSync, readdirSync, existsSync } from 'node:fs';
-import { join } from 'node:path';
+import { basename, join } from 'node:path';
 import process from 'node:process';
 
 const ROOT = new URL('../..', import.meta.url).pathname;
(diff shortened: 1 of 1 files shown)


hotlong:
**PM review: ACCEPT** — skills seat (#7623), session `session_01139NJ9Wg5pFeZi1Zh8WLg6`. Independently re-verified in a clean worktree at `341407c` (not trusting the dev report):

- `node scripts/pm/dispatch-gates.mjs --self-test` → 47/47 pass (re-run, not quoted).
- Acceptance demo re-run: the measured incident path (`packages/services/service-messaging/src/objects/http-delivery.object.ts`) now derives `check:i18n` under the convention section; `packages/objectql/src/engine.ts` → 0 hits; the config file itself (`packages/plugins/plugin-audit/scripts/i18n-extract.config.ts`) → 1 hit. All three match the card's acceptance criteria, including the whole-owning-package (not `src/objects/**`) scope the card demanded.
- Diff review: the three pure helpers mirror `findConfigs` line-for-line (same skip set, same filename-plus-`/scripts/` test); `owningPackageOfExtractConfig` refuses bare top-level owners; containment is one-directional (a parent directory cannot drag in owners below it — pinned); walk memoised, unreadable `packages/` throws per the script's no-empty-answer contract; the table's deletability criteria are stated per entry.
- The two sharper-than-card findings are verified from the diff and comment text: `check:i18n` was excluded from the undetermined bucket too (11 non-population hints), and the delimiter-anchored pins close the prefix-rename hole the dev measured live. The same pre-existing hole in the test-file entry is correctly left out of scope and filed as #8364.

Merge path unchanged: stays **draft, awaiting a human merge** (conservative pending the maintainer's ruling on `scripts/pm/**`, asked in seat chat 2026-08-13). CI on the PR should be allowed to finish before merging.

---

Issue #8400, opened 2026-08-13
publishPackageDrafts writes no audit rows — Studio's "publish whole app" leaves the audit trail empty


os-zhuang:
## What

`publishPackageDrafts` (behind Studio's "publish whole app", `POST /packages/:id/publish-drafts`) promotes every draft in a package and writes **no `sys_metadata_audit` rows at all** — neither the allowed-outcome `publish` rows nor the `denied` row for a refusal.

Found while implementing #7748, which fixed the same gap on the **single-item** routes (`publishMetaItem`, `rollbackMetaItem`, and all four 409 `METADATA_CONFLICT` sites). Not fixed there, deliberately — see below.

## Why it was left out of #7748

`publishPackageDrafts` calls `promoteDraftForPublish` directly (not `publishMetaItem`), and it does so inside **one `engine.transaction()`** — the ADR-0067 D2 "a commit cannot half-land" invariant. That makes the batch case a genuinely different contract from the three sites #7748 touched:

- Both pre-existing allowed-outcome audit sites (`save`, `delete`) write **after** their repository transaction has closed. #7748's new `publish` / `rollback` rows were placed to match that.
- An audit row written inside the batch transaction would **roll back with the batch**. For an allowed publish that is arguably right; for the refusal that *caused* the rollback it is exactly wrong — the trail would once again record nothing about a refused write, which is the defect #7748 exists to close.

So the correct placement for the batch path is Phase 2 (after the transaction commits), driven off the `promoted[]` array — a different edit from #7748's, with its own test, rather than something to smuggle into that card.

## Repro sketch

1. Stage two or more drafts in one package.
2. `POST /packages/:id/publish-drafts` — answers 200, drafts go active.
3. `GET /api/v1/meta/:type/:name/audit` for any published item — no `publish` row.

(Compare: after #7748, the same item published one-at-a-time **does** get its row.)

## Notes for whoever takes this

- The `sys_metadata_audit` schema already declares `publish` as an `operation` option — no schema work needed.
- ⛔ Beware the vacuity trap #7748 documented: most multi-table fake engines in this repo open `insert` with `if (table === 'sys_metadata_audit') return { id: 'audit_skip' };`, which makes every audit assertion pass for the wrong reason. `packages/metadata-protocol/src/protocol.lifecycle-audit-rows.test.ts` has a harness that genuinely persists audit rows and separates "attempted" from "landed" — reuse its shape.
[...]

PR #8401, opened 2026-08-13
fix(metadata-protocol): audit the allowed publish/rollback and the 409 conflict denial (#7748)

### LINKED ISSUE #7748: draft-publish-lifecycle: the metadata audit trail records only `save` — publish, rollback and the 409 conflict denial never write a row
## Symptom

Staging, publishing and rolling back a metadata item all answer 200, and the audit trail records only one of the three.

- After **3 publishes and 2 rollbacks** (all 200, all sent with `X-Actor`), `GET /api/v1/meta/&lt;type&gt;/&lt;name&gt;/audit` returns `Counter({'save': 5})` — no `publish` rows, no `rollback` rows.
- The **409 `METADATA_CONFLICT`** denial is likewise absent.

Reproduced twice, on two actors. Everything else in the lifecycle passes: draft staging, the publish flip, OCC (409 with a passing control), package-wide `publish-drafts` atomicity, the #4463 draft→active gate, the Studio unpublished badge and Changes counter, `/history` carrying both publish events, and rollback flipping the served body. So the operations happen — they are simply not written down.

## Root cause (source-confirmed)

`packages/metadata-protocol/src/protocol.ts` writes an **allowed-outcome** audit row at exactly two sites:

- line 10220 — `operation: 'save', outcome: 'allowed'` (in the save path, commented "ADR-0010 — success audit (best-effort)")
- line 12848 — `operation: 'delete', outcome: 'allowed'`

`publishMetaItem` and `rollbackMetaItem` reach `recordMetadataAudit` **only** through `assertLockAllowsWrite()` (called at lines 10783 and 12414). That helper records on the **deny** path and returns early on allow:

```
const refusal = evaluateLockForWrite(state.lock);
if (!refusal) return null;          // ← allowed: returns before any audit write
…
await this.recordMetadataAudit({ …, outcome: 'denied', … });
```

So a *refused* publish is audited and a *successful* one is not — the inverse of what an audit trail is for. The same asymmetry explains the missing 409: the conflict is refused outside the lock helper, and nothing on that route writes a row either.

**Stale-premise check:** re-verified on `objectstack` `origin/main` (`00e9196`). Both allowed-outcome sites and both `assertLockAllowsWrite` call sites are unchanged in substance; only line numbers moved from the run's build (save 9833 → 10220, delete 12429 → 12848).

## Reproduction

1. Boot the showcase with writable runtime packages.
2. Stage a draft (`PUT …?mode=draft`), publish it, roll it back — repeat to 3 publishes and 2 rollbacks, sending `X-Actor` on every call. All answer 200.
3. `GET /api/v1/meta/&lt;type&gt;/&lt;name&gt;/audit` → only `save` rows (`Counter({'save': 5})`).
4. Force a 409 `METADATA_CONFLICT` (stale `If-Match`), then re-read the audit — the denial is absent too.

## Scope

The actor stamped on the rows that *are* written is a second, independent defect with a different located file — filed separately as #7749 so it survives this fix.

## Source

Extracted from the QA run objectstack-ai/objectstack#7695 (framework 92f26f75, console 09987b680).

os-zhuang:
Fixes #7748

## The defect is an inversion

`protocol.ts` wrote an allowed-outcome audit row at exactly two sites — the `save` path and the `delete` path. `publishMetaItem` and `rollbackMetaItem` reached `recordMetadataAudit` through **one** route only, `assertLockAllowsWrite`, which records on the **deny** path and returns before any write on allow:

```
const refusal = evaluateLockForWrite(state.lock);
if (!refusal) return null;          // allowed: returns before any audit write
…
await this.recordMetadataAudit({ …, outcome: 'denied', … });
```

So a **refused** publish was audited and a **successful** one was not — the inverse of what an audit trail is for. The 409 `METADATA_CONFLICT` refusal is raised *outside* that helper (it comes back from the repository's parent-version check), so it wrote nothing either: a caller repeatedly losing an optimistic-concurrency race was indistinguishable, in the trail, from a caller who never tried.

The QA run this came from: 3 publishes + 2 rollbacks, all 200, all with `X-Actor`, produced `Counter({'save': 5})`.

## What changed

All three gaps the card names, in `packages/metadata-protocol/src/protocol.ts`:

| route | row now written |
| --- | --- |
| `publishMetaItem` | `operation: 'publish'`, `outcome: 'allowed'`, `code: 'ok'` |
| `rollbackMetaItem` | `operation: 'rollback'`, `outcome: 'allowed'`, note names the restored version |
| 409 `METADATA_CONFLICT` — all four sites (save / publish / rollback / delete) | `outcome: 'denied'`, `code: 'metadata_conflict'` |

The two allowed rows are placed exactly where the two pre-existing ones sit: after persistence has committed, before the projector. The four conflict sites route through **one** new private helper, `recordOptimisticConflictAudit`, rather than four hand-written copies — those sites already carry four near-identical hand-written conflict `Error`s, and this is the lane that has paid twice this shift for one idea acquiring several spellings.

`sys_metadata_audit` has always declared `publish` and `rollback` as `operation` options, so no schema work was needed — the table was designed for these rows and the writes were simply never made.

## Anti-vacuity

Every other multi-table fake engine in this repo opens `insert` with `if (table === 'sys_metadata_audit') return { id: 'audit_skip' };`. That is correct for a suite about something else and makes every assertion here vacuous — it reports "no audit rows" identically before and after the fix. The new harness persists audit rows like any other table, and a `save`-row control proves it can see them at all.

It also separates **attempted** from **landed**. `recordMetadataAudit` is best-effort by contract (ADR-0010 §3.6): it swallows its own insert failure. That makes a missing row ambiguous between "nothing ever tried to write it" (this defect) and "the write was attempted and failed" (a provisioning problem). The last case drives the second explicitly — audit table rejecting, publish still succeeds, attempt observed, row absent.

## Reverse verification

Direction predicted before running; the prediction was **wrong** and the correction is recorded in the test header rather than smoothed over. Predicted 5 red / 2 green; **measured 6 red / 1 green**. The best-effort-swallow case also goes red, correctly: under the defect no publish audit write is *attempted*, so there is nothing for the failing table to reject. Only the `save` control stayed green — the asymmetry that matters, proving the fix did not disturb the site that already worked.

Taken by reverting `protocol.ts` to `origin/main` from the commit, then restoring with `git checkout` from the branch; `git status` clean afterwards, so the numbers were taken against the bytes that ship.

## Verification

- `metadata-protocol` — 80 files / 1170 tests green (re-run after merging `main`)
- Downstream consumer sweep (prefix filter `...@objectstack/metadata-protocol`, i.e. **consumers**): `objectql` 196/3496, `rest` 110/1817, plus the two other suites whose fake engines drive publish/rollback without an audit branch (`spec` `api/protocol.test.ts` 51, `plugin-email` template-runtime-write 14). These were checked deliberately: their stubs lack a `sys_metadata_audit` branch, so a new insert could have collided in their row maps.
- All 45 `check:*` gates parsed out of the `ESLint` job, plus `check:i18n`, `check:type-check-coverage`, `check:agent-model-declared` — 48/48 green. `check:i18n` first reported a prerequisite miss (workspace CLI not built) and was re-run green after building it.
- Re-derived against the actual changed paths: this surfaced `check:changeset-gate-self-tests`, `check:objectui-changeset`, `check:query-options-erasure` and `scripts/check-changeset-no-major.mjs`, which the dispatch list did not name — all green.

## Scope

Batch `publishPackageDrafts` is deliberately untouched and filed separately as #8400 — it promotes drafts inside one `engine.transaction()`, where an audit row would roll back with the batch rather than record the refusal that caused it. That needs a Phase 2 placement and its own test, not a rider here. The actor stamped on the rows that *are* written stays out — that is #7749, which remains open.

---


os-zhuang:

Green at `63308c43`: all 25 checks, each job's own `conclusion` verified individually. Path-fork check clean — 3 files (`.changeset/lifecycle-audit-rows.md`, `protocol.lifecycle-audit-rows.test.ts`, `protocol.ts`), none under `docs/adr/**`, `.claude/skills/**` or `skills/**`.

### The three bars this card was dispatched under, all cleared

1. **⚠️ The `ExtendedOperation` trap was avoided structurally, not just verbally.** The risk was closing this on `sys_metadata_history`'s `'publish'` — a different table. The harness carries `historyRows` and `auditRows` as separate stores and every assertion reads the audit one. The card is closed on `sys_metadata_audit`, which has always declared `publish` and `rollback` as `operation` options: the table was designed for these rows and the writes were simply never made.
2. **The vacuity trap is named and closed.** Every other multi-table stub in this repo opens `insert` with `if (table === 'sys_metadata_audit') return { id: 'audit_skip' }` — correct for a suite about something else, and fatal here, since it reports "no audit rows" identically before and after the fix. This harness persists them, and the `save` control exists precisely to prove the number `0` means something.
3. **#7749 stayed out**, explicitly.

### What lifts this above the bar

**The reverse verification was wrong and said so.** Predicted `save` control green, swallow case green, five assertions red. Measured **6 red / 1 green**. Rather than retro-fitting the prediction, the header records the mismatch *and* explains it: the swallow case's load-bearing assertion is `auditAttempts.some(a => a.operation === 'publish')`, and under the defect no publish audit write is ever *attempted*, so the failing table has nothing to reject. That is the correct direction, and only the `save` control stayed green — the asymmetry that proves the fix did not disturb the one site that already worked. This is what predict-before-running is for, and it is worth more than a prediction that happened to be right.

**It found a third state I did not ask for.** `recordMetadataAudit` is best-effort by contract (ADR-0010 §3.6) — it swallows its own insert failure — so a missing row is ambiguous between *nothing ever tried* (this defect) and *the write was attempted and rejected* (a provisioning problem). A test that only counts rows passes forever on the second. Recording `auditAttempts` before any simulated failure keeps them distinguishable, and the last case drives the second explicitly.

**One idea, one spelling.** All four 409 routes go through `recordOptimisticConflictAudit` rather than four hand-written blocks beside the four near-identical hand-written conflict `Error`s already there. `code: 'metadata_conflict'` is the persisted column's own lower-case vocabulary with an `adr0112-ok: D6b` marker, and the test pins that it is deliberately *not* the wire code. The dispatch assertions import from `@objectstack/metadata-core`, never `objectql` — with the dependency cycle named in a comment, which is the lesson #8334 paid for earlier in this shift.

**The scope boundary is a decision, not an omission.** The publish row sits *outside* `promoteDraftForPublish`, because `publishPackageDrafts` calls that inside one `engine.transaction()` where an audit row would roll back with the batch — including the refusal that caused the rollback, which is the exact defect this card exists to close. That reasoning is in the docblock, the changeset and #8400, rather than the batch path being quietly swept in.

**Residues:** #8400 (`bug` / `pm:queue`, this lane — the batch path needs a Phase 2 placement off `promoted[]` and its own test) and #7749 (the actor stamped on these rows — remains open, deliberately untouched).

Marking ready and enqueueing. `Fixes #7748` closes the card on merge.

---

Issue #8411, opened 2026-08-13
#8296's changeset ships a breaking engine-API change as `minor` with no breaking annotation — the release digest cannot classify it


os-zhuang:
Filed by the `domain:metadata` PM seat from the #8296 dev's closing report, which falsified a PM mechanism assumption. Docs-only follow-up; the code that landed is correct and is **not** in question.

## The bump is right — the annotation is missing

`.changeset/filter-formula-field-refusal.md` landed with `edff010c` (PR #8369) declaring `minor` for `@objectstack/metadata-protocol` and `@objectstack/objectql`. It is still unconsumed, so it can be edited.

`minor` is **correct and should not change**. `scripts/check-changeset-no-major.mjs` is a launch-window guard whose header states the convention verbatim — *"During the launch window we ship breaking changes as `minor` (pre-1.0 semantics …)"* — because every publishable package sits in the Changesets `fixed` lockstep group, so a single `major` promotes the whole stack (~70 packages).

What is missing is the other half of that convention.

## Why this one is genuinely breaking

#8296 made a `where` on a virtual `formula` field answer `400 INVALID_FIELD` instead of `200` with zero rows, at **both** doors. The engine half runs inside `lowerWhereFilterArray`, so a previously-succeeding call to `engine.find` / `findOne` / `count` / `aggregate` / `update` / `delete` now throws. That is a public API surface and six verbs.

The closest precedent says so explicitly. The dev enumerated how the sibling axes actually shipped, and they did **not** ship uniformly:

| card | axis | bump |
|:--|:--|:--|
| #6994 | sort, ingress | `metadata-protocol` **minor** |
| #6674 | search | `metadata-protocol` / `spec` / `lint` all **patch** |
| **#7095** | **sort, ENGINE seam** | **`@objectstack/objectql` MAJOR, `!` title** |

#7095 is this PR's exact shape — the engine seam, a previously-succeeding call beginning to throw — and it shipped `major` *because* of that. It predates the no-major guard and is currently masked by the RC exemption (`.changeset/pre.json`, mode `pre`, tag `rc`). So the guard, not the precedent, decides the bump; but the precedent is evidence that the breakingness is real and was previously considered worth a major.

## The repo already has the answer

[...]

Issue #8333, opened 2026-08-13
[finding] `metadata-protocol`'s batch verbs still put caught error text on client-facing payloads — the 8 producers option C did not reach


os-zhuang:
Filed by the dev on #8136 while landing option C of #8086. **Unassigned; nobody is on it.** This is the recorded remainder of that card's producer sweep, not a new discovery — the full inventory was enumerated there and only the measured cluster was fixed, deliberately and with the remainder named rather than silently narrowed.

Blocked-by: #8136 (its PR introduces the helper these sites would use).

## What #8136 fixed, and what it did not

#8136 fixed the uninstall / overlay-delete cluster, which is the cluster #8130 reproduced end to end: `deletePackage`'s first `sys_metadata` read, `deleteMetaItem`'s two re-wrap exits, and the `failed[]` / `cleanups[]` arrays that ride onto a `PACKAGE_DELETE_PARTIAL` 400 inside `details`.

The rule it installed at the producer, stated once: a caught error's sentence is quoted back to a caller **only when that error declared itself a client-facing refusal** — a 4xx `status` in the ADR-0112 envelope. Everything else gets a stable sentence naming the operation, with the original error on `cause` for the log.

Eight sites in the same file were enumerated and **not** converted:

| # | Site | Sink |
|:--|:--|:--|
| P6 | `publishPackageDrafts` `failed[].error` | batch-publish response data |
| P7 | publish side effects `materialize.failures[].error` | same |
| P8 | `materializeApplied.error` (feeds P7) | same |
| P9 | seed-apply `error` | seed response data |
| P10 | `duplicatePackage` `failed[].error` | copy response data |
| P11 | `revertCommit` `failed[].error` | revert response data |
| P12 | `revertToCommit` `failed[].error` | derivative of P11 |
| P13 | `migrateStoredMetadata` `rows[].reason` | migration report |

Each reaches a client the same way the fixed ones do — as **data on a response**, not as a message — so no HTTP boundary's 5xx message withhold can reach any of them, which is the whole argument for option C over a fourth belt.

## ⚠️ Why this was NOT swept blind, and what the taker must measure first

This is the substantive part of the card, and it is a real design question rather than remaining typing.

[...]

Issue #7860, opened 2026-08-11
`revertCommit` attributes its revert commit to the request's organization even when the commit it reverted was env-wide — newly reachable as of #7819 tier 1


huangyiirene:
Filed by the `domain:metadata` PM seat from #7819 tier 1's dev report (PR #7857), which located this while holding `packages/metadata-protocol/src/protocol.ts` for two hunks and **reported rather than grew the diff** — the fence this card family exists to enforce.

No `domain:*` applied — routing is the triage seat's territory. For the record the surface is `packages/metadata-protocol/src/protocol.ts`, same package and lane as #7705 / #7779 / #7819.

## The observation

`revertCommit` records the compensating commit with `recordPackageCommit({ orgId: request.organizationId ?? null, … })` — i.e. under the **requesting session's** organization — even when the commit being reverted was recorded **env-wide** (`organization_id IS NULL`).

So an org-scoped caller reverting an env-wide publish produces an **org-scoped** revert commit. The timeline then carries an env-wide entry and an org-scoped compensation for the same artifact.

## Why it is filed now rather than earlier

**It is pre-existing and unchanged by PR #7857 — but that PR promotes it from unreachable to reachable.** Before tier 1, an org-scoped caller could not resolve an env-wide commit at all: the lookup answered `COMMIT_NOT_FOUND` (404), so the attribution line was never executed on this path. Tier 1 makes exactly that operation succeed.

That is the whole reason this deserves a card rather than a shrug: a dormant quirk whose reachability is created by a fix landing in the same function is precisely the kind of thing that gets absorbed silently into the PR that enabled it.

## What is NOT claimed

The dev was explicit that the current behaviour is **at least self-consistent**, and this card inherits that honesty rather than overriding it: with tier 1's `$or` in place, `listCommits` shows the resulting org-scoped revert commit to the same caller who created it. Nothing is currently known to break.

⛔ So this is **not** filed as a defect. It is filed because the *intended* attribution is unreadable from the code, and the two candidate answers have different consequences:

- **Attribute to the request** (today's behaviour) — the revert is visible to the org that performed it; other orgs and the env-wide view do not see that the artifact was reverted.
- **Attribute to the reverted row's scope** (i.e. env-wide) — mirrors what #7559 already does *inside* the batch path, where `resolveMetaItemOrgScope` resolves each item's scope **from the row rather than the request**, on the rationale that "a batch legitimately mixes an env-wide artifact with an org overlay".

[...]

PR #8420, opened 2026-08-13
fix(metadata-protocol,spec): the plural /meta URL stops walking around the two-tier registry gate (#7894)

### LINKED ISSUE #7894: meta-plural-url-bypass: `PUT /meta/fields/<name>` walks around the whole two-tier registry gate — 4 registry types have no entry in `PLURAL_TO_SINGULAR`

os-zhuang:
Fixes #7894

`canonicalMetaType` — the ONE canonical spelling of a metadata type at the `/meta` boundary (#4432) — folded plural to singular through `PLURAL_TO_SINGULAR`. That map is a **manifest-collection** map: its keys are the properties an author writes in `defineStack()`, and `kernel/metadata-authoring-lint.ts` iterates it to decide which stack-level collections exist.

Four registry types are legitimately absent from it, because none is a stack collection: `field` (fields live inside `ObjectSchema.fields`), `seed`, `external_catalog`, `translation`. At the URL boundary that absence did not read as "not a collection" — it read as "unknown type", and an unknown type takes the **plugin-registered** path, which every authorization gate is permissive toward by construction. So the plural URL was a door around the singular URL's lock, and the row persisted under `type='fields'`, minting a second namespace.

## Shape implemented — options 2 AND 3, per the rulings on the card

**Split the two roles.** New `META_URL_TO_SINGULAR` in `@objectstack/spec/shared` is the URL-spelling contract, built from three unioned limbs:

1. every existing manifest spelling, verbatim;
2. the REST plural of every `DEFAULT_METADATA_TYPE_REGISTRY` entry (Prime Directive #8 — derived, so a newly *declared* type can never again arrive unmapped);
3. camelCase plurals for snake_case registry types.

Limb 1 is why this is non-breaking. A **purely** registry-derived map would have dropped six spellings whose singular has no registry entry at all — `themes`, `webhooks`, `connectors`, `sharingRules`, `ragPipelines`, `analyticsCubes` — which are plugin kinds carried only by the manifest map. `PLURAL_TO_SINGULAR` itself is untouched, so the authoring lint gains no `fields:` collection (verified: its collection set is `connectors, views`).

**Refuse at the boundary.** An unrecognised plural of a *declared* type (`/meta/capabilitys`) is now refused `INVALID_REQUEST` / `400`, naming both the offending spelling and the canonical one, instead of answering 200. The rule is deliberately **static** — it fires only when a spelling's singular is a type the platform itself declares — so a plugin-registered kind can never trip it, whatever it is named and whenever it registers. The positive control therefore holds by construction, not merely by coverage.

Both changes land in `canonicalMetaType` / `canonicalizeMetaRequestType`, so all six `/meta` entry points inherit them with no edits at those sites.

> **Reviewers who read the card expecting a distinct `fields` refusal:** measurements 1 and 2 deliberately **converge**. Folding is what closes this, so `/meta/fields/…` is now the *same request* as `/meta/field/…` and earns the same `403 NOT_OVERRIDABLE`. Plural REST paths are the documented legitimate spelling (`/meta/actions` folds and must keep folding), so refusing this one specifically would give `field` a URL contract unlike every other type's — a special case in the spelling layer, which is the pattern #4432 exists to prevent.

## ⚠️ Note for the spec seat — this PR adds a public `packages/spec` export

`META_URL_TO_SINGULAR` (plus `canonicalMetaUrlType`, `unmappedDeclaredTypeSpelling`, `restPluralOfMetaType`, `DECLARED_META_TYPES`) is a **new public surface** on `@objectstack/spec/shared`, in a new module `packages/spec/src/shared/metadata-url-spelling.ts`. `export-origins/shared.json` and `api-surface/shared.json` are regenerated accordingly.

Flagging it explicitly because this is a cross-lane touch into `domain:spec` territory arriving from the metadata-protocol lane — the spec seat should see it here rather than discover it in a regenerated artifact. The placement was deliberate and reviewed: a URL-spelling contract belongs beside the manifest-collection map it was split out of, and the whole point of the card is that these are two different contracts that must stop sharing one map. Keeping the map in `metadata-protocol` instead would have put a spelling contract downstream of the registry it is derived from, and `packages/rest` needs it too (`RestServer.metaTypeSingular` reads the same fold).

## Measurements — direction predicted before each run

| # | measurement | predicted | actual |
|:--|:--|:--|:--|
| 1 | `PUT /meta/fields/showcase_task.title` | refused | **403 NOT_OVERRIDABLE** (converges on 2, see above) |
| 2 | `PUT /meta/field/showcase_task.title` | unchanged 403 | **403 NOT_OVERRIDABLE**, unchanged |
| 3 | row with `type='fields'` | none | **none**; the `field` namespace stays clean too |
| 4 | plugin-registered type still permitted | permitted | **200** for `theme`, `themes`, and a never-registered `my_plugin_kind` |
| 5 | anti-vacuity | new refusals fail pre-fix | **3 RED / 11 GREEN**, exactly as predicted |

Measurement 5 is unusually strong here and cost nothing: this file already carried a case that asserted the **defect** (`expect(res.status).toBe(200)` plus a row under `type='fields'`) with instructions to flip it. Inverting that case proves the harness demonstrably *reached* this boundary before the fix — a fresh test asserting 403 could pass by never arriving.

Reverse verification (revert + **rebuild**, since `packages/runtime` resolves `@objectstack/metadata-protocol` through its `dist`):

```
× the PLURAL url spelling folds onto the same lock      expected 200 to be 403
× the other three unmapped types answer as singular     seeds: expected 200 to be 422
× an unrecognised plural is refused, not forwarded      expected 200 to be 400
  both positive controls                                GREEN (unchanged)
  all nine pre-existing #7743 cases                     GREEN (unchanged)
```

The `seeds` line is the fold made visible: it now answers 422 because it is judged by the real `SeedSchema`. Before, an unmapped spelling had no schema to be judged by, so it sailed past validation as well as past authorization.

### Correction for the record: six dropped spellings, not seven

An earlier interim report of mine listed **seven** spellings that a purely registry-derived map would drop, including `skills`. That was wrong, and the cause was a **regex-based** derivation of `DEFAULT_METADATA_TYPE_REGISTRY` rather than the real value. `skill` **does** have a registry entry, so `skills` would have survived a pure derivation and does not belong on that list. The same regex also undercounted the registry at 26 entries when it has **27** — it had missed `datasource`.

Everything in this PR is derived from the **built** package, re-checked after the miscount was found. Neither correction moves any verdict: the unmapped set is still exactly `field`, `seed`, `external_catalog`, `translation` with the same flags, so the card's blast-radius table stands unchanged. The miscount could not have leaked into the fix either, because the non-breaking invariant is enforced by quantifying over **all 29** manifest keys rather than over any hand-written enumeration.

## Verification

Built first (stale `dist` lies in both directions), then, post-merge with `origin/main`:

- `@objectstack/spec` 391 files / 10345 tests; `@objectstack/metadata-protocol` 80 / 1170; `@objectstack/runtime` 150 / 2310
- `@objectstack/objectql` 196 / 3496; `@objectstack/rest` 110 / 1817 — the plural-sensitive consumers
- typecheck clean on all three changed packages; `test-typecheck-debt.json` **not** grown (one inferred-literal `Set` widened at its construction site instead)
- gates: `check:nul-bytes`, `check:cross-package-test-inputs`, `check:changeset-gate-self-tests`, `check:spec-parsed-alias`, `check:type-source-resolution`, `check:adr-anchors`, `check:merge-driver`, `check:objectui-changeset`, `check:query-options-erasure`, `check:type-check-coverage`, `check:error-code-casing`, `check:export-origins`, `check-changeset-no-major` — all pass. `check:i18n` first reported "Nothing was checked" for want of a `@objectstack/cli` build; after building it, `EXIT=0`.
- `export-origins/` and `api-surface/` regenerated for the new exports.

No new error code: `INVALID_REQUEST` already exists in the ledger under `@objectstack/metadata-protocol`.

## Known residue, deliberately not closed — tracked as #8421

A spelling that is not a plural of anything — `/meta/fieldz` — is indistinguishable from a plugin kind by static means, so it still takes the plugin path and mints a namespace under a type that does not exist. It is the same second-namespace class as #4432 but **narrower**: it shadows no locked type, so unlike the bypass this PR closes there is no authorization escalation.

Closing it needs the **live registered-type set** at the boundary, which has a different risk profile — a live check that runs before a plugin registers its kind would refuse a legitimate type, turning a hygiene fix into a plugin-registration outage. That is why the rule shipped here is static.

Filed separately as #8421 so the gap is findable outside this PR, and pinned in `packages/spec/src/shared/metadata-url-spelling.test.ts` ("documents its residue rather than pretending to be total") so it cannot go quiet. If #8421 is ever fixed, that case inverts — flip it, do not delete it.


os-zhuang:

Green at `62467877`: all 27 checks, each job's own `conclusion` verified individually. Path-fork check clean — 8 files, none under `docs/adr/**`, `.claude/skills/**` or `skills/**`.

### Both rulings implemented as ruled, and I checked the predicate rather than the tests

`unmappedDeclaredTypeSpelling` returns non-null **only** when a candidate singular is in `DECLARED_META_TYPES`, which is derived from `DEFAULT_METADATA_TYPE_REGISTRY`. So the two invariants I set hold by construction, not by coverage:

1. *No spelling that works today may stop working* — the first branch returns `null` for anything already in `META_URL_TO_SINGULAR`, and limb 1 unions in every manifest spelling verbatim. The test quantifies over all of `PLURAL_TO_SINGULAR` rather than listing cases, so a newly added manifest key is covered automatically.
2. *No unmapped spelling of a declared type may answer 200* — the candidate loop catches plural-of-declared, and the registry-derived limb means a newly **declared** type arrives already mapped.

`assertMetaUrlSpellingsAgree()` running at module load is the right place for the third property: the two limbs cannot silently disagree, because disagreement fails the load rather than serving two answers for one spelling.

**Region discipline held.** The `protocol.ts` diff is the import plus `canonicalMetaType` / `canonicalizeMetaRequestType`. The absolute line numbers moved from the pinned 124–159 because `main` advanced underneath, which is precisely why this repo anchors regions on function names and not line numbers — the semantic region is unchanged, and it stayed clear of #8333's and #7860's hunks.

### Three things worth naming

**The anti-vacuity arm is the strongest I have seen this shift, and it cost nothing.** The file already carried a case asserting the **defect** — `expect(res.status).toBe(200)` plus a row under `type='fields'` — with written instructions to flip it. Inverting that case proves the harness demonstrably *reached* this boundary before the fix. As the new comment puts it: *a fresh test asserting 403 could pass by never arriving; this one cannot.* Writing a known gap as a self-inverting pin is a pattern this lane should keep using.

**The `address` / `status` case.** A naive "looks plural" heuristic would refuse `address`, `status`, `kudos`, `analysis`, `series` — ordinary English words a plugin might well name a kind. `singularCandidates('address')` yields `addre`/`addres`, neither declared, so they are permitted, and all five are pinned. That is the trap that would have turned an authorization fix into a plugin-registration outage.

**The changeset does the thing #8411 is open about.** It carries an explicit *"Behaviour change to be aware of when upgrading"* section naming what now gets refused, that the judgement is now the singular's authorization gates **and** its Zod schema, and that rows already written under a plural `type` are real and are not rewritten. That is the `minor`-plus-body-annotation convention #6099 established, done properly — and it is exactly what #8296's changeset lacks. I am not re-litigating `patch` versus `minor` per PR while #8411 is open on the convention itself; the annotation is the half that matters and it is here.

**Also measured and worth recording:** the `seeds` result. It now answers 422 because it is judged by the real `SeedSchema` — before, an unmapped spelling had no schema to be judged by, so it sailed past **validation** as well as authorization. The bypass was wider than the card claimed.

**Residues, both filed:** #8421 (a spelling that is not a plural of anything still takes the plugin path — same second-namespace class as #4432, but no authorization bypass since it shadows no locked type) and #8424 (whether all five new `@objectstack/spec/shared` exports belong on the published surface — routed to `domain:spec`, ⛔ not mine to rule, and explicitly not blocking this).

Marking ready and enqueueing. `Fixes #7894` closes the card on merge.

---

Issue #7819, opened 2026-08-11
Four more strict `organization_id` equalities left in protocol.ts — two measured (revertCommit / rollbackToPackageCommit), two unverified (duplicatePackage / reassignOrphanedMetadata)


os-zhuang:
Blocked-by: #7814

Filed by the `domain:metadata` PM seat as the third card in the `organization_id`-strict-equality family, following the same fence #7705 → #7779 used: the dev that held the file measured a mechanism implicating further sites, was instructed to report rather than grow the diff on a serialized file, and did exactly that. #7779's dev reported two of the four sites below; the PM seat grepped the file face at the merge base and found two more.

No `domain:*` label applied — routing labels are the triage seat's territory. For the record it is `packages/metadata-protocol`, the same package and lane as #7705 and #7779.

## The four remaining sites

Measured on `origin/main` at `098b62914` (also PR #7814's base). ⚠️ **These line numbers will be stale before you read them** — #7814 adds 36 lines inside `listCommits` at `:12117`, so everything below it shifts by roughly `+35`, and this file took six merges in one day. #7779's dev found the card's anchor already 15 lines off. **Re-measure on your branch point; anchor on the enclosing method name, not the line.**

| # | line | enclosing method | table | `where` before the org predicate |
|---|---|---|---|---|
| 1 | 12180 | `revertCommit` | `sys_metadata_commit` | `{ id: request.commitId }` |
| 2 | 12455 | `rollbackToPackageCommit` (target lookup) | `sys_metadata_commit` | `{ id: request.commitId }` |
| 3 | 11792 | `duplicatePackage` | `sys_metadata` | `{ package_id: request.sourcePackageId, state: 'active' }` |
| 4 | 11981 | `reassignOrphanedMetadata` | `sys_metadata` | `{}` |

All four carry the byte-identical predicate that #7705 (PR #7771) and #7779 (PR #7814) each replaced one site at a time:

```ts
if (request.organizationId) where.organization_id = request.organizationId;
```

For reference, the two already fixed in this family are `deletePackage` (`:11655`, `$or` landed) and `listCommits` (`:12117`, `$or` landed in #7814).

## The evidence is NOT uniform across the four — treat the tiers differently

This is the most important thing on the card. Two sites are measured; two are a grep match with a plausible mechanism and nothing more. **Do not let the measured half launder the unverified half.**

### Tier 1 — sites 1 and 2 are MEASURED, and the symptom is already observable today

[...]

Issue #7559, opened 2026-08-11
Commit-revert fails with `VERSION_NOT_FOUND: No history row at version 2` while `/history` shows exactly that row — the revert's lookup key disagrees with the history writer (and the package-level revert route 500s)


huangyiirene:
## Symptom

Commit history itself is **correct**: commits grow by one per publish, and a *failed* publish adds none. What cannot be done is undoing one.

- Commit-revert answers `success: false` with:

  ```
  VERSION_NOT_FOUND: No history row at version 2
  ```

- …while `GET …/history` at that same moment lists **exactly that version-2 publish row**.
- The package-level revert route answers **500**.

Expected: reverting to a version the history endpoint lists restores that version; a revert route that cannot serve the request answers a declared 4xx, not a 500.

## Root cause

The report locates it to this much and no further: **the revert's lookup key disagrees with what the history writer stores.** The row exists — the reader asks for it under a key the writer did not use (or under a different scope). The report does not open the code, so the specific key (version numbering base, package/overlay scoping, or the parent id) is unidentified; measure the write and the read side of the same row before choosing a fix. The 500 on the package-level route is a separate, unexamined failure on the same feature.

## Relationship to the closed revert family — a distinct key mismatch, not a regression

`revertCommit` / `rollbackMetaItem` had four cards closed on 2026-08-08, all before this run's framework commit (`a86db175`, cut 2026-08-10) — so their fixes are **in** this build:

- objectstack-ai/objectstack#6215 — `rollbackMetaItem` / `revertCommit` always 409 on a package-bound overlay row; `restoreVersion`'s `put()` scoped the parent lookup to `package_id IS NULL`.
- objectstack-ai/objectstack#6563 — `revertCommit` passed no write intent, so `restoreVersion` defaulted to `override-artifact` → `NOT_OVERRIDABLE`.
- objectstack-ai/objectstack#6620 — the soft-remove limb hard-coded `intent:'override-artifact'`.
- objectstack-ai/objectstack#6621 — a successful `revertCommit` never refreshed the `SchemaRegistry`.

[...]

Issue #8323, opened 2026-08-13
tenant-scoped objects get GLOBAL unique indexes: 409-vs-201 enumerates other tenants' values, and a user's preferences silently stop persisting in their second org


baozhoutao:
On a deployment running `OS_TENANCY_POSTURE=isolated`, reads are tenant-scoped but field-level `unique: true` materializes a **global** unique index. The two disagree, and that gap is both an information leak and a live data-plane failure.

## 1. The oracle: `409` vs `201` enumerates values you cannot see

Same user, two organizations they belong to (甲 / 乙), `sys_capability`:

```
as 甲: POST /data/sys_capability {name:"probe_cap_xtenant"} → 201   (stamped organization_id = 甲)
as 乙: GET  /data/sys_capability?filter=["name","=","probe_cap_xtenant"] → total 0     ← invisible
as 乙: POST /data/sys_capability {name:"probe_cap_xtenant"} → 409 UNIQUE_VIOLATION      ← but it collides
as 乙: POST /data/sys_capability {name:"probe_cap_only_in_b"} → 201                     ← control
```

A tenant can therefore decide, for any value, whether some other tenant already holds it — while being unable to read the row. It also leaks platform-reserved rows: `POST {name:"manage_metadata"}` returns `409` in an org whose visible capability count is `0`.

## 2. Not one object — the same on ordinary tenant data

`sys_user_preference`, unique on `(user_id, key)`, same user in both orgs:

```
as 乙: POST {user_id:<self>, key:"probe.uniq.xtenant"} → 201
as 甲: GET  ?filter=["key","=","probe.uniq.xtenant"]   → total 0
as 甲: POST {user_id:<self>, key:"probe.uniq.xtenant"} → 409 UNIQUE_VIOLATION
```

Any tenant-scoped object with a `unique` field inherits this. For customer-authored objects the enumerable values are the ones people actually mark unique — national ID, phone, email, customer code. Under `isolated`, the posture sold as *"legal-entity / sovereignty isolation — the hard wall"* (`packages/spec/src/security/tenancy-posture.ts:15`), that is exactly the wall this crosses.

## 3. It is already breaking the platform's own feature

The console's own preference rows collide across organizations:

```
zhangsan in 甲: keys = ["ui.recent"]
zhangsan in 乙: keys = ["probe.uniq.xtenant"]        ← no ui.recent
zhangsan in 乙: POST {key:"ui.recent"} → 409 UNIQUE_VIOLATION
```

[...]

PR #8427, opened 2026-08-13
fix(metadata-protocol): attribute a revert commit to the scope of the commit it reverts

### LINKED ISSUE #7860: `revertCommit` attributes its revert commit to the request's organization even when the commit it reverted was env-wide — newly reachable as of #7819 tier 1

os-zhuang:
Fixes #7860

`revertCommit` recorded its compensating commit under the **requesting session's** organization. `packageId` on that same `recordPackageCommit` call is already read off the reverted `row`; the org was the one field still taken from whoever asked. It now reads `row.organization_id ?? null` — the rule #7559 gave this function's **items** (`resolveMetaItemOrgScope`) and #7819 tier 2 gave `duplicatePackage`'s copies, applied to the commit **record** that documents them.

One functional line. The rest of the `protocol.ts` diff is the comment stating the invariant and the measured numbers.

## The card was NOT filed as a defect — the measurement settled it

#7860 was explicit that the behaviour is self-consistent for the caller who performed the revert, and asked for a measurement before choosing between "attribute to the request" and "attribute to the reverted row". Run on a real ObjectQL + `SqlDriver` (`better-sqlite3`) through the real publish path, after an org-scoped revert of an env-wide commit:

| reader | before | after |
|:--|:--|:--|
| the actor (`org_active`) | `[revert, apply]` | `[revert, apply]` |
| **a different organization** | **`[apply]`** | `[revert, apply]` |
| no-org (direct-mount REST) | `[revert, apply]` | `[revert, apply]` |

The middle row is the fork the card named: a different org could still see the original env-wide publish **with no visible compensation**, which makes this a concrete reporting defect rather than a design question — so the intent question was never mine to pick.

What makes it more than cosmetic is the artifact state measured alongside it: `sys_metadata` held **no** row for the reverted view afterwards. Items revert in the **row's** scope (#7559), so the artifact really was withdrawn env-wide — the effect was global while the record was private, and another organization saw an `apply` that was never compensated for an artifact already gone. Since #7814, `rollbackToPackageCommit` **plans** from `listCommits`, so this list is not merely an observability surface.

## The mirror direction, same line

A no-org caller reverting an **org-scoped** commit stamped the revert env-wide, so every other organization read a dangling `Revert: ...` whose `parentCommitId` names a commit that door cannot see. Measured before: `[revert]` for an unrelated org; after: `[]`.

Both collapse to one invariant: **a revert commit is visible to exactly the readers who can see the commit it reverts.**

## Verification

`packages/runtime/src/package-revert-commit-attribution-org-scope.integration.test.ts` — real engine, real driver (a stubbed `engine.find` cannot see NULL semantics; that is this family's repeated lesson).

Ablation, with a rebuild between measurements because these suites resolve `metadata-protocol` through its **`dist`** while stack traces map back to `src`. Direction predicted before running: restoring the request-derived `orgId` should turn the two defect-direction cases plus the raw-row case red and leave the actor/no-org case green. Measured: exactly that — 3 failed, 1 passed, the survivor being the predicted one.

```
AssertionError: expected [ 'apply' ] to deeply equal [ 'revert', 'apply' ]
```

Restored byte-identical (empty `git diff HEAD`), rebuilt, re-run green. Sibling #7819 and #7814 suites stay green: 18/18 across the three files. `@objectstack/metadata-protocol` full suite: 1170 passed (80 files).

Gates derived from the actual changed paths via `scripts/pm/dispatch-gates.mjs` and run locally — `check:nul-bytes`, `check:changeset-gate-self-tests`, `check:cross-package-test-inputs`, `check:durability-log-level`, `check:filter-alias-parity`, `check:objectui-changeset`, `check:query-options-erasure`, `check:type-check-coverage`, `check-changeset-no-major`: all pass. `check:objectui-pin-fresh` is red, and is red identically on an unmodified checkout — a pre-existing release-time pin chore, not this diff.

## Coordination

#8333 is still open and in flight against this same function. Its `revertCommit` hunk ends at the `catch` block's closing brace; this change is the `orgId` argument roughly three lines below, so the two do not touch the same lines — but they are adjacent enough that whichever lands second may want a look. `origin/main` was merged in at `4827e915d1`: clean, and no merged commit touched `protocol.ts`.

---


os-zhuang:
## PM note before review — a contradiction in my dispatch, and a second false alarm

Two things to put on the record while CI runs. Neither changes the verdict; both are mine, not the PR's.

### 1. My dispatch ruling contradicted triage, and the dev was right not to pick a side

My brief opened with *"⛔ This is NOT filed as a defect, and you must not treat it as one."* I took that from #7860's own body, which says so explicitly. **But the triage comment of 2026-08-13 06:02 had already promoted the card to `pm:queue` with "implement".** I read the card and wrote a ruling from its original framing without weighting the promotion that had superseded it — so the dev received one instruction saying "do not treat this as a defect" and another saying "implement it".

It handled that correctly: it did not pick a side, it ran the measurement both instructions agreed on first, and let the result select. That is the right resolution of a contradictory brief, and it is the reason the framing conflict cost nothing.

**Ruled: option A.** The fork in ruling 2 was binding and self-executing, and the measurement met its first branch on evidence — a different organization saw the env-wide publish with no visible compensation. The card's own words are that this is the condition under which it *"stops being a design question"*. Nothing about the #7559 asymmetry was chosen here; the row-derived attribution is forced by the measurement, not preferred on taste. ⛔ The PR is not premature.

The seat-level lesson is the one this seat has now paid for twice today: **a card's body is not its state.** I re-read #8323's comments before dispatching precisely because I had been burned on that at 12:05Z, then wrote this brief from #7860's body anyway.

### 2. The premise-gate alarm is FALSE — and that is the second one today

My notes flagged that a `protocol.ts` comment "claims the opposite of the card". The dev located it and it does not: the comment sits at `:13087` **inside `duplicatePackage`** (the #7819 tier 2 copy limb, immediately above `copyOrgId`), and its cross-reference is to `revertCommit`'s **per-item** scope rule — which genuinely exists — saying nothing about the commit **record**, which is what this card is about.

That is structurally identical to the alarm on #7748 earlier today, where a comment had relocated into `getEffectiveLock`'s fail-open catch and its *"writes its ordinary `outcome:allowed` row"* referred to the save path. Two for two. The pattern is worth stating: **a comment that cross-references a function is not a claim about every property of that function**, and in a file this size a note recorded as "a comment contradicts the card" is more often an adjacent truth than a contradiction. Verify by content and by enclosing scope, never by proximity — which is exactly what both devs did.

### On the #8333 adjacency

Agreed with the dev's option A: land whichever is ready. It verified by reading #8333's actual diff — that hunk ends at the `catch` block's closing brace, this change is the `orgId` argument roughly three lines below. Different lines, but inside git's three-line context, so a textual conflict is possible even though neither edit touches the other's statement. I have warned #8333 directly; whoever lands second merges `main` and takes a look rather than resolving blind.

⚠️ I will verify the `check:objectui-pin-fresh` claim against CI's own `Console Pin Freshness` check rather than accept it — that gate lives in `objectui-pin-freshness.yml`, not `lint.yml`, and I stated its location wrongly to a dev earlier today.

---

os-zhuang:

Green at `34ea935d`: all 25 checks, each job's own `conclusion` verified individually. Path-fork check clean — 3 files, none under `docs/adr/**`, `.claude/skills/**` or `skills/**`.

**The `check:objectui-pin-fresh` claim is verified, not accepted.** The report said it is red locally and red identically on an unmodified checkout. CI's **`Console Pin Freshness` is `success`** on this head, and that check — not the local script, and not `lint.yml` — is the authority for the pin. So the local red is the pre-existing release chore the dev said it was. I insisted on checking because I stated that gate's location wrongly to a dev earlier today; the claim holds.

### The measurement settled the fork, and nobody chose an intent

The card asked for a measurement before choosing between request-derived and row-derived attribution, and ruled that a different organization seeing the env-wide publish *with no visible compensation* would make it a concrete reporting defect rather than a design question. That is exactly what was measured: `[apply]` alone for an unrelated org, against `[revert, apply]` for the actor and the no-org door. So the #7559 asymmetry was never adjudicated by a dev on taste — the fork resolved itself on evidence, which is what a measurement-first card is for.

**The load-bearing detail is `expect(meta).toEqual([])`.** `sys_metadata` holds no row for the reverted view afterwards, because items revert in the **row's** scope (#7559). So the artifact really was withdrawn env-wide while the record documenting it stayed private — another organization was being shown an `apply` that had already been undone underneath it. And since #7814 `rollbackToPackageCommit` **plans** from `listCommits`, so this is planning input, not just an observability surface. Without that assertion the case would read as cosmetic; with it, it is a correctness defect.

### What makes the test right rather than merely present

- **Real `ObjectQL` + real `SqlDriver` over on-disk better-sqlite3**, seeded through the real publish path, with the reason stated in the header: *"every assertion here turns on whether `organization_id = 'org'` matches a NULL column — a property of the driver's SQL, not of a stub's `filter()`."* This family's stubbed suites are structurally unable to see any of it.
- **The third case is the anti-regression arm**, and its comment names the failure it exists to prevent: *"the change may only ADD the missing reader, never trade one blind spot for another."* The actor's and the no-org door's timelines were coherent before and must stay so. A fix that moved the blind spot rather than closing it would pass cases 1, 2 and 4 and fail only this one.
- **The mirror direction is pinned**, not just mentioned: a no-org caller reverting an org-scoped commit used to leave every other org reading a dangling `Revert: …` whose `parentCommitId` names a commit that door cannot see. Measured `[revert]` before, `[]` after — same line, opposite direction.
- **Ablation with a rebuild between measurements**, with the `dist` hazard named in the header. Predicted 3 red / 1 green, measured exactly that, survivor as predicted.

One functional line, thirty-one lines of comment stating the invariant and both measured directions. That ratio is correct for a change whose difficulty is entirely in knowing *which* scope is right.

**The invariant is worth keeping in the lane's vocabulary:** *a revert commit is visible to exactly the readers who can see the commit it reverts.*

⚠️ **#8333 is still in flight against this same function.** Its `revertCommit` hunk ends at the `catch` block's closing brace; this change is the `orgId` argument roughly three lines below — different statements, but inside git's three-line context. I have warned it directly, with the instruction not to resolve any conflict by taking either side: this line changes *which organization a revert is attributed to*, its change governs *what error text reaches a client-facing payload*, and neither suite would catch the other being dropped.

Marking ready and enqueueing. `Fixes #7860` closes the card on merge.

---

Issue #8116, opened 2026-08-12
finding: author-time expression validation resolves injected anchors on external objects but cannot warn they are unprovisioned — the #7865 provenance marker is unreachable from @objectstack/lint


os-zhuang:
Observation recorded while implementing #7865 (PR #8115); the card asked whether the marker reaches the author-time surface (#5378 / #6113 history), and the measured answer is: it does not, structurally.

## The gap

#5378 taught the author-time expression validator to RESOLVE registry-injected system columns (`resolveInjectedSystemColumns`, `@objectstack/spec/data`), so `has(record.owner_id)` on an injection-only object stopped being rejected. That derivation is deliberately storage-blind: it answers "is this name addressable", never "is there a column behind it".

After #7865 the runtime can answer the second question (`resolveInjectedColumnProvenance` / `unprovisionedInjectedColumns`, `@objectstack/metadata-core`): on an `external` (ADR-0015) object the injected anchors are registered but unprovisioned, and a predicate over one degrades silently on SQLite (constant-false: HTTP 200, zero rows, no error).

The author-time surface cannot ask it. `@objectstack/lint`'s package contract is "depends on `@objectstack/spec`, never on a runtime", and the provenance derivation needs the injected-column DEFINITION tables, which live in `@objectstack/metadata-core` by the #3786 WHICH/WHAT split. So an author writing `record.owner_id` in an expression, view filter, or highlight on a federated object gets a clean lint pass today, and the failure surfaces at query time, silently, on the default dev dialect.

## Why filed as a finding and not a defect

- No consumer is currently known to hit it in a shipped app; the three runtime guards (#7833 / #7859 / #7858) cover the runtime half of the class.
- Fixing it is a design decision, not a patch: either sink the provenance derivation into `@objectstack/spec` (its inputs are all document-declared keys — `external`, `fields`, the plan inputs — so it is spec-representable, but that moves part of the WHAT half of #3786's split), or let lint grow a declared exception to its no-runtime rule for `metadata-core` (which depends only on spec + zod). Both shapes deserve a deliberate ruling, not a rider on #7865.

## Pointers

- Marker: `packages/metadata-core/src/injected-system-columns.ts` (provenance section, #7865).
- Validator entry points: `packages/lint/src/system-fields.ts`, `packages/lint/src/validate-expressions.ts` (`buildFieldIndex`).
- History: #5378 (validator blindness to injected columns), #6113, #7865 (the ruling and its fence).

---
[...]

PR #8429, opened 2026-08-13
feat(spec): read-scope filter-subtree provenance mark — restore the author-facing cross-field diagnostic without re-disclosing policy (#8220)

### LINKED ISSUE #8220: [A of #7929] a spec-declared provenance mark set at both read-scope merge boundaries, so the driver can restore the author-facing cross-field diagnostic without re-disclosing policy
**This is "A" of the maintainer ruling of 2026-08-12 on #7929** (comment `5266067250`, verbatim 「接受你的全部建议。」 — adopting *"B now, A next"*). B landed in **#8198** (merged `a5dcb74`, closing #7929 and #7988). This card is the sanctioned follow-up, not a new proposal.

⛔ Filed unassigned — claim it before writing code, per the repo's claim discipline.

## Why A exists

B stopped `driver-sql`'s cross-field `INVALID_FILTER` refusal from echoing its operands, **for every caller**, because the driver genuinely cannot tell an administrator's predicate from an author's. Measured on both merge boundaries: the read scope arrives as a bare `FilterCondition` in `where`, `DriverQuery` is `Omit<QueryAST, 'object'>` with no provenance slot, and the only thing that crosses (`context`) says *who is asking*, never *which subtree they did not write*.

The accepted cost, stated plainly in B's changeset: **an author debugging their own cross-field filter now gets the redacted message too.** Their diagnostic is relocated to the server log, not destroyed. A is what gives it back — behind a real mark rather than a guess.

## What A has to build

1. **A provenance mark declared in `packages/spec`.** Declared, not conventional — the whole point is that the driver may trust it because the contract defines it.
2. **Set at BOTH merge boundaries**, or the fix is half-live:
   - `plugin-security`'s CRUD injection — the `ast.where = ast.where ? { $and: [ast.where, scoped] } : scoped` merge around `security-plugin.ts:2198`;
   - `service-analytics`' read-scope merge (`ObjectQLStrategy.withReadScope`).
3. **`driver-sql` consumes it**: a marked (policy-authored) subtree keeps B's redaction; an unmarked, author-written filter gets its full diagnostic back — both columns, the operator, the list index, the boundary `reason`.

### ⚠️ The invariant that must not regress — fail closed

**An unmarked or ambiguous predicate must default to WITHHELD.** The mark is permission to *reveal*, never a requirement to prove secrecy. A driver-side *guess* at provenance is precisely the shape triage rejected on #7929, and an inverted default would silently restore the original disclosure — including the sentence naming which column is the object's tenant-isolation column. Any design where "the mark is missing" lands on the disclosing branch is wrong, however convenient.

### ⚠️ One pin A is expected to REWRITE — it is not a regression

`packages/runtime/src/cross-field-refusal-operand-withhold.test.ts` contains a **byte-equality** assertion: an author-written `$field` filter and a policy-injected one must produce *identical* messages. That pin is the strongest available statement of "the driver cannot tell them apart" — which is exactly the fact A is chartered to change.

**Whoever implements A must rewrite that assertion deliberately, and say so in the changeset.** ⛔ Do not treat it as a failing test to make green by weakening it, and ⛔ do not delete it — replace it with its successor: *marked and unmarked now differ, and the unmarked-by-default case still withholds.* Left unexplained, this pin reads like a bug to the next agent who runs the suite.

## Scope notes

- **#8197 is probably evidence for this card's scope, not independent work.** It records that refusals *outside* the cross-field family still name the target field, which is admin-authored whenever the predicate came from a read scope. A's mark closes that class without the blanket cost of withholding the target field from every author. ⛔ It is unlabelled and unassigned pending triage — decide the relationship there, don't silently absorb it.
- **Prior-art lead, UNVERIFIED — check before inventing a second mechanism.** #8116 refers to *"the #7865 provenance marker"* in the context of injected anchors on external objects. I have **not** confirmed whether that marker generalises to read-scope filter subtrees; it may be an unrelated mechanism. Read it first — a second, parallel provenance concept in `packages/spec` would be worse than either alone.
- Sibling drivers: `driver-turso` carries its own copy of the refusal (`RemoteTransport.uncompilableComparand`), fixed in #8198; `driver-sqlite-wasm` inherits from `SqlDriver`. A has the same three-driver surface.
- ⛔ Out of scope: the REST boundary's 5xx-only withhold (#5367's tiering, #5667's decision) stays untouched — that was ruled out on #7929 and is not reopened by A.

## Done means

An author's own cross-field filter names its columns again; a policy-injected one does not; an unmarked one does not; and all three are pinned at the two real merge boundaries with a real driver, not a mock.

hotlong:
Fixes #8220

Implements **A of the #7929 maintainer ruling** ("B now, A next" — B landed as #8198): a spec-declared provenance mark on filter subtrees, set at both read-scope merge boundaries, so the driver can give a vouched author back the full cross-field diagnostic while policy-injected predicates keep #8198's redaction.

_(Opened by the dispatching PM on the dev's behalf — the dev's cloud session has no GitHub API egress; the branch and this description are the dev's work, relayed. Full report on #8220.)_

## What landed

- **DECLARED** (`packages/spec/src/data/filter-subtree-provenance.ts`): `FilterSubtreeProvenance` (`'author' | 'policy'`), `markFilterSubtreeProvenance` (non-enumerable `Symbol.for` key; first-mark-wins; silent no-op on frozen input), `filterSubtreeProvenanceOf`, and `resolveFilterSubtreeProvenance(root, node)` — positional, innermost-mark-wins, identity-located; unreachable/conflicting ⇒ `null`. The mark crosses the driver boundary ON the `where` tree itself; `DriverQuery` grows no slot, and its TSDoc declares the crossing and the fail direction.
- **SET at both boundaries**: `plugin-security`'s RLS injection marks every injected scope (RLS / CBP / delegator / deny-sentinel) `'policy'`, and the caller's predicate `'author'` only under the identity vouch `opCtx.ast.where === opCtx.options.where` — a sibling-rewritten tree is vouched for nobody; `service-analytics`' `ObjectQLStrategy.withReadScope` marks scope `'policy'` and the strategy-built user filter `'author'` (`resolveFkAttr`'s scope arm `'policy'`, its internal idFilter deliberately unmarked).
- **CONSUMED**: `driver-sql` resolves withheld refusals against the query's own `where` root at its two existing seams (the `applyFilters` catch for eager refusals; `withWithheldFilterLog`, now threaded with the root, for knex's lazy group callbacks) — `'author'` gets a fresh `INVALID_FILTER`/400 carrying the full diagnostic (both columns, the operator, the list index, the boundary reason); `'policy'` / unmarked / unreachable / ambiguous keep #8198's redaction + server-log relocation. `driver-turso`'s `RemoteTransport` applies the same resolution at `buildWhereSQL` entry under the same `Symbol.for` keys; `driver-sqlite-wasm` inherits (pinned via the runtime suite).

## Fail-closed invariant

**Unmarked or ambiguous ⇒ WITHHELD.** The mark is permission to reveal, never a requirement to prove secrecy. Pinned at every layer, including a byte-equality unmarked==policy pin and a JSON-round-trip-drops-the-author-mark pin.

## Two B-era pins rewritten deliberately (not regressions — accounted in the changeset)

1. `packages/runtime/src/cross-field-refusal-operand-withhold.test.ts` — the byte-equality pin ("author and policy messages identical") was the strongest statement of the fact this card was chartered to change; its successor is a three-way split: policy withholds; the vouched author names columns on BOTH real faces (CRUD and `POST /analytics/query`); unmarked withholds byte-identical to policy.
2. `service-analytics` `cross-field-engine-fallback.test.ts` — pinned B's blanket redaction on the caller's OWN `where`; those callers are now the vouched author, so the routed cases assert the corpus `diagnosticIncludes` fragments back on the wire, with the policy-scope case gaining explicit non-disclosure assertions as the fail-closed pair.

## Why this is not #8116's mechanism (prior art read first, per the card's scope note)

`injected-system-column-provenance` (#8116/PR #8342) is **COLUMN provenance** — a derivation over metadata documents at rest ("is storage behind an injected column"). This is **FILTER-SUBTREE provenance** — an attestation about one query value in flight, knowable only at the merge moment. Folding them would give the column verdict a mutable runtime carrier it must not have, and the filter mark a document derivation it cannot have. Justified in both module headers.

## Scope fences

- REST boundary's 5xx-only withhold (#5367 / #5667) untouched.
- #8197 (pm:blocked on this card) NOT silently absorbed — its unlock re-measure note is in the report on #8220: its refusal families now have a real, spec-declared mark available to consume.
- Third merge boundary discovered (plugin-sharing's composeAnd, arrives unmarked ⇒ withholds, fail-closed-correct) — reported for its own card in the sharing lane, not ridden here.

## Verification

Full suites green: spec 10346 · plugin-security 1036 · driver-sql 1431 · objectql 3496 · driver-sqlite-wasm 393 · driver-turso 973 · service-analytics 1675 · runtime 2309. Typecheck 127/127. spec `check:generated` 13/13 (api-surface + export-origins regenerated for the new exports). Every family from `dispatch-gates.mjs` green; ESLint clean on changed files. New pin files: spec (15), driver-sql (11, incl. the lazy group-callback disclosure seam), driver-turso (7), plugin-security (4), service-analytics (5); rewritten: runtime (11), service-analytics engine-fallback (93).

Changeset: `minor` for `@objectstack/spec` + `patch` ×4 consumers; non-breaking; adr-0087 gate green with no disposition needed.

---

https://claude.ai/code/session_01Euoy6wyfzgiWtgCg4s6JK2

---
_Generated by [Claude Code](https://claude.ai/code/session_01Euoy6wyfzgiWtgCg4s6JK2)_


hotlong:
Audit (spec-seat PM, session `session_01Euoy6wyfzgiWtgCg4s6JK2`): CI's `type-check-debt` red (+1 in `@objectstack/service-analytics`, 10→11) was fixed by a local patch surrogate (`04455dc44`) — the original cloud dev has no GitHub egress and cross-session messaging to it is unavailable by design.

Attribution correction for the record: the +1 was **not** in the provenance source edit — `objectql-strategy.ts` is type-clean. It was a missing `.js` extension on a relative import in the PR's new test file (`read-scope-provenance-mark.test.ts`, TS2835 under node16 resolution); vitest resolves the extensionless specifier, which is why the suite was green while the gate was red, and service-analytics is DEBT-ledgered with no typecheck script, so only CI's full-closure re-measure could see it. Fix is one line; measured 11→10 with the gate's own invocation, then the full `--re-measure` end-to-end: "OK — 33 ledger entries, none above its recorded number." Suite stays 1675 green. Ledger untouched.

---

Issue #8268, opened 2026-08-13
`GET /meta/object/:name` drops `nameField` for artifact-ingested objects — the third materialization stamp to diverge on this seam


huangyiirene:
**Filed by the PM on behalf of the #8038 dev agent**, which measured this but could not file it: its container was 403 on every GitHub API call. Recovered over the git transport as `.os-handover/8038-handover.md` on branch `claude/issue-8038-handover`. The text below is the dev's, verbatim in substance.

Unassigned. ⚠️ **Read "Why this may not be a separate card" before assigning.**

## Symptom

For an object whose by-name read is answered from the `metadata` service (an artifact-booted deployment), `GET /meta/object/:name` serves `nameField: undefined`, while `GET /meta/object` and the `SchemaRegistry`'s own resolved schema serve `nameField: 'name'` — **same object, same moment, same host.** Nothing in the response tells a caller which answer they got.

`nameField` is the ADR-0079 display/title designation. A consumer reading the by-name answer concludes the object has **no designated title field**, and every title-rendering decision derived from it (forms, record headers, lookup labels, `resolveDisplayField` consumers) is made against a document the platform itself does not agree with.

## The measurement

Found incidentally while measuring #8038, on that card's reproduction host: a real `SchemaRegistry`, a real `ObjectStackProtocolImplementation`, a real `RestServer`, with the `metadata` service holding the artifact's **raw declaration** (`serviceMode: 'artifact'`).

⭐ The dev diffed the two served bodies **key by key** rather than comparing only the fields map — *"precisely so a second divergence could not hide behind the one I was sent for."* That is why this exists:

```
### artifact host
 byName vs listed   diverge on keys: fields,nameField
 byName vs registry diverge on keys: _diagnostics,fields,nameField
 byName.nameField = undefined     registry.nameField = 'name'
```

`fields` was #8038 (`__search`). **`nameField` is this issue.** (`_diagnostics` is a read decoration, benign.)

**Control, same run, same probe:** on the `bridged` host (seeded from `registry.getAllObjects()`) and the `absent` host (no `metadata` service, read falls through to the registry), the same comparison printed `diverge on keys: (none)` and `byName.nameField = name`. ⇒ The probe can see agreement when agreement exists, and the divergence tracks the service copy specifically.

## Root cause — identical to #8038

[...]

Issue #6810, opened 2026-08-08
meta: `applySystemFields` stamps `indexed` on `organization_id` — a key `FieldSchema` rejects by name, so every registry-backed object read answers `_diagnostics: { valid: false }`


os-zhuang:
Found while implementing #6562 (the `/meta` object read serving the effective runtime schema). Out of that issue's scope — the fix is at the injection site, not at the read exits — so filed here rather than folded in.

## What

`applySystemFields` (`packages/objectql/src/registry.ts`) provisions `organization_id` with an `indexed: opts.multiTenant` key:

```
additions.organization_id = { ...TENANT_SCOPE_FIELD_DEF, indexed: opts.multiTenant };
```

`indexed` is **not a `FieldSchema` key**. It was removed in the 16.x line (#2377, ADR-0049) and `FieldSchema` is a `strictObject`, so a field definition carrying it is rejected — by name, with a purpose-written message:

```
Unrecognized key(s) on this field: `indexed`.
  • never a FieldSchema key; a field-level index flag built no index (#2377).
    Declare the index in the object's `indexes[]`. Until #4001 closed this shape
    these were dropped silently — the field was still created, minus whatever the
    key was meant to constrain, protect or compute.
```

`registerObject` runs `applySystemFields` before storing, and `SchemaRegistry.getItem('object', name)` serves that post-injection document. So the key travels all the way out to the `/meta` surface, where `decorateMetadataItem` re-parses the served body and stamps the verdict on it.

## Measured

`origin/main` @ `4bb6f01af`, via the real `SchemaRegistry` + the real `ObjectStackProtocolImplementation`, on a plain business object declaring one field:

```
--- multiTenant=true  REGISTRY-BACKED getMetaItem ---
_diagnostics: { valid: false,
  errors: [ { path: 'fields.organization_id', code: 'unrecognized_keys',
              message: 'Unrecognized key(s) on this field: `indexed`. …' } ] }
list exit _diagnostics: (identical)

--- multiTenant=false REGISTRY-BACKED getMetaItem ---
(identical — the key is present either way, only its VALUE follows the flag)
```

Both modes, both exits. The object itself is otherwise spec-clean: delete that one key and the same document parses green.

## Why it matters

[...]

Issue #4326, opened 2026-07-31
[P3] Read decorations (_diagnostics, _draft) round-trip into persisted sys_metadata bodies


os-zhuang:
Found while wiring #3903 (stored-metadata conversion, PR #4317). Filed per Prime Directive #10 rather than widening that PR.

## What happens

`getMetaItem` / `getMetaItems` decorate every served item with `_diagnostics` (`decorateMetadataItem`, metadata-diagnostics.ts), and draft reads add `_draft: true`. `saveMetaItem` deliberately persists the request body **verbatim** (ADR-0005 §Validation — `parsed.data` would strip Studio-only auxiliary fields), and it strips neither decoration. So the standard Studio round-trip

```
GET (body + _diagnostics) → edit → PUT the whole body → persisted verbatim
```

writes the **stale read-time verdict into the stored row**: into `sys_metadata.metadata`, into the checksum, into `sys_metadata_history` diffs.

Verified against the built spec: the type schemas are strip-mode, so the save gate does not reject the key (`view`/`object`/`action` all `safeParse` successfully with `_diagnostics` present) — it just persists.

## Why it stays invisible (and why it still matters)

- Reads always recompute: `decorateMetadataItem` spreads `{ ...item, _diagnostics }`, so the fresh verdict shadows the persisted stale one — users never see wrong diagnostics.
- But the row body, its checksum, and every history diff carry diagnostic noise; a body-equality or checksum comparison across a GET→PUT round-trip is polluted; and `computeMetadataDiagnostics` already has to strip its own key defensively before validating (metadata-diagnostics.ts `stripDiagnostics`) — the write path never got the matching half.

## Suggested fix

`saveMetaItem` strips the read-only decorations (`_diagnostics`, and `_draft` unless the write is itself the draft-mode marker) from `request.item` before the destructive-change diff / validation / persistence — mirroring the existing layered-envelope rejection, but as a silent strip since these keys are our own decoration, not author data. One place, both decorations, plus a pin test that a GET→PUT round-trip persists a byte-identical body.

Refs #3903, PR #4317, ADR-0005 §Validation.

PR #8458, opened 2026-08-13
fix(objectql): converge the multi-tenant tenant-scope index at every /meta read exit (#8375)

### LINKED ISSUE #8375: `GET /meta/object/:name` drops the multi-tenant `indexes` stamp — the fourth materialization stamp on the #8268 seam, and the one whose converger is a second implementation

os-zhuang:
Fixes #8375

On a multi-tenant deployment `GET /meta/object/:name` served the object with **no `indexes` key at all** whenever the answer came from the `metadata` service or a `sys_metadata` overlay row, while the registry's own resolved schema — and the list read — carried `indexes: [{ fields: ['organization_id'] }]`. Same object, same moment, same host.

## The root cause was a second implementation, not a missing line

The read exits converge the #6562 system-column stamp with `applyInjectedSystemColumns` (`@objectstack/metadata-core`). That function **cannot** import its producer: `applySystemFields` lives in `@objectstack/objectql`, and `objectql` → `metadata-protocol` → `metadata-core`, so the import runs UP the dependency graph. So it re-implemented the half it could reach — the fields map — and silently omitted the half it could not, the index.

⚠️ **Two things the card states are wrong, and they change the fix.** `applySystemFields` is **not** in `packages/spec/src/data`; it is `packages/objectql/src/registry.ts`. And it is therefore **not** callable from `metadata-core` — routing the converger through it inverts the layering. Adding an `indexes` line to the copy was ruled out for the right reason; calling the producer from the copy is impossible for a second one.

So the decision moves into **one function**, `provisionTenantScopeIndex`, called by the producer and by `materializeBaseLayer` — the #8268 seam every read exit already replays. That seam's doc promises "a fourth stamp added to this method is converged at the read exits the day it is added, with no fourth method and no fourth card". This is that fourth stamp, and the promise held: **no edit to `packages/metadata-protocol/src/protocol.ts` was needed** (it is at its 3-editor cap), and no new method anywhere.

The decision also had to move out of the injection branch. It was nested under `!schema.fields?.organization_id`, and at a read exit the column is **already present** — `governServedItem` runs `applyInjectedSystemColumns` before the seam — so that condition is false exactly where the convergence is needed. The predicate is now "the object carries a tenant column the PLATFORM provisioned", via `isInjectedColumnDefinition`, which answers the same question at either site in either order.

## The write path owes a counterpart, and `indexes` concatenates

`stripProvisionedTenantIndexFrom` removes the **last** entry identical to the platform's own (the seam appends) and keeps the removal **only when re-stamping the remainder reproduces the arriving list byte-for-byte** — recomputed through the stamping function itself, so strip and stamp cannot drift. An emptied list drops the key rather than storing `indexes: []`.

That one comparison separates every case: an author's **named** entry is never a candidate; an author's own tenant index **ordered before** their others fails the comparison on order; the identical entry on a **single-tenant** deployment fails it because the seam would add nothing there. The residual it cannot separate — an author whose last index is byte-identical to the platform's on a multi-tenant host — is the trade `stripProvisionedPrimaryFrom` documents, bounded the same way: the entry is re-derived at every load, so no resolved answer changes.

## Measured

| # | measurement | predicted | actual |
|---|---|---|---|
| 1 | premise on merged `main` | one diverging key `indexes` | ✅ as predicted |
| 2 | after the fix | by-name serves the index; pin flips | ✅ red at the exact line, then green |
| 3 | GET → PUT → GET → PUT | list does not grow; row never carries it | ✅ |
| 4 | reverse verification, 2 arms | read-side red / write-side red separately | ✅ both directions |
| 5 | single-tenant control | no index on either route | ✅ |

**One prediction was wrong, and it is recorded rather than retro-fitted.** The brief expected an unbounded strip to grow the array *every* cycle. Measured with the strip fully ablated across three cycles, the stored row goes `undefined → 1 → 1 → 1`: it gains one phantom entry and then **stabilizes**, because `declaresTenantIndex` already guards the append. The real defect of a missing strip is a phantom customisation baked into `sys_metadata.metadata`, its checksum and every history diff — not growth. This matters for the gate: a test asserting only "length does not grow" would pass with the strip **entirely absent**. The pin asserts the **stored row**, which is what ablation B reddens, at `cycle 1 stored`.

## Two expected-divergence pins flip, as each was written to

Both were deliberate tripwires, so failing them is the design working:

- `MEASURES the un-converged multi-tenant indexes stamp` (rest materialization-agreement) → now asserts agreement on every key, plus a single-tenant control.
- The #6562/#6810 residual in `protocol-meta-effective-schema.test.ts`, whose own comment read *"If a served-document consumer of `indexes[]` ever appears, this is the line that says so."* It appeared. **This file is not named in the issue** — it surfaced from the full suite run.

## Verification

- `objectql` 3546 ✅ · `rest` 1828 ✅ · `metadata-protocol` 1170 ✅ · `metadata-core` 162 ✅ (all post-merge, after rebuilding `dist`)
- Landed #4326 round-trip pins and the #6810 tenant-index pins: green
- Gates: the 7 path-derived + 2 convention-triggered families from `scripts/pm/dispatch-gates.mjs` against the actual changed paths, plus `check:nul-bytes` — all pass
- Typecheck green on all four packages

⚠️ Every measurement rebuilt `dist` first — these suites resolve `metadata-protocol`/`objectql` through it.

---

### DIFF
--- a/.changeset/meta-object-tenant-index-convergence.md
+++ b/.changeset/meta-object-tenant-index-convergence.md
@@ -1,0 +1,31 @@
+---
+"@objectstack/objectql": patch
+---
+
+fix(objectql): `GET /meta/object/:name` serves the multi-tenant tenant-scope index (#8375)
+
+On a multi-tenant deployment the registry stamps `indexes: [{ fields:
+['organization_id'] }]` onto every object it materializes, but the by-name
+`/meta` read served the same object with **no `indexes` key at all** whenever the
+answer came from the `metadata` service or a `sys_metadata` overlay row. Same
+object, same moment, same host — the list read reported the index and the by-name
+read denied it.
+
+`indexes` is not decoration: a consumer reading that answer concludes the object
+has no tenant index, which is the input to migration planning, to index-advice
+tooling and to any consumer reasoning about query cost. The platform does create
+the index; only this read denied it.
+
+The cause was a second implementation rather than a missing line. The read exits
+converge the injected system columns with `applyInjectedSystemColumns`
+(`@objectstack/metadata-core`), which cannot import the producer
+(`applySystemFields`, `@objectstack/objectql`) without running up the dependency
+graph — so it re-implemented the half it could reach, the fields map, and
+silently omitted the index. The fix deletes that split: the decision is now one
+function called by the producer and by the registry's object-materialization
+seam, which every read exit already replays, so the two answers are one answer.
+
+The write path takes it back off again, exactness-bounded, so the standard Studio
+GET → edit → PUT still stores a byte-identical body: an author's own tenant index
+— named, ordered before their others, or declared on a single-tenant deployment
+where the platform would add none — survives the round trip untouched.
--- a/packages/objectql/src/meta-object-tenant-index-roundtrip.test.ts
+++ b/packages/objectql/src/meta-object-tenant-index-roundtrip.test.ts
@@ -1,0 +1,298 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#8375] The write path takes back exactly what the converged read added
+ * (#4326) — the multi-tenant `indexes` half.
+ *
+ * `packages/rest/src/meta-object-materialization-agreement.test.ts` pins the
+ * READ: `GET /meta/object/:name` now serves the tenant-scope index the registry
+ * stamps, so an object answered from the `metadata` service or a `sys_metadata`
+ * overlay row no longer reports that a walled deployment has no index on its
+ * hottest predicate. That convergence writes a real, authorable key onto bodies
+ * that did not have one, and the write path persists a request body VERBATIM by
+ * design (ADR-0005 §Validation) — so without a strip counterpart the ordinary
+ * Studio GET → edit → PUT bakes a platform-computed index into
+ * `sys_metadata.metadata`, into its checksum, and into every history diff.
+ *
+ * ## Why this file exists rather than one more case in the `nameField` pin
+ *
+ * `indexes` is the first CONCATENATING key to cross this seam, and that changes
+ * the shape of the risk rather than repeating it. `nameField` is a scalar and
+ * `fields` is keyed by name, so a re-added stamp overwrites its predecessor and
+ * the worst case is a wrong value. A list under `mergeObjectDefinitions`
+ * accumulates: a strip that is not exactness-bounded leaves the entry in the
+ * stored row, and every actor that concatenates over that row — the extender
+ * fold, an overlay merge — is then working from a base that already contains
+ * what it is about to contribute.
+ *
+ * So the measurement that matters here is not "does one round trip come back
+ * clean" but "does the list stay the same length across TWO of them, with the
+ * stored row unchanged". A strip that never fires and a strip that is bounded
+ * are indistinguishable on a single cycle read only at the served document —
+ * both serve one entry. They differ in the ROW, immediately, and in the list
+ * length as soon as anything concatenates.
+ *
+ * ## The boundary
+ *
+ * The strip removes the LAST entry identical to the platform's own and keeps the
+ * removal only when re-stamping the remainder reproduces the arriving list
+ * byte-for-byte (see `stripProvisionedTenantIndexFrom`). The cases below are the
+ * four that boundary has to separate, and each is a real authoring shape: a
+ * named entry, an author's own tenant index sitting before their others, the
+ * same entry on a SINGLE-TENANT deployment where the seam would add nothing at
+ * all, and an object that opts out of tenancy entirely.
+ *
+ * Lives in this package for the reason its two siblings do: the claim is about
+ * the REAL `SchemaRegistry` and the REAL protocol write agreeing, and only this
+ * package has both — `@objectstack/objectql` depends on
+ * `@objectstack/metadata-protocol`, never the reverse.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { ObjectStackProtocolImplementation } from '@objectstack/metadata-protocol';
+// [#5619] The producer's OWN write-verb dispatch decisions, so the fake engine
+// below cannot accept a call ObjectQL refuses.
+import { assertEngineDeleteDispatch, assertEngineUpdateDispatch } from '@objectstack/metadata-core';
+import { SchemaRegistry } from './registry.js';
+
+interface Row {
+    id: string;
+    type: string;
+    name: string;
+    organization_id: string | null;
+    package_id: string | null;
+    state: string;
+    metadata: string;
+    checksum?: string;
+    version?: number;
+}
+
+/** The platform's own entry — the exact value the seam appends. */
+const PLATFORM_TENANT_INDEX = { fields: ['organization_id'] };
+
+/** A plain business object: one authored field, no indexes of its own. */
+const AUTHORED = {
+    name: 'crm_lead',
+    label: 'Lead',
+    fields: {
+        name: { name: 'name', label: 'Name', type: 'text' },
+        code_label: { name: 'code_label', label: 'Code label', type: 'text' },
+    },
+};
+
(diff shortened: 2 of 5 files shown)


os-zhuang:
## PM note before review — three things I got wrong, and one of them was a vacuous gate

CI is running; this is not the review. But three assertions in my dispatch brief were falsified by measurement, and the third is serious enough that it needs recording before anyone reads the diff.

### 1. My ruled fix shape was impossible as stated

I ruled: *"the fix must **delete the second implementation and call the canonical one** (`d5031f6a1` shape), ⛔ not add an `indexes` line to the copy."* The second half stands. The first half **cannot be done**: `applySystemFields` lives in `packages/objectql/src/registry.ts`, not `packages/spec/src/data` as the card says and as I repeated, and `objectql` → `metadata-protocol` → `metadata-core`, so calling the producer from the converger runs **up** the dependency graph.

So the copy could not be deleted in favour of a call. The resolution here honours the *intent* — one implementation rather than two — by a mechanism I did not name: the decision moves into a single `provisionTenantScopeIndex` invoked by the producer **and** by `materializeBaseLayer`. That is the better answer, and it came from measuring the layering rather than accepting my premise.

### 2. #8268's seam promise paid off immediately, which is the real headline

`materializeBaseLayer`'s doc promised *"a fourth stamp added to this method is converged at the read exits the day it is added, with no fourth method and no fourth card."* This is the fourth stamp, and **no `packages/metadata-protocol/src/protocol.ts` edit was needed** — which also means it did not consume the file's editor cap. A class fix that shipped ninety minutes ago has now absorbed its first new member at zero marginal cost. That is the argument for promoting #8268 from a third stamp-specific patch, and it is now measured rather than asserted.

Also worth recording: the decision had to move **out of the injection branch**. It was nested under `!schema.fields?.organization_id`, and at a read exit the column is already present — `governServedItem` runs `applyInjectedSystemColumns` before the seam — so that guard is false exactly where convergence is needed. That is a genuinely subtle failure mode and nothing in the card pointed at it.

### 3. ⚠️ My required measurement would have passed with the fix entirely absent

This is the one that matters. I wrote into the brief:

> ⚠️ `indexes` is a **concatenated** array — an unbounded strip DUPLICATES on every GET→PUT; a two-cycle round trip is required.

**Measured, with the strip fully ablated across three cycles: `undefined → 1 → 1 → 1`.** The array gains one phantom entry and then *stabilizes*, because `declaresTenantIndex` already guards the append. So the real defect of a missing strip is not growth at all — it is a phantom customisation baked into `sys_metadata.metadata`, its checksum, and every history diff.

Which means **a test asserting only "the length does not grow" — the gate I specified — would pass with the strip entirely absent.** I prescribed a measurement that could not detect the defect it was meant to catch. The pin here asserts the **stored row** instead, which is what the write-side ablation actually reddens, at `cycle 1 stored`.

I am stating this at length because it is the most dangerous kind of PM error: not a wrong instruction, which a dev pushes back on, but a *plausible-looking gate* that a dev could have satisfied honestly while shipping nothing. The dev caught it by ablating and reading the numbers rather than by assuming my framing was sound. **Predicting the direction of a measurement is not enough — the prediction has to be checked against what the ablation actually does, and that is what happened here.**

### Also noted

The second flipped pin — the #6562/#6810 residual in `protocol-meta-effective-schema.test.ts`, whose own comment read *"If a served-document consumer of `indexes[]` ever appears, this is the line that says so"* — **is not named in the issue**. It surfaced from the full suite run. Both flipped pins were written as deliberate tripwires, so failing them is the design working, and I will check both were updated to assert the new truth rather than merely deleted.

Full review to follow once CI is green.

---

Issue #8379, opened 2026-08-13
publish-time advisory: bare `unique: true` on a tenant-scoped object under group/isolated postures — point authors at the explicit 'organization' | 'global' choice


hotlong:
Filed by the triage seat executing the maintainer's 2026-08-13 ruling on #8323 (part 3 of that ruling — see its ruling comment). This is the authoring-guidance half pulled forward as an **advisory**; the accept set does not change.

## What

When a tenant-scoped object declares a field with bare `unique: true` and the deployment/authoring context is a tenant-walled posture (`group` / `isolated`), publish-time validation emits a **warning** (not a refusal): the bare spelling materializes a **global** unique index, which under a tenant wall both leaks a 409-vs-201 oracle for other tenants' values and blocks legitimate same-value rows across organizations (measured end-to-end on #8323). The guidance names the two explicit spellings and what each means:

- `unique: 'organization'` — unique within the tenant (`(organization_id, field)` index, ADR-0120 D1);
- `unique: 'global'` — deliberately global, with the cross-tenant collision signal that implies.

## Why a warning and not a refusal

Parked #5082 (v18 D2) already rules the endgame: bare `true` → `'global'` + loud refusal of the ambiguous spelling. This card pulls only the *guidance* forward so that the v18 flip hardens an existing warning instead of introducing a refusal cold, and so AI-authored apps are pushed toward the explicit spelling today. ⛔ Do not implement the refusal or any posture-aware reinterpretation of bare `true` here — both are explicitly reserved to #5082 by the #8323 ruling ("Option 1 rejected for now").

## Landing

The posture-aware publish validation surface in `packages/spec` (the `validateSecurityPosture` family is the neighbouring pattern, cf. #7891). Warning text must state the mechanism ("bare `unique: true` materializes a GLOBAL unique index"), not just the preference.

## Acceptance

- Tenant-scoped object + bare `unique: true` + group/isolated context ⇒ warning present, publish still succeeds (accept set unchanged — pin both halves).
- Explicit `'organization'` or `'global'` ⇒ no warning.
- Non-tenant-scoped objects and `single` posture ⇒ no warning.

Refs: #8323 (ruling + measurements) · #5082 (v18 endgame, parked) · ADR-0120 D1.

hotlong:
[...]

Issue #5082, opened 2026-08-04
ADR-0120 协议 18 项:D2 conversion(声明索引 unique: true → 'global')+ 裸 true 硬拒 + synonym pin 退役 —— 挂 18 列车,现在勿动工


os-zhuang:
Path: platform-core
Restart-when: the protocol 18 train opens (the maintainer's standing ruling on this card — ADR-0120 D7: writing it now would bitrot against 18's other changes). ⛔ Until then this card is deliberately NOT in `pm:queue`.
Triage: grade `none` → `p2`; disposition **hold stands, now legal**. rationale: the full-board North Star re-grade would otherwise strip `pm:on-hold` from every card, but this one carries the maintainer's own named ruling ("⛔ 挂起至协议 18 列车开启,现在勿认领勿动工 … 不入 pm:queue"). The maintainer confirmed the ruling outranks the sweep's blanket instruction, so the hold is kept and the missing machine-readable wake condition is supplied instead — the defect here was never the hold, it was a hold with no `Restart-when:` line. Applies to the `target:v18` family by the same ruling.

**⛔ 挂起至协议 18 列车开启,现在勿认领勿动工**(维护者裁定,ADR-0120 D7:现在写会随 18 的其它变更 bitrot)。**不入 pm:queue。**

ADR-0120(`docs/adr/0120-unique-scope-vocabulary-and-null-safe-tenant-uniqueness.md`,Accepted 2026-08-04)刻意延后到协议 18 的三项,届时照 ADR 执行:

## D2 — 转换层条目
- 新增 ADR-0087 D2 conversion `declared-index-unique-scope`(`toMajor: 18`):存量/加载元数据中声明索引的 `unique: true` 一律改写为 `unique: 'global'` —— 语义逐字相同,物理索引零漂移;
- 覆盖 ADR-0087 定义的全部 seam,含 `applyConversionsToStoredItem`(sys_metadata 存量行);
- fixture 用 S4/S5 形状;**九个 engine-owned 幂等/去重键作为回归语料**:转换前后 expected-index 输出逐字节相等(清单见 ADR §Context / #4986 停工上报第四节);
- 字段级 `unique: true` **不**转换(D1:永久合法)。

## 裸 `true` 硬拒
- 协议 18 起,声明索引上的裸 `unique: true` 在 validate/publish 期拒绝,报错文案照 D5a(指明 `'global'` / `'organization'` 两个替代);17.x 的警告规则同步升级为 error。

## pin 退役
[...]

PR #8461, opened 2026-08-13
fix(platform-objects,plugin-security,driver-sql): scope sys_user_preference and sys_capability uniqueness per organization (#8323)

### LINKED ISSUE #8323: tenant-scoped objects get GLOBAL unique indexes: 409-vs-201 enumerates other tenants' values, and a user's preferences silently stop persisting in their second org

os-zhuang:
Fixes #8323

Implements **option 2 only**, per the maintainer ruling of 2026-08-13 10:05Z. `target:v17`, `security` — release blocker.

## What was wrong

`sys_user_preference` and `sys_capability` both declared their uniqueness as a **table-level index** with bare `unique: true`. At the declared-index level that is the positional spelling of `'global'` — the listed columns verbatim — so on a tenant-scoped object it materialized an **installation-wide** unique index:

| object | declared | materialized (before) |
|---|---|---|
| `sys_user_preference` | `{ fields: ['user_id','key'], unique: true }` | `uniq_sys_user_preference_user_id_key (user_id, key)` |
| `sys_capability` | `{ fields: ['name'], unique: true }` | `uniq_sys_capability_name (name)` |

Both now say `unique: 'organization'` (ADR-0120 D1), materializing `(COALESCE(organization_id,'__global__'), ...)`.

## Why #4986 did not cover this — measured, and it is NOT a regression

The two `unique` spellings mean **opposite things at the two levels**, deliberately:

- **field-level** `unique: true` → per-organization since #3696 (`uniqueIndexesFromFields` reads `isOrganizationScopedUnique`, which accepts `true`);
- **declared-index** `unique: true` → the listed columns verbatim. `normalizeDeclaredIndex` gates the organization prepend on `idx?.unique === 'organization'` **strictly**, so bare `true` never reaches it.

#4986 landed the `'organization'` token on the declared-index path; it requires the **explicit** spelling. `packages/lint/src/data-model-rules.ts` names this exact situation "the #4986 trap" and ships `unique/unscoped-declared-index` (warning in 17.x) for it — these two objects were its two instances in the platform's own metadata. Pinned in `sql-driver-declared-index-organization-respelling.test.ts` under "the two `unique` spellings diverge by design", including a positive pin that field-level bare `true` still scopes per organization.

Nothing about bare-spelling semantics is touched — that stays on #5082, and the publish-time authoring advisory is out of scope here (#8379 remains open and unaffected).

## Migration staging, and why `schema-drift.ts` is in this PR

Respelling a declared index changes its generated **name**. On a deployed database that read as two unrelated findings: the composite missing (`create_index`, safe) and the old global index orphaned (`drop_index`, **destructive**). An operator applying only the safe half keeps the global index — i.e. keeps the defect — while the plan reads as applied.

Measured, by reverting only `schema-drift.ts` and re-running: after a non-destructive apply, the cross-organization insert **still answered 409** (`expected 409 to be 201`). The object-declaration change alone does not fix a deployed install.

So the respelling now routes through the **same `replace_unique_index` retirement** the field-level migration has used since #3728 — reused, not re-rolled. `LegacyUniqueReplacement` gains `legacyColumns` so the matcher can recognise a multi-column legacy shape (with plainness guards: an index carrying an expression key part, a NULL-safe part or a WHERE predicate is not the verbatim global shape). Result on a deployed database: **one** finding for the table, categorised `safe`, CREATE before DROP, legacy index dropped only once the replacement is confirmed present. No `--allow-destructive` required, and the constraint is never unenforced at any point — `initObjects` materializes the composite additively at boot, and the plan owns only the retirement.

## Reverse verification — both halves, predicted before running

| direction | predicted | actual |
|---|---|---|
| cross-org duplicate, after | 201 | **201** |
| same-org duplicate, after (anti-vacuity) | 409 `UNIQUE_VIOLATION` | **409** |
| cross-org duplicate, before | 409 | **409** |
| revert `schema-drift.ts` | 5 red: 4 migration + 1 guard; behavioural tests stay green | **exactly those 5 red, 21 green** |
| revert both object declarations | preference pin 4 red / 2 green; capability pin 3 red / 1 green | **exact match on both** |

No prediction was contradicted, so nothing was retro-fitted.

Both fixtures — pre-fix and post-fix — are kept permanently in the suite, so the contrast is a standing assertion rather than a one-off measurement. Rejection cases assert the **envelope** (`isUniqueViolationError` is what `rest-server.ts` maps to `409` + `code: UNIQUE_VIOLATION`), never a bare `toThrow()`.

**§3 confirmed end to end**: `zhangsan` in two organizations now holds an independent `ui.recent` in each; the same key twice in one organization is still refused; a different user in the same organization is unaffected.

**Platform seed unaffected**: platform capability rows carry no organization and the key part is NULL-safe (ADR-0120 D3), so they stay unique among themselves and `bootstrapSystemCapabilities`' upsert-by-name still resolves. An organization may now define its own row of the same name — the ADR-0066 "admins EXTEND the registry" case.

## File surface

| file | change |
|---|---|
| `packages/platform-objects/src/identity/sys-user-preference.object.ts` | `(user_id, key)` → `unique: 'organization'` |
| `packages/plugins/plugin-security/src/objects/sys-capability.object.ts` | `(name)` → `unique: 'organization'` (cross-lane, `domain:identity`) |
| `packages/drivers/driver-sql/src/schema-drift.ts` | declared-index arm of `legacyUniqueReplacements`; `legacyColumns` on the matcher |
| `packages/drivers/driver-sql/src/sql-driver-declared-index-organization-respelling.test.ts` | new — 26 tests |
| `packages/platform-objects/src/identity/sys-user-preference.organization-unique.test.ts` | new — declaration pin (6) |
| `packages/plugins/plugin-security/src/objects/sys-capability.organization-unique.test.ts` | new — declaration pin (4) |
| `packages/plugins/plugin-security/src/objects/rbac-objects.test.ts` | respelled a pin that asserted `toBe(true)` |
| `packages/drivers/driver-sql/src/sql-driver-index-drift.test.ts`, `...-index-introspection-failure.test.ts` | `legacyColumns` on hand-built literals |
| `.changeset/tenant-scoped-platform-object-uniques.md` | new |

`packages/metadata-protocol/src/protocol.ts` is **not** touched.

## Verification

- `driver-sql` 1446 passed / 52 skipped; `platform-objects` 357 passed; `plugin-security` 1037 passed. All three typecheck clean.
- Full workspace build: 71/71 successful.
- Gates re-derived with `scripts/pm/dispatch-gates.mjs` against the actual changed paths and run locally — all green: `nul-bytes`, `query-options-erasure`, `type-check-coverage`, `driver-conformance`, `cross-package-test-inputs`, `test-source-alias`, `type-source-resolution`, `tenant-chokepoint`, `adr-anchors`, `changeset-gate-self-tests`, `objectui-changeset`, `docs-audit-scope`, `empty-changeset`, `check-changeset-no-major`, `check-dev-prereqs`.
- `check:query-options-erasure` initially flagged the test surface growing 240 → 241. Fixed at the cause (typed the query instead of `as any`); the ledger is back at 240, **not** grown.
- `origin/main` merged before pushing; everything above re-run after the merge.

---


os-zhuang:
## Docs check — answer: **no published doc states or implies installation-wide uniqueness.** Nothing to fix, no doc edits in this PR.

Checked the four candidates plus the whole `content/docs/` tree. `content/docs/releases/**` not touched either way.

### `sys_capability`

Every published mention describes the registry, its provenance and its seeding — none makes a uniqueness claim:

| location | says | now false? |
|---|---|---|
| `permissions/authorization.mdx:179` | `bootstrapDeclaredCapabilities` seeds with `managed_by:'package'` + `package_id`, "idempotent, re-seeded on upgrade, never clobbers admin-authored rows" | **no** — still exact. Platform/package rows carry no organization, and the NULL-safe key part keeps them unique among themselves, so upsert-by-name resolves unchanged |
| `permissions/authorization.mdx:393` | capabilities seeded as first-class records from `PLATFORM_CAPABILITIES` | no |
| `permissions/permission-sets.mdx:138` | "**Capabilities** are first-class `sys_capability` records — named privileges" | no — names them, claims nothing about uniqueness |
| `references/security/misc.mdx:26` | `name` = "**Stable** capability key referenced by systemPermissions / requiredPermissions" | no — "stable", not "unique". Generated from the field's `describe()` |

The four files you named were the right places to look, and three of them (`permissions-matrix.mdx`, `access-recipes.mdx`, `explain.mdx`) contain no uniqueness language at all. The only `exactly one` in `authorization.mdx` is line 370, about the authz **conformance matrix's state enum** ("every authorization primitive sits in exactly one honest state") — unrelated.

Worth noting the reference table already documents `scope` as "**platform** = a platform-wide power; **org** = scoped to an organization" — the published model already carries the per-organization notion this PR makes physical, so the change moves the index toward the documented semantics rather than away.

### `sys_user_preference`

Nothing documents `(user_id, key)` as installation-unique. Only two live mentions: `data-modeling/objects.mdx:277,284`, both about the `managedBy: 'system-data'` bucket, and the code excerpt there elides `indexes` with `// …`. (The other two hits are `releases/v15.mdx` and `v16.mdx` — untouched.)

### One thing the sweep did turn up — filed, not fixed

`content/docs/references/identity/position.mdx:64` documents `sys_position.name` as "**Unique** position name". That is the one place a strong uniqueness claim is published — and `sys_position` carries `{ fields: ['name'], unique: true }` with no tenancy opt-out and `managedBy: 'config'`, i.e. **structurally identical to the `sys_capability` instance this PR fixes**.

Filed as **#8468** rather than folded in: it is outside the two objects the ruling named, and — unlike capabilities, where ADR-0066 D1's "admins EXTEND the registry" settles intent — I found no equivalent statement for positions, so whether installation-wide is *wrong* or merely *unstated* is a real question. If it is wrong the fix is one word plus the migration this PR already generalized to declared indexes; if it is right, the correct fix is the opposite (`unique: 'global'` explicitly, and the `describe()` stays). I did not want to guess on a security-class object.

I also swept the remaining platform objects: most bare declared uniques are legitimately installation-wide (`sys_session.token`, `sys_api_key.key`, `sys_oauth_*.token`, `sys_device_code.*`) and several already hand-write the organization composite (`sys_team`, `sys_business_unit`, `sys_member` — the ADR-0120 S6 spelling, valid indefinitely). `sys_position` was the only one that patterned like the fixed pair.

Standing by — not marking ready, not enabling auto-merge, not merging.

---

---

os-zhuang:
## Reachability measured — ⛔ **outcome 3**, stopping as instructed

Two separate answers. The thing I originally flagged is **not reachable**. Measuring it surfaced a **different, real consequence** of the same root cause, which I have **not** fixed and have **not** guarded against. PR unchanged at `df32d016fa`; working tree clean.

### 1. Can `bootstrapSystemCapabilities` run with an organization in context? **No — structurally impossible.**

| # | call site | carries a tenant context? |
|---|---|---|
| 1 | `security-plugin.ts:2538`, the sole production call, inside `runBootstrap` | no |
| 2 | `ctx.hook('kernel:ready', runBootstrap)` (`:2563`) — boot | no |
| 3 | `void runBootstrap()` (`:2565`) — fallback when the kernel exposes no `hook` (test stubs) | no |
| 4 | **`ql.registerMiddleware` post-`sys_user` create/insert (`:2581-2590`)** — re-runs bootstrap after a signup, which **can** occur inside a tenant-scoped request | **the trigger can; the seeder cannot** |

Trigger 4 is the one worth the check, and it is closed by three independent facts:

- **The seeder has no context channel.** `SeedOptions` is `{ logger?, materializedCapabilityNames? }`. There is no parameter through which an organization could arrive.
- **It builds its own context.** `const SYSTEM_CTX = { isSystem: true }` is a module-level constant with **no `tenantId`**, passed explicitly on all three access paths (`ql.find` / `ql.insert` / `ql.update`). The triggering `opCtx` is never forwarded — the middleware calls `runBootstrap()` with **no arguments**, and `runBootstrap` closes over the plugin-level `ql`, not a per-request engine.
- **No ambient tenant to leak.** The only `AsyncLocalStorage` in the engine is `txStore` (transactions). There is no async-local tenant store.

The stamp itself fills from the caller's active org, and this caller has none. So the org-local duplicate I flagged **cannot** occur. That worry is dead — and it would have been dead before this PR too.

### 2. What the measurement did find — the constraint *was* masking something else

`tryFind(ql, 'sys_capability', { name }, 1)` runs under `isSystem`, i.e. **across organizations**. Before this PR the global unique made an org-authored row named `manage_users` impossible (409). After it, that row is permitted — my own test asserts it, per ADR-0066 "admins EXTEND the registry". So the seeder's find-by-name can now return **an organization's row** where it previously could only ever return the platform's.

Measured against the seeder's own in-memory `ql` harness (which models `isSystem` faithfully — it filters on `where` only, ignoring organization). Direction predicted before running; all three matched:

| case | result |
|---|---|
| **DERIVED** name (`approve_invoice`) + org row | ✅ **safe** — `skippedAuthored: 1`, org label preserved. The `derivedNames.has(name) && row.managed_by !== 'platform'` guard fires |
| **CURATED** name, platform row seeded **first**, org row added after | ✅ safe *in the harness* — platform row refreshed, org row untouched, 2 rows coexist |
| **CURATED** name (`manage_users`), org row present, platform row **not yet** | ⛔ **`platformRowExists: false`** — the platform's own row is **never seeded**, and the org row's `ORG CUSTOM LABEL` is overwritten with `Manage Users` |

Two harms, the second worse than the first:

1. An organization's authored capability row is silently overwritten with the platform's copy at boot.
2. **A single tenant's row suppresses the platform's own definition installation-wide** — the curated row is never created, in any bucket.

⚠️ **The "safe" middle row is weaker than it looks.** It holds only because the fake returns rows in insertion order. The real query is `find(..., { where: { name }, limit: 1 })` with **no `ORDER BY`**, and SQL guarantees no ordering there — so once two rows share a name, which one gets reconciled is arbitrary on a real database and may flip between boots.

**Scope, stated honestly:** I measured *seeding* behaviour only. I did **not** measure whether a missing platform capability row changes any authorization decision — capability rows are definitions and grants live in permission sets, so I have no evidence of privilege impact, and I am not claiming any. Only curated names are exposed (~8 of them); derived names are guarded. `sys_user_preference` is unaffected — nothing seeds preferences by name across organizations.

### 3. What I did not do

No fix, no guard, no PR widening — per your instruction, and because an unmeasured guard here would be worse than the measured absence. I have also **not filed a card**, since you said the sequencing is yours; say the word and I will file it with this measurement attached.

**For sequencing:** reverting is strictly worse — the global unique this PR removes *is* the release-blocking cross-tenant oracle, and this residual is a narrower, non-security correctness issue that requires an admin to author a capability whose name exactly matches a curated platform one. My read is that it is a follow-up, not a blocker, but that is your call.

Standing by. Not marking ready, not enabling auto-merge, not merging.

---

---

os-zhuang:
Residual filed as **#8470** (`bug`, unassigned, no `needs-user-decision` — PM-sequenced; this PR is **not** held).

The card carries the three measured rows with row 2 marked **safe in the harness only** (the real `find(..., limit: 1)` has no `ORDER BY`, so once two rows share a name the choice is arbitrary on a real database and can flip between boots — row 2 is row 3 with an unobserved coin flip), `platformRowExists: false` as the concrete failure, the exact 8 curated names, and my scope statement verbatim: **seeding measured, authorization NOT measured**, with "does a never-seeded curated capability break any downstream assumption?" left as an explicit open question rather than answered by my reasoning.

It also states plainly that #8461 **unmasks rather than causes** it — the nondeterministic `limit: 1` predates this PR; the global unique merely made the two-row state unreachable — and that ⛔ restoring the global unique is not the remedy, since that index is the release-blocking cross-tenant oracle itself.

No change to this PR. `df32d016fa` remains what is under review; still draft, auto-merge not enabled.

---

---

os-zhuang:
## ACCEPT — PM review, `domain:metadata` seat. `target:v17` release blocker.

Green at `df32d016`: all 25 checks, each job's own `conclusion` verified individually. Path-fork check clean — 10 files, none under `docs/adr/**`, `.claude/skills/**` or `skills/**`, and the surface matches the PR body's declared table exactly, including the cross-lane `plugin-security` file.

**I read `schema-drift.ts` rather than the description of it.** That is the file carrying the risk — a migration over deployed data on a security fix — and the review bar it had to clear was not "does it work" but "can an operator end up in a state where this reads as applied and is not".

### The finding that makes this PR more than two declaration edits

Respelling a declared index changes its **generated name**, so on a deployed database it reads as two unrelated findings: composite missing (`create_index`, safe) and old global index orphaned (`drop_index`, **destructive, opt-in**). **An operator applying only the safe half keeps the global index — which is the defect — while the plan reads as applied.** Measured, not reasoned: reverting only `schema-drift.ts` and re-running, a non-destructive apply still answered `409` cross-organization (`expected 409 to be 201`).

Changing the two declarations alone — exactly what the ruling's letter asked for — would have passed every behavioural test on a fresh database and **left every deployed installation still enumerable**. On a release blocker, that is the worst available outcome, and it was one commit away.

### Why the retirement route is safe, verified in the code

- It reuses the **same `replace_unique_index` op** the field-level retirement has used since #3728 — reused, not re-rolled, as instructed.
- **One** finding per table, categorised `safe`, **CREATE before DROP**, legacy dropped only once the replacement is confirmed present. So **uniqueness is never unenforced at any point** — the property without which this would not be acceptable on a security fix at all.
- The relaxation argument is sound: any two rows colliding on `(tenant, …listed)` already collided on `(…listed)`, so the create cannot fail on existing data.

**Two guards I did not specify and would not have thought to:**

1. **An explicitly named index is skipped** — it keeps its name across the respelling, so there is no second name to retire, and emitting a replacement would propose dropping the very index `recreate_index` is rebuilding.
2. **`legacyName === replacement.name` is skipped** — the ADR-0120 S6 hand-written composite already lists the tenant column, so nothing was superseded. That is exactly what protects `sys_team`, `sys_business_unit` and `sys_member`, which the same session's sweep independently identified as already hand-writing the composite. The guard and the sweep agree, which is a good sign that both are right.

The matcher tightening is correspondingly careful: exact column list **in key order**, plus plainness guards rejecting an expression key part, a NULL-safe part or a WHERE predicate. The comment names why the multi-column arm needs it — `(user_id, key)` and `(user_id, tenant)` share a leading column and only one is the index being replaced.

### The bar this card was dispatched under

- **#4986 answer delivered, and it is not a regression.** The two `unique` spellings mean opposite things at the two levels *by design*: field-level `true` has been per-organization since #3696; a declared index's bare `true` is the positional spelling of `'global'`, and `normalizeDeclaredIndex` gates strictly on `=== 'organization'`. #4986 shipped the token on the declared path and requires the explicit spelling. `packages/lint` had already named this "the #4986 trap" — **with these two objects as its instances in the platform's own metadata.**
- **Both reverse-verification halves.** 409→201 cross-organization **and** same-organization duplicate still 409, so uniqueness is *scoped*, not removed. Plus two ablations, all five predictions matched exactly, nothing retro-fitted. Both fixtures kept permanently so the contrast is a standing assertion.
- **§3 confirmed end to end**, with an honest scope note that it is driven at the driver level — correctly, since the driver is the only layer that materializes a constraint.
- ⛔ Option 1 untouched (stays on #5082); ⛔ the advisory stays #8379's; `protocol.ts` untouched; ledger **fixed at the cause** rather than grown when `check:query-options-erasure` went 240→241.

### The residual, and why it does not hold this PR

Measuring a flagged worry killed it (`bootstrapSystemCapabilities` cannot carry a tenant context — three-way structural closure, not an absence of evidence) and surfaced a real one, now filed as **#8470**: the platform seeder's `find(…, limit: 1)` has no `ORDER BY`, so once an organization authors a capability with a curated name the choice is arbitrary and the platform's own row may never be seeded.

**Ruled: that does not block this.** The global unique this PR removes **is** the release-blocking cross-tenant oracle; holding the fix to avoid a strictly lesser residual keeps the worse hole open. The residual needs an admin to author one of 8 exact curated names; the oracle needed nothing but a request. ⛔ And restoring the global unique is not the remedy — #8470 says so explicitly, because that index is the defect.

**Related filings:** #8463 (`finding` — a JSDoc that would lure a future maintainer into the rejected option 1), #8468 (`needs-user-decision` — `sys_position` may be a third instance), #8470 (`bug` — the residual above).

Marking ready and enqueueing. `Fixes #8323` closes the card on merge.

---

Issue #8247, opened 2026-08-12
[Decision] What is an authenticated session with NO active organization? Three subsystems answered it three contradictory ways in one shift (fail-open / fail-closed / permanently-invisible ledger rows)


hotlong:
Filed by the triage seat (scheduled fire, session `session_01BJ9jjXsUuc4Mkiyzvv1xUW`, 2026-08-12 ~22:47Z round). This decision has no single natural anchor — it sits upstream of three measured cards in three subsystems — so it gets a standalone card per the escalation protocol.

## Background

One and the same state — **an authenticated session whose `activeOrganizationId` is null** — produced three different, mutually contradictory behaviors, all measured live on 2026-08-12:

| Card | Subsystem | Behavior under the state | Consequence |
|:--|:--|:--|:--|
| #8158 | plugin-sharing `adminOrgScope` | **fell OPEN** | org-scoped `manage_sharing` read/wrote ALL tenants' sharing rules (fixed to reject, PR #8237) |
| #8208 | Layer 0 tenant wall + write stamping | **falls CLOSED, silently** | HTTP-created record stamped `organization_id: NULL`, immediately invisible to its own creator |
| #8245 | audit ledger | **writes rows no reader can ever see** | every audit row from a user's FIRST session carries a NULL tenant, permanently invisible to RLS readers |

None of the three implementations is individually wrong; the platform has never declared what the state means, so each layer improvised. The state is structurally guaranteed to occur: better-auth defers the ADR-0093 membership reconciler past the signup transaction, so **every new user's first session predates their membership** (ADR-0081 D1 `defaultActiveOrg` finds no `sys_member` row). It can also arise later (member removed from their org, etc.), so it cannot be defined away by fixing signup alone.

## Premises (re-check before ruling)

- The signup ordering is still live: `git log --oneline -5 -- packages/plugin-auth/src` and re-read `AuthManager.composeDatabaseHooks` / the ADR-0093 reconciler for a synchronous-settle change.
- #8158's reject-direction fix is merged: check PR #8237 state.
- #8208 / #8245 are still open findings with no disposition: re-read both cards.

## The question

Is "authenticated, but no active organization" a **legal session state** with defined semantics, or an **illegal intermediate** to be eliminated at the mint point?

## Options

[...]

Issue #8245, opened 2026-08-12
audit: a user's FIRST session predates their membership, so every audit row written in that window carries a NULL tenant and is invisible to RLS readers


os-zhuang:
Observation-class finding, measured while implementing #8144 (PR #8244). **Not caused by that change** — it is a pre-existing ordering property that #8144 made visible because it put a row on the ledger at exactly that moment.

## What was measured

`databaseHooks.session.create.before` derives a session's `activeOrganizationId` from the caller's `sys_member` row (ADR-0081 D1, `AuthManager.composeDatabaseHooks` → `defaultActiveOrg`). The membership itself is written by the ADR-0093 reconciler composed into `user.create.after`, and better-auth **defers that past the signup transaction** — `membership-actor-attribution.dogfood.test.ts` already polls up to 10s for it, and calls the delay out in a comment.

So the session minted by sign-up is created **before** the user has any membership. It carries `activeOrganizationId: null`, legitimately: at that instant there is no organization to name.

Reproduced while writing #8144's dogfood test: signing up a member and immediately reading the `login` audit row for that first session yields `tenant_id: null` (and `organization_id: null`). Waiting for the `sys_member` row before signing in produces a correctly tenanted row on every subsequent session. The test now measures the tenant half on an ordinary sign-in and documents the first-session case rather than asserting on it.

## Why it is worth a card

The SecurityPlugin's RLS predicate gates on `organization_id`. `audit-writers.ts` already names the consequence in its own comment, for the sibling cause:

> Without this fallback, audit rows are written with `organization_id=NULL` and the SecurityPlugin's RLS predicate (`organization_id = current_user.organization_id`) hides them forever — making the audit log UI appear permanently empty even though writes succeed.

Its fallback reads the audited record's own `organization_id`, which cannot help here: a session row has no organization column either, and the account/user rows being written in that same window are in the same state.

The window is narrow (a user's first session) but the loss is permanent — nothing back-fills a written ledger row — and the affected rows are exactly the ones describing account creation.

## Related, not duplicate

- #8208 — a record created with no active organization is invisible to its own creator (same RLS consequence, different producer: an HTTP write rather than the signup ordering).
- #8158 — `adminOrgScope` falls open when a session has no active organization (same "no active org" precondition, opposite failure direction).

This card is about the **ordering** that produces the no-active-org state for every new user, and about ledger rows specifically.

## Not proposed here

[...]

Issue #8158, opened 2026-08-12
plugin-sharing: a `manage_sharing` holder whose session has no ACTIVE organization reads every tenant's sharing rules (adminOrgScope falls open)


os-zhuang:
Found while implementing #7795 (platform-global delete gate). **Out of scope there and deliberately not fixed in that PR** — #7795's ruling holds the read surface open and untouched, and this is a different axis: it is about the ABSENCE of an org on the context, not about the `organization_id = null` row class.

Filed unassigned.

## The shape

`SharingRuleService.adminOrgScope` falls open when the caller has no org id:

```ts
private adminOrgScope(where: Record(string, unknown), orgId: string | null | undefined) {
  if (!orgId) return where;   // ← unscoped: every tenant's rows
  return { ...where, $or: [{ organization_id: orgId }, { organization_id: null }] };
}
```

That branch exists for `SYSTEM_CTX` (boot seeding, hooks, backfills), which legitimately reads unfiltered. But it is reached on **capability**, not on system-ness, and `assertCanManageRules` admits any caller holding org-scoped `manage_sharing`:

```ts
const orgId = (context as any)?.organizationId ?? context?.tenantId;   // may be undefined
const rows = await this.engine.find('sys_sharing_rule', {
  where: this.adminOrgScope(where, orgId), ... context: SYSTEM_CTX,
});
```

So an authenticated, non-system caller arriving with neither `organizationId` nor `tenantId` gets the system-context read scope.

## Why that context is reachable

`resolveAuthzContext` (`packages/core/src/security/resolve-authz-context.ts`) derives the tenant from the session's active organization and only stamps it when truthy:

```ts
tenantId = tenantId ?? sessionData?.session?.activeOrganizationId;
...
if (tenantId) ctx.tenantId = tenantId;
```

A better-auth session with no `activeOrganizationId` — a user who has not selected an organization, or whose active org was cleared — therefore produces an authenticated context carrying neither field. Nothing on the sharing-rule path re-derives one.

## Impact

For a caller holding only the org-scoped `manage_sharing` capability, `listRules` returns **every organization's** sharing rules, and `getRule` resolves any of them by id or name. `evaluateRule` then reaches those rows too, which is a cross-tenant **write** (it reconciles `sys_record_share` grants). This is the same class #7761 closed for the by-id branch — an org-level capability reaching rows outside the caller's tenant — arriving through a different door: there the filter was missing, here it is skipped because the org id is absent.

[...]

PR #8471, opened 2026-08-13
ADR-0123 + enforcement: an authenticated session with no active organization — reads resolve to nothing, tenant-scoped writes are refused loudly

### LINKED ISSUE #8208: A record an admin creates over HTTP is immediately invisible to its own creator when the caller has no active organization and the Layer 0 wall is active (`organization_id` stamped NULL)
Found incidentally while building #8119's dogfood control — it is what made a "grant a share on a local record" control fail with 404. Filed unassigned. **Scope caveat up front: I have not established whether this reproduces on a real deployment** (see "What I did not check"), which is exactly why this is a finding rather than a bug report.

## Measured, on a booted showcase stack

Boot: `bootStack(showcaseStack, { multiTenant: 'posture-only' })`. The boot log reports the wall is on:

```
INFO [security] tenancy posture 'isolated' — the Layer 0 organization wall is ACTIVE
                (organization_id = active organization)
```

The seeded platform admin (`stack.signIn()`) resolves an authz context carrying **no `tenantId` at all** — its keys are `positions, permissions, systemPermissions, org_user_ids, accessible_org_ids, accessToken, email, userId, posture`.

That admin then creates a private record through the real HTTP path and reads it back:

```
POST /api/v1/data/showcase_private_note      -> 2xx, row created
  stored row: { owner_id: 'PK07N5…', created_by: 'PK07N5…',   ← the admin, both
                organization_id: null, owning_business_unit_id: null }

GET  /api/v1/data/showcase_private_note      -> 200 {"records":[],"total":0}
GET  /api/v1/data/showcase_private_note/:id  -> 404
```

The row exists (a system-context `find` returns it in full). Its creator owns it. And its creator cannot see it through either read surface.

## Why it happens, as far as I traced it

The write stamps `organization_id` from the caller's active organization, which is absent, so the row is born org-less. The read then applies the Layer 0 wall, and "no active organization" is deliberately **not** "every organization" — the fail-closed rule stated in `activeOrganizationId`'s docblock and ADR-0095 D1 / ADR-0105 D1. Both halves are individually correct and defensible. Together they produce a write that succeeds and a record that is unreachable by the person who just made it.

The asymmetry is the interesting part: the **write** path treats a missing active org as "stamp NULL and proceed", while the **read** path treats the same missing value as "deny". If the write is allowed to proceed without an organization, something has to be able to read the result back.

## What I did not check

- Whether a real (non-harness) deployment ever puts an admin in this state. A seeded admin with no default organization membership may be specific to the verify harness's boot, in which case this is a test-fixture sharpness issue and not a product defect. **Checking that is the first triage step.**
- Whether the same happens under `single` / `group` postures, or only `isolated`.
- Whether the write path should refuse, stamp a default, or the read path should admit org-less rows to their owner — three different fixes with different blast radii, which is why this carries no recommendation.

## Related

- #5494 (closed) — the adjacent shape from the other direction: automation `create_record` under `runAs:'system'` inserting rows with `owner_id`/`organization_id`/`created_by` all NULL, "records born untouchable even by admin". Here `owner_id` and `created_by` ARE set correctly and only the org is null, so it is not the same defect, but it is the same failure family.
- #8158 — the other open card about a session with no active organization, from the opposite failure direction (falling open rather than closed).

---
_Generated by [Claude Code](https://claude.ai/code/session_01PEVB6w7D7uCszR9Mw1BL73)_

os-zhuang:
Fixes #8208
Part of #8247 — the **declared-semantics half** (ruling rules 1 and 3). The signup-ordering half (rule 2) is #8473, which declares #8245.

⛔ **This PR touches `docs/adr/**` — Prime Directive #14 reserves its landing to the maintainer.** Do not merge it, do not add it to the merge queue, do not arm auto-merge. It is deliberately left awaiting a human merge.

### ⚠️ Cross-lane touch, absorbed with PM authorisation — declared here, veto open

Two files in **`packages/qa/dogfood`** are `domain:cli`'s surface, not `domain:identity`'s. They are absorbed into this PR rather than handed to that seat, with the PM's explicit authorisation, because the only reason either file changes is this diff and a cross-seat round-trip buys lane purity on a PR that is maintainer-gated anyway:

- `test/federated-phantom-share-grant.dogfood.test.ts` — **setup repair** (`beforeAll` only). Its control-note write is a caller-path tenant-scoped write under an active wall, i.e. exactly the state D2 now refuses.
- `test/no-active-organization-write-refusal.dogfood.test.ts` — **new**, and the reason the repair is safe to make (below).

If `domain:cli` would rather own either file, say so and I will split them out — nothing else in this PR depends on them.

## The ruling this implements

Maintainer ruling on #8247, 2026-08-13 — **Option C**: the state is legal, with declared fail-closed semantics, plus a targeted signup-ordering fix. This PR carries the ADR that records it and the enforcement of its write half.

## What was open

`security-plugin.ts` step 3.7 — the Layer 0 write-side wall — validated **supplied** `organization_id` values only. Its own comment says so:

> This validates SUPPLIED values only; it never fills an absent one.

That is the right guard for what it guards (a payload naming *another* tenant, ADR-0095 / ADR-0105 D5), and it leaves the opposite case wide open: a payload naming *no* tenant, written by a caller who *has* no tenant. Nothing filled it downstream either — auto-stamping lives in the enterprise `@objectstack/organizations` runtime and has nothing to stamp when the caller carries no active organization.

So the row landed with `organization_id` NULL, and the read wall then hid it from every reader **including the author who had just created it**. Both halves were individually defensible, which is exactly why it sat.

## The rule (ADR-0123)

Under an authenticated session with no active organization:

- tenant-scoped **reads resolve to nothing** — Layer 0's deny sentinel, unchanged, silent, HTTP 200;
- tenant-scoped **`insert` / `update` are refused loudly** — `PERMISSION_DENIED` / 403, with a message that **names the missing active organization**;
- ⛔ **no silent NULL stamping**, anywhere.

`delete` is deliberately absent and the ADR says so rather than leaving it as an omission: it places no row and decides no tenant, so its target is selected through the Layer 0 **row** wall, which already resolves to nothing.

**The code and status are #8237's, on purpose.** #8158's landed fix already chose *refuse, naming the missing organization* over *answer empty*. Reusing `PERMISSION_DENIED`/403 rather than registering a synonym is what makes ruling rule 3 — "#8158's landed fix conforms and stands" — a measured statement rather than an assertion, and it follows ADR-0112 D3 (a generic permission-class condition uses the standard catalog).

## The verdict is derived, never re-derived

The refusal asks `computeWriteTenantCheckFilter` — the same Layer 0 the read side computes — so every exemption stays in one place and cannot drift: a `tenancy.enabled:false` platform-global object, an object with no `organization_id` column, a federated phantom anchor (#7835), a true `PLATFORM_ADMIN` on a posture-permitting object (ADR-0095 D3), and the whole `single` posture all yield `null` and are untouched. System / boot writes never reach the line (`isSystem` short-circuits the middleware). Under `group`, "no organization scope" means an **empty membership set** — ADR-0105 D2's own fail-closed rule, not a missing `tenantId`.

A cheap posture/context pre-test runs first so the common deployment (`single`) and the common caller (has an organization) never pay for the layered compile.

**Ordered ahead of the forge guard, deliberately.** A caller with no organization scope who *also* supplies a foreign `organization_id` satisfies both guards, and both answer 403 `PERMISSION_DENIED`. "You have no active organization" is the actionable half — the supplied value is not why the write cannot land, since no value could have satisfied the wall.

## #8208 is closed END TO END, over real HTTP — and how that got measured

When this PR first ran, the HTTP-level proof arrived as a **sibling dogfood fixture going red on its own setup step**:

```
FAIL |isolated| test/federated-phantom-share-grant.dogfood.test.ts
AssertionError: admin creates the control note: expected 403 to be less than 300
```

That `beforeAll` is #8208's scenario verbatim — `posture-only` boot (wall ACTIVE), the seeded harness admin whose resolved context carries no `tenantId`, `POST /data/showcase_private_note` through the real HTTP path. `showcase_private_note` declares `sharingModel: 'private'`, the row-ownership OWD and **not** `access.default: 'private'`, so it is an ordinary business tenant object and the ADR-0095 D3 platform-operator exemption correctly does not apply.

**Attribution was verified, not assumed.** A 403 alone cannot separate this refusal from the object's own CRUD gate, which answers `403 PERMISSION_DENIED` on the same route — and grepping the CI log for the refusal's own marker returned **0 hits**. The discriminator: **#8473 shares this PR's base and carries no `plugin-security` change, and its `Dogfood Regression Gate (3/3)` is success.** So the red was the refusal.

**The measurement is now a permanent assertion, and it was committed BEFORE the fixture was repaired** — a repaired fixture is green again and carries no record of what it was red for. `no-active-organization-write-refusal.dogfood.test.ts`, 6 cases, all green:

- PRECONDITION — `tenancy` reports `posture: 'isolated'`, `isolationActive: true`, `degraded: false` (a degraded boot resolves to `single`, where Layer 0 contributes nothing and the file would be vacuous while green);
- PRECONDITION — the admin's resolved `tenantId` is `null`, #8208's own fact re-measured;
- **the refusal** — 403, `code: 'PERMISSION_DENIED'`, message naming the missing active organization;
- **nothing landed** — a system-context read, which no wall narrows, finds zero rows (the defect was a write that *succeeded* and then hid; a refusal that still landed the row would reproduce it while looking fixed);
- **the control** — the same caller, object and route SUCCEED once a `sys_member` row exists and a fresh session is minted. One fact differs between refusal and success: the organization.

That upgrades `Fixes #8208` from enforcement-layer evidence to end-to-end evidence. Recorded on #8208 itself too.

**Why the fixture repair is a bind, not an assertion change.** The alternative — expecting 403 there — would have deleted #8119's own control. Binding the admin restores the setup's intent (a record created through the caller's own HTTP path). It is done by hand because `bootStack`'s `orgContext` option **refuses** to compose with `multiTenant`: under a walled posture it would be a no-op that reads like a feature, the #7762 vacuity class. And #8473's ordering fix cannot do it either — `tenancy.defaultOrgId()` returns `null` under **any** walled posture by design, so nothing auto-binds under `posture-only`.

**One adjacent property, noted and NOT closed here:** under `posture-only` an *org-bound* caller's write still lands `organization_id: null`, because auto-stamping lives in the enterprise runtime. Pre-existing, deliberate, documented at the stamping site ("a broken deployment rather than a working unlicensed one"), and out of scope — recorded because it is visible in the control case above.

## Tests — `no-active-organization-write-refusal.test.ts`, 19 unit cases

Built around the two vacuity traps this card carries.

**Trap 1 — "reads resolve to nothing" and the bug look identical from outside.** No read-side case asserts "empty". They take **one concrete row** and assert the *same* row is admitted under a system context and excluded under the org-less caller's own composed filter, and that the excluding filter is the **deny sentinel specifically** — not an absent policy (`undefined`, unrestricted) and not an ordinary wall (`organization_id = ...`). A row that was never written would be admitted by neither, so the system-context leg is what makes empty-by-rule and empty-by-accident different verdicts. A third leg pins that one membership turns the sentinel into a real wall that **admits** the row, ruling out "denied to everyone but system".

**Trap 2 — a status-only 4xx assertion cannot tell a conforming refusal from any other 4xx.** Every write case pins `code`, `statusCode`, **and** message content (naming the missing active organization, and *not* the forge guard's sentence), never a bare `toThrow()`. The sharpest case is the ordering one, where both guards fire and only the message tells them apart.

Every exemption is paired with a sibling differing in one fact where the refusal **does** fire — including the one that matters most: the same platform operator, same superuser bits, same platform capabilities, **is** refused on an ordinary business tenant object, because ADR-0095 D3's exemption is posture-scoped and never a property of the person (the W2 hole).

### Ablation (predictions written before the mutation, then measured)

Restoring `security-plugin.ts` to `origin/main` and re-running: **6 red, 13 green**, exactly the write-refusal pins and nothing else. The first failure reads `the write was not refused at all: expected undefined to be defined` — the pre-fix build letting the org-less insert through. Every read-side pin, every exemption control, and the existing forge-guard pins stayed green.

Suites: `plugin-security` **52 files / 1052 tests green**. Dogfood: the new pin **6/6**, the repaired fixture **18/18**, and every *other* walled-posture dogfood file (`attachments-permission-matrix`, `federated-anchor-provenance`, `federated-rls-injectors`, `org-create-default-team`, `rls-multitenant`, `showcase-external-autoconnect`) **green** — a scoped sweep by the change's real consumption radius, confirming `federated-phantom-share-grant` was the only casualty.

## Why two PRs, and why this one waits

Ruling rule 2 (signup ordering) is #8473, which can land through the normal flow. The split is not stylistic — it is forced, and measurably so:

- `check:adr-anchors` requires **every `ADR-NNNN` cited in a tracked file to name a real record under `docs/adr/`**. Prime Directive #13 requires code realizing an ADR's decision to carry its id. So rule 1's enforcement, which cites ADR-0123, **cannot** be separated from the ADR that defines it — CI would reject it.
- #8473 has no such coupling. It stands on **already-accepted** ADR-0093, cites no unmerged record, and its disposition of #8245 stands regardless of how this ADR's text is finally worded. Confirmed empirically: `ADR maintainer approval` passes on #8473, so a `scripts/adr-anchors/**` registry file is not a governed ADR document — only `docs/adr/**` is (`ADR_PATH_PREFIX`).

The landing order is also the favourable one: #8473 landing first *reduces how often* this state occurs at signup, before this PR starts refusing writes in it.

`skip-changeset` is **not** applicable — this PR changes runtime behaviour and carries its own changeset, declared breaking with an ADR-0087 `not-required (no-migration-prescription)` disposition (gate green).

## Files

- `docs/adr/0123-no-active-organization-session-semantics.md` — the record (D1 legal state, D2 read/write asymmetry, D3 no NULL stamping + the ledger's ordering carve-out, D4 the refusal names what is missing), with the rejected options and their reasons.
- `packages/plugins/plugin-security/src/security-plugin.ts` — the refusal, plus two helpers (`callerHasOrganizationScope`, posture-shaped by ADR-0105 D2; `isTenantWallDenial`, which compares the sentinel by **value** because `computeTenantLayer0Filter` spreads it and a reference check would answer `false` for every real denial).
- `packages/plugins/plugin-security/src/no-active-organization-write-refusal.test.ts` — new, 19 unit cases.
- `packages/qa/dogfood/test/no-active-organization-write-refusal.dogfood.test.ts` — new, 6 HTTP cases *(cross-lane, declared above)*.
- `packages/qa/dogfood/test/federated-phantom-share-grant.dogfood.test.ts` — `beforeAll` setup repair *(cross-lane, declared above)*.
- `scripts/adr-anchors/packages__plugins__plugin-security__src__security-plugin.ts.json` — ADR-0123 added, with an invariant that teaches why the refusal cannot be folded into the forge guard.
- `.changeset/no-active-organization-write-refusal.md`.

## Dispositions

- **#8208** — declared above; closed end to end, evidence recorded on the card.
- **#8245** — #8473's, not this PR's.
- **#8158 / PR #8237** — conforms and stands, untouched.
- **#8247** — `Part of` only. Neither PR closes the decision card alone; it closes when both have landed.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PEVB6w7D7uCszR9Mw1BL73)_


os-zhuang:
## ADR-0123 D3 scope clarification — `5c36d7591`

Added ahead of maintainer review so the text under review states its own scope accurately. **This narrows the ADR's wording, not the ruling** — reasoning below, because that distinction is the whole question.

### What was wrong with my draft

The #8247 ruling reads:

> under an authenticated session with `activeOrganizationId == null`, tenant-scoped **reads resolve to nothing** and tenant-scoped **writes are refused loudly** (4xx naming the missing active organization). ⛔ No silent NULL stamping anywhere.

Its "anywhere" never leaves that antecedent — it scopes **which layers the rule binds** (no subsystem gets an exemption), which is exactly the point of a card filed because three subsystems each improvised.

My D3 opening sentence dropped the antecedent:

> ~~No path may write a tenant-scoped row with a NULL tenant on behalf of an authenticated caller and report success.~~

That reads as covering **any** authenticated caller, org-bound ones included — a promise the ruling never made and the open packages cannot keep. So this is a correction of my own over-reach, not a trim of the maintainer's decision.

### The measurement behind it

Under `multiTenant: 'posture-only'` — walled posture, no enterprise organizations runtime — an **org-bound** caller's write still lands `organization_id: null`. Auto-stamping belongs to that runtime, which is also what *activates* every walled posture, so a real walled deployment has the stamper by construction. Visible in this PR's own dogfood control case, where the control row lands null-org even though its creator holds an organization.

Pre-existing and deliberate; the stamping site in `security-plugin.ts` already gives the reason — keeping the stamp there means a forged `org-scoping` registration "yields NULL-org rows that the wall hides, i.e. a broken deployment rather than a working unlicensed one".

### What changed

`docs/adr/0123-*.md` only — two places, no code:

- **D3's opening** now carries the ruling's antecedent explicitly, and keeps its force verbatim: *"…no path may write a tenant-scoped row with a NULL tenant and report success, **in any subsystem**; D2's refusal is what makes that hold, and **no layer gets an exemption from it**."*
- **A scope paragraph** states that the guarantee is about the *caller's missing organization*, not a promise that every tenant-scoped row is stamped; that D2 does not reach the org-bound case **and is not meant to** (the caller has an organization, so nothing is missing for a refusal to name); and that the record "neither changes that property nor endorses it — it declines to restate a packaging decision as a security guarantee the open packages cannot keep alone."
- The TL;DR bullet was tightened the same way, so the summary and the decision cannot drift apart.

No decision is added, removed or reversed; D1, D2 and D4 are untouched; no code changed, so every test and gate result already reported stands.

⛔ **If you read this as changing the ruling's meaning rather than stating its scope, it should be reverted and asked as a question instead** — that call is yours, not mine, and `git revert 5c36d7591` restores the broader wording exactly.

`check:adr-anchors` OK (48 anchored files, 23637 citations resolve) · `check:nul-bytes` OK (7632 files) · no new links added.

---

hotlong (approved):

Issue #8501, opened 2026-08-13
A licensed `max_nodes` cap still cannot be enforced by any replica: the cluster has no membership view and no slot claim, so a count-carrying gate verdict stays advisory


os-zhuang:
Filed **unassigned, observation-class** from the `domain:services` dev seat while implementing #8367 (session `session_01ARidKDYSCD56LaygrvDPnk`). Not triaged or routed by me.

Blocked-by: #8367 (the seam widening this would consume; it is necessary but not sufficient on its own)

## Fact (measured on `origin/main` @ `ff1e9b6a9`)

The maintainer ruled 2026-08-13 (recorded on `objectstack-ai/cloud#1275`) that a licensed `max_nodes` overflow must **refuse the excess replicas, run up to the paid limit, and warn loudly** — explicitly not a whole-cluster degrade. #8367 makes that verdict *expressible* at the `MultiNodeGate` seam. It does **not** make it *enforceable*, and no consumer-side change can, because of three measured properties of the surrounding mechanism:

1. **The gate is consulted once per process, at boot, by each replica independently.** Sole call site: `packages/cli/src/commands/serve.ts:1234-1247`, inside the `OS_CLUSTER_DRIVER` branch of `os serve`.
2. **There is no cluster membership view.** `nodeId` is generated randomly per process — `generateNodeId()` at `packages/services/service-cluster/src/cluster.ts:123`, used via `parsed.nodeId ?? generateNodeId()` at `cluster.ts:60`. Nothing registers a node, tracks liveness, or counts live members.
3. **There is no ordinal.** The only count available at the call site is `OS_CLUSTER_REPLICAS` (`packages/services/service-cluster/src/split-brain-guard.ts:33-34,47-51`) — an operator-*declared* desired count, **identical in every replica**, not a live membership count.

## Why that leaves the cap advisory

With a cap of 3 and 5 replicas booting, every replica calls the gate with the same input and computes the **same** verdict ("3 admitted"). None of them can determine whether *it* is one of the admitted 3 or one of the excess 2. The only outcomes reachable from that verdict locally are:

- every replica joins — nothing is actually refused, the cap remains advisory (status quo), or
- every replica refuses — whole-cluster degrade, precisely the outcome the ruling rejects.

So "run N, refuse N+1" cannot be produced by any per-replica decision over a count alone.

## What would make it binding

[...]

Issue #8299, opened 2026-08-13
ADR-0087's changeset disposition has no category for published runtime TS interfaces with no metadata surface — the #8277 exemption argument is correct, unverifiable, and will be re-litigated


hotlong:
Filed by the triage seat from the `domain:services` seat's round-12 hand-off (seat post #6015, comment of 2026-08-13 03:51Z), which routed the observation here and left the filing decision to triage. Dedup: searched ADR-0087 + changeset + exemption; zero open hits.

## The gap

ADR-0087's changeset disposition covers artefacts with a metadata surface — things `objectstack migrate meta` can reach, with a Zod schema, a spec declaration, or a stored representation. PR #8277 hit a shape outside that taxonomy: **a published runtime TypeScript interface with no metadata surface at all.**

The PR argued its exemption from the migration-note requirement essentially correctly:

- no Zod schema, no spec declaration, no stored representation ⇒ `objectstack migrate meta` has nothing to rewrite;
- the channel that actually reaches affected consumers is the **compiler** (measured: TS2339 on the removed member).

The argument was accepted in review. But it was **prose, judged by hand** — there is no named category in the disposition it could cite, and nothing mechanical verifies its two premises (no schema / no stored form). The next author who touches a published runtime interface faces the same wall and either re-derives the argument, or (worse) pattern-matches "no metadata ⇒ no changeset discipline at all", which is not what #8277 argued.

## What "fixed" looks like

A named category in the ADR-0087 disposition (e.g. *runtime-interface-only: compiler-carried, migration-exempt*) with a mechanically checkable predicate — "the touched exported symbol appears in no `*.zod.ts`, no spec `contracts/**` entry, and no object definition" is grep-shaped and could live in the changeset gate rather than in reviewer judgment. The #8277 review thread is the worked example to encode.

⚠️ Scope note: the disposition text lives with ADR-0087, so the eventual PR is ADR-class (maintainer-merged). The checkable-predicate half is gate tooling. If promoted, decide then whether to split along that boundary — same shape as other ADR+gate pairs.

## Provenance

#8277 (the worked example) · #8131 (the card it delivered) · ADR-0087 registries.

hotlong:
Concentrated findings round (maintainer-directed, 2026-08-13): **promoted `finding` → `pm:queue`** (`domain:spec-tooling` unchanged — the disposition taxonomy and its gate predicate are the machine face of the spec contract). Type: Task.

[...]

PR #8503, opened 2026-08-13
feat(service-cluster): carry an admitted node count through the multi-node gate (#8367)

### LINKED ISSUE #8367: MultiNodeGate is count-blind: `allowMultiNode()` returns a boolean verdict, so a license's `max_nodes` cap can only deny the whole cluster, never the excess replicas
Filed **unassigned and unlabeled for triage** from the repo:cloud seat (session `session_01TPHe3S9uZYC9wtUwdkRZ46`) while accepting cloud#1278 (ADR-0022 Phase 1 runtime online client, PR cloud#1291). I am not this repo's triage and do not grade or route here.

## Fact (measured against `62b6a2fb`, the SHA cloud pins today)

`registerMultiNodeGate` consumes `{ allowMultiNode(): { allowed, reason } }` — a boolean verdict with **no node count in the contract**. The EE license layer (`@objectstack/security-enterprise`) registers into this seam.

## Why it matters now

The maintainer ruled on 2026-08-13 (recorded on cloud#1275) that a licensed `max_nodes` overflow must **refuse the excess replicas — run up to the paid limit — plus a loud warning**, explicitly NOT whole-cluster degrade. cloud#1291 shipped everything expressible from the consumer side: the per-replica verdict (`checkNodeCap`), the loud warning, and node-count telemetry to the control plane. But the refusal cannot be made **binding** through a boolean gate: denying `allowMultiNode` is precisely the whole-cluster degrade the ruling rejected, so today the cap is advisory at the clustering seam.

## Suggested shape (not a decision)

Widen the seam to carry a count — e.g. `allowMultiNode(nodeCount: number)` returning an admitted count, or a per-node admission callback — so a license cap can admit N and refuse N+1..M. Cloud's consumer change afterwards is one line (cloud#1291's `checkNodeCap` already computes `{ allowed, refused }`).

Refs: cloud#1275 (the ruling), cloud#1278 / PR cloud#1291 (the consumer half, with the measurement in its report), `docs/adr/` ADR-0022 in the cloud repo (D4 multi-node gating).

os-zhuang:
Fixes #8367

The multi-node gate could not express the verdict the maintainer ruled for. This widens it so a license can admit N nodes and refuse the excess — and is explicit, in code and in the changeset, that the resulting counts are **advisory at this seam** rather than enforcement.

## Measured signature before the change

`packages/services/service-cluster/src/multi-node-gate.ts:22` on `origin/main` @ `ff1e9b6a9`:

```ts
allowMultiNode(): { allowed: boolean; reason?: string };
```

The card measured this against `62b6a2fb` (what cloud pins). Re-verified on today's `main` — unchanged, so the premise holds.

## Consumer enumeration

Complete, both repos:

| Site | Consumption |
| --- | --- |
| `packages/cli/src/commands/serve.ts:1234-1247` | the **only** runtime consumer. Calls `checkMultiNodeAllowed()` zero-arg, reads `.allowed` and `.reason` |
| `packages/services/service-cluster/src/index.ts:70-77` | re-export |
| `packages/services/service-cluster/src/multi-node-gate.test.ts` | tests |
| cloud `apps/objectos-ee/objectstack.config.ts:92-94` | the EE registration — a **zero-arg arrow** returning bare `{ allowed, reason }` from `resolveMultiNodeEntitlement` |

Two properties of the CLI site shaped the design: it passes **no count**, and it re-declares the function's type locally inside a dynamic-import cast rather than importing it — so a widening does not reach it implicitly and cannot break it implicitly either.

## Shape chosen: optional parameter, optional `admitted`, totalized result

```ts
interface MultiNodeVerdict { allowed: boolean; reason?: string; admitted?: number }
interface MultiNodeGate { allowMultiNode(requested?: number): MultiNodeVerdict }

function checkMultiNodeAllowed(requested?: number): {
  allowed: boolean; reason?: string; admitted?: number; refused: number; capped: boolean
}
```

- `requested` optional and `admitted` optional is what preserves the EE provider: a function of fewer parameters stays assignable, and a verdict with no `admitted` reads as "no cap" — it admits everything asked for rather than having a refusal invented for it.
- `refused` / `capped` are **totalized** (always present) so no consumer writes `?? 0` over a third-party gate's output. Normalization of non-finite, fractional and negative counts happens at the seam — contract-first, per AGENTS.md PD #12.
- `capped` marks only a **partial** refusal and stays `false` for an outright `allowed: false`, so the licensed-overflow case and the unlicensed case cannot be conflated.

**The per-node admission callback was measured and rejected.** A gate is a module singleton inside one replica's process, with no cross-process state. Each replica's provider would start its own counter at zero and admit itself, so the callback would look like per-replica enforcement while enforcing nothing. An honestly advisory verdict beats a silently vacuous callback.

## The "advisory at the seam" claim held — and it is the bigger finding

Verified, and it is stronger than the card stated: a count-carrying verdict is **necessary but not sufficient**, and no consumer-side change can close the gap. The gate is consulted once per process at boot, where a replica has no membership view (`nodeId` is random per process, `cluster.ts:123`; nothing tracks live nodes) and no ordinal (`OS_CLUSTER_REPLICAS` is a declared count, identical in every replica). With a cap of 3 and 5 replicas, all five compute the same verdict and none can tell whether it is one of the admitted 3 — leaving only "all join" (cap advisory) or "all refuse" (the whole-cluster degrade the ruling rejects).

Binding enforcement needs an atomic slot claim against the `ILock`/`ICounter`/`IKV` primitives this package already ships. That is a new runtime mechanism with its own design decisions, filed separately as #8501 (out of scope here, not addressed by this PR). This PR therefore documents the limitation in the module doc and the changeset instead of implying enforcement.

## Reverse verification

Predicted before running, then measured.

**Runtime** — tests added with the implementation byte-identical to `origin/main` (`git diff --quiet` confirmed): predicted RED, read **9 failed / 47 passed**. The diagnostics name the mechanism:

- `expected [ undefined ] to deeply equal [ 5 ]` — the count is dropped entirely, never reaching the gate.
- `expected { allowed: true, admitted: 5 } to deeply equal { allowed: true, admitted: 3, ...}` — the provider's raw object passes through unclamped, with no `refused`/`capped`.
- `expected { allowed: true, admitted: -2 } to match object { admitted: 0, refused: 3 }` — degenerate counts pass through unnormalized.

After the implementation: **56 passed (56)**, up from a 45-test baseline.

**Vacuity trap, named and closed.** An assertion reading only `allowed`/`reason`, or one whose requested count never exceeds the cap, passes verbatim against a completely unwidened tree — `refused` would be 0 either way. Every count pin therefore requests strictly more than the cap, and the partial-cap pins assert `allowed === true` in the **same** expectation, because a gate denying the whole cluster would also yield `refused > 0`. Measured proof this matters: 2 of the 11 new tests (the uncounted compat pins) were GREEN on `main` by construction, and are reported as non-discriminating rather than counted as evidence of the widening.

**Type level, against the rebuilt `.d.ts`** (`dist` regenerated first, so this is not a cached read): the EE provider's exact zero-arg shape, the count-aware follow-up shape, and today's zero-arg call site all compile clean. The rejection direction fires as predicted — a string `admitted`, a verdict missing `allowed`, a string count, and treating `refused` as optional are all refused, the errors naming `MultiNodeVerdict`.

**Consumer sweep, downstream direction** (`--filter '...@objectstack/service-cluster'` — the prefix form; 16 packages downstream): `@objectstack/cli`, `@objectstack/runtime` and `@objectstack/service-cluster-redis` typecheck clean **after** building the CLI's 55-package closure. Before that build they showed `Cannot find module` errors that read exactly like broken imports — a false red from the fresh worktree, not from this change.

## The consumer signature this leaves for the cloud follow-up

Mechanical, as promised. `resolveMultiNodeEntitlement` keeps its current shape; only the registration changes, and `maxNodes` is already on the entitlement (`packages/security-enterprise/src/license/entitlement.ts:173,258`):

```ts
registerMultiNodeGate({
  allowMultiNode: (requested) => {
    const ent = resolveMultiNodeEntitlement(process.env.OS_LICENSE_KEY);
    if (!ent.allowed) return ent;
    const cap = getActiveEntitlement()?.maxNodes;
    return cap === undefined ? ent : { ...ent, admitted: Math.min(requested ?? cap, cap) };
  },
});
```

Not doing so is safe: the current zero-arg registration keeps working unchanged and reads as uncapped.

## Changeset and ADR-0087

`.changeset/multinode-gate-admitted-count.md`, `minor` — additive public API on `@objectstack/service-cluster`.

ADR-0087 reading, stated rather than guessed: it governs **metadata protocol** surface (spec properties, `engines.protocol`, conversion/deprecation registries) and contains no changeset disposition at all. This change widens a TypeScript interface in a service package with **no metadata or authorable surface** — no spec property, no Zod schema, no protocol version — so no ADR-0087 registration applies. `check-adr-0087-registration` agrees on the real diff. The category gap itself is tracked in #8299 and is untouched here.

## Gates

Path-derived via `node scripts/pm/dispatch-gates.mjs` against the actual changed paths, plus the convention-scoped ones: `check:changeset-gate-self-tests`, `check:docs-audit-scope`, `check:objectui-changeset`, `check:test-source-alias`, `check:type-source-resolution`, `check-adr-0087-registration`, `check-changeset-fixed`, `check-changeset-no-major`, `check-empty-changeset`, `check:query-options-erasure`, `check:type-check-coverage`, `check:nul-bytes`, `check:error-code-casing` — all exit 0.

One pre-existing red, not from this change: `check:objectui-pin-fresh` exits 1. It reads only `.objectui-sha`, which this PR does not touch; it entered the derived list solely because the diff touches `.changeset/`, and its own output cites #3340 as its tracking issue.

---

### DIFF
--- a/.changeset/multinode-gate-admitted-count.md
+++ b/.changeset/multinode-gate-admitted-count.md
@@ -1,0 +1,46 @@
+---
+"@objectstack/service-cluster": minor
+---
+
+feat(service-cluster): the multi-node gate can carry an admitted node count, so a license cap can refuse the excess replicas instead of the whole cluster (#8367)
+
+`registerMultiNodeGate` consumed `{ allowMultiNode(): { allowed, reason } }` — a
+bare boolean verdict with **no node count in the contract**. The maintainer ruled
+on 2026-08-13 (recorded on `objectstack-ai/cloud#1275`) that a licensed
+`max_nodes` overflow must **refuse the excess replicas, run up to the paid limit,
+and warn loudly** — explicitly *not* a whole-cluster degrade. Through a boolean
+gate that verdict could not be stated at all: the only refusal a license could
+express was `allowed: false`, which is precisely the whole-cluster degrade the
+ruling rejects.
+
+A gate verdict may now carry `admitted` — how many nodes it admits — and
+`checkMultiNodeAllowed(requested?)` forwards the caller's intended node count to
+the gate and returns a normalized verdict:
+
+```ts
+{ allowed: boolean; reason?: string; admitted?: number; refused: number; capped: boolean }
+```
+
+`refused` and `capped` are **totalized** (always present), so no consumer writes
+`?? 0` over a third-party gate's output — the seam normalizes non-finite,
+fractional and negative counts itself. `capped` marks only a **partial** refusal:
+it stays `false` for an outright `allowed: false`, so the licensed-overflow case
+and the unlicensed case cannot be conflated by a consumer.
+
+**Backward compatible.** `requested` is an optional parameter and `admitted` an
+optional return field, so an existing zero-arg, boolean-shaped provider — the
+shape `@objectstack/security-enterprise` registers today — remains valid and is
+interpreted as "no count-based cap": it admits everything requested rather than
+having a refusal invented for it. Existing zero-arg call sites are unaffected.
+
+**⚠️ The count is advisory at this seam — it is not yet enforcement.** The gate
+is consulted once per process, at boot, by each replica independently, and at
+that moment a replica has no cluster membership view (`nodeId` is random per
+process; nothing tracks live nodes) and no ordinal — the only count available is
+the operator-declared `OS_CLUSTER_REPLICAS`, identical in every replica. So every
+replica computes the same verdict and none can tell whether *it* is one of the
+admitted N or one of the excess. Binding enforcement additionally requires an
+atomic slot claim against the shared cluster primitives this package already
+ships (`ILock`/`ICounter`/`IKV`), which is tracked separately. Until then,
+consumers should treat `refused > 0` as the trigger for the loud warning the
+ruling requires, never as grounds to deny the cluster.
--- a/packages/services/service-cluster/src/index.ts
+++ b/packages/services/service-cluster/src/index.ts
@@ -72,4 +72,6 @@ export {
     checkMultiNodeAllowed,
     __resetMultiNodeGate,
     type MultiNodeGate,
+    type MultiNodeVerdict,
+    type ResolvedMultiNodeVerdict,
 } from './multi-node-gate.js';
--- a/packages/services/service-cluster/src/multi-node-gate.test.ts
+++ b/packages/services/service-cluster/src/multi-node-gate.test.ts
@@ -10,12 +10,18 @@ afterEach(() => __resetMultiNodeGate());
 
 describe('multi-node gate', () => {
     it('allows when no gate is registered (open framework)', () => {
-        expect(checkMultiNodeAllowed()).toEqual({ allowed: true });
+        expect(checkMultiNodeAllowed()).toEqual({ allowed: true, refused: 0, capped: false });
     });
 
     it('honors a denying gate with reason', () => {
         registerMultiNodeGate({ allowMultiNode: () => ({ allowed: false, reason: 'unlicensed' }) });
-        expect(checkMultiNodeAllowed()).toEqual({ allowed: false, reason: 'unlicensed' });
+        expect(checkMultiNodeAllowed()).toEqual({
+            allowed: false,
+            reason: 'unlicensed',
+            admitted: 0,
+            refused: 0,
+            capped: false,
+        });
     });
 
     it('honors an allowing gate', () => {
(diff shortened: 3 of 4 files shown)


os-zhuang:
**PM review — `domain:services` seat #6021, session `session_01ARidKDYSCD56LaygrvDPnk`. Verdict: ACCEPT**, flip + arm pending CI (23 checks still running at time of writing; ⛔ nothing enqueued until every one concludes `success`).

## ⭐ The dispatch's named failure mode was hit, and reported instead of papered over

The dispatch order told this dev to watch for exactly one thing:

> "a widened contract that looks like it implements the ruling while the excess replicas still start."

That is what was found. The gate is consulted once per process at boot; `nodeId` is random per process (`cluster.ts:123`), nothing tracks membership, and `OS_CLUSTER_REPLICAS` is an operator-*declared* count identical in every replica. So with a cap of 3 and 5 replicas, all five compute the same verdict and **none can tell whether it is one of the admitted 3**. Only "all join" (advisory) or "all refuse" (the degrade the ruling rejects) are reachable locally.

⚠️ **This falsifies the card's own load-bearing premise.** #8367 states the widening is what makes the refusal binding, and triage's routing repeated it ("this card is the seam-widening that makes it enforceable"). It is not — it is necessary but not sufficient, and **no consumer-side change can close the gap**. Had this shipped quietly, the repo would carry a count-carrying gate, a satisfied-looking card, a closed ruling, and five replicas still starting under a cap of three.

⭐ Equally right: the PR **documents the limitation in the module doc and changeset instead of implying enforcement**, and files the real mechanism as #8501 rather than smuggling a new runtime subsystem (slot claims, TTL, refusal posture) into a signature-change card. That is the correct scope discipline.

## On `Fixes #8367` — deliberate, and I am ruling it correct

⚠️ I want this on the record because it is the #8131 shape and I got that one wrong. The argument for `Part of` is real: the card's stated goal is bindingness, and this PR does not deliver bindingness.

I am letting `Fixes` stand. The card's **named deliverable** — the count-blind seam — is fully delivered, and its *motivation* rested on a belief measurement has now falsified. Holding the card open would leave an anchor whose entire remaining scope is #8501's, i.e. a duplicate. ⛔ But the closure must be loud, not quiet: the post-merge comment on #8367 will state plainly that the 2026-08-13 ruling is **not yet enforced** and point at #8501, so nobody auditing "is the ruling implemented?" reads a closed card as yes.

## Review points, all clearing

- **Backward compatibility** — the hard constraint from the dispatch. Verified the right way: against the **rebuilt** `.d.ts`, not a cached read, with the EE provider's exact zero-arg arrow shape compiling clean. A fewer-parameter function staying assignable is correct TS, and the rejection direction was tested too. ⭐ The reasoning that an absent `admitted` reads as "no cap" rather than having a refusal invented for it is the right default — the opposite would have made every existing provider start refusing.
- **Per-node callback rejected on measurement**, not taste: each replica's provider starts its own counter at zero in its own process, so the callback would enforce nothing while looking like enforcement. Rejecting the card's own suggested alternative for a stated reason is what Zone 3 is for.
- **Vacuity trap closed honestly** — every count pin requests strictly more than the cap, and ⭐ 2 of the 11 new tests are reported as **GREEN on `main` by construction and non-discriminating** rather than counted as evidence. Self-reporting a weak pin is worth more than the pin.
- **The 16-package downstream sweep's false red** (`Cannot find module` before the CLI's 55-package closure was built) is correctly identified as a fresh-worktree artifact, not a break. ⚠️ This is the third local-red-that-CI-disagrees-with this shift; naming it as an artifact rather than escalating it is the right call and the one I got wrong earlier today.
- **ADR-0087 reading is stated, not guessed**, and lands where #8299 says the gap is. Correct.

## Docs drift advisory — measured, no action needed

The bot flagged 4 docs. I checked rather than forwarding it: **no doc anywhere under `content/docs` mentions `max_nodes`/`maxNodes` at all** (zero hits), and `content/docs/kernel/cluster.mdx`'s only claim is that "when it denies, `os serve` logs a warning and downgrades" — which stays exactly true, since an outright `allowed: false` is unchanged and `capped` is deliberately kept `false` for it.

⭐ So there is nothing misleading to correct here, and ⛔ no doc edit belongs in this PR. **Note for whoever lands #8501**: `cluster.mdx` is where the enforcement story will need to appear, because that doc currently describes only the deny/downgrade path and an operator has no other place to learn whether a node cap binds.

---

PR #8506, opened 2026-08-13
fix(plugins): retire the `i?.content ?? i` unwrap family from plugin read paths (#8378)

### LINKED ISSUE #8378: [finding] The `i?.content ?? i` unwrap family survives in plugin read paths (bootstrap-declared-* and friends) — same envelope presumption #7519 removed from MetadataFacade
Observation-class finding, filed unassigned by the #7519 dev (session `session_01RDTnVvsgA6cUZ4xFVtPZRy`) from that card's out-of-scope sweep. Dup search: zero hits beyond #7519 itself.

## Mechanism (verified at file:line on `origin/main` @ `a8940e401`)

#7519 / PR #8377 removed the `item?.content ?? item` unwrap from `MetadataFacade` after measuring that the `{ name, content }` storage envelope it presumed has **no producer** anywhere in the tree (the only writer that ever produced one was the facade's own interim boxing of non-object values, removed by #8349; DB hydration registers the parsed document itself). The identical unwrap survives in production plugin read paths over the same `SchemaRegistry.listItems` stores:

- `packages/plugins/plugin-security/src/bootstrap-declared-permissions.ts:65`
- `packages/plugins/plugin-security/src/bootstrap-declared-positions.ts:55`
- `packages/plugins/plugin-security/src/permission-set-projection.ts:307`
- `packages/plugins/plugin-sharing/src/bootstrap-declared-sharing-rules.ts:81`
- `packages/plugins/plugin-webhooks/src/bootstrap-declared-webhooks.ts:101` and `:110`
- `packages/plugins/plugin-email/src/bootstrap-declared-email-templates.ts:114` and `:123`
- `packages/plugins/plugin-email/src/email-plugin.ts:1171` and `:1231`

Also: `packages/mcp/src/mcp-server-runtime.ts:276` carries a docblock already reasoning the same unwrap is a no-op on its path — precedent that this family is being retired seam by seam rather than in one sweep.

## Severity notes — why observation-class rather than a concrete defect

The types these readers consume (`capability`, `permission`, `position`, `sharingRule`, `webhook`, `emailTemplate`, `permission_set`) do not declare a stored `content` field in their spec schemas, so the unwrap is dormant dead code on spec-valid data — the same "no producer" measurement #7519's fix rests on.

One nuance worth grading: `packages/spec/src/system/email-template.zod.ts:62` accepts `content` as an authorable **alias** spelling (`content` maps to `bodyHtml` in the conversion layer). Whether an item can reach `bootstrap-declared-email-templates.ts`'s read still carrying its pre-conversion `content` key was NOT measured here — if it can, the unwrap answers the HTML string instead of the template on that path, the exact #7519 defect shape, live at boot rather than dormant. That measurement is the first step of any fix.

Note the unwrap also corrupts a hypothetical valid item whose `content` is any truthy value — and `''` (falsy but non-nullish) also passes the `??` — so the guard is not merely dead, it is wrong whenever the key exists.

Test helpers replicating the pattern (pinning the production expression, so they move with it): `engine-capability-provenance.test.ts:55`, `engine-nested-plugin-collections.test.ts:86`/`:92`, `engine-nested-plugin-view-expansion.test.ts:103` in `packages/objectql/src`.

## Refs

#7519 (the facade half, PR #8377) · #8349 (removed the envelope's only producer) · #6725 (silent-loss family).

Filed unassigned — recording the finding, not claiming it.

---

os-zhuang:
Fixes #8378

Retires the `i?.content ?? i` unwrap family from the ten enumerated production read paths plus the three pinned test helpers, as one cross-domain sweep (triage's designation — the enumeration spans identity's `plugin-security` / `plugin-sharing` and services' `plugin-webhooks` / `plugin-email`, and it is one mechanical criterion).

## Step 1 — the email-template alias measurement, and what it falsified

The card's prescribed first step was: *can an item reach `bootstrap-declared-email-templates.ts`'s read still carrying its **pre-conversion** `content` key?*

**The question's premise is false — there is no conversion.** The card (and triage) read `email-template.zod.ts:62` as an authorable alias that "maps to `bodyHtml` in the conversion layer". Measured, it does not:

| claim | measurement |
|---|---|
| `content` maps to `bodyHtml` in the conversion layer | **false.** `packages/spec/src/conversions/registry.ts` has **zero** `email_template` entries. `normalizeStackInput` emits 0 notices and leaves the key untouched — independently re-confirmed in-tree at `metadata-protocol/src/protocol.ts`: *"no ADR-0087 conversion entry touches `email_template` — re-checked against `packages/spec/src/conversions/registry.ts`"* |
| ...so `content` is silently renamed before the read | **false.** `content` sits in `EmailTemplateDefinitionSchema`'s `strictObject({ aliases: … })` table, which feeds `strictUnknownKeyError`. That runs **only** on the `unrecognized_keys` path and **only** builds a message string. It never rewrites input. `content` is a **rejection** alias |

So the honest reformulation — *can an item carrying `content` reach that read?* — answers **no through any validating door**, and every door validates:

- `defineStack` strict (the default) parses through `ObjectStackDefinitionSchema` and throws;
- `saveMetaItem` (`PUT /meta`, Studio) resolves `email_template` to its schema and refuses with **422 `INVALID_METADATA`** before persisting;
- `loadMetaFromDb` registers `convertStoredItem(JSON.parse(record.metadata))` — the parsed body, never the `sys_metadata` row (whose body column is `metadata`, not `content`).

It survives only where validation is skipped by choice: `defineStack(…, { strict: false })`, a hand-built manifest, a direct registry write.

**But that is not "dormant", and this is the PR's headline.** On exactly that path the unwrap was *actively destructive*, in a way the card did not anticipate:

1. **It destroyed the author's own prescription.** The schema is built to answer this mistake — *Unrecognized key(s) on this email template: `content`. Did you mean `content` → `bodyHtml`?* The unwrap replaced the document with the body string, so `EmailTemplateDefinitionSchema.parse()` saw a string and answered `Invalid input: expected object, received string` — and the boot warning's `name` field came back `undefined`, so an operator could not even tell **which** template failed. The unwrap was the one thing standing between the author and their own fix.
2. **`content: ''` vanished outright.** Falsy but non-nullish, so it passed `??` and then died at the reader's own `filter(Boolean)`: no row, no warning, no count. The card predicted `''` "passes `??`"; it does, and then the item is silently dropped — worse than stated, and the ADR-0078 silent-loss shape.

## The card's dormancy assumption — verified per type, and extended

Verified mechanically against each bound schema rather than trusted:

| type | declares a stored `content`? | verdict |
|---|---|---|
| `permission` (PermissionSetSchema) | no — rejects as `unrecognized_keys` | dormant |
| `position` | no — rejects | dormant |
| `capability` | no — rejects | dormant |
| `sharing_rule` | no — rejects | dormant |
| `webhook` | no — rejects | dormant |
| `email_template` | no — rejects (alias ⇒ prescription) | the live-harm case above |
| **`object`** | no — rejects | dormant — **not on the card's list** |

**One addition to the card's enumeration of consumers.** `readDeclared` in `bootstrap-declared-permissions.ts` is exported and generic, and is called with three types, not one: `permission` (also from `suggested-audience-bindings.ts`), `capability` (`bootstrap-declared-capabilities.ts`), and **`object`** (`security-plugin.ts`, feeding `applyManagedWriteDenies`). `object` is absent from the card's six-type list. Measured: `ObjectSchema` declares no `content` and rejects it, so the assumption holds — but it held by luck, not by the card's reasoning.

### Production sites (10)

| 落点 | before | after |
|---|---|---|
| `plugin-security/src/bootstrap-declared-permissions.ts:65` | `(reg.listItems(type) ?? []).map((i) => i?.content ?? i).filter(Boolean)` | `(reg.listItems(type) ?? []).filter(Boolean)` |
| `plugin-security/src/bootstrap-declared-positions.ts:55` | same expression | `(reg.listItems(type) ?? []).filter(Boolean)` |
| `plugin-security/src/permission-set-projection.ts:307` | `for (const i of items) { const body = i?.content ?? i; … }` | `for (const body of items) { … }` |
| `plugin-sharing/src/bootstrap-declared-sharing-rules.ts:81` | same expression | `(reg.listItems(type) ?? []).filter(Boolean)` |
| `plugin-webhooks/src/bootstrap-declared-webhooks.ts:101` | registry read, mapped through the unwrap | `.filter(Boolean)` only |
| `plugin-webhooks/src/bootstrap-declared-webhooks.ts:110` | metadata-service fallback, same map | `arr.filter(Boolean)` |
| `plugin-email/src/bootstrap-declared-email-templates.ts:114` | registry read, mapped through the unwrap | `.filter(Boolean)` only |
| `plugin-email/src/bootstrap-declared-email-templates.ts:123` | metadata-service fallback, same map | `arr.filter(Boolean)` |
| `plugin-email/src/email-plugin.ts:1171` | `upsertDeclaredEmailTemplate(engine, (raw as any)?.content ?? raw, …)` | `upsertDeclaredEmailTemplate(engine, raw, …)` |
| `plugin-email/src/email-plugin.ts:1231` | `stripReadDecorations((item as any)?.content ?? item)` | `stripReadDecorations(item)` |

### Pinned test helpers (3 files, 4 sites)

| 落点 | before | after |
|---|---|---|
| `objectql/src/engine-capability-provenance.test.ts:55` | `readDeclaredShape` mirrored the unwrap | `.filter(Boolean)` only |
| `objectql/src/engine-nested-plugin-collections.test.ts:86` | `registeredNames` mirrored it | `.filter(Boolean)` only |
| `objectql/src/engine-nested-plugin-collections.test.ts:92` | `registeredItem` mirrored it | `.filter(Boolean)` only |
| `objectql/src/engine-nested-plugin-view-expansion.test.ts:103` | `viewItems` mirrored it | `.filter(Boolean)` only |

These four drive a **real** `ObjectQL`, so the unwrap was already a no-op for them — which is the measurement that retired it.

### Forced by the removal — the fixtures that manufactured the envelope (4 files, 6 sites)

Not on the card's list, and **inseparable**: these fakes boxed every declared item as `{ content: item }`, which made them **the only producers of that envelope anywhere in the tree**. That fiction is precisely what kept the production unwrap looking load-bearing. They now register documents the way the real engine does.

| 落点 | before | after |
|---|---|---|
| `plugin-email/src/bootstrap-declared-email-templates.test.ts:44` | `listItems: … .map((content) => ({ content }))` | returns the documents |
| `plugin-email/src/bootstrap-declared-email-templates.test.ts:264` | `list: () => [{ content: declaredTemplate() }]` | `list: () => [declaredTemplate()]` |
| `plugin-webhooks/src/bootstrap-declared-webhooks.test.ts:47` | same boxing | returns the documents |
| `plugin-security/src/bootstrap-declared-capabilities.test.ts:16` | `declared.map((c) => ({ content: c }))` | `[...declared]` |
| `plugin-security/src/bootstrap-declared-capabilities.test.ts:62` | inline `[{ content: {…} }]` | the document |
| `plugin-security/src/bootstrap-declared-positions.test.ts:23` | `declared.map((c) => ({ content: c }))` | `[...declared]` |

## Reverse verification — predicted before running, then measured

Two new pins in `bootstrap-declared-email-templates.test.ts`. **The vacuity trap closed:** both fixtures put a `content` key genuinely *in play*. A template that never spells `content` exercises only the unwrap's `?? i` arm and would pass against a completely unfixed tree — which is exactly what the pre-existing suite did, since its fake supplied the envelope itself.

Predicted **RED** on `origin/main` (tests added, implementation reverted), **GREEN** with the fix. Measured, with the implementation restored from `origin/main` and the tests kept:

```
 ❯ src/bootstrap-declared-email-templates.test.ts (17 tests | 2 failed)
   × reaches the schema as a DOCUMENT, so the rejection carries the `content` → `bodyHtml` fix
     AssertionError: expected undefined to be 'auth.welcome'
   × does not silently vanish when `content` is the empty string
     AssertionError: expected { seeded: +0, skipped: +0 } to deeply equal { seeded: +0, skipped: 1 }
 Test Files  1 failed | 22 passed (23)
      Tests  2 failed | 367 passed (369)
```

Both failures are the predicted direction and the predicted *mechanism*: the diagnostic could not name the template (`undefined`) because the parse was handed a bare string, and the empty-string template disappeared with `skipped: 0`. With the fix restored: **23 files / 369 tests passed**.

## Verification

| | result |
|---|---|
| `plugin-email` test | 23 files / 369 passed |
| `plugin-security` test | 54 files / 1060 passed (on the rebased base, incl. #8461's) |
| `plugin-sharing` test | 21 files / 569 passed |
| `plugin-webhooks` test | 5 files / 56 passed |
| `objectql` test | 197 files / 3539 passed |
| `typecheck` (all five) | clean |

Gates re-derived against the actual changed paths with `node scripts/pm/dispatch-gates.mjs` — the derivation surfaced families beyond the ones named at dispatch, all run: `check:nul-bytes`, `check:cross-package-test-inputs`, `check:test-source-alias`, `check:type-source-resolution`, `check:durability-log-level`, `check:engine-double-contract`, `check:docs-audit-scope`, `check:query-options-erasure`, `check:error-code-casing`, `check:i18n` (9 packages in sync), `check:type-check-coverage`, `check-engine-split-ratio.mjs` — **all pass**. `check-dev-prereqs.mjs` reports the workspace is not fully built (12 packages this branch never touches have no `dist/`); that is a local environment precondition, not a verdict on this change.

Rebased onto current `main`. **#8461 (#8323) has landed** and does touch `plugin-security` — re-checked as triage asked: no file overlap, no conflict, and `plugin-security`'s suite is green on the rebased base with its 23 new tests included.

## Scope

Zero changes outside the declared surface, other than the six fixture sites above, which the removal forces. No `content/docs/releases/` edit. Changeset added.

---

### DIFF
--- a/.changeset/retire-content-unwrap-family.md
+++ b/.changeset/retire-content-unwrap-family.md
@@ -1,0 +1,53 @@
+---
+"@objectstack/plugin-email": patch
+"@objectstack/plugin-security": patch
+"@objectstack/plugin-sharing": patch
+"@objectstack/plugin-webhooks": patch
+---
+
+fix(plugins): a declared item reaches its schema intact — retire the `i?.content ?? i` unwrap from plugin read paths (#8378)
+
+Ten production reads over `SchemaRegistry.listItems` unwrapped every declared
+item as `i?.content ?? i`, presuming a `{ name, content }` storage envelope.
+That envelope has **no producer**. Re-measured at these seams rather than
+inherited from #7519's measurement of `MetadataFacade`:
+
+- `registerMetadataCollections` (objectql) registers each stack-collection
+  element as-is — `registerItem(type, item, 'name')`, no boxing;
+- `loadMetaFromDb` registers `convertStoredItem(JSON.parse(record.metadata))` —
+  the parsed body, never the `sys_metadata` row (whose body column is
+  `metadata`, not `content`);
+- the facade's own interim boxing of non-object values, the one writer that ever
+  produced the shape, was removed by #8349.
+
+**Removal is a fix, not a cleanup.** None of the types read through these seams
+— `permission`, `position`, `capability`, `object`, `sharingRule`, `webhook`,
+`emailTemplate` — declares a stored `content` key; every one of them rejects it
+as an unrecognized key. So wherever the key did appear the unwrap replaced a
+whole authoring document with one of its values, and `''` — falsy but
+non-nullish — passed `??` and then died at the reader's own `filter(Boolean)`,
+dropping the item with no warning, no count and no row.
+
+**On email templates the harm was sharpest, and it is the one users will
+notice.** `content` really is a spelling an author can write there:
+`EmailTemplateDefinitionSchema` lists it in its `strictObject` **aliases** table
+(`content: 'bodyHtml'`). That table is a *rejection* facility, not a conversion —
+it feeds `strictUnknownKeyError`, which runs only on the `unrecognized_keys`
+path and only builds a message; nothing rewrites the key, and the ADR-0087
+conversion layer has no `email_template` entry either. The schema was therefore
+always ready with the author's fix, and the unwrap was the one thing standing
+between the author and it: the HTML string reached
+`EmailTemplateDefinitionSchema.parse()`, which answered `Invalid input: expected
+object, received string`, and the boot warning's `name` field came back
+`undefined` — so an operator could not even tell **which** template had failed.
+
+A template authored with `content` now yields what it was always meant to:
+
+> Unrecognized key(s) on this email template: `content`. Did you mean
+> `content` → `bodyHtml`?
+
+…named against the template it came from, and counted as `skipped` rather than
+vanishing.
+
+No behaviour changes for spec-valid metadata: the reads hand back exactly the
+documents they always did.
--- a/packages/objectql/src/engine-capability-provenance.test.ts
+++ b/packages/objectql/src/engine-capability-provenance.test.ts
@@ -24,8 +24,10 @@
  * Consequence before the fix: `plugin-security`'s `bootstrapDeclaredCapabilities`
  * resolves the owner as `cap._packageId ?? cap.packageId` and reads its input
  * with `readDeclared(ql, 'capability')`, which is
- * `engine.registry.listItems('capability')` mapped through `i?.content ?? i`
- * (`packages/plugins/plugin-security/src/bootstrap-declared-permissions.ts:61-69`).
+ * `engine.registry.listItems('capability')`, filtered
+ * (`packages/plugins/plugin-security/src/bootstrap-declared-permissions.ts`).
+ * That read used to map through `i?.content ?? i`; #8378 retired the unwrap
+ * across this family after measuring the envelope has no producer.
  * With `capabilities` outside the seam that list was ALWAYS empty, so the
  * `_packageId` half of that `??` could never be satisfied and the author-side
  * `packageId` — documented in `CapabilityDeclarationSchema` as the *fallback*
@@ -48,11 +50,15 @@ import { ObjectQL } from './engine';
 /**
  * The exact read `bootstrapDeclaredCapabilities` performs on the engine, kept
  * in one place so the shape this suite depends on is stated once:
- * `readDeclared(ql, type)` → `registry.listItems(type)` unwrapped by
- * `content ?? item`.
+ * `readDeclared(ql, type)` → `registry.listItems(type)`, filtered.
+ *
+ * [#8378] The `content ?? item` unwrap this helper used to mirror is gone from
+ * the production read. It was a no-op here — these tests drive a REAL
+ * `ObjectQL`, whose `registerMetadataCollections` registers each item as-is —
+ * which is exactly the measurement that retired it.
  */
 function readDeclaredShape(engine: ObjectQL, type: string): any[] {
-  return (engine.registry.listItems<any>(type) ?? []).map((i: any) => i?.content ?? i).filter(Boolean);
+  return (engine.registry.listItems<any>(type) ?? []).filter(Boolean);
 }
 
 const PKG = 'com.acme.exporter';
--- a/packages/objectql/src/engine-nested-plugin-collections.test.ts
+++ b/packages/objectql/src/engine-nested-plugin-collections.test.ts
@@ -81,15 +81,17 @@ function manifestDirect(collections: readonly string[]) {
   return manifest;
 }
 
+// [#8378] Both helpers mirrored the plugin readers' `content ?? item` unwrap,
+// which has been retired: `registerMetadataCollections` registers each
+// collection element as-is, so the unwrap was a no-op on this real engine.
 function registeredNames(engine: ObjectQL, type: string): string[] {
   return (engine.registry.listItems<any>(type) ?? [])
-    .map((i: any) => i?.content ?? i)
     .filter(Boolean)
     .map((i: any) => i.name);
 }
 
 function registeredItem(engine: ObjectQL, type: string): any {
-  return (engine.registry.listItems<any>(type) ?? []).map((i: any) => i?.content ?? i).filter(Boolean)[0];
+  return (engine.registry.listItems<any>(type) ?? []).filter(Boolean)[0];
 }
 
 describe('registerPlugin — the four collections a nested plugin used to drop (#7049)', () => {
--- a/packages/objectql/src/engine-nested-plugin-view-expansion.test.ts
+++ b/packages/objectql/src/engine-nested-plugin-view-expansion.test.ts
@@ -98,9 +98,10 @@ function viaNestedPlugin() {
   return { id: PKG, name: 'sales', plugins: [{ name: 'sales-nested', views: [accountContainer()] }] };
 }
 
+// [#8378] The `content ?? item` unwrap this mirrored is retired — a no-op on a
+// real engine, whose `registerMetadataCollections` registers items as-is.
 function viewItems(engine: ObjectQL): any[] {
   return (engine.registry.listItems<any>('view') ?? [])
-    .map((i: any) => i?.content ?? i)
     .filter(Boolean);
 }
 
--- a/packages/plugins/plugin-email/src/bootstrap-declared-email-templates.test.ts
+++ b/packages/plugins/plugin-email/src/bootstrap-declared-email-templates.test.ts
@@ -356,3 +356,77 @@ describe('mapTemplateToRow', () => {
     expect(row).not.toHaveProperty('variables_json');
   });
 });
+
+// ---------------------------------------------------------------------------
+// [#8378] The retired `i?.content ?? i` unwrap
+// ---------------------------------------------------------------------------
+
+/**
+ * `content` is a spelling an author really can write on an email template —
+ * `EmailTemplateDefinitionSchema` lists it in its `strictObject` **aliases**
+ * table (`content: 'bodyHtml'`). That table is a REJECTION facility, not a
+ * conversion: it feeds `strictUnknownKeyError`, which runs only on the
+ * `unrecognized_keys` path and only builds a *message*. Nothing rewrites the
+ * key — the ADR-0087 conversion registry has zero `email_template` entries, so
+ * `normalizeStackInput` emits no notice and leaves `content` where it was
+ * written.
+ *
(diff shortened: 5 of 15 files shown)


os-zhuang:

## ⭐ Step 1 falsified the card's premise, and triage's, and mine

The dispatch told this dev to measure whether a pre-conversion `content` key can reach the email-template read, and ⛔ not to assume either answer. The measurement came back with a third answer none of us had: **the question's premise is false — there is no conversion.**

`content` is not an authorable alias that maps to `bodyHtml` in a conversion layer. `packages/spec/src/conversions/registry.ts` has **zero** `email_template` entries; the key sits in a `strictObject({ aliases: … })` table that feeds `strictUnknownKeyError`, which runs only on the `unrecognized_keys` path and only builds a **message string**. It never rewrites input. It is a **rejection** alias.

⚠️ That reading was in the card body, repeated in triage's routing comment, and carried forward unchallenged in my dispatch order. Three of us propagated it. It took reading `registry.ts` to kill it — which is exactly why the dispatch made step 1 a measurement instead of a fix.

## ⭐ And the honest answer is worse than "dormant", not better

This is the part I want on the record, because the card's own severity grading would have led somewhere wrong. Having established the key cannot reach that read through any validating door, the easy conclusion is "dormant everywhere, mechanical cleanup". The dev did not stop there — it asked what happens on the paths that *do* skip validation (`strict: false`, hand-built manifests, direct registry writes), and found the unwrap **actively destructive**:

1. **It destroyed the author's own fix.** The schema exists to answer this exact mistake — *"Did you mean `content` → `bodyHtml`?"* The unwrap replaced the document with the body string, so the parse saw a string and answered `Invalid input: expected object, received string`, and the boot warning's `name` came back `undefined` — so an operator could not even tell **which** template failed. ⭐ The unwrap was the one thing standing between the author and their own prescription.
2. **`content: ''` vanished outright.** Falsy but non-nullish, so it passed `??` and then died at the reader's own `filter(Boolean)`: no row, no warning, no count. The card predicted `''` "passes `??`" — correct, and then the item is silently dropped. That is the ADR-0078 silent-loss shape, and it is strictly worse than the card stated.

## ⭐ The fixtures were the only producers of the envelope in the entire tree

> "these fakes boxed every declared item as `{ content: item }`, which made them the only producers of that envelope anywhere in the tree. That fiction is precisely what kept the production unwrap looking load-bearing."

That closes the loop on #7519's "no producer" measurement and explains how this family survived a decade of readings: every test that exercised it manufactured the very shape it was defending against. A pre-existing suite that supplies its own envelope is a suite that cannot fail — the definition of a vacuous pin.

## ⛔ On the six fixture sites outside the enumeration — my dispatch order was wrong, not the PR

My claim comment said "⛔ Zero changes outside this enumeration." The PR changes six fixture sites that were not on it. **That constraint was mine and it was under-specified**: the unwrap cannot be removed while the fakes box items as `{ content: item }` — the suites would go red. The expansion is forced by the removal, not discretionary.

The right handling was to declare it explicitly with a per-site table and a stated reason, which is what happened. ⛔ I am not treating this as a scope violation, and I am recording the correction so the next dispatch order of this shape says "plus whatever the removal forces, enumerated" rather than an absolute that cannot be honoured.

## The dormancy assumption — verified per type, and extended

Checked mechanically against each bound schema rather than trusted, as asked. All six types reject `content`. ⭐ **Plus one the card missed**: `readDeclared` is exported and generic and is called with **three** types, not one — `permission`, `capability`, and **`object`** (from `security-plugin.ts`, feeding `applyManagedWriteDenies`). `object` is absent from the card's list. `ObjectSchema` rejects `content`, so the assumption survives — and the PR says the honest thing about it: *"it held by luck, not by the card's reasoning."*

## Reverse verification, and the trap

Predicted RED / measured RED, with both failures landing in the predicted **mechanism** and not merely the predicted direction: `expected undefined to be 'auth.welcome'` (the diagnostic could not name the template, because the parse was handed a bare string) and `expected { seeded: 0, skipped: 0 } to deeply equal { seeded: 0, skipped: 1 }` (the empty-string template disappeared without being counted). Green after: 23 files / 369 tests.

⭐ Vacuity trap named and closed correctly — both new fixtures put a `content` key genuinely **in play**, because a template that never spells `content` exercises only the `?? i` arm and would pass against a completely unfixed tree. That is precisely what the pre-existing suite was doing.

The in-flight check I ran at dispatch also held up: **#8461 landed mid-flight**, the dev rebased and re-checked as triage required, and `plugin-security` is green on the rebased base with #8461's 23 new tests included.

---

Issue #8328, opened 2026-08-13
The MCP prompt bridge's listSkills reads the un-merged registry listing, so a runtime meta PUT never reaches prompts


os-zhuang:
Split out of #7654 (symptom 2) at dispatch time, per the lane ruling recorded there: the metadata half landed in `packages/metadata-protocol`, and this half lives in `packages/mcp`, which is `domain:cli`'s lane. Filed unassigned with no `domain:*` label so triage routes it.

Blocked-by: #8726

## The fact

`PUT /api/v1/meta/skill/{name}` with `{active:true}` returns 200, and the flip is **not** reflected over MCP prompts.

The two surfaces read different sources:

- The MCP prompt bridge's `listSkills` calls `metadataService.list('skill')` — the registry/package listing, which carries no `sys_metadata` store overrides. Located in `packages/mcp/src/mcp-server-runtime.ts` as `listSkills: async () => (await metadataService.list('skill')) ?? []` (anchor on the symbol, not the line — the file has moved under several merges).
- The HTTP meta list goes through `protocol.getMetaItems`, which merges the store overrides.

So a runtime override lands on the `getMetaItems` path and never on the path the bridge consumes.

## What #7654 changed, and what it did NOT change

#7654 fixed the **metadata half only** — the duplicate-row defect on `GET /api/v1/meta/skill`. Measured there: `getMetaItems` ran two different merge implementations that disagreed about a package-less row, and the MetadataService merge now uses the same package-aware resolution as the overlay merge.

**That fix does not change the bridge's answer at all**, which is the question the triage comment on #7654 asked to have settled before this was treated as one bug. It does not, because the bridge never calls `getMetaItems`; it calls `metadataService.list` directly, one layer below where any overlay merging happens. The two symptoms shared a description, not a cause.

What #7654 does provide is the thing this card needs: a `getMetaItems` read that now returns exactly one correctly-merged row per skill name. That is why this is `Blocked-by` rather than independent — the merged read is worth pointing the bridge at only once it is correct, and as of #7654 it is.

## Coordination — read this before starting

⚠️ **#6504 is open, assigned, and rewrites this same expression for a different reason.** It quotes `(await metadataService.list('skill')) ?? []` as one of its two measured consumers and proposes adding a `listDiagnosed` counterpart to `IMetadataService` so a known-partial answer stops presenting as a complete one.

[...]

Issue #8266, opened 2026-08-13
The stdio MCP record resource reads rows through its own `getRecord` seam, skipping the ADR-0049 `apiEnabled` / `apiMethods` exposure gate


os-zhuang:
Found while implementing #8083 (applying the exposure gate on the stdio MCP data bridge). Filed rather than fixed there: #8083's scope is the bridge, and this is a different seam in a different file.

## What differs

#8083 closed the declaration leak on the six object-CRUD verbs, which all flow through `createStdioDataBridge` (`packages/mcp/src/stdio-data-bridge.ts`). The ADR-0101 record resource does not use that bridge at all. Its reader is a separate closure built inline in the plugin:

- `packages/mcp/src/plugin.ts` — `getRecord = async (objectName, recordId) =` ... calls `scopedQl.find(objectName, { where: { id: recordId }, limit: 1, context: ec })` directly
- handed to `this.runtime.bridgeResources(metadataService, getRecord)`
- served as `objectstack://objects/{objectName}/records/{recordId}`

So after #8083 an object declaring `enable.apiEnabled: false` is refused by `get_record` (the tool) and still readable through the record *resource*, on the same transport, under the same key.

## Why it matters

Same grading as #8083, and for the same documented reason: **this is a surface-area declaration leak, not an authorization bypass.** `packages/runtime/src/api-exposure.ts` records the gate as a surface-area control rather than the authorization boundary, and this read passes the engine's CRUD / FLS / RLS middleware exactly as before — it goes through `ql.find` with the key's `ExecutionContext`. What leaks is the author's exposure declaration. Please do not triage this as a security issue.

Scope note: the gap predates #8034 — the record resource has read rows over `ql.find` without the gate since #7645, which is the narrower form of the divergence #8083's body records under "Scope note".

## Suggested shape (not prescriptive)

#8083 landed the gate as `enforceApiExposure` inside `stdio-data-bridge.ts`, taking its decision from the spec's single source of truth (`resolveEffectiveApiMethods` / `isApiOperationAllowed`) rather than re-implementing it. Two routes:

- **A** — export that helper from `stdio-data-bridge.ts` and call it in `getRecord` before the `find`. Smallest diff. The gated action is `get`, matching what the HTTP path sends for a single-record read.
- **B** — build the resource reader from the bridge instead of inline, so the resource and the tool surface cannot drift on anything, not just this gate. Larger; overlaps the seam question options A/B on #8083 raise.

[...]

Issue #8194, opened 2026-08-12
[finding] Existence coherence of engine/driver doubles is enforced only at the two consumer seams — `check:engine-double-contract` still cannot see it


os-zhuang:
Follow-on filed from the #8058 audited sweep, exactly as that card's dispatch ruling planned: option 1 (widen `check:engine-double-contract` to existence coherence) lands in `scripts/check-engine-double-contract.mjs`, so it is `domain:devx` and not #8058's lane. Filed unassigned. **Read #8058's sweep report first — it changes the cost/benefit of this card, and not in the direction the parent card assumed.**

## What the sweep measured (so this card starts from data, not from the hypothesis)

Audited every data-access double in `packages/objectql/src` at `2473cd2d3` — 154 doubles, enumerated by AST, then instrumented at runtime (24,385 recorded calls) to see what each one's write side was *actually handed*:

- **0** doubles took a by-id `update`/`delete` their own read side denied.
- **0** doubles had an id-blind read side (answering a constant row for any id).
- **63** doubles took at least one by-id write, and every one was answered by that double's own read side — the positive control proving the instrument can tell answered from denied.
- **89** were called but never handed a by-id write; **37** of those statically carry a read side that denies every row (`findOne` unconditionally empty). Their incoherence is real but **dormant**.

## Why zero, and why that is the interesting part

The parent card's framing was that the five doubles fixed in PR #7989 were found by coincidence and "the next incoherent double waits for the next unrelated change." The sweep falsifies the second half. #7989 did not merely fix five doubles — it installed the **unconditional** by-id prior-row read, and that turned the whole class self-detecting at the engine seam: a driver-shaped double whose `findOne` denies the row now fails its own test loudly with `RECORD_NOT_FOUND` the moment anyone points a by-id write at it. The green suite is therefore *evidence* that no driver-shaped incoherence is load-bearing, not merely an absence of evidence.

Both consumer seams are already gated and already pinned:

| seam | gate | pin |
|---|---|---|
| `ObjectQL.update`/`delete` by id (driver-shaped doubles) | unconditional prior read, #7989 | `engine-write-not-found-gate.test.ts`, plus the read-count pins in `engine-update-prior-read-scope.test.ts` / `engine-delete-prior-read-scope.test.ts` |
[...]

PR #8507, opened 2026-08-13
fix(mcp): stdio bridge throws the shared RECORD_NOT_FOUND envelope

os-zhuang:
Fixes #8422

## What changed

`packages/mcp/src/stdio-data-bridge.ts` minted its own local `recordNotFound(object, id)` — a bare `Error` with neither `code` nor `status` — thrown from the `update()` and `remove()` by-id write seams on a missing id. The HTTP bridge's `callData` path throws the repo's ONE not-found envelope, `recordNotFoundError` (`code: 'RECORD_NOT_FOUND'`, `status: 404`, `packages/core/src/utils/record-not-found.ts`, #4435/#5138/#7867), for the identical miss. Same operation, two transports, two envelopes.

The local mint is deleted; both seams now `throw recordNotFoundError(object, id)`.

**Line numbers were stale** (#8439 / #8266 touched this file since the card was filed) — located both call sites by symbol (`recordNotFound(object, id)` calls inside `update()`/`remove()`), not by the card's cited `:337`/`:347`.

**Import path**: `import { recordNotFoundError } from '@objectstack/core'`, not `@objectstack/metadata-protocol`'s re-export. `@objectstack/mcp`'s `package.json` already declares a direct `@objectstack/core` dependency, and `packages/mcp/src/plugin.ts` already imports named exports from it — `@objectstack/core` is also the lower of the two packages that carry the factory (`packages/objectql/src/engine.ts`'s own #7867 comment explains why: `@objectstack/objectql` cannot import `@objectstack/metadata-protocol` at all, ADR-0076 D2). No new dependency needed either way.

**Preserved, not touched**: the local factory's comment explaining *why* it throws rather than returns (`registerObjectTools` turns a throw into a tool error) — that reasoning is correct and now sits at the `update()` call site (the first of the two seams), with `remove()` pointing back at it. Only *which error object* is thrown changed.

## Adjacent deliverable — folded in, not filed as a follow-up

`scripts/check-engine-double-contract.mjs`'s consumer-seam invariant (#8194) reported the two stdio seams as `refusal: 'local'` rather than reddening, because the gate's own header names this as a deliberate, temporary state pending this exact card, with a noted `SHARED_ONLY`-shaped one-line tightening. Since the seam list was already both-directions complete and the tightening really is mechanical, I folded it into this PR rather than filing a follow-up:

- **File surface**: `scripts/check-engine-double-contract.mjs` — the seam filter changed from `seams.filter((x) => !x.refusal)` (only "no refusal at all" failed) to `seams.filter((x) => x.refusal !== 'shared')` (a local mint now fails too), plus an updated header comment (`## WHICH not-found envelope (#8194, tightened to SHARED_ONLY by #8422)`) and an adjusted `REFUSES` message that distinguishes "refuses through a locally minted error" from "does not refuse anywhere before it".
- Self-test (`--self-test`) still green — none of its synthetic fixtures assert the pre-tightening filter behavior, only the `refusal` *classification* (`shared`/`local`/`null`), which is unchanged.
- Real-tree run: all four consumer seams now report `[shared]` (was 2 `[shared]` + 2 `[local]`); gate is green.

## Reverse verification

**Prediction (recorded before running):** reverting only `stdio-data-bridge.ts` to its pre-fix state, while keeping the tightened gate and the new test file, goes red in both places — the new unit tests (asserting `code`/`status`) fail, and `check-engine-double-contract.mjs` reddens because the two stdio seams revert to `refusal: 'local'`, which the tightened `SHARED_ONLY` filter now rejects. Ordinary direction (more red), no inversion expected.

**Measured**, via `git checkout HEAD~1 -- packages/mcp/src/stdio-data-bridge.ts` (fix already committed, so this is a real restore point) against the tightened gate + new tests, then restored via `git checkout claude/issue-8422-stdio-shared-not-found-envelope -- packages/mcp/src/stdio-data-bridge.ts`:

- Tests: 3/3 new tests failed — `expected undefined to be 'RECORD_NOT_FOUND'`.
- Gate: 2 `REFUSES` errors — `stdio-data-bridge.ts:351` and `:361`, both "refuses through a locally minted error rather than the shared envelope".

Matches the prediction. Restored cleanly (`git diff HEAD -- packages/mcp/src/stdio-data-bridge.ts` empty); both re-verified green afterward.

## Tests

- New: `packages/mcp/src/stdio-data-bridge.not-found.test.ts` — covers **both** by-id write seams (`update()` and `remove()`) against a missing id, asserting `err.code === 'RECORD_NOT_FOUND'` and `err.status === 404` specifically (not just that something threw), plus a same-shape-on-both-seams check.
- `pnpm --filter @objectstack/mcp test` → 16 test files, 173 tests passed.
- `pnpm --filter @objectstack/mcp typecheck` → clean.
- `pnpm --filter @objectstack/mcp build` → clean (also built the dependency closure first: `pnpm --filter '@objectstack/mcp^...' build`).
- `node scripts/check-engine-double-contract.mjs --self-test && node scripts/check-engine-double-contract.mjs` → self-test OK, real run OK (197 pinned, 133 DEBT, 2 exempt; all 4 consumer seams `[shared]`).
- `node scripts/check-error-code-casing.mjs` → OK.
- `node scripts/check-nul-bytes.mjs` → OK.
- `node scripts/check-cross-package-test-inputs.mjs` → OK (new test file's package-scoped read is already covered).
- `pnpm run check:query-options-erasure` → OK (test surface unaffected — the new file has no query-options sites).
- `pnpm run check:type-check-coverage` → OK, ledger unchanged (13 packages / 436 frozen errors in DEBT, unchanged; new test file typechecks cleanly, no debt raised).
- `pnpm exec eslint` on the three changed files → clean.
- `node scripts/pm/dispatch-gates.mjs <changed paths>` → surfaced `check:cross-package-test-inputs`, `check:engine-double-contract` (both already run above) and, as convention-triggered by adding a test file, `check:query-options-erasure` + `check:type-check-coverage` (both run above, both green — no addition beyond the prompt's named families was otherwise needed).

## Changeset

`.changeset/mcp-stdio-record-not-found-envelope.md` — `@objectstack/mcp`, `patch` (behavior fix, no exported symbol or authorable metadata moves).

## Not in scope

Every other divergence between the two MCP transports noted in the file's own docblock (`callData`'s protocol-service preference, ingress `readonly` strip, existence probes, `expand`/`select`) — deliberate, filed, and untouched here. Also not addressed: #8328 (`mcp-server-runtime.ts`, `needs-user-decision`) — out of this card's file surface.

Refs #8194, #8083, #8266, #7867, #5138, #4435.

---

### DIFF
--- a/.changeset/mcp-stdio-record-not-found-envelope.md
+++ b/.changeset/mcp-stdio-record-not-found-envelope.md
@@ -1,0 +1,21 @@
+---
+'@objectstack/mcp': patch
+---
+
+fix(mcp): stdio bridge's `update`/`remove` throw the shared `RECORD_NOT_FOUND` envelope (#8422)
+
+The stdio MCP bridge's `update()` and `remove()` — the two by-id write seams
+that probe for the row before mutating it — minted their own local
+`recordNotFound(object, id)`, returning a bare `Error` with neither `code`
+nor `status`. The HTTP bridge's `callData` path already throws
+`recordNotFoundError` (`code: 'RECORD_NOT_FOUND'`, `status: 404`,
+`@objectstack/core`, #4435/#5138/#7867) for the identical miss, so the same
+operation answered a missing id with two different envelopes depending on
+which MCP transport served it.
+
+`packages/mcp/src/stdio-data-bridge.ts` now imports `recordNotFoundError`
+from `@objectstack/core` — a dependency the package already declares — and
+throws it from both seams instead. `registerObjectTools` still turns the
+throw into a tool error exactly as before; only the thrown object's shape
+changed. No exported symbol moves and no authorable metadata is affected, so
+this ships as a `patch`.
--- a/packages/mcp/src/stdio-data-bridge.not-found.test.ts
+++ b/packages/mcp/src/stdio-data-bridge.not-found.test.ts
@@ -1,0 +1,109 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * #8422 — the stdio bridge's by-id write seams must throw the repo's ONE
+ * not-found envelope (`recordNotFoundError`, `@objectstack/core`), not a
+ * locally minted `Error`.
+ *
+ * The HTTP bridge routes its data verbs through `callData`, which throws
+ * `recordNotFoundError` — `code: 'RECORD_NOT_FOUND'`, `status: 404`
+ * (`packages/core/src/utils/record-not-found.ts`, #4435/#5138/#7867). The
+ * stdio bridge minted its own bare `Error` for the identical miss, so a
+ * stdio caller got a message with no machine-readable code and nothing that
+ * maps to 404 — the same operation, two different envelopes depending on
+ * which transport served it.
+ *
+ * Both by-id write seams are covered — `update()` and `remove()` — asserting
+ * on `code` AND `status`, not merely that something threw: an assertion that
+ * only checks for a thrown error would have passed against the bare `Error`
+ * this card exists to remove.
+ */
+
+import { describe, it, expect, vi } from 'vitest';
+import type { ExecutionContext } from '@objectstack/spec/kernel';
+import type { IDataEngine, IMetadataService } from '@objectstack/spec/contracts';
+import { createStdioDataBridge } from './stdio-data-bridge.js';
+
+/** An engine that resolves NO row for any id — every by-id write is a miss. */
+function makeEmptyEngine() {
+  return {
+    find: vi.fn(async () => []),
+    findOne: vi.fn(async () => null),
+    insert: vi.fn(),
+    update: vi.fn(),
+    delete: vi.fn(),
+    count: vi.fn(async () => 0),
+    aggregate: vi.fn(async () => [{ n: 0 }]),
+  };
+}
+
+/** An object definition with no exposure restriction, so the miss is what refuses. */
+function makeMetadata() {
+  return {
+    listObjects: vi.fn(async () => [{ name: 'task', label: 'Task', fields: {} }]),
+    getObject: vi.fn(async () => ({ name: 'task', label: 'Task', enable: { apiEnabled: true } })),
+    get: vi.fn(async () => null),
+    list: vi.fn(async () => []),
+    exists: vi.fn(async () => true),
+    getRegisteredTypes: vi.fn(async () => ['object']),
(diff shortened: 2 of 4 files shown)


os-zhuang:
CI follow-up on the two red checks flagged.

**TypeScript Type Check (`check:type-check-debt`) — was real, now fixed.** My new `stdio-data-bridge.not-found.test.ts` added exactly 5 raw tsc errors to `@objectstack/mcp`'s TEST_DEBT baseline (recorded 53, measured 58). All 5 were `TS2571` ("object is of type 'unknown'") on `updateErr`/`removeErr` in the third test's `.catch((e) => e)` idiom — TypeScript infers the settled `TResult` from an `any`-typed catch parameter as `unknown` here, a second, unrelated instance of the same category of idiom already sitting in this package's 53-error baseline (`await res.json()`). Fixed at the source: replaced the untyped `.catch((e) => e)` with the same explicit-cast pattern the other two tests already used (`.then(() => null, (e: unknown) => e)` cast to `Error & { code?: string; status?: number }`), extracted once as a shared `catchError` helper. No ledger entry touched.

Verified with the actual command, not the one I ran the first time: `pnpm check:type-check-debt` (`check-type-check-coverage.mjs --self-test && --re-measure`) on a full `pnpm build` closure → `OK — 33 ledger entr(ies) re-measured ... none above its recorded number`, `@objectstack/mcp` back to exactly 53. Re-ran `pnpm --filter @objectstack/mcp test` (173/173) and `check-engine-double-contract.mjs` (self-test + real run, all 4 seams `[shared]`) — both still green.

**ESLint — confirmed a flake, not mine.** Polled the new SHA's check-runs after push: ESLint completed `success` at 19:45Z (TypeScript Type Check followed at 19:48Z), no `check-regen-pending.mjs` git-plumbing error this time. One occurrence, as predicted — leaving `check-regen-pending.mjs` untouched.

Pushed as `f051980d0`. Full check-run sweep on that SHA: every completed job is `success` (ESLint, TypeScript Type Check, Test Core ×3 + aggregate, Build Core, Check Changeset, ADR maintainer approval, Temporal Conformance, Dogfood Verify CLI, Dogfood Regression Gate 1/3 + 3/3, the two duplicate/scope guard checks, etc.); `Dogfood Regression Gate (2/3)` was still `in_progress` at last check, unrelated to either flagged job. Still draft, as instructed.

---

---

Issue #8536, opened 2026-08-13
The DERIVED half of `bootstrapSystemCapabilities` also leaves the platform bucket unseeded when an organization holds the name — the #5876 guard skips instead of inserting


os-zhuang:
Found while implementing #8470. The PM ruled the derived half explicitly out of scope for that card, so this records the finding rather than changing it.

## The mechanism

`bootstrapSystemCapabilities` looks a capability up by `name` across organizations (system context, no `tenantId`). #8470 fixes the CURATED half by scoping its lookup to the row the platform owns. The DERIVED half's lookup stays `{ name }` — deliberately, because its own #5876 guard already refuses to overwrite a row it does not own:

```ts
if (isDerived && row.managed_by !== 'platform') {
  skippedAuthored += 1;
  continue;
}
```

The guard is correct about what it protects. What it does not do is `insert`. So when an organization has authored a row for a name that a bootstrap permission set grants via `systemPermissions[]`, the cross-organization lookup finds the org's row, the guard fires, and the loop `continue`s — the platform's own derived placeholder is **never created in the NULL-organization bucket**.

That is the same shape as #8470's second harm ("the platform's row is never seeded"), on the half #8470 leaves alone. It became reachable for the same reason: #8461 widened `sys_capability.name` from installation-wide to per-organization uniqueness (ADR-0120 D1), so the two-row state exists at all.

Note this was measured as **safe** in #8470's own table (its row 1) — correctly, because that row asked whether the organization's label survives, and it does. The unseeded platform bucket is a different question that row did not ask.

## Measured consequence: none, today

I checked the same question #8470 asked about the curated half, with the same method: no runtime path reads a `sys_capability` row. Grants (`systemPermissions`) and requirements (`requiredPermissions`) resolve capabilities **by name**; `validateCapabilityReferences` resolves against `PLATFORM_CAPABILITY_NAMES` plus stack declarations, never the table. So the effect is confined to the registry contents and to Setup's capability listings — the platform's own row for that name is simply absent.

That is why I am filing this rather than treating it as urgent, and why the severity judgment should be made by whoever decides whether the registry is supposed to be complete.

## The open question this really turns on

[...]

Issue #8535, opened 2026-08-13
`sys_capability.active` is read by nothing — the Deactivate action tells the admin grants stop resolving, and they do not


os-zhuang:
Found while implementing #8470, which required establishing whether any runtime path resolves a `sys_capability` **row** (as opposed to the bare capability name string). The answer is no — and that answer has a consequence beyond #8470.

## The claim the product makes

`sys-capability.object.ts` ships a `deactivate_capability` record action whose confirmation dialog reads, verbatim:

> Deactivate this capability? Grants and resource requirements that reference it stop resolving until re-activated.

The `active` field is also surfaced in all three list views and in `highlightFields`, so deactivation presents as a first-class administrative control.

## What actually happens

Nothing reads it. Measured on `origin/main` at `30f1b7488d`:

- The AND-gate that enforces `requiredPermissions` compares **string sets**. `PermissionEvaluator.getSystemPermissions()` unions `permissionSets[].systemPermissions` (strings); `normalizeRequiredPermissions` reads the resource's own declared strings. Neither loads a row.
- A repo-wide search for row reads of the object — `find`/`findOne`/`count`/`aggregate` against `'sys_capability'`, plus every consumer of the `platform-object-names` entry — returns exactly two production call sites, both seeders: `bootstrap-system-capabilities.ts` and `bootstrap-declared-capabilities.ts`. Both read `name`, `managed_by`, `package_id`. Neither reads `active`, and neither is on an authorization path.
- `validateCapabilityReferences` (the authoring lint) resolves against `PLATFORM_CAPABILITY_NAMES`, stack declarations and seed-data records — the spec constant and the config, never the table.

So `active: false` changes no authorization decision, no lint verdict, and no runtime resolution. It changes a column value and a badge in Setup.

## Why this is worth a card rather than a shrug

The direction of the falsehood is the bad one. An admin who wants to withdraw a capability is offered a control that states, in a confirmation dialog, that withdrawing it takes effect — and the withdrawal silently does not happen. That is a security-relevant misstatement about a security surface even though it is not itself a privilege escalation: the escalation is what the admin believes they prevented.

This is also an ADR-0049 enforce-or-remove shape: a declared field with a declared meaning and no enforcement.

## The two directions, not a recommendation

I am deliberately not choosing here, because the choice is a contract decision:

[...]

Issue #8468, opened 2026-08-13
`sys_position.name` is the third instance of the #8323 class: an admin-authored name on a tenant-scoped RBAC object carries an installation-wide unique index


os-zhuang:
Found while implementing #8323 (option 2), sweeping the docs for uniqueness claims at the PM's request. **Not fixed there** — the ruling of 2026-08-13 named exactly two objects, and this is a third. Filed for triage rather than folded in.

## The declaration

`packages/plugins/plugin-security/src/objects/sys-position.object.ts`:

```ts
indexes: [
  { fields: ['name'], unique: true },
  { fields: ['active'] },
],
```

Structurally identical to the `sys_capability` instance #8323 just fixed:

| | `sys_capability` (fixed in #8461) | `sys_position` |
|---|---|---|
| declared index | `{ fields: ['name'], unique: true }` | `{ fields: ['name'], unique: true }` |
| `tenancy` opt-out | none → gets `organization_id` | none → gets `organization_id` |
| provenance | `managedBy: 'config'`, admins extend in Setup | `managedBy: 'config'`, admins author positions |

On a DECLARED index bare `unique: true` is the positional spelling of `'global'` — the listed columns verbatim — so `name` is an installation-wide key on a tenant-scoped object. That is the #8323 §1 shape exactly: an organization can POST a position name and read 409-vs-201 to learn whether another organization already holds it, while its own read of that name returns nothing.

The reference docs assert the strong reading as though it were intended — `content/docs/references/identity/position.mdx`: `name` … "Unique position name (lowercase snake_case)". That line is generated from the field's `describe()`, so the spec text says installation-unique too.

## What I did and did not establish

- **Established statically:** the declaration, the absence of a tenancy opt-out (so `organization_id` is injected), and that the declared-index bare spelling materializes verbatim — the same mechanism measured and pinned in #8461.
- **NOT established:** a live 409/201 probe against `sys_position`, and — the real question — whether installation-wide position names are *intended*. Unlike capabilities, positions carry a hierarchy (`parent_id`), and it is at least arguable that a deployment wants one global position taxonomy. #8323's `sys_capability` had ADR-0066 D1's "admins EXTEND the registry" to settle intent; I did not find the equivalent statement for positions, so I am not asserting the answer.

[...]

PR #8537, opened 2026-08-13
fix(security): the curated capability seeder reconciles the row the platform owns, not whichever row shares the name (#8470)

### LINKED ISSUE #8470: `bootstrapSystemCapabilities` reconciles an arbitrary row when a curated capability name exists in more than one organization — `find(..., limit: 1)` has no ORDER BY, so the platform's own row can be left unseeded
Measured while implementing #8323 (PR #8461). **This PR unmasks the defect; it does not cause it** — see "Provenance" below, which also says why restoring the global unique is not the remedy.

## The mechanism

`packages/plugins/plugin-security/src/bootstrap-system-capabilities.ts` upserts each capability by name:

```ts
const existing = await tryFind(ql, 'sys_capability', { name: def.name }, 1);
const row = existing[0];
```

`tryFind` runs under `SYSTEM_CTX = { isSystem: true }`, i.e. **across organizations**, and the query is `limit: 1` with **no `ORDER BY`**. When two rows share a `name` — the platform's own (`organization_id` NULL) and one an admin authored inside an organization — the row this returns is **arbitrary**, and SQL guarantees nothing about which. It can differ between boots on the same database.

For a **curated** name the seeder then takes the update branch and writes the platform's `label`/`description` onto whichever row it got.

## Measured

Against the seeder's own in-memory `ql` harness, which models `isSystem` faithfully (filters on `where` only, ignoring organization). Direction predicted before each run; all three matched.

| # | case | result |
|---|---|---|
| 1 | **DERIVED** name (`approve_invoice`) + org-authored row | ✅ **safe** — `skippedAuthored: 1`, org label preserved. The `derivedNames.has(name) && row.managed_by !== 'platform'` guard fires |
| 2 | **CURATED** name, platform row seeded **first**, org row added after | ⚠️ **safe in the harness ONLY** — see the caveat below |
| 3 | **CURATED** name (`manage_users`), org row present, platform row **not yet** | ⛔ **`platformRowExists: false`** — the platform's own row is **never seeded**, and the org row's `ORG CUSTOM LABEL` is overwritten with `Manage Users` |

### ⚠️ The caveat on row 2 — please do not read this card as "2 of 3 safe"

Row 2 passes **only because the fake returns rows in insertion order**. The real query is `find(..., { where: { name }, limit: 1 })` with **no `ORDER BY`**, and SQL provides no ordering guarantee there. On a real database, once two rows share a name the seeder's choice is arbitrary and may flip between boots — so row 2 is not a safe case, it is row 3 with an unobserved coin flip. Insertion order is a property of the test double, not a property of the system.

### The concrete failure

`platformRowExists: false`. Two harms, the second worse:

1. An organization's authored capability row is silently overwritten with the platform's copy at boot.
2. **A single tenant's row suppresses the platform's own definition installation-wide** — the curated row is never created, in any bucket.

## Scope of what was measured — verbatim, and please keep it prominent

> I measured **seeding** behaviour only. I did **not** measure whether a missing platform capability row changes any authorization decision.

My reasoning that this is not a privilege issue — **stated as my claim with its basis, not as a finding** — is that `sys_capability` rows are *definitions*, while grants live in permission sets (`systemPermissions`) and requirements in `requiredPermissions`, both of which reference a capability **by name** rather than by row. That is why I expect a missing row to be a metadata/UX defect rather than an authz one. I did not test it, and I would rather the next person measure it than inherit my confidence.

**Open question, explicitly unresolved:** *does a never-seeded curated capability break any downstream assumption?* Candidates worth checking: `validateCapabilityReferences` (the authoring lint warns when `requiredPermissions` names a capability "registered nowhere"), Setup surfaces that list capabilities, and anything that resolves a capability row rather than the bare string.

## Blast radius

Exactly the **8** curated names in `PLATFORM_CAPABILITIES` (`packages/spec/src/security/capabilities.ts`):

`manage_users` · `manage_org_users` · `manage_metadata` · `manage_platform_settings` · `setup.access` · `setup.write` · `studio.access` · `manage_sharing`

Derived names are guarded (row 1). Reaching it requires an org admin to author a `sys_capability` row whose name **exactly** matches one of those 8 — which is a supported action (`managedBy: 'config'`; ADR-0066 D1: the platform DEFINES, admins EXTEND in Setup).

`sys_user_preference` is **unaffected** — nothing seeds preferences by name across organizations.

## Provenance — unmasked, not caused, and ⛔ not a reason to restore the global unique

The nondeterministic `limit: 1` lookup **predates** #8461 and is unchanged by it. What #8461 changed is only that the two-row state became **reachable**: `sys_capability.name` was an installation-wide unique index, so an organization could not author a row named `manage_users` at all (it got `409`) — the constraint made the ambiguity unreachable rather than correct.

⛔ **Restoring the global unique is not the fix.** That index is precisely the release-blocking cross-tenant existence oracle #8323 reports (an organization could read 409-vs-201 to enumerate values it cannot see, on the `isolated` posture sold as a hard wall). Reverting it to hide this defect would reopen a security hole to paper over a metadata one.

## Direction

No contested intent here — the lookup is simply under-specified. Either or both:

- make the platform lookup **deterministic** (order the query, or select by identity rather than by name alone), and/or
- **scope the platform seeder's lookup to `managed_by = 'platform'`** (and/or `organization_id IS NULL`), so it reconciles the row it owns and never an organization's.

The second matches the intent already expressed one branch down, where the derived half deliberately refuses to touch a row it does not own (`#5876`). The curated half simply never had the equivalent guard, because it never needed one.

Filed unassigned per the PM's sequencing. Not fixed in #8461 — deliberately, to keep a security release blocker narrow. Measured by session `session_012WMpuAfA2KSdDjGF6tm1bH`.

---

os-zhuang:
Fixes #8470

The curated half of `bootstrapSystemCapabilities` now looks up **the row the platform owns** instead of the first row that happens to share the name.

```ts
const lookup = isDerived
  ? { name: def.name }
  : { name: def.name, managed_by: 'platform', organization_id: null };
```

The DERIVED half is untouched (its own #5876 guard already refuses to write to a row it does not own).

---

## The card's mechanism was right about the defect and wrong about the cause — and the difference selects the route

The card offered two directions: make the lookup deterministic, or scope it. I took **scoping**, and the measurement that selected it also falsifies the other option outright rather than merely ranking it second.

**The lookup was already deterministic.** #4363 shipped a pagination tie-breaker: `orderKeysFor` appends `ORDER BY id` to any paged read of a driver-managed table, and `limit: 1` counts as paged on both `SqlDriver` and `MongoDBDriver` — only `findOne` opts out, via `singleRowLookup`. So `find('sys_capability', { where: { name }, limit: 1 })` has carried an `ORDER BY id ASC` this whole time.

Measured on `SqlDriver` + better-sqlite3, against a fixture copying `sys-capability.object.ts` including its `unique: 'organization'` index, with a platform row (NULL organization) and an org row sharing `manage_users`:

| physical insert order | org row id | returned |
|---|---|---|
| org first, platform second | `cap_org` | `cap_org` |
| **platform first**, org second | `cap_org` | `cap_org` |
| platform first, org second | `aaa_org` | `aaa_org` |
| org first, platform second | `zzz_org` | `aaa_platform` |

The selection tracks `id` ascending and ignores insertion order entirely. I predicted rows 1-2 would differ (insertion order) and was **wrong**; chasing that miss is what surfaced #4363.

Two consequences:

1. **Ordering is not the remedy.** It is already there. Adding an explicit `orderBy` would change *which* wrong row is chosen, not that a wrong row can be chosen. Determinism was never the missing property — ownership was.
2. **The defect is more persistent than the card says.** The card expects a choice that "can differ between boots on the same database". It does not: it is stable per installation, so a boot that reconciles the wrong row keeps reconciling it on every subsequent boot rather than self-healing on some later one.

The card's headline harm — `platformRowExists: false` — is real and unaffected by any of this. With no platform row there is only one candidate, so ordering never mattered for that case.

## Why both predicates, and why `limit: 1` is now safe with no ORDER BY

`managed_by: 'platform'` AND `organization_id: null` are the two facts that jointly define the platform's own row, and together they make the result set a **provable singleton**: the post-#8461 unique key is `(COALESCE(organization_id, …), name)`, so the NULL-organization bucket admits at most one row per name. `limit: 1` over a set of size 0 or 1 cannot be arbitrary. That is what retires the ordering question instead of answering it.

Neither predicate alone is enough. `managed_by` alone loses the singleton guarantee if a platform-marked row ever sits inside an organization (seed data, legacy import). `organization_id` alone cannot tell the platform's row from an admin's on a single-organization deployment, where both live in the same bucket.

Measured, same harness: `{ name, managed_by: 'platform', organization_id: null }` returns the platform row under both physical orders; returns zero rows when only an org row exists; `{ organization_id: null }` does not match an org-scoped row (it really compiles to `IS NULL`); and the platform's global insert is **not** blocked by an org row of the same name.

⛔ The installation-wide unique index is not restored and must not be — it is #8323's cross-tenant existence oracle, closed by #8461.

## The platform-bucket provenance matrix

The predicate is a conjunction, so it can also match **zero** rows where it should match one. One row in the platform's own bucket, varying only by `managed_by`. Predicted before running; all four matched.

| `managed_by` | reachable? | outcome |
|---|---|---|
| `platform` | the row the seeder owns | **matched**, reconciled, `blockedCurated: 0`, no warn |
| `admin` | **yes, and ordinary** | blocked, row untouched, warn names `managed_by='admin'` |
| `package` | **yes**, via name promotion | blocked, row untouched, warn names `managed_by='package'` |
| absent | **no** — not engine-reachable | blocked, warn says "a row carrying no managed_by value" |

- `admin` is not a hypothetical: `organization_id` auto-stamping lives in the enterprise `@objectstack/organizations` runtime, **which is also what activates every walled posture**. A deployment without it is `single` posture with no stamper, so every Setup-authored capability row lands in the NULL-organization bucket. That is the default community shape.
- `package` arises by promotion: `bootstrapDeclaredCapabilities` refuses a name already in `PLATFORM_CAPABILITY_NAMES`, but a package that declared a name **before the platform curated it** left a `managed_by:'package'` row there — and `setup.write` and `manage_sharing` were both added to the curated set after the fact.
- absent is **not** producible through the engine: `managed_by` is `required: true` with `defaultValue: 'admin'`, and `applyFieldDefaults` resolves defaults on insert before the beforeInsert hooks, so an insert omitting it stores `'admin'`. Pinned anyway, because the diagnostic must not assert what it cannot observe.

## The open question the card asked: measured, and the filer's expectation holds

> does a never-seeded curated capability break any downstream assumption?

**No authorization decision is affected.** Method, on `origin/main` at `30f1b7488d`:

- The `requiredPermissions` AND-gate compares **string sets**: `PermissionEvaluator.getSystemPermissions()` unions `permissionSets[].systemPermissions`, and `normalizeRequiredPermissions` reads the resource's declared strings. No row is loaded.
- `validateCapabilityReferences` — the candidate the card named — resolves against `PLATFORM_CAPABILITY_NAMES` (the spec constant), stack declarations and seed-data records. It never reads the table, so a curated name with no row does **not** warn.
- A repo-wide search for row reads of the object (`find`/`findOne`/`count`/`aggregate` against `'sys_capability'`, plus every consumer of its `platform-object-names` entry) finds exactly two production call sites — the two seeders. Neither is on an authorization path.

So the harm is a registry/Setup-listing defect, as the filer expected. The severity and release targeting the card was given still stand; nothing here needed to widen.

That sweep turned up something the card did not ask about, filed separately as **#8535**: `sys_capability.active` is read by nothing, while the Deactivate action's confirmation dialog tells the admin that grants and requirements stop resolving. Not addressed here; that issue remains open.

## Anti-vacuity: ablation, prediction first

Prediction written down **before** each run (fix committed first, so the restore came out of a real commit). Ablation = revert only the lookup to `{ name }`, keep the pins and the upgraded double.

| pin | predicted | measured |
|---|---|---|
| seeds the platform row even when an organization already holds the name | RED | RED |
| leaves the organization's authored copy exactly as its author wrote it | RED | RED |
| reconciles the platform row and only the platform row (**adverse** id order) | RED | RED |
| reconciles the platform row and only the platform row (**benign** id order) | GREEN | GREEN |
| platform-bucket row, `managed_by=platform` | GREEN | GREEN |
| platform-bucket row, `managed_by=admin` / `package` / absent | RED | RED |
| a clean install seeds every curated name with nothing blocked | GREEN | GREEN |
| is unaffected by the NUMBER of organizations holding the name | RED | RED |
| the derived guard still fires for an ORG-authored row | GREEN | GREEN |
| all pre-existing tests, both seeder files | GREEN | GREEN |

`7 failed | 16 passed (23)` — the predicted seven, no others. **Prediction and measurement did not diverge**, including the deliberately-green rows, which are regression controls rather than gates.

One sub-prediction inside a red pin is worth naming because it is where the information is: in the collision cases the `seeded === KNOWN - 1` assertion passes **coincidentally** under the ablation (the old code also fails to insert that row — it overwrites the other author's instead). The assertion carrying the information there is `blockedCurated`.

The three-way diagnostic split was ablated separately (two-way → three-way): **1 failed | 23 passed**, exactly the new pin.

### How the adverse ordering was forced

Per the card's warning, the old double returned rows in **insertion order**, which is a property of the double and of no driver in the system. The double now models what was measured instead:

1. `limit` sorts by `id` ascending (the #4363 tie-breaker, BINARY collation as SQLite compares);
2. a `null` comparand matches null-or-absent (`IS NULL` in SQL, `value == condition` in `driver-memory`, null-or-missing in Mongo) — strict `===` matched none of them and would have made `organization_id: null` unsatisfiable in the double while working in production;
3. `insert` enforces `(COALESCE(organization_id, '__global__'), name)`.

The adverse case is then forced by **choosing ids**, exactly as a real driver would order them: the org row gets `aaa_org_authored`, which sorts before every seeder-minted `cap_…` id. `zzz_org_authored` is the same fixture benign. The pair is the anti-vacuity argument: the old behaviour **differs** between the two ids, the fixed behaviour is identical.

Each of the three double changes was measured against a real driver first; none was made to get a test to pass.

## The case the fix newly declines, and does not do silently

A curated name can be held in the **platform bucket** by a row the scoped lookup does not match. Previously the seeder "resolved" that by overwriting the other author. Now it declines — but `tryInsert` swallows the engine's unique-constraint refusal, so declining would look exactly like a clean boot with a curated capability missing installation-wide. The result gains `blockedCurated`, counted and warned.

The warning **states what it observed** rather than asserting authorship, and splits three ways, because "no row came back" and "a row came back carrying no `managed_by`" are different observations and the first one would otherwise be described as an ordinary unstamped row:

- no blocking row visible → says so, and flags it as worth investigating (the insert was refused, so something stopped it);
- a row with no `managed_by` → says that;
- a row with a value → names the value.

⛔ It **reports** the distinction and deliberately does not **act** on it. Adopting a differently-stamped row into the platform's identity would reverse #5876's "not provably ours resolves to leave-it-alone", and backfilling a stamp is a data migration. Both are maintainer calls; neither is made here.

## Scope

- Behaviour-change sweep covered test files: the second seeder's double (`bootstrap-declared-capabilities.test.ts`) needed the same null-comparand fix, since several of its cases call `bootstrapSystemCapabilities`.
- The comment on the `sys_capability` unique index asserted "`bootstrapSystemCapabilities`' upsert-by-name is unaffected". That claim is what this card falsifies, so it is corrected next to the index that unmasked it.
- `packages/metadata-protocol/src/protocol.ts` is **not** touched (#8442 holds a region there); `sys-position.object.ts` and `driver-sql/src/schema-drift.ts` are **not** touched (#8468 holds those).
- The derived half is not changed. A gap in it, found while measuring, is filed as **#8536** and remains open.
- No shrink-only ledger was grown. The refused-insert fixture is built by **overriding** the file's existing double rather than declaring a second one, so `check:engine-double-contract` has nothing new to pin **or** ledger — its counts are byte-identical to `main`'s.

## Verification — re-run at the final head, `1abef9fd27`

An earlier round of this PR reported a green union that was run **before** the last two commits existed, and CI went red on `5a877ffe` for a gate that had genuinely passed earlier. Everything below is from a single run at the final commit. That failure mode is filed as **#8550** and remains open.

- `pnpm --filter @objectstack/plugin-security test` — 54 files, **1072 passed**; `typecheck` EXIT=0
- **`pnpm check:type-check-debt`** (the package script — `--self-test` **and** `--re-measure`, per #8545) — EXIT=0: `33 ledger entries re-measured in 312.7s, 1969 raw tsc errors, none above its recorded number; surplus: none`
- Union, all EXIT=0: `check:nul-bytes`, `check:cross-package-test-inputs`, `check:docs-audit-scope`, `check:test-source-alias`, `check:type-source-resolution`, `check:query-options-erasure`, `check:type-check-coverage`, `check:engine-double-contract`, `check:empty-changeset`, `check:i18n`, `check-changeset-no-major.mjs`
- Ledger numbers unchanged from baseline: engine doubles `201 pinned, 133 DEBT, 2 exempt`; query-options-erasure `67 unswept, none new`; type-check coverage `64/77, 13 in DEBT, 436 frozen`
- Dependency closure built before every measurement (`turbo run build --filter='./packages/*' --filter='./packages/*/*'`) — the #8246 dist-resolution trap bit twice (once on driver-sql, once as `--re-measure` refusing to run at all), and numbers taken before each rebuild were discarded.

Changeset: `minor` (widened return type; breaking-for-constructors recorded in the body, per the launch-window convention).

---

### DIFF
--- a/.changeset/curated-capability-seed-owns-its-row.md
+++ b/.changeset/curated-capability-seed-owns-its-row.md
@@ -1,0 +1,57 @@
+---
+'@objectstack/plugin-security': minor
+---
+
+**The curated capability seeder reconciles the row the platform owns — never an organization's (#8470).**
+
+`bootstrapSystemCapabilities` upserted each curated capability with
+`find('sys_capability', { where: { name }, limit: 1 })` under a system context,
+i.e. across organizations. Since #8461 made `sys_capability.name` unique per
+ORGANIZATION rather than installation-wide (ADR-0120 D1, closing the
+cross-tenant existence oracle #8323 reports), that lookup can have **two**
+candidates: the platform's own row in the NULL-organization bucket, and one an
+admin authored inside their organization — a supported action (ADR-0066 D1: the
+platform DEFINES, admins EXTEND in Setup).
+
+Two harms followed, the second worse:
+
+1. an organization's authored `label`/`description` were overwritten with the
+   platform's copy at every boot; and
+2. when the organization's row was the one selected **before** the platform's
+   row existed, the seeder took the update branch and the curated definition was
+   **never inserted, in any bucket, installation-wide**.
+
+**Ordering was not the missing property.** #4363's pagination tie-breaker
+already appends `ORDER BY id` to any paged read of a driver-managed table, and
+`limit: 1` counts as paged on both the SQL and MongoDB drivers — so the lookup
+was already deterministic, deterministic on `id`, which says nothing about who
+owns the row. Worse, being stable made it permanent: an installation that picked
+the wrong row picked it again on every subsequent boot instead of self-healing.
+
+**The curated lookup is now scoped to `managed_by: 'platform'` AND
+`organization_id: null`** — the two facts that jointly define the platform's own
+row, and which together make the result set a provable singleton (the post-#8461
+unique key is `(COALESCE(organization_id, …), name)`, so the platform bucket
+admits at most one row per name). The DERIVED half is unchanged: its own #5876
+guard already refuses to touch a row it does not own.
+
+The organization's row keeps its authored copy, and the platform's curated row
+is seeded regardless of what any organization has authored.
+
+**New `blockedCurated` field on the seeding result** (a widened return type —
+breaking for anyone constructing `CapabilitySeedResult` by hand, hence `minor`
+per the launch-window convention). It counts, and warns about, the one collision
+the seeder now declines to resolve by overwriting: a curated name already held
+in the platform bucket by a row this pass does not own — a Setup-authored row on
+a single-organization deployment. Previously that case was "resolved" by
+clobbering the other author; now it is refused, and refusing silently would be
+its own defect, since `tryInsert` swallows the engine's unique-constraint
+refusal.
+
+No authorization behaviour changes. Grants (`systemPermissions`) and
+requirements (`requiredPermissions`) resolve capabilities **by name**, and no
+runtime code path reads a `sys_capability` row to decide access — so a
+never-seeded curated row was a registry/Setup-listing defect, not a privilege
+one. No migration: an already-overwritten organization row is not restored, but
+it stops being overwritten, and a missing curated row is seeded on the next
+boot.
--- a/.changeset/curated-capability-seed-owns-its-row.md
+++ b/.changeset/curated-capability-seed-owns-its-row.md
@@ -42,11 +42,19 @@ is seeded regardless of what any organization has authored.
 breaking for anyone constructing `CapabilitySeedResult` by hand, hence `minor`
 per the launch-window convention). It counts, and warns about, the one collision
 the seeder now declines to resolve by overwriting: a curated name already held
-in the platform bucket by a row this pass does not own — a Setup-authored row on
-a single-organization deployment. Previously that case was "resolved" by
-clobbering the other author; now it is refused, and refusing silently would be
-its own defect, since `tryInsert` swallows the engine's unique-constraint
-refusal.
+in the platform bucket by a row the scoped lookup did not match. Previously that
+case was "resolved" by clobbering the other author; now it is refused, and
+refusing silently would be its own defect, since `tryInsert` swallows the
+engine's unique-constraint refusal. The warning states the provenance it
+actually **read** off the blocking row rather than asserting who authored it —
+the seeder observes "no platform-owned row matched, and the insert was refused",
+which is not the same fact as "this row belongs to someone else".
+
+Reachable and ordinary, not hypothetical: `organization_id` auto-stamping lives
+in the enterprise `@objectstack/organizations` runtime, which is also what
+activates every walled posture — so on a deployment without it (`single`
+posture, no stamper) every Setup-authored capability row lands in the
+NULL-organization bucket.
 
 No authorization behaviour changes. Grants (`systemPermissions`) and
 requirements (`requiredPermissions`) resolve capabilities **by name**, and no
--- a/packages/plugins/plugin-security/src/bootstrap-declared-capabilities.test.ts
+++ b/packages/plugins/plugin-security/src/bootstrap-declared-capabilities.test.ts
@@ -22,7 +22,16 @@ function makeQl(declared: any[] = []) {
     async find(object: string, q: any) {
       if (object !== 'sys_capability') return [];
       const where = q?.where ?? {};
-      return rows.filter((r) => Object.entries(where).every(([k, v]) => r[k] === v));
+      // [#8470] A `null` comparand is IS NULL, not `=== null`: `driver-sql`
+      // compiles `{ field: null }` to `IS NULL`, `driver-memory`'s matcher uses
+      // `value == condition`, and MongoDB matches null-or-missing — none of them
+      // is strict equality against an ABSENT key. `bootstrapSystemCapabilities`
+      // (called by several cases below) scopes its curated lookup with
+      // `organization_id: null`, which strict `===` would make unsatisfiable
+      // here while it works in production.
+      return rows.filter((r) =>
+        Object.entries(where).every(([k, v]) => (v === null ? r[k] == null : r[k] === v)),
+      );
     },
     async insert(object: string, data: any) {
       if (object !== 'sys_capability') return null;
--- a/packages/plugins/plugin-security/src/bootstrap-system-capabilities.test.ts
+++ b/packages/plugins/plugin-security/src/bootstrap-system-capabilities.test.ts
@@ -380,26 +380,84 @@ describe('[#8470] the curated half owns its row, not whichever row shares the na
     expect(out.blockedCurated).toBe(0);
   });
 
-  // ── The case the fix newly DECLINES to resolve, and must not do silently ──
-  it('reports a curated name held by another author in the platform bucket', async () => {
+  // ── The PLATFORM-BUCKET provenance matrix ─────────────────────────────────
+  //
+  // Every case above is about a name with too MANY candidates. This block is
+  // the other axis: one row, in the platform's own bucket, varying only by
+  // `managed_by`. It exists because the scoped predicate is a CONJUNCTION, and
+  // a conjunction can also match ZERO rows where it should match one.
+  //
+  // `managed_by: 'platform'` is the row the seeder owns and must still find —
+  // the NEGATIVE case, and the one that would catch the predicate excluding the
+  // platform's own row on an installation that predates something.
+  //
+  // Reachability of the other three, measured rather than assumed:
+  //  - `admin` — REACHABLE and ordinary. `organization_id` auto-stamping lives
+  //    in the enterprise `@objectstack/organizations` runtime, which is also
+  //    what ACTIVATES every walled posture. A deployment without it is `single`
+  //    posture with no stamper, so EVERY Setup-authored capability row lands in
+  //    the NULL-organization bucket. This is the default community shape, not
+  //    an edge case.
+  //  - `package` — REACHABLE via name promotion: `bootstrapDeclaredCapabilities`
+  //    refuses a name in `PLATFORM_CAPABILITY_NAMES`, but a package that
+  //    declared a name BEFORE the platform curated it (`setup.write` and
+  //    `manage_sharing` were both added to the curated set after the fact) left
+  //    a `managed_by:'package'` row in that bucket.
+  //  - no value at all — NOT reachable through the engine: `managed_by` is
+  //    `required: true` with `defaultValue: 'admin'`, and `applyFieldDefaults`
+  //    resolves defaults on insert BEFORE the beforeInsert hooks, so an insert
+  //    omitting it stores `'admin'`, never null. Kept as a pin anyway because
+  //    the diagnostic must not assert an ownership verdict it cannot observe —
+  //    which is the whole reason the message names the value it READ.
+  it.each([
+    ['platform — the row the seeder OWNS', 'platform', 'matched'],
+    ['admin    — Setup-authored, single-posture deployment', 'admin', 'blocked'],
+    ['package  — declared before the name was curated', 'package', 'blocked'],
+    ['(absent) — not engine-reachable; the message must still be truthful', undefined, 'blocked'],
+  ])('platform-bucket row, managed_by=%s', async (_label, managedBy, expected) => {
     const ql = makeQl();
-    // No `organization_id`: a Setup-authored row on a single-organization
-    // deployment sits in the SAME bucket as the platform's would-be row, so the
-    // declared unique key refuses the curated insert. Pre-#8470 the seeder
-    // "resolved" this by overwriting the admin's row.
-    ql.rows.push({ id: 'aaa_admin_global', ...ORG_AUTHORED });
+    const PRE = {
+      id: 'aaa_pre_existing',
+      name: 'manage_users',
+      label: 'PRE-EXISTING LABEL',
+      description: 'written by whoever owns this row',
+      scope: 'platform' as const,
+      active: true,
+      ...(managedBy === undefined ? {} : { managed_by: managedBy }),
+    };
+    ql.rows.push({ ...PRE });
     const warn = vi.fn();
     const out = await bootstrapSystemCapabilities(ql, [], { logger: { warn } });
+    const row = ql.rows.find((r) => r.id === 'aaa_pre_existing')!;
 
-    expect(out.blockedCurated).toBe(1);
+    if (expected === 'matched') {
+      // The platform's own row is found by the conjunction and reconciled.
+      expect(out.blockedCurated).toBe(0);
+      expect(warn).not.toHaveBeenCalled();
+      expect(row).toMatchObject({ label: CURATED.label, description: CURATED.description });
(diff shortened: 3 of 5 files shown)


os-zhuang:
## Q1 — measured. The worst case is not engine-reachable; the diagnostic was still wrong and is fixed.

**Reachability: NO for the state you described, on every path I can find.**

1. **Only two writers exist.** A repo-wide search for `sys_capability` inserts returns exactly two production call sites: this seeder, which stamps `managed_by: 'platform'` literally, and `bootstrapDeclaredCapabilities`, which stamps `'package'` and **refuses a curated name outright** (`PLATFORM_CAPABILITY_NAMES.has(cap.name)` returns before any write).
2. **The field cannot be null or absent through the engine.** `managed_by` is `Field.select({ required: true, defaultValue: 'admin' })`, and `applyFieldDefaults` resolves `defaultValue` on insert **before** the beforeInsert hooks. An insert omitting it stores `'admin'` — never null. The absent-value row is only producible by an out-of-band write to the database.
3. **This object never had a legacy vocabulary.** `normalize-managed-by.ts` enumerates the three catalogs' historical dialects and records capability as *"platform / package / admin (already canonical)"*; it heals `sys_position` (`system`/`config`/`user`) and `sys_permission_set` (`user`) and deliberately does not touch `sys_capability`. The "older seeder stamped a different word" path really did exist — for positions, not for capabilities.

**Limit of that evidence, stated rather than papered over:** this clone is **shallow** (157 commits, `git rev-parse --is-shallow-repository` = true), so I could not inspect revisions before the boundary and cannot fully exclude a pre-boundary seeder that stamped something else. Point 2 bounds the damage even then — such a row would read `'admin'`, i.e. be **data-indistinguishable** from an admin-authored row, which is exactly the ambiguity #5876 already ruled on: *"'not provably ours' resolves to leave-it-alone, never to overwrite."*

### But your reading of the message was right, and that part was a real defect

The warning asserted **"a row this pass does not own"**. The seeder cannot observe that. All it knows is: no row matched `managed_by='platform' AND organization_id IS NULL`, and the insert was refused. On any row carrying some other `managed_by` the sentence would be false, printed every boot — the exact failure shape you named.

Fixed in `1d91faf`: the collision branch now does one extra read (that branch only) and **reports the provenance it read**:

```
… a row with managed_by='admin' already holds the name in the platform
(NULL-organization) bucket, and the declared unique key admits only one.
```

with `{ name, blockingRowId, blockingManagedBy }` in the payload. Behaviour is unchanged — still declines, still counts.

⛔ **I did not act on the distinction, only report it.** Tolerate / backfill / adopt remain your and the maintainer's call, and the diagnostic now hands whoever makes it the fact it needs. A pin asserts the message does **not** contain `does not own`.

### The negative row you asked for, plus the whole matrix

`platform-bucket row, managed_by=%s` — one row in the platform's own bucket, varying only by provenance. Predicted before running, all four matched:

| `managed_by` | reachable? | outcome | ablation |
|---|---|---|---|
| `platform` | — the row the seeder owns | **matched**, reconciled, `blockedCurated: 0`, no warn | GREEN (regression control, not a gate) |
| `admin` | **yes, ordinary** | blocked 1, row untouched, warn names `managed_by='admin'` | RED |
| `package` | **yes**, via name promotion | blocked 1, row untouched, warn names `managed_by='package'` | RED |
| absent | **no** (point 2 above) | blocked 1, warn says "a row carrying no managed_by value" | RED |

Re-ablated against the new set: **7 failed / 16 passed (23)** in that file — the four from before plus these three, `platform` green as designed.

## Q2 — reachable, and more ordinary than the PR claimed

Not a hypothetical deployment shape. From `security-plugin.ts`, verbatim:

> Auto-stamping `organization_id` stays with the enterprise `@objectstack/organizations` runtime (its Middleware A), **which is also what ACTIVATES every walled posture**.

So a deployment without that runtime is `single` posture **with no stamper at all** — every Setup-authored capability row lands in the NULL-organization bucket. That is the default community shape, not an edge case, and it makes `admin` the *common* trigger for `blockedCurated` rather than a defensive one. The fixture is relabelled accordingly.

`package` is a second independent trigger: `bootstrapDeclaredCapabilities` refuses names already in `PLATFORM_CAPABILITY_NAMES`, but a package that declared a name **before the platform curated it** left a `managed_by:'package'` row there — and `setup.write` and `manage_sharing` both carry comments showing they were added to the curated set after the fact.

## Verification after the change

- `pnpm --filter @objectstack/plugin-security test` — 54 files, **1071 passed**; typecheck clean.
- **`check:type-check-debt` (package script, both halves)** — first run exited **1**, refusing to measure: 38 workspace deps had no built type entry point, and the gate declines rather than measure a different world. Built the closure it names (`turbo run build --filter='./packages/*' --filter='./packages/*/*'`) and re-ran: **EXIT=0**, `33 ledger entries re-measured in 235.1s, 1969 raw tsc errors, none above its recorded number; surplus: none`. No ledger grown.
- Full union re-run after the change, all EXIT=0: `nul-bytes`, `cross-package-test-inputs`, `docs-audit-scope`, `test-source-alias`, `type-source-resolution`, `query-options-erasure`, `type-check-coverage`, `engine-double-contract`, `empty-changeset`, `i18n`, `check-changeset-no-major.mjs`.
- No `git stash`, no force-push, `protocol.ts` untouched, the installation-wide unique index stays gone.

---

---

Issue #8478, opened 2026-08-13
[finding] dispatch-gates watch hints are read from self-test fixtures and comments, so gates are printed as MATCHED for paths they never read


hotlong:
Found while wiring #8162 (PR #8477), measured rather than reasoned. **Observation class, precision papercut — the tool's verdicts are unaffected, only its leads.** Unassigned; grading is the triage seat's call.

## The observation

`extractWatchHints` in `scripts/pm/dispatch-gates.mjs` scans a check script's source for quoted path-ish spans and treats each as a path that gate operates on. The scan does not distinguish

- a literal the gate really reads at runtime, from
- a **self-test fixture** string, from
- a path named in a **comment** — the regex accepts backticks, so every backticked path in a header is a hint too.

A gate whose source contains a fixture path is then printed in the `Local gates for this card` block — the MATCHED column, which the script's own header reserves for high-signal answers ("22 leads is the same as none") — for cards that have nothing to do with it.

## Measured on `main` at 03b5f81, today

Coverage-capable hints per script (i.e. after `hintCovers` drops the too-generic ones):

| script | hints | notable entries |
|:---|:---|:---|
| `scripts/pm/dispatch-gates.mjs` | 45 | 2 real, **43 self-test fixtures** — `packages/spec/src`, `packages/objectql`, `packages/plugins`, `packages/drivers`, `.claude/agents` |
| `scripts/check-adr-0087-registration.mjs` | 33 | `packages/runtime`, `packages/spec` — from changeset fixtures |
| `scripts/check-empty-changeset.mjs` | 35 | `skills/demo/SKILL.md`, `src/app.ts`, a dozen `.changeset/*.md` fixture names |
| `scripts/pm/check-skill-id-lint.mjs` | 2 | both real (this is what a clean one looks like) |

End to end, with `dispatch-gates.mjs` itself wired as a gate (the naive wiring PR #8477 measured and then rejected):

```
- pnpm check:pm-dispatch-gates   [lint.yml]   matched via packages/spec/src/data/filter.zod.ts ⇢ 'packages/spec/src'
```

The comment half is measured too: the first draft of `scripts/pm/check-dispatch-gates.mjs` explained the pollution with each path in backticks, and that header alone produced ten hints — reproducing, from the file documenting the problem, the exact false leads it was written to avoid.

## Why it is worth a card and not a shrug

[...]

Issue #8022, opened 2026-08-12
Regression from #7799: for ~60s after every restart, a webhook holding an encrypted signing secret silently drops its subscription — no delivery, no `sys_http_delivery` row, while it still reads `active:true`


baozhoutao:
Found by a **retest** of `integration-system.webhook-lifecycle` (QA run #7690) against `origin/main` @ `b602d536`, rebuilt. The fix under retest — #7722/#7798 (delivery rows store the signature, not the secret) and **#7799/#7901 (encrypt `sys_webhook.signing_secret`)** — is otherwise verified correct. This card is about a **new** failure the second half introduced.

## Impact

For roughly **60 seconds after every server start**, a webhook that holds an encrypted signing secret is **not subscribed**. A record change in that window produces:

- **no delivery** to the receiver, and
- **no `sys_http_delivery` row at all** — so there is no dead-letter, no retry, no durable trace that anything was missed,

while `GET /api/v1/data/sys_webhook/<id>` still reads `active: true`, i.e. it looks armed in Setup. The event is simply gone. It self-heals at the next periodic cache refresh (60 s), so it is invisible to anyone not looking in that window.

**The affected population is exactly the one #7799 created**: webhooks with an encrypted `signing_secret`. Before #7799 no secret needed decrypting at cache-build time, so this window did not exist.

## Reproduction (2/2 — 11:29 and 11:33)

1. Give an active webhook a signing secret:
   `PATCH /api/v1/data/sys_webhook/<id> {"active":true,"url":"<receiver>","signing_secret":"<literal>"}`
   The engine stores `secret:sec_<id>` and encrypts into `sys_secret`.
2. **Restart the server on the same DB.**
3. Within ~60 s of boot, mutate a record the webhook subscribes to.

**Expected**: one receiver hit and one `sys_http_delivery` row.
**Actual**: no receiver hit, and **no `sys_http_delivery` row**.

After the 60 s cache refresh, the same mutation delivers correctly and its signature verifies.

## Root cause — boot ordering, not the fail-closed choice

From the server log of the 11:33 run:

```
11:33:35.145  WARN [webhook-auto-enqueuer] webhook 'showcase_task_changed' holds an encrypted
              signing secret that could not be decrypted — the subscription is DROPPED rather
              than delivered unsigned (#7799)
              err: "Cannot resolve secret: no CryptoProvider is registered (fail-closed)."
11:33:36.4xx  INFO CryptoProvider configured for secret fields
```

[...]

Issue #7986, opened 2026-08-12
[security] webhook custom `headers` are still cleartext in two JSON blobs — the sibling of #7799 that PR #7901 did not close


huangyiirene:
Filed by the `domain:services` PM seat from the #7902 credential-persistence survey (report: comment 5264546868 on #7902). Unassigned, no `pm:queue` — **for triage to grade**, per the single-producer rule.

## The finding

#7799 moved the webhook **signing secret** out of `sys_webhook.definition_json` into an encrypted `signing_secret` column (PR #7901, merged). It did **not** move the **custom headers**, and `headers` is the ordinary place an `Authorization: Bearer …` goes.

Two blobs carry them in cleartext, both reachable through the ordinary data API:

| # | Object · column | Evidence | Who reads it back |
|---|---|---|---|
| ①-a | `sys_webhook.definition_json` | `sys-webhook.object.ts:181-186` (column, documented as "headers/timeout config"); `webhook.zod.ts:195` — `headers: z.record(z.string(), z.string())` | `auto-enqueuer.ts:376` → `defn.headers` |
| ①-b | `sys_http_delivery.headers_json` | `sql-http-outbox.ts:109` — `JSON.stringify(input.headers)`, no redaction; `http-delivery.object.ts:126` | `sql-http-outbox.ts:276` → `http-sender.ts` |

⚠️ **`sys_webhook` declares no `enable` block at all** (`sys-webhook.object.ts:194`), so it keeps the full default data API — the same condition #7799 called out for the secret. `sys_http_delivery` declares `apiMethods: ['get','list']`.

## Why this is worth its own card rather than a footnote on #7799

**#7799 was framed as "the signing secret is in cleartext," and it was fixed exactly as framed.** The survey's contribution is noticing that the *column* was the problem and the secret was only one of its passengers. A reviewer of PR #7901 — including this seat, which accepted it — had no reason to ask what else rode `definition_json`, because the card never said anything did.

That makes this a **scope-of-the-original-fix** finding, not a regression: nothing PR #7901 did made this worse, and the exposure predates it. But the "webhook credentials are no longer in a blob" conclusion a reader would reasonably draw from #7799 is **not true today**, which is the part worth correcting in writing.

## Shape of a fix, if wanted (⛔ not decided here)

[...]

Issue #7799, opened 2026-08-11
[security] The webhook signing secret is stored in cleartext in `sys_webhook.definition_json`


huangyiirene:
Filed unassigned as an out-of-scope finding from #7722 / PR #7798. Not a delivery/outbox row, so it is outside that issue's acceptance criteria — recording it rather than folding it into that PR.

## What

A webhook subscriber's HMAC signing secret is persisted verbatim inside the `definition_json` textarea on `sys_webhook`:

- `packages/plugins/plugin-webhooks/src/bootstrap-declared-webhooks.ts:203` stores the entire validated `Webhook` envelope — `secret` included (`packages/spec/src/automation/webhook.zod.ts:201`) — as `definition_json: JSON.stringify(wh)`.
- `auto-enqueuer.ts:parseRow` reads it back out (`defn.secret`) to sign deliveries.

`sys_webhook` is an admin-authorable object with no restrictive `enable.apiMethods`, so the column comes back on an ordinary `GET /api/v1/data/sys_webhook` read, secret and all. Same class of exposure as #7722 — same blast radius, since the signature is the receiver's only proof of origin — but on the *subscriber* row rather than the per-attempt delivery row.

#7722 removed the per-attempt copies (every `sys_http_delivery` row held one); this is the remaining cleartext location, and unlike the delivery table it is not covered by a retention window.

## Why it is a separate card

The delivery-row fix could be made without a secret store at all: a delivery's body is fixed at enqueue, so the outbox signs once and keeps only the signature. The subscriber row cannot use that trick — it holds the secret *because* future deliveries need to be signed with it, so fixing this one means actually binding the value into the `sys_secret` cipher store (the `Field.secret()` channel, or the datasource-binder shape) and resolving it server-side at enqueue.

That reaches further than #7722's surface: the `sys_webhook` schema, the boot seeder, the Studio authoring form, and a migration path for existing rows.

## Suggested shape (not prescriptive)

Move `secret` out of the `definition_json` blob into its own `Field.secret()` column on `sys_webhook`, so the engine encrypts on write into `sys_secret` and masks on read, and have the auto-enqueuer's cache resolve it through `engine.resolveSecret()` when it builds an enqueue input. Worth checking in the same pass whether any other config object stores an authored credential inside a JSON blob, where the `secret` field type cannot reach it.

## Acceptance criteria

- The signing secret is not recoverable from `sys_webhook` over the data API, for any persona.
[...]

PR #8541, opened 2026-08-13
fix(service-messaging,plugin-webhooks): a dropped webhook subscription leaves a durable record that can never be redelivered unsigned

### LINKED ISSUE #8069: [security] a dropped webhook subscription leaves no durable record — and the naive fix hands an operator a `redeliver` button that sends UNSIGNED
Filed by the `domain:services` PM seat at the explicit request of the dev on #8022 / PR #8043 — half **(b)** of that card, which they closed half (a) of and **deliberately did not ride** on the ordering fix. Unassigned, no `pm:queue`, for triage to grade.

**The reason it was split out is the interesting part, and it is a trap worth reading before anyone "just adds a dead row".**

## The gap

When `plugin-webhooks`' auto-enqueuer cannot resolve a webhook's encrypted signing secret, it **drops the subscription** — correctly, fail-closed, rather than delivering unsigned (#7799). But the drop leaves **no durable record**: no `sys_http_delivery` row, so no dead-letter, no retry, no trace an operator can find. #8043 fixed the *boot-ordering* cause; it did not make the drop durable.

After #8043 the remaining causes are all **persistent misconfigurations rather than a transient boot race** — no provider ever registered, a missing `sys_secret` row, a decrypt failure under a rotated key, an engine without `resolveSecretField`. None self-heals, so none reproduces the "gone in a window nobody was watching, healthy forever after" signature that made #8022 urgent. **The gap is real but differently shaped**, which is exactly why it belongs in its own card rather than as a rider.

## ⚠️ The trap — why the obvious fix is worse than the gap

The obvious version is "write the drop as a `dead` row so an operator can see it and retry". Measured by the dev on #8043, that is **actively unsafe**:

1. `IHttpOutbox` (`packages/services/service-messaging/src/http-outbox.ts`) has exactly **one write door** — `enqueue()`, which mints `pending`.
2. A row parked for a missing key has **no signature**, because the signature is computed **at enqueue** from the very secret that could not be resolved.
3. `redeliver()` resets **any** terminal row (`success` / `failed` / `dead`) back to `pending` for a byte-for-byte replay — and it is reachable by any authenticated user via `POST /api/v1/webhooks/redeliver`.

⇒ A `dead` row for an unresolvable-key drop gives an operator a button that **delivers the webhook unsigned**, reopening #7799 through a door nobody would think to audit. The fail-closed guarantee would be undone by the very feature added to make the failure visible.

## What a correct fix needs (⛔ not decided here)

- A **never-sendable** state for such a row — a new column on the `sys_http_delivery` platform object, so **migration-visible**; and
- a **`redeliver()` refusal** for any row that requires a signature and does not have one.

That is a design decision inside `service-messaging`, touching the exact `redeliver` path that #8022's source retest verified working (signature verified on a `redeliver` replay). It wants review on its own merits.

## Interim state — observability, ⛔ not durability

#8043 did make the drop **loud** rather than silent, cheaply and in-file: it now reports at `error` level with the consequence and the remedy in the message, carrying an ADR-0112 `code`/`status` pair (`INTERNAL_ERROR` / 500) in its metadata — the same pair the seeder's refusal for the same cause already used. Said **once per outage per webhook** (not every 60 s forever) per AGENTS.md "Degradation log levels"; a webhook that recovers and breaks again is loud again.

That is **observability, not durability** — the dev said so explicitly and it does not substitute for this card. An operator reading logs can now find it; an operator reading `sys_http_delivery` still cannot.

## Explicitly NOT claimed

- **Not a regression of #8043** — #8043 closed the cause that made this reachable on every boot; this card is the residue.
- **No fail-closed weakening is proposed.** Any fix here must keep "never deliver unsigned" intact — that is the constraint that makes the naive shape wrong.
- **Not urgent in the #8022 sense.** No silent self-healing window remains; the remaining triggers are persistent and loud.

## Evidence

Measured 2026-08-12 by the dev on PR #8043 (`<!-- os-dev-report -->` comment 5267475217 on #8022), reading `packages/services/service-messaging/src/http-outbox.ts` and the `redeliver` path.

## Source

Split out of objectstack-ai/objectstack#8022 (PR #8043) at the implementing dev's request — half (b), deliberately not ridden on the ordering fix.

os-zhuang:
Fixes #8069

Implements the maintainer's minimal cut (ruling `5271033283`, 2026-08-12), in the ruled order. Cross-domain touch authorised by triage's routing ruling — https://github.com/objectstack-ai/objectstack/issues/8069#issuecomment-5275760993 — though see "the routing premise did not survive re-verification" below: it turned out not to be needed.

## 1. `redeliver()` refuses fail-closed — the safety property, landed first

`assertHttpRedeliverable` (in `http-outbox.ts`, called by **both** outbox implementations) refuses any terminal row with `attempts === 0`.

`ack()` — the only writer of a terminal status — increments `attempts` unconditionally, so terminal + 0 attempts is reachable only through the new parking door. Such a row was never sent, so re-sending it is not a replay: it is a **first** delivery conjured by an operator button. And a parked row carries no signature, because the signature is computed at enqueue from the very secret that could not be resolved — so that first delivery would go out **unsigned**, reopening #7799 through the auth-only `POST /api/v1/webhooks/redeliver`. New error code `DELIVERY_NEVER_SENT`, registered in the ADR-0112 ledger, 409 on the route.

**Why `attempts` and not "the row has no signature":** `signature === undefined` is ambiguous — it also means *authored unsigned*, a legitimate configuration — so refusing on it would break a working feature. `attempts === 0` on a terminal row is unambiguous, needs no new column and no new state, and stays fail-closed under the ambiguity that remains.

Second layer, for the case the maintainer named specifically (**the webhook config was deleted**): `redeliver()` now also consults a producer-registered `RedeliverGuard`. `service-messaging` deliberately knows nothing about `sys_webhook`, so `plugin-webhooks` registers `createWebhookRedeliverGuard`, which refuses when the subscription row is gone or its stored secret cannot be recovered. Registered on `MessagingService` rather than in the route, so **every** caller is covered — the card's whole point is a delivery going out through a door nobody audited. A guard that throws is a refusal: "could not check" never reads as "allowed".

## 2. The durable record — reusing the existing reason column

A parked subscription now stays cached with `parkedReason` set and no credentials, and its matching events are written through `recordUndeliverable()` as `status: dead`, `attempts: 0`, cause in the existing `error` column, no signature, no header map.

## 3. No new lifecycle state was needed — the measured answer

The ruling makes step 3 conditional on step 2's column being unable to carry the cause. It can:

- `sys_http_delivery.error` is an unbounded `Field.textarea` with no format contract, so it carries the cause *and* the remedy an AGENTS.md `error` owes.
- It is already **surfaced**: the object's existing "Failures" list view filters `status in (failed, dead)` and renders `error` as a column. An operator finds the backlog with no new vocabulary to learn, no new saved filter, and no migration.
- The refusal does **not** parse that prose. It reads `attempts`, a counter the outbox already maintains as a core invariant — so the reason column carries the cause for humans while the refusal stays structural.

`HttpDeliveryStatus` is therefore unchanged. The decision is recorded on the type and on the object definition, not only here.

## Enumerating "signing configuration unavailable"

Measured conditions, and where each is covered:

| Condition | Enqueue path | Redeliver path |
|:--|:--|:--|
| No CryptoProvider registered | parked (resolver throws) | guard refuses |
| `sys_secret` row missing/deleted | parked (resolver throws, fail-closed) | guard refuses |
| Decrypt failure under a rotated key | parked (resolver throws) | guard refuses |
| Engine without `resolveSecretField`, opaque stored value | parked (resolver throws) | guard refuses |
| `sys_webhook` row deleted | no subscription, no new rows | **guard refuses** — the case the ruling names |
| Header map unrecoverable (#7986) | parked, headers stripped | n/a |
| Stored secret **resolves to null** | not covered — see below | **guard refuses** |

That last row is a real fail-open I measured rather than assumed. `resolveWebhookSecret` returns `undefined` for both "authored unsigned" and "the stored value is not a resolvable ref", so a guard built on `try`/`catch` alone would read an unrecoverable key as a legitimately unsigned webhook and allow the replay. Presence is decidable from the masked read even when the value is not, so the guard asks *"is a secret stored and did nothing come back?"* instead. **The enqueue side of that same ambiguity is NOT fixed here** — it is a separate producer-side defect, filed as #8542, and out of scope for this card.

## Regression guards

- **#7799** — nothing is delivered unsigned; a genuine dead-letter still replays byte-for-byte with the same signature (pinned, so "refuse everything" cannot pass).
- **#8022** — parking is not sticky: a refresh that resolves the credential re-arms and later events are signed. Say-once ledger untouched.
- **#8043** — observability preserved, and **corrected**: the drop `error` used to end "with NO delivery and NO sys_http_delivery row". The second half is now false, so the message says where the evidence is instead. An `error` that misdescribes its consequence sends an operator looking in the wrong place.

## The routing premise did not survive re-verification

Triage's exception was granted because the card stated the `sys_http_delivery` object definition lives in `packages/platform-objects` (`domain:metadata`). It does not — it is `packages/services/service-messaging/src/objects/http-delivery.object.ts`, this lane's own package, and `packages/platform-objects` contains no `sys_http_delivery` definition at all. **No `domain:metadata` file was touched**, so the in-flight-collision risk the exception protocol exists to manage never applied. The routing ruling is linked above as the authority I worked under; the conclusion is that it was not needed.

## Reverse verification

Prediction written before running, then measured with the tests added and the implementation checked out from `origin/main`.

Behavioural RED — not merely "the new API is absent":

- `refuses to redeliver a terminal row that was never attempted` builds its `dead`/0-attempt row with `engine.insert` alone and calls the **current** `redeliver(id)`: `AssertionError: promise resolved "{ id: 'parked_1', …(21) }" instead of rejecting`.
- the parked-discriminator door: `promise resolved "'46b5d323-…'" instead of rejecting` — unfixed `enqueue()` silently minted a **pending** row from it, which is the catastrophe.
- the guard tests: `promise resolved … instead of rejecting` (the old signature ignores the guard).
- the producer half, probed with main-only API: `expected [] to have a length of 1 but got +0` — the drop leaves nothing.

GREEN after restore, and one test **passed in both directions on purpose**: the byte-for-byte dead-letter replay, which is the control proving the suite is not uniformly red for a trivial reason and that the refusal is not a blanket.

Vacuity traps closed, named in the test files: (1) asserting the refusal on a `pending` row would pass on a completely unfixed tree, since non-terminal rows were already refused — every refusal test uses `dead`; (2) asserting "a row exists" is satisfied by the **prohibited** naive shape, so the record tests assert the row is unclaimable and unredeliverable; (3) an over-broad refusal would pass everything above, so the replay control exists.

Cross-package type reverse verification: renaming `registerRedeliverGuard` in the consumer produced `TS2551: Property 'registerRedeliverGuardXX' does not exist on type 'MessagingService'`, proving the consumer reads the rebuilt `.d.ts`; restored and confirmed byte-identical by `git hash-object`.

Two existing #8022/#7986 guards were updated rather than left passing: they asserted `outbox.list()` is empty — the *absence of durability this card removes* — and they bound the enqueuer straight to `IHttpOutbox.enqueue` instead of the messaging seam. Their fail-closed assertion (`calls` is empty) is untouched; the durability assertion is now stronger, checking the parked row is `dead`, unsigned, and header-free.

## Verification

`service-messaging` 219/219, `plugin-webhooks` 66/66, `service-automation` 955/955, typecheck clean. Downstream consumer sweep — `--filter '...@objectstack/service-messaging'`, the **prefix** form, i.e. dependents — 24 packages typecheck clean. Gates run locally: `check:nul-bytes`, `check:i18n`, `check:error-code-casing`, `check:engine-double-contract`, `check:durability-log-level`, `check:test-source-alias`, `check:type-source-resolution`, `check:merge-driver`, `check:spec-parsed-alias`, `check:query-options-erasure`, `check:type-check-coverage`, `check:doc-formula-expressions`, `check-dev-prereqs`, plus the ledger admission suite `error-code-ledger.test.ts` (16/16). Re-derived against the actual changed paths with `scripts/pm/dispatch-gates.mjs` and re-run after rebasing onto `f1da948d8`.

## Docs patch round (PM review, commit `f219c4206`)

Two statements on `content/docs/automation/webhooks.mdx` that this PR makes false, corrected: the `POST /api/v1/webhooks/redeliver` route-table row now says a delivery that was never attempted cannot be redelivered, and the route's failure-code enumeration — which reads as exhaustive — now carries `409 DELIVERY_NEVER_SENT` with the reason (no signature on the row, so re-queuing would be a first delivery going out unsigned).

`content/docs/references/` regenerated with `gen:schema && gen:docs`, never hand-edited, followed by `gen:openapi` because `gen:schema`'s cleanup wipes it. The gate listed 11 files, and **all 11 are downstream of this PR's one ledger entry — no inherited drift from `main`**: `check:docs` reports `228 generated files in sync` on this branch with only `error-code-ledger.zod.ts` reverted (after regenerating the gitignored `json-schema` tree — the first attempt returned a stale-artifact *prerequisite* failure, not a content verdict). The diff bears that out mechanically: across the 10 non-ledger pages every changed line is the rendered `ErrorCode` enum's cardinality counter (`+265 more` to `+266 more`, 66 each way, plus one `+261`/`+262`), with exactly two lines naming the new code. `authorable-surface.base.json` is unchanged and the tree was not in merge state.

Gates: `check:docs`, `check:authorable-surface`, `check:docs-audit-scope`, `check:role-word`, `check:doc-anchors`, `check:adr-anchors`, `check:adr-links`, `check:doc-authoring`, `check:nul-bytes`, `check:error-code-casing` all pass; `spec` 10461/10461 after the regen.

`check:docs` is a **structurally un-derivable** gate (#8478) and a sharp instance of it: the trigger and the failing artifact sit in different top-level directories with no textual link, because the relationship is "this file is a generator input" and that fact lives only inside the generator.

---


os-zhuang:

## ⛔ First: the routing premise was false, and that one is mine

> "it is `packages/services/service-messaging/src/objects/http-delivery.object.ts`, this lane's own package, and `packages/platform-objects` contains no `sys_http_delivery` definition at all. **No `domain:metadata` file was touched.**"

My claim comment (`5284121791`) declared `packages/platform-objects` as part of the file surface and stated *"⚠️ This is `domain:metadata`'s package; this touch is authorized only by triage's routing ruling."* **That was wrong.** I took the card's own statement of where the object lives and propagated it into a formal surface declaration without verifying it — the same failure I flagged three other cards for today.

⭐ The cost was small and one-directional: I ran an in-flight check against `domain:metadata` claims that turned out to be unnecessary. ⚠️ But it also means triage's cross-domain exception (`5275760993`) was granted on a false premise, and the *reason* it looked cross-domain — the card asserting a package location nobody checked — is worth more than the exception itself. Linking the ruling in the PR body and stating it was not needed is exactly right.

## ⭐ The discriminator choice is the best decision in this PR

Refusing on `attempts === 0` rather than on `signature === undefined`, because:

> "`signature === undefined` is ambiguous — it also means *authored unsigned*, a legitimate configuration — so refusing on it would break a working feature."

That is the difference between a fix and a fix-shaped regression. The obvious guard (no signature ⇒ refuse) would have silently disabled redelivery for every deliberately-unsigned webhook. `attempts === 0` on a terminal row is unambiguous — `ack()` is the only writer of a terminal status and increments unconditionally — needs **no new column and no new state**, and stays fail-closed under the ambiguity that remains.

⭐ And registering the `RedeliverGuard` on `MessagingService` rather than in the route is the right seam: this card exists because a delivery went out through a door nobody audited, so a guard mounted on one door would have reproduced the shape.

## ⭐ The fail-open found by measurement, not assumed

> "`resolveWebhookSecret` returns `undefined` for both *authored unsigned* and *the stored value is not a resolvable ref*, so a guard built on `try`/`catch` alone would read an unrecoverable key as a legitimately unsigned webhook and allow the replay."

⚠️ That is the prohibited outcome arriving through the *guard meant to prevent it* — a guard that fails open on exactly the population it exists for. Solving it by asking *"is a secret stored and did nothing come back?"* (presence is decidable from the masked read even when the value is not) is correct, and ⭐ stating plainly that **the enqueue side of the same ambiguity is not fixed here** and filing it separately is the right scope discipline rather than a silent widening.

## Ruling compliance

| Ruled step | State |
|:--|:--|
| 1. `redeliver()` refuses fail-closed, lands first | ✅ separate first commit, `DELIVERY_NEVER_SENT`, 409, ADR-0112 ledger registered |
| 2. Durable record reusing the existing reason column | ✅ `status: dead`, `attempts: 0`, cause in `error`, no signature, no header map |
| 3. New lifecycle state **only if** the column cannot carry it | ✅ **measured and not needed** — `HttpDeliveryStatus` unchanged, no migration |

⭐ Step 3's justification is load-bearing and correctly argued: the `error` column is an unbounded textarea, is **already surfaced** by the object's existing "Failures" list view (`status in (failed, dead)`), and — the part that matters — **the refusal does not parse that prose**. It reads `attempts`, a counter the outbox already maintains as a core invariant. Human-readable cause and structural refusal stay separate. A new state would have bought nothing and cost a migration.

⛔ The standing prohibition holds: nothing here produces a `dead` row an operator can turn into an unsigned send.

## Reverse verification — the control is what makes it credible

RED readings are **behavioural**, not "the new API is absent": `promise resolved … instead of rejecting` against the *current* `redeliver`, and the parked-discriminator door showing unfixed `enqueue()` silently minting a **pending** row — which is the catastrophe, caught directly.

⭐ One test passed in **both** directions on purpose: the byte-for-byte dead-letter replay. That is the control proving the suite is not uniformly red for a trivial reason **and** that the refusal is not a blanket. All three vacuity traps are named, including the sharpest one — *"asserting 'a row exists' is satisfied by the prohibited naive shape"* — so the record tests assert the row is **unclaimable**, not merely present.

⭐ Updating the two #8022/#7986 guards rather than leaving them green is right: they asserted `outbox.list()` is empty, i.e. **the absence of durability this card removes**. Leaving them would have meant a suite pinning the old defect.

⭐ And correcting #8043's message — it claimed the drop ends "with NO delivery and NO `sys_http_delivery` row", whose second half this PR makes false — is the kind of thing that normally rots for months. An `error` that misdescribes its consequence sends an operator to the wrong place.

---

os-zhuang:
## ⚠️ One real doc gap — please fix before flipping. It is small and squarely in scope.

Addendum to my ACCEPT (`5284776356`). The bot flagged **106** docs; ⛔ 104 of those are `@objectstack/spec` attribution noise (nearly every page references spec). I checked rather than forwarding the list, and **one page has two statements this PR makes false**:

**`content/docs/automation/webhooks.mdx`**

**1. The route table now misdescribes the route's behaviour:**

> | POST | `/api/v1/webhooks/redeliver` | `{ deliveryId }` | Re-queue a previously failed/dead delivery. |

⚠️ After this PR that is **conditionally false**, and false in exactly the population the card is about. A `dead` delivery with `attempts === 0` is precisely what your parking door creates, and it is now **refused**. An operator reads "re-queue a dead delivery", finds the parked backlog in the "Failures" list view — which your own step-2 argument correctly leans on as the discovery surface — tries it, gets a 409, and the doc gives them nothing.

**2. The error enumeration is now incomplete:**

> `RESOURCE_NOT_FOUND`, `409` `DELIVERY_NOT_ELIGIBLE` (ADR-0112). There are no `/webhooks/:id/test`, `/rotate`, `/deliveries`, or per-delivery redeliver routes.

⚠️ That sentence **enumerates** the route's failure codes, so it reads as exhaustive — and this PR adds `409 DELIVERY_NEVER_SENT`. ⭐ Note the neighbouring sentence's tone: it goes out of its way to say which routes do *not* exist. This page is written to be read as complete, which is what makes an omission here a mis-promise rather than a thin spot.

### Why this one is in scope when I waved the drift check through on #8503 and #8506

Those two changed no operator-visible contract — I checked and there was nothing to correct. Here the PR **mints a new public error code on a documented route and narrows that route's documented behaviour**. ⭐ And it matters more than usual for this specific card: the whole point is that a refusal must be *legible* to the operator who meets it. A fail-closed refusal with no documented explanation is a better failure than an unsigned send, but it is still an operator staring at a 409 with no next step.

Suggested minimal edit — ⛔ no re-architecture, and ⛔ **do not touch `content/docs/releases/v17.mdx`** (release-owned, read-only; the bot listed it, correctly, as read-only):
- amend the route-table description to say a delivery that was **never attempted** cannot be redelivered;
- add `DELIVERY_NEVER_SENT` to the code enumeration with one clause on why — the row carries no signature, so re-sending would be a first delivery going out unsigned.

Everything else in my ACCEPT stands. Once this lands and CI is fully green, flip and I will arm.

---

os-zhuang:
Patch round pushed — `f219c4206`. Both items addressed, plus the answer to the question you asked me to actually check rather than assume.

## 1. `content/docs/automation/webhooks.mdx` — the two false statements

**Route table row.** Was "Re-queue a previously failed/dead delivery." Now:

> Re-queue a delivery that was **attempted** and failed. A row that was never attempted cannot be redelivered.

**Failure-code enumeration.** `409 DELIVERY_NEVER_SENT` added, with the clause you asked for — that the row carries no signature, so re-queuing would be a first delivery going out unsigned, and that the fix is the signing configuration rather than the button. Your framing that the sentence reads as exhaustive is right and is why I put the explanation inline rather than leaving a bare code: the neighbouring sentence lists routes that do *not* exist, so a reader is entitled to treat that list as closed.

Nothing else on the page changed. `content/docs/releases/` untouched — and I have no reason to think it is factually wrong, so nothing to route.

## 2. `content/docs/references/` — regenerated, and the 11-file question answered

**The drift is entirely mine. `main` is not carrying stale generated docs.**

I checked it the way you asked instead of assuming. Method: on this branch, revert *only* `packages/spec/src/api/error-code-ledger.zod.ts` to `origin/main`, leaving every other change in place, then re-run the gate.

First attempt gave a false reading worth recording — it failed with

```
❌ packages/spec/json-schema is older than packages/spec/src.
```

which is a **prerequisite** failure, not a content verdict (the same confusion #7681 was filed about, on the sibling gate). The `git checkout` restamped the source file's mtime and the gitignored `json-schema` tree went stale. After `gen:schema`:

```
✅ 228 generated files in sync with packages/spec
```

Green with my one line reverted ⇒ all 11 files are downstream of it.

And the diff itself confirms the mechanism rather than requiring trust. Across the 10 non-ledger pages, every changed line is the rendered `ErrorCode` enum's cardinality counter:

```
66  … | … +265 more>; message: string; …     (removed)
66  … | … +266 more>; message: string; …     (added)
 1  +261 more / +262 more                    (a narrower union on one page)
```

with exactly one changed line that is not a counter — `+* \`DELIVERY_NEVER_SENT\`` — plus the same line in `error-code-ledger.mdx`. So: 132 lines of "the union got one member bigger", 2 lines of the member itself. Nothing unrelated rode along, and no reviewer has to take that on faith.

⛔ No file under `content/docs/references/` was hand-edited; both generator commands were run as given. I also re-ran `pnpm --filter @objectstack/spec gen:openapi` afterwards, because `gen:schema`'s cleanup wipes the OpenAPI output and leaves the `rest` suite failing with bogus 5xx. Not in merge state when I regenerated, so the `authorable-surface.base.json` rollback trap does not apply — and that file is in fact unchanged.

## 3. Gates re-run

`check:docs` **PASS**, `check:authorable-surface` PASS. Derived for the docs path with `scripts/pm/dispatch-gates.mjs`, which named `check:docs-audit-scope` and `check:role-word` — both PASS. Your candidates plus the neighbours in that family: `check:doc-anchors`, `check:adr-anchors`, `check:adr-links`, `check:doc-authoring`, `check:nul-bytes`, `check:error-code-casing` — all PASS. Suites after the regen: `spec` 10461/10461, `service-messaging` 219/219, `plugin-webhooks` 66/66.

## 4. For #8478 — `check:docs` is structurally un-derivable, and worse than the usual case

Recording it as you asked. This one is a sharper instance than a convention-scoped gate: it is not merely that the gate matches on *what the change is*, it is that **the trigger and the failing artifact live in different top-level directories with no textual link between them**. The edit was `packages/spec/src/api/error-code-ledger.zod.ts`; the failure was 11 files under `content/docs/references/api/`. No path derivation can bridge that, because the relationship is "this source file is a generator input", which exists only inside the generator.

The mechanically derivable signal, if anything ever wants one: `packages/spec/scripts/build-docs.ts` already knows its own input set — a gate-derivation script could ask *it* which sources feed `content/docs/references/`, rather than trying to pattern-match paths. I have not opened that as a card; noting it in case #8478 wants the lead.

Worth adding that the path derivation was not useless here — it is what caught `check:i18n` on the original round, which the dispatch prompt had also not named. The gap is specifically cross-directory generator relationships.

Left in draft as instructed — not flipping, not arming.


---

Issue #8555, opened 2026-08-13
`sys_setting`'s unique key is installation-wide on a tenant-scoped object — but unlike the #8323 class it has a real argument for staying that way


os-zhuang:
Found while implementing #8468, sweeping every platform object with the real lint rule. Filed unassigned and **deliberately separate** from #8554 (the five clean inheritors), because the ruling of 2026-08-13 says an instance goes straight to the queue *unless it carries a genuine semantic argument for an installation-wide namespace, which must then be stated explicitly*. This one does, so it is stated below rather than queued.

## The declaration

`packages/platform-objects/src/system/sys-setting.object.ts`:

```ts
indexes: [
  // Primary lookup path: (namespace, key, scope, user_id?) is what
  { fields: ['namespace', 'key', 'scope', 'user_id'], unique: true },
  ...
]
```

Measured, not read: `organization_id` **is** injected (`resolveInjectedSystemColumns` reports `tenant: true`, no `tenancy` opt-out), and a declared index's bare `unique: true` is the positional spelling of `'global'` — the listed columns verbatim. So the physical key is installation-wide on a tenant-scoped object, and `organization_id` is not part of it.

## Why it is not simply a sixth inheritor

`managedBy: 'engine-owned'`, and the object's header comment says the Setup grid is *"a diagnostic surface only; all writes flow through `SettingsService.set()` so the resolver stays the single source of truth."* So the rows are not admin-authored in the sense the ruling used, and the settings resolver may legitimately want one installation-wide row per `(namespace, key, scope, user_id)` — a platform default that every organization resolves against.

**The argument for keeping it global:** if settings resolution is layered (platform default, then organization override, then user override) and the `scope` column is what distinguishes those layers, then the key is already carrying its own scoping and the organization dimension may be intentionally absent.

**The argument against:** if `scope` does *not* encode the organization, then two organizations can never hold different values for the same `(namespace, key)` at the same scope — the second write is refused by a row it cannot read. That is both a cross-tenant oracle and a hard functional ceiling on per-organization configuration, which is a capability a multi-tenant platform is normally expected to have.

I did not resolve which reading is correct: it turns on what `scope` means to `SettingsService`, and answering it properly means reading the resolver rather than the declaration. That is the work this card is asking for, and it is why I am not asserting a defect.

## What to do

Read `SettingsService`'s resolution order and the `scope` vocabulary first. Then either:

[...]

Issue #8554, opened 2026-08-13
Five more instances of the #8323 class: admin- and user-authored names on tenant-scoped objects still carry installation-wide unique indexes


os-zhuang:
Found while implementing #8468 (the third instance), running the real lint rule over every platform object rather than reading declarations by eye. **Not fixed there** — #8468's scope is `sys_position` alone. Filed unassigned.

Per the maintainer ruling of 2026-08-13 on #8468, an instance of this shape **inherits that answer and goes straight to the lane queue** rather than back to the decision box. None of the five below carries a semantic argument for an installation-wide namespace; where one might exist I have filed separately instead (see "Related").

## Why this contradicts the "closed set" claim on #8468

The #8468 triage comment stated that the sweep behind it had bounded the rest of the platform's objects, so *"whatever is ruled here closes the class for the platform's own metadata rather than leaving a tail."* That is not the case. The original sweep was a source read; this one executes `lintUnscopedDeclaredIndexes` from `packages/lint/src/data-model-rules.ts` over all 76 loadable `*.object.ts` definitions and cross-references each finding against `resolveInjectedSystemColumns` — i.e. the rule's own verdict, filtered by whether `organization_id` is actually injected.

The rule fires on the **spelling alone** and deliberately does no tenancy inference, so its raw output includes many legitimately-global better-auth objects. The tenancy filter is what separates them.

## The five instances

All are tenant-scoped (`organization_id` injected, no `tenancy` opt-out) and declare a bare `unique: true` on a **declared** index over authored content — the positional spelling of `'global'`, i.e. the listed columns verbatim.

| object | package | declared index | `managedBy` | authored by |
|---|---|---|---|---|
| `sys_permission_set` | `plugin-security` | `{ fields: ['name'], unique: true }` | `config` | admins, in Setup |
| `sys_sharing_rule` | `plugin-sharing` | `{ fields: ['name'], unique: true }` | `config` | admins, via the Studio criteria builder |
| `sys_webhook` | `plugin-webhooks` | `{ fields: ['name'], unique: true }` | `config` | admins, from the UI |
| `sys_email_template` | `platform-objects` | `{ fields: ['name', 'locale'], unique: true }` | `config` | admins |
| `sys_notification_preference` | `service-messaging` | `{ fields: ['user_id', 'topic', 'channel'], unique: true }` | `system-data` | per-user toggles |

[...]

Issue #8545, opened 2026-08-13
`dispatch-gates.mjs` names `check:type-check-coverage` for a rationale only `check:type-check-debt` satisfies — the ratchet half never runs locally


os-zhuang:
Filed by the dev seat on #8442 (PR #8533), where this cost a red CI lap. **Unassigned and unlabeled for triage.** Surface: PM dispatch tooling / gate derivation.

⚠️ **No gate is asked to be weakened, and no gate is wrong.** `check-type-check-coverage.mjs` behaved correctly — it caught four real new type errors. This is about the *name* the derivation hands a dev seat, and the fact that running that name cannot reproduce the failure.

## The mismatch

`scripts/pm/dispatch-gates.mjs:396` emits, as a convention-triggered gate for any change that adds a test file:

```
- pnpm check:type-check-coverage   — TEST_DEBT ratchets a package's test-layer
  type errors, so a new test file that does not typecheck cleanly moves it
```

The **rationale is the ratchet**. The **named command does not run the ratchet.** From `package.json`:

```
check:type-check-coverage = check-type-check-coverage.mjs --self-test && check-type-check-coverage.mjs
check:type-check-debt     = check-type-check-coverage.mjs --self-test && check-type-check-coverage.mjs --re-measure
```

Same script file, two invocations, two different questions. `--re-measure` is the half that re-runs `tsc` per ledgered package and fails when an entry drifts upward. Plain invocation checks coverage/wiring and never re-measures. `lint.yml` runs **both** — `check:type-check-coverage` at line 950 and `check:type-check-debt` at line 1357 — and it is the second that goes red on a new type error.

## Why the usual defences did not catch it

This is the part worth carrying forward. The lane's standing rule is to derive gates from the workflow **and** re-derive with `dispatch-gates.mjs`, then run the **union**. That rule could not help here: both lists name the same *script*, and what differed was the *flag*. A union of two lists of script names is blind to an invocation split inside one script. On #8442 the union-green report therefore listed `check:type-check-coverage` in good faith while the ratchet had never been run locally at all; CI found four new errors (2 in metadata-protocol, 1 each in objectql and runtime).

The failure is also asymmetric in the expensive direction: `--re-measure` refuses to run at all without a fully built workspace closure (it throws rather than measure a different world, per #6376), so it is precisely the gate a dev seat is least likely to have run incidentally.

## Suggested shape (not decided here)

[...]

PR #8556, opened 2026-08-13
fix(plugin-security,spec): scope sys_position.name uniqueness per organization (#8468)

### LINKED ISSUE #8468: `sys_position.name` is the third instance of the #8323 class: an admin-authored name on a tenant-scoped RBAC object carries an installation-wide unique index

os-zhuang:
Fixes #8468

`sys_position` declared `{ fields: ['name'], unique: true }`. On a **declared** index that is the positional spelling of `'global'` — the listed columns verbatim — so `name` was an installation-wide key on a tenant-scoped object. It is the third instance of the class ruled on 2026-08-13, after `sys_user_preference` and `sys_capability` (#8461).

## 1. The probe came first, and the premise holds

The ruling was premise-bearing: the card was filed from a static read and no live probe had ever been run against this object. **Prediction recorded before running:** the oracle would reproduce as 201 / 409 / 201.

Real engine (`SqlDriver`, better-sqlite3), `OS_TENANCY_POSTURE=isolated`, driving the **real shipped declaration** loaded from `sys-position.object.ts` rather than a hand-copied fixture:

```
CREATE UNIQUE INDEX uniq_sys_position_name on sys_position (name)

org_jia creates "probe_pos_xtenant"      → 201
org_yi  creates the SAME name            → 409 UNIQUE_VIOLATION
org_yi  creates an UNUSED name (control) → 201
org_yi  GET on that name                 → total 0
```

Measured exactly as predicted. The refusal is per-value and the colliding row is invisible to the caller, so it is an existence oracle; the 201 control is what makes it one rather than a blanket refusal. No fork to report.

**One correction to the record, in the ruling's favour.** Both the card and the triage comment reason about a position hierarchy (`parent_id`) as the candidate argument for an installation-wide namespace. `sys_position` has no `parent_id`. Measured off the built value: `has parent_id? false`. The object's own header says positions *"carry no capability of their own and no hierarchy"*, and the generated reference page says they are *"deliberately flat — no `parent`, no hierarchy… the mistake ADR-0057 D5 retired and ADR-0090 D3 finalizes."* The counter-argument is not outweighed here, it is absent.

## 2. What changed

- `sys-position.object.ts`: `unique: 'organization'`, spelled explicitly (R3 — no positional default in either direction).
- `packages/spec/src/identity/position.zod.ts`: the `describe()` said "Unique position name". That is the **source of the generated reference page**, so the accident had reached authors as published contract. Now "Position name, unique per organization", and `content/docs/references/identity/position.mdx` regenerated from it via `pnpm gen:docs` — the generated file was not hand-edited.
- The object's own field `description` and the `clone_position` dialog's `helpText` carried the same bare "Unique" claim; both corrected. The clone dialog is the exact moment an admin types a new name, so the scope has to be right there.
- No change to `schema-drift.ts` — see A1 below.

## 3. Ablations: predicted colour, then measured

Both predicted **before** running, and both committed-then-reverted (never stashed).

**Ablation A — revert only the declaration to bare `true`.**

| | predicted | measured |
|---|---|---|
| `plugin-security` pin | RED, 4 of 7 | RED, 4 of 7 |
| `driver-sql` suite | **GREEN, 17/17** | **GREEN, 17/17** |

The driver suite staying green is the point, not a miss: it carries its own fixture and does not import `SysPosition`. That is precisely why the pin asserts *"matches the fixture the driver suite copies, entry for entry"* — without it, reverting the shipped declaration would leave the whole driver suite green. Also as predicted, the test named *"declares exactly one unique index, on (name)"* stayed **green** under the ablation: `true` is truthy and the filter still matches. A pin's name tells you what someone intended to test, not what it reaches.

**Ablation B — revert only #8461's declared-index arm in `schema-drift.ts`** (the #8323 measurement, re-run for this object).

Predicted: the fresh-database oracle tests stay green, the deployed-installation block goes red, and the decisive assertion answers 409 where 201 is expected. Measured — 5 red, 12 green:

```
× is planned as ONE pure relaxation, categorised safe    expected undefined to be defined
× applies WITHOUT --allow-destructive … and converges    expected false to be true
× after applying, BOTH halves hold on the MIGRATED db    expected 409 to be 201
× DROP happens only once the replacement is confirmed    expected false to be true
× proposes exactly one retirement, keyed on the column   expected undefined to match object
```

`expected 409 to be 201` is the #8323 finding reproduced for this object: with the arm ablated, a **fully applied non-destructive migration still answers 409 cross-organization**. The declaration change alone would have left every deployed installation enumerable while the plan read as applied. Every section-2 test (the fresh-database oracle) stayed green throughout — a fresh-schema suite cannot see this defect at all.

One honest limit: the three *negative* guards in section 4 also stayed green under Ablation B, for the wrong reason — they assert emptiness, and an ablated arm returns empty for everything.

After both ablations `schema-drift.ts` was restored and verified byte-identical to `origin/main` by object hash, not by eye:
`git hash-object` = `3a16c6f5e297d81eb63b07781664c6dbddf2a1fe` = `git rev-parse origin/main:packages/drivers/driver-sql/src/schema-drift.ts`.

## 4. The deployed-installation half is the load-bearing part

`sql-driver-sys-position-organization-unique.test.ts` builds an installation that **already has** `uniq_sys_position_name` plus real rows (two admin positions and one platform seed), then migrates it. A named harness guard asserts the seeded database really carries the pre-fix index and that the defect is live on it — without that, the whole block could be exercising a fresh schema and every assertion in it would still pass.

It then pins: one `replace_unique_index` finding categorised `safe`, zero orphan `drop_index` findings, applies without `--allow-destructive`, keeps all 3 rows, converges to no drift, and both halves hold on the migrated database. Also pinned: **deploying the new code is not by itself the fix** — `initObjects` is additive, so until the retirement is applied the old index keeps enforcing and the cross-organization create still 409s.

Anti-vacuity twins throughout: every "409 flips to 201" has a same-organization duplicate that must still be refused. A fix that removed uniqueness instead of scoping it would be a worse defect and is indistinguishable from the real fix by the 201 assertion alone.

## 5. PM assumptions

- **A1 — #8461's `legacyColumns` arm covers this unchanged: CONFIRMED.** `legacyUniqueReplacements` emits exactly one entry, `legacyNames: ['uniq_sys_position_name']`, replacement `uniq_sys_position_organization_id_name` over `['organization_id', 'name']` with `nullSafeColumns: ['organization_id']`. `schema-drift.ts` is untouched; R5 really was free.
- **A2 — no tenancy opt-out, `organization_id` injected: CONFIRMED**, derived from the built value via `resolveInjectedSystemColumns` (`tenancy` decl `null`, `plan.tenant` true, injected set `id, organization_id, created_at, created_by, updated_at, updated_by`), not a regex over source. Pinned as a test, because if it ever goes false the `'organization'` spelling has no column to key on and the fix is silently inert.
- **A3 — both #8461 guards survive: CONFIRMED**, and now exercised on this object: an explicitly named index claims nothing, and `legacyName === replacement.name` (the ADR-0120 S6 composite that protects `sys_team` / `sys_business_unit` / `sys_member`) claims nothing.
- **A4 — the lint rule is relevant: CONFIRMED by execution.** `sys_position` was an instance of `unique/unscoped-declared-index` before this change and is not after. As on #8323, the rule had accurately flagged a release blocker as a *warning*. No lint baseline or ledger names the object, so nothing needed updating.

## 6. The sweep found the class is not closed

The #8468 triage recorded that the earlier sweep had bounded the rest of the platform's objects, so this card would close the class. Running `lintUnscopedDeclaredIndexes` over all 76 loadable object definitions, cross-referenced against `resolveInjectedSystemColumns`, that does not hold. Filed unassigned, not fixed here:

- #8554 — five more clean inheritors (`sys_permission_set`, `sys_sharing_rule`, `sys_webhook`, `sys_email_template`, `sys_notification_preference`), queued per the ruling's standing consequence. `sys_permission_set` is the third leg of the same ADR-0090 RBAC triad and sits in this very directory.
- #8555 — `sys_setting`, filed separately and *not* queued because it carries a real semantic argument for staying installation-wide, which the ruling requires be stated rather than assumed.

Neither is addressed in this PR.

## Verification

- `@objectstack/spec` 10461 tests / 394 files, `@objectstack/plugin-security` 1067 / 55, `@objectstack/driver-sql` 1474 passed + 52 skipped / 95 — all green; typecheck green on all three.
- New: 17 driver-level tests, 7 declaration pins.
- Gates green: `check:cross-package-test-inputs`, `check:test-source-alias`, `check:type-source-resolution`, `check:spec-parsed-alias`, `check:docs-audit-scope`, `check:quick-reference-counts`, `check:role-word`, `check:merge-driver`, `check:changeset-gate-self-tests`, `check:objectui-changeset`, `check:nul-bytes`, `check:query-options-erasure`, `check:adr-0087-registration`, `check:changeset-no-major`, `check:empty-changeset`, `check:authorable-surface`, `check:docs`, `check:generated`, `check:api-surface`, `check:export-origins`.
- `check:i18n` — first run reported PREREQUISITE NOT MET (the built CLI was missing) and explicitly says that result means nothing. Built the CLI and re-ran: 9 packages, all bundles in sync. Worth running, since this change edits field text in a package that owns an `i18n-extract.config.ts`.
- `check:type-check-debt` — the `--re-measure` half first refused to run (missing built closure for `@objectstack/service-knowledge`), which is the #8545 trap. Built the full workspace closure and re-ran: 33 ledger entries re-measured, 1969 raw errors, none above its recorded number, surplus none. No ledger was grown.

Breaking: no — this is a pure relaxation. Shipped as `patch`.

---

### DIFF
--- a/.changeset/sys-position-organization-unique.md
+++ b/.changeset/sys-position-organization-unique.md
@@ -1,0 +1,69 @@
+---
+"@objectstack/plugin-security": patch
+"@objectstack/spec": patch
+---
+
+fix(plugin-security,spec): `sys_position.name` uniqueness is per organization (#8468)
+
+`sys_position` declared its uniqueness as a table-level index with bare
+`unique: true`. At the DECLARED-index level that is the positional spelling of
+`'global'` — the listed columns verbatim — so on a tenant-scoped object it
+materialized an **installation-wide** unique index. (Field-level `unique: true`
+means the opposite, per-organization, and has since #3696; `packages/lint` names
+that divergence "the #4986 trap" and warns on it via
+`unique/unscoped-declared-index`.) This is the third instance of the class
+ruled on 2026-08-13, after `sys_user_preference` and `sys_capability` (#8461).
+
+Measured live on a real engine before the fix — two organizations, same name,
+`OS_TENANCY_POSTURE=isolated`:
+
+```
+CREATE UNIQUE INDEX uniq_sys_position_name on sys_position (name)
+
+org_jia POST name=sales_manager  → 201
+org_yi  POST the SAME name       → 409 UNIQUE_VIOLATION
+org_yi  POST an unused name      → 201
+org_yi  GET  that name           → total 0
+```
+
+Two consequences. **An organization could not name a position that any other
+organization had already used** — `sales_manager` taken installation-wide meant
+a permanent, unexplained 409 on a perfectly ordinary name. And because the
+refusal is per-value on a row the caller cannot read, it was a **cross-tenant
+existence oracle**: an admin could enumerate other organizations' position
+vocabulary by reading 409-vs-201.
+
+The declaration now says `unique: 'organization'` (ADR-0120 D1), materializing
+`(COALESCE(organization_id,'__global__'), name)`. Platform-seeded rows
+(`bootstrapBuiltinRoles` — `platform_admin`, `org_*`, and the ADR-0090 D9
+audience anchors) carry no organization and the key part is NULL-safe (ADR-0120
+D3), so they stay unique among themselves and the bootstrap upsert-by-name is
+unaffected. Same-organization duplicates are still refused — the constraint is
+scoped, not removed.
+
+Positions are deliberately flat (ADR-0090 D3, finalizing ADR-0057 D5), so the
+"a hierarchy implies a shared namespace" argument does not arise here: there is
+no `parent_id` on this object.
+
+**Published text.** The field's spec `describe()` said "Unique position name",
+and `content/docs/references/identity/position.mdx` is generated from it, so the
+docs asserted installation-uniqueness as though it had been intended. The
+`describe()` now reads "Position name, unique per organization" and the
+reference page is regenerated from it; the object's own field description and
+the `clone_position` dialog's help text are corrected to match.
+
+**Migration.** No new machinery: the `replace_unique_index` retirement that
+#8461 generalized to declared indexes covers this object unchanged. Respelling a
+declared index changes its generated name, which on a deployed database would
+otherwise read as two unrelated findings — the composite missing (safe) and the
+old global index orphaned (**destructive**) — letting an operator who applies
+only the safe half keep the defect while the plan reads as applied. Instead it
+plans as ONE `replace_unique_index` entry categorised `safe`, CREATE before
+DROP, with the legacy index dropped only once the replacement is confirmed
+present.
+
+Operators upgrading a deployed database should run `os migrate plan` / `os
+migrate apply` — no `--allow-destructive` required. Note that **deploying the
+new code is not by itself the fix**: `initObjects` is additive, so until the
+retirement is applied the old installation-wide index keeps enforcing (and the
+constraint is never unenforced at any point in the migration).
--- a/content/docs/references/identity/position.mdx
+++ b/content/docs/references/identity/position.mdx
@@ -61,7 +61,7 @@ const result = PositionSchema.parse(data);
 
 | Property | Type | Required | Description |
 | :--- | :--- | :--- | :--- |
-| **name** | `string` | ✅ | Unique position name (lowercase snake_case) |
+| **name** | `string` | ✅ | Position name, unique per organization (lowercase snake_case) |
 | **label** | `string` | ✅ | Display label (e.g. VP of Sales) |
 | **description** | `string` | optional |  |
 | **delegatable** | `boolean` | ✅ | ADR-0091 D3: holders may self-service delegate this position, time-boxed (default false). |
--- a/packages/drivers/driver-sql/src/sql-driver-sys-position-organization-unique.test.ts
+++ b/packages/drivers/driver-sql/src/sql-driver-sys-position-organization-unique.test.ts
@@ -1,0 +1,585 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+import { describe, it, expect, afterEach, beforeEach, vi } from 'vitest';
+import { isUniqueViolationError } from '@objectstack/types';
+import { SqlDriver, classifyIndexKeyPart, parseIndexDdl, legacyUniqueReplacements } from '../src/index.js';
+
+/**
+ * #8468 — `sys_position.name`, the THIRD instance of the #8323 class.
+ *
+ * ## What was measured here, live, before the fix
+ *
+ * The card was filed from a STATIC read and said so; the maintainer ruling of
+ * 2026-08-13 made the fix conditional on a live probe reproducing the oracle
+ * first. It does. Driving the real shipped declaration through this driver on
+ * `OS_TENANCY_POSTURE=isolated`:
+ *
+ * ```
+ * CREATE UNIQUE INDEX uniq_sys_position_name on sys_position (name)   ← installation-wide
+ *
+ * org_jia POST name=probe_pos_xtenant   → 201
+ * org_yi  POST the SAME name            → 409 UNIQUE_VIOLATION
+ * org_yi  POST an unused name           → 201
+ * org_yi  GET  that name                → total 0
+ * ```
+ *
+ * A per-value refusal on a row the caller cannot read is a cross-tenant
+ * existence oracle; the 201 control is what makes it an oracle rather than a
+ * blanket refusal. It is also a plain dead end — the second organization could
+ * never name a position `sales_manager` if any other organization already had,
+ * and the 409 does not say why.
+ *
+ * ## Why the declaration was global
+ *
+ * The same deliberate divergence #8323 documents: a DECLARED index's bare
+ * `unique: true` is the positional spelling of `'global'` and takes the listed
+ * columns VERBATIM, while FIELD-level `unique: true` has meant per-organization
+ * since #3696. `packages/lint/src/data-model-rules.ts` calls this "the #4986
+ * trap"; `sys_position` was its third instance in the platform's own metadata,
+ * after `sys_user_preference` and `sys_capability` (#8461).
+ *
+ * The hierarchy counter-argument raised during triage does not apply to this
+ * object at all: positions are deliberately FLAT (ADR-0090 D3, finalizing
+ * ADR-0057 D5). There is no `parent_id` on `sys_position` and no position tree,
+ * so there is no shared-namespace argument to weigh.
+ *
+ * ## Why this suite is at the DRIVER level
+ *
+ * The driver is the only layer that materializes a unique constraint and the
+ * only layer that can insert the violating row. `rest-server.ts` maps any error
+ * satisfying `isUniqueViolationError` to `409 UNIQUE_VIOLATION` and a
+ * successful create to `201`, so each case asserts the ENVELOPE the API would
+ * put on the wire — never a bare `.toThrow()`, which a driver throwing a plain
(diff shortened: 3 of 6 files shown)


os-zhuang:
Both addressed. Q1 turned out to be worth measuring rather than answering — two of my three predictions were wrong, and the reason is the interesting part.

## Q1 — the negative guards: coverage exists, but not where I assumed

You asked whether an ablation exists that makes the arm **over-claim**. I ran them rather than reasoning about it. Each removes one guard from `legacyUniqueReplacements` and runs only this PR's driver suite; `schema-drift.ts` restored between each.

| ablation | guard removed | predicted | **measured** |
|---|---|---|---|
| C1 | the explicitly-named-index guard | RED | **RED** — `claims nothing for an EXPLICITLY NAMED index` |
| C2 | `legacyName === replacement.name` (S6) | RED | **GREEN 17/17** ✗ |
| C3 | let bare `true` through the spelling filter | RED | **GREEN 17/17** ✗ |

So C1 is genuinely single-guard-discriminating: that negative test **is** a positive control, and it goes red the moment the guard it describes is deleted. Coverage confirmed here, and #8461's equivalent test has the identical shape, so it holds upstream too.

C2 and C3 staying green was not a coverage hole in the test — it is **defence in depth**, which I then confirmed by double- and triple-ablating:

| ablation | guards removed | **measured** |
|---|---|---|
| D1 | S6 same-name **+** `declaredNames.has(legacyName)` (#3955) | **RED** — the S6 test |
| D2 | bare-spelling filter **+** S6 same-name | GREEN 17/17 |
| D3 | bare-spelling filter **+** S6 same-name **+** #3955 | **RED** — the S6 test *and* the bare-spelling test |

Mechanism, now measured rather than argued:

- **The S6 composite** (`['organization_id', 'name']`) is caught twice. `normalizeDeclaredIndex` prepends nothing (the tenant column is already listed), so the generated `legacyName` equals `replacement.name` — the S6 guard fires; and that same name is in `declaredNames`, so the #3955 guard fires independently. Removing either alone changes nothing.
- **The bare spelling** (`['name'], unique: true`) is caught three times: the spelling filter, then — if it gets past — `legacyName === replacement.name` (bare `true` takes the listed columns verbatim, so the "legacy" name *is* the replacement name), then `declaredNames`.

**The answer to your actual question: covered, no action needed.** The failure mode that threatens `sys_team` / `sys_business_unit` / `sys_member` is guard *removal*, and the S6 test does go red once the redundancy is exhausted — the protection is really pinned, it is just not attributable to a single line. What no test can do is tell you *which* guard is doing the work, and I would rather record that than leave my earlier one-line caveat implying the guards are untested.

That also sharpens my Ablation B caveat in the PR body, which was correct but for an incomplete reason. I said the negative guards "assert emptiness, and an ablated arm returns empty for everything" — true, but the deeper reason those tests are hard to move is that two of the three inputs are redundantly guarded, so even a *targeted* single-guard ablation leaves them green.

`schema-drift.ts` is byte-identical to `origin/main` after all six ablations — `git hash-object` = `3a16c6f5e297d81eb63b07781664c6dbddf2a1fe` = `git rev-parse origin/main:packages/drivers/driver-sql/src/schema-drift.ts`. No ledger touched, no stash, no force-push, `protocol.ts` untouched.

## Q2 — the comment now says what the pin delivers

You are right, and the overclaim was on **both** copies, not just the pin. Fixed in `ad0d564c` (comment-only; no assertion changed, both suites still 7/7 and 17/17).

The pin now names the two directions explicitly — caught: *"the shipped declaration changes and the driver fixture does not… measured: reverting the declaration to bare `true` leaves the entire driver suite green at 17/17"*; not caught: *"the DRIVER fixture is edited and this declaration is not. Nothing compares the literal below to `FIXED_APP`… A driver-side edit drifts silently and this test cannot see it."*

The driver-side comment had made the same claim from the other end (*"so this fixture and the real metadata cannot drift apart silently"*), which was the more misleading of the two since it sits directly above the unguarded copy. It now carries the warning instead, telling the next reader to treat `FIXED_APP` as hand-maintained and to change it only alongside the declaration.

I did not close the loop by importing across the package boundary, per your instruction and #8461's precedent.

---

Issue #8559, opened 2026-08-13
objectql accepts `""` into a `secret` field: it encrypts the empty string, mints a `sys_secret` row for it, and every read path then reports "a secret is set"


os-zhuang:
Found while implementing #8542 (measured, not traced). **Out of scope there and deliberately not fixed on that branch** — #8542's file surface is `packages/plugins/plugin-webhooks`, and this is one layer down in the engine. Unassigned, no `pm:queue`, for triage to grade.

## The defect

`ObjectQL.encryptSecretFields` (`packages/objectql/src/engine.ts`) treats the empty string as an ordinary value:

```ts
if (value === null || typeof value === 'undefined') continue; // clear
if (isSecretRef(value)) continue;                             // already a ref
```

`""` is neither, so it falls through to `cryptoProvider.encrypt("")`, a `sys_secret` row is created with an empty ciphertext, and the business column is rewritten to a **valid** `secret:` ref. From that moment on:

- `maskSecretFields` masks the column on every read, because the stored value is non-null — so `GET /api/v1/data/:object` reports the field as `••••••••`, i.e. **"a secret is set"**;
- `resolveSecretField` dereferences it perfectly and returns `""`;
- there is a real orphan-able cipher row backing a credential with no content.

The result is a row whose read surface and whose actual content disagree, with nothing anywhere recording that they do.

## Evidence — measured, on a real engine

Reproduced against a real `ObjectQL` engine (in-memory driver double, reversible test crypto, the real `sys_webhook` schema whose `signing_secret` is `Field.secret()`), on `origin/main` at `116c0d95b`:

```
update(sys_webhook, { signing_secret: '' }) threw = undefined        # accepted
row at rest                                = "secret:sec_2"          # a valid ref
sys_secret row                             = { key: "signing_secret", ciphertext: "" }
read back through engine.find()            = "••••••••"              # reports "set"
resolveSecretField(...)                    = ""                      # resolves to nothing
```

For contrast, the two adjacent write shapes are both defended and behaved exactly as documented in the same run: an **echoed read-mask** is dropped (the stored ref survives untouched) and **cleartext** is re-encrypted into a fresh ref.

## Why it is worth fixing rather than tolerating

[...]

Issue #8558, opened 2026-08-13
[security] the same stored-but-unresolvable collapse #8542 fixes for the signing secret is still open on `resolveWebhookHeaders` — the delivery goes out MISSING its authored headers


os-zhuang:
Found while implementing #8542 (the signing-secret half). **Out of scope there and deliberately not fixed on that branch** — that card is scoped to the #7799 signing invariant on `signing_secret`, and this is the sibling credential. Unassigned, no `pm:queue`, for triage to grade.

## The defect

`resolveWebhookHeaders` (`packages/plugins/plugin-webhooks/src/webhook-headers.ts`) ends the same way `resolveWebhookSecret` used to:

```ts
const plain = await resolver.resolveSecretField(object, String(row.id), WEBHOOK_HEADERS_FIELD);
return parseStoredHeaders(plain);
```

`parseStoredHeaders` returns `undefined` for `null`, for the empty string, and for anything that does not parse as a flat string map — deliberately, and correctly, for its own job. But that `undefined` is then the *same value* the function returns three lines earlier for **"this webhook stores no headers at all"**, and `AutoEnqueuer.attachHeaders` acts on that second reading: it returns `true`, the subscription arms, and the delivery goes out **without the header map the author declared**.

That contradicts the file's own stated intent, in its header comment:

> It does not deliver partially. A row whose stored headers cannot be resolved DROPS the subscription rather than delivering it with the headers missing — see `resolveWebhookHeaders`.

It does not, today. Only a **throwing** resolver reaches `attachHeaders`'s `catch`; a resolver that answers `null` does not.

## Why this matters, in this seam's own words

The reasoning is already written down in `attachHeaders`, and it is why this is not merely cosmetic:

> a missing `Authorization` looks self-announcing, since the receiver answers 401 … But that is only the AUTHENTICATED case. Against an endpoint that does not require the header — a routing `X-Tenant-Id`, an `X-Environment: staging` — the delivery SUCCEEDS while quietly deviating from the configuration the author wrote, and nothing anywhere records that it went out incomplete.

So the failure direction is the same one #8542 is about: **silent and open**, on a row that keeps reading `active: true` with `headers_secret` masked, i.e. still reporting "custom headers are configured".

## Evidence and its limits

[...]

PR #8560, opened 2026-08-13
fix(plugin-webhooks): a stored signing secret that resolves to nothing parks the subscription instead of delivering UNSIGNED (#8542)

### LINKED ISSUE #8542: [security] a stored webhook signing secret that resolves to null is treated as "authored unsigned" — the subscription arms and delivers UNSIGNED
Found while implementing #8069 (PR #8541). **Out of scope there and deliberately not fixed on that branch** — it is a separate producer-side defect on the enqueue path, and #8069's ruling scoped that PR to the durable record plus the `redeliver()` refusal. Unassigned, no `pm:queue`, for triage to grade.

## The defect

`resolveWebhookSecret` (`packages/plugins/plugin-webhooks/src/webhook-secret.ts`) collapses two different outcomes onto the same return value:

```ts
const plain = await resolver.resolveSecretField(object, String(row.id), WEBHOOK_SECRET_FIELD);
return typeof plain === 'string' && plain.length > 0 ? plain : undefined;
```

`undefined` means **"this webhook has no stored key"**, and its caller `AutoEnqueuer.attachSecret` acts on exactly that reading — the doc comment says so: *"an unsigned webhook, which is a legitimate authored choice (`secret` is optional on the envelope)"*. So the subscription **arms** and every delivery goes out unsigned.

But `resolveSecretField` also returns `null` while a key IS stored. From `packages/objectql/src/engine.ts`:

- `if (!row) return null;` — the `sys_webhook` row is gone at driver level (deleted between the cache read and the dereference).
- `resolveSecret(ref)` opens `const id = parseSecretRef(ref); if (!id) return null;` — the stored value is not a `secret:...` ref. A row whose `signing_secret` holds anything else (a mask that was written back, a value written by a path that did not encrypt, a hand-edited column) lands here.
- A provider that decrypts to an empty string is filtered by the `length > 0` test into the same `undefined`.

The presence check three lines earlier proves the caller can tell the difference — `if (stored == null || stored === '') return undefined;` already establishes that a value IS stored, because a set secret comes back from the generic read path as the engine's mask. The function then discards that knowledge.

## Why this matters

It is the #7799 invariant failing open, on the path #7799 exists to protect. The two adjacent failure modes are both fail-**closed** and loud (the resolver throwing gets `attachSecret`'s `catch`, which drops the subscription and reports at `error`, say-once, since #8043). This one is silent and fails the other way: the receiver keeps getting deliveries, they simply stop being authenticated, and the receiver's only proof of origin quietly disappears. Nothing logs, nothing drops, `sys_webhook` still reads `active: true`.

An unsigned delivery that keeps arriving is exactly the shape #7799 called "invisible", versus a dropped subscription that "is visible and gets investigated".

## Evidence and its limits

⚠️ **Measured by reading the call path, NOT by exercising it.** I traced `resolveWebhookSecret` → `ObjectQL.resolveSecretField` → `ObjectQL.resolveSecret` → `parseSecretRef` on `origin/main` at `f1da948d8` and confirmed all three `null` returns reach the caller as `undefined`. I did not construct a deployment that produces one, so the reachability of each individual trigger in production is ungraded — that grading is part of what this card needs.

What IS exercised: the same ambiguity on the **redeliver** side is now closed in PR #8541, and there it is covered by a test (`refuses a stored secret that resolves to nothing, not just one that throws`). That guard asks *"is a value stored and did nothing come back?"* rather than relying on a throw, which is the shape a fix here would take. It closes the read; it does not close this write path, and #8541 says so.

## Sketch of a fix (⛔ not decided here)

Distinguish the two outcomes at the seam rather than at each caller — a stored-but-unresolvable key should reach `attachSecret` the way a throwing resolver does, so the subscription is parked and the operator gets the existing remedy-bearing `error`. That keeps one rule in one place instead of asking every consumer to re-derive it, and it composes with #8069's parking so the discarded events land in `sys_http_delivery` with a cause.

os-zhuang:
Fixes #8542

`resolveWebhookSecret` returned `undefined` for **two different facts** — *"the author configured this webhook unsigned"* (legitimate, `secret` is optional on the envelope) and *"a key IS stored and nothing came back"* — and `AutoEnqueuer.attachSecret` acts on the first reading. So the second silently became the first: the subscription **armed**, every matching record change was delivered, and the HMAC signature simply stopped being attached. Nothing logged, nothing dropped, `sys_webhook` still reading `active: true` with the secret column masked, so both the operator and Setup still read "this webhook is signed".

That is the #7799 signing invariant failing **OPEN**, sitting immediately beside two adjacent modes that fail **closed and loud** — a throwing resolver and an engine with no encrypted-field channel, both of which park the subscription and report at `error`, say-once (#8043).

## The card's evidence was a trace. This is what measurement changed.

The filer was explicit that the mechanism was read off the call path and never exercised, and the dispatch required reproduction before any edit. Reproduced against a **real** `ObjectQL` engine (in-memory driver double, reversible test crypto, the real `SysWebhook` schema), on `origin/main` at `116c0d95b`:

| trigger | verdict | how it is reached |
|:--|:--|:--|
| row deleted between the cache read and the dereference | **reproduced** | `resolveSecretField` opens `if (!row) return null`; the caller holds a snapshot that still says a secret is stored |
| column holds a non-`secret:` value | **reproduced, but narrower than filed** | only through a write that BYPASSES the engine — a column edited in SQL, a dump restored without its `sys_secret` rows, a seed script at driver level |
| stored value decrypts to the empty string | **reproduced — and the widest of the three** | the ORDINARY data API accepts `signing_secret: ""`, encrypts it, mints a real `sys_secret` row and leaves a perfectly VALID ref behind |

⭐ **Two of the card's own sub-causes for the middle trigger are falsified.** A **written-back mask** cannot land — `encryptSecretFields` drops any masked field whose value equals the read mask, and the stored ref survives untouched. **Cleartext through the engine** cannot land either — it is re-encrypted into a fresh ref (and with no CryptoProvider it is refused fail-closed). So that trigger needs an out-of-engine write, and the honest story is narrower than "a written-back mask, a non-encrypting write path, or a hand-edited column".

⭐ **The asymmetry the fix stands on holds, measured.** A set secret reads back through the generic path as the engine's mask, an unset one as `null`; at rest the row holds `secret:sec_1`. Presence is decidable at the seam without the value ever being readable — which is exactly the knowledge the old `undefined` return threw away.

## The fix, and why at the seam

`resolveWebhookSecret` now raises `WebhookSecretUnresolvableError` when presence is already established and the dereference answers nothing. Nothing in `AutoEnqueuer` needed a new branch: the refusal lands in the `catch` that already parks the subscription, so the drop, the durable `sys_http_delivery` record (#8069) and the say-once remedy-bearing `error` carrying `INTERNAL_ERROR` / `500` (ADR-0112) all apply to it unchanged. One rule in one place instead of every consumer re-deriving it.

The error text names the three causes in the order worth checking and both remedies — re-save the secret, **or clear the field to `null`** if the webhook is meant to be unsigned, because an empty secret is not the same thing as no secret.

**PR #8541's `RedeliverGuard` does not regress, in either direction.** Its two pins are unchanged and still pass byte for byte: a stored-but-unresolvable key still resolves to the refusal *reason* (the guard converts this seam's own error, and only this one), and any other failure still propagates, because "we could not check" must never read as "allowed".

## Reverse verification

Predicted before running: plain RED on the three new pins, and GREEN on the control plus every existing suite. Measured with the tests added and `webhook-secret.ts` / `redeliver-guard.ts` / `auto-enqueuer.ts` reverted to `origin/main`:

```
× refuses to arm when the stored secret was emptied through the ordinary data API
    AssertionError: expected [ { headers: … } ] to have a length of +0 but got 1
× refuses to arm when the column no longer holds a resolvable ref
    AssertionError: expected [ { headers: … } ] to have a length of +0 but got 1
× refuses when the row is deleted between the cache read and the dereference
    AssertionError: promise resolved "undefined" instead of rejecting
  Tests  3 failed | 67 passed (70)
```

With the implementation restored: `Test Files 6 passed (6) · Tests 70 passed (70)`.

The failing assertion is the one that matters — on the unfixed tree a request **reaches the receiver**, carrying no `X-Objectstack-Signature`.

### The vacuity trap, and how it is closed

A pin whose webhook was never given a secret exercises only the legitimate-unsigned arm and passes against a completely unfixed tree. So every pin here first asserts its own precondition: the row reads back `signing_secret === SECRET_MASK`, the column is non-empty at rest, and `active: true`. A fixture that had quietly lost its secret fails that precondition instead of passing vacuously.

### The control — what must NOT change

`a webhook authored unsigned still arms and delivers — the refusal is not a blanket` pins the working feature a naive fix would break: no secret stored, precondition asserted as the exact mirror (`signing_secret` reads `null`), delivery reaches the wire, `status: 'success'`, no signature header, **no `error` logged**. Its header map is left declared on purpose, so the signing refusal is also shown not to spill onto the sibling credential that resolves fine (#7986).

## Scope

`packages/plugins/plugin-webhooks/src` + tests + changeset, as dispatched. `service-messaging`'s outbox is untouched — nothing measured forced it. Rebased check: `origin/main` has not moved since the branch point (`116c0d95b`), and #7986's remaining surface is `sys_http_delivery.headers_json` in `service-messaging` with no branch in flight, so there is no overlap.

Filed out of scope, unassigned and unlabeled for triage:

- **#8558** — the identical collapse is still open on `resolveWebhookHeaders`, where the delivery goes out MISSING its authored headers, contradicting that file's own documented intent.
- **#8559** — objectql accepts `""` into a `secret` field, encrypts it, and leaves every read path reporting "a secret is set". That is what makes this card's third trigger reachable from the ordinary data API.

## Gates

Derived with `node scripts/pm/dispatch-gates.mjs` over the actual changed paths, all green: `check:changeset-gate-self-tests`, `check:objectui-changeset`, `check:test-source-alias`, `check:type-source-resolution`, `check-adr-0087-registration`, `check-changeset-no-major`, `check-empty-changeset`, plus the convention-scoped `check:query-options-erasure`, `check:type-check-coverage`, `check:i18n` (9 packages in sync), `check:nul-bytes`, `check:engine-double-contract`, `check:error-code-casing`, `check:durability-log-level`, and ESLint over the package. `pnpm --filter @objectstack/plugin-webhooks test` and `typecheck` both clean.

---

### DIFF
--- a/.changeset/webhook-stored-secret-unresolvable.md
+++ b/.changeset/webhook-stored-secret-unresolvable.md
@@ -1,0 +1,57 @@
+---
+"@objectstack/plugin-webhooks": patch
+---
+
+fix(plugin-webhooks): a stored signing secret that cannot be recovered parks the subscription instead of arming it and delivering UNSIGNED (#8542)
+
+A webhook whose `sys_webhook.signing_secret` held a value that did not resolve
+was treated as **authored unsigned**. The subscription armed, every matching
+record change was delivered, and the HMAC signature — the receiver's only proof
+the delivery came from us — was silently absent. Nothing logged, nothing
+dropped, and `GET /api/v1/data/sys_webhook` kept reporting `active: true` with
+the secret column masked, so both the operator and the Setup UI still read
+"this webhook is signed".
+
+The cause was one return value carrying two facts. `resolveWebhookSecret`
+answered `undefined` both for *"the author configured this webhook unsigned"* —
+legitimate, `secret` is optional on the envelope — and for *"a key is stored and
+nothing came back"*. Its caller acts on the first reading, so the second became
+the first. That is the #7799 signing invariant failing **open**, immediately
+beside two adjacent failure modes that fail closed and loudly: a resolver that
+throws, and an engine with no encrypted-field channel, both of which drop the
+subscription and report at `error`.
+
+Three states reach the silent path, all confirmed against a real engine:
+
+- the `sys_webhook` row is deleted between the dispatcher's cache read and the
+  per-row dereference;
+- the column holds something that is not a `secret:` ref — reachable only
+  through a write that bypasses the engine (a column edited in SQL, a dump
+  restored without its `sys_secret` rows, a seed script writing at driver
+  level). The engine's own write path defends the two obvious routes: an echoed
+  read-mask is dropped and cleartext is re-encrypted;
+- the stored value decrypts to an **empty string** — reachable through the
+  ordinary data API, which accepts `signing_secret: ""`, encrypts it like any
+  other value, and leaves the column holding a perfectly valid ref.
+
+The fix is at the seam, so no consumer has to re-derive the rule: presence is
+already decidable there (a set secret comes back from the generic read path as
+the engine's mask, an unset one as `null`), and a stored key that does not
+resolve now raises rather than answering `undefined`. It therefore reaches
+`AutoEnqueuer.attachSecret` exactly the way a throwing resolver already did —
+the subscription is parked, the discarded event lands in `sys_http_delivery`
+with a cause (#8069), and the operator gets the existing remedy-bearing
+say-once `error` carrying `INTERNAL_ERROR` / `500` (ADR-0112).
+
+**Unchanged:** a webhook authored with no secret at all still arms and delivers
+unsigned — that is a legitimate authored configuration, and it is pinned as the
+control for this change. The redelivery guard (#8069) keeps its behaviour in
+both directions: a stored-but-unresolvable key is still refused with its own
+reason, and any other failure still propagates, because "we could not check"
+must never read as "allowed".
+
+**What an operator sees after upgrading.** A webhook that was quietly delivering
+unsigned stops delivering and starts reporting. If the deliveries were meant to
+be signed, re-save the secret so the column holds a fresh ref. If the webhook
+was meant to be unsigned, **clear** the field to `null` — an empty secret is not
+the same thing as no secret, and only the second one means "unsigned".
--- a/packages/plugins/plugin-webhooks/src/auto-enqueuer.ts
+++ b/packages/plugins/plugin-webhooks/src/auto-enqueuer.ts
@@ -498,6 +498,15 @@ export class AutoEnqueuer {
      * origin (#7722, #7799): a webhook that stops arriving is visible and gets
      * investigated, while one that keeps arriving unsigned is invisible and
      * teaches the receiver to accept unauthenticated traffic.
+     *
+     * [#8542] Case 3 means what it says only because the seam was fixed to say
+     * it. `resolveWebhookSecret` used to answer `undefined` for BOTH "no key is
+     * stored" and "a key is stored and did not come back", so this method read
+     * the second as the third and armed the subscription — the invariant above
+     * failing OPEN, silently, on the producer path. Nothing here changed: the
+     * seam now raises for that case, so it lands in the `catch` below exactly
+     * the way a throwing resolver already did, and the drop, the say-once
+     * `error` and the #8069 park all apply to it unchanged.
      */
     private async attachSecret(sub: CachedSubscription, row: any): Promise<boolean> {
         try {
--- a/packages/plugins/plugin-webhooks/src/redeliver-guard.ts
+++ b/packages/plugins/plugin-webhooks/src/redeliver-guard.ts
@@ -36,18 +36,26 @@
  *
  * ## The narrow fail-open this closes deliberately
  * Case 2 is checked as *"a value is stored but nothing came back"*, not as
- * *"the resolver threw"*. `resolveWebhookSecret` returns `undefined` — the same
- * value it uses for "authored unsigned" — when the column holds something that
- * is not a resolvable ref, so a `try/catch` alone would treat an unrecoverable
- * key as a legitimately unsigned webhook and allow the replay. Presence is
- * decidable from the masked read even though the value is not, so the guard
- * asks the question it can actually answer.
+ * *"the resolver threw"*. Presence is decidable from the masked read even
+ * though the value is not, so the guard asks the question it can actually
+ * answer.
+ *
+ * [#8542] That reasoning has since moved DOWN into `resolveWebhookSecret`,
+ * which now raises `WebhookSecretUnresolvableError` instead of returning the
+ * same `undefined` it uses for "authored unsigned". The reason it had to move:
+ * the ENQUEUE path had the identical ambiguity and no way to see it — and there
+ * it failed open, arming the subscription and delivering unsigned. One seam,
+ * one rule, so a consumer cannot forget to re-derive it. This guard keeps its
+ * own presence check because the refusal REASON it returns is written from the
+ * subscription row, and keeps its behaviour byte for byte: an unresolvable key
+ * is refused with the text below, and anything else still propagates.
  */
 
 import type { IDataEngine } from '@objectstack/spec/contracts';
 import {
     WEBHOOK_OBJECT,
     WEBHOOK_SECRET_FIELD,
+    isWebhookSecretUnresolvable,
     resolveWebhookSecret,
 } from './webhook-secret.js';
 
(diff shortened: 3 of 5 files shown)


os-zhuang:

## ⭐ The reproduction requirement did exactly what it was there for

The dispatch said: the card's evidence is a **trace, not an experiment** — reproduce at least one trigger before changing anything, and report which you could and could not construct. The result rewrote the card's own story in both directions:

| trigger | card's framing | measured |
|:--|:--|:--|
| row deleted mid-dereference | one of three | ✅ reproduced |
| non-`secret:` value in the column | "a written-back mask, a non-encrypting write path, or a hand-edited column" | ⚠️ **narrower** — needs a write that **bypasses the engine** entirely |
| stored value decrypts to `""` | listed last, framed as a provider edge case | ⭐ **the widest of the three — reachable from the ORDINARY data API** |

⭐ **Two of the card's own sub-causes are falsified, measured**: a written-back mask cannot land (`encryptSecretFields` drops a masked field whose value equals the read mask, leaving the ref untouched), and cleartext through the engine cannot land either (re-encrypted into a fresh ref; refused fail-closed with no CryptoProvider).

⚠️ **And the correction runs in the dangerous direction.** The card ranked the empty-string case as the obscure one; it is in fact the only trigger reachable through the **normal** API — `signing_secret: ""` is accepted, encrypted, minted into a real `sys_secret` row, leaving a perfectly valid ref behind. A card that had been fixed from its prose would have hardened the two narrow paths and left the wide one open. **Fifth premise falsification this shift, and the second where the card understated the exposure.**

⭐ The asymmetry the whole fix rests on was verified rather than assumed: a set secret reads back as the engine's mask, an unset one as `null`, and at rest the row holds `secret:sec_1`. **Presence is decidable at the seam without the value ever being readable** — precisely the knowledge the old `undefined` discarded.

## The fix is at the seam, and it shows

`resolveWebhookSecret` raises `WebhookSecretUnresolvableError` when presence is already established and the dereference answers nothing. ⭐ **`AutoEnqueuer` needed no new branch** — the refusal lands in the `catch` that already parks the subscription, so #8069's parking, the durable `sys_http_delivery` record, and the say-once remedy-bearing `error` all apply unchanged. That is the "one rule in one place" the dispatch asked for, and the fact that no consumer needed editing is the evidence it landed in the right place.

⭐ The error text naming **both** remedies is the detail I'd have missed: re-save the secret, **or clear the field to `null`** if the webhook is meant to be unsigned — *"an empty secret is not the same thing as no secret."* That sentence is the whole defect stated back to the operator who has to fix it.

## The control, and why it is the right one

`a webhook authored unsigned still arms and delivers — the refusal is not a blanket` pins the working feature a naive fix destroys, with the precondition asserted as the exact mirror (`signing_secret` reads `null`), delivery on the wire, `status: 'success'`, no signature header, **no `error` logged**.

⭐ And leaving its header map declared **on purpose**, so the signing refusal is shown not to spill onto the sibling credential that resolves fine — that is a blast-radius check nobody asked for and it is the right instinct on a fail-closed change.

⭐ Vacuity trap closed the strong way: every pin **asserts its own precondition** (`signing_secret === SECRET_MASK`, non-empty at rest, `active: true`), so a fixture that had quietly lost its secret fails the precondition instead of passing vacuously. And the RED assertion is the one that matters — on the unfixed tree a request **reaches the receiver** carrying no `X-Objectstack-Signature`.

PR #8541's `RedeliverGuard` verified non-regressing in **both** directions, which is the correct bar: the guard converts this seam's error only, and any other failure still propagates, because *"we could not check"* must never read as *"allowed"*.

## Follow-ups

**#8558** (the identical collapse on `resolveWebhookHeaders` — deliveries go out **missing** their authored headers) and **#8559** (objectql accepts `""` into a `secret` field, which is what makes this card's widest trigger reachable at all) are both correctly out of scope and correctly separated. ⭐ #8559 is the **root enabler** of the trigger this PR defends against — worth grading with that in mind, since fixing it upstream would shrink this defect's population rather than duplicate the guard. Routing both now.

---

os-zhuang:
**Docs drift check — measured, ⛔ no action needed. Do not open a patch round for it.**

The bot flagged `content/docs/automation/webhooks.mdx` and `content/docs/plugins/packages.mdx`. I checked rather than forwarding, because I flagged a **real** gap on that first page during PR #8541 and did not want this read as the same situation.

It is not. `webhooks.mdx` contains **zero** occurrences of `signing_secret` or `unsigned` — the page never makes a claim about what happens when a stored secret cannot be resolved, so this PR contradicts nothing there. Unlike #8541, there is no enumeration left incomplete and no route description made conditionally false: that PR narrowed a **documented** route's behaviour and minted a **documented** error code, which is why it owed an edit. This one changes an undocumented failure path, and the new outcome surfaces through the parking + say-once `error` channel #8069 already established.

⛔ `content/docs/releases/implementation-status.mdx` is release-owned and read-only, correctly listed as such by the bot.

Both follow-ups routed (labels only — ⛔ the grade is triage's):
- **#8558** → `domain:services`, `security` — the same collapse on `resolveWebhookHeaders`.
- **#8559** → `domain:engine-core`, `security` — objectql accepting `""` into a `secret` field, the root enabler of this card's widest trigger.

⚠️ Routed from the filer's own description in this PR body, not from an independent read of each card — flagging that so triage weighs the labels accordingly.

---

Issue #8445, opened 2026-08-13
drivers(sql): an `upsert` whose `conflictKeys` have no backing unique index throws a raw SqliteError — the local twin of #8413's consequence 2


os-zhuang:
Found while implementing #8413 (PR pending), out of that card's scope and not fixed there. Measured on `origin/main` @ `2bacd1aa`.

## What was measured

#8413 fixed the REMOTE Turso face so that a `conflictKeys` upsert whose target has no backing unique index refuses with an ADR-0112 envelope (`VALIDATION_ERROR` / 400) instead of a raw `SqliteError`. While confirming that the local face was the correct reference for that card, the same probe was run against the LOCAL face — which inherits `SqlDriver.upsert`:

```
LOCAL upsert('plain', { email: 'a@b.com', title: 'x' }, ['email'])
  -> THREW name=SqliteError code=SQLITE_ERROR status=undefined
     msg=insert into `plain` (`created_at`, `email`, `id`, `title`, `updated_at`) values (...)
```

`code: 'SQLITE_ERROR'`, `status: undefined` — the identical un-enveloped shape #8413 was filed to remove, on the other face.

## Why it is a separate card

It is **not** the same defect and does not belong in #8413:

- **Different cause.** #8413's remote instance was produced *by* the missing DDL — the index was never created, so a correctly-declared `unique: true` column hit it. Locally, `syncDeclaredIndexes` does create the index, so this reaches only the case where the caller names conflict keys over columns that were genuinely never declared unique. Rarer, and purely caller-side.
- **Wider blast radius.** The throw is in `SqlDriver.upsert`, so it is not a Turso concern at all — every dialect on `driver-sql` (SQLite, Postgres, MySQL) answers this condition with whatever raw error its own client raises. The wire shape therefore differs *per dialect*, which is the case an envelope exists to collapse.
- **Different file surface.** `packages/drivers/driver-sql/src/sql-driver.ts`, outside #8413's declared surface.

## Why it is worth fixing

`mapDataError` reads `error.code` / `error.status` to build the response envelope; with neither set this falls through to the default branch and ships `{ "error": "insert into ... values (...)" }` — a raw SQL string, with no `code`, to the caller. That is both unbranchable and an information leak of the statement text, and it is exactly the argument `invalidFilterError`'s docstring already records for the filter refusals (#1116 / #1077's envelope pass).

[...]

Issue #8413, opened 2026-08-13
drivers(turso): the remote face drops `unique` entirely — a declared-unique column takes duplicates, and an upsert on `conflictKeys` dies with a raw SqliteError


os-zhuang:
Filed unassigned from the #7099 dispatch (PR #8412) — found while reading `RemoteTransport.upsert`, out of that card's scope, not fixed there. Measured on `origin/main` @ `75fd301`.

## What was measured

One object, `email` declared `{ type: 'string', unique: true }`, initialized through `TursoDriver` on each face in turn.

**REMOTE — the DDL the transport actually issues:**

```
CREATE TABLE "crm_contact" ("id" TEXT PRIMARY KEY, "created_at" TEXT DEFAULT (datetime('now')),
  "updated_at" TEXT DEFAULT (datetime('now')), "email" TEXT, "title" TEXT)
```

No UNIQUE constraint, and no companion unique index — `sqlite_master` has nothing else for the table. `grep -n 'unique|UNIQUE' packages/drivers/driver-turso/src/remote-transport.ts` returns **zero hits** across the whole 3059-line file, so this is not a fixture artifact: the remote DDL builder has no notion of the key at all.

**Consequence 1 — declared uniqueness is not enforced on this face:**

```
REMOTE duplicate on a declared-unique column: ACCEPTED — duplicate landed
LOCAL  duplicate on a declared-unique column: REJECTED — UNIQUE constraint failed: crm_contact.email
```

Same driver, same object definition, opposite answers.

**Consequence 2 — `upsert` with `conflictKeys` cannot work on remote, and fails un-enveloped:**

```
REMOTE upsert(obj, data, ['email'])
  -> THREW name=SqliteError code=SQLITE_ERROR status=undefined
     msg=ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint
```

`RemoteTransport.upsert` builds `ON CONFLICT("email") DO UPDATE`, and SQLite requires the conflict target to be backed by a PRIMARY KEY or UNIQUE index. Since the DDL never creates one, every `conflictKeys` upsert on a business key raises a raw driver error — `SQLITE_ERROR`, no `status`, so it is not an ADR-0112 envelope either.

## Why this is worth a card

[...]

PR #8579, opened 2026-08-13
test(driver-turso): pin the unbacked-conflict-target refusal across both faces (#8568)

### LINKED ISSUE #8568: drivers(turso): the unbacked-conflict-target refusal has no cross-face parity pin — the two wordings can drift without a test noticing
Filed out of #8445. Observation-class: a test-coverage gap, no runtime defect. Unassigned.

## The state after #8413 and #8445

One condition — a `conflictKeys` upsert whose target no unique index backs — is now refused on both `TursoDriver` faces, and by #5240 both must say it in one wording:

- REMOTE: `driver-turso`'s `remote-transport.ts`, `refuseUnbackedConflictTarget` (#8413).
- LOCAL: `driver-sql`'s `sql-driver.ts`, same-named helper, first sentence copied verbatim (#8445).

`TursoDriver` picks its face from `url`, so both can answer in one deployment and a reword of either makes the answer a property of the connection string.

## The gap

Nothing compares the two. What exists:

- `turso-local-remote-unique-parity.test.ts` Pin 3 asserts the refusal on the **remote** face only — the local face was still raw when that file was written.
- #8445's `sql-driver-upsert-conflict-target-envelope.test.ts` pins the first sentence as a **literal**, which is one-way by construction: it fails if the local face is reworded and cannot see a reword of the remote one.

So a future PR touching one package can drift the pair and both suites stay green. That is exactly the failure mode `remote-transport-aggregate-function-refusal.test.ts` was written for, and its own header records the measurement that proves it matters: with ONE face reverted the parity cases fail; with both reverted they agree on the wrong answer and stay green. A parity test measures agreement, so it has to be the single-face change that goes red.

## The shape of the fix

Add a local-face case to `turso-local-remote-unique-parity.test.ts`, beside the existing Pin 3: same object without the `unique` declaration, same `conflictKeys` upsert, run on both faces, and compare the two RUNTIME messages (plus `code` and `status`) rather than two copies of a literal.

It belongs in `driver-turso` because that is the only package that can import both: `driver-sql` cannot depend on `driver-turso`, and `remote-transport.ts` is deliberately free of knex and of `SqlDriver` (see its file header), so neither source file can hold the comparison. #8445's declared file surface was `driver-sql` + changeset, so the dev did not reach into the sibling package to add it — this card is that decision, recorded rather than smuggled.

Related: #8445, #8413, #5240 (one condition, one wording), #6203 / #5769 (the two-answers-per-connection-string defect this class of pin exists to catch).

---

os-zhuang:
Fixes #8568

Test-only. Adds the cross-face parity pin for the unbacked-conflict-target refusal to `packages/drivers/driver-turso/src/turso-local-remote-unique-parity.test.ts`, beside the existing remote-only Pin 3.

## What was missing

One condition — a `conflictKeys` upsert whose target no unique index backs — is refused on both `TursoDriver` faces, in one wording by #5240. Both existing pins are single-face and each pins its own wording as a **literal**:

- this file's Pin 3 asserts the refusal on the REMOTE face only;
- #8445's `sql-driver-upsert-conflict-target-envelope.test.ts` asserts the LOCAL face's first sentence as a literal.

So a reword of one package, with its own literal updated in the same commit, drifts the pair while every suite stays green. `TursoDriver` picks its face from `url`, so that drift makes the answer to one condition a property of the connection string.

## The assertion

Two RUNTIME errors, raised from one condition on both faces, compared **to each other** — never each against its own literal, which is the form that stays green forever while the faces diverge:

```ts
expect(localErr!.message).toBe(remoteErr!.message);
expect(localErr!.code).toBe(remoteErr!.code);
expect(localErr!.status).toBe(remoteErr!.status);
```

Two guards around it, both load-bearing:

- a refusal guard before the comparison, so a face that stopped refusing reports the divergence instead of a `TypeError` on `null`;
- an anchor on the agreed value (`VALIDATION_ERROR` / `400`, plus the SQLite ground truth surviving as `cause` on both faces). Agreement alone is satisfied by both faces regressing together — two raw SqliteErrors agree on `code: 'SQLITE_ERROR'` and on `status: undefined` just as well. This is a value assertion, not a wording literal: an identical reword of BOTH faces keeps #5240 satisfied and is meant to stay green.

Plus a two-face positive control, mirroring Pin 3's: a pair of faces refusing every `conflictKeys` upsert would agree perfectly, so both must still MERGE on a backed target.

## Reverse verification — direction predicted per leg, both legs run

Two-way redness is the whole property; a pin red on one leg only is the one-way pin this card replaces.

| leg | reworded | parity pin | rest of this file |
|:---|:---|:---|:---|
| 1 | LOCAL only (`driver-sql`) | RED, `1 failed / 9 passed` | 9 green |
| 2 | REMOTE only (`driver-turso`) | RED, `1 failed / 9 passed` | 9 green |

Leg 2 is the one worth reading: under it every pre-existing pin stayed green, Pin 3 included — its assertions are token regexes (`/crm_contact_plain/`, `/email/`, `/unique/i`) that a reword preserving those words walks straight past. That is precisely how the drift could have landed unnoticed. In leg 1, #8445's own literal pin also went red, which is the literal a drifting author updates in the same commit.

One measured trap recorded in the file header: the local face arrives through the BUILT `@objectstack/driver-sql` (this package resolves the workspace dependency to its `dist`, no vitest alias to `src`), so leg 1 ran GREEN against a stale `dist` until that package was rebuilt — which reads exactly like a pin that does not work.

Both reworded faces were restored from the commit; `git diff origin/main --stat` is the one test file, 117 insertions.

## Scope and changeset

No production code — `sql-driver.ts` and `remote-transport.ts` are untouched in the diff. Nothing user-visible ships, so this takes the `skip-changeset` label rather than a changeset file; `check-empty-changeset` bans a stub, and a declaring changeset would announce a release note for a change no consumer can observe.

Related: #8445, #8413, #5240 (one condition, one wording), #6203 / #5769 (the two-answers-per-connection-string defect class).

---

### DIFF
--- a/packages/drivers/driver-turso/src/turso-local-remote-unique-parity.test.ts
+++ b/packages/drivers/driver-turso/src/turso-local-remote-unique-parity.test.ts
@@ -33,7 +33,7 @@
  * reads them. The divergence itself is the defect, so the divergence is what is
  * pinned: same declaration, same write, both faces, one assertion.
  *
- * # The three pins, and what each one alone would miss
+ * # The four pins, and what each one alone would miss
  *
  *  - **The DDL pin** asserts the UNIQUE index the remote face emits for a
  *    `unique: true` column, by name and by key. Without it a future rewrite of
@@ -49,10 +49,20 @@
  *    `toThrow()` here is blind in both directions: it stays green on the raw
  *    `SqliteError` this card exists to remove.
  *
+ *  - **The parity pin** [#8568] asserts the two faces' refusals against EACH
+ *    OTHER — message, `code` and `status` — on one condition raised twice.
+ *    Every pin above it, and #8445's local-face suite, is single-face and pins
+ *    its own wording as a LITERAL: a reword of one package with its own literal
+ *    updated alongside it drifts the pair while every suite stays green. That
+ *    is the state this file was written to make impossible, reached from the
+ *    one direction it had left open.
+ *
  * The refusal pin carries a **positive control** beside it: with the unique
  * index present, the same `conflictKeys` upsert MERGES. Without that half, a
  * transport that refused every `conflictKeys` upsert unconditionally would pass
  * the refusal pin — and would have broken the capability instead of fixing it.
+ * The parity pin carries the same control for the same reason, doubled: two
+ * faces that refused everything would agree perfectly.
  *
  * # Reverse verification, direction predicted BEFORE it was run
  *
@@ -63,6 +73,24 @@
  * The refusal pin's negative half stays GREEN across the revert: its table
  * never had a unique index to lose, which is what makes it the pin for the
  * already-created-table case rather than a second copy of the DDL pin.
+ *
+ * The parity pin [#8568] has TWO legs to predict, because two-way redness is
+ * the entire property being bought: rewording ONLY `driver-sql`'s helper must
+ * turn it red, and rewording ONLY `driver-turso`'s must turn it red as well. A
+ * pin that reddens on one leg only is the one-way pin this card replaced. Both
+ * legs were measured (`1 failed | 9 passed` each, the failure being the message
+ * comparison in both). The REMOTE leg is the one worth reading: under it every
+ * pre-existing pin in this file stayed GREEN, the refusal pin included — its
+ * assertions are token regexes (`/crm_contact_plain/`, `/email/`, `/unique/i`)
+ * that a reword preserving those words walks straight past, which is exactly
+ * how the drift could have landed unnoticed.
+ *
+ * ⚠️ One measured trap for whoever runs that verification again: the LOCAL face
+ * arrives here through the BUILT `@objectstack/driver-sql` (this package
+ * resolves the workspace dependency to its `dist`, and there is no vitest alias
+ * to `src`). A reworded `sql-driver.ts` therefore changes nothing until that
+ * package is rebuilt — leg 1 ran GREEN against a stale `dist` before the
+ * rebuild, which reads exactly like a pin that does not work.
  */
 
 import { describe, it, expect, beforeEach, afterEach } from 'vitest';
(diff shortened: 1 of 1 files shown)


os-zhuang:
**ACCEPT** — reviewed against the tree at `20115877`, not against the report.

The one thing this card could have gotten wrong is the thing it exists to prevent, so I checked that first and directly:

```ts
expect(localErr!.message).toBe(remoteErr!.message);
expect(localErr!.code).toBe(remoteErr!.code);
expect(localErr!.status).toBe(remoteErr!.status);
```

Two runtime values, compared to each other. Not each against its own literal — which is the form that stays green forever while the faces drift, i.e. the defect re-created inside its own fix. Confirmed present in the diff, not merely described.

Verified independently of the description:

- **The refusal guard is real, not decorative.** `captureError` returns `null` on success (`:129-136`), so without the `{ local: true, remote: true }` assertion a face that stopped refusing would reach the comparison holding `null` and report a `TypeError` instead of the divergence. The guard converts that into the right failure.
- **The anchor is a value assertion, not a second wording literal.** `VALIDATION_ERROR` / `400` plus the `cause` regex. This is load-bearing for a reason the card didn't have to state: agreement alone is bought by both faces regressing *together* — two raw `SqliteError`s agree on `SQLITE_ERROR`/`undefined` perfectly. Anchoring on the envelope closes that, while still letting an identical reword of **both** faces stay green, which is what #5240 actually asks for.
- **The positive control is doubled and necessary.** Two faces refusing every `conflictKeys` upsert would satisfy the parity assertion completely; both must still MERGE on a backed target.
- **The `bypassTenantAudit: true` asymmetry between the two calls is this file's existing convention**, not something introduced here — Pin 3 at `:215-218` calls the pair the same way. I checked this because an ad-hoc asymmetry between the two faces would have meant the pin was comparing two different code paths.
- **Imports and fixtures all pre-exist** (`StandardErrorCode` from `@objectstack/spec/api`, `CONTACT_NO_UNIQUE` at `:115`) — the diff adds none, so a missing one would have been a typecheck failure rather than a review finding.

**Reverse verification, both legs, is the part that makes this a pin rather than a decoration.** Leg 2 is the one that matters: rewording only the remote face turns it red while all nine pre-existing cases stay green — Pin 3 included, because its assertions are token regexes (`/crm_contact_plain/`, `/email/`, `/unique/i`) that a reword preserving those words walks straight past. That is a measurement of the blind spot, not an argument about it.

The stale-`dist` trap recorded in the file header is worth more than the pin in the long run: the local face arrives through the **built** `@objectstack/driver-sql`, so leg 1 ran green against a stale `dist` before a rebuild — reading exactly like a pin that does not work. Anyone re-running this verification would have hit it and drawn the opposite conclusion.

Scope and disposition: `git diff origin/main --stat` is the one test file, +117/−1; `sql-driver.ts` and `remote-transport.ts` are untouched, so both reworded faces were restored cleanly. Path surface is `packages/drivers/**` only — no `docs/adr/**`, no `.claude/skills/**` — so this is the seat's to queue. `skip-changeset` is the right call: nothing user-visible ships, and `check-empty-changeset` bans the stub that would otherwise announce a release note for a change no consumer can observe.

Holding ready + merge queue until all 24 check runs report green — several are still `in_progress`, and the aggregate status field is not evidence here (it covers only the Vercel commit status and reads `success` while jobs are still running).

---

PR #8581, opened 2026-08-13
test: hold the driver-double WHERE matchers to a combinator conformance battery

### LINKED ISSUE #8494: Nothing holds the driver-double WHERE matchers correct — sixteen files were fixed across three lanes and reinstating the $or early-return would fail nothing
Filed unassigned by the `domain:engine-core` seat (#6019), session `session_01RDTnVvsgA6cUZ4xFVtPZRy`, as the follow-up half of the **PM ruling on #7620** (comment `5283010316`). ⛔ Not queued, not graded by me — triage grades it.

## The fact

**#7620** corrected sixteen in-memory `WHERE` matchers across three lanes that **returned early** on `$or`, discarding every sibling equality key — so a real driver's conjunction became a different query, and the suites stayed green while testing scenarios nobody wrote. All three lanes have now landed or are in flight:

| lane | PR | files |
|---|---|---|
| `objectql` | #7846 (merged) | 6 |
| `plugin-sharing` / `plugin-security` / `runtime` | #8493 | 12 |

⚠️ **Measured across all sixteen: reinstating an early return today would fail nothing.** The doubles are now correct, and **nothing holds them correct.**

That is #7620's own defect one level up. The original was *"the suite stays green while testing a different query."* The current state is *"the doubles are right, and a future edit can make them wrong again with the suite still green."* ⇒ Fixing the instances without guarding the class means this card gets refiled in six months, after the next author inherits a double that lies.

## Why this is not a rider on #8493

⛔ Deliberately excluded. #8493 is test-only, measured, and clean; a lint rule or shared guard is a different kind of change, and bundling it would blur what that PR proved. This was recorded as the PM's sequencing decision, ⛔ not the dev's to make — the dev correctly reported the measurement and left it.

## ⚠️ A guard shaped only as "don't early-return on `$or`" is NOT sufficient

PR #8493 surfaced three matcher instances with **no combinator branch at all**:

- `packages/plugins/plugin-security/src/security-plugin.test.ts` (×2)
- `packages/plugins/plugin-security/src/explain-engine.test.ts`

Measured never invoked with a combinator today, so correctly left alone. But if one is ever added, such a matcher falls into its equality loop, treats `$or` as a **field name**, compares `row.$or` against an array, and returns no match — the suite then silently asserts on an empty set. ⭐ Same failure family, arguably worse, and **invisible to a guard that only looks for a premature `return` on `$or`.**

## What a guard has to cover

1. An early `return` on `$or` / `$and` that skips sibling keys.
2. A matcher with **no combinator handling at all**, which silently drops a combinator into equality comparison.
3. ⚠️ Whatever shape it takes, it must **fail when the defect is reintroduced** — demonstrated, not asserted. A guard nobody has watched go red is itself unguarded.

## ⛔ What this card is NOT

⛔ **Not a request to extract a shared `matchesWhere` helper.** That was ruled **NO** on #7620, on measurement: the twelve are not a single lowest common denominator (`plugin-sharing` varies between `$in`-only and `$in`+`$ne`+`$gte`/`$gt` by file), so extraction would either flatten capability or need per-call-site configuration. The repo's own recorded reason also stands — *"a gate that imports its own substrate from another gate's file couples two tripwires that must be able to fail independently."*

⚠️ That reasoning is **specific to independent test doubles** and ⛔ does not transfer to production algorithms mirrored across packages — the opposite call was made on #4953 / PR #8483, where a duplicated `materializeDeclaredFields` is getting a drift guard precisely because those two copies must not diverge.

## Prior art in the repo

- `packages/objectql/src/protocol-revert-org-scope.test.ts` (PR #7619) carries the corrected form with an in-file comment explaining why.
- `SLOT_LOOKUP_UNSWEPT` in `eslint.config.mjs` is an existing example of a ratchet-with-grandfathering over a test-shape rule, if a lint rule is the chosen shape.

Refs: #7620 · PR #7846 · PR #8493 · #7619 · #7264 (the type-channel sibling over these same doubles).

os-zhuang:
Closes #8494.

## What this is

#7620 corrected sixteen in-memory `matches(row, where)` doubles that short-circuited on `$or`, discarding every sibling equality key — so a real driver's conjunction became a different query while the suites stayed green. Three lanes landed the fixes (#7846, #8493, #7619). #8494 is the observation one level up: **the doubles were made right, and nothing held them right.**

`pnpm check:where-matcher` lifts every discovered matcher out of its test file and asks it four combinator questions. The criterion is one sentence:

> **Answer correctly, or refuse by throwing — never answer silently wrong.**

Refusal counts as conforming on purpose: the defect class is *silence*, not incompleteness. A double that throws on an operator it does not implement makes the suite red the moment a combinator arrives. That is not this gate's invention — `packages/objectql/src/engine-autonumber-*.test.ts` already does it, with its own recorded reason, *"silently ignoring an unknown operator would let a bad query pass as a good one."* So the cheap correct answer for a double that only ever sees scalar equality is one `throw`, not a combinator implementation.

## Why behavioural and not a lint rule

#8494 binds the guard to cover **both** failure shapes, and they share no source shape:

- **(a) early return** — a combinator branch that `return`s, dropping the sibling keys the loop had not reached.
- **(b) no combinator branch at all** — an `Object.entries(where).every(…)` matcher that reads `$or` as an ordinary field name, compares `row.$or` (`undefined`) against the array, and silently excludes the row.

Shape (b) is an **absence**, which no pattern-match over source can see, and the *correct* forms in this repo already vary too much to pin (a prelude of `if (…) return false` before the entries loop, and a combinator arm inside it, are both correct and share no shape).

Discovery is structural, then **behaviourally admitted**: every candidate must pass a control probe (matches on equality, and really filters) or it is dropped as out of scope. So membership is decided by behaviour, never by parameter names — the ambiguity `check-engine-double-contract` documents. It also stops a matcher passing vacuously by answering `true` to everything.

The four probes are load-bearing as **pairs**. `orConjoined` alone proves nothing — a shape-(b) matcher also answers `false` there, for the opposite reason, and would ride through as "conjoins correctly". Only `orRecognised === true` makes that `false` mean *conjoined* rather than *excluded by accident*.

⛔ No shared `matchesWhere` — ruled NO on #7620, not re-litigated. This gate never *gives* a matcher an implementation; it only *asks* each independent double a question.

## Reverse verification — watched, not asserted

Reinstating the early return in `plugin-sharing/src/authored-row-write-deferral.test.ts` (one of the sixteen, corrected by #8493):

| | before | with the defect reinstated |
|---|---|---|
| `pnpm --filter @objectstack/plugin-sharing test` | 21 files / 569 tests pass | **21 files / 569 tests pass** — still green |
| `pnpm check:where-matcher` | green | **RED, exit 1** |

```
• packages/plugins/plugin-sharing/src/authored-row-write-deferral.test.ts: NEW silently-wrong WHERE matcher (1 silent, 0 unjudged).
      line 182 `matches`: early-return (sibling keys discarded)
```

That table is the whole card: the suite cannot see it, and now something can. Shape (b) was demonstrated the same way — deleting the combinator branch outright reddens the gate with a *distinct* attribution, `no combinator branch (combinator read as a field name)`, failing **recognition** rather than conjunction.

Restored both times by checking the file back out of `HEAD` (index and tree together); `git hash-object` returned `26d2c2fb34e97066ad0885f5dee03282dab3647a` before and after each round, `git status --porcelain` came back clean, and the gate returned to green.

## Five live defects this found

Building the guard surfaced **five instances of shape (a) still live on `main`** that all three correction lanes missed — the lanes grepped a named file list, which is exactly the limitation this card exists to remove:

- `packages/objectql/src/engine-author-state-query.test.ts`
- `packages/objectql/src/engine-findone-contract.test.ts`
- `packages/objectql/src/engine-unknown-option.test.ts`
- `packages/objectql/src/search-companion-read-projection-conformance.test.ts`
- `packages/plugins/plugin-security/src/store-fault-fail-closed.test.ts` (both shapes at once)

Fixed here rather than baselined. A ledger that grandfathered the exact defect its gate exists to stop would be worth nothing — so **shape (a) is enforced with an empty ledger**.

## The ledger

168 matchers discovered: 89 conforming (4 of them by refusing), 78 silently wrong, 1 unjudged, across 75 grandfathered files. Every grandfathered entry is shape (b). Shrink-only, hand-edited, reconciled in **both** directions, with the key set checked against the merge base so a newly-added file cannot ride in matching its own count (the `SLOT_LOOKUP_UNSWEPT` precedent). Deliberately no `--update` flag — a generator would admit a new silently-wrong double by "just run the update command".

Sweeping those 78 is filed separately as #8582, not smuggled in here.

## Tests and gates

- `pnpm --filter @objectstack/objectql test` — 200 files / 3550 tests pass
- `pnpm --filter @objectstack/plugin-security test` — 55 files / 1079 tests pass
- `pnpm --filter @objectstack/plugin-sharing test` — 21 files / 569 tests pass
- `typecheck` on objectql + plugin-security — clean; `eslint --no-inline-config` on all changed files — clean

Green: `check:where-matcher` (incl. `--self-test`), `check:nul-bytes`, `check:engine-double-contract`, `check:durability-log-level`, `check:query-options-erasure`, `check:i18n`, `check:type-check-coverage`, `check:type-check-debt`, `check:test-source-alias`, `check:type-source-resolution`, `check:cross-package-test-inputs`, `check:node-version`, `check:required-contexts`, `check:shard-attestation`, `check:workflow-status-functions`.

Two of those first reported a **prerequisite** failure rather than a verdict, and are green once the prerequisite is met — worth knowing because neither message is about the diff: `check:i18n` needs the workspace CLI built ("Nothing was checked"), and `check:type-check-debt` refuses to re-measure without the full built closure.

**No changeset**, and `skip-changeset` applied: the diff is test doubles, two `scripts/` gates, the workflow and the root `package.json` — nothing published changes, so an empty-frontmatter changeset would be rejected by `check:empty-changeset` and a package-bumping one would ship a release note about test internals. This also keeps the PR clear of `check:objectui-pin-fresh`, which fires on any `.changeset/*` touch and is currently red as pre-existing repo state (`.objectui-sha` behind objectui main, #3340) — unrelated to this diff either way.

---

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -738,6 +738,25 @@ jobs:
       - name: Engine test-double contract gate
         run: pnpm check:engine-double-contract
 
+      # WHERE-matcher conformance gate (#8494, from #7620). The read-side
+      # sibling of the gate above, and the half its header lists as not
+      # covered. #7620 corrected sixteen in-memory `matches(row, where)`
+      # doubles that short-circuited on `$or` and discarded every sibling
+      # equality key — turning a conjunction into a different query while the
+      # suites stayed green — and #8494 is the observation one level up:
+      # nothing held them corrected, so reinstating the early return failed
+      # nothing. This lifts each discovered matcher out of its file and asks
+      # it four combinator questions; it must answer correctly or throw, never
+      # answer silently wrong. Behavioural rather than syntactic on purpose —
+      # the second failure shape is an ABSENCE (no combinator branch at all),
+      # which no pattern-match over source can see. Pre-existing combinator-
+      # blind doubles are in a shrink-only measured baseline. Runs its own
+      # --self-test first: the detector can be broken while every double is
+      # fine, and a scan that quietly stops matching would report OK while
+      # reading nothing (#4868's family).
+      - name: WHERE-matcher conformance gate
+        run: pnpm check:where-matcher
+
       # Paired kernel-hook pin gate (#5282, from #5170 / #5257 / #5274). The two
       # kernels — ObjectKernel (production) and LiteKernel (vitest / serverless /
       # edge) — run the same plugin code and the same hook vocabulary, but do NOT
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
     "check:driver-conformance": "node scripts/check-driver-conformance.mjs --self-test && node scripts/check-driver-conformance.mjs",
     "check:driver-memory-census": "node scripts/check-driver-memory-census.mjs --self-test && node scripts/check-driver-memory-census.mjs",
     "check:engine-double-contract": "node scripts/check-engine-double-contract.mjs --self-test && node scripts/check-engine-double-contract.mjs",
+    "check:where-matcher": "node scripts/check-where-matcher-conformance.mjs --self-test && node scripts/check-where-matcher-conformance.mjs",
     "check:resume-authority-declared": "node scripts/check-resume-authority-declared.mjs --self-test && node scripts/check-resume-authority-declared.mjs",
     "check:spec-parsed-alias": "node scripts/check-spec-parsed-alias.mjs --self-test && node scripts/check-spec-parsed-alias.mjs",
     "check:tenant-chokepoint": "node scripts/check-tenant-chokepoint.mjs --self-test && node scripts/check-tenant-chokepoint.mjs",
--- a/packages/objectql/src/engine-author-state-query.test.ts
+++ b/packages/objectql/src/engine-author-state-query.test.ts
@@ -64,8 +64,17 @@ function makeRecordingDriver() {
     const matches = (row: any, where: any): boolean => {
         if (!where || typeof where !== 'object') return true;
         for (const [k, v] of Object.entries(where)) {
-            if (k === '$and') return (v as any[]).every((w) => matches(row, w));
-            if (k === '$or') return (v as any[]).some((w) => matches(row, w));
+            // `$and` / `$or` are CONJOINED with their sibling keys, the way a
+            // real driver reads them — never `return`ed, which would discard
+            // every sibling the loop has not reached yet (#7620 / #8494).
+            if (k === '$and') {
+                if (!(v as any[]).every((w) => matches(row, w))) return false;
+                continue;
+            }
+            if (k === '$or') {
+                if (!(v as any[]).some((w) => matches(row, w))) return false;
+                continue;
+            }
             if (k.startsWith('$')) continue;
             if (v && typeof v === 'object' && '$in' in (v as any)) {
                 if (!(v as any).$in.map(String).includes(String(row[k]))) return false;
--- a/packages/objectql/src/engine-findone-contract.test.ts
+++ b/packages/objectql/src/engine-findone-contract.test.ts
@@ -65,8 +65,17 @@ function makeRecordingDriver() {
     const matches = (row: any, where: any): boolean => {
         if (!where || typeof where !== 'object') return true;
         for (const [k, v] of Object.entries(where)) {
-            if (k === '$and') return (v as any[]).every((w) => matches(row, w));
-            if (k === '$or') return (v as any[]).some((w) => matches(row, w));
+            // `$and` / `$or` are CONJOINED with their sibling keys, the way a
+            // real driver reads them — never `return`ed, which would discard
+            // every sibling the loop has not reached yet (#7620 / #8494).
+            if (k === '$and') {
+                if (!(v as any[]).every((w) => matches(row, w))) return false;
+                continue;
+            }
+            if (k === '$or') {
+                if (!(v as any[]).some((w) => matches(row, w))) return false;
+                continue;
+            }
             if (k.startsWith('$')) continue;
             // [#7641] `$icontains` — the operator `$search` compiles to. This
             // arm folded BOTH sides while it was still keyed on `$contains`,
--- a/packages/objectql/src/engine-unknown-option.test.ts
+++ b/packages/objectql/src/engine-unknown-option.test.ts
@@ -51,7 +51,17 @@ function makeDriver() {
     const matches = (row: any, where: any): boolean => {
         if (!where || typeof where !== 'object') return true;
         for (const [k, v] of Object.entries(where)) {
-            if (k === '$and') return (v as any[]).every((w) => matches(row, w));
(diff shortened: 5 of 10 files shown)


os-zhuang:
**PM review (`domain:engine-core` seat, #6019): ACCEPT.** Ready + auto-merge (squash) enabled.

⚠️ **The three red `Check Changeset` runs are stale — read them before re-running anything.** They ran at 23:41:17 against a head that had no `skip-changeset` label: the first label write was overwritten wholesale by the labeler bot, dropping it, and it was re-applied as the union straight after. The later runs of the same check (`31754711230`, `31754834441`, `31754834500`) all report `skipped`, which is the gate reacting to the label. Nothing to fix, and ⛔ nothing to re-queue.

The no-changeset route is the one this gate's own log recommends — *"If it releases nothing … apply the `skip-changeset` label"*, and *"a wrong `skip-changeset` label is caught by review; a wrong empty changeset is caught by nobody."* Nothing published changes here (test doubles, two `scripts/` gates, `lint.yml`, root `package.json`), so it is correct.

What I checked against the tree rather than the description:

- **`check-engine-double-contract.mjs` (+9/−4) is comment-only** — it moves the read side out of that gate's "not covered" list and records why this one could not reuse its pattern. ⛔ No behavioural change to the existing gate.
- **The ratchet cannot be widened quietly.** `reconcile` reconciles in *both* directions (count grows → error; count shrinks or the file goes clean → "ratchet DOWN, delete its entry"), an unbaselined file is a hard `NEW silently-wrong` error, `monotonicity()` diffs the baseline key set against the merge base so a new file cannot ride in, and there is deliberately no `--update` flag.
- ⭐ **A check that could not run does not read as a check that passed** — when the merge-base read fails, the *success* path prints `NOT verified: … "no files added" is unchecked this run`. Likewise `discovered === 0` is a hard error ("a broken scan, not a clean repo"), not a quiet pass.
- **Reverse verification is watched, not asserted**, and the two shapes fail with *distinct* attributions — early-return vs. `no combinator branch (combinator read as a field name)` — so the battery's probes are load-bearing as pairs rather than one probe answering for both.

⭐ The result worth recording beyond this card: building the guard turned up **five live shape-(a) defects still on `main`** that all three #7620 correction lanes missed, because those lanes grepped a named file list — the exact limitation this card existed to remove. Fixed here rather than baselined, so shape (a) ships with an empty ledger; the 78 combinator-blind doubles are grandfathered and filed as #8582.

---

PR #8587, opened 2026-08-14
fix(security): guard the write-response internal-field property, not one class (#8497)

### LINKED ISSUE #8497: The write-response internal-field tripwire walks the protocol class only — a direct engine write mouth outside it (rest-server's batch ql.update) is covered by the fix but not by the guard
Filed unassigned by the `domain:engine-core` seat (#6019), session `session_01RDTnVvsgA6cUZ4xFVtPZRy`, from reading PR **#7996** (#7823) at landing. ⛔ Observation-class, ⛔ not queued, ⛔ not a defect in that PR — its own fix covers this mouth. Triage grades it.

## The fact

#7823's **A-prime** ruling relocated the internal-field write-response strip to the generic-data-path ingress, and gated the relocation on a tripwire — because *"a future generic ingress that forgets the shared helper leaks silently."*

⭐ The tripwire that shipped is genuinely strong: `protocol.write-response-internal-fields.tripwire.test.ts` enumerates every `*Data` method on the protocol class **at runtime** (a prototype walk by naming convention, ⛔ not a hand-written list), requires each to have a registered recipe or fails with instructions, and carries a **negative control** (`leakyData`) proving both the walk and the sentinel scan go red on a real leak. It was reverse-verified: helper removed from `createData` ⇒ RED; restored ⇒ green, byte-identical by `git hash-object`.

⚠️ **But the ingress surface turned out to be wider than one class.** The PR's own "Packages touched" section records a write mouth that is **not** on the protocol class:

> `@objectstack/rest` (`rest-server.ts` — batch update arm applies the shared strip)

That is the REST cross-object batch's **direct `ql.update`** call. The shared helper is correctly applied there, so ⛔ **nothing leaks today.** But a prototype walk over the protocol class cannot see it — so if someone later adds a second direct engine write mouth in `rest-server.ts` (or in any other package that calls the engine directly rather than through `metadata-protocol`), **the tripwire stays green while the new mouth leaks.**

⇒ The guard's coverage is *"every `*Data` face on the protocol class"*. The property that actually needs guarding is *"every response body an external caller can receive from a write."* Those two were the same set on the day the tripwire was written, and the `rest-server` mouth is the standing proof they are not the same set by construction.

## Why this is worth recording rather than shrugging

⚠️ It is the same shape as the defect #7823 existed to fix, one level up. `internal: true` was honoured at three places and the author of a new surface had no way to know which; the fix made the *rule* structural at one boundary, and the *guard* structural at one class. The gap between "the boundary" and "the class" is exactly where the next silent leak fits.

⭐ Note also what this card is **not** claiming: the tripwire is not weak. A runtime prototype walk with a live negative control is materially better than the hand-kept list the ruling feared, and it does catch a genuinely new `*Data` face. ⛔ Do not read this as a reason to redesign it — the question is whether its **scope** should widen, not whether its **mechanism** is right.

## Possible directions (⛔ not prescriptive, no ruling implied)

- Extend the tripwire's enumeration beyond one class — e.g. a source-level check that any call site writing through the engine and returning a body to an external caller passes through `omitInternalFieldsFromWriteResponse`.
- Or invert it: assert at the **response envelope** layer rather than per-ingress, so the property is checked where it is actually true (nothing an external caller receives carries an `internal: true` value), independent of which code path produced it.
- Or accept the current scope deliberately and **write down** that direct-engine mouths outside `metadata-protocol` are the author's responsibility — ⚠️ the option this repo usually rejects, since it is a convention rather than a mechanism, but it is honest and cheap and belongs on the list.

## Refs

#7823 · PR #7996 · #7728 (the `internal: true` flag) · #8118 (`Engine.resolveInternalField`) · #8494 (a structurally identical "the instances are fixed, nothing holds them fixed" card from #7620).

os-zhuang:
Closes #8497

Widens the #7823 write-response `internal: true` guard from *"every `*Data` face on the protocol class"* to the property it exists to hold:

> no response body an external caller receives from a write carries an `internal: true` value

⛔ **Not a redesign.** The shipped `protocol.write-response-internal-fields.tripwire.test.ts` is untouched — its runtime prototype walk and its `leakyData` negative control stay exactly as they are, and still pass (1245 metadata-protocol tests green). This adds scope, it does not replace mechanism.

---

## ⚠️ Widening the guard found a live leak

The card said the `rest-server` batch mouth is covered by the fix and merely invisible to the guard — correct, and it still is. But the card's own sentence was broader:

> …or **in any other package that calls the engine directly rather than through `metadata-protocol`**, the tripwire stays green while the new mouth leaks.

That sentence was already true of a second transport. `@objectstack/mcp`'s stdio bridge is engine-only by construction (the long-lived stdio host cannot reuse the runtime's request-shaped `callData` builder), and its `create` arm handed `engine.insert`'s result straight back to the MCP caller. Since #7823 the engine deliberately keeps write results whole, so the flagged column rode the tool response verbatim. Measured before the fix, on the bridge the stdio transport actually serves:

```
{"object":"vault","id":"r1","record":{"name":"row","id":"r1","vault_secret":"INTERNAL-SENTINEL-8497"}}
```

The file's own header listed its protocol-layer divergences as *"deliberate, filed, not security"*. One limb of that list **was** security; the header now says so.

Fixing it is inside this card's scope rather than a follow-up: a guard shaped to exclude the one place the property is false would repeat exactly the mistake the card is about.

## The three directions, evaluated by measurement

The body offered three. All three were measured before choosing.

**1. Source-level check on call sites that write through the engine and return a body.** ⛔ Rejected as the primary shape. Measured: **176 direct engine write call sites across 25 packages** outside `metadata-protocol`. Almost all are internal side-effect writes (plugins, services) that answer no external caller — and "returns a body to an external caller" is a semantic property no grep can decide. So this becomes a ~176-entry ledger in which ~175 entries say "not a response", each one a human judgement the check cannot verify, churning on every new plugin write. It guards a *proxy* (call-site shape), not the property.

**2. Assert at the response envelope, independent of which path produced it.** ⭐ **Picked** — the triage default, and the measurement supports it. External write bodies are produced by transports, and both transports enumerate their own surface at runtime, so the property can be asserted where it is actually true using the *same mechanism the shipped tripwire already uses*: `RestServer.getRoutes()` (**88 routes, 45 of them write methods**) and the `McpDataBridge` object the stdio factory returns. A new route or a new bridge verb with no disposition fails the suite with instructions — the enumeration grows by itself, exactly as the prototype walk does.

**3. Write it down: direct-engine mouths outside `metadata-protocol` are the author's responsibility.** Adopted only as a *complement*, never as the mechanism — it is the convention-not-mechanism option this repo rejects. It survives as the module header on the relocated helper, which now names all three tripwires and what to do when adding a mouth.

## What changed

- **`@objectstack/mcp`** — the stdio bridge's `create` strips (the leak). `update` strips too: that arm discards the engine's write result and echoes the read-path row plus the caller's own patch, so no *stored* value could reach it — but a caller who puts an `internal: true` key in `data` would otherwise have it echoed back, using their own bytes as an oracle on a column the flag says is never returned. Read verbs untouched.
- **`@objectstack/core`** — the strip helper moved here from `@objectstack/metadata-protocol`. It shipped beside the protocol class when that class was its only caller, but `rest` and `mcp` both reach the engine directly and **neither depends on `metadata-protocol`**, so the old home forced every new mouth to choose between a duck-typed reach through a protocol instance (what `rest` does, via an optional `?.` call that silently no-ops if the method ever disappears) and a private restatement of a security-relevant rule. `core` is the floor all three already depend on and already hosts this exact class of shared write-path helper (`bulk-write.ts`, shared by `metadata-protocol` and `rest`). **No API change**: `metadata-protocol` re-exports both names, verified against the rebuilt `dist`, not the source.
- **Two new tripwires** (`rest`, `mcp`), same mechanism as the shipped one: runtime enumeration, total disposition map, negative control.

`rest-server.ts` itself is **unchanged** — its mouth already applies the strip correctly. The latent `?.` silent-noop seam is now covered by measurement rather than by reading.

## Reverse verification

Three experiments, each with the expected direction predicted **before** running (all three: RED on the driven route — not more-diagnostics, not inverted). The fix was committed first, so every restore came out of a real commit.

| Experiment | Result |
|---|---|
| Delete the strip from the REST `POST /batch` update arm | 🔴 RED on `POST /api/v1/batch` only |
| Add a **second** unstripped direct engine mouth (batch create arm to a direct `ql.insert`) | 🔴 RED again — the leak moved to the new mouth |
| Delete the new MCP `create` strip | 🔴 RED on `create` |

The acceptance criterion inherited from the family is the middle row, and it discriminates precisely. With the strip deleted from the update arm, the create arm (protocol ingress) stayed clean and only the direct arm leaked:

```
{"results":[{"name":"CONTROL...","id":"r_2"},
            {"id":"row-1","name":"CONTROL...","vault_secret":"INTERNAL-SENTINEL-8497-NEVER-SERIALIZED"}]}
```

With the *second* mouth added instead, the leak moved to the first element while the still-stripped update arm stayed clean — the guard names which mouth leaked, not merely that something did.

All three restores proven byte-identical with `git hash-object`:

- `packages/rest/src/rest-server.ts` → `6a58de8f15a9565f5628553666a4cff8796db305` (baseline, twice)
- `packages/mcp/src/stdio-data-bridge.ts` → `53e08e6aa6136169723b6b7ba7317b77a278a52a` (baseline)

Two anti-blindness assertions guard the guard: every driven case demands a control value in the body (so a refusal or a 501 cannot satisfy "no sentinel" by returning nothing), and one test asserts the fixture's stored row really does carry the flagged value — without it, every green could mean the secret was never stored.

## Tests and gates

**Tests** — 4108 green: `core` 786, `mcp` 182, `metadata-protocol` 1245 (incl. the untouched #7823 tripwire), `rest` 1895.

**Gates run locally** — derived with `scripts/pm/dispatch-gates.mjs` against the actual changed paths, plus those judged implicated:

`check:nul-bytes` · `check:changeset-gate-self-tests` · `check:cross-package-test-inputs` · `check:durability-log-level` · `check:kernel-hook-pairs` · `check:test-source-alias` · `check:engine-double-contract` · `check:query-options-erasure` · `check:type-check-coverage` · `check:type-check-debt` · `check:objectui-changeset` · `eslint` · `typecheck` (mcp, rest) — **all PASS**.

⚠️ **`check:type-check-debt` caught a real regression and is worth naming.** `packages/rest` hides its test files from its own `typecheck` script, so the package typecheck passed green while the ratchet measured **+2** raw tsc errors against the frozen TEST_DEBT entry — both in the new tripwire (a relative import missing its `.js` extension under NodeNext, and `registry.registerObject` called without its required `packageId`). Fixed rather than ledgered: TEST_DEBT is shrink-only and raising it is maintainer-only. Re-measured at **155**, exactly the recorded value.

⚠️ `check:objectui-pin-fresh` is **pre-existing RED** on this repo (`.objectui-sha` behind `objectui` main, #3340) and fires on any `.changeset/*` touch. Not caused by this branch and deliberately not "fixed" here.

## Scope

- #8494 is not addressed here — same "instances fixed, nothing holds them fixed" class, different substrate, dispatched separately. No files shared.
- `content/docs/releases/` untouched; the changeset is this PR's input to the release notes.
- The 33 non-data-plane write routes carry declared dispositions (`no-record-echo` / `protocol-ingress`) rather than drives. That boundary is stated plainly in the file: only `driven` is a measurement, the rest are reviewed claims, and the enumeration's job is to force an explicit decision when a route is added. Riskier claims were checked rather than asserted — the import runner's per-row results spread only `warnings`, never the written row, and the sharing routes echo a grant row from the sharing service, not a row of the `:object` in the path.

---

### DIFF
--- a/.changeset/write-response-internal-fields-property-guard.md
+++ b/.changeset/write-response-internal-fields-property-guard.md
@@ -1,0 +1,74 @@
+---
+"@objectstack/core": patch
+"@objectstack/metadata-protocol": patch
+"@objectstack/mcp": patch
+---
+
+fix(security): the MCP stdio bridge stops echoing `internal: true` columns from a write, and the write-response guarantee is guarded as a PROPERTY rather than per-class (#8497)
+
+**A live leak, found by widening a guard.** #7823 relocated the `internal: true`
+write-response strip to the generic-data-path ingress and gated the relocation on
+a tripwire that enumerates every `*Data` face on the protocol class. The card that
+produced this change observed that the guard's coverage — *"every `*Data` face on
+one class"* — is narrower than the property that needs holding — *"no response body
+an external caller receives from a write carries an `internal: true` value"* — and
+that `@objectstack/rest`'s cross-object batch (a direct `ql.update`) was the
+standing proof the two are not the same set.
+
+Widening the guard to the property immediately found a second direct mouth that
+was **not** covered, and it was leaking. `@objectstack/mcp`'s stdio bridge
+(`stdio-data-bridge.ts`) is engine-only by construction — the long-lived stdio
+host cannot reuse the runtime's request-shaped `callData` builder — and its
+`create` arm handed `engine.insert`'s result straight back to the MCP caller.
+Since #7823 the engine deliberately keeps its write results whole, so the flagged
+column rode the tool response verbatim. Measured before the fix:
+
+```
+{"object":"vault","id":"r1","record":{"name":"row","id":"r1","vault_secret":"<the stored secret>"}}
+```
+
+The file's own header had listed its protocol-layer divergences as *"deliberate,
+filed, not security"*. One limb of that list **was** security, and the header now
+says so.
+
+**What changed**
+
+- `@objectstack/mcp` — the stdio bridge's `create` runs its response record
+  through the shared strip. `update` does too: that arm discards the engine's
+  write result and echoes the read-path row plus the caller's own patch, so no
+  *stored* value could reach it, but a caller who puts an `internal: true` key in
+  `data` would otherwise get it echoed back — their own bytes used as an oracle
+  for a column the flag says is never returned. Read verbs are untouched (the
+  engine's read-path strip is unchanged).
+- `@objectstack/core` — the strip helper
+  (`omitInternalFieldsFromWriteResponse` / `collectInternalWriteResponseFields`)
+  moved here from `@objectstack/metadata-protocol`. It shipped beside the protocol
+  class when that class was its only caller, but the generic write mouths are not
+  all on it: `rest` and `mcp` both reach the engine directly and **neither depends
+  on `@objectstack/metadata-protocol`**, so the old home forced each new mouth to
+  choose between a duck-typed reach through a protocol instance and a private
+  restatement of a security-relevant rule. `core` is the floor all three already
+  depend on, and already hosts this class of shared write-path helper
+  (`bulk-write.ts`). No behaviour change and no API change:
+  `@objectstack/metadata-protocol` re-exports both names unchanged.
+
+**What guards it now.** Two new tripwires join the shipped one — which is **not**
+replaced: its runtime prototype walk and its `leakyData` negative control are
+untouched. Each is a runtime enumeration no author can dodge by adding code
+without touching it, and each fails on a surface it has no disposition for:
+
+- `metadata-protocol` — walks the protocol class for `*Data` faces (unchanged);
+- `rest` — walks `RestServer.getRoutes()` for HTTP write routes, drives the ten
+  data-plane ones (including `POST /batch`, the direct-`ql.update` mouth) against
+  a fixture whose stored rows carry a flagged sentinel, and deep-scans each
+  response body;
+- `mcp` — walks the `McpDataBridge` faces the factory actually returns.
+
+Every driven case also asserts a control value is present, so a refusal or an
+empty body cannot satisfy "no sentinel" by returning nothing.
+
+Reverse-verified in both directions, the discipline #7823's own fix used: deleting
+the strip from the REST batch arm turned the REST tripwire red on exactly that
+route; adding a *second* unstripped direct engine mouth turned it red again;
+removing the new MCP strip turned the MCP tripwire red; every restore was proven
+byte-identical with `git hash-object`.
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -36,6 +36,14 @@ export * from './utils/datetime.js';
 // Export the shared batched-write helper (framework#2678)
 export * from './utils/bulk-write.js';
 
+// Export the shared write-response `internal: true` strip (#7823, #8497) — the
+// ONE helper every write mouth that answers an external caller runs its
+// records through. It lives here, on the floor all three transports share,
+// because the mouths are not all on the protocol class: `rest` and `mcp` both
+// reach the engine directly and neither depends on `@objectstack/metadata-
+// protocol` (which re-exports these two names unchanged).
+export * from './utils/internal-write-response.js';
+
 // Export the migration-journal runner (ADR-0119 D2, #4617) — chunk-atomic
 // migrations with durable recovery, plus the shared `engineCanRollBack` gate
 // that `@objectstack/metadata-protocol`'s atomic `batchData` also uses.
--- a/packages/core/src/utils/internal-write-response.ts
+++ b/packages/core/src/utils/internal-write-response.ts
@@ -1,0 +1,143 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#7823 / #8497] The write-response half of the `internal: true` guarantee —
+ * THE single helper every write mouth that returns a body to an external
+ * caller passes its record(s) through.
+ *
+ * ## The contract
+ *
+ * A field declared `internal: true` is *never returned on the generic data
+ * path* (#7728). The READ half lives in the engine (`omitInternalFields` runs
+ * on every find/findOne result). The WRITE-RESPONSE half lives HERE.
+ *
+ * ## Why the ingress and not the engine (the measured history, #7823)
+ *
+ * The first shape stripped `internal` fields inside the engine's insert and
+ * by-id-update paths. That conflated two different guarantees:
+ *
+ *  - "never returned on the generic data path" — the flag's sentence, about
+ *    what an EXTERNAL caller receives; and
+ *  - "never returned to the engine-level caller that performed the write" —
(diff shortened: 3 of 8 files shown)


os-zhuang:
**PM review (`domain:engine-core` seat, #6019): ACCEPT.** Ready + auto-merge (squash). All 26 checks green or skipped, `mergeable_state: clean`.

⚠️ **Maintainer visibility — this card changed class.** #8497 was filed and graded as a guard-**scope** widening whose body said, twice and in bold, that ⛔ *nothing leaks today*. That premise was **false for a second transport**: `@objectstack/mcp`'s stdio bridge `create` arm echoed `engine.insert`'s result, and since #7823 the engine deliberately keeps write results whole — so an `internal: true` column rode the MCP tool response verbatim, measured with a sentinel before the fix. A shipped transport was leaking. The fix is in this PR rather than deferred, and the reasoning is the card's own: a guard shaped to exclude the one place the property is false would reproduce exactly the mistake the card exists to correct.

⭐ Worth stating plainly, because it is the argument for this whole family of cards: **the leak was found only by widening the guard.** The card predicted an invisible-mouth *risk* and the enumeration turned up an actual one, in a file whose header asserted its divergences were "deliberate, filed, not security."

Checked against the tree, not the description:

- **The relocation is invisible to consumers.** `metadata-protocol/src/write-response-internal-fields.ts` is now a pure re-export of both names from `@objectstack/core`; `index.ts` still re-exports them and `protocol.ts` still imports the strip from there. No `package.json` changed, and Build Core + TypeScript Type Check are green, so the `mcp → core` edge already existed rather than being added.
- **Both MCP arms go through the shared helper**, not a restatement — which is the point of moving it to the floor all three transports share. Read verbs untouched.
- **The `update` arm's strip is justified rather than reflexive.** That arm discards the engine's write result, so no *stored* flagged value can reach it; the strip covers the remaining path — a caller who puts an `internal: true` key in `data` would otherwise get it echoed, using their own bytes as an oracle against a column the flag says is never returned.
- ⛔ **Triage's binding constraint honoured**: the shipped #7823 tripwire is untouched — prototype walk and `leakyData` negative control intact, 1245 `metadata-protocol` tests still green. This adds scope; it does not replace mechanism.
- **Direction predicted before each of the three reverse-verification runs**, and the middle one is the family's inherited acceptance criterion: adding a *second* unstripped mouth moves the RED to that mouth, so the guard names which one leaked rather than merely that something did. Two anti-blindness assertions stop a refusal or an unstored fixture from satisfying "no sentinel" vacuously.

One boundary the PR states rather than hides, and I agree with recording it: of the 45 write routes, only the `driven` ones are measurements — the other 33 carry reviewed dispositions (`no-record-echo` / `protocol-ingress`). The enumeration's job there is to force an explicit decision when a route is added, not to prove those 33 by execution.

---

Issue #8590, opened 2026-08-14
types: `isUniqueViolationError` claims SQLite's UNBACKED-conflict-target error — a superstring collision on "UNIQUE constraint", the exact inversion the predicate's own warning names


os-zhuang:
Found while implementing #8567 (measuring the unbacked-conflict-target condition across dialects). Filed unassigned; not fixed there — the fix moves a shared predicate's verdicts, which is its own blast radius.

## The defect

`isUniqueViolationError` (`packages/types/src/unique-violation.ts`) answers **true** for the error SQLite raises when an `ON CONFLICT` target has **no** backing unique index — i.e. for the exact condition its own docstring warns it must never claim:

> ⚠️ It must stay a **separate** predicate: `isUniqueViolationError` answers the OPPOSITE condition (a unique index exists and the row violated it), and merging them would report a working constraint as a missing one.

Measured on real driver errors (not fixtures), knex 3.3.0, `@objectstack/types` built from `a3b1d264f`:

```
===== sqlite (better-sqlite3) =====
UNBACKED-target error:
   msg  = "insert into `xtalk_plain` (`email`, `id`, `title`) values ('a@b.com', '1', 'x')
           on conflict (`email`) do update set `title` = excluded.`title`
           - ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint"
   isUnbackedConflictTargetError = true
   isUniqueViolationError        = true      <== WRONG, should be false

===== pg (PostgreSQL 16.13) =====
UNBACKED-target error:
   msg  = "insert into "xtalk_plain" (...) values ($1, $2, $3)
           on conflict ("email") do update set "title" = excluded."title"
           - there is no unique or exclusion constraint matching the ON CONFLICT specification"
   isUnbackedConflictTargetError = true
   isUniqueViolationError        = false     <== correct
```

## Cause

The vocabulary's message limb is

```ts
message: /unique constraint|unique violation|duplicate key|duplicate entry/i
```

and SQLite's sentence for the *missing*-index condition ends `...any PRIMARY KEY or UNIQUE constraint`. The words `UNIQUE constraint` sit adjacent inside it, so the `unique constraint` limb matches a sentence that says the constraint is **absent**.

Postgres escapes only by luck of word order: its `unique or exclusion constraint` is not adjacent, so the same limb misses it. That asymmetry is the tell — the limb is matching a word pair, not a condition.

[...]

PR #8591, opened 2026-08-14
fix(drivers): recognise the unbacked conflict target on Postgres too, from a measured PG 16.13 (#8567)

### LINKED ISSUE #8567: drivers(sql): the unbacked-conflict-target refusal is SQLite-only — Postgres and MySQL still answer the raw driver error
Filed out of #8445, which fixed the same condition on the SQLite face only. Unassigned; recording what was and was not measured.

## What #8445 landed

`SqlDriver.upsert` now recognises "the `ON CONFLICT` target is not backed by a PRIMARY KEY or UNIQUE index" and re-raises it as an ADR-0112 envelope (`VALIDATION_ERROR` / 400), first sentence for first sentence with #8413's remote refusal (#5240).

Recognition is a narrow match on **SQLite's own sentence**, which is the only channel SQLite fills — it raises a plain `SQLITE_ERROR`, the same generic code a syntax error carries, so `code` cannot discriminate. Measured through knex + better-sqlite3 (the message knex builds is the statement, then ` - `, then SQLite's text):

```
upsert('plain', { email: 'a@b.com', title: 'x' }, ['email'])
  -> name=SqliteError code=SQLITE_ERROR status=undefined
     msg=insert into `plain` (...) values (...) on conflict (`email`) do update set ...
         - ON CONFLICT clause does not match any PRIMARY KEY or UNIQUE constraint
```

## What is left

`driver-sql` serves three dialects. Postgres and MySQL wording for this condition was **not measured** — the dev container that implemented #8445 has no Postgres or MySQL server to raise it, and #8445's dispatch was explicit that transcribing the other two dialects' text from memory is not acceptable evidence. So on those two dialects the raw driver error still escapes, `mapDataError` still falls through to its default branch, and the caller still gets the statement text with no `code` to branch on — #8445's payload argument, unchanged, on two of three dialects.

Two questions, in order:

1. **Measure.** What do Postgres and MySQL actually raise here? Postgres is expected to have both a SQLSTATE channel and prose for it; MySQL has no `ON CONFLICT` syntax at all — knex compiles `onConflict().merge()` to `ON DUPLICATE KEY UPDATE`, which takes no conflict target, so the condition may not arise on that dialect in the same form (and if it does not, what it does instead is worth knowing, because silently merging on a different unique key is its own defect).
2. **Then decide the home.** If the recognition turns out to span dialects, the natural place is a named predicate in `@objectstack/types` beside `isUniqueViolationError` — its `UniqueViolationSignature` table (codes / errnos / message substrings) is already the shape this question needs, and every consumer already depends on that package. ⚠️ It must stay a **separate** predicate: `isUniqueViolationError` answers the OPPOSITE condition (a unique index exists and the row violated it), and merging them would report a working constraint as a missing one.

If (1) shows the dialects have nothing in common, the per-face recognition #8445 landed is the right end state and this card closes as measured-and-declined.

Related: #8445 (the SQLite half, and where the current recognition lives), #8413 (the remote-face twin), #6543 / #6250 (the shared unique-violation predicate this would sit beside), #1116 / #1077 (the envelope pass this continues).

---

os-zhuang:
Fixes #8567

The card's first deliverable was the **measurement**, and it decided the rest. Both questions answered, in the order the card set.

## (1) MEASURE — Postgres, raised for real

A throwaway PostgreSQL 16.13 cluster from the system PG16 binaries (`initdb` + `pg_ctl`, no container runtime — the docker daemon is unreachable here), driven through **knex + pg**, the same path `SqlDriver.upsert` takes: a table with no unique index on the target column, then `upsert(obj, row, ['email'])`. Read off the thrown error object, verbatim:

```
name        = error            (constructor DatabaseError)
code        = "42P10"          (typeof string)
severity    = "ERROR"
routine     = "infer_arbiter_indexes"
constraint  = undefined
detail      = undefined
status      = undefined
message     = insert into "plain" ("email", "id", "title") values ($1, $2, $3)
              on conflict ("email") do update set "title" = excluded."title"
              - there is no unique or exclusion constraint matching the ON CONFLICT specification
```

The card's stated expectation held — SQLSTATE `42P10`, and prose matching "there is no unique or exclusion constraint matching the ON CONFLICT specification" — so nothing had to be overturned. knex wraps it the same way it wraps SQLite's: STATEMENT, then ` - `, then the server's own sentence.

**`42P10` IS reachable by other conditions**, which the card asked to be checked and which changes the design. Measured on the same cluster, same session:

```
select id from plain order by 7  -> code=42P10  "ORDER BY position 7 is not in select list"
select id from plain group by 9  -> code=42P10  "GROUP BY position 9 is not in select list"
select nope from plain           -> code=42703  (a different code entirely)
```

`42P10` is `invalid_column_reference`, not "unbacked conflict target". So a code-only predicate over-matches, and the message must carry the verdict — the code channel is left deliberately unread on **both** dialects (SQLite's is the generic `SQLITE_ERROR`). Also measured: a **partial** unique index (`... where email is not null`) raises the identical 42P10 and sentence, which is correct — a partial index cannot serve as an arbiter.

**MySQL — the sub-question that needs no server, answered from the compiled SQL.** knex compiles the driver's exact call on the `mysql2` dialect to `ON DUPLICATE KEY UPDATE`, which takes no conflict target:

```
mysql2 -> insert into `plain` (`email`, `id`, `title`) values (?, ?, ?)
          on duplicate key update `title` = values(`title`)          <- no conflict target
pg     -> insert into "plain" (...) values ($1, $2, $3)
          on conflict ("email") do update set "title" = excluded."title"
```

The named keys never leave the process, so the server is never asked to find an index for them: **the condition cannot arise on MySQL**. What it does instead — merging on whichever unique key the row happens to collide with — is the separate defect the card anticipated; not fixed here, and not asserted here either, because nobody has watched a MySQL server do it. The live MySQL cell is **declared un-run** through `live-dialect-matrix.testkit.ts` rather than dropped.

## (2) THEN the home — the dialects share a shape, so the predicate graduates

Two dialects, two unrelated sentences, one channel: that is precisely the per-dialect alternation `UniqueViolationSignature` already encodes. `isUnbackedConflictTargetError` now lives in `@objectstack/types` beside `isUniqueViolationError`, one measured limb per dialect, no new dependency edge.

It stays a **separate** predicate. `isUniqueViolationError` answers the opposite condition, and putting the two side by side is what puts the warning where someone reaching for the wrong one will read it.

`driver-turso`'s private copy deliberately stays: that package does not depend on `@objectstack/types`, and its face speaks libsql/SQLite and only ever will, so the Postgres limb could never apply. Noted in place with the condition under which it should be deleted.

## The wording moved one clause, on both faces

"…and SQLite refuses the statement" is now "…and **the database** refuses the statement". Once recognition covers Postgres, that sentence pointed a Postgres operator at the wrong engine. Per #5240 both faces carry one wording, so `driver-turso`'s copy moved in the same commit. Nothing else in the refusal changed.

## ⚠️ Found while pinning: a pre-existing inversion, filed not fixed

The disjointness suite went red on its first run, and the measurement won. On **SQLite**, `isUniqueViolationError` already claims the unbacked-target error — its `unique constraint` limb matches SQLite's *missing*-index sentence, which ends `...any PRIMARY KEY or UNIQUE constraint`. Postgres escapes only on word order. Measured on real driver errors:

```
sqlite unbacked-target error -> isUniqueViolationError = true    <- wrong
pg     unbacked-target error -> isUniqueViolationError = false   <- correct
```

Filed as #8590 and left alone here — narrowing that predicate moves verdicts in six consuming packages. It is latent today (`upsert` is the only site compiling a caller-supplied conflict target, and recognition runs first in that catch), but a comment in `sql-driver.ts` asserted the opposite as fact and reasoned from it; that comment is corrected in place, and the ordering is now documented as load-bearing. The test pins both predicates' verdicts per dialect, so #8590's fix announces itself instead of landing silently. #8590 is not addressed by this PR.

## Reverse verification — direction predicted before each leg

- **Remove the Postgres limb** ⇒ predicted the Postgres cell reddens and every SQLite pin holds. Observed: `3 failed | 11 passed | 1 skipped` — envelope, wording and leak pins red on the live-postgres cell only; all 6 SQLite pins green; **both positive controls green on both cells**, which is what proves the limbs are independent rather than one regex covering both.
- **Reword one face only** (turso back to "SQLite") ⇒ predicted #8568's parity pin reddens. Observed: `1 failed | 9 passed`, the failure being `expect(localErr.message).toBe(remoteErr.message)`. Restored; both faces moved together, pin green.
- The `driver-sql` dist rebuild trap from #8568 was respected — `driver-turso` was only ever run against a freshly built `dist` carrying the reword (verified by grepping the built bundle).

## Tests

Full suites, live Postgres cell provisioned and **non-vacuous** (server `Asia/Shanghai`, process `TZ=America/New_York`, so the three-way zone skew guard passes rather than the matrix proving nothing):

```
@objectstack/types        11 files,  280 tests  passed
@objectstack/driver-sql   96 files, 1703 tests  passed | 1 skipped (the MySQL cell)
@objectstack/driver-turso 37 files,  986 tests  passed
typecheck: types, driver-sql, driver-turso — all clean
```

Gates derived from the actual changed paths via `scripts/pm/dispatch-gates.mjs`, all green: `check:changeset-gate-self-tests`, `check:objectui-changeset`, `check:test-source-alias`, `check:type-source-resolution`, `check:query-options-erasure`, `check:type-check-coverage`, `check:error-code-casing`, `check:nul-bytes`, `check-adr-0087-registration`, `check-changeset-no-major`, `check-empty-changeset`.

Changeset: a real `patch` across the three packages. The accept/reject set does not move — the same upserts fail, they fail legibly — but a Postgres caller's response body changes from raw statement text to a coded envelope, which is consumer-visible.

---

### DIFF
--- a/.changeset/unbacked-conflict-target-dialects.md
+++ b/.changeset/unbacked-conflict-target-dialects.md
@@ -1,0 +1,55 @@
+---
+"@objectstack/types": patch
+"@objectstack/driver-sql": patch
+"@objectstack/driver-turso": patch
+---
+
+fix(drivers): a `conflictKeys` upsert with no backing unique index refuses legibly on **Postgres** too, not only SQLite (#8567)
+
+`SqlDriver.upsert` recognised "the `ON CONFLICT` target is not backed by a
+PRIMARY KEY or UNIQUE index" on **SQLite only** (#8445). `driver-sql` serves
+three dialects, so on Postgres the raw driver error still escaped: `mapDataError`
+fell through to its default branch and served the thrown message as the whole
+response body — and that message is the **statement**, with no `code` for any
+client to branch on.
+
+**What a Postgres caller got, and now gets.** Measured against a real
+PostgreSQL 16.13 through the same knex + `pg` path the driver uses:
+
+```
+before:  code=42P10  status=undefined
+         message=insert into "plain" ("email", "id", "title") values ($1, $2, $3)
+                 on conflict ("email") do update set "title" = excluded."title"
+                 - there is no unique or exclusion constraint matching the ON CONFLICT specification
+
+after:   code=VALIDATION_ERROR  status=400
+         message=Cannot upsert into "plain" on conflict keys ("email"): no PRIMARY KEY or
+                 UNIQUE index backs them, … Fix by declaring the column(s) "unique: true" …
+```
+
+The accept/reject set does not move: the same upserts fail, they fail legibly.
+The server's own sentence is preserved on the error's `cause`, which no error
+mapper puts on the wire, so an operator debugging the table keeps the ground
+truth while the caller stops receiving SQL text.
+
+**One clause of the refusal wording changed, on both faces.** "…and SQLite
+refuses the statement" is now "…and **the database** refuses the statement".
+Once recognition covers Postgres, naming SQLite points a Postgres operator at
+the wrong engine. `driver-turso`'s remote-face copy moved in the same commit —
+the two are held word-for-word identical (#5240) by #8568's cross-face parity
+pin. No other sentence of the refusal changed.
+
+**Recognition is now a named, shared predicate.**
+`isUnbackedConflictTargetError` is exported from `@objectstack/types` beside
+`isUniqueViolationError`, carrying one measured message limb per dialect that
+can raise the condition. ⚠️ It is deliberately a **separate** predicate:
+`isUniqueViolationError` answers the opposite condition (an index exists and the
+row violated it), and a merged one would report a working constraint as a
+missing one.
+
+**MySQL is unaffected, by measurement rather than omission.** knex compiles
+`onConflict(...).merge(...)` on that dialect to `ON DUPLICATE KEY UPDATE`, which
+takes no conflict target — the named keys never leave the process, so the server
+is never asked to find an index for them and the condition cannot arise. The
+compiled statement is pinned; the live MySQL cell is reported as un-run rather
+than passing vacuously.
--- a/packages/drivers/driver-sql/src/sql-driver-upsert-conflict-target-dialects.test.ts
+++ b/packages/drivers/driver-sql/src/sql-driver-upsert-conflict-target-dialects.test.ts
@@ -1,0 +1,326 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#8567] The unbacked-conflict-target refusal, across the DIALECTS
+ * `driver-sql` actually serves — not just the one #8445 could raise it on.
+ *
+ * # What this card measured, and why measuring was the deliverable
+ *
+ * #8445 landed the refusal against SQLite's sentence and said so in the code:
+ * the container had no other server, and its dispatch ruled that transcribing
+ * another dialect's wording from memory is not evidence. That left Postgres
+ * and MySQL answering the raw driver error — `mapDataError` falling through to
+ * its default branch and shipping the STATEMENT, bound values included, with no
+ * `code` for any client to branch on. Two of three dialects, the same payload
+ * argument #8445 made for the third.
+ *
+ * So this file's first job is to be the place a dialect's wording is
+ * OBSERVED rather than assumed. Postgres was raised for real — system PG 16
+ * binaries, `initdb` + `pg_ctl`, no container runtime — through the same
+ * knex + `pg` path `upsert` takes, and the sentence it produced is now a limb
+ * of `isUnbackedConflictTargetError` in `@objectstack/types`.
+ *
+ * ```
+ * # PostgreSQL 16.13, knex 3.3.0 + pg 8.22.0
+ * upsert('plain', { email: 'a@b.com', title: 'x' }, ['email'])
+ *   -> name=error (DatabaseError) code=42P10 severity=ERROR status=undefined
+ *      routine=infer_arbiter_indexes  constraint=undefined  detail=undefined
+ *      msg=insert into "plain" ("email", "id", "title") values ($1, $2, $3)
+ *          on conflict ("email") do update set "title" = excluded."title"
+ *          - there is no unique or exclusion constraint matching the ON CONFLICT specification
+ * ```
+ *
+ * # The three cells are not symmetric, and pretending otherwise would lie
+ *
+ *  - **SQLite** runs everywhere, in-process. It is the cell that always
+ *    executes, so a regression in the shared predicate cannot hide behind an
+ *    unprovisioned matrix.
+ *  - **Postgres** runs when `OS_TEST_POSTGRES_URL` is set, and is REPORTED as
+ *    an un-run cell when it is not (`declareUnprovisionedCell`) — never a
+ *    silent pass. This is the cell whose wording this card added.
+ *  - **MySQL** cannot raise the condition at all, which is a measurement, not
+ *    an excuse. knex compiles `onConflict(...).merge(...)` on that dialect to
+ *    `ON DUPLICATE KEY UPDATE`, which takes **no conflict target**: the named
+ *    keys are dropped before the statement leaves the process, so the server is
+ *    never asked to find an index for them. That is checkable with no server at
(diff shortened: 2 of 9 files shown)


os-zhuang:
**ACCEPT** — reviewed against the tree at `91bdacb`, and the measurement re-run independently rather than taken from the report.

This card's whole deliverable was question (1), so I did not accept the Postgres numbers on the report's word. I started my own throwaway PG 16.13 cluster and raised the condition:

```
ERROR:  42P10: there is no unique or exclusion constraint matching the ON CONFLICT specification
LOCATION:  infer_arbiter_indexes, plancat.c:920
```

Byte-for-byte the sentence that went into the predicate's regex, and `infer_arbiter_indexes` matches the reported `routine`. The three follow-on claims corroborate too:

```
select id from plain order by 7   -> ERROR: 42P10: ORDER BY position 7 is not in select list   <- over-match, real
select nope from plain            -> ERROR: 42703: column "nope" does not exist                <- different code
create unique index ... where email is not null;
  insert ... on conflict (email)  -> ERROR: 42P10: there is no unique or exclusion constraint…  <- partial index is no arbiter
```

So **`42P10` really is reachable from an unrelated condition**, and the decision to leave the code channel unread on both dialects and let the message carry the verdict is forced by measurement, not stylistic. (Aside: `initdb` refuses to run as root, so this needed an unprivileged user — a real hurdle, cleared the same way on both sides.)

Verified in the diff, independently of the description:

- **`isUniqueViolationError` was documented, not changed.** `unique-violation.ts` is +22 lines of comment and **zero** code — so no verdict moves in the six consuming packages, exactly as the PR claims. This was the highest-risk thing to get wrong here and it is clean.
- **The new predicate refuses a loose limb on purpose.** Rejecting a bare `on conflict` fragment because knex prefixes statement text onto *every* upsert failure — including a unique violation, the opposite condition — is the sharpest call in this PR. A looser regex would have passed every test in it.
- **One wording on both faces (#5240).** `sql-driver.ts` and `remote-transport.ts` both now read "and the database refuses the statement". Naming SQLite to a Postgres operator would have been wrong the moment recognition widened.
- **#8568's parity pin is not in the diff at all** — and that is the point. Both faces moved together, so the runtime-vs-runtime comparison stayed green untouched, while #8445's single-face literal had to be updated. The literal was **updated, not weakened**: still a full-sentence `toBe`. The pin built yesterday did exactly the job it was built for, on its first real encounter.
- **Path surface** is `packages/**` + `.changeset/**` only. Not `docs/adr/**`, not `.claude/skills/**`. Graduating a predicate into `@objectstack/types` is triage-sanctioned for this card (its dispatch note names it), and #8590's triage confirms `types` is the cli lane's — so this crossing was authorised, not incidental.
- **The MySQL cell is declared un-run, not vacuously green**, through the existing `live-dialect-matrix.testkit.ts` convention.

**On #8590 — filing it instead of fixing it was the right call**, and the reasoning is better than "out of scope": the fix moves a consolidated predicate's verdicts across six packages, and the naive narrowing would silently drop Postgres' `violates unique constraint "..."`, which that limb has covered since it was inherited verbatim. Correcting the `sql-driver.ts` comment that asserted the opposite **as fact and reasoned from it** was the necessary half to do here — a wrong comment that a future author trusts is how the latent bug becomes live.

**One thing I can resolve that the dev could not.** The report declines to file the MySQL-behaviour card because no `mysqld` exists in the dev container and describing unobserved behaviour is the evidence standard #8445 ruled out — correct reasoning, incomplete premise. CI provisions a live MySQL:

```yaml
# .github/workflows/ci.yml:583
image: mysql:8.0
# :673
OS_TEST_MYSQL_URL: mysql://root:root@127.0.0.1:3306/conformance
```

So that half is measurable — in CI, not on a dev box. Nothing to change in this PR (the file here runs under Test Core, which has no such service, so its declared-un-run cell is honest for the job that runs it). I am filing the follow-up as a card whose **first deliverable is the measurement**, pointed at that runner, rather than one asserting what MySQL does.

Holding ready + queue until all 24 checks report green.

---

Issue #8566, opened 2026-08-13
`sys_webhook.headers_secret` has a required plaintext SHAPE that no write path enforces — the ordinary data API accepts any string, and the author only finds out at the next delivery


os-zhuang:
Found while implementing #8558 (**measured**, not traced). **Out of scope there and deliberately not fixed on that branch** — #8558 is scoped to the consumer seam (`resolveWebhookHeaders` must not fold "stored but unusable" onto "no headers authored"), and this is about not creating the state at the door. Unassigned, no `pm:queue`, for triage to grade.

## The defect

`sys_webhook.headers_secret` is a `Field.secret()` whose plaintext is **not** an opaque blob: it is a serialized header map with a required shape — a flat JSON object of string values — and `parseStoredHeaders` is the only reader.

Nothing validates that shape on the way in. The ordinary data API accepts any string, encrypts it like any other secret, mints a real `sys_secret` row, and leaves the column holding a perfectly valid `secret:` ref that reads back as the mask with `active: true`.

That field is directly admin-authorable, and its own description tells the author what to type into it:

> Custom HTTP headers sent with each delivery, as a JSON object ({"Authorization": "Bearer …"}). Encrypted at rest into sys_secret; reads return a mask, never the headers.

So the product asks for a JSON object, accepts anything, and stores it.

## Evidence — measured on a real engine

Reproduced against a real `ObjectQL` engine (in-memory driver double, reversible test crypto, the real `sys_webhook` schema), on `origin/main` at `719a21bfc`, through `engine.update()` — the ordinary data API, no privileged access:

| written to `headers_secret` | write | column at rest | read back | resolves to |
| --- | --- | --- | --- | --- |
| `{}` | accepted | valid `secret:` ref | mask, `active: true` | `{}` |
| `[]` | accepted | valid `secret:` ref | mask, `active: true` | `[]` |
| `{"X-Count":5}` | accepted | valid `secret:` ref | mask, `active: true` | `{"X-Count":5}` |
| `{"X-Team":{"name":"crm"}}` | accepted | valid `secret:` ref | mask, `active: true` | the nested object |
| `{X-Team: crm}` (a typo) | accepted | valid `secret:` ref | mask, `active: true` | the typo, verbatim |

Every one of them is a value the plugin can never use.

## Why it is worth fixing rather than tolerating

**Before #8558** this was silent and open: the subscription armed and the delivery went out missing its entire authored header map, correctly signed, `status: 'success'`, nothing logged.

[...]

PR #8616, opened 2026-08-14
fix(objectql): refuse empty-string writes to secret and password fields

### LINKED ISSUE #8559: objectql accepts `""` into a `secret` field: it encrypts the empty string, mints a `sys_secret` row for it, and every read path then reports "a secret is set"

os-zhuang:
Closes #8559

## What

The credential write door now refuses `""` (maintainer ruling recorded on the card — option 2, loud refusal over silent reinterpretation):

- **secret arm** — `encryptSecretFields` refuses before any crypto side effect: no `sys_secret` row is minted and no CryptoProvider is consulted, so the verdict is about the payload, never about what crypto happens to be wired.
- **password arm** — its own seam, `refuseEmptyPasswordFields`, guards the same doors (single/bulk insert, single-id update, multi update) without routing plaintext-at-rest fields through the encryption path. Scope is the new `collectMaskedPasswordFields` collector, kept beside the read-mask collector so the two cannot drift: `managedBy: 'better-auth'` objects are exempt, mirroring the read mask — with no mask there, the "reads as set, holds nothing" contradiction cannot arise, and that write path belongs to the auth subsystem.

The refusal is a located ADR-0112 envelope: `EmptyCredentialWriteError` with `code: 'VALIDATION_ERROR'` (standard catalog member — no ledger entry needed) and `status: 400`, message naming the field (`object.field`) and the correct spellings — write `null` to clear the stored credential, omit the field to leave it unchanged. The class, its `EMPTY_CREDENTIAL_REFUSAL_CODE` / `EMPTY_CREDENTIAL_REFUSAL_STATUS` pair, and `collectMaskedPasswordFields` are exported from the package root so consumers branch on `code`/`status` rather than message text.

On the per-row outcome path (`insertMany`), the refusal is reported per row and surviving rows land with their own cipher rows.

Rows that already decrypt to `""` are historical and stay reachable — this closes the door, it does not migrate history; the consumer-side defenses (#8542, #8558) remain load-bearing.

## Pin inventory (the card's measured contrast table)

Pinned unchanged alongside the refusal, all in `packages/objectql/src/secret-fields.test.ts`:

| shape | pin |
|---|---|
| echoed read-mask dropped (secret + password) | pre-existing pins, untouched |
| cleartext re-encrypts into a fresh ref | pre-existing pin, untouched |
| `null` clears — stored null, reads report null (now truthfully) | new pins, both arms |
| better-auth password column takes `""` unjudged | new pin (deliberate scope edge) |

New refusal pins (8): insert / update / multi update on both arms, each asserting the full envelope (class, `code`, `status`, `object`/`field`/`fieldType`, located first sentence, null prescription) plus no-side-effect facts (no business row, no `sys_secret` row, zero encrypt calls); a payload-verdict pin (refusal fires with NO CryptoProvider wired, not the fail-closed crypto error); and a per-row `insertMany` outcome pin. The stub driver gained `updateMany` (affected-count contract) so the multi:true path exists on the double at all — it reuses the file's existing grandfathered matcher; `check:where-matcher-conformance` verified: no new matcher, baseline key set unchanged.

## Reverse verification

Fix committed first, then `git checkout origin/main -- packages/objectql/src/engine.ts`, suite re-run. Predicted direction: the 8 refusal pins go red (writes accepted again), everything else stays green. Observed exactly that — 8 failed / 28 passed; the failing set is precisely the new refusal pins (both arms' insert, update, multi update, the no-CryptoProvider payload-verdict pin, and the per-row `insertMany` pin). Fix restored from the commit: 36/36 green.

## Verification

- `pnpm --filter @objectstack/objectql test` — 202 files, **3571/3571 pass**
- `pnpm --filter @objectstack/objectql typecheck` — clean
- eslint over the four changed source files — clean
- Gates, derived from actual changed paths via `scripts/pm/dispatch-gates.mjs` plus convention: `check:where-matcher-conformance` ✓ · `check:durability-log-level` ✓ · `check:nul-bytes` ✓ · `check:error-code-casing` ✓ · `check:changeset-gate-self-tests` ✓ · `check:objectui-changeset` ✓ · `check:stack-collection-maps` ✓ · `check-adr-0087-registration` ✓ (non-breaking changeset) · `check-changeset-no-major` ✓ · `check-empty-changeset` ✓ · `check-engine-split-ratio` ✓ (informational) · `check:query-options-erasure` ✓ (at ceiling, none added) · `check:type-check-coverage` ✓ · `check:type-check-debt --re-measure` ✓ with the full package closure built — no ledger entry above its recorded number, i.e. the new test code adds zero test-layer type errors to the hidden-test package.

## Notes for review

- `check:objectui-pin-fresh` is red as pre-existing repo state (#3340) and fires on any `.changeset/*` touch — deliberately not addressed here; #3340 remains open.
- #8566 is not addressed here (same ruling batch, disjoint files); this PR lands first so that shape hook never needs an empty-string special case. #8566 remains open.
- Changeset: `patch` on `@objectstack/objectql` — fix-typed acceptance narrowing at the data API, following the `action-body-write-not-found-gate` precedent.

---

### DIFF
--- a/.changeset/secret-empty-string-refusal.md
+++ b/.changeset/secret-empty-string-refusal.md
@@ -1,0 +1,46 @@
+---
+"@objectstack/objectql": patch
+---
+
+fix(objectql): the credential write door refuses `""` — a secret/password field can no longer be set to an empty credential that every read reports as present (#8559)
+
+`ObjectQL.encryptSecretFields` treated the empty string as an ordinary value: it
+encrypted `""`, minted a real `sys_secret` row whose ciphertext decrypts to
+nothing, and rewrote the business column to a perfectly valid `secret:` ref.
+From that moment every read surface reported "a secret is set" (the field masks
+as `••••••••`) while `resolveSecretField` returned `""` — the contradictory
+state the webhook seam (#8542) and its siblings had to defend consumers
+against. A `password`-typed field stored `""` verbatim, producing the same
+contradiction without the cipher row.
+
+**Both arms now refuse `""` at the write door** (maintainer ruling on #8559,
+option 2 — loud refusal over silent reinterpretation):
+
+- `secret` fields refuse in `encryptSecretFields`, before any crypto side
+  effect — no cipher row is minted, no CryptoProvider is consulted.
+- `password` fields refuse at their own write seam
+  (`refuseEmptyPasswordFields`), guarding the same doors (single/bulk insert,
+  single-id update, multi update) without routing plaintext-at-rest fields
+  through the encryption path. `managedBy: 'better-auth'` objects are exempt,
+  mirroring the read mask's scope — their credential writes belong to the auth
+  subsystem.
+
+The refusal is a located ADR-0112 envelope: `EmptyCredentialWriteError` with
+`code: 'VALIDATION_ERROR'`, `status: 400`, and a message naming the field
+(`object.field`) and the correct spellings — **write `null` to clear the stored
+credential; omit the field to leave it unchanged**. The class and its
+`EMPTY_CREDENTIAL_REFUSAL_CODE` / `EMPTY_CREDENTIAL_REFUSAL_STATUS` pair are
+exported from `@objectstack/objectql`, as is `collectMaskedPasswordFields`, so
+consumers branch on `code`/`status` rather than message text. On the per-row
+outcome path (`insertMany`), the refusal is reported per row and surviving rows
+land normally.
+
+Everything else at the door is pinned unchanged: echoing the read mask back
+still means "unchanged" (the key is dropped), cleartext still re-encrypts into
+a fresh ref, and `null` still clears — after which reads truthfully report "not
+set". Rows that already hold an empty credential are untouched: this closes the
+door, it does not migrate history, and the consumer-side defenses (#8542,
+#8558) remain load-bearing for those.
+
+If a client meant "clear this credential" and sent `""` (the natural form-
+control spelling), send `null` instead — the refusal message says exactly that.
--- a/packages/objectql/src/engine.ts
+++ b/packages/objectql/src/engine.ts
@@ -133,11 +133,13 @@ import type { ICryptoProvider, CryptoHandle } from '@objectstack/spec/contracts'
 import {
   collectSecretFields,
   collectMaskedReadFields,
+  collectMaskedPasswordFields,
   collectInternalReadFields,
   collectCredentialFields,
   makeSecretRef,
   parseSecretRef,
   isSecretRef,
+  EmptyCredentialWriteError,
   SECRET_MASK,
 } from './secret-fields.js';
 import { pluralToSingular, ExternalWriteForbiddenError } from '@objectstack/spec/shared';
@@ -4928,6 +4930,16 @@ export class ObjectQL implements IObjectQLEngine {
    *    overwrite the stored value.
    *  - No secret fields on the object ⇒ no further work (fast path, no crypto).
    *  - `null`/`undefined` secret value ⇒ left as-is (clears the secret).
+   *  - [#8559] `""` secret value ⇒ REFUSED with {@link EmptyCredentialWriteError}
+   *    (ADR-0112 `VALIDATION_ERROR`/400, message names `null` as the clear
+   *    spelling). Encrypting it would mint a `sys_secret` row that decrypts to
+   *    nothing behind a perfectly valid ref, so every read reports "a secret is
+   *    set" while the dereference returns nothing — the contradictory state
+   *    #8542/#8558 defend consumers against. Refused, not folded into the
+   *    `null` branch: silently reinterpreting the write was explicitly rejected
+   *    (maintainer ruling 2026-08-13 on #8559). The `password` half of the same
+   *    ruling lands at its own seam — {@link refuseEmptyPasswordFields} — since
+   *    the two types share the read mask but not this encryption path.
    *  - Secret value already a ref (re-save of an unchanged ref) ⇒ left as-is.
    *  - **Fail-closed:** any other secret value with no CryptoProvider registered,
    *    or no reachable `sys_secret` store, THROWS — never persists cleartext.
@@ -4960,6 +4972,11 @@ export class ObjectQL implements IObjectQLEngine {
       const value = row[field];
 
       if (value === null || typeof value === 'undefined') continue; // clear
+      // [#8559] The empty string is neither a clear (`null` is) nor a storable
+      // credential — refuse it BEFORE the crypto gates below: this is a verdict
+      // on the payload, so it must not depend on whether a CryptoProvider or
+      // sys_secret store happens to be wired.
+      if (value === '') throw new EmptyCredentialWriteError(object, field, 'secret');
       if (isSecretRef(value)) continue; // already encrypted ref
 
       if (!this.cryptoProvider) {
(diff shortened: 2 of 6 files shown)


os-zhuang:
Patch round — `Test Core (2/3)` red at `d586be23` (the second red was the shard attestation reporting the failed shard, one root cause).

**The actual failing assertions** — `packages/plugins/plugin-webhooks/src/webhook-secret-at-rest.test.ts`, two tests: the #8542 pin "refuses to arm when the stored secret was emptied through the ordinary data API" and its #8558 twin for `headers_secret`. Neither failed on its consumer-side assertion: each died at its **setup** line — `engine.update('sys_webhook', { signing_secret: '' })` (resp. `headers_secret`) now gets `EmptyCredentialWriteError` (`VALIDATION_ERROR`/400) from the door this PR closes. The pins manufactured their trigger state through exactly the road the ruling shut.

**What changed (tests only — the ruling is untouched, no carve-outs added)**: both pins now plant the decrypts-to-nothing state **below the engine** — a driver-level `sys_secret` row with empty ciphertext plus a valid `secret:` ref on the column — the same route as the file's adjacent hand-pasted-column pin, with names and comments restated on the historical-row premise: #8559 closes the door but does not migrate history, so pre-refusal rows can still hold this shape and the consumer-side refusal-to-arm stays load-bearing for them. #8542/#8558 coverage is preserved, not weakened. The non-empty unusable spellings in the `it.each` (`'{}'`, `'[]'`, bad JSON…) still write through the ordinary data API unchanged — only the literal empty string moved to historical-only.

**Class sweep before pushing** — repo-wide grep over test files for empty-string writes to credential-named keys (literal and computed-key forms), plus every package whose test fixtures carry `secret`/`password`-typed fields. `plugin-webhooks` was the **only** class member. Adjudicated non-members, confirmed by their green suites: `runtime` http-dispatcher (`password: ''` in a better-auth sign-in HTTP payload — auth subsystem, not an engine-typed-field write), `plugin-auth` admin endpoint (`password: ''` beside `generatePassword: true`), `service-datasource` turso config literal (`secret: ''` in driver config, not a row write), `service-settings` (`token: ''` through the `sys_setting` channel, its own encryption path). `spec` excluded with reason: it is **upstream** of the door (objectql depends on spec), so no spec test can reach the engine write path; its credential-typed fixtures feed schema parsing only.

**Packages re-run, all green**: plugin-webhooks 81/81 · runtime 2369/2369 · plugin-auth 1195/1195 · service-settings 451/451 · service-datasource 376/376 · service-messaging 219/219 · qa/dogfood 770 passed / 1 expected-fail (pre-existing marker) / 3 skipped · objectql 3571/3571 (unchanged from the base commit run). An initial broad dogfood red reproduced as an unbuilt `@objectstack/verify` dist in the fresh worktree — environmental, gone after building its closure.

**Gates re-run for the new changed path** (derived via `scripts/pm/dispatch-gates.mjs` + convention): `check:test-source-alias` ✓ · `check:type-source-resolution` ✓ · `check:query-options-erasure` ✓ (at ceiling, none added) · `check:i18n` ✓ · `check:where-matcher-conformance` ✓ (169 matchers, none new) · `check:nul-bytes` ✓ · `check:type-check-coverage` ✓ · `check:type-check-debt --re-measure` ✓ (no ledger entry above its recorded number) · eslint + `tsc --noEmit` on plugin-webhooks ✓.

Head: `e8065a932`, same branch, same PR.

---

---

os-zhuang:
**PM review (`domain:engine-core` seat, #6019): ACCEPT.** Ready + auto-merge (squash); it cannot fire until the last `TypeScript Type Check` job passes.

⭐ **The patch round's fix was better than the one I asked for, and the difference matters.** My instruction said a test writing `""` to a credential field "should now write `null` (clear) or omit the field." That would have been wrong here: the two `plugin-webhooks` pins were not *asserting* on the write — they were using it as **setup** to manufacture a decrypts-to-nothing row, and `null` clears rather than creating that state. Following my instruction would have quietly hollowed out the #8542/#8558 coverage into a test of a cleared field. Planting the state **below the engine** (cipher row with empty ciphertext + a valid `secret:` ref, matching the file's own adjacent hand-pasted-column pattern) is the correct call, and it keeps the premise honest: **the door is closed, history is not migrated**, so the consumer-side refusal-to-arm stays load-bearing for pre-refusal rows. The renamed test titles now say so.

Checked against the tree:

- **The secret refusal precedes the crypto gates**, deliberately — a payload verdict must not depend on whether a `CryptoProvider` or `sys_secret` store happens to be wired. Pinned by a test asserting `EmptyCredentialWriteError` (not the fail-closed "no CryptoProvider" error) on a bare engine. That ordering is the whole reason the refusal is trustworthy.
- **The password arm is its own seam** (`refuseEmptyPasswordFields`), not folded into the encryption path — right, since `password` is plaintext at rest and shares the read mask but not the cipher path. `managedBy: 'better-auth'` is exempt, mirroring the read mask's scope, and that exemption is pinned rather than assumed.
- **Drift guard on scope**: `collectMaskedPasswordFields` is asserted to be a subset of `collectMaskedReadFields`, so the refusal's scope cannot drift from the mask that justifies it.
- **All four doors guarded** (single/bulk insert, single-id update, multi update), with the `insertMany` call inside the same `try` so partial mode reports the refusal per row instead of killing the batch.
- ⭐ **The stub driver's new `updateMany` carries its reason**: without it the engine's capability check rejects a multi update *before* any write-door seam is reached, so the multi-update pin would have passed vacuously. Naming that is the difference between a pin and a decoration.
- **The class sweep is reported as a negative result with its evidence** — `plugin-webhooks` was the only member, with four adjudicated non-members named individually (runtime http-dispatcher, plugin-auth admin endpoint, service-datasource turso config, service-settings token channel) and `spec` excluded as upstream of the door.

One observation, ⛔ not a blocker: the changeset is `patch` on `@objectstack/objectql` for a change that turns previously-accepted writes into `400`s. That is semver-relevant in the abstract, but it matches this repo's convention (`check:changeset-no-major`) and the maintainer ruled the accept/reject change knowingly. Recording it rather than re-opening it.

---

Issue #8615, opened 2026-08-14
check-where-matcher-conformance's discovery heuristic misses single-param inline filter closures — 16 files carry the same shape-(b) blind spot, undetected


os-zhuang:
Filed while working #8582 (session `session_01RDTnVvsgA6cUZ4xFVtPZRy`), which swept the 78 test-double `matches(row, where)` fakes that `pnpm check:where-matcher` discovered and grandfathered in `scripts/where-matcher-conformance.baseline.json`.

## The gap

`scripts/check-where-matcher-conformance.mjs`'s discovery only considers a candidate matcher when it is a function with **two or more identifier parameters** — see `discoverInSource`'s `params.length >= 2 && params[0] && params[1]` gate. That is deliberate (it is how the control probe stays name-independent), but it has a structural blind spot: a **single-param** `.filter()` callback that closes over `where`/`filter` from its *enclosing* scope, rather than taking it as its own second parameter, is invisible to discovery — even when its body is textually identical to a matcher the gate already knows how to grade.

Measured on `main` (as of this filing), this exact shape —

```
.filter((r) => Object.entries(where).every(([k, v]) => r[k] === v))
```

— or a same-file `filter` variant, appears **17 times across 16 files**, none of them reachable by `pnpm check:where-matcher` and none in its ledger:

- `packages/plugins/plugin-email/src/attachment-reclaim.test.ts:58`
- `packages/plugins/plugin-email/src/template-locale-resolution.test.ts:67`
- `packages/plugins/plugin-security/src/bootstrap-declared-positions.test.ts:29`
- `packages/plugins/plugin-security/src/bootstrap-declared-permissions.test.ts:18`
- `packages/plugins/plugin-security/src/bootstrap-platform-admin.test.ts:30`
- `packages/plugins/plugin-auth/src/admin-import-users.test.ts:36`
- `packages/core/src/security/api-key.test.ts:20`
- `packages/metadata-protocol/src/migrations/recorded-by-sentinel.test.ts:76`
- `packages/core/src/utils/migration-journal.test.ts:84`
- `packages/rest/src/import-runner-selfref.test.ts:47`
- `packages/rest/src/import-runner-idempotency.test.ts:56`
- `packages/services/service-storage/src/lax-deviation-reclamation-gate.test.ts:61` and `:88` (two instances)
[...]

PR #8618, opened 2026-08-14
test(where-matcher): empty the combinator-blind ledger — 75 files, refuse unrecognised $-keys

### LINKED ISSUE #8582: finding: 78 test-double WHERE matchers have no combinator branch at all — they read `$or` as a field name and silently drop the row
Filed unassigned by the `domain:engine-core` dev seat while building #8494's guard (PR #8581), session `session_01RDTnVvsgA6cUZ4xFVtPZRy`. ⛔ Not queued, not graded by me — triage grades it.

## The fact

#8494 landed `pnpm check:where-matcher` (`scripts/check-where-matcher-conformance.mjs`), which lifts every in-memory WHERE matcher out of its test file and asks it four combinator questions. Measured across `packages/**/*.test.ts` on `main`:

| verdict | count |
|---|---|
| discovered matchers | 168 |
| conforming | 89 (4 of them by **refusing** — throwing on an unrecognised `$`-key) |
| **silently wrong** | **78** |
| unjudged (could not be lifted and executed) | 1 |

The 78 are all **failure shape (b)** from #8494: an `Object.entries(where).every(…)` matcher with no combinator branch at all. It treats `$or` as an ordinary FIELD NAME, compares `row.$or` (`undefined`) against the array, gets no match, and silently excludes the row — so a suite that ever passes one asserts on an **empty result set with nothing erroring**.

They are grandfathered in `scripts/where-matcher-conformance.baseline.json`, which is the worklist: 75 files, shrink-only, reconciled in both directions. This issue is the ledger's other half — the sweep that empties it.

⭐ Note what is deliberately **not** in that ledger: failure shape (a), the `$or` early return. PR #8581 fixed the five live instances it found rather than baselining them, so shape (a) is enforced with an **empty** ledger. Only shape (b) is owed.

## Dormant, and that is not a reason to close it

#8493 measured its three instances as never invoked with a combinator today, and the same is very likely true of most of these 78 — the liveness of the remaining 75 is **unmeasured**, and measuring it per file is part of the work, not a precondition for filing.

The card this came from already answered the "it's dormant" objection in its own words, and it applies unchanged here: *"the next test that adds an `$or` inherits a double that lies."*

## The cheap fix is usually the right one

Two ways to clear an entry, and option 2 is one line:

1. Make the matcher conjoin `$or`/`$and` with its sibling keys.
2. Make it **refuse** what it does not implement — `throw` on an unrecognised `$`-prefixed key.

Option 2 is already the recorded practice in `packages/objectql/src/engine-autonumber-*.test.ts`, with its own rationale: *"silently ignoring an unknown operator would let a bad query pass as a good one."* A double that only ever sees scalar equality does not need a combinator implementation — it needs to stop lying. Those four files are exactly the "4 conforming by refusing" above, so the pattern is proven against the gate.

## Suggested shape

Sweep per package as separate lanes rather than one 75-file PR (`metadata-protocol`, `objectql`, `plugin-*`, `runtime`, `services/*`, `triggers/*`), lowering the baseline counts in the same PR each time — the gate fails if a count falls without the ledger following it down. The 1 unjudged entry is its own small question: whether the extractor should reach it, or whether that matcher should be restructured.

⛔ Not a request to extract a shared `matchesWhere` — ruled NO on #7620, reasoning specific to independent test doubles, and not re-litigable here.

Refs: #8494 · PR #8581 · #7620 · PR #7846 · PR #8493 · #7619

---

os-zhuang:
Closes #8582

## What

Sweeps all 75 files (78 matchers + 1 unjudged) grandfathered in `scripts/where-matcher-conformance.baseline.json` — test-double `matches(row, where)` fakes with **no combinator branch at all**, which read `$or`/`$and` as an ordinary field name, compared `row.$or` (`undefined`) against the array, and silently dropped every row a combinator query touched.

Default remedy (per the issue): make the double **refuse** — `throw` on an unrecognised `$`-prefixed key, the practice already pinned in `packages/objectql/src/engine-autonumber-*.test.ts`. Implemented the combinator instead only where a suite genuinely exercises it (see below).

`pnpm check:where-matcher` (final state):
```
✓ where-matcher conformance holds: 168 matcher(s) discovered, 168 answer the combinator battery
  correctly or refuse it loudly (82 refuse).
  0 silently-wrong and 0 unjudged matcher(s) in 0 grandfathered file(s); none new.
```
The ledger is now `{}` — empty, matching failure-shape-(a)'s steady state.

## Full file surface (cross-domain exception, per the claim comment)

75 test files across `metadata-protocol`, `objectql`, `plugin-approvals`, `plugin-auth`, `plugin-email`, `plugin-hono-server`, `plugin-reports`, `plugin-security`, `plugin-sharing`, `rest`, `service-analytics`, `service-automation`, `service-messaging`, `service-queue`, `service-settings`, `service-storage`, `trigger-schedule` — **test doubles only, zero production files.** Disjoint from the two other in-flight production-file claims this batch (#8559 `packages/objectql/src/engine.ts`, #8460 `packages/objectql/src/registry.ts`) — neither touched here.

Commits, per-package batches with the ledger lowered in the same commit each time (self-proving against the gate):
1. `metadata-protocol`, `objectql`, `plugin-approvals`, `plugin-auth` (39 files)
2. `plugin-email`, `plugin-hono-server`, `plugin-reports`, `plugin-security`, `plugin-sharing`, `rest` (22 files)
3. `service-analytics`, `service-automation`, `service-messaging`, `service-queue`, `service-settings`, `service-storage`, `trigger-schedule` (14 files)
4. One follow-up fix (below)

## Two files needed more than the mechanical guard

- **`packages/plugins/plugin-sharing/src/business-unit-graph.test.ts`** already implements `$or` correctly (a real, exercised recursive conjunction). Added a refusal for any *other* `$`-key (`$and` included) rather than growing a second combinator implementation nothing in the suite exercises.
- **`packages/services/service-analytics/src/__tests__/objectql-daterange.test.ts`** — the blanket refusal I first landed (batch 3) turned a passing test red: `query()` drives through `fieldLeaves`'s real NULL-safe `$ne` expansion (#5298), which emits `$and: [{ $or: [{field:null},{field:{$ne}}] }]` — this double genuinely receives `$or` at runtime, it is not dormant here. Per the issue's own guidance ("option 1 only where a suite genuinely exercises combinators"), implemented `$or` the same way the file already implements `$and` (a `.some()` over the branches) instead of refusing it. Caught by running the full test suite before reporting, not by the gate (the gate only asks the four synthetic battery questions against the lifted function in isolation — it has no way to see what a *real* caller in this file actually hands the matcher).

## The 1 unjudged matcher

`packages/plugins/plugin-email/src/outbox-sweep.test.ts`. Its nested-operator throw message (`` `fakeEngine: unsupported operator ${op}` ``) coincidentally repeats its enclosing factory function's exact name (`function fakeEngine(seed: RowSeed[])`), and the gate's `identifiersIn()` text-scanner reads that string-literal word as if it were a reference to the factory — pulling the whole closure (and its unsuppliable `seed` parameter) into the standalone extraction, which then threw `ReferenceError: seed is not defined` no matter how the retry/drop loop iterated.

**Disposition: RESTRUCTURE**, not extractor-reach — renamed the message so it no longer shadows the enclosing scope (one line, zero blast radius) rather than changing the shared `identifiersIn()` text-scanner every one of the 168 discovered matchers depends on. The matcher now judges `CONFORMING` (refuses).

## Out-of-scope finding

Filed #8615: the gate's discovery heuristic requires a matcher to be a function with **two or more identifier parameters**, so a single-param `.filter((r) => Object.entries(where).every(...))` closure that captures `where`/`filter` from an *enclosing* scope is structurally invisible to it — 17 such sites across 16 files carry the identical shape-(b) pattern, completely ungraded by `check:where-matcher`. Not fixed here (outside #8582's worklist, which is scoped to the ledger); left for triage.

## Verification

- `pnpm check:where-matcher` — green (`0 silently wrong, 0 unjudged`), see above.
- `pnpm test` scoped to all 17 touched packages (`metadata-protocol`, `objectql`, `plugin-approvals`, `plugin-auth`, `plugin-email`, `plugin-hono-server`, `plugin-reports`, `plugin-security`, `plugin-sharing`, `rest`, `service-analytics`, `service-automation`, `service-messaging`, `service-queue`, `service-settings`, `service-storage`, `trigger-schedule`) — **all green**, real numbers:
  - metadata-protocol 85/85 files, 1261/1261 tests
  - objectql 202/202 files, 3559/3559 tests
  - plugin-approvals 21/21 files, 458/458 tests
  - plugin-auth 53/53 files, 1195/1195 tests
  - plugin-reports 3/3, 70/70 · plugin-hono-server 18/18, 211/211 · plugin-sharing 21/21, 569/569
  - service-queue 3/3, 53/53 · service-messaging 20/20, 219/219 · service-settings 23/23, 451/451
  - service-analytics 75/75, 1675/1675 (after the $or fix) · service-storage 24/24, 361/361
  - plugin-email 23/23, 369/369 · plugin-security 55/55, 1079/1079
  - rest 114/114, 1881/1881 · service-automation 80/80, 955/955 · trigger-schedule 4/4, 46/46
- `pnpm typecheck` on every touched package that declares its own `typecheck` script — clean (`service-analytics`, `service-automation`, `service-storage` carry no `typecheck` script; pre-existing DEBT/EXEMPT-ledger packages, untouched by this test-only diff).
- Derived gates (`node scripts/pm/dispatch-gates.mjs <changed paths>`), all green: `check:cross-package-test-inputs`, `check:durability-log-level`, `check:test-source-alias`, `check:type-source-resolution`, `check:engine-split-ratio`, `check:query-options-erasure`, `check:type-check-coverage` (436 frozen DEBT / 1533 frozen TEST_DEBT — unchanged from `main`, no regression), `check:i18n` (9 packages, all in sync), `check:nul-bytes`.

## Not in scope

- No shared `matchesWhere` extraction — ruled NO on #7620, not re-litigated.
- Test-only surface — no changeset (`skip-changeset` label applied).
- `#8559` / `#8460`'s production files (`packages/objectql/src/engine.ts`, `packages/objectql/src/registry.ts`) — not touched.

---


os-zhuang:
**PM review (`domain:engine-core` seat, #6019): ACCEPT.** Ready + auto-merge (squash). All 35 checks green or skipped, including `check:where-matcher` itself and all three Test Core shards.

⭐ **Two things here are better than the card asked for, and both came from running the real suites rather than trusting the gate's synthetic battery.**

1. **`service-analytics/objectql-daterange.test.ts` genuinely receives an `$or`** — production `fieldLeaves`' NULL-safe `$ne` expansion hands it one. The gate's four-probe battery could not have told you that; only the real suite did. So it correctly got a real `$or` implementation instead of the default `throw`. ⚠️ That is the standing caveat on this whole remedy: "refuse" is right only where the combinator never arrives, and the *suite*, not the gate, is what decides that.
2. **The 1 unjudged matcher was a detector bug, not a double bug.** Its throw-message text coincidentally repeated its enclosing factory's exact name, so the gate's text-scanning `identifiersIn()` pulled in an unsuppliable closure parameter. Fixed by restructuring the message — ⛔ correctly *not* by touching the shared extractor.

⚠️ **The empty ledger (`{}`) must not be read as "the repo is swept clean."** The same sweep filed **#8615**: the discovery heuristic misses single-param inline filter closures (`.filter((r) => Object.entries(where).every(...))` capturing `where` from an enclosing scope), and **17 sites across 16 files carry the identical shape-(b) pattern, invisible to this gate**. So `0 silently-wrong in 0 grandfathered files` is a true statement about what the gate can *see*, and the gate's field of view is now the open question. That is exactly the "the instances are fixed, nothing holds them fixed" shape one level up again — ⛔ triage's to grade, not this seat's, but it belongs in the record next to the empty ledger.

Cross-domain conditions met: full file surface declared, 17 packages re-run green, ledger lowered in the same commit so the PR self-proves.

---

Issue #8601, opened 2026-08-14
`sys_position`'s shipped en bundle still asserts installation-wide uniqueness that #8556 corrected at source


os-zhuang:
Measured during #8554 / PR #8599, on `origin/main`.

#8556 (landed `6b441a84`) corrected `sys_position`'s published uniqueness claim at source — the spec `describe()` went from "Unique position name" to "Position name, unique per organization", and the generated reference page was regenerated from it.

**The translation bundle was not updated, and `main` currently ships the old claim.** `packages/plugins/plugin-security/src/translations/en.objects.generated.ts` still reads *"Unique machine name for the position"* — so the corrected source and the English text an admin actually reads in Setup disagree, right now.

## Why the gate did not catch it

`check-i18n-bundles.mjs --write` printed `regenerated` for all nine packages and **changed nothing**, and `check:i18n` reports `all bundles in sync` either way. That is **#8543** — the merge mode never updates an *existing* field description, so editing one leaves the bundle silently stale with the gate green.

This card is the **data residue**, not the generator defect. It does not need #8543 to land first: the bundle header documents the workflow — *"Edit translations in place; re-run extract (with --merge) to fill new gaps. Do not hand-edit the structure — only the leaf string values"* — and PR #8599 used exactly that route for its own five objects, correcting the leaf values in all four shipped locales.

## What to do

Correct `sys_position`'s leaf values in all four shipped locales (`en`, `es-ES`, `ja-JP`, `zh-CN`) to match the source `describe()` that #8556 landed. Leaf values only; ⛔ do not hand-edit the structure, and ⛔ do not touch the generator (that is #8543's).

⚠️ Worth sweeping while there: #8556 was one card. Any other landed source correction whose bundle went stale the same way has the same residue, and the gate will not tell you. A short sweep for bundle strings that contradict their current source `describe()` would bound it.

## Related

- #8556 / #8468 — where the source was corrected.
- #8543 — the generator defect that let this happen silently (`needs-user-decision`, another seat).
- #8599 / #8554 — where this was measured.

Filed by the `domain:metadata` PM seat.

hotlong:
Triage confirmation: native type **Bug** set (the shipped en bundle contradicts the corrected source `describe()` — declared vs shipped, measured on `origin/main`). Labels accepted as filed.

[...]

PR #8619, opened 2026-08-14
fix(plugin-security): withdraw the sys_capability Deactivate dialog's false grant-revocation claim (#8535)

### LINKED ISSUE #8535: `sys_capability.active` is read by nothing — the Deactivate action tells the admin grants stop resolving, and they do not

### LINKED ISSUE #8613: `sys_permission_set.active` and `sys_position.active` are unenforced too — both Deactivate dialogs promise access stops, and it does not
Found while implementing #8535 (which withdraws the identical false claim on `sys_capability` only). Filed unassigned — **out of scope for #8535's PR**, which deliberately touches nothing but `sys_capability`.

#8535 established that `sys_capability.active` is read by nothing while its Deactivate dialog claims grants stop resolving. The same shape exists on the **other two RBAC catalogue objects**, and on those it is worse: a permission set and a position are the **grant** itself, not a catalogue entry.

## The claims

`sys-permission-set.object.ts`, `deactivate_permission_set`:

> Deactivate this permission set? Existing assignments stay in place but stop granting access until re-activated.

`sys-position.object.ts`, `deactivate_position`:

> Deactivate this position? Users keep their assignment but the position stops granting permissions until re-activated.

Both ship translated into all four locales.

## What actually happens (measured on `b45c71e85a`)

Nothing filters on `active` anywhere in the resolution chain:

- `SecurityPlugin.resolvePermissionSetsForContext` (`security-plugin.ts:3626`) builds `requested` from `context.positions` + `context.permissions` plus the additive baseline. No `active` check.
- `PermissionEvaluator.resolvePermissionSets` (`permission-evaluator.ts:398`) matches metadata sets by `name`, then bootstrap sets by `name`, then falls through to the DB loader. No `active` check in any of the three sources.
- The DB loader (`security-plugin.ts:812`) queries `sys_permission_set` with `where: { name: { $in: names } }` — **no `active` predicate** — and the row is mapped to a `PermissionSet` that does not carry `active` at all, so nothing downstream could filter on it even if it wanted to.
- Every non-test `sys_position` row read (`delegated-admin-gate.ts`, `security-plugin.ts`, the bootstrap seeders) looks up by `name` or `id`. None reads `active`.

So `active: false` on a permission set or a position changes a column value and a badge in Setup. The assignments keep granting.

## Why this is worth its own card

Same direction of falsehood as #8535, higher blast radius. An admin revoking a compromised or over-broad permission set is told in a confirmation dialog that access stopped; it did not. The admin's likely next action is to *not* do the thing that would actually work (delete the set, or remove the assignments).

Note `sys_position` was partially touched by #8556/#8601, but that was different text — the `deactivate_position` dialog above is still on `main` making the claim.

## Not a recommendation

The direction is the same contract decision the maintainer already ruled on for `sys_capability` (2026-08-13, ADR-0049 enforce-or-remove, option B: withdraw the claim rather than put the registry on the authorization hot path). Whether that ruling extends to these two is **not** something this finding assumes: unlike a capability catalogue, a permission set's `active` flag has a plausible real enforcement point (the DB loader's `where`, one predicate) and a plausible real business pull (switching a grant off without deleting it). Both options are live and the trade-off is different from #8535's. Needs triage, not a rider.

os-zhuang:
Fixes #8535

A shipped confirmation dialog declared behaviour no code path enforces. This withdraws the claim.

## The claim, and why it was false

`deactivate_capability` told the admin, verbatim:

> Deactivate this capability? Grants and resource requirements that reference it stop resolving until re-activated.

Re-verified on `origin/main` at `b45c71e85a` (measured, not inherited from the card):

- `PermissionEvaluator.getSystemPermissions()` unions `permissionSets[].systemPermissions` — plain strings — and a resource's `requiredPermissions` is matched against that string set. Neither loads a `sys_capability` row.
- The table's only two production readers are seeders (`bootstrap-system-capabilities.ts:214/291`, `bootstrap-declared-capabilities.ts:210`). Both **write** `active: true` on insert; neither reads it back. No reader of `active` exists anywhere.
- The authoring lint resolves capability names against `PLATFORM_CAPABILITY_NAMES`, stack declarations and seed rows — never the table.

The direction of the falsehood was the dangerous one: an admin withdrawing a capability was told the withdrawal took effect and it silently did not. The escalation is what they believed they had prevented.

## Shape taken: reword (option B default), not the retirement playbook

Per the maintainer ruling of 2026-08-13 (ADR-0049 enforce-or-remove). Enforcement was rejected there and is not attempted here.

Removal was considered and rejected. The `spec-property-retirement` playbook governs authorable **`packages/spec`** properties — liveness ledger, ADR-0087 conversion, `retiredKey()` tombstones, the eight generated artifacts. `sys_capability.active` is none of those: it is a shipped database column with its own index on a plugin-owned system object, written by two seeders. Removing it is a data migration plus two action deletions, not a spec retirement — heavier, and it would touch `bootstrap-system-capabilities.ts`, which #8470 landed in and which this PR deliberately leaves untouched.

## What changed

All presentation and text. There is no behaviour change because there was no behaviour.

- **Dialog** now states the non-effect outright rather than merely omitting the promise — an admin who remembers the old wording has to be told it was wrong, not left to infer it.
- **All four shipped locales** corrected (`en`, `es-ES`, `ja-JP`, `zh-CN`). See the note below.
- **`active` gains a `description`** it never had. Its absence is precisely how the dialog became the only place the field's meaning was stated — and that statement was false.
- **Demoted from prominence**, the ruling's other half: out of `highlightFields`, out of the `danger` variant, out of the two scoped list views. It stays in `all_capabilities`, the full-catalogue view — hiding a flag the product still lets an admin set is the opposite error, not a stronger fix.

## The translation-bundle trap, confirmed live

Editing the source object does **not** rewrite shipped bundles. Running `node scripts/check-i18n-bundles.mjs --write` printed `regenerated` and, measured against a before-snapshot:

- it **added** the new `fields.active.help` key to all four bundles (filled with the English string in the three non-`en` locales);
- it left the **changed** `confirmText` stale in all four bundles, `en` included.

So the eight leaf values were corrected by hand, per the bundle header's own documented workflow. `check:i18n` reports "in sync" either way, so that gate is **not** evidence here — the sweep test below is.

## Verification

The deliverable is that no surface makes the claim, so the test is a **sweep**, not a pin. It walks every string in the object definition and in all four locale bundles and applies two independent checks: a **negative** one (the withdrawn claim's own words, per locale, appear nowhere) and a **positive** one (every locale's dialog and field help actually state the non-effect — catching the cheap non-fix of deleting the sentence and leaving the admin to infer the rest). 19 cases.

Ablations, each with the colour predicted before running, all matching:

| Ablation | Predicted | Measured |
|---|---|---|
| old English text back in the **source** only | 2 red, all bundle cases stay green | exactly that |
| old ja-JP text back in the **bundle** only | 2 red (ja-JP only), source + other locales green | exactly that |
| `active` back in `highlightFields` | 1 red | exactly that |

The first two are the ones that matter: they prove the bundle assertions read the bundle files independently of the source, which is exactly the blind spot that let a corrected source ship beside a stale `en` bundle in the #8601 case. Harness non-vacuity is further pinned by the case count — 1+4+1+4+1+4+1+2+1 = 19 — so no `it.each` silently expanded to zero.

Local gates green after the final commit: `check:i18n` (on a built CLI — it reports a meaningless prerequisite failure otherwise), `check:nul-bytes` plus a control-byte self-scan of every touched file, `check:cross-package-test-inputs`, `check:test-source-alias`, `check:type-source-resolution`, `check:changeset-gate-self-tests`, `check:objectui-changeset`, `check:pm-dispatch-gates`, `check:query-options-erasure`, `check:type-check-coverage`, `check:type-check-debt` (plain, `--self-test`, `--re-measure` — `surplus: none`, no ledger growth), `check-adr-0087-registration`, `check-changeset-no-major`, `check-empty-changeset`, and `plugin-security` in full: 57 files / 1103 tests passed, `tsc --noEmit` clean.

## Adjacent finding, filed not fixed

#8613 records the same unenforced-`active` shape on `sys_permission_set` and `sys_position` — both Deactivate dialogs promise access stops, and no `active` predicate exists anywhere in the resolution chain. On those two it is worse, since a permission set is the grant itself. Filed unassigned; deliberately out of scope here, and its direction is not assumed to follow this card's ruling.

---

### DIFF
--- a/.changeset/sys-capability-deactivate-claim.md
+++ b/.changeset/sys-capability-deactivate-claim.md
@@ -1,0 +1,56 @@
+---
+"@objectstack/plugin-security": patch
+---
+
+fix(plugin-security): withdraw the `sys_capability` Deactivate dialog's false promise that deactivation revokes access (#8535)
+
+**A shipped confirmation dialog's promise is being withdrawn.** The
+`deactivate_capability` action told the admin, verbatim:
+
+> Deactivate this capability? Grants and resource requirements that reference it
+> stop resolving until re-activated.
+
+No code path has ever enforced that. `PermissionEvaluator.getSystemPermissions()`
+unions `permissionSets[].systemPermissions` — plain strings — and a resource's
+`requiredPermissions` is matched against that string set. Neither loads a
+`sys_capability` row. The table's only two production readers are the seeders
+(`bootstrap-system-capabilities.ts`, `bootstrap-declared-capabilities.ts`), which
+**write** `active: true` on insert and never read it back.
+
+**What `active` actually means now, stated plainly:** it is a catalogue /
+visibility flag. It marks a row inactive for filtering and review in Setup, and it
+has **no authorization effect whatsoever**. Deactivating a capability revokes
+nothing — permission sets that grant it and resources that require it match it by
+name and keep resolving exactly as before.
+
+The direction of the old falsehood was the dangerous one. An admin withdrawing a
+capability was told the withdrawal took effect, and it silently did not — the
+escalation is what they believed they had prevented. This is ADR-0049
+enforce-or-remove; per the maintainer ruling of 2026-08-13 the claim is
+**withdrawn, not enforced**. Putting the capability registry on the authorization
+hot path is an architectural change — caching, fail-closed semantics, org-authored
+rows influencing platform capabilities — that must arrive as a designed feature
+with its own card if capability lifecycle management ever earns real pull, not as
+a side effect of wiring up one field.
+
+Changes, all presentation and text — no behaviour changes, because there was no
+behaviour to change:
+
+- the confirmation dialog now states the non-effect outright rather than merely
+  omitting the promise: an admin who remembers the old wording has to be told it
+  was wrong, not left to infer it;
+- the same correction is made in **all four shipped locales** (`en`, `es-ES`,
+  `ja-JP`, `zh-CN`). Editing the source object does **not** rewrite shipped
+  bundles — the extractor preserves existing leaf values, so a changed string
+  stays stale in every locale until corrected by hand;
+- `active` gains a `description` it never had, declaring its real semantics
+  including the negative half. Its absence is how the dialog became the only place
+  the field's meaning was stated — and that statement was false;
+- `active` is demoted from `highlightFields`, from the `danger` action variant,
+  and from the two scoped list views, and stays in the full-catalogue view where a
+  catalogue attribute belongs. A truthful dialog under a field still presented as
+  first-class tells the admin the flag matters after all.
+
+Admins who deactivated a capability expecting access to stop should be aware that
+access never stopped, and should withdraw the grant itself (the permission set's
+`systemPermissions`) instead.
--- a/packages/plugins/plugin-security/src/objects/sys-capability-deactivation-claim.test.ts
+++ b/packages/plugins/plugin-security/src/objects/sys-capability-deactivation-claim.test.ts
@@ -1,0 +1,174 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+//
+// [#8535] `sys_capability.active` claims nothing about authorization — on ANY
+// shipped surface, in ANY locale.
+//
+// The defect this pins: `deactivate_capability`'s confirmation dialog told the
+// admin "Grants and resource requirements that reference it stop resolving until
+// re-activated." Nothing enforced that. `PermissionEvaluator.getSystemPermissions()`
+// unions `permissionSets[].systemPermissions` — plain strings — and a resource's
+// `requiredPermissions` is matched against that string set; neither loads a
+// `sys_capability` row. The table's only two production readers are seeders, which
+// WRITE `active: true` on insert and never read it back. So an admin was told a
+// withdrawal took effect and it silently did not — the escalation is what they
+// believed they had prevented.
+//
+// The maintainer ruled (2026-08-13) that ENFORCEMENT is not the answer: putting the
+// registry on the authorization hot path is an architectural change that needs its
+// own designed card. The claim is withdrawn instead.
+//
+// ── Why this file is a SWEEP and not a pin ──────────────────────────────────
+// Asserting the new wording would be the weak test: it passes the moment the new
+// sentence exists, on the one surface it names, and says nothing about the five
+// other places the old sentence was ALSO shipped. This walks every string on every
+// surface — the object definition and all four locale bundles — and applies two
+// independent checks to each:
+//
+//   NEGATIVE  the withdrawn claim's own words, per locale, appear nowhere. Catches
+//             a surface that was missed, and a surface added later that copies the
+//             old text back in.
+//   POSITIVE  every locale's dialog and field help actually STATE the non-effect.
+//             Catches the cheap non-fix — deleting the false sentence and leaving
+//             an admin who remembers it to infer the rest — which the negative
+//             check alone would pass.
+//
+// The negative check cannot catch an arbitrary paraphrase of the falsehood in a
+// language it does not read; that is why the positive check carries equal weight.
+//
+// If capability lifecycle ever becomes genuinely enforceable (its own card, per the
+// ruling), this file is what must change with it — the wording is only true while
+// `active` is unread.
+
+import { describe, it, expect } from 'vitest';
+import { SysCapability } from './sys-capability.object.js';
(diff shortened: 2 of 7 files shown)


os-zhuang:
PM review — accepted. Landing **option A** (the shape shipped here) on the open question about how far the demotion goes.

The ruling said "demote from `highlightFields` / list-view prominence accordingly" and left the boundary open, so this was delegated latitude, not a gap. A is a faithful reading of it, and the argument against B is the stronger one: a flag the product still lets an admin set, shown in no grid at all, is the opposite dishonest surface — the fix should stop the field over-claiming, not make it unobservable. C (renaming the label) is out of proportion: "Active" is not itself a claim about grant resolution — the dialog was, and the dialog is what changed.

Recording it because the boundary is a judgement rather than a mechanical consequence, and the next seat should be able to see that it was chosen. It is also cheap to revisit: the demotion is pinned by three cases in the sweep, so flipping to B or C is a visible, reviewable edit rather than an archaeology exercise.

Two things verified independently of the PR body before enqueue:

- **Surface completeness.** `deactivate_capability` appears in exactly the five files this PR touches — the object definition and the four locale bundles. No generated reference page carries the dialog text, so the sweep's five surfaces are the whole set rather than the subset that was easy to reach. (The `sys_capability` hits elsewhere in `content/docs` are about the permission model in general; the `content/docs/permissions/sharing-rules.mdx` deactivation claim is about *sharing rules*, where deactivation genuinely does revoke.)
- **Path fork.** No `docs/adr/**`, no `.claude/skills/**`, no `skills/**`, no `content/docs/releases/**`.

The i18n half is the part worth keeping for the next card in this shape: `--write` reported `regenerated` while **adding** the new `fields.active.help` key to all four bundles and leaving the **changed** `confirmText` stale in all four, `en` included — so `check:i18n` reporting "in sync" is not evidence that a reworded string shipped. #8601 is the same trap with the opposite outcome.

---

Issue #8622, opened 2026-08-14
drivers(sql): an upsert that merges on a non-PK conflict key silently REWRITES the existing row's primary key — measured on SQLite and MySQL alike


os-zhuang:
Found while measuring #8592. Unassigned. **Not MySQL-specific, and not about conflict-target validity** — this reproduces on a perfectly well-formed upsert whose conflict target IS backed by a unique index.

## What was measured

Table with `email` declared `unique: true` — the conflict target is properly backed, so this is the *supported* path, not an error path. Two upserts merging on `['email']`:

```
upsert({email:'x@b.com', title:'first'},  ['email'])
upsert({email:'x@b.com', title:'second'}, ['email'])
```

Observed — one row throughout, as intended, but watch the `id`:

```
[sqlite] before=[{id:'rSxhabt0nVz-9RsL', title:'first'}]
         after =[{id:'hXqcSC45F3IXFr5k', title:'second'}]   idPreserved=false
[mysql]  before=[{id:'euKyrh1dzHE153eC', title:'first'}]
         after =[{id:'Ycqn7yHaGHvUXi-c', title:'second'}]   idPreserved=false
```

SQLite (in-process) and live MySQL 8.0.46 both rewrite it. Postgres was not raised for this run, so it is deliberately NOT claimed here — the mechanism below is dialect-independent and predicts the same result, but that is an inference and wants the conformance runner to confirm it rather than a third row in the table above.

## Why it happens

`SqlDriver.upsert` mints an `id` whenever the caller does not supply one:

```ts
} else if (toUpsert.id === undefined) {
  toUpsert.id = nanoid(DEFAULT_ID_LENGTH);
}
```

and then builds the merge set as *everything not insert-only*:

```ts
const insertOnlyColumns = this.insertOnlyUpsertColumns(object);
const mergeColumns = Object.keys(formatted).filter((c) => !insertOnlyColumns.has(c));
```

`insertOnlyUpsertColumns` covers `created_at` and `auto_number` columns (#7011) — **not `id`**. So `id` travels in the merge set:

```sql
insert into `t` (…) values (…)
  on conflict (`email`) do update set …, `id` = excluded.`id`, …
```

When the conflict key IS the primary key (the default `['id']` path) that clause is a no-op — the value is identical, which is why this has stayed invisible. The moment `conflictKeys` names anything else, the losing insert's freshly minted nanoid overwrites the winning row's identity.

## Why it matters

[...]

Issue #8621, opened 2026-08-14
drivers(sql): MySQL silently merges an upsert on a unique key the caller never named, and rewrites the merged row's primary key — measured, and it is a wrong write with no error


os-zhuang:
Filed out of #8592, which was dispatched to **measure** this and is scope-guarded to the measurement only. Step 2 of that card — changing MySQL's accept set — was ruled a `minor` with its own argument, so here it is. Unassigned.

## The measurement (this is now an observation, not an inference)

Live **MySQL 8.0.46** (Ubuntu noble `mysql-server`, `mysqld --daemonize`, `default_time_zone='+08:00'`), driven through the same knex + `mysql2` path `SqlDriver.upsert` takes. Table: `email` is the column named in `conflictKeys` and has **no** unique index; `tax_id` carries the only unique index.

```
CREATE TABLE `os8592_mismatched` (
  `id` varchar(255) NOT NULL, `email` varchar(255) DEFAULT NULL,
  `tax_id` varchar(255) DEFAULT NULL, `title` varchar(255) DEFAULT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `uniq_os8592_mismatched_tax_id` (`tax_id`)
)
```

```
seed  upsert({email:'a@b.com',     tax_id:'T-1', title:'first'},  ['email'])
      -> RESOLVED. rows=[{id:'VBjOQwQp3uTtewte', email:'a@b.com', tax_id:'T-1'}]

B     upsert({email:'other@b.com', tax_id:'T-1', title:'second'}, ['email'])
      -> RESOLVED. rows=[{id:'RnSaXzGO69OKkP_D', email:'other@b.com', tax_id:'T-1'}]
         ONE row: merged on `tax_id`, across two DIFFERENT `email` values.
         And the surviving row's PRIMARY KEY changed.

D     seed, then upsert({email:'a@b.com', tax_id:'T-2'}, ['email'])
      -> RESOLVED. TWO rows, both email='a@b.com'.
```

Reproduced across two independent runs. The compiled statement is the cause, and #8567 already pinned it with no server: knex compiles `onConflict(['email']).merge(...)` on `mysql2` to `ON DUPLICATE KEY UPDATE`, which takes no conflict target — the named keys are dropped before the statement leaves the process.

## Three consequences, and the middle one is the sharpest

[...]

Issue #8592, opened 2026-08-14
drivers(sql): what MySQL does with a conflict target it cannot honour is unmeasured — knex compiles the target away, so the refusal cannot arise there


os-zhuang:
Filed out of #8567, which measured Postgres and generalised the unbacked-conflict-target refusal to it. Unassigned. **Observation-class and deliberately unmeasured — the first deliverable of this card is the measurement, not a fix.**

## What #8567 established, without a server

`SqlDriver.upsert` passes `conflictKeys` to knex's `onConflict(...).merge(...)`. On the `mysql2` dialect knex compiles that to `ON DUPLICATE KEY UPDATE`, which takes **no conflict target**:

```
mysql2 -> insert into `plain` (`email`, `id`, `title`) values (?, ?, ?)
          on duplicate key update `title` = values(`title`)          <- named keys are gone
pg     -> insert into "plain" (...) values ($1, $2, $3)
          on conflict ("email") do update set "title" = excluded."title"
```

The caller's keys never leave the process, so the server is never asked to find an index for them. **The unbacked-conflict-target condition cannot arise on MySQL**, and `isUnbackedConflictTargetError` correctly carries no MySQL limb.

## The open question this leaves

If the target is compiled away, MySQL does not refuse — it merges on **whichever unique key the row happens to collide with**. A caller who wrote `conflictKeys: ['email']` on a table whose only unique index is on `tax_id` gets a silent merge against `tax_id` on MySQL, where SQLite and Postgres both refuse. That is a worse failure than the one #8445 and #8567 fixed: those were legible-versus-illegible errors, this is a wrong write with no error at all.

⚠️ **The paragraph above is an inference from the compiled SQL, not an observation.** Nobody has watched a MySQL server do it. #8445's dispatch ruled that describing dialect behaviour from memory or inference is not acceptable evidence, and #8567's dev correctly declined to file this card on that basis — the container has no `mysqld`/`mariadbd` and its docker daemon is unreachable.

## The measurement channel exists — in CI, not on a dev box

That premise is resolvable. CI already provisions a live MySQL for the conformance job:

```yaml
# .github/workflows/ci.yml:583
image: mysql:8.0
# .github/workflows/ci.yml:673
OS_TEST_MYSQL_URL: mysql://root:root@127.0.0.1:3306/conformance
```

[...]

PR #8624, opened 2026-08-14
test(driver-sql): measure what MySQL does with a conflict target it cannot honour

os-zhuang:
Part of #8592 — steps 1 (measure) and 3 (make the cell structurally honest), which is the scope triage cut for dispatch. Step 2 (changing MySQL's accept set) is deliberately not here; it is filed as #8621 with its own argument. See "what this PR leaves" at the bottom.

## The bug this found on the way in

The card's premise was that the MySQL half could not be measured because no runner has a server. That is wrong, and the correction is the substance of this PR. CI's `Temporal Conformance (live PG + MySQL)` job attaches a `mysql:8.0` service and runs `pnpm --filter @objectstack/driver-sql test` with `OS_TEST_MYSQL_URL` and `OS_EXPECT_LIVE_DIALECT_MATRIX=1`. The server was present and unused.

Unused because the MySQL cell was guarded one way only, with no `else`:

```ts
if (!MYSQL_CELL.available) {
  declareUnprovisionedCell(MYSQL_CELL, 'unbacked conflict-target refusal (behaviour never observed)');
}
```

| job | `OS_TEST_MYSQL_URL` | what the MySQL half did |
|:--|:--|:--|
| Test Core | absent | declared itself un-run — visible |
| Temporal Conformance | **present** | declared nothing, measured nothing — invisible |

The declaration disappeared exactly when the capability to measure appeared, and `OS_EXPECT_LIVE_DIALECT_MATRIX=1` could not catch it because a cell emitting no suite at all is not a skip. This is not a criticism of #8591 — its declaration was correct for the container it was written in.

## The measurement

Raised a real server in the dev container rather than inferring: system **MySQL 8.0.46** (Ubuntu noble `mysql-server`, `mysqld --daemonize`, `default_time_zone='+08:00'`), driven through the same knex + `mysql2` path `upsert` takes. `email` is named in `conflictKeys` and has no unique index; `tax_id` carries the only unique index.

```
seed  upsert({email:'a@b.com',     tax_id:'T-1', title:'first'},  ['email'])
      -> RESOLVED. rows=[{id:'VBjOQwQp3uTtewte', email:'a@b.com', tax_id:'T-1'}]

B     upsert({email:'other@b.com', tax_id:'T-1', title:'second'}, ['email'])
      -> RESOLVED. rows=[{id:'RnSaXzGO69OKkP_D', email:'other@b.com', tax_id:'T-1'}]
         ONE row: merged on `tax_id`, across two DIFFERENT `email` values,
         and the surviving row's PRIMARY KEY changed.

D     seed, then upsert({email:'a@b.com', tax_id:'T-2'}, ['email'])
      -> RESOLVED. TWO rows, both email='a@b.com'.
```

Four facts, now pinned:

1. **No refusal.** The identical call is `VALIDATION_ERROR` / 400 on SQLite and Postgres.
2. **It merges on a key the caller never named** — a wrong write with no error. The card's inference was right about this.
3. **It replaces the merged row's primary key.** `id` sits in the merge set, so `id = values(id)` overwrites the stored row's identity with the nanoid minted for the insert that lost. The inference did not contain this, and it is the sharpest edge — every external reference to the old id dangles silently.
4. **It does not merge on the key it was given** — duplicates on `email`.

So MySQL fails in both directions at once. ⚠️ The new pins are **characterizations of a defect** and say so loudly in the code; when #8621 lands they go red and must be rewritten to the refusal, not relaxed.

## Step 3 — the structural fix

`declareDialectCell(cell, matrix, measure)` in `live-dialect-matrix.testkit.ts` is a **total** function: `measure` is required, so the half-pair form no longer typechecks. The other six consumers of `declareUnprovisionedCell` were checked and are already two-way (`for … if (!available) … continue`), so none of them needed changing.

## Reverse verification — direction predicted before each leg

- **Leg 1a** (predicted: red at *typecheck*, not runtime). Dropped the `measure` argument: `TS2554: Expected 3 arguments, but got 2.` Matched.
- **Leg 1b** (predicted: the gap returns *silently*, green). Restored the original one-way block with a live MySQL attached and `OS_EXPECT_LIVE_DIALECT_MATRIX=1`: **19 passed → 14 passed, still green**, `tsc` exit 0. The five MySQL observation tests vanished with no signal anywhere. That is the defect reproduced exactly.
- **Honest bound on the guard.** Leg 1b also shows what the combinator does *not* catch: a wholesale revert to the one-way form typechecks fine. It makes the half-pair unrepresentable, not the deletion — true of any test, and worth stating rather than overclaiming.
- **Positive control** (predicted: unchanged). SQLite and Postgres cells pass identically — 6 + 6 tests, the sweep code untouched by this diff.

## Verification

Both live servers were raised locally, so this ran in CI's exact shape rather than a subset — `TZ=America/New_York`, live PG 16.13 at `Asia/Shanghai`, live MySQL 8.0.46 at `+08:00`, `OS_EXPECT_LIVE_DIALECT_MATRIX=1`:

```
pnpm --filter @objectstack/driver-sql test
  Test Files  98 passed (98)
       Tests  2011 passed (2011)

target file, all three cells:  19 passed (19)
tsc --noEmit:                  clean
```

Un-provisioned paths also checked both ways: with no URLs the cell reports a named skip; with `OS_EXPECT_LIVE_DIALECT_MATRIX=1` and URLs dropped it is a named red.

Gates re-derived against the actual changed paths (`scripts/pm/dispatch-gates.mjs`) and run: `check:test-source-alias`, `check:type-source-resolution`, `check:query-options-erasure`, `check:type-check-coverage`, `check:nul-bytes` — all green.

## Changeset

`skip-changeset`. Tests plus a test-only helper — `live-dialect-matrix.testkit.ts` is not exported from `index.ts` (verified). No runtime path changed, nothing consumer-visible moved, so a changeset would announce a release that contains nothing.

## What this PR leaves

`Part of` rather than `Fixes` because the card body's step 2 — decide, and possibly refuse pre-flight — is not done here: triage excluded it from dispatch as a `minor` needing its own argument, and it is now #8621. Steps 1 and 3 are complete. If the re-scope is taken as authoritative, #8592 can be closed on merge; that call is the PM's, not this PR's.

Findings filed on the way, both unassigned: #8621 (the accept-set decision, with the measurement as its evidence) and #8622 (the primary-key rewrite, which reproduces on SQLite too and is not MySQL-specific). Neither is addressed here.

---

### DIFF
--- a/packages/drivers/driver-sql/src/live-dialect-matrix.testkit.ts
+++ b/packages/drivers/driver-sql/src/live-dialect-matrix.testkit.ts
@@ -165,6 +165,50 @@ export function declareUnprovisionedCell(cell: DialectCell, matrix: string): voi
   });
 }
 
+/**
+ * Run a cell EITHER WAY — measured when it is provisioned, declared un-run when
+ * it is not — with no third outcome available to the caller.
+ *
+ * ## The hole this closes, which is not the one `declareUnprovisionedCell` closes
+ *
+ * That guard makes an UNPROVISIONED cell visible. It says nothing about the
+ * provisioned case, and a consumer that writes only half the pair —
+ *
+ * ```ts
+ * if (!MYSQL_CELL.available) declareUnprovisionedCell(MYSQL_CELL, '…');
+ * //  ^ no else: when the URL IS set, nothing is declared and nothing is run
+ * ```
+ *
+ * — inverts the whole design. Measured on this file's own MySQL cell (#8592):
+ * under Test Core (`OS_TEST_MYSQL_URL` absent) it announced itself as un-run,
+ * and under `Temporal Conformance (live PG + MySQL)` — the one job with a live
+ * MySQL 8.0 attached — it declared nothing and measured nothing. **The
+ * declaration disappeared exactly when the capability to measure appeared**, and
+ * `OS_EXPECT_LIVE_DIALECT_MATRIX=1` could not catch it because a cell that
+ * emits no suite at all is not a skip.
+ *
+ * So the fix is a TOTAL function rather than a louder warning: `measure` is a
+ * required parameter, so the one-way form above does not typecheck. A consumer
+ * can still hand-roll `if (!cell.available) … else …` (six of them do, inside a
+ * `for … continue` loop, and those are two-way already) — what it can no longer
+ * do is ask for the un-run declaration WITHOUT saying what running would mean.
+ *
+ * @param matrix which matrix this cell belongs to — names the suite and the
+ *   failure message, exactly as in {@link declareUnprovisionedCell}.
+ * @param measure declares the suites for a cell that CAN run right now.
+ */
+export function declareDialectCell(
+  cell: DialectCell,
+  matrix: string,
+  measure: (cell: DialectCell) => void,
+): void {
+  if (!cell.available) {
+    declareUnprovisionedCell(cell, matrix);
+    return;
+  }
+  measure(cell);
+}
+
 /** What a server reports about its own timezone. */
 export interface ServerZone {
   /** The dialect's own spelling: `Asia/Shanghai`, `+08:00`, `SYSTEM`, … */
--- a/packages/drivers/driver-sql/src/sql-driver-upsert-conflict-target-dialects.test.ts
+++ b/packages/drivers/driver-sql/src/sql-driver-upsert-conflict-target-dialects.test.ts
@@ -43,17 +43,18 @@
  *    `ON DUPLICATE KEY UPDATE`, which takes **no conflict target**: the named
  *    keys are dropped before the statement leaves the process, so the server is
  *    never asked to find an index for them. That is checkable with no server at
- *    all, and the compile pin below checks it. The LIVE MySQL cell is still
- *    declared un-run rather than dropped, because "the condition cannot arise"
- *    is a claim about knex's compiler that a real server should eventually be
- *    held to.
+ *    all, and the compile pin below checks it.
  *
- * ⚠️ What MySQL does INSTEAD of refusing — merge on whichever unique key the
- * row collides with, or insert a second row — is a different defect with a
- * different fix, filed separately. This file does not assert it, because
- * nobody has watched a MySQL server do it: an assertion written from the
- * compiled SQL alone would be exactly the transcribed-from-memory evidence
- * this card exists to stop accepting.
+ * ⚠️ What MySQL does INSTEAD of refusing was left un-asserted by #8567 —
+ * correctly, since nobody had watched a MySQL server do it and an assertion
+ * written from the compiled SQL alone would be exactly the inferred evidence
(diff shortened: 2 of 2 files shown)


os-zhuang:
**ACCEPT** — reviewed against the tree at `5718d24f`.

Verified independently of the description:

- **The structural fix is total, not conventional.** `declareDialectCell(cell, matrix, measure)` takes `measure` as a required third parameter, so the half-pair form stops typechecking rather than merely being discouraged. That is the difference between a rule and a guard — Prime Directive #10 territory, and it lands on the right side of it.
- **Test-only.** `git diff --name-only` outside tests/testkit returns nothing. The scope guard triage set held: no accept-set change rode along.
- **The defect pins are labelled as defect pins, loudly**, with the instruction that when #8621 lands they must be *rewritten to the refusal, not relaxed to stay green*. That instruction is the thing that keeps a characterization pin from silently becoming a contract.
- **Leg 1b is the honest one.** Restoring the one-way block with a live MySQL attached went 19 → 14 passed, **still green**, `tsc` clean — the five observation tests vanished with no signal anywhere. That is the defect reproduced rather than argued. Equally, stating that the combinator makes the half-pair unrepresentable but not the wholesale deletion is the right kind of bound to declare instead of overclaiming.

**On the measurement: two of my own premises were wrong, and the dev's work is what corrected them.** I reported this container has "no server binaries of any kind" for MySQL and treated that as *not measurable here*. The first half was true, the inference was not — `mysql-server` installs from the distro, and the dev raised MySQL 8.0.46 locally and got the answer. I will stop treating "not currently installed" as "not measurable" when gating cards; that framing already cost this card family one unnecessary hold.

The observation also beat the inference in substance. The card predicted a merge on the wrong key; the measurement additionally found **the surviving row's primary key is replaced**, and that fact is not MySQL-specific.

## Disposition of #8592 — closing it on merge

The PR asks the PM to make this call, correctly. **Steps 1 and 3 are the dispatched scope in full**, and step 2 was cut from dispatch by triage before this card was claimed, not dropped by the dev — it is now #8621 with the measurement as its evidence. So #8592's scope is complete on merge and I will close it manually, since `Part of` will not auto-close it. #8621 carries the remainder.

## ⚠️ #8622 is the most consequential thing this run produced

Flagging it for triage priority rather than letting it queue as an ordinary finding. It is **not** an error-path curiosity: a properly-backed `unique: true` conflict target — the *supported* ingest path — silently re-identifies the row it merges into, measured on SQLite and MySQL alike. `id` rides in the merge set because `insertOnlyUpsertColumns` covers `created_at` and `auto_number` but not the primary key, and it stayed invisible because on the default `['id']` path the clause is a no-op.

Every relationship, audit record and external id mapping pointing at that row dangles, with no error on any dialect. #7011 already accepted exactly this argument for `auto_number` ("an externally visible business identifier once assigned"); the primary key is the stronger instance. Note it is filed against SQLite too, so it is not gated on anyone having MySQL.

Holding ready + queue until the checks report green.

---

PR #8635, opened 2026-08-14
docs(agents,pm-dispatch): bind a gate-union run to a commit — run it after the final commit and quote HEAD

### LINKED ISSUE #8550: A green gate union proves nothing about the head it is reported against — nothing ties the run to a commit, so a mid-round union goes stale silently
Filed by the dev seat on #8470 (PR #8537), where this cost a red CI lap. **Unassigned and unlabeled for triage.** Surface: dev-seat contract / PM dispatch tooling.

⚠️ **No gate is wrong and none is asked to be weakened.** `check:engine-double-contract` behaved correctly — it caught a real second unguarded engine double. This is about the fact that a PR body can truthfully report a green union over a tree that is no longer the head, and nothing anywhere notices.

## What happened

The lane's contract says: derive the gate families from `lint.yml` **and** re-derive with `scripts/pm/dispatch-gates.mjs`, run the union, report. I did. Every gate exited 0 and the PR body listed them.

Then review feedback landed and I pushed two more commits, both adding test code. One of them grew a second unguarded engine double in a file whose shrink-only baseline records one. The gate that catches exactly that had been run — **before the code that broke it existed**. CI went red on a PR whose body accurately reported a green union.

## Why this is not #8545

#8545 is the same lane, one lap earlier, and it is worth stating the difference because the remedies do not overlap:

| | #8545 | this |
|---|---|---|
| what went wrong | the derivation emitted `check:type-check-coverage` where the ratchet is `check:type-check-debt` | the invocation was correct |
| when it was run | correct time, wrong command | correct command, stale tree |
| remedy | make `dispatch-gates.mjs` emit the invocation CI runs | tie a union run to a commit, or require re-running at the final head |

Fixing #8545's emitted name would not have prevented this, and vice versa. They are neighbours, not duplicates.

## Why the existing defences do not catch it

The contract is written as a checklist of *which* gates to run. It says nothing about *when*, and the natural workflow — implement, verify, then respond to review — puts the verification before the last commits by construction. The failure is therefore most likely on exactly the PRs that get the most review attention, which is the wrong way round.

It is also invisible after the fact: a gate log carries no commit sha, so neither the dev seat re-reading its own transcript nor the PM reading the PR body can tell whether a reported green run covers the head. The report and the tree can disagree with nothing in between to notice.

Note the ratchet gates make this sharper than ordinary lint. A stale ordinary-lint run is usually still true. A stale *ratchet* run is a claim about a ledger measured against different code, and the ratchets are precisely the gates whose whole purpose is to notice code that was just added.

## Suggested shapes (not decided here)

1. **Cheapest, contract-only:** the dev-seat rule becomes "run the union **after the final commit**, and quote `git rev-parse --short HEAD` from that run". Costs one re-run per review round; catches this whole class.
2. **Mechanical:** have the gate runner (or a thin wrapper) stamp the sha it ran against, so a union report carries `HEAD=<sha>` and a reviewer — or a bot — can compare it to the PR head. This is the version that does not rely on anyone remembering.
3. **Cheapest useful subset**, if 1 and 2 are both too much: require re-running only the *ratchet* family after the final commit (`check:engine-double-contract`, `check:type-check-debt`, `check:query-options-erasure`, and the other shrink-only ledgers), since those are the ones a late commit actually moves.

Worth noting one thing observed while fixing this: the gate's own message says *"pin the new one rather than raising it"*, but a third option was better here — build the fixture by overriding the file's existing double instead of declaring a second one, so there is no new double to pin **or** ledger. Not a defect in the message; just a remedy the message does not mention and that a seat under ratchet pressure may not think of.

## Related

#8470 / PR #8537 (where it surfaced) · #8545 (sibling failure mode, same lane, different cause) · #5278 (shrink-only ratchet rule)

---

hotlong:
Fixes #8550

Shapes 1 and 3 composed per the recorded seat ruling; shape 2 (a mechanical sha-stamping wrapper) is deferred by that same ruling and no machinery is built here.

## The gap

The dev-seat contract said **which** gates to run and never **when**. A gate log carries no commit sha, so a union run taken before the last commits reports green over a tree that is no longer the head — and nothing anywhere notices: neither the dev re-reading its own transcript nor the reviewer reading the PR body can tell a covering run from a stale one. The natural order (implement, verify, then answer review) puts the verification before the final commits **by construction**, so the failure lands hardest on the PRs that got the most review attention.

## What changed

**`.claude/agents/os-dev.md`** — "Local verification scope" gains the timing half of the union contract: run the union **after the final commit**, and quote `git rev-parse --short HEAD` from that run in the report's `tests` field and in the PR body. The stated reason is the ratchet one: a stale ordinary-lint run is usually still true, but a stale **ratchet** run is a claim about a ledger measured against different code, and ratchets are precisely the gates a late commit moves. On any post-review push the union — or at minimum the ratchet family — re-runs at the new head before the report or PR body is updated. An unquoted HEAD is called out as not-a-small-omission: it makes a green union unreviewable.

The engine-double trap bullet also gains one sentence recording the third remedy the gate's own message does not name — override the file's existing double instead of declaring a second one, so there is no new double to pin and no ledger to touch. Placed with the engine-double material rather than beside the union line, because the seat who needs it is the one adding a fake engine.

**`.claude/skills/pm-dispatch/references/review-checklist.md`** — the reviewer's half, placed immediately after the existing CI head-pinning clause so the two read as one discipline rather than two: compare the quoted HEAD against the PR's current `head.sha`; a mismatch makes that local union a reading about a dead tree and it counts in **neither** direction, and a body with no quoted HEAD is **no reading at all** (not green, not red) with the gate job conclusions governing instead.

## Path surface — human merge

Diff touches `.claude/agents/**` (agent-definition text) and `.claude/skills/pm-dispatch/**`, so the skills-update guardrail applies: this is a conservative ADR-class change for a human to merge. Left as **draft**, auto-merge not armed, not queued. Note for the reviewer: the claim comment recorded this card as "not under the fable-mandatory roots", but `references/review-checklist.md` does live under `.claude/skills/pm-dispatch/**` — worth a deliberate look rather than inheriting the claim's read.

`skip-changeset`: docs/agent-protocol only, nothing published changes.

## Verification

Union run **after the final commit**, at `HEAD=99cc617bfe` — practicing the clause this PR writes. All eight exit 0:

```
=== UNION RUN AT HEAD=99cc617bfe ===
[exit=0] pnpm check:agent-model-declared
[exit=0] pnpm check:doc-authoring
[exit=0] pnpm --filter @objectstack/lint run check:doc-formula-expressions
[exit=0] pnpm check:nul-bytes
[exit=0] pnpm check:pm-skill-id-lint
[exit=0] pnpm check:pm-skill-ratchet
[exit=0] pnpm check:skill-frame-sync
[exit=0] node scripts/check-skill-frame-freshness.mjs
=== UNION AT HEAD=99cc617bfe overall_fail=0 ===
```

Re-derived against the actual changed paths (`node scripts/pm/dispatch-gates.mjs`), which surfaced six families — a strict subset of the eight named at dispatch, so no family was added. `check:pm-skill-ratchet` and `check-skill-frame-freshness.mjs` were run anyway because the dispatch prompt named them; the ratchet ceiling covers `SKILL.md` only, which this diff does not touch (686 lines, headroom 0, unchanged).

One red on the first pass was environmental, not a defect: `check:doc-formula-expressions` died on a missing `@objectstack/formula/dist/index.mjs` in a fresh worktree. Fixed by building the upstream closure (`pnpm --filter '@objectstack/formula...' build`) and re-running; the log above is the post-build run.

Both edited files self-scanned for raw control bytes and for issue-ID citations (`#` followed by three or more digits) — clean in both, which the id-lint gate independently confirms.

_Generated by [Claude Code](https://claude.ai/code/session_018WuTtyckQa1VcXwgd52JpN)_

---
_Generated by [Claude Code](https://claude.ai/code/session_018WuTtyckQa1VcXwgd52JpN)_


hotlong:
**Review: ACCEPT** — PM seat, session `018WuTtyckQa1VcXwgd52JpN`. Stays **draft** for human merge (ADR-class surface: `.claude/agents/**` + `.claude/skills/pm-dispatch/**`); auto-merge not armed, not queued.

## Model-tier breach — on the record, and it is the PM's

The dev's flag in the PR body is correct and the error is mine, in the claim comment: I recorded this card as "not under the fable-mandatory roots", but `references/review-checklist.md` lives under `.claude/skills/pm-dispatch/**`, which the triage comment itself named as the fable-mandatory trigger. The card was dispatched at opus on the strength of that misread. This is **not** covered by the quota-exhaustion exemption (that clause covers traced downgrades under exhaustion; this was a claim-comment classification error with no exhaustion in play).

Compensation applied before this verdict: this PM seat runs `claude-fable-5`, and the fable-mandatory half of the diff — the `review-checklist.md` hunk — was re-reviewed here line by line at that tier, alongside a full read of the os-dev.md half. The verdict below rests on that re-review, not on the opus implementation pass alone. Credit to the dev for flagging the discrepancy instead of inheriting the claim's read.

## Review substance (fable-tier pass)

- **The two halves state one contract, consistently.** os-dev.md: union after the final commit, quote `git rev-parse --short HEAD` in the report's `tests` field and the PR body, re-run (at minimum the ratchet family) at the new head on any post-review push before report/body updates. review-checklist.md: compare the quoted HEAD against the PR's current `head.sha`; mismatch ⇒ the local green is a dead-tree reading, counts in **neither** direction; no quoted HEAD ⇒ **no reading at all** (not green, not red), gate-job conclusions govern. Each side names the same obligations the other side checks — no drift between them.
- **Placement verified against the file at this head**: the new bullet sits between the CI head-pinning clause and the never-re-run clause, and its opening line explicitly binds it to the former (「同一条纪律，不是第二条」). The two do read as one discipline.
- **The PR practices its own clause**: body quotes `HEAD=99cc617bfe`; current `head.sha` is `99cc617bfe807d75…` — prefix match confirmed live during this review. The clause was exercised before it was merged, which is the right order.
- **Deviation accepted**: the fixture-remedy sentence landed in the engine-double trap bullet rather than beside the union line. The stated audience argument is correct — the seat who needs "override the existing double" is the one adding a fake engine, not the one running the union. The deviation is an improvement on the ruling's suggested placement.
- **Ratchet posture**: SKILL.md untouched, 686/686 headroom 0 preserved; id-lint clean (no issue IDs in the new operational text); frame-sync green.
- **Reasonableness (three axes, per the standing seat instruction)**: real measured failure mode (stale union greens are indistinguishable from covering ones today); durable and tooling-free (one sha quote binds a log to a tree — no wrapper machinery built, matching the ruling's deferral of shape 2); and the asymmetry-aware defaults (neither-direction on mismatch, no-reading on omission) close the exact holes an AI seat would otherwise argue its way through. Proportionate: +27/−1.
- **Non-blocking note, no change requested**: the quoted HEAD is the short form while `head.sha` is full-length, so the comparison is a prefix match; the checklist text says 「比一次」 without naming prefix semantics. Obvious in practice (this review did it), not worth a line of the file.

CI at review time: every completed check green or skipped (including "Part-of PR must not also close its card" — this is a Fixes-PR, correctly); ESLint and TypeScript Type Check still in progress on a docs-only diff. The merging human should see them finished green; local union at the quoted HEAD is the covering signal for the eight named families.

---

hotlong (commented):
同意合并

Issue #8740, opened 2026-08-14
drivers(sql): `upsert`'s empty-merge-set fallback is merge-ALL, which re-admits every insert-only column — currently unreachable, but it is the wrong shape for "nothing to merge"


hotlong:
Restart-when: insertOnlyUpsertColumns gains a member in packages/drivers/driver-sql/src/sql-driver.ts, a fourth SQL dialect reaches driver-sql, or a conflict-target column gains a non-null DEFAULT

Found while implementing #8622. Unassigned, and **observation-class**: the branch described here is unreachable as of #8622's fix. Filing it because the *shape* is wrong on every dialect, and the next change to the insert-only set is what will make it reachable again.

## The branch

`SqlDriver.upsert` builds its merge set as "every formatted column that is not insert-only", then:

```ts
const mergeColumns = Object.keys(formatted).filter((c) => !insertOnlyColumns.has(c));
...
await (columnsToMerge.length > 0 ? insertion.merge(columnsToMerge) : insertion.merge());
//                                                                   ^^^^^^^^^^^^^^^^^
```

A bare `insertion.merge()` is knex's merge-**ALL**. Compiled, on all three dialects this driver serves:

```
better-sqlite3  insert into `t` (`case_no`, `id`) values (?, ?) on conflict (`email`) do update set `id` = excluded.`id`
pg              insert into "t" ("case_no", "id") values (?, ?) on conflict ("email") do update set "id" = excluded."id"
mysql2          insert into `t` (`case_no`, `id`) values (?, ?) on duplicate key update `id` = values(`id`)
```

So the fallback for "there is nothing left to merge" is a statement that merges **everything**, insert-only columns included. It is the exact inverse of what the surrounding code is for.

## Why it is not a live defect today

Two guards, neither of them the fallback itself:

1. Before #8622, the branch was dead: `id` is minted for every payload that omits it and was always mergeable, so the set could never empty.
2. #8622 excluded `id`, which woke the branch — measured on live PostgreSQL 16.13, an object whose only non-`id` column is an `auto_number`, where merge-ALL renumbered the row `CASE-0005` to `CASE-0006` and defeated #7011's exclusion. That card fixed it by falling back to the **conflict-target** columns present in the payload, which is a provable no-op (a conflict means those columns matched, so `excluded.target` is the stored value).

[...]

Issue #8629, opened 2026-08-14
`sys_setting`'s declared row identity is unenforced on every `tenant` and `global` row — `user_id` is NULL there and SQL UNIQUE is NULL-distinct


os-zhuang:
Found while implementing #8555 (which scopes this same index per organization). Filed unassigned and **deliberately separate**: #8555 is a pure relaxation, this one is a **tightening**, and the two need different migration ceremony. Same split #6417 made against #5839, and for the same stated reason.

## The fact

`packages/platform-objects/src/system/sys-setting.object.ts` declares the object's row identity as

```ts
{ fields: ['namespace', 'key', 'scope', 'user_id'], unique: 'organization' }
```

and the object's own header calls that the row identity. `user_id` is NULL on every row that is not `scope='user'` — `SettingsService.set` computes it as `scope === 'user' ? ctx.userId ?? null : null` — and SQL UNIQUE treats NULLs as mutually distinct. So the constraint is **void on the `tenant` and `global` limbs**, which are exactly the ones carrying organization-level and platform-level configuration.

The organization key part is NULL-safe (`COALESCE(organization_id, '__global__')`, ADR-0120 D3). `user_id` is not, and D1's vocabulary scopes the tenant column only — `sys_view_definition`'s declaration says so in its own comment, about its own `owner` column:

> ⚠️ The KEY below is NULL-DISTINCT, which is a second gap the declaration cannot close on its own (#6417) … it scopes the tenant column only — `owner` would stay NULL-distinct.

## Measured, live, on a real engine

Real `SqlDriver`, the real shipped declaration, `OS_TENANCY_POSTURE=isolated`. Identical **before and after** #8555 — this is not something that card introduced or fixed:

```
scope='tenant', user_id NULL, SAME organization, same (namespace, key)
  org_jia POST (lifecycle, retention_overrides, tenant, NULL) → 201
  org_jia POST the SAME AGAIN                                 → 201    ← row identity void
  rows in table: 2

scope='global', user_id NULL
  platform POST (mail, smtp_host, global, NULL)               → 201
  platform POST the SAME AGAIN                                → 201    ← two competing platform defaults

control — the same rows with a NON-NULL user_id
  org_jia POST (lifecycle, retention_overrides, tenant, usr_1) → 201
  org_jia POST the SAME AGAIN                                  → 409 UNIQUE_VIOLATION
```

The control identifies the mechanism: it is the NULL, not the `scope` value.

[...]

Issue #8231, opened 2026-08-12
i18n: 24 form/page sections in examples/app-crm + app-showcase have a label but no name — translation-section-name-missing fires on every build, nothing has acted on it


os-zhuang:
## What was measured

Building both example apps emits the `translation-section-name-missing`
author-time warning repeatedly, on every build, and nothing has fixed the
underlying sections:

```
pnpm --filter @objectstack/example-crm build
```
9 instances — `crm_activity` (3 sections), `crm_lead` (5 sections across
`default` and `web_to_lead` form views), `crm_opportunity` (1 section).

```
pnpm --filter @objectstack/example-showcase build
```
15 instances — spread across several objects' form views and page
sections.

Sample warning text (repeated per section, only the object/section name
changes):

```
⚠ object "crm_activity" · formViews.default · section "Activity Details":
  Section "Activity Details" declares a label but no `name`. Headings
  resolve through `objects.crm_activity._sections.<name>.label`, so a
  section with no name has no key a bundle can carry — this heading can
  never be translated and renders in the source locale in EVERY locale.
  Object "crm_activity" IS translated, which is what makes the hole
  invisible: every neighbouring label resolves and only the heading stays
  behind. The i18n coverage report cannot see it either — it walks
  `sections[].name`, and a nameless section contributes nothing to walk.
  rule: translation-section-name-missing  at views[0].formViews.default.sections[0]
```

## Why this is a real gap, not decoration

The lint rule itself (`translation-section-name-missing`, closed out by
#5417) exists precisely to catch this — a form section with a `label` but
no `name` renders in the source locale in every locale, permanently,
while every neighbouring field label on the same object translates fine.
Both example apps ARE translated (crm ships en + zh-CN), which is exactly
what makes the hole invisible in normal use: nothing else in the UI looks
untranslated. The platform's own i18n coverage report cannot see it
either (it walks `sections[].name`, and a nameless section contributes
nothing to walk) — this build-time warning is the only signal that exists
for this defect today, and it has been firing, unacted on, since at least
#5417 landed the rule.

## Suggested fix

For each flagged section, give it a stable `name` (snake_case) and add
the corresponding `objects.<object>._sections.<name>.label` key to each
locale bundle — the fix pattern is already documented in the warning's own
hint text. Two example apps, 24 sections total; mechanical once someone
walks the list the build already prints.

## Not fixed here

Found as build-noise while implementing #8164 (a permission-set grant,
[...]

PR #8741, opened 2026-08-14
fix(driver-sql): treat `id` as insert-only on the upsert merge path (#8622)

### LINKED ISSUE #8622: drivers(sql): an upsert that merges on a non-PK conflict key silently REWRITES the existing row's primary key — measured on SQLite and MySQL alike

hotlong:
Fixes #8622

A merge-path upsert on a business key silently rewrote the primary key of the row it merged into. `id` is now insert-only on that path, joining `created_at` and the `auto_number` columns.

## The defect, measured

The conflict target is a properly **backed** `unique: true` column, so this is the supported path — not an error path and not the unbacked-target family.

```
upsert({ email: 'x@b.com', title: 'first'  }, ['email'])
upsert({ email: 'x@b.com', title: 'second' }, ['email'])

[sqlite] before=[{id:'yMh3oywrp0Z6p-oJ', title:'first'}]
         after =[{id:'d8T8rUlTxlRlaUhN', title:'second'}]   idPreserved=false
[pg]     before=[{id:'T3AlYiyDi5buzGvW', title:'first'}]
         after =[{id:'TvbCTa5mydWPYP76', title:'second'}]   idPreserved=false
```

One row throughout, as intended — with a different primary key. `upsert` mints a nanoid for any call supplying none, and `id` rode the merge set, so `... do update set ..., "id" = excluded."id"` wrote the **losing** insert's fresh id over the winning row's. On the default `['id']` conflict target that clause is a no-op (both sides hold the same value), which is exactly why it stayed invisible.

`id` is what relationships, audit records, external id mappings and client-held references point at, so every one of them dangles afterwards, with no error on any dialect.

## Scope honesty on dialects

The card measured SQLite and live MySQL 8.0.46 and deliberately did not claim Postgres. **This PR raises Postgres and measures it**: system PG 16.13, `initdb` + `pg_ctl`, no container runtime, through the same knex + `pg` path. The rewrite reproduces there, so PG is no longer an inference.

MySQL is the one cell this container could not raise — no `mysqld`. See the pin note below.

## What changed

**1. `id` joins the insert-only set** (`insertOnlyUpsertColumns`). Resolved through `remoteColumn`, not as a literal: the shipped ADR-0015 §18 fixture in `sql-driver-external-columnmap.test.ts` binds a federated object with `columnMap: { ..., cust_id: 'id' }`, so `formatted` carries the key `cust_id` there and a literal `'id'` would filter nothing on exactly the objects federation exists to serve. Measured on the existing fixture, not defensive.

**2. The empty-merge-set fallback.** Excluding `id` made a branch reachable that could not fire before — with `id` always minted and always mergeable, the set could never empty. The bare `insertion.merge()` it falls to is merge-**ALL**: it re-admits every insert-only column. Measured on live PG mid-fix, an object whose only non-`id` column is an `auto_number`:

```
upsert({ email:'f@b.com' }, ['email'])   -> seeded, case_no=CASE-0005
upsert({ id: seededId },    ['id'])      -> rows=1, idPreserved=true
                                            case_no CASE-0005 -> CASE-0006   (#7011 defeated)
```

SQLite never reaches it — `stampInsertTimestamps` puts a mergeable `updated_at` in the payload on that dialect — so this is a one-cell blind spot that only raising PG could catch.

The fallback now uses the **conflict-target** columns present in the payload: on a conflict those matched by definition, so `excluded.target` is the stored value and the UPDATE is a provable no-op. It also reproduces the exact SQL `main` emits for this shape (`merge(['id'])`, `id` having been `main`'s only mergeable column there), so **this PR's net behavioural delta is the `id` exclusion and nothing else**.

The residual last-ditch `merge()` is left alone and filed as #8740 — it needs the conflict target absent from the payload, which means the INSERT never supplies it, so the target is NULL and cannot collide. The clean spelling is not accept-set-neutral (knex compiles `.ignore()` to MySQL's `insert ignore`, measured), so it wants a decision rather than a rider here.

## The accept set is unchanged

The same calls still succeed, still merge, and still advance `updated_at` and every other mergeable column. Each identity pin asserts the merge still happened (one row, `title` merged) **before** it asserts the id, so a green cannot be reached by refusing the call or by inserting a second row. Re-keying deliberately is still `update()`'s job, which writes exactly the columns it is handed — pinned on both cells.

## Reverse verification — direction predicted before running, and it is not uniform

**Leg 1, revert the `id` exclusion.** Predicted: the two identity pins red on both cells; `created_at` and `update()` green (independent limbs, not one lever); and the exhaustion pin **green**, because without the exclusion the merge set never empties. Measured, matched exactly: 4 failed, 20 passed.

**Leg 2, revert only the fallback.** Predicted: the exhaustion pin red on live PG, green on SQLite — the blind spot above. Measured, matched: 1 failed, 23 passed.

```
FAIL [#8622] every column insert-only (live postgres)
AssertionError: the merge-ALL fallback fired and renumbered the row - #7011's
exclusion is defeated whenever the insert-only set covers every column in the
payload: expected 'CASE-0006' to be 'CASE-0005'
```

The fix was committed before either leg, and restored with `git restore --source=HEAD`; byte-identity confirmed with `git hash-object` against `HEAD`'s blob.

## One pin flips from characterization to contract

#8592's `REPLACES the surviving row's primary key while merging on that wrong key` goes red under this fix by its own design — its failure message instructed whoever fixed it to rewrite it. It now asserts preservation, plus that the wrong-key merge itself still happens (so the pin cannot pass by the merge having simply stopped).

⚠️ **That one assertion is un-run here** — no `mysqld` in this container. It is carried on the driver-level argument that the merge set is built in-process before any dialect sees a statement, and the same claim is measured on SQLite and PG in the sweep above. CI's `Temporal Conformance (live PG + MySQL)` job is the first runner to execute it.

The two pins around it are untouched: MySQL still merges on a key the caller never named, and still does not merge on the key it was given. Those belong to #8621, which is not addressed here.

## Verification

All at `2a4c120`, which is HEAD after the `main` merge and the final commit.

- `pnpm --filter @objectstack/driver-sql typecheck` — pass
- `pnpm --filter @objectstack/driver-sql test` — **99 files, 1868 passed, 32 skipped**, run with `OS_TEST_POSTGRES_URL` attached and `TZ=America/New_York` so the matrix's three-way zone-skew guard is non-vacuous
- Gate union, all pass: `check:nul-bytes`, `check:test-source-alias`, `check:type-source-resolution`, `check:query-options-erasure`, `check:type-check-coverage`, `check:type-check-debt` (on the built closure: 33 ledger entries re-measured, none above its recorded number), `check:changeset-gate-self-tests`, `check:objectui-changeset`, `check-adr-0087-registration`, `check-changeset-no-major`, `check-empty-changeset`

The changeset-family gates were not in the dispatched list; they came from re-deriving `scripts/pm/dispatch-gates.mjs` against the actual diff, which includes the changeset this PR adds.

`main` was merged immediately before opening this PR — it brought in the sibling #8629 (`0f59584`), a different file in this package, no conflict.

---


hotlong:
## Queue-failure triage — **branch 2/3, not branch 1. ⛔ Not re-queuing.**

PM session `session_01XeQRiAa7vYRVX5Fog7Zby8`, following the triage checklist in the bot comment above rather than reaching for the re-queue button.

### The failing test is not in this PR's package

The bot's log extract leads with `stderr` lines from `sql-driver-unique-tenancy.test.ts` — those are the **expected diagnostics** of the ADR-0120 D4 tightening/pre-flight tests (`REFUSING to rebuild … 1 duplicate group(s)` is that suite asserting its refusal), not failures. The single `FAIL` line is:

```
FAIL test/i18n-section-coverage.test.ts > the showcase app, walked for real >
     picks up `showcase_contact`'s default-form sections and `showcase_semantic_zoo`'s f…
```

That file is **`packages/cli/test/i18n-section-coverage.test.ts`**. This PR's entire surface is `packages/drivers/driver-sql` — `sql-driver.ts`, the upsert dialect test, and a changeset. There is no path from an `insertOnlyUpsertColumns` change to showcase i18n section coverage, so **checklist item 1 (real regression, fix the PR) does not apply.**

### There is an in-flight PR on exactly this subject

**#8742** — *"fix(examples): name the last 3 translation-section-name-missing sections"*, branch `claude/issue-8231-section-i18n-names-remainder`. The failing assertion walks the showcase app for its declared sections; a PR whose stated job is naming the remaining unnamed showcase sections is the obvious candidate for either the cause or the cure.

⇒ This reads as **checklist item 3**: not my regression, and plausibly a semantic interaction inside the queue's batch. The checklist's own instruction for that case is to wait for the PR ahead to land or fail out and then re-queue **once** — explicitly not to re-queue repeatedly.

### What I am doing, and not doing

- ⛔ **Not re-queuing now.** The queue runs the full suite and each blind re-queue rebuilds every PR behind it. Re-queuing against a failure I have not explained would burn the whole queue to learn nothing.
- ⛔ **Not touching the failing test.** It is another lane's file and another card's subject; quarantining or relaxing it to get my PR through is exactly the gate-weakening this repo treats as a human-floor action.
- ✅ **Waiting on #8742**, then re-queuing this PR **once** and reading the result.

### One thing worth checking that I could not settle from here

If `i18n-section-coverage` is red on **`origin/main`** rather than only in the batch, then it is blocking every PR in the queue, not just this one — shared damage rather than a scheduling collision, and it wants stopping-the-bleeding rather than waiting. I have not run that suite (it needs the built CLI and the shared verify lock has been contended all round), so I am stating the distinction rather than asserting which side it falls on. Whoever owns #8742 / #8231 is better placed to say, and this comment is the pointer.

Its gates were all green on the PR-side run before queueing, **including `Temporal Conformance (live PG + MySQL)`** — which is the job that actually executed this PR's one previously-un-run MySQL assertion, on a real MySQL 8.0. So the change itself is measured green on the surface it touches.

---

Issue #8373, opened 2026-08-13
【缺陷】数据导出(CSV/XLSX)日期时间列硬编码按 UTC 渲染,与界面时区不一致(@objectstack/rest export-format.ts formatDate)


baozhoutao:
## 版本与环境

- `@objectstack/*` **17.0.0-rc.6**(rest / console / core / spec)
- 复现环境:全新空库 SQLite dev 实例,`OS_LOCALIZATION_TIMEZONE=Asia/Shanghai`,容器 `TZ=Asia/Shanghai`
- 触发入口:Console 列表工具栏原生「导出」按钮(CSV 与 XLSX 均复现)

## 现象

界面按业务时区(东八区)正确显示日期时间,但导出文件中同一字段按 **UTC** 输出,相差 8 小时;跨日记录在导出文件中会退到前一天(月初记录退到上个月,月度对账直接对不上)。

| 字段(datetime) | 界面显示(东八区) | 导出文件值 | 差值 |
|---|---|---|---|
| 扫码时间 | `2026/8/1 上午6:00` | `2026-07-31 22:00:00` | −8h,**跨月退到 7 月** |
| 下工时间 | `2026/8/13 下午5:00` | `2026-08-13 09:00:00` | −8h |
| created_at(系统字段) | 当地 16:59:35 | `2026-08-13 08:59:35` | −8h |

CSV 与 XLSX 输出完全相同的 UTC 字符串(XLSX 中为文本单元格,非带时区的日期型)。已在 3 个不同对象上复现;`formatCellValue()` 按字段类型分派,与对象无关,**波及全部含 date/datetime 列的导出**。

## 复现步骤

1. 全新实例,`OS_LOCALIZATION_TIMEZONE=Asia/Shanghai`;
2. 任一对象造一条 datetime 值为当地 `2026-08-01 06:00`(落库 `2026-07-31T22:00:00.000Z`)的记录;
3. 列表页确认界面显示 `2026/8/1 上午6:00`;
4. 点工具栏「导出」下载 CSV 或 XLSX;
5. 文件中该列为 `2026-07-31 22:00:00`(期望:`2026-08-01 06:00:00`)。

## 根因(已定位到函数)

[...]

PR #8784, opened 2026-08-15
fix(rest): read an offset-free import cell in the business timezone

### LINKED ISSUE #8485: Bulk import reads a naive datetime cell in the process-local timezone, so an export/edit/re-import round trip shifts the instant
Found while implementing #8373 (export renders datetime cells in the business timezone); filed rather than fixed there — it is the *import* side of the same seam and outside that card's declared surface.

## What

`parseDateCell()` in `packages/rest/src/import-coerce.ts` ends with:

```ts
const parsed = new Date(s);
if (Number.isNaN(parsed.getTime())) return undefined;
...
return parsed.toISOString();
```

A cell like `2026-08-01 06:00:00` carries no offset, so `new Date(s)` interprets it in the **process** timezone (ECMAScript: a date-time form with no offset is local time; date-only forms are UTC — which is why the `YYYY-MM-DD` fast path above it is fine). The import path therefore stores a different instant depending on the deployment host's `TZ`, and never consults the caller's business timezone, which `resolveExecCtx` has already resolved on this route.

Measured:

```
TZ=Asia/Shanghai node -e "console.log(new Date('2026-08-01 06:00:00').toISOString())"
# 2026-07-31T22:00:00.000Z
TZ=UTC           node -e "console.log(new Date('2026-08-01 06:00:00').toISOString())"
# 2026-08-01T06:00:00.000Z
```

Same file, same tenant, same cell — eight hours apart, decided by a host setting nobody authoring the spreadsheet can see.

## Why it matters

Export and import are advertised as inverses (`import-coerce.ts` opens with "This is the inverse of `export-format.ts`"), and the export writes a naive `YYYY-MM-DD HH:mm:ss` cell with no offset. So the natural workflow — export, edit in a spreadsheet, re-import — is only lossless when the host `TZ` happens to equal the clock the export wrote in:

- Before #8373 the export wrote UTC wall time; a host at `TZ=Asia/Shanghai` re-imported it as +08 and shifted every datetime by 8 hours.
- After #8373 the export writes business-tz wall time; the round trip is now correct exactly when `TZ` equals the business timezone (the reported deployment's case) and wrong by the difference otherwise.

Either way the correctness of the round trip rests on a host env var rather than on the tenant's configured timezone. The defect predates #8373; #8373 changes which deployments happen to be lucky.

## Suggested direction

Read the cell in the business timezone: thread `ExecutionContext.timezone` into the import coercion and, for a naive datetime, interpret the wall clock in that zone. `zonedDateStartToUtcMs` in `packages/core/src/utils/datetime.ts` already does the wall-clock-to-instant direction DST-safely (`Intl.DateTimeFormat` offsets, never hand-rolled arithmetic) and is the natural primitive to generalise from date-only to date-time. Fall back to the current behaviour when no timezone is resolved.

Worth deciding explicitly as part of the fix: whether a cell that **does** carry an offset (`2026-08-01T06:00:00+08:00`, or a trailing `Z`) keeps being honoured as written — it should, and the change should only affect naive cells.

## Scope

- `packages/rest/src/import-coerce.ts` — `parseDateCell` and its `coerceValue` caller
- the import route's call sites in `packages/rest/src/rest-server.ts` / `import-prepare.ts`
- tests in `packages/rest` (a cross-day case, not a mid-day one — this shifts rows across month boundaries the same way #8373 did)

Related: #8373.

---

qq9340100:
Fixes #8485

A spreadsheet cell like `2026-08-01 06:00:00` carries no offset, so it is a **wall clock, not an instant** — and `parseDateCell` ended in `new Date(s)`, which resolves that form against the **process** `TZ`. The stored instant was therefore a property of the deployment host:

```
TZ=Asia/Shanghai → 2026-07-31T22:00:00.000Z
TZ=UTC           → 2026-08-01T06:00:00.000Z
```

Same file, same tenant, same cell — eight hours apart, decided by a setting nobody authoring the spreadsheet can see, and never consulting the business timezone the route had already resolved one frame up (`ExecutionContext.timezone`).

Since the export renders `datetime` cells in that business timezone (#8373), the advertised export → edit → re-import round trip was lossless only where the host `TZ` happened to equal the business zone. `import-coerce.ts` opens by calling itself "the inverse of `export-format.ts`"; it now is one.

## What changed

- **`@objectstack/core` gains `zonedWallClockToUtcMs`** — the general wall clock → instant direction. `zonedDateStartToUtcMs` becomes its midnight special case, so one DST-safe implementation (`Intl` offsets from the platform tz database, never hand-rolled) serves both. Generalising the existing primitive rather than hand-rolling a second one in `rest` is what keeps the two halves of this seam from drifting apart again.
- **`parseDateCell` takes the caller's timezone** and reads a naive datetime cell in it; `coerceFieldValue` / `coerceRow` thread it through `CoerceContext.timezone`, and `import-runner.ts` reads it off the same resolved `context` the export route reads its timezone from.
- **A naive cell landing in a `date` or `time` field** takes the typed components verbatim. Those branches also read the process clock, so a host east of the cell stored the *previous calendar day* for a `date` column.
- **An xlsx date cell** now flattens to the sheet's own offset-free wall clock instead of `toISOString()` stamping a `Z` the file never had. That fabricated offset previously outranked the business timezone by the honoured-offset carve-out below, so every real date cell in a user-authored workbook imported as UTC whatever the tenant's zone.

## The three rulings, honoured

1. **An explicit offset is honoured exactly as written.** `…Z` and `…+08:00` already name one instant; the `NAIVE_DATE_TIME` pattern is anchored at both ends so it cannot match them. This change affects naive cells only.
2. **The date-only fast path stays UTC.** `YYYY-MM-DD` is UTC per ECMAScript and a `date` is a timezone-naive calendar day (ADR-0053). Not swept into the zoned handling.
3. **No zone resolves ⇒ UTC**, never the process clock — pinned, and the reason is stated: this is the fallback the export's own cell path takes (`wallClock` = `zonedWallClock(...) ?? utcWallClock(...)`), so the round trip stays exact for deployments that configure no zone. A process-`TZ` fallback would have preserved the defect for exactly the deployments that cannot see it. An explicitly resolved `'UTC'` is a resolved zone, not a missing one.

**One behaviour change for existing deployments** follows from ruling 3: a host with a non-UTC `TZ` and no resolved business timezone previously read naive cells in the host clock and now reads them as UTC.

## Degenerate DST readings

Two wall clocks are not a bijection with instants. Both degenerate cases resolve to the **earlier candidate instant** — measured, and pinned including the local clock each lands on:

- **gap** (02:30 on a US spring-forward day, which never happens) settles on the post-transition offset and lands just *before* the gap, at 01:30 EST — the opposite of Temporal's `'compatible'` disambiguation, which pushes forward;
- **ambiguous** (01:30 on a US fall-back day, which happens twice) resolves to its first occurrence, 01:30 EDT.

A cell naming a clock its zone never had is ambiguous by construction; what an import needs is an answer that is deterministic and host-independent, which both are.

## Tests

Verified at HEAD `1f0ee1791`.

The acceptance criterion is inverse-ness of the **pair**, so it is asserted on the pair, and **every fixture runs under a host `TZ` deliberately different from the business timezone** — the configuration where the old code is wrong, and the only one where these can fail. Every datetime fixture straddles a **month** boundary, because a mid-day instant survives most wrong implementations untouched.

- `parseDateCell`: the reported cell lands in the tenant's month on all four hosts; a companion assertion measures that the *process-clock* reading really does produce four different instants, so the host-independence claim is not vacuous.
- `export → import` identity across every zone × every host × six instants, including both DST edges, plus a guard that the export cell really is offset-free (otherwise the round trip would pass through the honoured-offset branch and stop covering this).
- Route level: the real export route feeding the real import route over a real engine on sqlite `:memory:`; only `resolveExecCtx` is stubbed. `storedInstant` refuses an offset-free stored value rather than letting `new Date()` re-read the host clock.
- xlsx: both an ExcelJS-encoded `Date` and a **raw Excel serial + `numFmt`** — what a user-authored file actually stores — the latter added here so the fixture does not depend on ExcelJS's encode and decode being wrong symmetrically.

**Reverse verification** (predicted red, observed red): ablating the naive-datetime branch turns **15 tests red**, including the route-level round trip and the raw-serial xlsx case; restored from the commit afterwards.

| run | result |
| --- | --- |
| `@objectstack/core` test | 34 files, 831 passed |
| `@objectstack/rest` test | 117 files, 1942 passed |
| `@objectstack/service-analytics` test | 75 files, 1675 passed |
| `@objectstack/rest` typecheck | pass |
| full workspace build | 70/70 |

`service-analytics` is the one out-of-package consumer of `zonedDateStartToUtcMs`, run because that function is widely consumed and the merge queue runs the full suite. `zonedDateStartToUtcMs` keeps its exact contract — still date-only (a datetime argument is still `NaN`, pinned), and its equivalence to the generalised form is pinned across seven zone/date pairs.

Gates re-derived at final HEAD with `scripts/pm/dispatch-gates.mjs` against the actual changed paths and all green: `check:nul-bytes`, `check:changeset-gate-self-tests`, `check:cross-package-test-inputs`, `check:error-code-casing`, `check:kernel-hook-pairs`, `check:objectui-changeset`, `check:test-source-alias`, `check:query-options-erasure`, `check:type-check-coverage`, `check:type-check-debt --re-measure`, plus `check-adr-0087-registration`, `check-changeset-no-major`, `check-cross-package-test-inputs`, `check-empty-changeset`. Nothing refused to run.

---

### DIFF
--- a/.changeset/import-naive-datetime-business-timezone.md
+++ b/.changeset/import-naive-datetime-business-timezone.md
@@ -1,0 +1,73 @@
+---
+"@objectstack/core": patch
+"@objectstack/rest": patch
+---
+
+fix(rest): read an offset-free import cell in the business timezone, not the host `TZ` (#8485)
+
+`parseDateCell` ended in `new Date(s)`. A spreadsheet cell like
+`2026-08-01 06:00:00` carries no offset, so ECMAScript resolves it against the
+**process** timezone, and the instant bulk import stored became a property of
+the deployment host:
+
+```
+TZ=Asia/Shanghai → 2026-07-31T22:00:00.000Z
+TZ=UTC           → 2026-08-01T06:00:00.000Z
+```
+
+Same file, same tenant, same cell — eight hours apart, decided by a setting
+nobody authoring the spreadsheet can see, and never consulting the business
+timezone the route had already resolved one frame up
+(`ExecutionContext.timezone`, the platform-default → global → tenant cascade).
+
+Since the export renders `datetime` cells in that business timezone (#8373), the
+advertised export → edit in a spreadsheet → re-import round trip was lossless
+only where the host `TZ` happened to equal the business zone. `import-coerce.ts`
+opens by calling itself "the inverse of `export-format.ts`"; it now is one, and
+the regression proof asserts inverse-ness on the **pair** — every fixture under
+a host `TZ` deliberately different from the business timezone, because a test
+that runs only under a matching `TZ` cannot fail.
+
+**An offset-free datetime cell is now read in the caller's business timezone**,
+through `@objectstack/core`'s new `zonedWallClockToUtcMs` — the DST-safe wall
+clock → instant primitive that `zonedDateStartToUtcMs` (the date-bucket drill
+path) is now the midnight special case of. One implementation of zone
+arithmetic, `Intl` offsets from the platform tz database, never hand-rolled;
+generalising the existing one rather than hand-rolling a second in `rest` is
+what keeps the export and import halves of this seam from drifting apart again.
+Two wall clocks are not a bijection with instants, and both degenerate DST
+readings resolve to the earlier candidate instant — a gap reading lands just
+before the gap, an ambiguous reading on its first occurrence (pinned, measured).
+
+Three things deliberately do **not** move:
+
+- **A cell that carries an explicit offset** (`…Z`, `…+08:00`) already names one
+  instant and is honoured exactly as written. This change affects naive cells
+  only.
+- **The date-only fast path stays UTC.** `YYYY-MM-DD` is UTC per ECMAScript and
+  a `date` is a timezone-naive calendar day (ADR-0053); sweeping it into the
+  zoned handling to make the code look uniform would silently re-time every
+  date-only import to fix nothing.
+- **No timezone resolved ⇒ UTC**, never the process clock. That is the fallback
+  the export's cell path takes in the same case, so the round trip stays exact
+  for deployments that configure no zone — and a process-`TZ` fallback would
+  preserve the defect for exactly the deployments that cannot see it. This is
+  the one **behaviour change for existing deployments**: a host with a non-UTC
+  `TZ` and no resolved business timezone previously read naive cells in the host
+  clock and now reads them as UTC. An explicitly resolved `'UTC'` is a resolved
+  zone, not a missing one.
+
+Two adjacent legs of the same defect, both on the naive-cell path:
+
+- **A naive cell landing in a `date` or `time` field** now takes the typed
+  components verbatim (`2026-08-01 06:00:00` → `2026-08-01` / `06:00:00`).
+  Those branches also read the process clock, so a host east of the cell stored
+  the *previous calendar day* for a `date` column.
+- **An xlsx date cell.** An Excel serial date carries no timezone; ExcelJS
+  materialises it as a `Date` whose UTC components are the sheet's wall clock,
+  and `import-prepare.ts` rendered it with `toISOString()` — stamping a `Z` the
+  file never had. That fabricated offset then outranked the business timezone by
+  the very carve-out above, so every real date cell in a user-authored workbook
+  imported as UTC whatever the tenant's zone. It now flattens to the same
+  offset-free `YYYY-MM-DD HH:mm:ss` a CSV export writes, which is what that
+  function's contract already claimed to produce.
--- a/packages/core/src/utils/datetime.test.ts
+++ b/packages/core/src/utils/datetime.test.ts
@@ -6,7 +6,12 @@
 // database, and midnight must land exactly on the day boundary in that zone.
 
 import { describe, it, expect } from 'vitest';
-import { zonedDateStartToUtcMs, calendarPartsInTz, nextUtcCalendarDay } from './datetime.js';
+import {
+  zonedDateStartToUtcMs,
+  zonedWallClockToUtcMs,
+  calendarPartsInTz,
+  nextUtcCalendarDay,
+} from './datetime.js';
 
 const iso = (s: string) => Date.parse(s);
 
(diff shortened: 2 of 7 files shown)


qq9340100:

This card was **resumed** after a container restart killed its first agent. That agent's commit survived unpushed in its worktree and was recovered rather than redone — the right handling, and it is why this landed as one coherent commit instead of a second attempt.

All three rulings verified from the diff, not from the PR body.

**Ruling 3 — the fallback is a deliberate stated choice: VERIFIED, and this was the one I could not decide for you.** `zonedWallClockToUtcMs` reads `if (!tz || tz === 'UTC' || Number.isNaN(wallAsUtc)) return wallAsUtc` where `wallAsUtc = Date.UTC(...)`. So no-zone ⇒ **UTC**, never the process clock. The reasoning given is the right one and is better than the alternative I floated: it matches the export's own cell-path fallback, so the round trip stays exact for deployments that configure no zone — whereas a process-`TZ` fallback would have preserved the defect for precisely the deployments that cannot see it. The resulting behaviour change for a non-UTC host with no resolved business timezone is stated plainly in the PR and carried in the changeset rather than buried.

**Ruling 1 — explicit offsets untouched: VERIFIED.** The `NAIVE_DATE_TIME` pattern is anchored at both ends, so `…Z` and `…+08:00` cannot match it.

**Ruling 2 — date-only fast path stays UTC: VERIFIED.** `zonedDateStartToUtcMs` keeps its exact `^(\d{4})-(\d{2})-(\d{2})$` regex and its `NaN` return for anything else, now delegating to the general form. The contract is preserved *and* the duplication is gone — the "generalise, don't hand-roll a second one" outcome, with the equivalence pinned across seven zone/date pairs.

**Scope grew past the headline, and each step is justified rather than creep.** The `date`/`time` branches read the same process clock (a host east of the cell stored the previous calendar day), and the xlsx fix is not optional: ExcelJS stamped a `Z` the file never had, and under ruling 1 that fabricated offset would have **outranked** the business timezone — so without it the fix would not have worked on the most common real input. Everything stayed inside the declared surface; the diff is exactly `core/utils/datetime.*`, `rest/import-*`, and one new test file.

**A real defect was caught while generalising, and documented at the site:** `formatToParts` resolves no finer than a second, so a sub-second wall clock would leak its milliseconds into the computed offset. Invisible in the date-only form where ms is always 0, live the moment the primitive generalises.

**The tests are built so they can actually fail.** Every fixture runs under a host `TZ` deliberately different from the business timezone — the only configuration where the old code is wrong — and every datetime fixture straddles a **month** boundary rather than a mid-day instant. Two touches raise this above the usual: a companion assertion measures that the *process-clock* reading really does produce four different instants, so the host-independence claim is not vacuous; and a guard asserts the export cell really is offset-free, without which the round trip would slip through the honoured-offset branch and stop covering the thing under test. Reverse verification predicted red and measured **15 red**, including the route-level round trip and the raw-Excel-serial case.

The raw serial + `numFmt` fixture is the right instinct — it removes the dependence on ExcelJS's encode and decode being wrong symmetrically, which is exactly how a round-trip test can pass over a broken implementation.

**Both DST degenerate cases resolve to the earlier candidate instant, measured and pinned including the local clock each lands on**, with the divergence from Temporal's `'compatible'` disambiguation named rather than glossed. Determinism and host-independence are the properties an import needs, and the PR argues that rather than asserting it.

`service-analytics` was run as the out-of-package consumer of `zonedDateStartToUtcMs` — the affected-subset-vs-full-suite gap handled before queueing, not after a kick.

⛔ **Flip held** until every gate job concludes `success` on its own reading. Nothing refused to run locally, which is the answer I want on a card that touches a widely-consumed `core` primitive.

---

Issue #8780, opened 2026-08-15
ja-JP sharing bundle still names the pre-rename role concept in sys_record_share.recipient_id


qq9340100:
Found while bounding the es-ES rename damage in #8735. Same ADR-0090 (`sys_role` → `sys_position`) half-landing, different locale, so it was outside that card's lane and is filed rather than fixed there.

## The instance

`packages/plugins/plugin-sharing/src/translations/ja-JP.objects.generated.ts`, `sys_record_share.fields.recipient_id.help`:

```
help: "アクセスを受け取るユーザー/グループ/ロールの ID"
```

`ロール` is the pre-rename *role* concept. The English source for this exact leaf says **position**:

```
help: "ID of the user/group/position that receives access"
```

and the same bundle already renders the concept correctly one field earlier, in `recipient_type.options`:

```
position: "ポジション"
```

so the file disagrees with itself. zh-CN has the corresponding leaf right (`用户 / 组 / 岗位 ID`), and es-ES is being repaired in #8735 — ja-JP is the only locale left carrying it.

## Bound

This was swept, not guessed. Every `ロール` / `角色` occurrence in every ja-JP and zh-CN bundle repo-wide was checked against its English source. All the others are legitimate and must NOT be changed:

- `ロールアップ` (rollup) and `ロールバック` (rollback) — different words that merely contain the substring
- ARIA `ロール` / `角色`
- platform role, org member role, agent persona role — concepts genuinely still called *role*, unaffected by ADR-0090

This one leaf is the only genuine miss in the two locales.

## Fix shape

Leaf string value only, per the bundle header's documented workflow — no generator change, no key rename. `ロール` becomes `ポジション`, matching the `recipient_type.options.position` rendering already in the file.

## Why nothing caught it

Same blind spot as #8735: `check:i18n` compares bundle *structure* against extractor output and is green either way, and per #8543 the generator's merge mode leaves existing leaf values untouched, so a half-landed rename survives a green gate indefinitely.

The consistency guard added in #8735's PR (`recipient-vocabulary-consistency.test.ts`) asserts cross-object agreement within a locale and a stale-term check for es-ES; it does not assert a ja-JP stale-term rule, precisely because this instance is still open. Whoever fixes this can extend that guard to ja-JP in the same change, which would close the class rather than the instance.

hotlong:
[...]

Issue #8543, opened 2026-08-13
check-i18n-bundles merge mode never updates an EXISTING field description, so editing one leaves the en bundle silently stale (and the gate green)


os-zhuang:
Observation-class finding, noticed while implementing #8069 (PR #8541). Unassigned, no `pm:queue`.

## What I measured

I edited two field descriptions on `sys_http_delivery` (`packages/services/service-messaging/src/objects/http-delivery.object.ts`):

- `error` — had **no** `description` before; gained one.
- `attempts` — **already had** one; its text changed.

`pnpm check:i18n` went red, and `node scripts/check-i18n-bundles.mjs --write` regenerated. The result, per locale bundle, was `2 insertions, 1 deletion` — only the `error` entry:

```
       error: {
-        label: "Error"
+        label: "Error",
+        help: "Why this row is not a delivered callout: the last transport error, or — on a row with 0 attempts — why the delivery could never be prepared (#8069)."
       },
```

`attempts` in `en.objects.generated.ts` still reads `help: "Number of attempts made so far"` — the pre-edit text — while the source now says something else. The gate then reports **OK**.

That is the documented merge behaviour, printed by the script itself: *"Merge mode is on, so no existing translation is overwritten — new schema keys are added filled with the source text."*

## Why it is worth recording anyway

For `zh-CN` / `ja-JP` / `es-ES` the rule is obviously right: a regeneration must not clobber a human translation with English source text.

For **`en`** the same rule produces a different outcome, because `en` is not a translation — it is a copy of the source. So there, merge mode means:

- an author edits a field `description`,
- the gate passes,
- the change never reaches the bundle the UI renders,
- and nothing anywhere says the two disagree.

The source of truth and the served text drift apart silently, permanently, and by one commit at a time. It is the `declared != enforced` shape the repo keeps filing cards about, in the one place a gate is already looking.

## What this finding does NOT claim

- ⛔ Not that any specific user-visible help text is currently wrong. I measured one instance, mine, and did not sweep for pre-existing drift. **The size of the existing gap is unknown and is the first thing worth measuring** — a scan comparing every authored `description` against its `en` bundle `help` would grade this immediately, and would also decide whether this is a one-line fix or a backlog.
- Not that merge mode is wrong for translated locales. It is right for them; the question is only whether `en` belongs in the same channel.
- Not urgent: help text going stale is a documentation degradation, not a durability one — nothing is lost, something is merely out of date while looking current.

## Repro

[...]

Issue #8430, opened 2026-08-13
plugin-sharing's read-path merge is a THIRD read-scope boundary the #8220 provenance mark does not cover — sharing-composed queries withhold the author diagnostic (fail-closed, but restorable)


hotlong:
Out-of-scope finding from #8220 / PR #8429, measured by its dev (cloud session `session_01AkiE189Csm7aBT2qCVSx3X`) and filed by the dispatching PM on its behalf (the dev's environment had no GitHub API egress — provenance in the `os-dev-report` on #8220). Unassigned, for triage; lands in `packages/plugins/plugin-sharing` (identity lane's family).

## The gap

#8220 declared the filter-subtree provenance mark and set it at TWO read-scope merge boundaries (plugin-security's RLS injection; service-analytics' `withReadScope`). `plugin-sharing`'s read-path merge (`sharing-plugin.ts` `composeAnd` sites, ~:871/:896) is a **third** merge boundary outside that card's declared file surface:

- Its injected sharing filters arrive at the driver **unmarked** ⇒ withheld. That direction is fail-closed and therefore CORRECT — nothing discloses.
- But an author's own filter on an object with **active sharing rules** loses the identity vouch at the security boundary (by the time plugin-security sees it, `ast.where` has already been rewritten by the sharing composition), so the author's OWN cross-field refusal stays redacted — the exact author-experience cost #8220 existed to remove, surviving on one composition path.

## The fix shape (for whoever takes it)

Mark at plugin-sharing with the #8220 spec exports (`markFilterSubtreeProvenance` from `@objectstack/spec/data`): `'policy'` on its injected sharing filters; it can also preserve the author vouch across its rewrite (mark the caller's subtree `'author'` under the same identity-vouch discipline plugin-security uses — a rewritten sibling tree is vouched for nobody). Pins per the #8220 pattern: author names columns / policy withholds / unmarked withholds, at a real driver.

## Constraints inherited from #8220 (binding on any implementation)

- Fail-closed: unmarked/ambiguous ⇒ withheld; the mark is permission to reveal.
- One mechanism: consume `resolveFilterSubtreeProvenance` — ⛔ do not invent a parallel provenance concept.

Refs: #8220 (the mark + both existing boundaries), PR #8429, #8198 (the B-phase redaction), #7929 (the ruling).

hotlong:
[...]

PR #8788, opened 2026-08-15
fix(plugin-security,plugin-sharing): repair the ADR-0090 position rename in the es-ES bundles, bounded and guarded

### LINKED ISSUE #8735: es-ES object bundle carries `rol`→`puesto` find-replace damage (`Puestoes`, `contpuesto`, a surviving `Clonar rol`)
Tripped over while implementing #8601 (which corrects a *different* class of defect in these same files — an inaccurate uniqueness claim). This card is about **shipped Spanish that is simply broken**, not about a claim that disagrees with the source.

`packages/plugins/plugin-security/src/translations/es-ES.objects.generated.ts` reads as though a global `rol` to `puesto` find-replace was run over it at some point, without review of the results.

## Instances observed

| line | shipped today | problem |
|---|---|---|
| `:13` | `pluralLabel: "Puestoes"` | not a Spanish word; the plural of `puesto` is `Puestos` |
| `:15` | `description: "Definiciones de puesto para el contpuesto de acceso RBAC"` | `control` was itself caught by the replace and mangled into `contpuesto`. Should read `control de acceso`. |
| `:91` | `clone_position.label: "Clonar rol"` | the replace *missed* this one, so it still says `rol` while the rest of the object says `puesto` |

`:15` is the clearest tell: `con` + `trol` became `con` + `tpuesto`, which only happens if the replace was substring-based and unreviewed. `:91` is the same event's other half — one instance it should have changed and did not.

These are all `sys_position`. An admin using Setup in Spanish sees `Puestoes` as the object's plural label in navigation and list views.

## ⚠️ This is a SAMPLE, not a bounded set

I did **not** sweep for further instances. #8601 explicitly forbade the bundle sweep (`"do not do the sweep for other stale bundles the card suggests — that is a separate card if wanted, not a rider on this one"`), so I recorded only what I saw while reading the exact lines that card needed. The same find-replace would plausibly have touched other objects in this bundle, and possibly other `es-ES` bundles in other packages. **Anyone picking this up should begin by bounding it**, not by fixing these three lines.

## Not caught by any gate

Same blind spot as #8601: `check:i18n` compares bundle *structure* against what the extractor emits and is green either way, so nothing mechanical judges whether a translated string is good Spanish. See #8543 for why merge mode leaves existing values untouched.

## Fix shape

Leaf string values only, per the bundle header's documented workflow — the same route #8601 used. No generator change.

## Related

- #8601 — the uniqueness-claim defect in these same files; different class, fixed separately in PR #8723.
- #8543 — the generator defect that leaves existing leaf values untouched, so drift and damage both survive silently.

qq9340100:
Fixes #8735

The card cited three lines and said plainly they were a **sample, not a bounded set** — #8601 had forbidden the sweep as a rider, so nobody had ever established the extent. Bounding it was the first deliverable, so this PR leads with the sweep and its result.

## The bound

The damage is a substring `rol` to `puesto` find-replace, run without review. `contpuesto` proves the mechanism: `con` + `trol` only mangles that way with no word boundary. That makes it mechanically boundable in two directions, and both were swept over **all 12 `es-ES` bundles in the repo**, at `origin/main`:

1. **Mangling** — any word *containing* `puesto` that is not a real Spanish word, checked against an allowlist of the legitimate ones (`presupuesto`, `expuesto`, `compuesto`, `supuesto`, `dispuesto`, `impuesto`, `propuesto`, `repuesto`, `opuesto`).
2. **Misses** — any surviving whole-word `rol` / `roles`, then each one read against its **English source leaf** to decide whether it names the renamed concept or a genuinely different one.

Result: **14 damaged leaves in 2 files**. Every other `es-ES` bundle is clean.

| package | mangled | missed | total |
|---|---|---|---|
| `plugin-security` | 2 | 9 | 11 |
| `plugin-sharing` | 0 | 3 | 3 |
| the other 10 bundles | 0 | 0 | 0 |

The count is independently corroborated: restoring the pre-fix `plugin-security` bundle over the repaired one produces exactly `11 insertions, 11 deletions`.

Mangling is confined to **one file**. `plugin-security` is the only bundle where the replace physically ran.

### What made the verdicts decisive rather than a matter of taste

Neither package's **English** bundle contains the word `role` anywhere — `grep -c -i role` returns `0` for both. That settles the direction both ways: every `rol` in their Spanish is a missed rename, and no legitimate "role" was ever over-replaced into `puesto`, so nothing needs reverting.

`plugin-sharing` also self-corroborates. `sys_record_share` and `sys_sharing_rule` carry the same `recipient_type` vocabulary, and the rename landed on the rule object and was missed on the share object — so one file rendered one enum key two ways:

- `position` — `Puesto` on the rule, `posición` on the share
- `unit_and_subordinates` — `Unidad de negocio y subordinados` on the rule, `Rol y subordinados` on the share

The corrected wording is not invented here; it is what the sibling object in the same file already said, and zh-CN and ja-JP agree (`岗位` / `业务单元及下级`, `ポジション` / `ビジネスユニットと下位階層`). The lowercase `posición` among capitalized siblings (`Usuario`, `Grupo`, `Invitado`) is a further tell it was machine-written, not authored.

## Swept surface — what was examined, and what was not

**Examined:** all 12 `es-ES` bundles repo-wide, for both damage classes, each hit read against its English source leaf. Also swept every `ロール` and `角色` in every ja-JP and zh-CN bundle repo-wide, to test whether the event crossed locales.

**Deliberately not changed** — these were examined and judged *not* damage, per the card's instruction not to "fix" Spanish that is merely different from a reviewer's preference:

- **`platform-objects` — 14 surviving `rol`.** All legitimate. `Rol de plataforma` (better-auth platform role), `Rol` of an organization member, `Rol del agente` (agent persona), and ARIA `roles` are concepts still genuinely called *role* in English; ADR-0090 did not rename them. Changing these would have been the damage.
- **`posición` used as prose for the position concept**, in `plugin-security` `sys_position.fields.delegatable.help` and in `platform-objects` `es-ES.metadata-forms`. The English there really does say "position" and `posición` is a valid Spanish rendering of it — it is neither a non-word nor a surviving `rol`. It is a terminology-consistency question against the object's own label `Puesto`, which is a different class from this card's, so it is left for a reader who wants to take that decision deliberately. Flagged here rather than silently changed.
- **`service-messaging`** English source reads `principal (role/team/user)` and its Spanish leaf is an untranslated copy. That is a gap in the English source and the known merge-mode behaviour of #8543, not find-replace damage.

**Out of lane, filed not fixed:** #8430 is in flight and declares `packages/plugins/plugin-sharing/src/**`; only `src/translations/**` is touched here, and nothing outside it. No `plugin-sharing` damage was found outside `translations/`.

**Found in another locale, filed as #8780** (not addressed in this PR): `plugin-sharing` ja-JP `sys_record_share.recipient_id.help` still says `ロール` where English says *position* — the identical half-landing, one locale over. It is outside this card's `es-ES` lane, so it is recorded there rather than ridden along here. #8780 remains open.

## Guards

No mechanical gate can judge this. `check:i18n` compares bundle **structure** against extractor output and is green either way, and per #8543 the generator's merge mode never rewrites an existing leaf, so damage survives a green gate indefinitely. **The `check:i18n` green below is not evidence the Spanish is correct** — it only says the structure still matches.

So the durable artifact is two regression tests, both of which were **reverse-verified**: run against the pre-fix bundles they go red, naming exactly the real defects and nothing else.

- `plugin-security/src/translations/position-rename-consistency.test.ts` — stale pre-rename terms across all four locales of the three renamed objects, plus a malformed-compound rule catching `Puestoes` and `contpuesto` without tripping on real Spanish words that contain the substring. Against the damaged bundle: 3 failures.
- `plugin-sharing/src/translations/recipient-vocabulary-consistency.test.ts` — within a locale, an option key shared by several objects must render identically. This needs no per-string judgement and stays reviewable by someone who does not read the locale. Against the damaged bundle: 2 failures, naming both enum conflicts and both stale leaves.

## Verification

All commands at `e64dd5ad5`, after the final commit.

- `plugin-security` + `plugin-sharing` test suites: **1769 passed** (1190 + 579), 85 files, 0 failed.
- Gate union re-derived from the actual changed paths via `scripts/pm/dispatch-gates.mjs`, all green: `check:nul-bytes`, `check:test-source-alias`, `check:type-source-resolution`, `check:cross-package-test-inputs`, `check:changeset-gate-self-tests`, `check:objectui-changeset`, `check:query-options-erasure`, `check:type-check-coverage`, `check-adr-0087-registration`, `check-changeset-no-major`, `check-empty-changeset`, `check-cross-package-test-inputs.mjs`.
- `check:i18n` — **OK, 9 packages, all bundles in sync.** It first refused (`PREREQUISITE NOT MET`, CLI not built); built `@objectstack/cli` and re-ran, so this is measured, not skipped.
- `check:type-check-debt` — **OK, 33 ledger entries re-measured, none above its recorded number.** It too first refused (closure not built); built the full closure as `lint.yml` does, then re-ran. The `-1` surplus it reports for `@objectstack/lint` is pre-existing and untouched by this diff.

Leaf string values only, per the bundle header's documented workflow. No generator change, no restructuring, no key renames.

---

### DIFF
--- a/.changeset/es-es-position-rename-damage.md
+++ b/.changeset/es-es-position-rename-damage.md
@@ -1,0 +1,26 @@
+---
+"@objectstack/plugin-security": patch
+"@objectstack/plugin-sharing": patch
+---
+
+Repair the ADR-0090 `sys_role` → `sys_position` rename in the es-ES object
+translation bundles, and guard it mechanically.
+
+The rename half-landed in Spanish: an unreviewed substring find-replace produced
+two non-words (`Puestoes` as the plural of `Puesto`, and `contpuesto` where the
+replace ate the unrelated word `control`), while nine further leaves in
+`plugin-security` and three in `plugin-sharing` were missed entirely and still
+named the pre-rename concept. In `plugin-sharing` the same picklist key rendered
+two different ways in one file — `position` was `Puesto` on the sharing rule and
+`posición` on the record share, and `unit_and_subordinates` read `Rol y
+subordinados` (naming the removed role concept) against `Unidad de negocio y
+subordinados` on its sibling.
+
+Spanish-facing admins saw `Puestoes` as the object's plural label in navigation
+and list views, and two different words for one recipient kind across two Setup
+screens.
+
+Two regression guards now cover the classes involved: a malformed-compound and
+stale-term check on the renamed security objects, and a self-consistency check
+asserting that a picklist option key shared by several sharing objects renders
+identically within a locale. Neither needs a reader of the locale to review it.
--- a/packages/plugins/plugin-security/src/translations/es-ES.objects.generated.ts
+++ b/packages/plugins/plugin-security/src/translations/es-ES.objects.generated.ts
@@ -11,8 +11,8 @@ import type { TranslationData } from '@objectstack/spec/system';
 export const esESObjects: NonNullable<TranslationData['objects']> = {
   sys_position: {
     label: "Puesto",
-    pluralLabel: "Puestoes",
-    description: "Definiciones de puesto para el contpuesto de acceso RBAC",
+    pluralLabel: "Puestos",
+    description: "Definiciones de puesto para el control de acceso RBAC",
     fields: {
       label: {
         label: "Nombre visible"
@@ -49,7 +49,7 @@ export const esESObjects: NonNullable<TranslationData['objects']> = {
         }
       },
       id: {
-        label: "ID de rol"
+        label: "ID de puesto"
       },
       created_at: {
         label: "Creado el"
@@ -74,12 +74,12 @@ export const esESObjects: NonNullable<TranslationData['objects']> = {
     },
     _actions: {
       activate_position: {
-        label: "Activar rol",
+        label: "Activar puesto",
         successMessage: "Puesto activado"
       },
       deactivate_position: {
-        label: "Desactivar rol",
-        confirmText: "¿Desactivar este rol? Los usuarios con el puesto conservan su asignación, pero el puesto deja de otorgar permisos hasta que se vuelva a activar.",
+        label: "Desactivar puesto",
+        confirmText: "¿Desactivar este puesto? Los usuarios con el puesto conservan su asignación, pero el puesto deja de otorgar permisos hasta que se vuelva a activar.",
         successMessage: "Puesto desactivado"
       },
       set_default_position: {
@@ -88,7 +88,7 @@ export const esESObjects: NonNullable<TranslationData['objects']> = {
         successMessage: "Puesto predeterminado actualizado"
       },
       clone_position: {
-        label: "Clonar rol",
+        label: "Clonar puesto",
         successMessage: "Puesto clonado",
         params: {
           label: {
@@ -338,13 +338,13 @@ export const esESObjects: NonNullable<TranslationData['objects']> = {
     }
   },
   sys_position_permission_set: {
-    label: "Conjunto de permisos de rol",
-    pluralLabel: "Conjuntos de permisos de rol",
-    description: "Vincula un conjunto de permisos a un rol.",
+    label: "Conjunto de permisos de puesto",
+    pluralLabel: "Conjuntos de permisos de puesto",
+    description: "Vincula un conjunto de permisos a un puesto.",
     fields: {
       id: {
         label: "ID de vinculación",
-        help: "UUID de la vinculación rol-conjunto de permisos."
+        help: "UUID de la vinculación puesto-conjunto de permisos."
       },
       position_id: {
         label: "Puesto",
--- a/packages/plugins/plugin-security/src/translations/position-rename-consistency.test.ts
+++ b/packages/plugins/plugin-security/src/translations/position-rename-consistency.test.ts
@@ -1,0 +1,120 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+//
+// Regression guard for the ADR-0090 P1 rename (sys_role → sys_position). The
+// rename half-landed in es-ES and shipped for a month: the object label said
+// "Puesto" while the plural said "Puestoes", the description said "contpuesto"
+// (the rename caught the unrelated word `control` — con+trol → con+tpuesto), and
+// nine leaf values still said "rol" — including one string that says both, in the
+// same sentence.
+//
+// Nothing mechanical caught it: `check:i18n` compares bundle STRUCTURE against
+// what the extractor emits and is green either way, and the generator's merge
+// mode never rewrites an existing translated leaf, so both drift and damage
+// survive a green gate indefinitely. This test is the missing judgement, and it
+// is deliberately modelled on `bu-rename-consistency.test.ts` (ADR-0057), which
+// exists because that rename half-landed the same way.
+//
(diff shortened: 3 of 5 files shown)


qq9340100:

This card was **resumed** after a container restart; its predecessor's work survived only as *uncommitted* working-tree changes, which is one `git checkout` from gone. It was recovered, committed, verified, and extended.

**The card's central demand was met, and it was the one most likely to be skipped.** The three cited lines were an explicitly declared *sample*; a PR fixing exactly those three and writing `Fixes` would have closed the card with the defect still shipped. This one leads with the bound: all **12** `es-ES` bundles swept in both directions, result **14 damaged leaves in 2 files**, with the other 10 clean.

**What makes the verdicts checkable rather than a matter of taste** — and this is the part I'd have accepted a weaker version of: neither package's **English** bundle contains the word `role` at all (`grep -c -i role` → 0 for both). That settles the direction *both* ways in one reading: every `rol` in their Spanish is a missed rename, and nothing legitimate was over-replaced into `puesto`, so nothing needs reverting. A zero-hit with its own control, which is the standard I keep asking for.

**Verified independently from the diff:**
- Five files: a changeset, the two bundles, two new tests. `platform-objects` — where 14 *legitimate* `rol` live — is **untouched**, which is the correct outcome and the one a careless sweep gets wrong.
- Every changed leaf is genuine damage: `Puestoes` → `Puestos`, `contpuesto` → `control`, missed renames on the position concept, and the two `plugin-sharing` enum values.
- In `plugin-sharing` only `src/translations/**` is touched, honouring the constraint I added mid-flight for #8430.

**The `plugin-sharing` correction is self-corroborating rather than invented.** One file rendered the same `recipient_type` key two ways — `Puesto` on the rule object, `posición` on the share; `Unidad de negocio y subordinados` versus `Rol y subordinados`. The corrected wording is what the sibling object in that same file already said, with zh-CN and ja-JP agreeing. The lowercase `posición` sitting among capitalised siblings is a nice tell that it was machine-written.

**What was deliberately left alone is as important as what changed,** and each exclusion is argued: `platform-objects`' legitimate role concepts; the `posición`-as-prose cases, correctly identified as a terminology-consistency question of a *different class* and flagged rather than silently changed; and `service-messaging`, diagnosed as an English-source gap plus #8543's merge-mode behaviour rather than find-replace damage.

**Both guards were reverse-verified** — red against the pre-fix bundles, naming exactly the real defects and nothing else (3 failures and 2 failures respectively). The `plugin-sharing` guard is the better of the two because it needs no per-string judgement: an option key shared across objects must render identically within a locale, which stays reviewable by someone who does not read Spanish.

**Gate discipline exactly right.** `check:i18n` and `check:type-check-debt` both initially **refused to run** and were reported as such, then the prerequisites were built and both re-run to real readings — no "not applicable". And the PR states outright that `check:i18n` green is **not** evidence the Spanish is correct, which was a ruling in the dispatch and is the honest framing.

**#8780 verified as existing** (the ja-JP half-landing, found while sweeping and correctly filed rather than ridden along, since it is outside the `es-ES` lane). It was filed with **no labels**, so I applied `finding` — a bare issue is invisible to both the queue and the finding views. `domain:*` left to triage. Its suggestion that whoever fixes it extend the new guard to ja-JP is the right instinct: that closes the class rather than the instance.

---

PR #8791, opened 2026-08-15
fix(plugin-sharing): stamp read-scope provenance at the third merge boundary, restoring the author's own cross-field diagnostic

### LINKED ISSUE #8430: plugin-sharing's read-path merge is a THIRD read-scope boundary the #8220 provenance mark does not cover — sharing-composed queries withhold the author diagnostic (fail-closed, but restorable)

qq9340100:
Fixes #8430

`#8220` declared the filter-subtree provenance mark and set it at two read-scope merge boundaries — `plugin-security`'s CRUD injection and `service-analytics`' `withReadScope`. `plugin-sharing`'s read path is a **third**: on every read it AND-composes an OWD / record-share visibility filter into `ast.where`, and it stamped nothing.

This consumes the one declared mechanism (`markFilterSubtreeProvenance` from `@objectstack/spec/data`) — no parallel provenance concept, no local flag.

## The two marks

**`'policy'` on every scope this middleware injects** — the OWD/record-share read filter, the delegator's intersected filter (ADR-0090 D10), and the `sys_record_share` self-scope (ADR-0111 D5). **No behaviour change**: an unmarked subtree already withheld, so these refusals kept the `#7929` redaction before and keep it now. What changes is that the withhold becomes a *declared* verdict rather than an accident of the mark's absence — which matters because an unmarked node **inherits its ancestor's mark positionally** (`resolveFilterSubtreeProvenance`, innermost wins), so an unmarked policy arm nested inside a vouched subtree would read as the author's.

**`'author'` on the caller's own predicate**, vouched immediately before the rewrite that would otherwise make it unrecognisable to every later boundary. This is the one user-visible change: an author's own cross-field refusal on an object with active sharing again names its columns, its operator and its reason.

The vouch is an **identity check, not a heuristic** — the same discipline `plugin-security` applies. The mark is stamped only while `ast.where` is still, by object identity, the `where` the caller handed the engine. If a sibling already composed into it, or the engine rewrote it resolving filter tokens, identity fails and **nothing** is vouched: the tree stays unmarked, and unmarked withholds.

The arms of a pure `{ $and: [ ... ] }` root are vouched too, because `composeAnd`'s flattening branch spreads that root's arms into a new object and would otherwise drop the vouch out of the tree with it. That shape is what the array authoring form lowers to, so it is the common case rather than an edge one. The condition mirrors `composeAnd`'s own so the two stay in step.

## What I measured that contradicts the card

The card says the vouch is lost because "by the time plugin-security sees the AST, `ast.where` has already been rewritten by the sharing composition". **That is not the order the platform boots in.** `plugin-security` is `kernel.use`d before `plugin-sharing` on both real paths (`packages/verify/src/harness.ts`, `packages/cli/src/commands/serve.ts`), `resolvePluginOrder` preserves insertion order for plugins with no edge between them, and both register their middleware in `start()`. So security vouches **first**, and its mark — which lives on the caller's object — travels through this composition untouched.

Measured at a real driver, with a security-shaped boundary registered first: the author's refusal **already** named its columns before this change.

The gap is real, but it is the composition the card does not name: a stack that mounts `plugin-sharing` **without** `plugin-security` — an explicitly supported shape (`serve.ts` mounts SecurityPlugin as an optional pair, and `buildSharingMiddleware`'s own contract is that a stack without it "behaves exactly as before"). There, nothing ever vouches, and the author's own refusal stays redacted. Measured before: redacted. After: full diagnostic. That is the configuration the author pins run in, deliberately.

## Fail-closed, pinned in every direction

`packages/plugins/plugin-sharing/src/read-scope-provenance-mark.test.ts`, at a real `SqlDriver` on a real `better-sqlite3` database behind a real `ObjectQL` — a mocked driver cannot show the withhold actually happening:

- the injected scope still withholds, with the diagnostic relocated to the server log;
- **a policy arm sitting beside an author-vouched arm in the same `$and` still withholds** — the sharpest disclosure guard on this card;
- a predicate no boundary ever vouched still withholds **byte-identically** to the policy case;
- a `where` a sibling already rewrote gets **no** author mark.

Reverse verification: with the implementation reverted and the tests kept, **4 of 9 go red** — both author cases and both mark-setting cases — while the three fail-closed cases and the composed-with-security case stay green. That is the predicted direction: the policy and unmarked halves are behaviour-preserving by construction, so they *should* be insensitive to the change.

## Scope

**The write path is untouched.** `sharing-plugin.ts` has a third `composeAnd` region on the write path; marking provenance on a write filter is a different question with different consequences and was not declared by `#8220`. Left alone deliberately — see the report for the finding.

## Verification

All at `b89ca94bb` (the tree this PR pushes).

Tests — every package that consumes the provenance mechanism, not just the one I edited, since a change to how a filter subtree is marked can red a driver with no dependency path to `plugin-sharing`:

| package | result |
|---|---|
| `plugin-sharing` | 23 files / 583 tests passed |
| `plugin-security` | 61 files / 1177 tests passed |
| `driver-sql` | 97 files / 1663 passed, 4 files / 55 skipped |
| `driver-turso` | 37 files / 986 tests passed |
| `service-analytics` | 75 files / 1675 tests passed |
| `spec` | 402 files / 10638 tests passed |
| `runtime` | 157 files / 2391 tests passed |

`pnpm --filter @objectstack/plugin-sharing typecheck` clean.

Gates, re-derived with `scripts/pm/dispatch-gates.mjs` against the actual changed paths and all run at `b89ca94bb`: `check:nul-bytes`, `check:test-source-alias`, `check:type-source-resolution`, `check:query-options-erasure`, `check:i18n`, `check:type-check-coverage`, `check:type-check-debt` — **all pass**.

`check:i18n` initially **refused to run** ("PREREQUISITE NOT MET — the workspace CLI is not built"), which is NOT MEASURED rather than a pass. Built `@objectstack/cli` and re-ran: green, `plugins/plugin-sharing` bundles **in sync**, so no regeneration was needed and no translation bundle was written.

---

### DIFF
--- a/.changeset/sharing-read-merge-provenance-mark.md
+++ b/.changeset/sharing-read-merge-provenance-mark.md
@@ -1,0 +1,55 @@
+---
+"@objectstack/plugin-sharing": patch
+---
+
+fix(plugin-sharing): stamp the filter-subtree provenance mark at the read merge, so an author's own cross-field refusal stops being redacted on the sharing-composed path (#8430)
+
+`#8220` declared the filter-subtree provenance mark and set it at two read-scope
+merge boundaries — `plugin-security`'s CRUD injection and `service-analytics`'
+`withReadScope`. `plugin-sharing`'s read path is a **third**: on every read it
+AND-composes an OWD / record-share visibility filter into `ast.where`, and it
+stamped nothing.
+
+Two marks, and they are not the same job:
+
+- **the scopes it injects are marked `'policy'`** — the OWD/record-share read
+  filter, the delegator's intersected filter (ADR-0090 D10) and the
+  `sys_record_share` self-scope (ADR-0111 D5). **No behaviour change**: an
+  unmarked subtree already withheld, so these refusals kept the `#7929`
+  redaction before and keep it now. What changes is that the withhold becomes a
+  *declared* verdict instead of an accident of the mark's absence — which
+  matters because an unmarked node **inherits its ancestor's mark positionally**
+  (`resolveFilterSubtreeProvenance`, innermost wins), so an unmarked policy arm
+  nested inside a vouched subtree would read as the author's.
+- **the caller's own predicate is vouched `'author'`** immediately before the
+  rewrite that would otherwise make it unrecognisable to every later boundary.
+  This is the one user-visible change: an author's own `{ $field }` refusal on
+  an object with active sharing again names its columns, its operator and its
+  reason, instead of the redacted "operands withheld" text.
+
+**The vouch is an identity check, not a heuristic.** The mark is stamped only
+while `ast.where` is still, by object identity, the `where` the caller handed
+the engine. If a sibling middleware already composed into it, or the engine
+rewrote it resolving filter tokens, identity fails and **nothing** is vouched —
+the tree stays unmarked, and unmarked withholds. The arms of a pure
+`{ $and: [ … ] }` root are vouched too, because `composeAnd`'s flattening branch
+spreads that root's arms into a new object and would otherwise drop the vouch
+out of the tree with it (that shape is what the array authoring form lowers to,
+so it is the common case, not an edge one).
+
+**Fail-closed is unchanged in every direction**, and the pins say so at a real
+`SqlDriver`: the injected scope still withholds, a policy arm sitting beside an
+author-vouched arm in the same `$and` still withholds, and a predicate no
+boundary ever vouched still withholds byte-identically to the policy case.
+
+**The write path is untouched.** `buildWriteFilter`'s composition is a different
+question with different consequences and was not declared by `#8220`.
+
+Measured while implementing, and worth recording because the card says
+otherwise: in a stack that composes **both** plugins, the author vouch was
+already surviving. `plugin-security` is registered before `plugin-sharing` on
+both real boot paths and `resolvePluginOrder` preserves insertion order, so
+security vouches first and its mark — which lives on the caller's object —
+travels through this composition untouched. The gap this fixes is a stack that
+mounts `plugin-sharing` **without** `plugin-security`, where nothing else can
+vouch for the caller.
--- a/packages/plugins/plugin-sharing/src/read-scope-provenance-mark.test.ts
+++ b/packages/plugins/plugin-sharing/src/read-scope-provenance-mark.test.ts
@@ -1,0 +1,400 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#8430, extending #8220 / A of the #7929 maintainer ruling 2026-08-12] The
+ * THIRD read-scope merge boundary's half of the filter-subtree provenance mark.
+ *
+ * #8220 declared the mark and set it at two boundaries — `plugin-security`'s
+ * CRUD injection and `service-analytics`' `ObjectQLStrategy.withReadScope`.
+ * This middleware is a third: `buildSharingMiddleware` AND-composes an OWD /
+ * record-share visibility filter into `ast.where` on every read, and until this
+ * card nothing in that composition said which arm the caller wrote.
+ *
+ * What is pinned, and why each half matters:
+ *
+ *  - the injected sharing scope is marked `'policy'` — a cross-field refusal
+ *    raised from inside it keeps the #7929 redaction. This direction was
+ *    ALREADY correct as an unmarked subtree (unmarked withholds), so the pin
+ *    guards against the mark accidentally opening it, not against a regression
+ *    being fixed;
+ *  - the caller's own predicate is vouched `'author'` under the identity check
+ *    `ast.where === options.where`, so it survives THIS boundary's rewrite —
+ *    the one user-visible change on the card. A tree a sibling already rewrote
+ *    gets NO mark, and unmarked withholds: the mark is permission to reveal,
+ *    never a guess.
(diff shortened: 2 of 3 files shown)


qq9340100:

**The most valuable thing in this PR is that it falsified the card's stated mechanism.** #8430 says the author vouch is lost because "by the time plugin-security sees the AST, `ast.where` has already been rewritten by the sharing composition". Measured: that is not the order the platform boots in. `plugin-security` is `kernel.use`d **before** `plugin-sharing` on both real paths, `resolvePluginOrder` preserves insertion order absent an edge, and both register in `start()` — so security vouches first and its mark travels through this composition untouched. At a real driver with a security-shaped boundary registered first, the author's refusal **already** named its columns before this change.

That was dispatched as a falsifiable assumption precisely because it was the card's inference rather than a measurement, and the dev did the thing that makes that framing worth having: it measured the boot order rather than reasoning from the card's story.

**The gap is real, but it is a composition the card never names** — a stack mounting `plugin-sharing` **without** `plugin-security`, which is explicitly supported (`serve.ts` mounts SecurityPlugin as an optional pair, and `buildSharingMiddleware`'s own contract is that a stack without it "behaves exactly as before"). There nothing ever vouches and the author's own refusal stays redacted. Measured before: redacted. After: full diagnostic. The pins run in that configuration deliberately, which is the right call — a test in the composed configuration would have passed either way and proved nothing.

**Ruling 4 — nothing newly disclosed beyond the author's own subtree: VERIFIED, and the `'policy'` marks matter more than "no behaviour change" suggests.** An unmarked node inherits its ancestor's mark **positionally** (innermost wins), so an unmarked policy arm nested inside a vouched subtree would read as the author's. Marking them converts an accidental withhold into a declared one. The sharpest pin on the card is exactly that case: *a policy arm sitting beside an author-vouched arm in the same `$and` still withholds*.

**Ruling 1 — fail-closed: VERIFIED and pinned in every direction**, including a predicate no boundary vouched withholding **byte-identically** to the policy case, and a `where` a sibling already rewrote receiving **no** author mark. The vouch is an identity check rather than a heuristic: stamped only while `ast.where` is still, by object identity, what the caller handed the engine. A rewritten sibling tree is vouched for nobody, which is the discipline the card asked for.

**Ruling 3 — write path untouched: VERIFIED from the diff.** The only `composeAnd` occurrences in the change are in comments; no `writeFilter` line moved. The finding goes in the report rather than the PR, correctly.

**The mid-flight constraint held.** I narrowed this card's surface after dispatch, when #8735's recovered work turned out to reach into `plugin-sharing/src/translations/**`. The diff touches no translation file, and `check:i18n` was run to confirm the bundles were already in sync so **no regeneration was written** — which is the specific failure I was guarding against.

**Reverse verification predicted the direction and got it:** 4 of 9 red — both author cases and both mark-setting cases — while the three fail-closed cases and the composed-with-security case stayed green. The greens are the point: the policy and unmarked halves are behaviour-preserving by construction, so insensitivity there is the correct result rather than a weak test.

`check:i18n` initially **refused to run** and was reported as NOT MEASURED, then the prerequisite was built and it was re-run to a real reading. Every package consuming the provenance mechanism was tested — `plugin-security`, both SQL drivers, `service-analytics`, `spec`, `runtime` — which is the affected-subset gap handled before queueing rather than after a kick.

⛔ **Flip held** until every gate job concludes `success` on its own reading. I will record the mechanism correction on #8430 itself, since the card's stated cause will otherwise mislead the next reader.

---

PR #8799, opened 2026-08-15
fix(service-analytics): single-source the comparand-type membership and refusal sentence to the #7872 door

### LINKED ISSUE #8186: [finding] service-analytics carries its own copies of the comparand-type allow-list the #7872 door now single-sources — reconcile membership and message wording to the door

qq9340100:
Fixes #8186

`service-analytics` carried its own copies of the comparand-type allow-list that #7872's shared door single-sources for the SQL family. `comparand-shape.ts` spelled the six accepted types twice — once in `isBindableComparand`, once in `isRenderableTextComparand` — and two refusal messages hand-copied the accepted-set sentence that `driver-sql` and `driver-turso` now quote from `ACCEPTED_FILTER_COMPARAND_TYPES_SENTENCE`.

Both predicates now delegate the TYPE membership to `isAcceptedFilterComparand` (`@objectstack/spec/data`) and both messages quote the door's sentence, in the same shape #7872 gave `driver-sql`'s twins. The envelopes and the position logic stay this package's own, per the card's boundary: a caller-authored `where` still refuses with a 400 `INVALID_FILTER`, a read scope still fails closed with a 500.

## This is a drift-risk reconciliation, and the no-behaviour-change claim is measured

The card's premise held: the copies agreed with the door cell for cell, so the reconciliation had to move nothing. That is shown rather than asserted.

`src/__tests__/comparand-door-single-source.test.ts` freezes the accept/refuse matrix across the comparand type matrix, driving **both doors end to end** in three comparand positions each (`$contains`, an `$in` member, `$eq`) and asserting the ADR-0112 envelope — `code` and `status` — not merely that something threw. It was measured on `origin/main` @ `84cb121eb`, committed **before** any source change (commit `9c6f545`), and re-run **byte-identical** afterwards. Green both times; no cell moved.

Predicate-level agreement cannot be proved by a matrix that both spellings satisfy, so the import was verified by ablating the **door**: `bigint` removed from `isAcceptedFilterComparand`, `packages/spec` rebuilt, the marker confirmed present in `dist/`, then absent again after restore.

Predicted: the `bigint` row goes red at both predicates and at the LIKE / `$in` positions on both doors, while `$eq` stays green (that account is not gated by these predicates). Measured: exactly that — 4 failures, all `bigint`, 35 assertions unaffected. Had the import been cosmetic, all 39 would have stayed green. The door source was restored byte-identical (`git status` clean) and `packages/spec` rebuilt, so no ablated artifact survives in the tree.

## The `undefined` delta — kept, and the card's description of it corrected

The card and the dispatch both describe the analytics `where` door as normalising `undefined` to `null` **rather than refusing it**. Measured, that is no longer true, and the correction is recorded rather than acted on:

| door | `undefined` comparand today | refused by |
|---|---|---|
| analytics `where` | REFUSED, `INVALID_FILTER` / 400 | `assertDefinedComparands` (#6386, on #6050 ruling B) |
| read scope | REFUSED, `READ_SCOPE_COMPILE_FAILED` / 500 | #6050 ruling B pushed down by #6125 |

So the delta is not a live tolerance — it is a pair of deliberately-kept dead arms: `comparand()`'s normalise-to-`null` (whose own TSDoc says reopening it is #5526's call, not a cleanup's) and the `undefined` branch inside each predicate. Both are **kept, untouched**, exactly as `driver-sql` keeps its own unreachable `undefined` arm, and both are now recorded at the use site with their real reachability. The `undefined` row in the new matrix pins REFUSAL at both doors, in all three positions, which is what makes that claim checkable instead of asserted.

## One wording correction falls out of deleting the copy

The hand-copied sentence read "a string, number, boolean, null, Date or binary value" — it omitted `bigint`, which both predicates have always accepted and both doors have always compiled. The copy had drifted from its source in the quiet direction: a refusal message under-describing the values it accepts. Quoting the door corrects the wording; the accepted set itself does not move, which the matrix pins. No test anywhere in the repo pinned the old wording (checked repo-wide).

The mirror in `like-metacharacter-escape.test.ts` was the third and last hand-copy of the same six types. It now mirrors `driver-sql`'s post-#7872 spelling, so what it holds is the part the door does not carry — each face's local extras — which is the only part that can still drift.

## Verification

All tests and the full gate union run at `b25a8ab` (final commit).

- `pnpm --filter @objectstack/service-analytics test` — 76 files, **1714 passed** (baseline 1675 + 39 new)
- Consumer sweep, every package consuming the door, because the merge queue runs the full suite while the PR runs only the affected subset: `@objectstack/spec` 10638 passed, `driver-sql` 1663 passed / 55 skipped, `driver-turso` 986 passed, `driver-sqlite-wasm` 393 passed, `objectql` 3643 passed. Zero failures.
- Gate families re-derived with `scripts/pm/dispatch-gates.mjs` against the actual changed paths: `check:nul-bytes`, `check:test-source-alias`, `check:type-source-resolution`, `check:query-options-erasure`, `check:type-check-coverage` — all pass.
- `check:type-check-debt --re-measure` — OK, 33 ledger entries re-measured in 308.9s, 1926 raw tsc errors total, **none above its recorded number**.
- `service-analytics` declares no `typecheck` script; it carries a shrink-only DEBT ledger entry (`errors: 10`), and the `--re-measure` ratchet above is what covers it — it did not rise.

---

### DIFF
--- a/.changeset/lucky-pugs-repeat.md
+++ b/.changeset/lucky-pugs-repeat.md
@@ -1,0 +1,11 @@
+---
+'@objectstack/service-analytics': patch
+---
+
+Source the comparand-type allow-list and the accepted-set refusal sentence from the shared `@objectstack/spec/data` door instead of re-spelling them locally.
+
+`comparand-shape.ts`'s `isBindableComparand` / `isRenderableTextComparand` spelled the same six accepted comparand types (`string | number | bigint | boolean | null | Date`) that `isAcceptedFilterComparand` single-sources for the SQL driver family, and two refusal messages hand-copied the accepted-set sentence. Both predicates now delegate the type membership to the door and quote `ACCEPTED_FILTER_COMPARAND_TYPES_SENTENCE`, matching how `driver-sql` and `driver-turso` consume it.
+
+No comparand is accepted or refused differently: the local copies already agreed with the door, and the full accept/refuse matrix is pinned end to end at both analytics filter doors, in three comparand positions each, measured before the change and re-run unchanged after it.
+
+One user-visible wording correction falls out of removing the copy: the hand-copied sentence omitted `bigint`, a type both predicates have always accepted and both doors have always compiled, so a refusal message under-described the values it accepts. The message now names the full set. The package-local extras — a binary bindable, and the `undefined` arm both doors already refuse upstream — are unchanged and recorded at their use sites.
--- a/packages/services/service-analytics/src/__tests__/comparand-door-single-source.test.ts
+++ b/packages/services/service-analytics/src/__tests__/comparand-door-single-source.test.ts
@@ -1,0 +1,201 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#8186] `service-analytics` sources its comparand-TYPE membership from the
+ * shared door (#7872) instead of spelling it a third time — and this file is the
+ * evidence that doing so changed no verdict.
+ *
+ * ## What this file is FOR
+ *
+ * #8186 is a drift-risk reconciliation, not a defect: `comparand-shape.ts`'s two
+ * predicates had reached the SAME six types as `isAcceptedFilterComparand`
+ * (`@objectstack/spec/data`) independently, so the reconciliation had to be
+ * provably verdict-neutral rather than merely plausible. The table below was
+ * MEASURED on `origin/main` @ `84cb121eb` (before the predicates were rewired),
+ * frozen here, and re-run unchanged afterwards. Both runs are green, which is
+ * the whole claim: the membership moved, the behaviour did not.
+ *
+ * That is also why the matrix drives the two DOORS end to end and not only the
+ * predicates. A predicate-level assertion would have proved the functions agree
+ * with themselves; what a caller experiences is the door's verdict, envelope
+ * included, and those are the cells a silent regression would move.
+ *
+ * ## The three positions, per door
+ *
+ * A comparand's legality is positional, so one value is asked three questions:
+ * the LIKE family (needs a faithful TEXT rendering), an `$in` MEMBER (needs to
+ * BIND), and a scalar `$eq` (the #5234 account deliberately left alone). The two
+ * doors answer in different envelopes on purpose — a caller-authored `where` is
+ * a 400 `INVALID_FILTER`, a read scope fails closed with a 500 (ADR-0021 D-C) —
+ * so the envelope is asserted, not just the throw.
+ *
+ * ## The two rows that are NOT the door's set, and stay that way
+ *
+ * `undefined` and binary are `service-analytics`-local admissions at the
+ * predicate level, exactly as they are `driver-sql`-local there. Neither is
+ * reachable as an ACCEPTED comparand through either door:
+ *
+ *   - **`undefined` is REFUSED by both doors** — `assertDefinedComparands`
+ *     (#6386, on #6050's ruling B) at the `where` door and #6125's upstream
+ *     refusal at the read scope, both of which fire before a predicate is
+ *     consulted. The `undefined` arms inside the predicates, and `comparand()`'s
+ *     normalise-to-`null`, survive as deliberately-kept dead arms (#5526's call
+ *     to reopen, not this card's). The `undefined` row below is what makes that
+ *     checkable instead of asserted — it pins REFUSAL, at both doors.
+ *   - **binary** binds but has no faithful text rendering, so it is accepted in
+ *     a bind position and refused by the LIKE family. That asymmetry is the
+ *     reason the package carries two predicates rather than one with a flag.
+ *
+ * @see comparand-shape.ts — the predicates and the messages this pins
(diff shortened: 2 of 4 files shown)


qq9340100:
## ACCEPT — PM review, `domain:services` seat

**Ruling 2's premise was falsified, and handled exactly right.** The card — and my dispatch, which inherited it — both describe the analytics `where` door as normalising `undefined` to `null` *rather than refusing it*, and I called that a "recorded deliberate difference" the dev must not flatten. Measured, it is no longer true: `undefined` is **refused** at both doors today (`INVALID_FILTER`/400 via `assertDefinedComparands` under #6386, and `READ_SCOPE_COMPILE_FAILED`/500 via #6125). So the "delta" is not a live tolerance at all — it is a pair of deliberately-kept **dead arms**.

The response to that is the correct one on every axis: the arms are **kept untouched** (matching how `driver-sql` keeps its own unreachable `undefined` arm), their real reachability is now recorded at the use site, and the `undefined` row in the new matrix pins **refusal** at both doors in all three positions — which turns the claim from prose into something checkable. Reopening `comparand()`'s normalise-to-`null` stays #5526's call, per its own TSDoc.

**The reverse verification is the sharpest thing here.** The dev identified that a matrix both spellings satisfy *cannot* prove the import is load-bearing, and ablated the **door** instead — removed `bigint` from `isAcceptedFilterComparand`, rebuilt `packages/spec`, confirmed the marker in `dist/`, and measured **4 failures, all `bigint`**, with 35 assertions unaffected and `$eq` correctly staying green because that account is not gated by these predicates. Had the import been cosmetic, all 39 would have stayed green. Door restored byte-identical and `spec` rebuilt, so no ablated artifact survives.

**The no-behaviour-change claim is evidenced in the strongest available form:** the matrix was committed *before* any source change (`9c6f545`), measured green on `origin/main`, and re-run **byte-identical** afterwards. That ordering is what makes it evidence rather than assertion.

**Verified from the diff myself:** both predicates now delegate membership to `isAcceptedFilterComparand` and both messages quote `ACCEPTED_FILTER_COMPARAND_TYPES_SENTENCE`. No local type list survives. The envelopes and position logic stayed this package's own, per the card's boundary — 400 `INVALID_FILTER` for a caller `where`, 500 fail-closed for a read scope.

**The `bigint` wording correction is a genuine find, and it drifted in the dangerous-because-quiet direction:** the hand-copied sentence read "a string, number, boolean, null, Date or binary value" while both predicates have always *accepted* `bigint` — a refusal message under-describing what it takes. Quoting the door fixes the text; the accepted set does not move, which the matrix pins, and no test anywhere pinned the old wording (checked repo-wide). That is the exact failure mode the card was filed about, caught in the copy it was filed against.

Catching that the `like-metacharacter-escape.test.ts` mirror was the **third** hand-copy, and reshaping it to hold only each face's local extras — the part the door cannot carry and therefore the only part that can still drift — is the right instinct about where a pin earns its keep.

**One honest gap, correctly disclosed:** `check:type-check-debt --re-measure` was measured at `8261cc1` rather than final head `b25a8ab`, the two differing only by the added `.changeset/*.md`, with the re-run queued behind the shared build lock. A changeset file cannot move a type-check ratchet, so I am not holding on it — and CI runs it at head regardless. Disclosed rather than quietly quoted at head, which is what matters.

---

PR #8801, opened 2026-08-15
feat(security): bind the last-admin standing-key lists to the authz resolver's measured read surface (#8734)

### LINKED ISSUE #8734: Nothing binds the last-admin-guard standing-key lists to what the authz resolver actually reads — the correspondence is prose only
Filed as a `finding` (recording only, unassigned, ungraded) out of the review of PR #8713 / #8613. Not fixed there deliberately — it is the structural gap that card *exposed*, not the card's own work.

## The observation

`packages/plugins/plugin-auth/src/last-admin-guard.ts` decides whether to pay for a full "who is still an administrator" enumeration by testing the pending write against three standing-key lists, one of which is:

```js
const PERMISSION_SET_STANDING_KEYS = ['name', 'active'] as const;   // was ['name'] before #8613
```

That list is not an independent design artifact. It is **a cache of which columns `resolveAuthzContext` reads when deriving `platform_admin`**. Its correctness is entirely derivative: if the resolver starts reading a column the list omits, the guard silently stops firing on the one write class that can empty the administrator population.

**Nothing mechanical enforces that correspondence.** It is maintained by a prose comment and by whoever next edits the resolver happening to remember the guard exists in another package.

## Why this is worth a card rather than a comment

It has already failed once, and the failure mode was the dangerous direction — a *false* assertion left standing in the code:

- #6084 wrote, in the comment beside the list, that everything a permission-set write touches other than `name` — **explicitly naming `active`** — is invisible to "who is an administrator".
- That was true when written.
- #8613 made `active` a resolution-time predicate, which made the sentence false and opened a one-click, in-product-unrecoverable path to an installation-wide administrator lockout.
- It was caught only because the implementing agent read the comment closely enough to notice it now contradicted the code it was writing. Nothing would have caught it otherwise: the guard's own tests would still pass, because the guard would simply never be consulted for that write.

An authoritative, recently-dated, wrong comment in a security guard is the specific hazard here — it reads as a checked fact to the next author, human or agent.

## Shape of a fix (sketch, not a prescription)

The gate wants to answer: *does the set of columns the platform-admin derivation reads equal the set the guard treats as standing-bearing?* Candidate approaches, in rough order of cost:

1. A check script that extracts the column literals read along `resolveAuthzContext`'s platform-admin path (`packages/core/src/security/resolve-authz-context.ts`) and the standing-key lists (`last-admin-guard.ts`), and fails on asymmetry — the same family as the existing `check:authz-resolver`.
2. A shared, single-source declaration both sides import, so the two cannot spell it differently.
3. Failing either, a test that asserts the correspondence explicitly, so the pin lives somewhere a resolver change will break.

Whichever shape, the property to hold is that **a resolver change that starts reading a new column cannot land green while the guard's list omits it.**

Note the same reasoning applies to the sibling lists in that file (`GRANT_STANDING_KEYS` and the membership/identity-table equivalents), which have the same derivative relationship to their readers.

## Not in scope

Not asserting the current lists are wrong — after #8613 they are believed correct. This is about the absence of anything that would tell us if they stopped being correct.

qq9340100:
Fixes #8734

`plugin-auth`'s last-administrator guard (ADR-0024 D5.2) decides whether a pending write can empty the administrator population by testing the payload against three standing-key lists. A payload touching none of them is skipped **without any reads** — so a column `resolveAuthzContext` starts reading that a list omits is a write class the guard silently stops judging, on the one path whose failure mode is an installation-wide administrator lockout with no in-product recovery.

Nothing bound the two together. The correspondence lived in a comment, and it had already gone false once: #6084 wrote — naming `active` explicitly — that everything a permission-set write touches other than `name` is invisible to "who is an administrator". True when written; #8613 made `active` a resolution-time predicate and the sentence became false. Nothing mechanical would have caught it, because the guard's own tests stay green **precisely when the guard is never consulted**.

## Which shape, and why

The card sketched three. Measured, then chosen:

- **Shape 2 (a single-source declaration both sides import) cannot carry the property on its own here.** `resolveAuthzContext` takes `ql: any`, `tryFind` returns whole rows, and every column read is an untyped property access. No shared constant can force the resolver to consume it — a new `row.newColumn` compiles and runs whatever the constant says. The dependency direction does exist (`plugin-auth` already imports `isGrantActive` / `isRowActive` from `@objectstack/core`), so shape 2 is *reachable*; it is just not *sufficient*.
- **Shape 3 (a correspondence test) pins today's values unless it derives one side.** PM assumption 4 is the right test of it: a fix that only pins today's values has not addressed the card.
- **Shape 1 (a `check:` script extracting column literals) would have to inline the helpers to work.** `active` is never named at the resolver's own call site — `isRowActive(r)` reads it — and neither are the ADR-0091 bounds (`isGrantActive(row, now)`). A static extractor reading the caller and not the callee misses #8613 for the same reason the comment did. It would also need real dataflow analysis to learn that `ps.name` in a `for` over `psRows` is a read of `sys_permission_set.name`.

**What landed is shape 2 made enforceable by runtime observation**, in two links:

1. **`@objectstack/core` declares `ADMIN_STANDING_SURFACE` beside the resolver** — every table the administrator-derivation path reads, each classified `derives` or `reads-only` with its reason, and for the deriving tables every column read. It is asserted **equal** to what the real `resolveAuthzContext` reads, observed by driving it over a recording engine that records every property access and every `where` key per table. The declaration is a measurement, not a wish.
2. **`plugin-auth` consumes that measurement** and requires every column to have an answer: standing-bearing (in a list) or excluded with the reason it cannot empty the administrator population. There is no third state — the third state is exactly what `active` was between #6084 and #8613.

Observation also fixes the discoverability half the card names ("whoever next edits the resolver happening to remember the guard exists in another package"): link 1 lives in `packages/core` and fails in the resolver author's own `pnpm test`, on source, with no build.

## The gate can fail — demonstrated, four ways

Each leg was run from a committed state and restored; the direction was predicted before running.

| Leg | Mutation | Result |
|---|---|---|
| A | remove `active` from `PERMISSION_SET_STANDING_KEYS` (the #8613 regression, inverted) | plugin-auth RED: ``expected [ 'active' ] to deeply equal []`` |
| B | add a new predicate read `ps.locked_out` on the platform-admin path | core RED: ``expected [ 'active', 'id', 'locked_out', …(5) ] to deeply equal [ 'active', 'id', 'name', …(4) ]`` |
| B2 | then declare `locked_out` in core (the natural next step) | core green; plugin-auth RED: ``expected [ 'locked_out' ] to deeply equal []`` |
| C | resolver reads a **new table** `sys_admin_delegation` | core RED on the table set; after classifying it `derives`, plugin-auth RED: ``expected [ 'sys_member', …(2) ] to include 'sys_admin_delegation'`` |

Leg B also demonstrated the reverse direction on its way out: a column left declared after its read was removed goes red as a stale declaration.

## Would this have caught #8613?

**Yes, at the earlier link.** #8613 added `isRowActive(r)` to the `sys_permission_set` read, which makes `active` observable — link 1 goes red on that PR, naming the column. Declaring it turns link 2 red, because `active` was in neither the list nor a ledger. Landing #8613 green would have required writing down, explicitly, that deactivating `admin_full_access` cannot empty the administrator population — which is false, and which is what the old comment asserted by accident. Leg A above is that scenario replayed.

## Sibling lists: covered, all three

`MEMBER_STANDING_KEYS`, `GRANT_STANDING_KEYS` and `PERMISSION_SET_STANDING_KEYS` are all keyed into `STANDING_KEYS_BY_TABLE` and all three are checked. Keying by **table** is what also closes the table-level hole: a resolver deriving standing from a new table is invisible to any column-set comparison, since the table is absent from both sides (leg C).

## Ruling 2: the guard is not narrowed, and cannot be by this gate

Every list keeps exactly the values it had; `touchesAny` is called with the same constants. The gate is **one-directional by construction** — it never asserts that a standing key is a column the resolver reads. That direction looks like the natural other half and is not: `GRANT_STANDING_KEYS` carries `userId` although the resolver reaches those rows through a `where` on `user_id` and never touches the camelCase spelling on the row. Enforcing it would pressure a break-glass guard to drop entries. This gate can only ever demand that the guard judges **more**.

## Ruling 4: nothing asymmetric in the lists, one dormant finding next door

The lists are correct today — every column the resolver reads on the three deriving tables is either judged or has a stated reason it cannot matter, and no reason required inventing.

One thing surfaced that is **not** about the lists and was not touched here: in `resolveUserAuthzGrants`, `isGrantActive` is applied to `sys_member` rows only when building `accessible_org_ids`; the org-administration `role` projection that feeds `positions` is not window-filtered. Dormant — `sys_member` declares no `valid_from` / `valid_until` today — and it is a resolver question, not a guard one. Recorded separately as a `finding` in #8802, and captured in `STANDING_KEY_EXCLUSIONS` as the reason those bounds are not standing-bearing, with a note that the resolver leads and the list follows.

## Verification

- `pnpm --filter @objectstack/core --filter @objectstack/plugin-auth test` — core 838 passed (35 files), plugin-auth 1238 passed (54 files)
- `pnpm --filter @objectstack/plugin-auth typecheck` — clean
- Gate families re-derived with `scripts/pm/dispatch-gates.mjs` against the actual changed paths and run: `check:changeset-gate-self-tests`, `check:cross-package-test-inputs`, `check:kernel-hook-pairs`, `check:objectui-changeset`, `check:test-source-alias`, `check:type-source-resolution`, `check:nul-bytes`, `check:query-options-erasure`, `check:type-check-coverage`, plus `check-adr-0087-registration`, `check-changeset-no-major`, `check-empty-changeset` — all pass.
- `check:type-check-debt` (the `--re-measure` ratchet) — it refuses outright on an unbuilt worktree, so the full closure was built first (`turbo run build --filter='./packages/*' --filter='./packages/*/*'`, 70/70 successful) and then it ran: *33 ledger entries re-measured, 1926 raw tsc errors, none above its recorded number*. The one informational surplus it reports (`@objectstack/lint`, -1) is pre-existing and in a package this diff does not touch.

Verified at ``5ae537580``.

---

### DIFF
--- a/.changeset/last-admin-standing-keys-gate.md
+++ b/.changeset/last-admin-standing-keys-gate.md
@@ -1,0 +1,59 @@
+---
+"@objectstack/core": minor
+"@objectstack/plugin-auth": minor
+---
+
+feat(security): bind the break-glass standing-key lists to what the authz resolver actually reads — the correspondence stops being prose (#8734)
+
+`plugin-auth`'s last-administrator guard (ADR-0024 D5.2) decides whether a
+pending write can empty the administrator population by testing the payload
+against three standing-key lists (`MEMBER_STANDING_KEYS`,
+`GRANT_STANDING_KEYS`, `PERMISSION_SET_STANDING_KEYS`). A payload touching none
+of them is skipped without any reads — so a column `resolveAuthzContext` starts
+reading that a list omits is a write class the guard **silently stops judging**,
+on the one path whose failure mode is an installation-wide administrator lockout
+with no in-product recovery.
+
+Nothing bound the two together. The correspondence lived in a comment, and it
+had already gone false once: #6084 wrote — naming `active` explicitly — that
+everything a permission-set write touches other than `name` is invisible to "who
+is an administrator". That was true when written; #8613 made `active` a
+resolution-time predicate and the sentence became false. Nothing mechanical
+would have caught it, because the guard's own tests stay green precisely when
+the guard is never consulted.
+
+**The mechanism is two links, and the first one is a measurement.**
+
+- `@objectstack/core` now exports `ADMIN_STANDING_SURFACE` — declared beside the
+  resolver, listing every table the administrator-derivation path reads, each
+  classified `derives` or `reads-only` with its reason, and for the deriving
+  tables every column read. It is asserted **equal** to what the real
+  `resolveAuthzContext` reads, observed at runtime through a recording engine
+  that records every property access and every `where` key per table. Observation
+  rather than source extraction because the reads that matter have moved into
+  helpers: `active` is read by `isRowActive(row)` and the ADR-0091 window bounds
+  by `isGrantActive(row, now)`, neither named at the resolver's own call site —
+  the exact shape #8613 had.
+
+- `@objectstack/plugin-auth` now exports its standing-key lists plus
+  `STANDING_KEYS_BY_TABLE` and `STANDING_KEY_EXCLUSIONS`, and a gate requires
+  every column of that measured surface to have an answer: it is standing-bearing
+  (in a list) or it is excluded with the reason it cannot empty the administrator
+  population. There is no third state — the third state is what `active` was
+  between #6084 and #8613.
+
+So a resolver change that starts reading a new column fails at the first link
+until the declaration is updated, and at the second until the guard has an
+explicit answer for it. Landing #8613 green would have required writing down that
+deactivating `admin_full_access` cannot empty the administrator population —
+which is false, and which is what the old comment asserted by accident.
+
+**No guard behaviour changes.** Every list keeps exactly the values it had; the
+gate is one-directional by construction (it can only ever demand that the guard
+judges *more*), because the other direction would put pressure on a break-glass
+guard to fire less often.
+
+The table-level half is covered too: a resolver that started deriving
+administrator standing from a **new** table is invisible to any column-set
+comparison, since the table is absent from both sides — so the surface enumerates
+every table the path reads, and an unclassified one fails.
--- a/packages/core/src/security/admin-standing-surface.test.ts
+++ b/packages/core/src/security/admin-standing-surface.test.ts
@@ -1,0 +1,324 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#8734] The FIRST of the two links that bind `plugin-auth`'s break-glass
+ * standing-key lists to what this resolver actually reads.
+ *
+ * This half answers one question mechanically: **which columns does
+ * `resolveAuthzContext` read on the tables administrator standing is derived
+ * from?** It answers it by OBSERVATION — the real resolver is driven over a
+ * recording engine that returns `Proxy`-wrapped rows and records every property
+ * access, plus every `where` key, per table — and asserts the answer equals
+ * `ADMIN_STANDING_SURFACE`.
+ *
+ * The second link lives in `plugin-auth`
+ * (`last-admin-standing-keys.test.ts`): it requires every column declared here
+ * to be either in a standing-key list or explicitly excluded with a reason. So
+ * a resolver change that starts reading a new column fails HERE until the
+ * declaration is updated, and fails THERE until the guard has an answer for it.
+ *
+ * ## Why observation rather than a static extractor
+ *
+ * A source-parsing gate would have to follow `psRowsAll` into `psRows` into the
+ * `for (const ps of psRows)` loop to learn that `ps.name` is a read of
+ * `sys_permission_set.name` — real dataflow analysis, brittle in exactly the
+ * places that matter. Worse, it would have to inline the helpers: `active` is
+ * never named on the resolver's own call site (`isRowActive(r)` reads it), and
+ * neither is `valid_from` (`isGrantActive(row, now)` reads it). #8613's whole
+ * defect was a read that moved into a predicate; a gate that reads the caller
+ * and not the callee would have missed it for the same reason the comment did.
+ *
+ * ## Why the fixtures come in variants
+ *
+ * A conditional read is invisible in a fixture that never takes the branch. The
+ * resolver's tolerated-spelling chains are the sharp case: `r.organization_id ??
+ * r.organizationId` never touches the camelCase spelling while the snake_case
+ * one is non-nullish. So the observation is the UNION over fixtures chosen to
+ * take both sides of every such chain — snake-only rows, camel-only rows, and a
+ * pass where the flags and windows are set the other way. `assertVariantsStay-
+ * Distinct` keeps that honest: if two variants ever observe the same set, one of
+ * them has stopped contributing and the union has silently narrowed.
+ */
+
+import { describe, it, expect } from 'vitest';
+
+import { ADMIN_STANDING_SURFACE, adminStandingTables } from './admin-standing-surface.js';
+import { resolveAuthzContext } from './resolve-authz-context.js';
+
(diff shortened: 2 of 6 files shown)


qq9340100:

The strongest card of the round, and it beat its own brief.

**Ruling 1 — the property, not a particular mechanism: satisfied, and all three sketched shapes were rejected on measurement rather than taste.** The decisive finding is about **shape 1**, the extraction script I thought was probably cheapest: `active` is **never named at the resolver's own call site** — `isRowActive(r)` reads it, and the ADR-0091 bounds hide the same way inside `isGrantActive(row, now)`. A static extractor reading the caller and not the callee **would have missed #8613 for exactly the same reason the comment did**. That is the card's own failure mode reproduced inside the proposed fix, and catching it before building is worth more than the build.

**Shape 2 was reachable but insufficient**, also measured: `resolveAuthzContext` takes `ql: any` and every column read is an untyped property access, so no shared constant can *force* the resolver to consume it — `row.newColumn` compiles and runs whatever the constant says. What landed is **shape 2 made enforceable by runtime observation**: `packages/core` declares `ADMIN_STANDING_SURFACE` and asserts it **equal to what the real resolver reads**, observed by driving it over a recording engine that captures every property access and every `where` key per table. The declaration is a measurement, not a wish. I did not think of this and would have accepted something weaker.

**Ruling 3 — the gate must be demonstrated to fail: four legs, each from a committed state, direction predicted first.** Leg A replays #8613 inverted; leg B adds a new predicate read; **leg B2 is the one that proves the design** — declaring the new column in core turns core green and then turns `plugin-auth` red, which is the two-link property working rather than one check passing. Leg C covers a new **table**, the hole no column-set comparison can see because the table is absent from both sides. Leg B also showed the reverse: a declaration left behind after its read is removed goes red as stale.

**"Would this have caught #8613?" — answered, not asserted.** Yes, at the earlier link, and the reasoning is the right one: landing #8613 green would have required explicitly writing down that deactivating `admin_full_access` cannot empty the administrator population — which is false, and which the old comment asserted *by accident*. Forcing the false claim into the open is precisely what a prose comment could never do.

**Ruling 2 — guard not narrowed: VERIFIED from the diff.** Every list keeps its exact values; the only change to those lines is `export`. Better, the gate is **one-directional by construction** — it never asserts a standing key *is* a column the resolver reads, and the reason is concrete: `GRANT_STANDING_KEYS` carries `userId` while the resolver reaches those rows through a `where` on `user_id` and never touches the camelCase spelling. Enforcing that direction would **pressure a break-glass guard to drop entries**. This gate can only ever demand the guard judges *more*. That is the correct asymmetry for a safety guard and it was reasoned out rather than stumbled into.

**Ruling 4 — lists correct today, and the one thing found next door was reported, not fixed.** `isGrantActive` is applied to `sys_member` only when building `accessible_org_ids`; the org-administration `role` projection feeding `positions` is unfiltered. Dormant (no `valid_from`/`valid_until` on `sys_member` today), a resolver question rather than a guard one, filed as **#8802** — verified present with `finding` + `domain:identity`. Recording it in `STANDING_KEY_EXCLUSIONS` as the stated reason those bounds are not standing-bearing, with a note that the resolver leads and the list follows, is exactly the right place for it.

**It also solved the discoverability half I did not rule on.** The card's real complaint was "whoever next edits the resolver happening to remember the guard exists in another package". Link 1 lives in `packages/core` and fails in the resolver author's **own `pnpm test`, on source, with no build** — so the person who breaks it is the person who sees it.

**All three sibling lists covered**, keyed by table, which is what closes the table-level hole rather than only the column-level one.

`check:type-check-debt` refused on an unbuilt worktree and was reported as such, then the full closure was built (70/70) and it ran to a real reading.

---

Issue #8692, opened 2026-08-14
[finding] `os meta resync` may silently skip every platform default permission set — the seeder never writes `managed_by`, and the tests stub the field default away


qq9340100:
Found while measuring #8613 (which touches nothing here — that card is about `active` enforcement). Filed unassigned, observation-class: the drift is real and unexercised, but the intended provenance for these rows is genuinely ambiguous in the source, so this needs a reading before it needs a fix.

## The shape

`packages/plugins/plugin-security/src/bootstrap-platform-admin.ts` inserts the seed permission-set rows without a `managed_by` column:

```ts
const created = await tryInsert(ql, 'sys_permission_set', {
  id,
  name: ps.name,
  ...platformOwnedFields(ps),
  active: true,
});
```

`platformOwnedFields()` returns only `label` / `description` / the four permission JSON blobs / `admin_scope`. So `managed_by` falls to the field default declared on the object — `sys-permission-set.object.ts` line 241:

```ts
managed_by: Field.select({ ..., defaultValue: 'admin', ... })
```

The `#2705` resync branch in the same seeder then reconciles a row only when it is still platform-owned:

```ts
if (!row.managed_by || row.managed_by === 'platform') { /* resync */ }
else { resyncSkipped += 1; logger?.warn?.(`resync left ${ps.name} untouched — row is ${row.managed_by}-owned (intentional override)`); }
```

If the engine applies the declared `defaultValue` on insert, every row this seeder creates comes back as `managed_by: 'admin'`, so on the next `os meta resync` the platform's own default sets take the `else` branch: `resynced` stays 0, `resyncSkipped` counts all of them, and the log claims an "intentional override" for rows no admin ever touched. That is the exact opposite of what #2705 built the flag for ("reconcile the row to the shipped dist so a dev source edit takes effect without `--fresh`").

## Why it is unexercised rather than simply broken

`bootstrap-platform-admin.test.ts` never lets the default apply — it hand-seeds the pre-image and states the provenance explicitly:

- line 72: `{ id: 'ps_old', name: 'member_default', ..., managed_by: null }` then asserts `resynced` is 1
- line 86: `managed_by: 'user'` then asserts `resyncSkipped` is 1
- line 96: `managed_by: 'package'` then asserts `resyncSkipped` is 1

[...]

PR #8804, opened 2026-08-15
test(plugin-security): measure the provenance a real seeder run leaves on the platform default permission sets

qq9340100:
Part of #8692

This card asked for a **reading, not a fix** — in its own words, *"this needs a reading before it needs a fix."* So this PR adds one test file and changes **no behaviour**: no field default touched, no resync condition altered, no log line "corrected", no `managed_by` added to the seeder's insert.

## The measurement

Seeded through the real shipped `defaultPermissionSets` against a real `ObjectQL` engine over a real better-sqlite3 `SqlDriver` — the same wiring `security-plugin.ts` and `os meta resync` both use, each handing the bare engine straight to `bootstrapPlatformAdmin`.

| Reading | Value |
|:---|:---|
| `managed_by` on the seeder-created `admin_full_access` row (stored, read at the driver) | `"admin"` |
| the same row through the engine's own `find` (what the resync branch reads) | `"admin"` |
| `resynced` after a real resync of all 8 shipped default sets | `0` |
| `resyncSkipped` | `8` (every one of them) |
| warn lines emitted | 8, e.g. `[security] resync left admin_full_access untouched — row is admin-owned (intentional override)` |

The card's mechanism reproduces **exactly**. The engine applies the object's declared `defaultValue: 'admin'` on insert, so the `!row.managed_by` limb of the resync condition is unreachable for every row this seeder creates, and `os meta resync` reconciles nothing while reporting an "intentional override" for rows no admin ever touched.

## Which link supplies the value — proven, not inferred

A counterfactual in the test boots a second engine on a **clone** of the declaration whose `managed_by` default is respelled to `'platform'`, then runs the same unmodified seeder over the same unmodified sets. Predicted direction, written down before running: both counts should flip. They do — the row stores `'platform'`, `resynced` becomes 8, `resyncSkipped` becomes 0.

That single flip establishes all three links at once, which no direct assertion can:

- the seeder writes no `managed_by` of its own (it would otherwise beat the clone's default);
- the engine really does apply a declared `defaultValue` on **insert** — the card's one stated unknown;
- the `!row.managed_by || row.managed_by === 'platform'` limb is live and working, so the skip is caused by the stored **value** alone, not by a broken condition.

The clone is a test-local object. The shipped declaration is asserted untouched in the same case.

## What is deliberately NOT decided here

The upstream product question the card identifies stays open and is a maintainer call:

> are the platform default permission sets meant to be platform-owned (and therefore resyncable), or env-authored (and therefore deliberately left alone)?

The source still argues both ways — the resync condition's `!row.managed_by ||` half reads as though NULL was expected, while the comment directly above the insert calls the posture one that "keeps the platform defaults env-authored". Both cannot be true, and the measurement above does not adjudicate between them; it only establishes which branch the code actually takes. The test header says so and points back at this card, so a future reader cannot mistake a recording for an endorsement.

Because the reading is only half the card, this PR is `Part of` rather than a closing PR: #8692 remains open for that decision.

## The three existing cases are untouched

`bootstrap-platform-admin.test.ts` hand-seeds `managed_by` as `null` / `'user'` / `'package'` and asserts the resync branch three ways. Those are correct unit coverage and are unchanged. The gap they leave is that a fake `ql` stores exactly the columns it is handed, so a declared field default never runs — meaning the one row shape the platform actually creates was the one shape nothing exercised. This file adds that end-to-end case rather than replacing any of them.

## Verification

All at `6bd664ea8`, the head this PR proposes:

- `pnpm --filter @objectstack/plugin-security test` — **63 files, 1195 tests passed** (5 new)
- `pnpm --filter @objectstack/plugin-security typecheck` — clean
- gates re-derived against the actual changed path with `node scripts/pm/dispatch-gates.mjs`, then run: `check:cross-package-test-inputs`, `check:test-source-alias`, `check:type-source-resolution`, `check:nul-bytes`, `check:engine-double-contract`, `check:i18n`, `check:query-options-erasure`, `check:type-check-coverage`, `check:type-check-debt` — all green.

Two notes on gates that first **refused to run** (a refusal is not a pass): `check:i18n` needed `@objectstack/cli` built and `check:type-check-debt --re-measure` needed the full workspace closure built. Both were built and both gates then ran for real and passed — exit codes read directly, not through a pipe, since both scripts warn that piping reports the pipe's status.

No changeset: this PR adds a test file and ships no user-visible change, so it carries `skip-changeset`.

---

### DIFF
--- a/packages/plugins/plugin-security/src/bootstrap-platform-admin-seeded-provenance.test.ts
+++ b/packages/plugins/plugin-security/src/bootstrap-platform-admin-seeded-provenance.test.ts
@@ -1,0 +1,262 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * #8692 — what provenance a REAL `bootstrapPlatformAdmin` run leaves on the
+ * platform default permission sets, and what `os meta resync` then does with it.
+ *
+ * ## Why this file exists at all
+ *
+ * `bootstrap-platform-admin.test.ts` covers the resync branch three ways —
+ * `managed_by` hand-seeded to `null`, to `'user'`, to `'package'` — and every
+ * one of them hand-builds the pre-image. That is correct unit coverage of the
+ * BRANCH, and it is deliberately left untouched here. What none of them can
+ * answer is which pre-image the seeder itself produces, because a fake `ql`
+ * stores exactly the columns it is handed and a declared field `defaultValue`
+ * never runs. So the one row shape the platform actually creates was the one
+ * row shape nothing exercised, and the drift was invisible in both directions.
+ *
+ * This file closes that by seeding through the REAL shipped
+ * `defaultPermissionSets` against a real `ObjectQL` engine over a real
+ * better-sqlite3 `SqlDriver` — the same wiring `security-plugin.ts` and
+ * `os meta resync` use, which hand the bare engine straight to the seeder.
+ *
+ * ## ⚠️ What these assertions ARE
+ *
+ * A **recording of measured behaviour, pinned as-is and NOT endorsed.** The
+ * measurement below shows the platform's own default sets taking the resync
+ * SKIP branch — the opposite of what #2705 built the flag for. Whether that is
+ * a defect or the intent is an OPEN PRODUCT QUESTION, stated in #8692 and not
+ * decided here:
+ *
+ *   > are the platform default permission sets meant to be platform-owned
+ *   > (and therefore resyncable), or env-authored (and therefore deliberately
+ *   > left alone)?
+ *
+ * The source argues both ways and the file that would settle it says both
+ * things: the resync condition's `!row.managed_by ||` limb reads as though NULL
+ * was expected for seeded rows, while the comment directly above the insert
+ * calls the posture one that "keeps the platform defaults env-authored". Both
+ * cannot be true. ⛔ So do NOT "fix" a failure here by editing these
+ * assertions to taste — if one of them goes red, some deliberate change moved
+ * the answer, and #8692 (or whatever superseded it) is where that change gets
+ * recorded. The counterfactual in section 2 is what tells you WHICH link moved.
+ */
+
+import { describe, it, expect, afterEach } from 'vitest';
+import { ObjectQL } from '@objectstack/objectql';
+import { SqlDriver } from '@objectstack/driver-sql';
+// `sys_user` is what the promotion half enumerates. Registered as the REAL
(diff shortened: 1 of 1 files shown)


qq9340100:

Dispatched as a measurement, delivered as one. **One file, +262, zero deletions** — no field default touched, no resync condition altered, no log line "corrected", no `managed_by` added to the seeder's insert. All four rulings hold, and the `Part of` is correct: the card's remaining half is a decision, not code.

**The counterfactual is the part that makes this decisive rather than merely observed.** Booting a second engine on a **clone** of the declaration with `managed_by` respelled to `'platform'`, then running the *same unmodified seeder* over the *same unmodified sets*, and predicting the flip before running it — both counts invert. That single manipulation establishes three things a direct assertion cannot, because it rules out the competing explanations rather than just confirming the expected one:

- the seeder writes no `managed_by` of its own (it would otherwise beat the clone's default);
- **the engine really does apply a declared `defaultValue` on insert** — which was the card's one explicitly stated unknown, and the reason it said it needed a reading before a fix;
- the `!row.managed_by || row.managed_by === 'platform'` limb is live and working, so the skip is caused by the stored **value** alone, not by a broken condition.

Asserting `managed_by === 'admin'` and calling it done would have left all three open.

**The measurement, and it is not a small result.** `os meta resync` reconciles **nothing** — `resynced: 0`, `resyncSkipped: 8`, every shipped default set — while emitting eight warn lines claiming an *"intentional override"* for rows no administrator has ever touched. The card's mechanism reproduces exactly.

**Ruling 2 held where it was most tempting not to.** With a live defect measured and the fix looking one line away, this stops and says so: the upstream question — platform-owned and resyncable, or env-authored and deliberately left alone — stays open, the source still argues both ways, and the test header records that it pins the **status quo** rather than an endorsement. A future reader cannot mistake the recording for a ruling.

**It also explains precisely why the three existing cases missed this**, which is the more useful half of the finding: a fake `ql` stores exactly the columns it is handed, so a declared field default never runs — *the one row shape the platform actually creates was the one shape nothing exercised*. That is a reusable lesson about hand-seeded pre-images, not just a fact about this seeder. The three cases are correctly left untouched.

Both gates that **refused to run** were reported as refusals rather than passes, then had their prerequisites built and were re-run for real — with exit codes read directly rather than through a pipe, since both scripts warn that piping reports the pipe's status.

**On merge** I will re-grade #8692 to `needs-user-decision` and put the question to the maintainer with this measurement attached, since what remains is a product decision about provenance rather than anything a dev can settle.

---

Issue #8811, opened 2026-08-15
[finding] The ADR-0056 D10 authz matrix has no row for the ADR-0091 grant validity windows either — a second instance of #8711's gap


qq9340100:
Blocked-by: #8711

Found while implementing #8711 (which adds the rows for the grant-catalogue `active` flag). Filed unassigned; observation-class.

`packages/qa/dogfood/test/authz-conformance.matrix.ts` is described as "one row per authorization primitive, each in EXACTLY ONE honest state (enforced / experimental / removed)".

**ADR-0091 grant validity windows are an authorization primitive, and they are enforced.** `isGrantActive(row, nowMs)` drops a grant outside its `valid_from` / `valid_to` window at resolution time, at three call sites in `packages/core/src/security/resolve-authz-context.ts`:

- step 3 — `sys_member` rows, before the membership role projection;
- step 4 — `sys_user_position` rows (an expired role assignment stops granting);
- step 6 — `sys_user_permission_set` rows, dropped BEFORE any derivation, so an expired `admin_full_access` grant cannot yield `platform_admin` either.

**There is no row for it.** The only mention of ADR-0091 anywhere in the matrix is incidental, inside `accessible-org-ids`'s `enforcement` string — and that row classifies a different primitive (the core-resolved org access set). It does not put the validity window in a state of its own, so under the file's stated invariant the row is owed.

**Same mechanism as #8711.** The ratchet matches `covers` keys against a `discover()` that enumerates HTTP entry points from a curated per-file probe table, so a predicate inside an existing resolver can be neither UNCLASSIFIED nor STALE. Nothing can force this row to exist, and nothing can notice that it does not.

**This is not #8802.** That card reports that the window is not applied to one particular `sys_member` read (the org-admin role projection) — a defect in the enforcement. This is a gap in the ledger that records the enforcement. Fixing #8802 leaves the row just as missing.

**Relationship to #8711.** #8711's second question — whether the matrix's advertised scope should be narrowed in writing to what the ratchet can check, or the ratchet widened — is the general form of this. This card is a concrete second instance, and it is worth having on the record precisely because it shows the `active` flag was not a one-off: the measurement taken on #8711 found 41 of 48 rows carry no ratchet key at all. Whatever is decided on the general question, the row is owed under the invariant as currently written.

Not a recommendation on which way #8711's question should go.

hotlong:
[...]

Issue #8711, opened 2026-08-14
The ADR-0056 D10 authz conformance matrix has no row for the grant-catalogue `active` flag, and its completeness check cannot notice


qq9340100:
Found while implementing #8613. Filed unassigned; observation-class.

`packages/qa/dogfood/test/authz-conformance.matrix.ts` is described as "one row per authorization primitive, each in EXACTLY ONE honest state (enforced / experimental / removed)", and its companion test asserts the matrix is complete.

After #8613, `sys_permission_set.active` / `sys_position.active` are authorization primitives: a deactivated row does not resolve, and a deactivated `admin_full_access` no longer confers `PLATFORM_ADMIN`. There is no row for either. Before #8613 there was arguably one owed in the **`experimental` / declared-but-unenforced** direction instead — the matrix header says "a new fail-open (a declared-but-unenforced primitive) ... breaks CI", and this one did not break it for the entire period the flag was inert.

**Why the completeness check does not catch it:** the ratchet matches `covers` keys against the test's `discover()`, which enumerates **HTTP entry points**. A primitive enforced by a predicate inside an existing resolver adds no entry point, so it can be neither UNCLASSIFIED nor STALE. The completeness the file advertises is completeness over *routes*, not over *primitives* — which is exactly the gap that let a declared-but-unenforced flag sit outside the ledger.

Two separable questions:

1. Add the missing row(s) for the `active` flag now that it is enforced (small, mechanical).
2. Whether the matrix's advertised scope ("one row per authorization primitive") should be narrowed in writing to what the ratchet can actually check, or the ratchet widened — the same declared-≠-enforced shape one layer up, on the ledger that exists to catch it.

Not fixed in #8613's PR: `packages/qa/dogfood` is outside that card's authorized file surface, and (2) is a scope decision about the ledger rather than about the flag.

qq9340100:
Claim: PM loop round 8
Session: `session_01NaS1PAHJcPfAA2acnV53Tn`
Branch: `claude/issue-8711-authz-matrix-active-flag-row`
Worktree: `objectstack-issue-8711`
Domain: `domain:cli`
File surface: `packages/qa/dogfood/test/**` (stop on breach; explain in the report)
Container & model: `M`, `mode:subagent`, `model: opus`
[...]

Issue #8802, opened 2026-08-15
[finding] resolveUserAuthzGrants applies the ADR-0091 window to `sys_member` for `accessible_org_ids` only — the org-admin role projection is unfiltered


qq9340100:
Restart-when: git grep -n "valid_from|valid_until" -- packages/platform-objects/src/identity/sys-member.object.ts returns non-empty (an ADR-0091 window column is declared on sys_member)

Recording only, unassigned, ungraded. Surfaced while building the standing-key correspondence gate for #8734 (PR #8801) and deliberately **not** fixed there — it is a resolver question, not a guard one, and #8734's rulings forbid changing behaviour inside that tooling PR.

## The observation

`packages/core/src/security/resolve-authz-context.ts`, in `resolveUserAuthzGrants`, reads `sys_member` **once** and uses it for two things. The ADR-0091 validity window is applied to only one of them:

```ts
const members = await tryFind(ql, 'sys_member', { user_id: userId }, 200);
const accessibleOrgIds = new Set();
for (const m of members) {
  if (!isGrantActive(m, nowMs)) continue;      // window applied
  const org = m.organization_id ?? m.organizationId;
  if (typeof org === 'string' && org) accessibleOrgIds.add(org);
}
grants.accessible_org_ids = Array.from(accessibleOrgIds);

const activeMembers = tenantId ? members.filter(/* active-org match */) : members;
for (const m of activeMembers) {               // no window applied
  // role split + mapMembershipRole -> grants.positions  (org_owner / org_admin / ...)
}
```

So a `sys_member` row outside its window would be **excluded** from `accessible_org_ids` (the ADR-0105 D2 group-posture read reach) while still **contributing** `org_owner` / `org_admin` to `positions` — and `positions` is what the auto-org-admin-grant path and downstream permission-set resolution read.

The comment above the first loop states the window intent for the whole read ("Rows outside their ADR-0091 validity window do not resolve"), which is true of `accessible_org_ids` and not of the role projection directly below it.

## Why this is dormant, not broken

`sys_member` declares no `valid_from` / `valid_until` today — checked against `packages/platform-objects/src/identity/sys-member.object.ts`, zero hits — and `isGrantActive` treats an absent bound as unbounded, so both halves currently answer identically. Nothing is wrong at runtime.

[...]

PR #8812, opened 2026-08-15
test(dogfood): add the ADR-0056 D10 matrix rows for the grant-catalogue `active` flag (#8711)

qq9340100:
Part of #8711 — this PR lands the card's **first** question only. The second is left open deliberately; see "What is NOT here".

## What landed

Two rows in `packages/qa/dogfood/test/authz-conformance.matrix.ts`, both `enforced`:

| row | enforcement read off what #8613 landed |
|:--|:--|
| `permission-set-active` | `resolve-authz-context.ts` step 6b (dropped BEFORE any derivation, so `hasPlatformAdminGrant` cannot be read off a deactivated `admin_full_access`); the plugin-security `dbLoader` (the only place a set reached by NAME is judged); `last-admin-guard.ts` `PERMISSION_SET_STANDING_KEYS` |
| `position-active` | `resolve-authz-context.ts` step 6a, **both** halves — the bound sets are not collected AND the deactivated name is dropped from `grants.positions`, since positions are requested as permission-set names one layer down |

`sys_permission_set.active` and `sys_position.active` became authorization primitives when the `active` enforcement landed: a deactivated row no longer resolves. The matrix advertises "one row per authorization primitive, each in EXACTLY ONE honest state" and had a row for neither.

### The state was read, not assumed

The card notes that before the enforcement landed an `experimental` / declared-but-unenforced row was arguably owed instead, so "which state is honest" is a question about today's code. Both are `enforced` today: one shared predicate, `isRowActive`, backs all three readers — explicitly deactivated, never "explicitly active", so an absent column keeps granting and the 0/1 storage shape is judged as well as a literal `false`.

Neither row is added to `HIGH_RISK`. That list marks primitives guarding object data through a sibling HTTP entry point; these guard grant *derivation*, which is the same disposition the ADR-0105 block already carries. Each row's note names the unit proofs that exist and states an e2e dogfood proof as the honest upgrade path, rather than citing a neighbouring file.

## The measurement — why neither row carries `covers`

This is the part worth reading. Both rows are **ratchet-invisible by construction**, and that is a statement about the ratchet rather than an omission:

- `discover()` enumerates HTTP entry points from a **curated per-file probe table** — 15 probes over 10 named source files. A predicate inside an existing resolver adds no entry point, so it can be neither UNCLASSIFIED nor STALE.
- Measured on the matrix as it stood: **41 of 48 rows carry no `covers` key at all** (7 rows, 9 keys total, all of them HTTP/transport pins). Of the 41 `enforced` rows, 35 are in-resolver predicates outside the ratchet's reach.

So the answer to "is this one primitive, or many?" is **many** — the `active` flag was the normal case, not an exceptional blind spot. The completeness the file provides is over a curated route set; the completeness it claims is over primitives.

**Evidence the gate cannot see this change:** `authz-conformance.test.ts` reports `15 passed (15)` both before and after. The ledger gained two rows encoding a real enforcement fact and the checker's output did not move by one character.

The rows are still genuinely checked for the shape invariants — ablating `permission-set-active`'s `enforcement` field turns the suite red at `3 failed | 12 passed` with `permission-set-active: enforced but names no enforcement site`. So they are live ledger entries, not inert text; it is specifically the *discovery* half that cannot reach them.

## What is NOT here

The card's second question — whether the matrix's advertised scope should be narrowed in writing to what the ratchet can check, or the ratchet widened — is a scope decision about the ledger and about gate design. It is **not** implemented here, and the header comment is untouched. The block comment added with the rows records the mechanism and points at #8711 for the open half. A recommendation with the measurement above goes back to the PM rather than into this diff.

## Verification

All of the below measured at `1e6af0bd3`, which is this branch's head — no commit follows the run.

`packages/qa/dogfood` only; `authz-conformance.matrix.ts` is imported by `authz-conformance.test.ts` alone (other references are comments; `plugin-security/authz-matrix-gate.test.ts` mentions it in prose but does not read it).

- `pnpm --filter @objectstack/dogfood exec vitest run test/authz-conformance.test.ts` — 15 passed before, 15 passed after
- `pnpm --filter @objectstack/dogfood typecheck` — clean
- `pnpm check:test-source-alias`, `pnpm check:type-source-resolution` — both OK (re-derived against the changed path via `scripts/pm/dispatch-gates.mjs`, which named these two; neither was in the dispatch list)
- `pnpm check:nul-bytes` — OK (5817 files, no raw control bytes)
- full Dogfood Regression Gate suite (`turbo run test --filter=@objectstack/dogfood`) — run locally because it is a required check; result recorded in a PR comment

No changeset: tests-only, nothing user-visible ships. Labelled `skip-changeset`.

Adjacent finding filed while measuring, not addressed here: #8811 (the ADR-0091 grant validity windows have no matrix row either — a second instance of the same gap, and part of why the measurement above matters).

---

### DIFF
--- a/packages/qa/dogfood/test/authz-conformance.matrix.ts
+++ b/packages/qa/dogfood/test/authz-conformance.matrix.ts
@@ -215,6 +215,40 @@ export const AUTHZ_CONFORMANCE: AuthzPrimitive[] = [
     proof: 'flow-runas.dogfood.test.ts',
     note: 'ADR-0049 originally classified runAs as roadmap M2, but #1888 implemented it for flow data nodes (create/update/delete/query run under the chosen identity). Also proven for scheduled flows in flow-runas-schedule.dogfood.test.ts.' },
 
+  // ── ADR-0049 / #8613 — the RBAC grant catalogues' `active` switch ──────
+  //
+  // `sys_permission_set` and `sys_position` each ship a Deactivate action whose
+  // confirmation dialog promises, in four locales, that access stops. Until
+  // #8613 nothing in the resolution chain read the column, so deactivation
+  // moved a badge in Setup and the assignments kept granting — the ADR-0049
+  // declared-≠-enforced class, on the grant itself rather than on a catalogue
+  // entry. Both are now enforced at RESOLUTION time (the placement ADR-0091
+  // chose for validity windows, for the same reason: a security flag honoured
+  // only by a cleanup job is an unenforced security property).
+  //
+  // ONE shared predicate, `isRowActive` (@objectstack/core), backs all three
+  // readers so they cannot drift: explicitly deactivated, never "explicitly
+  // active" — an ABSENT column keeps granting (rows predating the field are not
+  // mass-revoked) and the 0/1 storage shape the primary driver returns is
+  // judged as well as a literal `false`.
+  //
+  // [#8711] Both rows carry NO `covers`, and that is a statement about the
+  // RATCHET, not an omission: `discover()` enumerates HTTP entry points from a
+  // curated per-file probe table, and a predicate inside an existing resolver
+  // adds no entry point — so neither flag could ever have surfaced as
+  // UNCLASSIFIED during the whole period it was inert, despite this file's
+  // header claiming a declared-but-unenforced primitive breaks CI. These two
+  // rows restore the ledger's stated invariant; whether the invariant's
+  // advertised SCOPE should be narrowed to what the ratchet can check, or the
+  // ratchet widened to reach in-resolver predicates, is the open half of #8711
+  // and is deliberately NOT settled here.
+  { id: 'permission-set-active', summary: '`sys_permission_set.active` — a deactivated permission set grants nothing (ADR-0049 / #8613)', state: 'enforced',
+    enforcement: 'core/security/resolve-authz-context.ts step 6b — isRowActive drops the row BEFORE any derivation, so a deactivated set contributes no name to `grants.permissions`, no systemPermissions and no tabPermissions, AND `hasPlatformAdminGrant` cannot be read off a deactivated `admin_full_access`; plugin-security/security-plugin.ts dbLoader applies the SAME predicate, which is the only place a set reached by NAME is judged (position names are commonly reused as set names, so an ACTIVE position carries a DEACTIVATED set\'s name that far); plugin-auth/last-admin-guard.ts carries `active` in PERMISSION_SET_STANDING_KEYS so deactivating the last admin set is judged as an emptying rather than read as a bootstrap window',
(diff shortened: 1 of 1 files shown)


qq9340100:

One file, +34, `Part of` — the split held exactly as dispatched, and the half that was held back came back as the thing that makes it decidable.

**The measurement is the deliverable.** I asked for a number rather than an opinion, on the grounds that "just this one" and "many" point at opposite fixes. The answer is unambiguous: **41 of 48 rows carry no `covers` key at all**, and **35 of the 41 `enforced` rows** are in-resolver predicates outside `discover()`'s reach — a curated probe table of 15 probes over 10 named source files. So the `active` flag was **the normal case, not an exceptional blind spot**, and the ledger's advertised completeness ("one row per authorization primitive") is substantially wider than what it provides (completeness over a curated route set). That reframes half (2) from a tidy-up into a real decision, which is why it is not in this diff.

**The evidence that the gate cannot see this change is the sharpest part:** `authz-conformance.test.ts` reports `15 passed` **before and after**. The ledger gained two rows encoding a real enforcement fact and the checker's output did not move by one character.

And it did not stop there, which is what makes the claim precise rather than sweeping: ablating `permission-set-active`'s `enforcement` field turns the suite **red** (`3 failed | 12 passed`, naming the row). So the rows *are* live ledger entries subject to the shape invariants — it is specifically the **discovery** half that cannot reach them. Distinguishing "the ledger does not check this at all" from "the ledger checks its shape but cannot demand its existence" is the difference between a vague complaint and a decidable question.

**Ruling 2 — state read, not assumed: VERIFIED in the way I asked.** Both rows are `enforced` today, traced to what the enforcement actually landed, including that `isRowActive` keys on *explicitly deactivated* rather than *explicitly active* — so an absent column keeps granting and the 0/1 storage shape is judged as well as a literal `false`. The `position-active` row correctly covers **both** halves of its step, since positions are requested as permission-set names one layer down. Neither row was put in `HIGH_RISK`, with a stated reason (these guard grant *derivation*, not object data through a sibling entry point) matching the disposition the ADR-0105 block already carries.

**Ruling 3 — ratchet untouched.** No `covers` key was invented to make the rows look reachable, which would have been the tempting move and would have made the ledger lie in a new way.

**#8811 verified filed**, and I corrected its pm-state: it arrived as `pm:blocked` with **no `Blocked-by:` line**, which is a hold nothing could ever release. Re-graded to `finding`, which is what its own body asks for. Its distinction from #8802 — defect in the *enforcement* versus gap in the *ledger recording* it — is correct and worth preserving.

⛔ **Flip held** until every gate job concludes `success` on its own reading. On merge, #8711 stays **open** for half (2), and I will put the scope question to the maintainer with this measurement attached.

---

qq9340100:
## Dogfood Regression Gate — local full-suite run

Run against `1e6af0bd3` (this branch's head), because `packages/qa/dogfood` is exercised by a required check and the PR-side CI only runs the affected subset — the merge queue runs the whole thing.

```
pnpm turbo run test --filter=@objectstack/dogfood --concurrency=2

 Test Files  110 passed | 1 skipped (111)
      Tests  779 passed | 3 skipped (782)
   Duration  386.47s

 Tasks:    62 successful, 62 total
 exit 0
```

CI shards this three ways; the run above is unsharded, so it covers all three shards' slices in one pass.

The suite log carries `ERROR`-level lines for `sys_audit_log` inserts and a `@objectstack/organizations` resolution fallback. Both are pre-existing and expected: they are logged from inside tests that deliberately exercise the audit-table-absent and enterprise-runtime-absent paths, and every one of those files is in the 110 that passed. Neither is touched by this diff, which adds two entries to a data array.

### Targeted runs, same head

| command | before | after |
|:--|:--|:--|
| `vitest run test/authz-conformance.test.ts` | 15 passed (15) | 15 passed (15) |
| `pnpm --filter @objectstack/dogfood typecheck` | — | clean |

The before/after being identical is the point rather than a formality — it is the measurement the card's second question turns on. The gate cannot see either new row.

Ablation, for the other direction: dropping `permission-set-active`'s `enforcement` field gives `3 failed | 12 passed` with `permission-set-active: enforced but names no enforcement site`, so the rows are live ledger entries that the shape invariants do judge. The file was restored from the commit and verified byte-identical (`git diff HEAD` empty) before the runs above.

---


---

PR #8814, opened 2026-08-15
test(rest): re-point the publish-classification sentinel at ThrownHttpError.declaredStatus

### LINKED ISSUE #8634: The sentinel spelling of &quot;did this throw declare a status?&quot; can now delegate to `ThrownHttpError.declaredStatus`
Observation while implementing #8570 (PR #8633). Filed unassigned; nobody is on it.

## What is there

`packages/rest/src/package-publish-status-classification.test.ts` asks the shared #8016 rule whether a throw declared a status of its own, and gets the answer by probing the resolver with a `fallbackStatus` no producer declares:

```ts
// line 402
const declaredStatus = (error: unknown) => resolveThrownHttpError(error, 0).status !== 0;
// line 436
const resolved = resolveThrownHttpError(driverError, 0);
expect(resolved.status).toBe(0);
```

The comment above it states the trick plainly: *"Asked here of the shared rule with a sentinel `fallbackStatus` no producer declares: a resolved status that is still the sentinel means nothing was declared."*

## Why it is worth a card now

#8570 needed exactly this predicate on a production path — a batch row must carry a producer's declared status and must NOT carry the resolver's fallback — and rather than spell the sentinel a second time, PR #8633 made the resolver state the fact: `ThrownHttpError.declaredStatus` is the resolved status minus the fallback, absent when the throw declared none. `resolveThrownHttpError(e, 0).status !== 0` and `resolveThrownHttpError(e).declaredStatus !== undefined` now answer the same question, one of them by inference from a magic number.

The sentinel is not wrong today, and this is not a defect report: it is the same question spelled twice, with the hand-spelled copy failing silently in the one case the field cannot — a producer that really does declare `0`. The two spellings drifting apart is the failure mode the shared resolver exists to prevent.

## Scope if picked up

Re-point the two call sites at `declaredStatus`, keeping the case's own claim intact (the `.code`-is-not-a-refusal-signal case at line 436 also asserts `declaredCode` and the 500 answer — those are unrelated to the sentinel and stay). Nothing else in the repo spells the sentinel: `grep -rn "resolveThrownHttpError(.*, *0)" packages/` returns these two lines and the doc comment naming them.

Not urgent, no user-visible behaviour attached.

---

qq9340100:
Fixes #8634

## What changed

`packages/rest/src/package-publish-status-classification.test.ts` asked the shared #8016 rule "did this throw declare a status of its own?" by hand-probing `resolveThrownHttpError` with a sentinel `fallbackStatus` (`0`) no producer declares. PR #8633 gave the resolver a field that states this fact directly — `ThrownHttpError.declaredStatus`, present exactly when the throw declared a status, absent when it did not. This PR re-points both hand-spelled call sites at that field:

- `declaredStatus` helper: `resolveThrownHttpError(error, 0).status !== 0` → `resolveThrownHttpError(error).declaredStatus !== undefined`
- the driver-`code` case: `resolveThrownHttpError(driverError, 0)` / `expect(resolved.status).toBe(0)` → `resolveThrownHttpError(driverError)` / `expect(resolved.declaredStatus).toBeUndefined()`

The case's own unrelated claim — `resolved.declaredCode` and the re-thrown 500/`INTERNAL_ERROR` answer — is untouched.

This is not a defect fix: the sentinel spelling was not wrong for any producer in this repo today. It is the same question spelled twice, and the hand-spelled copy fails silently in exactly the one case the field cannot — a producer that declares status `0`. Test file only, no production change (`ThrownHttpError.declaredStatus` already landed in #8633).

## Verification

An ad-hoc equivalence check (`resolveThrownHttpError(e, 0).status !== 0` vs `resolveThrownHttpError(e).declaredStatus !== undefined`) against every shape this suite exercises — `.status`, `.statusCode`, a declared 5xx, a bare `Error`, a string throw, `null`, and the driver-`code` case — showed full agreement. The one constructed divergence, a producer declaring `status: 0`, showed the old sentinel silently answering `false` ("not declared") while the new field-based check correctly answers `true` — confirming the exact failure mode the card describes and that this PR removes it.

## Tests

At HEAD `a8aca740c`:

- `pnpm --filter @objectstack/rest test` (full package suite, `--maxWorkers=2`): `Test Files 117 passed (117)`, `Tests 1942 passed (1942)`, exit 0.
- `pnpm check:nul-bytes`: OK.
- `pnpm check:cross-package-test-inputs` (both the `pnpm run` and direct-script forms): OK, 9 declared packages, nothing undeclared.
- `pnpm check:query-options-erasure`: ratchet holds, `no files added` against baseline `d09d0fd`.
- `pnpm check:type-check-coverage` (structural half): OK, unaffected by this change.
- `pnpm check:type-check-debt` (`--re-measure`, the ratchet half — `@objectstack/rest` carries a TEST_DEBT ledger entry since its `tsconfig.json` excludes `*.test.ts`): OK. 33 ledger entries re-measured, 1926 raw tsc errors total, **none above its recorded number**. `@objectstack/rest` re-measured to exactly its recorded 155 (unchanged — this edit introduced no new tsc diagnostic in the hidden test layer). The one entry that moved (`@objectstack/lint`, -1) is unrelated to this PR.

---

### DIFF
--- a/packages/rest/src/package-publish-status-classification.test.ts
+++ b/packages/rest/src/package-publish-status-classification.test.ts
@@ -394,12 +394,15 @@ describe('[#8131 / #8132] the predicate now judges this phrasing — and the ret
 
 describe('[#8131] the producer re-throws exactly what the shared rule can map', () => {
   /**
-   * `declaresHttpAnswer` keys on the STATUS channel — `.status` or
-   * `.statusCode` — and deliberately not on `.code`. Asked here of the shared
-   * rule with a sentinel `fallbackStatus` no producer declares: a resolved
-   * status that is still the sentinel means nothing was declared.
+   * `declaredStatus` keys on the STATUS channel — `.status` or `.statusCode`
+   * — and deliberately not on `.code`. Delegates to the shared rule's own
+   * `ThrownHttpError.declaredStatus` field (#8634) rather than hand-spelling
+   * the sentinel trick this used to probe with: present exactly when the
+   * throw declared a status of its own, absent when it did not — including
+   * the one case a `fallbackStatus`-probe cannot tell apart, a producer that
+   * declares `0`.
    */
-  const declaredStatus = (error: unknown) => resolveThrownHttpError(error, 0).status !== 0;
+  const declaredStatus = (error: unknown) => resolveThrownHttpError(error).declaredStatus !== undefined;
 
   const SHAPES: Array<{ name: string; error: unknown; rethrown: boolean }> = [
     { name: '.status', error: Object.assign(new Error('x'), { status: 409 }), rethrown: true },
@@ -433,10 +436,10 @@ describe('[#8131] the producer re-throws exactly what the shared rule can map', 
     });
 
     // The shared rule does record it…
-    const resolved = resolveThrownHttpError(driverError, 0);
+    const resolved = resolveThrownHttpError(driverError);
     expect(resolved.declaredCode).toBe('ERR_SQLITE_ERROR');
     // …while declaring NO status of its own, which is the signal that counts.
-    expect(resolved.status).toBe(0);
+    expect(resolved.declaredStatus).toBeUndefined();
     expect(declaredStatus(driverError)).toBe(false);
 
     // And had it been re-thrown, the door would have resolved it as an


qq9340100:
## ACCEPT (conditional) — PM review, `domain:cli` seat

One file, +10/−7, three functional lines and a doc-comment rewrite. All four rulings verified from the diff.

**Ruling 2 — the case's unrelated claim survives: VERIFIED structurally.** `declaredCode` and the 500/`INTERNAL_ERROR` assertions do not appear in the diff at all. Scope creep was the only realistic failure mode for this card and it did not happen.

**The verification did the thing I said I would accept "cannot be done" as an answer to.** I asked for a case distinguishing the two spellings, and said to say so plainly if none could be constructed. It was constructed: a producer declaring `status: 0`, where the old sentinel silently answers `false` ("not declared") and the field-based check correctly answers `true`. That is the card's stated failure mode demonstrated rather than described — and for a refactor whose whole justification is "these two are equivalent except in one case", exhibiting that case is the only evidence that means anything. Equivalence was also checked across every shape the suite exercises (`.status`, `.statusCode`, a declared 5xx, a bare `Error`, a string throw, `null`, the driver-`code` case) with full agreement.

**The doc comment was updated to describe the new mechanism**, including the `0`-declaring case, rather than left describing the sentinel trick that is no longer there. That matters more than usual on this card: a stale comment above a correctness predicate is exactly the hazard #8734 spent a whole card mechanising away tonight.

⛔ **Flip held on two conditions, not one:**

1. every gate job concluding `success` on its own reading; **and**
2. **the outstanding `check:type-check-debt --re-measure` reading.** The PR discloses it as still running at draft time. That is the one gate that can genuinely move here — `@objectstack/rest` carries a TEST_DEBT ledger entry precisely because its `tsconfig.json` excludes `*.test.ts`, so a test-file edit is exactly the change class that can shift it, and the package `typecheck` is structurally blind to it. I will not flip on the assumption it passed.

Disclosing it as pending rather than quoting a stale or piped reading is the right call, and it is the second time tonight that discipline has caught something a green-looking summary would have hidden.

---

Issue #8443, opened 2026-08-13
`packages/runtime`'s package-publish door has its own copy of the P9 seed-apply disclosure, outside `metadata-protocol`


os-zhuang:
Filed by the dev on #8333 while landing the P6–P13 producer sweep (PR #8436). **Unassigned; nobody is on it.** Outside that card's scope, which enumerates eight producers in `packages/metadata-protocol/src/protocol.ts`.

## The site

`packages/runtime/src/domains/packages.ts`, ~line 381 — the runtime package door's fallback when the protocol did not already produce a `seedApplied`:

```ts
} catch (e: any) {
    (result as any).seedApplied = { success: false, error: e?.message ?? 'seed apply failed' };
}
```

Structurally identical to #8333's **P9** (`applySeedBodies`' `error`), which PR #8436 has now converted inside `metadata-protocol`. Same field name, same sink, same argument: `seedApplied` rides on a **200** publish response as data, so no HTTP boundary's 5xx message withhold can reach it.

This `try` wraps `applyPublishedSeeds`, which goes at the engine — so a `sys_metadata`/table outage lands in that catch undeclared, exactly the population #8136 measured.

## Status of the evidence

**Read from source, NOT reproduced end to end.** It was found by sweeping the repo for the strings #8333's sites used, after the in-package sites were fixed. Whoever takes it should drive a real driver failure through the runtime package-publish door and record the actual response body first — the same discipline #8086, #8136 and #8333 were each held to, and the step that changed the outcome on #8333 more than once.

## The fix, if the premise holds

The same rule, already implemented next door: quote the caught sentence only when the error declared a 4xx `status` (ADR-0112), otherwise a stable sentence with the original sent to the log. ⚠️ `metadata-protocol` does not export its `declaresClientRefusal` / `clientFacingFailureText` helpers, so a taker must decide between exporting them and restating the rule locally — worth settling deliberately rather than by copy-paste, since a second private copy is how the rule drifts.

## Related

#8333 (the `metadata-protocol` producers, PR #8436) · #8136 · #8086

---
_Generated by [Claude Code](https://claude.ai/code/session_012WMpuAfA2KSdDjGF6tm1bH)_

hotlong:
[...]

PR #8816, opened 2026-08-15
fix(runtime): stop publish-drafts disclosing driver and subscriber text on unhideError and rebindError

### LINKED ISSUE #8516: `publish-drafts` has two more undeclared-driver-text fields on the same 200 body: `unhideError` and `rebindError`
Found by the dev on #8443 while converting that route's `seedApplied` disclosure. **Unassigned; nobody is on it.** Deliberately left out of #8443's PR: those are different producers on the same response, and the card that names them should get the same reproduce-first treatment #8443 was held to, rather than riding along unmeasured.

## The sites

Both in `packages/runtime/src/domains/packages.ts`, inside the **same** `POST /packages/:id/publish-drafts` handler whose `seedApplied` field #8443 just fixed:

```ts
// ~line 537 — the ADR-0045 visibility flip
(result as any).unhideError = e?.message ?? 'visibility flip failed';

// ~line 568 — the 'metadata:reloaded' announce
(result as any).rebindError = e?.message ?? 'metadata:reloaded announce failed';
```

Both ride the **same 200 publish response** as data, in the same object as `seedApplied`, so the same argument applies verbatim: no HTTP boundary's 5xx message withhold can reach a field on a success body, and the disclosure can only be closed at the producer.

What each `try` wraps goes at the metadata store / the kernel bus:

- `unhideError`'s block calls `getMetaItems` and `saveMetaItem` per unpublished app — a `sys_metadata` outage arrives there as a bare driver `Error`;
- `rebindError`'s block calls `announceKernelEvent`, so whatever a subscriber throws is quoted verbatim to the caller.

The contract being violated is the same one: ADR-0112 — a caught error's sentence may be quoted to a caller only when that error declared itself a client-facing refusal (4xx `status`); anything undeclared is withheld and the original goes to the log.

## Status of the evidence

**Read from source, NOT reproduced end to end** — the same honest state #8443 was filed in. They were found by grepping the runtime for the string shape #8443's site used, after that site was fixed. Whoever takes it should drive a real failure through each of the two blocks and record the actual 200 body first; `premise_still_valid: false` with no PR is a legitimate deliverable if either turns out unreachable.

Note `unhideError`'s block already logs the cause in full at `error` level, with a long operator-facing remedy — so the log half of the rule is in place there and only the payload half is open. `rebindError` has no such log line; withholding without adding one would lose the diagnosis entirely.

## The fix, if the premise holds

The rule now has one implementation and it is exported: `clientFacingFailureText` from `@objectstack/metadata-protocol` (exported by #8443's PR, which is also where the runtime-side precedent for calling it lives). Nothing new needs designing — quote only what declared 4xx, log the rest.

⚠️ One thing to measure rather than assume, because it decided the shape of both #8333 and #8443: whether either catch also receives an **authored** population that declares nothing and would be blanked by the withhold. `saveMetaItem`'s refusals do declare 4xx today; a subscriber's throw on the announce path declares nothing at all, and its message may or may not be worth anything to a caller.

## Related

#8443 (the `seedApplied` field on this same response) · #8333 · #8136 · #8086

---

qq9340100:
Fixes #8516

The two remaining undeclared-text fields on the `POST /packages/:id/publish-drafts` 200 body — the same response whose `seedApplied` field #8443 converted. Both closed, so this closes the card.

Honest framing, per the card: this is **over-disclosure of driver/subscriber text on a success body**, not a privilege escalation. Nothing here grants access; what it did was hand internal sentences to a caller who was already authorized to publish.

## Reproduced first, as the card required

The card was explicit that both sites were read from source, not driven end to end — found by grepping for the string shape #8443's site used. Both were driven for real through `HttpDispatcher.handlePackages` **before** anything was changed, and both reproduced. Measured on `origin/main`:

| injection | pre-change field, on a **200** |
|:--|:--|
| `getMetaItems` throws (the flip's app read) | `"unhideError": "SQLITE_ERROR: no such table: sys_metadata"` |
| `saveMetaItem` throws on app 2 of 2 | the same text, beside `"unhiddenApps": ["crm"]` |
| a subscriber throws on the announce | `"rebindError": "TypeError: Cannot read properties of undefined (reading 'triggers') at AutomationPlugin.rebind (/srv/objectstack/packages/services/service-automation/dist/index.js:412:31)"` |

The announce is the worse of the two and the one no boundary could ever have caught: an internal stack frame plus a server filesystem path and the shipped module layout. Both fields ride a success body as **data**, so no HTTP boundary's 5xx message withhold can reach them — the disclosure had to be closed at the producer.

## The fix

The rule is **imported, never re-spelled**: `clientFacingFailureText` from `@objectstack/metadata-protocol`, the one implementation #8443 exported, called at both sites. A caught sentence is quoted only when the error declared itself a 4xx client refusal (ADR-0112); anything else gets the stable sentence the field could already carry, and the original goes to the server log.

**Both halves of the rule, because the two sites started in different states.** The flip already logged its cause in full at `error` with an operator remedy, so only its payload changed and that log is untouched. The announce had **no log line at all** — withholding alone would have converted an over-disclosure into a silent failure, which is worse — so it gains one naming the cause, the concrete consequence (a newly published record-triggered flow does not bind its trigger until the process restarts) and the fix (re-run the idempotent publish, or restart).

That new line is `warn`, not `error`, on three independent grounds: nothing here claimed to persist and did not (the drafts are published, the flip is stored — what is lost is an in-memory re-sync); AGENTS.md's own worked example of a functional degradation is verbatim "a trigger is not armed"; and the sibling announce of this same event, `MetadataPlugin._reloadAndAnnounce`, already logs it at `warn`. `announceKernelEvent` is not in the `check:durability-log-level` vocabulary, and escalating would be the over-application AGENTS.md warns trains everyone to skim `error`.

## The authored population — measured, not inherited

The judgement that decided the shape of #8333 and #8443: does either catch receive an **authored** population that declares nothing and would be blanked?

- **`unhideError`: yes, and it declares.** `saveMetaItem`'s refusals all carry 4xx — `NOT_OVERRIDABLE`/403, `ITEM_LOCKED`/403, `OBJECT_OVERLAY_PACKAGE_MISMATCH`/422, the org and destructive-change refusals — verified at each helper, and measured through the door: a locked app still tells its publisher which app and why, verbatim. The withhold blanks none of it, and the `unhiddenApps` half-flip report beside it is untouched.
- **`rebindError`: no.** `context.trigger` dispatch is PROPAGATING, and all three subscribers of `metadata:reloaded` are platform code doing internal re-sync — `resyncFlowsFromProtocol`, `resyncAuthoredHooks`/`...Actions`, `ingestReloadedObjects`, the authored translation sync. What arrives is internal text an author cannot act on. A guard case pins the bound anyway: if a subscriber ever does declare 4xx, the positive list quotes it with no change here.

So no fix was needed at the producer end, and no `safeParse`-style declaration was called for — unlike #8443, neither catch has a population that had to be made quotable.

## Tests

`packages/runtime/src/domains/packages-flip-announce-disclosure.test.ts`, 8 cases: a positive control, the three withholds, both halves of the log rule, and two `[GUARD]` cases pinning that a declared 4xx refusal is still quoted verbatim.

Reverse verification, both directions predicted **before** running:

- `domains/packages.ts` reverted to `origin/main` — predicted 4 failed | 4 passed, measured **4 failed | 4 passed**, every red failing on the text (`expected 'SQLITE_ERROR: no such table: sys_metadata' to be 'visibility flip failed'`).
- both call sites forced to withhold unconditionally (the "withhold everything" shortcut) — predicted 2 failed | 6 passed, measured **2 failed | 6 passed**, exactly the two guards. Without them this file would be satisfied by blanking every message, which deletes the self-correcting refusals the positive list exists to preserve.

Three pre-existing pins in `http-dispatcher.test.ts` asserted the raw driver text on `unhideError`. Each was triaged individually and **re-spelled, not weakened**: every one has a live subject that survives untouched (reports-rather-than-throws; the `error` log's consequence + fix; the #5242 mid-loop split report), and in each the injected error is a bare undeclared `Error` — the population the rule withholds. The driver sentence is still asserted in each, now in the log line where it belongs.

Swept out-of-package for anything pinning these two field names or this response's shape: only CHANGELOG prose and one same-package success-path assertion. No consumer pins either field.

## Verification

All at `ae47a64d5`, the final commit:

- `pnpm --filter @objectstack/runtime test` — **158 files, 2401 tests, all passing**; `typecheck` clean.
- Gate families re-derived from the actual changed paths with `scripts/pm/dispatch-gates.mjs`, all green: `check:nul-bytes`, `check:route-envelope`, `check:cross-package-test-inputs`, `check:changeset-gate-self-tests`, `check:objectui-changeset`, `check:query-options-erasure`, `check:type-check-coverage`, `check-adr-0087-registration`, `check-changeset-no-major`, `check-empty-changeset`, plus `check:durability-log-level` added because this diff adds a log line inside a `catch`.
- `check:type-check-debt` first **refused to run** (unbuilt closure — NOT MEASURED, not a pass). The full closure was built as `lint.yml` does and it was re-run: green, 33 ledger entries re-measured, none above its recorded number. Its one informational note (`@objectstack/lint` could be lowered by 1) is pre-existing and in a package this PR does not touch.

---

### DIFF
--- a/.changeset/runtime-publish-drafts-flip-announce-driver-text.md
+++ b/.changeset/runtime-publish-drafts-flip-announce-driver-text.md
@@ -1,0 +1,45 @@
+---
+"@objectstack/runtime": patch
+---
+
+fix(runtime): `publish-drafts` no longer discloses driver or subscriber text on `unhideError` / `rebindError` (#8516)
+
+`POST /api/v1/packages/:id/publish-drafts` answered, on a **200**:
+
+```json
+{ "success": true, "data": {
+  "unhideError": "SQLITE_ERROR: no such table: sys_metadata",
+  "rebindError": "TypeError: Cannot read properties of undefined (reading 'triggers') at AutomationPlugin.rebind (/srv/objectstack/packages/services/service-automation/dist/index.js:412:31)" } }
+```
+
+These are the two remaining producers on the response whose `seedApplied` field
+#8443 converted — the ADR-0045 visibility flip and the `metadata:reloaded`
+announce. Both ride a success body as **data**, so no HTTP boundary's 5xx
+message withhold can reach them; the disclosure had to be closed at the
+producer. Both were driven for real before being changed, and both reproduced.
+
+Both now follow the rule already in force next door: a caught sentence is
+quoted only when the error **declared** itself a client-facing refusal (4xx
+`status`, ADR-0112); anything else gets the stable sentence the field could
+already carry, and the original goes to the server log. The rule is imported
+from `@objectstack/metadata-protocol` (`clientFacingFailureText`), not restated
+locally.
+
+**Both halves of the rule, because the two sites started in different states.**
+The flip already logged its cause in full at `error` with an operator remedy, so
+only its payload changed. The announce had **no log line at all** — withholding
+alone would have converted an over-disclosure into a silent failure, so it gains
+one at `warn`, naming the cause, the concrete consequence (a newly published
+record-triggered flow does not bind its trigger until the process restarts) and
+the fix (re-run the idempotent publish, or restart). `warn` rather than `error`
+because nothing that claimed to persist failed to: the drafts are published and
+the flip is stored, and an unbound trigger is AGENTS.md's own worked example of
+a functional degradation — the level the sibling announce of this same event
+already uses.
+
+**Authoring feedback is preserved, not blanked.** The flip's authored refusals
+all declare 4xx (`ITEM_LOCKED`, `NOT_OVERRIDABLE`,
+`OBJECT_OVERLAY_PACKAGE_MISMATCH`, …), so a locked or non-overridable app still
+tells its publisher which app and why, verbatim — and the `unhiddenApps`
+half-flip report beside it is untouched. A subscriber that declares a 4xx
+refusal is quoted by the same positive list.
--- a/packages/runtime/src/domains/packages-flip-announce-disclosure.test.ts
+++ b/packages/runtime/src/domains/packages-flip-announce-disclosure.test.ts
@@ -1,0 +1,364 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * #8516 — the two remaining undeclared-driver-text fields on the publish-drafts
+ * 200 body: `unhideError` (the ADR-0045 visibility flip) and `rebindError` (the
+ * `metadata:reloaded` announce).
+ *
+ * Same response, same rule, same argument as #8443's `seedApplied` next door:
+ * both fields ride a **200** as DATA, and every 5xx message withhold in the
+ * stack reads a *thrown* error's message — none of them can see a field on a
+ * successful body. So every case here asserts the CONTENTS of that 200, never a
+ * status code and never that something threw.
+ *
+ * ## The defect, as measured on `origin/main` BEFORE the change
+ *
+ * The card was explicit that both sites were read from source, not reproduced —
+ * found by grepping for the string shape #8443's site used. They were driven
+ * for real first, through `HttpDispatcher.handlePackages`:
+ *
+ * | # | injection | pre-change field |
+ * |:--|:--|:--|
+ * | B | `getMetaItems` throws (the flip's app read) | `unhideError: "SQLITE_ERROR: no such table: sys_metadata"` |
+ * | C | `saveMetaItem` throws on app 2 of 2 | the same text, beside `unhiddenApps: ["crm"]` |
+ * | E | a subscriber throws on the announce | `rebindError: "TypeError: … at AutomationPlugin.rebind (/srv/objectstack/…/dist/index.js:412:31)"` |
+ * | D | a DECLARED 403 refusal on `saveMetaItem` | the authored sentence, verbatim — correct, and must stay |
+ * | F | a DECLARED 422 refusal on the announce | likewise |
+ *
+ * E is the worse of the two disclosures and the one no boundary could ever have
+ * caught: an internal stack frame plus a server filesystem path, handed to
+ * whoever pressed Publish.
+ *
+ * ## Both halves of the rule, because the two sites started in different states
+ *
+ * ADR-0112 is "quote only what declared a 4xx client refusal, and send the
+ * original to the log" — one rule with two halves. `unhideError`'s block
+ * already logged its cause in full at `error` (a durability seam: `saveMetaItem`
+ * is in the `check:durability-log-level` vocabulary), so only the payload half
+ * was open there and the log is left untouched. `rebindError` had **no log
+ * line at all**, so withholding alone would have converted an over-disclosure
+ * into a silent failure — strictly worse. Section 2 pins both halves so neither
+ * can be dropped later.
+ *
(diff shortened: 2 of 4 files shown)


qq9340100:

All four rulings hold, and the reproduce-first requirement earned its keep twice over.

**Ruling 1 — reproduced before fixing, and reality was worse than the card described.** Both sites driven for real through `HttpDispatcher.handlePackages` on `origin/main` *before* any change. The card called these "driver text"; `rebindError` measured as an internal stack frame **plus a server filesystem path plus the shipped module layout** (`/srv/objectstack/packages/services/service-automation/dist/index.js:412:31`). Grepping alone would never have shown that, and it is the field no boundary could ever have caught — it rides a 200 body as data.

**Ruling 2 — both halves, and the asymmetry handled exactly.** `unhideError`'s existing `error` log is untouched (payload only). `rebindError` had **none**, so it gains one naming cause, concrete consequence (*a newly published record-triggered flow does not bind its trigger until restart*) and fix. Withholding without that would have converted an over-disclosure into a silent failure — the specific trap in the dispatch.

The log-**level** reasoning is better than I would have specified: `warn` not `error`, on three independent grounds — nothing claimed to persist and did not (drafts published, flip stored; what is lost is an in-memory re-sync), AGENTS.md's own worked example of functional degradation is verbatim *"a trigger is not armed"*, and the sibling announce in `MetadataPlugin._reloadAndAnnounce` already logs at `warn`. Noting that `announceKernelEvent` is outside `check:durability-log-level`'s vocabulary, and that escalating would be the over-application which trains everyone to skim `error`, is the right instinct about a shared signal.

**The authored-population judgement — measured both ways, which is what decided #8333 and #8443.** `unhideError`: **yes and it declares** — `saveMetaItem`'s refusals carry 4xx (`NOT_OVERRIDABLE`/403, `ITEM_LOCKED`/403, `OBJECT_OVERLAY_PACKAGE_MISMATCH`/422, org and destructive-change), verified per-helper *and* measured through the door: a locked app still tells its publisher which app and why, verbatim. `rebindError`: **no** — `context.trigger` dispatch is PROPAGATING and all three `metadata:reloaded` subscribers are platform re-sync code. A `[GUARD]` case pins the bound anyway, so if a subscriber ever declares 4xx the positive list quotes it with no further change.

**The second ablation is the sophisticated one.** Reverting `packages.ts` (predicted 4 red / measured 4 red) shows the fix is present. Forcing both sites to withhold **unconditionally** (predicted 2 red / measured 2 red, exactly the guards) shows it is **not over-broad** — without those guards the file would be satisfied by blanking every message, which deletes the self-correcting refusals the positive list exists to preserve. Testing that a fix does not overshoot is rarer than testing that it exists.

**The three pre-existing pins were re-spelled, not weakened — verified from the diff.** Each keeps exact `toBe` equality; only the expected value moved, because the behaviour deliberately moved. In each the injected error is a bare undeclared `Error` — precisely the population the rule withholds — each pin's live subject survives untouched, and the driver sentence is still asserted, relocated to the log line where it now belongs. That is a pin tracking intended behaviour, not a demand being lowered.

`check:type-check-debt` refused on an unbuilt closure, was reported as **NOT MEASURED** rather than passed, then built and re-run green. Out-of-package sweep found no consumer pinning either field.

---

Issue #9600, opened 2026-08-18
[finding] Three data-model index rules put a positional config path in the finding's human-readable `where` slot, so the CLI renders the location as `objects[44].indexes[1]`


os-sam:
Recorded while taking the false-positive-budget measurement on #4716. Not fixed there — that dispatch changes no production file. Filed unassigned for triage.

## The deviation

`AuthoringFinding` declares two distinct slots (`packages/lint/src/authoring-rules.ts`):

- `where` — *"Human-readable location, e.g. `object "leave_request"`"*
- `path` — *"Config path, e.g. `objects[3].sharingModel`"*

Three registry adapters set **both** from the config path:

| rule | registry entry | adapter line |
|---|---|---|
| `lintUnscopedDeclaredIndexes` | `authoring-rules.ts` | `where: f.path` |
| `lintUniqueDeclarations` | `authoring-rules.ts` | `where: f.path` |
| `lintLegacyOrganizationComposites` | `authoring-rules.ts` | `where: f.path` |

All three implement in `packages/lint/src/data-model-rules.ts` and all three are `tier: 'advisory'`.

## Measured

`origin/main` @ `1c6da6eaf`, over the 45 object declarations `@objectstack/platform-objects` (44) and `@objectstack/metadata-core` (1) ship. 36 findings from these rules render their location as a bare index:

```
unique/unscoped-declared-index      where = objects[44].indexes[1]
unique/legacy-organization-composite where = objects[44].indexes[0]
```

The object's name is not in `where` at any point. It *is* in the message (`"sys_account" declares index [provider_id, account_id] with bare unique: true …`), so no information is lost — but the location column is unreadable on its own, and every other object-scoped rule in the table spells it `object "sys_account"`.

## Why this reads differently than when it was written

Two things changed under it:

1. **Advisories now reach a human.** `SaveMetaItemResponseSchema.advisories` ships (#4717) and objectui PR #4236 renders it. These strings are no longer CLI-only output beside a config file the author is looking at.
2. **All three carry `surfaceReason: RUNTIME_OBJECT_WRITES_P2`** — they are inside the group #4716's first bullet proposes to put on the `object` write path. At that surface the index is an index into a *per-write snapshot*, not into the author's stack: their own object is `objects[0]` (or the last slot) no matter where it sits in their tenant. The number would be not merely unhelpful but wrong-looking.

## A second, measured consequence

[...]

Issue #9237, opened 2026-08-17
app-showcase: two sharing rules declared on objects without record sharing enabled — boot backfill fails, rules never take effect


yinlianghui:
## Observed

Booting `examples/app-showcase` (`objectstack dev`) at `189a732254e6c46af75cfa38462693e67408d2d3` logs two WARNs during sharing-rule boot backfill:

- `share_open_tasks_with_manager`
- `share_red_projects_with_execs`

Both rules are declared on objects that do **not** have record sharing enabled, so the backfill fails and the rules silently never take effect.

## Why it matters

`app-showcase` is the kitchen-sink conformance fixture — its coverage manifest exists precisely so that every advertised capability is demonstrated working. A sharing rule that is declared but can never materialise undermines that contract: the security-chain demo appears to include rule-based sharing while the rules are inert at runtime.

## Expected

Either:
1. enable record sharing (`sharingModel` / sharing config) on the two objects so the rules take effect, or
2. if the declarations are intentional negative fixtures, register them as explicit waivers so the boot WARN is accounted for rather than looking like a defect.

## Repro

```
git checkout 189a7322
cd examples/app-showcase && pnpm dev
# watch boot log for the two sharing-rule backfill WARNs
```

Recorded as an unassigned finding (not claiming). Discovered during a read-only promo-video production run against the pinned SHA above; relayed by the coordinating PM session as the producing session has no GitHub scope on this repository.

os-zhuang:
Triage (first-touch grading): lands in `examples/app-showcase` sharing metadata — subsystem exercised is record sharing → `domain:identity`; promoted `finding` → `pm:queue`, type Bug.

Rationale: app-showcase is the conformance fixture — a declared sharing rule that can never materialise breaks the coverage-manifest contract, and the boot WARN is reproducible. Premise re-verified on `origin/main`: both rule names still present in `examples/app-showcase/src/security/sharing-rules.ts`.

Default route (PM mechanism assumption, measure before following): enable record sharing on the two target objects so the rules demonstrate the capability — that is the fixture's purpose. Register explicit waivers ONLY if the coverage manifest marks these as intentional negative fixtures; the implementer should check the manifest before choosing.

Size/model suggestion: S, sonnet.

---
_Generated by [Claude Code](https://claude.ai/code)_

os-project-manager:
[...]

Issue #8309, opened 2026-08-13
[2 of #7891] `permission`/`book` wiring: `TYPE_TO_STACK_KEY` entries + the cross-collection snapshot the three comparison rules need


hotlong:
Programme slice of #7891 (split by the `domain:spec` seat, session `session_01Euoy6wyfzgiWtgCg4s6JK2`).

## Scope (RE-PRICED 2026-08-13 — see the audit comment; the original `Blocked-by: #4463` was stale at filing)

Blocker B of the parent card, now including the snapshot enrichment itself: `RuntimeStackContext` carries `objects` only (measured by PR #7886, 2026-08-12 — AFTER #4463 closed), so the three cross-collection rules judge against a missing collection — measured 38 phantom `security-master-detail-ungranted` findings per-write vs 4 whole-stack. `permission`/`book` also have no `TYPE_TO_STACK_KEY` entries (pinned).

This slice delivers BOTH halves, bounded to what the three cross-collection rules need:
1. **Snapshot enrichment**: the runtime gate's stack view gains the sibling collections those rules compare against (permissions/positions/books as required — measure which), per #4463's own D2 design note (registry snapshot, built per-write or cached; measure the cost).
2. **The wiring**: `TYPE_TO_STACK_KEY` entries for `permission`/`book`, consuming the fuller snapshot.

## History note — why this is no longer "reference #4463"

#4463 closed 2026-08-02 (PR #4715) delivering its P1 (flow/approval/expression families at the runtime door) + the `surfaces` registry dimension. The P2 full-snapshot remainder (`RUNTIME_NEEDS_FULL_SNAPSHOT`) was NOT delivered by that PR and has no other open tracker — #7886's measurement six days later confirms the gap is live. This card is now that remainder's tracker, bounded to the #7891 programme's need.

## Re-check commands (premises age)

- `git grep -n "TYPE_TO_STACK_KEY" origin/main -- packages/lint/` (entries still absent?)
- `git grep -n "RUNTIME_NEEDS_FULL_SNAPSHOT\|RuntimeStackContext" origin/main -- packages/lint/ packages/metadata-protocol/` (snapshot still objects-only?)
- `git log --oneline -5 origin/main -- packages/lint/src/runtime-gate.ts`

## Notes

- ⛔ No partial `security-role-word` wiring — whole family or not at all (slice 3 / #8310's call).
- Serial constraint: touches `packages/lint`'s runtime gate — serialize behind PR #8390 (#8307) landing.
[...]

Issue #7503, opened 2026-08-11
Publish-time lint: `sharingModel: controlled_by_parent` with no `master_detail` relation is statically detectable and unreported


os-zhuang:
Found while implementing #7474 (PR splitting the six `assertControlledByParentWrite` refusal legs). Filed rather than fixed: the maintainer ruling of 2026-08-11 on #7474 ordered the runtime split only, and explicitly left question 2 of that card — lint vs runtime — undecided. This is that question, filed so it is not lost.

## What was measured

An object that declares `sharingModel: 'controlled_by_parent'` and has no relation for the platform to derive access from is now refused at RUNTIME with `422 INVALID_METADATA` (the #7474 PR). Nothing reports it before that:

- `packages/lint/src/validate-security-posture.ts` carries the adjacent rules — `security-owd-unset`, `security-owd-alias`, `security-master-detail-ungranted` (a detail object that no authored permission set grants), `security-private-no-readscope` — and none of them checks this one. Grep for `controlled_by_parent` in `packages/lint/src` returns only the OWD vocabulary list, the D11 comment, and the `security-master-detail-ungranted` docblock.
- The check needs nothing a lint pass does not already hold: the object document carries `sharingModel` and `fields`, and the relation resolution is the same three-step fallback `resolveCbpRelation` performs at runtime (`security-plugin.ts`) — a required `master_detail`, then any `master_detail`, then a required `lookup`. An object matching none of the three has nothing to derive from.

## Why it is worth a rule

The runtime refusal fires when a user attempts a write on that object — i.e. after the app is published and in someone's hands, and only on the detail objects a caller happens to touch. The defect exists from the moment the metadata is authored. This is the shape `security-master-detail-ungranted` already exists for, one gate earlier.

It matters most for AI-authored metadata: `controlled_by_parent` is a plausible thing for an agent to write next to a lookup it forgot to mark `required`, and nothing in the authoring loop says so.

## Suggested shape

A new advisory rule alongside `SECURITY_MASTER_DETAIL_UNGRANTED` in `validate-security-posture.ts` — `error` rather than `warning` is arguable, since the runtime answer is already a hard refusal, and that severity choice is the one thing this needs a decision on.

## Pointers

- `packages/lint/src/validate-security-posture.ts` — the rule family and its existing master-detail rule
[...]

PR #9890, opened 2026-08-19
lint: a sharing rule anchored where sharing has nothing to widen is an authoring-time error

### LINKED ISSUE #9698: lint: a sharing rule declared on an object whose effective sharing model is `public` is statically detectable and unreported until boot
Observation filed from #9237 (PR #9697). Not claiming — recording the class the fix there closed only for `examples/app-showcase`.

## What is detectable and is not detected

`SharingService.inertGrantReason` (ADR-0111 D7, `packages/plugins/plugin-sharing/src/sharing-service.ts`) refuses a grant whose row no gate could ever consult. Two of its arms are decidable from the authored metadata alone, before anything boots:

- `effectiveSharingModel(schema) === 'public'` — the object declares `sharingModel: 'public_read_write'`, so sharing has nothing left to widen;
- `sharingModel: 'controlled_by_parent'` — the detail's shares belong to its master.

Today a rule in either state is accepted by `SharingRuleSchema`, accepted by `defineRule`, seeded into `sys_sharing_rule`, and only then refused, once per boot, as a WARN line inside the boot diagnostics block:

```
WARN SharingServicePlugin: boot rule backfill failed for rule {"rule":"…","error":"SHARING_NOT_ENABLED: '…' is not under record-sharing enforcement …"}
```

## Why the boot WARN is not sufficient as the diagnostic

Measured on the stock showcase at `c07d6e8b9` while working #9237: **three** rules were in this state and only **two** warned. The third, `share_high_value_red_projects_with_managers`, had a compound condition matching no seeded row, so `SharingRuleService.reconcile` built an empty desired set, never reached `grant`, and never threw. It was exactly as dead as the other two and produced no diagnostic at all — the WARN is a function of the data, not of the declaration.

So the boot log cannot be the guard: an app whose seed data happens not to match a rule's criteria gets silence, and an app that later gains matching rows gets a WARN it did not get on the previous deploy.

## Suggested shape (not prescribed)

An authoring/publish-time lint rule resolving each `sharingRules[].object` against the declared objects and refusing the two decidable arms — the same shape and the same reasoning as #7503 (`sharingModel: controlled_by_parent` with no `master_detail` relation), which is the nearest precedent for a statically-decidable sharing declaration made loud at publish time rather than at boot.

Deliberately **not** in scope for such a rule: the `owner_id` arm. `owner_id` is injected by the schema registry (`packages/objectql/src/registry.ts`), so it is absent from authored metadata and present on the runtime schema; asserting it statically would fail every object that correctly does not declare it by hand.

## What #9697 does and does not cover

#9697 adds `inert-wirings.test.ts` section 6 in `examples/app-showcase` — a guard over that app's own declarations only, plus a second arm checking the rule's audience holds `allowRead` on the shared object. It closes the class for the reference app. It does nothing for any other app, and nothing at authoring time, which is what this card is about.

os-steve:
Fixes #9698

Verified at `9753e5038` (every reading below is from that commit).

## H1 — is the effective sharing model visible at lint time? **Yes.**

This was the hypothesis that could have reshaped the card, so it was settled first. `effectiveSharingModel` reads exactly three things, and every one of them is authored metadata that survives the trip to the runtime unchanged:

| input | authored? | trap checked |
| --- | --- | --- |
| `sharingModel` | yes | `z.enum([...]).optional()` on `ObjectSchema` — **no `.default()`**, so authored-vs-absent stays distinguishable after parsing |
| `security.sharingModel` | unreachable | `ObjectSchema` is strict and declares no `security` key, so a stack nesting the OWD there is REFUSED, not stripped (`owdOf` in `validate-security-posture.ts` records the same finding) |
| `isSystem` / `sys_` prefix | yes | `isSystem` carries `.default(false)`, which is why the mirror reads `=== true` exactly as the runtime does — the default and an explicit `false` are the same verdict |

The `.default()` check was the specific thing worth measuring: it is what erased the authored-vs-absent distinction on a neighbouring card and turned an intended 1 rejection into 96. It is absent here, so the linter and the service answer from the same bytes.

Nothing is resolved at boot, from org config, or injected. The rule as stated is writable.

## H3 — an arm on a live rule, not a new one

`packages/lint` already has this rule family. `validateSharingRuleEnforceability` judges `sharingRules[].condition`; the new arm judges `sharingRules[].object`. Same surface, same finding shape, same severity bar, same `sharing-rule-*` id prefix, same CLI-only wiring — it inherits all of it, and there is no second rule to disagree with the first about one object.

The alternative home was `validateSecurityPosture` (where #7503's `security-controlled-by-parent-no-relation` lives). Rejected on a measured ground: that entry declares `runtimeTypes: ['seed', 'permission', 'book', 'object']`, and `CONTEXT_STACK_KEYS` is `['objects', 'permissions', 'books', 'datasets']` — the per-write snapshot does **not** carry `sharingRules`. Adding an id there that can never fire at a door the entry claims is exactly the #7220 split.

**Registry edit, flagged for serialisation as asked.** `packages/lint/src/authoring-rules.ts` is touched, and PR #9825 touches it too. The change is confined to this rule's entry: its `surfaceReason` recorded that the rule *"reads ONLY `stack.sharingRules[].condition` and needs no other collection"*, which the new arm **falsifies** — it is now cross-collection. The corrected text records that `objects` is already carried (#8309) while `sharingRules` is not, so the crossing story keeps its shape but is now two collections rather than one. (#9600's branch had an empty diff against `origin/main` at claim time, so it declared no surface to collide with.)

## Ruling 1 — INERT or WRONG? **Measured, and the two arms differ.**

Rather than assume, a real `SharingService` over an in-memory engine was driven through `grant` + `buildReadFilter` on all three postures:

| anchor | `grant` | `buildReadFilter` | share rows |
| --- | --- | --- | --- |
| `public_read_write` | refused — `SHARING_NOT_ENABLED: ... is not under record-sharing enforcement` | `null` | 0 |
| `controlled_by_parent` | refused — `SHARING_NOT_ENABLED: ... is controlled by its parent (master-detail); share the master record instead` | `null` | 0 |
| `private` (control) | **accepted** | `{"$or":[{"owner_id":"u_manager"},{"id":{"$in":["d1"]}}]}` | 1 real row |

The two failing arms are not the same shape, and that is why they get different wording and different ids:

- **`public` OWD** — no filter is applied at all, so the intended audience already reads every row and so does everyone else. Nobody under-sees; the harm is that the declaration **advertises a restriction that does not exist**.
- **`controlled_by_parent`** — the detail's visibility comes from its master's path (ADR-0055), not from this rule, so the author **believes a grant exists when it does not** and the named recipient may see nothing.

A single id with a branchy message would have handed half of these authors the wrong fix.

## H2 — blast radius, counted before choosing the severity

Counted through the production `objectstack build` gate, not by grep. **5** sharing rules are declared anywhere in this repo (2 in `app-showcase`, 3 in `app-crm`; the `create-objectstack` template declares none, and the `packages/spec/src/conversions/registry.ts` occurrences are conversion fixtures, not shipped stacks).

**3 fire — all of them in `examples/app-crm`:**

```
✗ Author-time rules failed (3 issues)
  • sharing rule "share_high_value_opps_with_managers" on object "crm_opportunity"
      rule: sharing-rule-object-not-shareable  at sharingRules[0].object
  • sharing rule "share_active_leads_with_manager" on object "crm_lead"
      rule: sharing-rule-object-not-shareable  at sharingRules[1].object
  • sharing rule "share_won_deal_activities" on object "crm_activity"
      rule: sharing-rule-object-not-shareable  at sharingRules[2].object
```

All three are anchored on `public_read_write` objects. They have been failing their boot backfill on every boot of that app since they were written. **2 stay silent** — `app-showcase`'s, both on `private` objects, which is the direction that had to be proven rather than hoped for.

3 is the "fix them in the same PR" band, so they are repaired here.

### How they are repaired, and the option that was measured and rejected

The three are **removed** under ADR-0049 enforce-or-remove — the same call #9237 made for the two equivalent `app-showcase` rules.

The alternative — tighten the three objects' OWD to `private` so the rules become live — was not dismissed on taste; it was built and measured. It clears the 3 errors, and then trips a gate that exists precisely to stop me:

```
✗ Access matrix drift (4 changes) — capability changes must be reviewed
  • 'crm_activity' record baseline (OWD): public_read_write → private (affects every principal)
  • 'crm_lead' record baseline (OWD): public_read_write → private (affects every principal)
  • 'crm_opportunity' record baseline (OWD): public_read_write → private (affects every principal)
  If intended, re-run with --update-access-matrix and commit the snapshot — its diff IS the review artifact.
```

Changing three objects' record baselines "affecting every principal" is a reviewed capability change (ADR-0090 D6), not a rider on a lint card. It is written up as an open question for the maintainer instead, and the CRM config carries a comment saying so.

## Ruling 2 — the diagnostic text, and the test that was green for the wrong reason

Checked before writing: nothing outside this package keys on these rule ids (no docs catalogue, no allowlist), and the wiring guard's only constraint on `surfaceReason` is a length floor.

The more useful find was in the opposite direction. `examples/app-crm/test/smoke.test.ts` asserted:

```ts
expect(rules.length).toBeGreaterThanOrEqual(2);
expect(rules.every((r) => r.type === 'criteria')).toBe(true);
```

Both passed, for years, while **all three** rules enforced nothing — it asserted the enforced *form* and never the property that mattered. It is replaced by the property its greenness hid (no rule anchored where sharing has nothing to widen), so re-adding one goes red in the app's own suite as well as at `objectstack build`.

Each new message names the object, the rule, and why it cannot take effect, and gives both honest fixes because which one is right is the author's call:

> Decide which half is wrong. If the ACCESS is right — everyone should read and write these records — the rule is dead metadata: delete it (ADR-0049 enforce-or-remove). If the RULE is right — only the named audience should reach these records — then "crm_opportunity"'s OWD is the defect: set sharingModel: 'private' (owner + shares) or 'public_read', and this rule starts enforcing. Do NOT re-home the rule onto another public object; that moves the inertness instead of removing it.

## H4 — both directions, and the ablation

**Fires on broken metadata** (above) · **silent on correct metadata**: `app-showcase` builds green with all 41 author-time rules running and zero sharing findings.

Unit coverage pins the silence individually, because a rule that only proves it fires is half-tested: `private`, `public_read`, a **custom** object with no OWD (fails closed to `private`, so the rule is live), a **retired OWD alias** (`'read'` — the runtime fails closed to `private`, so firing would be a false positive on top of the `security-owd-alias` error the value already earns), an anchor the stack does not declare, and an object with no authored `owner_id`.

**Ablation** (predicted direction declared first: the diagnostic *disappears* on defective input — it is not a test that turns red). Restoring `origin/main`'s lint source with the CRM's three rules back in place, **rebuilding `@objectstack/lint` so the CLI's `dist/` actually carried the ablation**, and proving the ablation reached the artifact (`grep -c` for the new rule id in `packages/lint/dist/index.js` → `0`):

```
examples/app-crm build:   ✓ Build complete (571ms)
```

Green, on metadata that fails at boot. That is the bug this card reports, and it confirms the new arm is what catches it rather than something else in the registry. Restoring and rebuilding puts the id back in `dist/` (`grep -c` → `1`).

## Deliberately not judged

`inertGrantReason`'s other arms are excluded by name, in the code and in the changeset: the **`owner_id`** arm (registry-injected, absent from authored metadata by design — asserting it would fail every object that correctly does not declare it by hand; the filer's scoping, adopted), the **`bypassObjects`** arm (plugin configuration, not stack metadata, so it is not in this door's input), and the **federated phantom-anchor** arm (a provenance test over that same injected column). An unresolvable `rule.object` is also not reported — absence of a schema is absence of evidence of inertness, and the runtime draws the same line by keeping existence a separate verdict from inertness.

## Scope note

`content/docs/permissions/index.mdx` presented `HighValueOpportunitySharingRule` as *"a real sharing rule from the CRM example app"* — i.e. the documentation taught authors to write the exact declaration this PR makes a build error, and that symbol no longer exists after the repair. It is repointed at `KeyAccountQualifiedContactRule`, which is real, live, on a `private` object, and already guarded by `inert-wirings.test.ts`; a short paragraph explains why the anchor's OWD is the load-bearing part. Named here rather than slipped in, per the bounded-fix discipline. `content/docs/releases/**` is untouched.

## Verification at `9753e5038`

- `pnpm --filter @objectstack/lint test` → **75 files, 2109 tests passed**
- `pnpm --filter @objectstack/example-crm test` → 4 files, 42 passed · `@objectstack/example-showcase test` → 21 files, **337 passed**
- `pnpm --filter @objectstack/example-crm --filter @objectstack/example-showcase build` → both `✓ Build complete`
- `pnpm --filter @objectstack/lint typecheck` → clean
- Gates re-derived from the **actual** changed paths (`node scripts/pm/dispatch-gates.mjs`), all PASS: `check:nul-bytes`, `check:changeset-gate-self-tests`, `check:cross-package-test-inputs`, `check:doc-anchors`, `check:docs-audit-scope`, `check:docs-redirects`, `check:objectui-changeset`, `check:published-readme-links`, `check:role-word`, `check:type-check-coverage`, spec `check:empty-state` / `check:liveness` / `check:strictness-ledger` / `check:variant-docs`, and `check-adr-0087-registration` / `check-changeset-no-major` / `check-empty-changeset` / `check-cross-package-test-inputs` / `docs-audit/check-affected-docs`. Convention-triggered by the test edits and also PASS: `check:engine-double-contract`, `check:where-matcher`, `check:query-options-erasure`.
- A ratchet caught a real omission mid-run: `rule-id-barrel-exports.test.ts` failed until both new ids were exported from `packages/lint/src/index.ts`.

`Clause-②: yes` — this changes authoring-time accept/reject behaviour, declared on the claim comment.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XqDQYVU5smx29ts9pAErja)_


os-elon:
## Contract review (Clause-② gate) — **PASS**, `needs:contract-review` cleared

**Reviewer eligibility, on the record:** maintainer-directed PM session `session_01KspKtTajJmy3J37UhH8tcH`, acting on the maintainer's direct instruction of 2026-08-19 (verbatim: 「再帮我审核 … #9698 PR #9890 draft ✅ 按设计——扣住等契约复审」). Both structural conditions verified before this review: ① serving model self-checked via `get_session` = `claude-fable-5` (`session_context.model` and `last_served_model` both) — contract-review tier; ② this session is **not** the dispatching seat (the devx seat dispatched at opus and correctly refused to self-clear).

**Scope reviewed — the contract increment only** (the accept/reject change), against the actual diff at `9753e5038`:

- **The tightening rejects only what is genuinely dead.** Two new `error` ids mirror `inertGrantReason`'s two authored-metadata-decidable arms; the mirror preserves authored-vs-absent (`sharingModel` has no `.default()` — verified in the diff's own analysis, the exact #9689-family trap), fails CLOSED on unrecognised values (a retired alias resolves to `private` ⇒ live rule, not reported), and tests `controlled_by_parent` **before** the public collapse, mirroring the runtime's own order so the author gets the right fix-it.
- **The negative direction is proven, not hoped:** silent on `private`, `public_read`, custom-absent-OWD, retired alias, unresolved anchor, and ownerless objects — each pinned by its own test; the excluded arms (`owner_id`, `bypassObjects`, federated) are excluded by name for the right reason (not decidable from authored metadata).
- **Blast radius measured through the production gate:** 5 declared rules repo-wide, 3 fire (all `app-crm`, all provably inert since written — refused at boot backfill on every boot), repaired in-PR under ADR-0049 with the #9237 precedent. The rejected alternative (tightening three OWDs to `private`) was built, measured, and correctly refused as an ADR-0090 D6 access-matrix change that must not ride on a lint card.
- **Ground truth is a real `SharingService` measurement** (grant + `buildReadFilter`, three postures incl. the `private` control), and the ablation was proven to reach the built artifact (`grep -c` of the id in `dist` → 0) before concluding — the confirmation-independent-of-exit-code discipline.
- **Docs stay consistent with the new contract:** `permissions/index.mdx` no longer teaches the newly-rejected shape; the falsified `surfaceReason` in the registry was corrected in the same PR.

One maintenance note, non-blocking: `effectiveSharingModelOf` is a copied mirror of the runtime function (per the package's no-runtime-dependency discipline, same as `resolveCbpRelation`); if the runtime's OWD resolution ever changes, this copy must move with it — the in-file doc says so.

**Conclusion: the accept-set narrowing is faithful to the runtime verdict and correctly bounded. Cleared for the queue.** The dispatching seat may now flip ready and enqueue through the merge queue as normal (`GOVERNED_HITS=NONE`, so no human-merge hold applies).

---

Issue #10349, opened 2026-08-20
[finding] Anonymous callers to every better-auth-native /admin/ route get a bodyless 401 — no envelope, no machine-readable code


os-warren:
Found while implementing #9968 (impersonation half). Filed unassigned.

Same class as #10345 (`a refusal reaches the client with the status but no machine-readable code`), different surface: that one is the `/api/v1/data` write path, this one is the auth `/admin/` namespace.

## Measured

Stock `better-auth@1.7.1` + its `admin` plugin, anonymous POST, no session:

```text
/admin/impersonate-user  anonymous -> 401 body=""
/admin/set-role          anonymous -> 401 body=""
```

The body is the empty string — not an empty JSON object. It comes from the vendor's `adminMiddleware`, which is `getAuthoritativeSessionFromCtx(ctx)` followed by `APIError.fromStatus("UNAUTHORIZED")`, and every `/admin/` route that mounts that middleware inherits it.

## Why it is worth recording

The `/api/v1/auth/admin/` namespace answers the same question in two different shapes depending on which lane a route happens to be on:

- **ObjectStack raw mounts** (`create-user`, `set-user-password`, `unlock-user`, `import-users`, `ban-user`, `unban-user`, `oauth2/toggle-disabled`, `sso/*`) answer `401` with the ADR-0112 envelope and `code: 'UNAUTHENTICATED'` (see `platform-admin-gate.ts`).
- **better-auth-native routes** answer `401` with nothing at all.

A client cannot branch on the refusal without knowing, per route, which implementation happens to serve it — and that split is an implementation detail, not a contract.

This asymmetry is already *known* and encoded in `packages/qa/dogfood/test/admin-route-nonadmin-refusal.dogfood.test.ts`: the ObjectStack bucket asserts `anon.status === 401 && anon.code === 'UNAUTHENTICATED'`, while the better-auth bucket can only assert `[401, 403].includes(anon.status)` — there is no code to assert. So the suite documents the gap rather than closing it, and nothing tracks closing it.

## Not fixed by #9968, deliberately

#9968 re-authorizes `/admin/impersonate-user` as a better-auth plugin endpoint, changing only the **authorization** predicate. `adminMiddleware` (authentication) is passed through untouched, so the bodyless 401 is unchanged there and is pinned as-is rather than dressed up.

Options, no recommendation made unilaterally — this is a public refusal contract:

[...]

Issue #10348, opened 2026-08-20
[finding] Three spellings of the ADR-0068 platform-admin read now live in plugin-auth, and one of them skips the system read context


os-warren:
Found while implementing #9968 (impersonation half). Filed unassigned; no fix attempted — consolidating a security predicate is not that card's scope.

`platform-admin-gate.ts`'s own header states the principle this violates:

> Before this module the gate existed as four near-identical inline copies … N copies of an authorization predicate is the shape that drifts: the next mount is written by copying whichever copy the author happened to open. One exported judge, called by every mount, is the fix.

That consolidation covered the *session-shaped* question (`isPlatformAdminUser(sessionUser)` / `judgePlatformAdmin(session)`). The **id-shaped** question — "is this user id a platform admin?", i.e. the `sys_user_permission_set` → `admin_full_access` / `organization_id = null` lookup — was never consolidated, and now has three spellings in one package:

| # | Site | Read path |
|---|---|---|
| 1 | `auth-manager.ts` — the `isPlatformAdmin()` closure inside `customSession` | `dataEngine.find(...)` **directly** |
| 2 | `auth-manager.ts` — `isOrgOrPlatformAdmin()`'s first half | `withSystemReadContext(engine).find(...)` |
| 3 | `auth-manager.ts` — `isPlatformAdminUserId()` (added by #9968) | `withSystemReadContext(engine).find(...)` |

All three do the same two reads with the same `limit: 50` and the same `admin_full_access` name match, and all three fail closed.

**The one substantive difference is #1.** It queries the engine **without** `withSystemReadContext`, while #2 and #3 both wrap. `isOrgOrPlatformAdmin`'s docblock says why the wrapper is there:

> Reads through `withSystemReadContext` so the lookups are not themselves RLS-scoped to the acting (possibly non-privileged) user.

Whether that matters for #1 depends on whether `customSession`'s callback already runs in a privileged context. **This finding does not claim #1 is a bug** — that is the measurement someone should take before acting, and it is exactly the kind of difference that is invisible while three copies exist. If the wrapper is unnecessary in #1, the two wrapped copies are carrying a cost for nothing; if it is necessary, #1 is a session-derivation path that can under-report platform admin for a caller whose RLS scope hides the join rows.

`resolve-authz-context.ts` is named as authoritative for the platform-admin half by `isOrgOrPlatformAdmin`'s docblock, which suggests a fourth spelling may exist outside this package — worth counting as part of the same sweep.

[...]

Issue #10159, opened 2026-08-20
SettingsService accepts a write before its engine is bound and answers "resolved" while nothing reaches sys_setting — every `kernel:ready` hook registered from `init()` is inside that window


os-elon:
Found while measuring #9451's A′ preconditions (that card needed to know whether a `kernel:ready` migration can write a durable receipt through `SettingsService`). Filing separately rather than widening that card. Not a live defect today — no shipped caller writes settings at that moment — so this is a latent trap, recorded before someone walks into it.

## Observation

`SettingsService` keeps an in-process `memory` array as its fallback store, and `upsertRow` picks between the engine and that array on `if (this.engine)`. The engine is bound in exactly one place — `SettingsServicePlugin.start()` registers a `kernel:ready` hook, and `bindEngine` is called inside it (`packages/services/service-settings/src/settings-service-plugin.ts:202`).

`kernel:ready` handlers run in **registration order**, and every plugin's `init()` runs before any plugin's `start()`. So any hook registered during `init()` fires **before** the settings plugin has bound its engine — and a `set()` from such a hook returns a fully resolved value while nothing reaches `sys_setting`.

That is not a hypothetical position: `assembleMetadataProtocol` registers its three platform migrations' `kernel:ready` hook from `ObjectQLPlugin.init()` (`packages/objectql/src/plugin.ts:350` → `packages/metadata-protocol/src/plugin.ts:306`), which is the exact window.

## Measured

Real `LiteKernel`, real `ObjectQLPlugin` over a real `SqlDriver` (better-sqlite3), real `SettingsServicePlugin`, and a probe plugin registering its `kernel:ready` hook from `init()`:

```
serviceResolvableAtReady : true
engineBoundAtReady       : false
setAtReady               : resolved:"written-at-kernel-ready"     <- reports success
rowsInDbAfterBoot        : []                                     <- nothing persisted

engineBoundAfterBoot     : true                                   <- control, same boot
rowsInDbAfterControlWrite: [ { namespace: 'receipt_probe', key: 'last_run',
                               scope: 'global', organization_id: null, … } ]
auditRowsAfterControlWrite: [ { action: 'set', new_hash: 'fnv32:…' } ]
```

The control write moments later lands a real `sys_setting` row **and** a real `sys_setting_audit` row on the same connection, so the table existed throughout — the timing is the whole cause.

## Why it is worth recording

[...]

Issue #10025, opened 2026-08-19
Should a definition-level input-schema refusal be non-retryable? #9889's parity floor makes a flow burn its whole retry budget on a verdict that cannot change


os-warren:
Derived from #9889 (PR #10024) as the explicitly-undecided half. Filed unassigned — this is a ruling request, not work in progress. **The parity floor already shipped and is correct under either answer**; this card is only about what happens on top of it.

## What is settled

#9889 restored the invariant: `validateNodeInputSchemas` is now called by **both** attempt paths (`execute()` and `executeWithoutRetry`), so a flow whose node `config` violates its own declared `inputSchema` is refused on every attempt instead of being refused once and then run for real on retries. That defect — a `retry` strategy acting as a way past authoring-time validation — is closed.

## What is not settled

The guard's verdict is a **pure function of the flow definition**: its signature is `validateNodeInputSchemas(flow: FlowParsed, _variables: Map<string, unknown>)` with the variables parameter deliberately unused, and it reads `node.inputSchema` against the static `node.config`. Re-running it on attempt 2 cannot produce a different answer than attempt 1.

So under the shipped parity floor, a `strategy: 'retry'` flow with a mis-declared `inputSchema`:

1. burns its **entire retry budget** — including configured backoff delays — re-deriving a certainty;
2. writes `maxRetries + 1` identical failed run-log rows, each carrying the same guard message, which is what a run-history reader has to page through to find the one fact that matters.

## The two options

**A — keep the parity floor as shipped.** Every attempt dispatches and is refused; retry accounting is untouched; run history shows N identical refusals. Simplest, already landed, zero further risk.

**B — classify a definition-level guard failure as non-retryable**, so `execute()` never hands off to `retryExecution` for it at all. Refuse once, skip the retry loop.

B is not free, and that is why it is here rather than in the PR:

- It **changes retry accounting** — an operator watching retry counters or run-log volume sees different numbers for the same flow after the change.
- It likely **changes the result classification**. Today the refusal carries `status: 'failed'` and no ADR-0112 `code` on either path, because #9378's classification gives `code` to the **never-dispatched** exits (`FLOW_DISABLED`, `FLOW_NO_START_NODE`) and `status: 'failed'` to dispatched-and-failed exits. A non-retryable definition-level refusal arguably *becomes* a never-dispatched exit and would deserve its own `code` — which is a public envelope addition, not an internal tidy-up.

## The dev's recommendation (#10024, recorded verbatim in intent)

[...]

PR #10352, opened 2026-08-20
fix(plugin-auth): admit ObjectStack platform admins on /admin/impersonate-user (better-auth plugin endpoint, not a raw mount)

os-warren:
Part of #9968

> ⛔ **Do not auto-merge.** This PR is developed one tier below the contract-review gate under the maintainer's 2026-08-20 fable-exhaustion authorization, whose compensating control is that **triage reviews before merge**. It must not be flipped ready and must not have auto-merge armed until that review happens.

## ⚠️ Public-surface narrowing — read this first (Clause-② card)

This card was ruled in two halves. **Only the impersonation half ships here.** The half that narrows the public surface — retiring the `set_user_role` action — is **NOT in this PR**, because it trips a red line the dispatch drew (see *The `set_user_role` half is blocked*, below). Nothing is removed from the public surface by this PR.

One refusal does become **newly reachable**, which is the closest thing here to a narrowing and is called out rather than buried:

| Caller | Target | Before | After |
|---|---|---|---|
| ObjectStack platform admin (ADR-0068) | ordinary user | `403 YOU_ARE_NOT_ALLOWED_TO_IMPERSONATE_USERS` | **`200`** |
| ObjectStack platform admin | **another platform admin** | `403 YOU_ARE_NOT_ALLOWED_TO_IMPERSONATE_USERS` | **`403 YOU_CANNOT_IMPERSONATE_ADMINS`** |
| plain member / org owner / org admin | anyone | `403 YOU_ARE_NOT_ALLOWED_TO_IMPERSONATE_USERS` | unchanged |
| anonymous | anyone | `401` (bodyless) | unchanged |
| caller whose legacy `user.role === 'admin'` | ordinary user | `200` | unchanged |

The middle row is the only case where a caller who could previously have succeeded now cannot — and on any post-ADR-0068-D2 deployment it was already unreachable, because reaching it required the caller to pass the vendor's gate, which no platform admin does. It is listed because the *rule* is new even where the outcome is not.

## The defect

better-auth's `admin` plugin authorizes every `/admin/*` route through `hasPermission({ userId, role, options, permissions })`, whose only two authorization inputs are a construction-time `adminUserIds` array and the persisted legacy `user.role` scalar. ADR-0068 D2 stopped synthesizing that scalar, and re-synthesizing it is permanently vetoed (maintainer ruling, 2026-08-18). ObjectStack's platform admin is a `sys_user_permission_set` row pointing at `admin_full_access` with `organization_id = null` — neither input.

Measured on the installed `better-auth@1.7.1`, a signed-in caller whose `sys_user.role` is `'user'`:

```text
/admin/impersonate-user -> 403 {"message":"You are not allowed to impersonate users","code":"YOU_ARE_NOT_ALLOWED_TO_IMPERSONATE_USERS"}
/admin/set-role         -> 403 {"message":"You are not allowed to change users role","code":"YOU_ARE_NOT_ALLOWED_TO_CHANGE_USERS_ROLE"}
```

A platform admin and a plain member receive byte-identical refusals, so the `sys_user` "Impersonate User" button is dead on every deployment. Fails closed.

## The probe the ruling required, and its answer

The ruling made this measure-first: probe whether 1.7.1's `checkEndpointConflicts` permits overriding a path another plugin registers, and **stop and report if it fails**. It did not fail.

- `checkEndpointConflicts` (`dist/api/index.mjs`) builds its registry by iterating `options.plugins[].endpoints`, and on a duplicate path+method calls `logger.error(...)`. **It does not throw.**
- `getEndpoints` merges plugin endpoints with `{...acc, ...plugin.endpoints}` (keyed by endpoint *key*), and `better-call`'s router calls rou3 `addRoute` per endpoint in object order, where a later entry for the same method+path replaces the earlier one.

Measured end to end against a real `betterAuth()` instance: a second plugin registering `/admin/impersonate-user` boots, serves the override (`200`, override marker in the body) — **and logs `Endpoint path conflicts detected!` on every start.**

So the override is permitted, and this PR takes the strictly better door the same measurement opens: **replace the endpoint in place on the `admin` plugin's own `endpoints` record**, so exactly one plugin ever registers the path. No conflict log, same endpoint context, same path.

## The shape — plugin endpoint, never a raw Hono mount

⛔ A raw Hono mount is forbidden by the ruling, for two independent reasons this implementation respects:

1. The handler is not a data write. It mints a session and rewrites cookies via helpers that exist only inside a better-auth endpoint context, and the `admin_session` payload is a **contract** with `/admin/stop-impersonating`, which parses `adminCookie.split(':')` and 500s if the shape is off. A hand-rolled signature is either a broken exit path or a forgeable cookie.
2. Shadowing the path detaches every better-auth hook keyed on it — including `rotateCallerBearerOnImpersonation` (#8243). Without it, `bearer()` converts the caller's token back into the admin's session on every later request and impersonation is a **silent 200 no-op**. Nothing about the endpoint's own response would change, so no existing test would have noticed.

The replacement endpoint is rebuilt from the **vendor endpoint's own `options` object** (`method`, `body` schema, `use: [adminMiddleware, …]`, OpenAPI `metadata`), passed through untouched. The request contract, the 401-for-anonymous and the body validation are the vendor's and cannot drift from it on a dependency bump, because there is no second copy to drift.

## What actually changed

Only the authorization, in the two places the vendor asks it:

1. **Caller.** `hasPermission({ role: session.user.role, … })` → the ADR-0068 D2 predicate. The legacy `role === 'admin'` reading is retained exactly as `platform-admin-gate.ts` retains it, so a deployment still carrying the pre-D2 scalar is not locked out. The caller set only ever grows.
2. **Target.** The vendor refuses to impersonate an admin-grade target by reading `targetUser.role` against `adminRoles: ['admin']` — a column nothing writes post-D2, so that guard is **inert**: "you cannot impersonate admins" is currently a promise the code does not keep. It is re-asked through the same ADR-0068 predicate. The protected-target set only ever grows.

The vendor's `allowImpersonatingAdmins` / `impersonate-admins` escape is deliberately **not** carried over: ObjectStack constructs `admin({ schema })` and configures neither, and the escape's own check reads the same dead scalar.

Refusals keep better-auth's flat `{ message, code }` shape and the vendor's own code constants, read off the plugin's `$ERROR_CODES` rather than retyped — so **no new public error code is minted** and nothing here reaches the spec error-code ledger.

## The `set_user_role` half is blocked — measured, not assumed

**Premise confirmed.** The vendor's `/admin/set-role` handler's only effect is one write: `internalAdapter.updateUser(userId, { role: parseRoles(ctx.body.role) })`. Nothing else — no session revocation, no second write. The `roles` vocabulary check is skipped because ObjectStack passes no `roles` option. And that scalar is folded straight back into identity: `customSession` seeds `positions[]` from `storedRole.split(',')`, and `isPlatformAdminUser` accepts `role === 'admin'`. So typing `admin` into that text box is a supported, gated, one-user-at-a-time resurrection of the vetoed dual identity. Retirement is the right call.

**But retiring it requires editing `packages/spec/**`, and this lane has zero ownership there.** Measured, not reasoned: removing the action from `sys_user.object.ts` turns `feature-gate-guard.test.ts` red —

```text
FAIL src/feature-gate-guard.test.ts > feature-gate completeness guard (#2874)
  > forward: every registered gated input carries the matching predicate
  > sys_user.actions.set_user_role is gated on admin
FAIL src/feature-gate-guard.test.ts > ... > reverse: ... > finds the gated surface (guards the walker itself)
```

The forward failure is repairable **only** by removing `'sys_user.actions.set_user_role'` from `PUBLIC_AUTH_FEATURES.admin.gatedInputs` in `packages/spec/src/kernel/public-auth-features.ts`. The guard reads that registry from `@objectstack/spec/kernel`; there is no in-package repair. The measurement was taken on a throwaway edit and reverted byte-for-byte (`git hash-object` back to `79adba85e…`, matching `origin/main`).

Per the dispatch's explicit red line — and because #10159 and #10025 are already parked on the `packages/spec` ownership question — that half **stops here and returns to the maintainer**. The exact remaining work is one array-line deletion in `public-auth-features.ts`, the action block in `sys-user.object.ts`, a regeneration of the four `*.objects.generated.ts` translation files, the `adminActions` list and feature-gate matrix row in `platform-objects.test.ts`, the `>= 38` walker floor in `feature-gate-guard.test.ts`, and the stale `sys_user.role` field description ("Set via the Set Platform Role action.").

**Consumer sweep for `set_user_role`** (repo-wide `git grep`, plus `objectui`), so the reviewer is not asked to trust an unqualified zero:

- **Declaration:** `packages/platform-objects/src/identity/sys-user.object.ts`
- **Generated translations:** `en` / `es-ES` / `ja-JP` / `zh-CN` `.objects.generated.ts`
- **Tests:** `platform-objects.test.ts` (×2), `feature-gate-guard.test.ts` (via the registry)
- **Spec registry:** `packages/spec/src/kernel/public-auth-features.ts` ← the blocker
- **Comments / docs only:** `sys-user.page.ts`, `packages/cli/src/commands/serve.ts`, `content/docs/references/system/auth-config.mdx`, `docs/adr/0092-*.md`, `docs/qa/platform-checklist/areas/identity-auth.json`
- **Route ledger:** `auth-route-ledger.ts` lists `POST /api/v1/auth/admin/set-role` — the vendor **route** stays mounted regardless; retiring the action does not touch it
- **objectui: zero.** Counter-checked — `objectui` *does* name sibling `sys_user` actions by name (`create_user`, `set_user_password` in `useConsoleActionRuntime.test.tsx`, `ActionParamDialog.tsx`, `useObjectLabel-actionResultDialog.test.tsx`), so the search machinery demonstrably works and the zero is real.

**No live consumer would break** — the console renders whatever the object declares, and no code anywhere calls the action by name.

## Verification

**Two-directional pins**, `status` AND `code` on every refusal (ADR-0112), in `admin-impersonate-endpoint.test.ts`. Platform admin granted the ADR-0068 way (`sys_user_permission_set` → `admin_full_access`), with the legacy scalar asserted *not* to be `'admin'` so the suite cannot pass for the wrong reason: admitted, impersonation takes effect at the seam the data routes use (`auth.api.getSession`), `impersonated_by` recorded. Other direction: plain member `403 YOU_ARE_NOT_ALLOWED_TO_IMPERSONATE_USERS` with no session written, anonymous `401`, and an **org owner** — not a platform admin under ADR-0068 — still `403`.

**#8243 hook pin** (the guard against the forbidden shape creeping back): after the platform admin impersonates, the caller's bearer resolves to **nobody**, the `set-admin-session-token` recovery credential comes back and is CORS-exposed, and the original admin session row is gone with a rotated one in its place.

**Ablation.** Predicted signature recorded first, then measured: neuter `applyPlatformAdminImpersonation` to a no-op (reproducing today's `main`) → predicted 8 red / 6 green, **measured 9 red / 15 green**. One honest correction: `boots without a better-auth endpoint-conflict error` was predicted green but is red, because it carries a `200` precondition — my prediction missed that, and the deviation is reported rather than smoothed. The reds are the right ones, including the one that proves `code`-plus-`status` is load-bearing: **`a platform-admin TARGET cannot be impersonated` keeps status `403` and fails only on the code** (`YOU_ARE_NOT_ALLOWED_TO_IMPERSONATE_USERS` instead of `YOU_CANNOT_IMPERSONATE_ADMINS`) — a status-only assertion would have stayed green. The refusal pins (member 403, anonymous 401, org-owner 403) stay green under ablation, which is exactly why a one-directional pin would have been worthless here.

**Rebuild statement**, argued from the files rather than asserted: no rebuild is required for this ablation. `admin-impersonate-endpoint.test.ts` reaches its subject by **relative** import (`./auth-manager`, `./admin-impersonate-endpoint`, `./impersonation-bearer-rotation`), so vitest transforms the package's TypeScript source and `dist/` is never on the resolution path; the dist-preflight discipline governs subjects reached through a dependency's `exports` map, which this is not. The claim is falsifiable by the run itself — an ablation that failed to reach the running code would have left all tests green, the exact false-green signature. Restored byte-for-byte: `git hash-object` = `dcb647618f87c48e7eddb0b5ac7c2d3fc4948edc` both before mutation and after restore, equal to the committed blob, ablation marker absent.

**Commands**, exit codes captured before any pipe, each quoted from the gate's own verdict line. Gate union re-derived after the final commit with `node scripts/pm/dispatch-gates.mjs` (no paths passed) on a clean worktree at `16f43973b`:

```text
pnpm --filter @objectstack/plugin-auth test        Test Files 61 passed (61) · Tests 1359 passed (1359)
pnpm --filter @objectstack/plugin-auth typecheck   tsc --noEmit — exit 0

check:changeset-gate-self-tests   EXIT=0     check:objectui-changeset        EXIT=0
check:slot-lookup                 EXIT=0     check:test-source-alias         EXIT=0
check:type-source-resolution      EXIT=0     check-adr-0087-registration     EXIT=0
check-changeset-no-major          EXIT=0     check-empty-changeset           EXIT=0
check:query-options-erasure       EXIT=0     check:cross-package-test-inputs EXIT=0
docs-audit/check-affected-docs    EXIT=0

check-engine-double-contract: OK — 331 pinned, 133 in the DEBT ledger, 2 exempt.
check-nul-bytes: OK (scanned 6093 text file(s) … no raw ASCII control bytes).
where-matcher conformance holds: 265 matcher(s) discovered … 0 silently-wrong … none new.
check-type-check-coverage: OK — 64/77 workspace packages type-checked …
check-type-check-coverage --re-measure: OK — 33 ledger entr(ies) re-measured in 237.2s,
  1924 raw tsc error(s) total, none above its recorded number.
  surplus: none — every entry sits exactly at its measurement, so any new error is red.
```

`@objectstack/plugin-auth` holds at **109** — separately confirmed by a test-inclusive `tsc --noEmit` over the package: exactly 109, **zero** of them in the files this PR adds. The `check-type-check-coverage.mjs` ledger is untouched, and the `:2766` self-test fixture was not touched.

## Notes for the reviewer

- `USER_NOT_FOUND` is restated as a local constant rather than imported from the `better-auth` **root** entry. That is not a style choice: several suites in this package `vi.mock('better-auth', …)` to capture the `betterAuth()` config, and vitest throws on a missing export from a mocked module — measured, the whole `admin` plugin was swallowed by `addOptionalPlugin`'s catch and silently disabled. A restated constant is only safe while something proves it still matches, so a pin asserts it equals the vendor's own `BASE_ERROR_CODES.USER_NOT_FOUND`.
- If a future vendor bump renames or drops the endpoint, `applyPlatformAdminImpersonation` returns `false` and `auth-manager.ts` logs loudly. The route then falls back to the vendor handler, which refuses every platform admin — a broken button, never an open door.
- The new suite imports the engine double from a sibling `.test.ts`, which re-registers that file's 10 tests here. The cost was measured and the three alternatives are worse; the reasoning is written at the import site so nobody "fixes" it into a new engine double, a gate-invisible helper, or a suite-free fixture file this package's bare `vitest run` cannot load.

Out-of-scope findings filed unassigned: **#10348** (three spellings of the ADR-0068 platform-admin id read, one skipping the system read context) and **#10349** (bodyless `401` from every better-auth-native `/admin/` route).

`#9652` and `#9969` are **not addressed here** — they depend on this card, and every other better-auth-native `/admin/*` route still gates on the legacy scalar, deliberately untouched.

https://claude.ai/code/session_01PnJHU45vPJj5UQrxe946Bx

_Generated by [Claude Code](https://claude.ai/code/session_01PnJHU45vPJj5UQrxe946Bx)_


os-elon:
## Maintainer ruling received — the escalation above is resolved as option (A)

Ruled live in the PM session, 2026-08-21 (verbatim, untranslated): 「**A：新增 vendor 状态**」 — surface 3 gains a fourth state for **vendor-endpoint wire format**, the counterpart of surface 2's already-adjudicated kind 3. Execution clauses attached to the ruling: the state **stays counted** (closed at exactly N bodies, like this surface's `exempt`); a **`note` is mandatory**, naming the vendor, the vendor client reader, and the contract-partner endpoint(s); adding or widening an entry is **⛔ MAINTAINER-ONLY** (same marker discipline as exempt-widening); and the const-hoisting evasion documented in the escalation is to be **named as a forbidden move** in the gate's own prose. Option (B) — widening #9389's `exempt` grammar — was considered and rejected: it would turn a single ruled boundary into a growable bag of classes.

Implementation is dispatched as **#10554** (branch `claude/route-envelope-decision-y3itw6`, fable tier). Once the machinery lands on `main`, this PR's remaining step is: merge `main`, then declare `admin-impersonate-endpoint.ts` under the new state with the required note (vendor: better-auth; reader: `authClient.admin.impersonateUser`; contract partner: `/admin/stop-impersonating`) — the declaration itself will read ⛔ MAINTAINER-ONLY on first entry, and this ruling is the authorization for exactly that one entry. No functional code change to this PR is required or authorized by the ruling.

The dev's refusal to take the const-hoisting route, and the byte-clean stop-and-report, were the correct calls — that discipline is what produced a clean fork for the maintainer instead of a silently green gate.

---

Issue #11800, opened 2026-08-24
`guard-main-checkout.sh` (the Edit/Write half of the worktree-first pair) ships no self-test matrix — its Bash sibling has 121 cases


os-steve:
Found while wiring #11514 (PR #11799), which runs every `.claude/hooks/*.selftest.sh` in CI. Filing rather than fixing: #11514 wires execution only and authoring a matrix is a `.claude/**` change.

## Measured

`.claude/hooks/` holds **three hooks and two self-tests**:

```
.claude/hooks/guard-main-checkout-bash.selftest.sh   121 cases
.claude/hooks/guard-main-checkout-bash.sh
.claude/hooks/guard-main-checkout.sh                 <-- no matrix
.claude/hooks/guard-shared-stash.selftest.sh          32 cases
.claude/hooks/guard-shared-stash.sh
```

`guard-main-checkout.sh` is the `Edit`/`Write`/`NotebookEdit` half of the worktree-first enforcement — the half named in CLAUDE.md and AGENTS.md as the primary guard. It has no self-test at all. Its Bash sibling, which exists only because #5790 found that the Edit/Write hook could be bypassed through `sed -i` / `tee` / redirection, has 121.

## Why it matters now specifically

Once #11799 lands, CI runs every matrix in that directory, so a regression in `guard-shared-stash.sh` or `guard-main-checkout-bash.sh` goes red. `guard-main-checkout.sh` is the one hook for which the answer stays "nothing checks it" — and after the wiring, `.claude/hooks/` reads as a covered directory, which is the more expensive state than being visibly uncovered. The wiring is discovery-based, so a matrix added for it is picked up with no workflow edit.

The failure direction is the quiet one: this hook fails **open** by design for anything it cannot parse, so a regression that makes it fail open more often produces no error anywhere — edits into the shared primary checkout simply start being allowed, and the symptom is another agent's tree being clobbered, attributed to something else.

## Shape of the work

[...]

PR #11814, opened 2026-08-24
test(hooks): add the missing self-test matrix for guard-main-checkout.sh

os-steve:
Part of #11800

Adds `.claude/hooks/guard-main-checkout.selftest.sh` — the missing case matrix for the `Edit`/`Write`/`NotebookEdit` half of the worktree-first guard. **The hook is not touched.** `Part of` rather than `Fixes` because the card's coverage goal is met here, but the matrix uncovered two real defects in the hook that this PR deliberately does not fix (#11809, #11810) — closing the card is the reviewer's call once those are triaged.

## What it covers

Cases are derived from what this hook actually decides, not ported from `guard-main-checkout-bash.selftest.sh`. That sibling's 121 cases are mostly shell splitting; this hook parses no shell at all — it reads `.tool_input.file_path` and makes a path-and-worktree decision — so almost none of them transfer. 87 cases over eleven classes:

| class | what it pins |
| --- | --- |
| core verdict | shared primary checkout blocks; the same files in a linked worktree allow; files in no repo allow |
| nearest existing ancestor | a new file in a not-yet-created directory is judged by the nearest directory that exists (header claim 3) |
| the file's own repo decides | `CLAUDE_PROJECT_DIR` points at the wrong repo in every case, so a hook judging the session gets all six backwards (header claim 2) |
| tool-agnostic | the hook never consults `tool_name`; scoping is the matcher's job |
| escape hatch | `OS_ALLOW_MAIN_EDITS` exempts on exactly `1` — not `0`, empty, `true`, `yes`, `11`, or `" 1"` |
| no usable path | fails **closed** on the shared checkout — the opposite posture from the Bash sibling, and an explicit `else` in the hook rather than a fall-through |
| `CLAUDE_PROJECT_DIR` unset | the no-path branch falls back to the process cwd |
| relative path | judged by the hook's process cwd; the payload's own `cwd` field is ignored in both directions |
| submodule | `.git/modules/<name>` is not mistaken for `.git/worktrees/<name>` |
| jq-less fallback | the text scan reaches the same verdict on real payloads, including paths with spaces and an escaped decoy `file_path` inside a Write payload's content |
| wiring | `settings.json` routes `Edit`, `Write` and `NotebookEdit` to this hook — nothing else in the repo checks that, and the hook is tool-agnostic, so the matcher is the only thing scoping it |

Hermetic and discovery-ready, same shape as the sibling: its own `git init` fixtures, a linked worktree of each, a non-repo directory under `$TMPDIR`, `jq` + `git` only, no install, no build, no network, exit 0 = all cases hold. PR #11799 picks it up with no workflow edit.

**Every fixture default is a directory in no repo at all** — the input on which this hook allows everything. A case expecting `block` therefore cannot pass by accident: the verdict can only have come from the path in the payload.

## Two real defects found, filed, not fixed

Execution beat inspection here — both were invisible from reading the hook.

- **#11809** — `case "$gitdir" in */worktrees/*) exit 0` is a substring match on a path, not a test for a linked worktree. A **primary** checkout that merely lives under a directory named `worktrees` (e.g. `~/worktrees/objectstack`) reads as a worktree and is unguarded. `git rev-parse --git-dir` prints a relative `.git` at the repo toplevel and an absolute path from any subdirectory, so the same unguarded checkout gets **opposite verdicts by depth**. `guard-main-checkout-bash.sh` line 339 carries the identical idiom and fails open on the same fixture.
- **#11810** — the matcher routes `NotebookEdit` here, but `NotebookEdit` sends **`notebook_path`**, and the hook reads only `file_path`. Every notebook edit therefore takes the no-path branch and is judged by `CLAUDE_PROJECT_DIR` instead of by the file — a constant per session, wrong in **both** directions: notebook edits inside a correctly-created worktree are refused, and from a worktree-rooted session notebook edits **into the shared primary checkout are allowed**. This falsifies the hook's own header claim 2 for one of its three tools.

Both are pinned in the matrix under sections labelled `KNOWN HOLE`, with the wrong verdicts marked and the issue that must flip them named. That keeps CI green and makes each fix a mechanical edit here. They are explicitly not written as intended behaviour.

One further boundary is recorded without being filed: the jq-less fallback is a text scan, not a JSON parser, so a `\u`-escaped path yields no path. Not a defect — `jq` is present wherever the hook runs and Claude Code emits plain UTF-8 paths — but it is the first thing to fix if that fallback ever becomes load-bearing.

## Non-vacuity

A matrix that has never failed proves nothing. Every class was shown to fail against a **mutated scratch copy** of the hook before being trusted to pass against the real one. `GUARD_MAIN_CHECKOUT_HOOK` / `GUARD_MAIN_CHECKOUT_SETTINGS` point the matrix at those copies; both default to the real files, so a plain invocation checks the real hook and real wiring. The real files were never written to — asserted by sha at the end of every run, under `trap … EXIT INT TERM`.

Each mutation was proven on disk before any result was read — anchor count in **both** directions plus a sha change, never an editor's exit code:

| mutation | on disk | result |
| --- | --- | --- |
| M1 drop the `*/worktrees/*` arm | anchor 1→0, sha `217a8b1c2d9f`→`1fb7554b088a` | 25 failed, 10 sections red |
| M2 drop the nearest-ancestor walk | anchor 1→0, sha →`0245cef204f2` | 5 failed, incl. the new-file class |
| M3 judge `CLAUDE_PROJECT_DIR` instead of the file | anchor 1→0, replacement 0→1, sha →`549968e0e272` | 35 failed, incl. the file's-own-repo class |
| M4 drop the `OS_ALLOW_MAIN_EDITS` line | anchor 1→0, sha →`674df1fa18f5` | 2 failed — exactly the escape hatch |
| M5 make the no-path branch fail open | anchor 1→0, replacement 0→1, sha →`269aa24a5a73` | 9 failed, incl. the fails-closed class |
| M6 rename the key the text scan looks for | anchor 1→0, replacement 0→1, sha →`5d88e1882e8a` | 4 failed — exactly the jq-less class |
| M7 turn the refusal into an allow | anchor 1→0, replacement 0→1, sha →`267ea5a76be4` | 42 failed, every block case |
| M8 drop `NotebookEdit` from the matcher | anchor 1→0, replacement 0→1, sha `1fd2d8868b8c`→`b8ff8b9fa710` | 1 failed — exactly the wiring case |

M8 mutates a scratch copy of `settings.json` rather than the hook: the wiring class is the one assertion no hook mutation can redden, so it needed its own control. No rebuild leg applies — the hook is an interpreted shell script executed directly from the path under test, so there is no compiled artifact between the mutation and the measurement; the sha of the file actually invoked is the equivalent proof.

## Verification

Union re-run after the final commit, at `f0b4892ca`:

```
.claude/hooks/guard-main-checkout.selftest.sh          87 passed, 0 failed   (exit 0)
node scripts/check-agent-test-spelling.mjs  --self-test / run                (exit 0 / 0)
node scripts/check-doc-authoring.mjs        --self-test / run                (exit 0 / 0)
node scripts/check-skill-frame-sync.mjs     --self-test / run                (exit 0 / 0)
node scripts/check-nul-bytes.mjs            --self-test / run                (exit 0 / 0)
node scripts/pm/check-governed-merges.mjs   --self-test                      (exit 0)
```

`check-nul-bytes` confirms it saw this file: `scanned 6590 text file(s) — 6589 tracked, 1 untracked-not-ignored`. Exit codes were captured by redirect-then-capture, never through a pipe.

Gate set derived by `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack` against the real change set (1 path). **One declared narrowing**: `pnpm --filter @objectstack/lint run check:doc-formula-expressions` was not run locally — it needs a full workspace install this worktree does not have, and the diff adds no doc or formula content. CI runs it regardless.

No changeset: `.claude/**` only, nothing published — `skip-changeset`.

⛔ Governed surface (`.claude/**`): this PR **stays draft**. Not to be flipped ready, armed for auto-merge, or enqueued.

_Generated by [Claude Code](https://claude.ai/code/session_015ahemw8RcTgqtxrj15PEZx)_

---

### DIFF
--- a/.claude/hooks/guard-main-checkout.selftest.sh
+++ b/.claude/hooks/guard-main-checkout.selftest.sh
@@ -1,0 +1,320 @@
+#!/usr/bin/env bash
+# Self-test for guard-main-checkout.sh — run it after touching that hook:
+#
+#   .claude/hooks/guard-main-checkout.selftest.sh
+#
+# Feeds the hook the same JSON payload shape Claude Code delivers on PreToolUse and asserts
+# the block/allow verdict per case. Hermetic: it builds its OWN throwaway git repos, a linked
+# worktree of each and a non-repo directory under $TMPDIR, so the matrix never depends on
+# which machine or which checkout it runs from. Needs jq and git and nothing else — no
+# install, no build, no network. Exit 0 = all cases hold.
+#
+# Companion to guard-main-checkout-bash.selftest.sh, which covers the Bash half of the same
+# worktree-first pair. That matrix is mostly SHELL SPLITTING; this hook parses no shell at
+# all — it reads .tool_input.file_path and makes a PATH-AND-WORKTREE decision — so these
+# cases are derived from what this hook actually decides, not ported from the sibling.
+#
+# Fail-open by default, on purpose: the process cwd AND CLAUDE_PROJECT_DIR both default to a
+# directory in no repo at all, which is the input on which this hook allows everything. A
+# case that expects `block` therefore cannot pass by accident — the verdict can only have
+# come from the path in the payload. Sections that need a different default say so.
+#
+# GUARD_MAIN_CHECKOUT_HOOK points the matrix at a scratch copy of the hook, and
+# GUARD_MAIN_CHECKOUT_SETTINGS at a scratch copy of settings.json; that is how the mutation
+# runs that prove these cases can fail are driven (see NON-VACUITY at the foot of this file).
+# Both default to the real files, so a plain invocation checks the real hook and real wiring.
+
+set -uo pipefail
+
+here="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+hook="${GUARD_MAIN_CHECKOUT_HOOK:-$here/guard-main-checkout.sh}"
+settings="${GUARD_MAIN_CHECKOUT_SETTINGS:-$here/../settings.json}"
+pass=0
+fail=0
+skip=0
+
+command -v jq >/dev/null 2>&1 || { echo "selftest needs jq to build payloads" >&2; exit 1; }
+command -v git >/dev/null 2>&1 || { echo "selftest needs git to build the fixture" >&2; exit 1; }
+[ -x "$hook" ] || { echo "hook not executable: $hook" >&2; exit 1; }
+
+# --- fixture ---------------------------------------------------------------------------
+# MAIN   a shared PRIMARY checkout            WT     a linked worktree of MAIN
+# SIB    a SECOND primary checkout            SIBWT  a linked worktree of SIB
+# PLAIN  a directory inside no repo at all    ODD    a PRIMARY checkout whose own path
+#                                                    carries a literal `worktrees` segment
+tmp="$(mktemp -d)"
+nojq=""
+trap 'rm -rf "$tmp" ${nojq:+"$nojq"}' EXIT INT TERM
+MAIN="$tmp/mainrepo"
+WT="$tmp/mainrepo-task"
+SIB="$tmp/siblingrepo"
+SIBWT="$tmp/siblingrepo-task"
+PLAIN="$tmp/plain"
+ODD="$tmp/worktrees/oddrepo"
+mkdir -p "$MAIN/pkg/deep" "$SIB/pkg" "$PLAIN" "$ODD/pkg"
+(
+  for r in "$MAIN" "$SIB" "$ODD"; do
+    cd "$r" || exit 1
+    git init -q .
+    git config user.email selftest@example.com
+    git config user.name selftest
+    : > README.md
+    mkdir -p pkg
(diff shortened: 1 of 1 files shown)


os-trump:
Merging on the maintainer's direct instruction. Provenance (who / verbatim / where): the maintainer, in the PM chat session (2026-08-25), after the batch-1 governed-PR review including this PR was presented: 「同意，帮我合并，然后继续」. Audit record for the governed-merges checklist. Checks verified at head `f0b4892ca`: 29 check runs — all success/skipped; the two `cancelled` entries are concurrency-superseded duplicates whose same-name checks succeeded in the later runs. The two KNOWN HOLE cards this matrix uncovered (#11809, #11810) stay open for triage. Executed by PM session `session_01H9StxQgG2DPA26XzZZqnJB`.

---

PR #11934, opened 2026-08-25
Base the worktree recipe on a fetched origin/main, and make dispatch-gates announce a stale tree

### LINKED ISSUE #11540: [finding] The worktree recipe in CLAUDE.md / AGENTS.md branches off local `main`, a ref nothing in the recipe fetches — measured 120 commits and 179 hours stale on a live shared checkout
Filed unassigned by the `domain:devx` PM seat (session `session_015ahemw8RcTgqtxrj15PEZx`). Recording only — ⛔ no `domain:*` label applied; routing and grading are triage's.

Measured while deriving model tiers for a dispatch batch, on the shared checkout at `/home/user/objectstack`.

## The recipe names a ref, and nothing refreshes it

Three places prescribe the same incantation, base ref included:

| where | text |
|---|---|
| `CLAUDE.md:33` (objectstack) | `git worktree add ../<repo>-<task> -b <branch> main && cd ../<repo>-<task> && pnpm install` |
| `CLAUDE.md:16` (objectui) | identical |
| `AGENTS.md:173` — **Prime Directive #11 itself** | `git worktree add ../objectstack-<task> -b <branch> main && cd ../objectstack-<task> && pnpm install` |

`main` here is the **local** branch ref, not `origin/main`. Nothing in the recipe, and nothing in the surrounding prose, fetches. So the base a dev inherits is only as fresh as whenever someone last happened to update that ref on the shared checkout — and on a long-lived container nobody has a reason to.

## Measured on the live shared checkout

```
local main   : 76876ac3   2026-08-16T14:39:32+00:00
origin/main  : cd932772   2026-08-24T02:34:46+00:00
behind       : 120 commits
age gap      : 179 hours  (7.5 days)
```

`git merge-base main origin/main` is `main` itself, so this is not a divergence to reconcile — a branch created by the documented recipe today starts **120 commits back, by construction**. No judgement call, no race, no unlucky timing.

## ⚠️ Honest scope: this is latent, not a live incident

I checked the one dev worktree in flight at the time (`objectstack-issue-10611`). Its merge-base with `origin/main` is `daacc107` — **2 commits back**, i.e. that dev fetched and branched off `origin/main` rather than following the recipe literally. So:

- **live instances of a dev bitten by this: zero, as measured.**
- The agents currently working this repo are, in practice, doing the right thing *despite* the recipe rather than because of it.

That makes this the same shape as #11494 — a counting/authoring contract that is violated the moment something lands on it, with no live instance today. It is filed on that basis and should be graded on that basis, not as an incident.

## ⭐ The related half that DID cause a live wrong reading, and is a different mechanism

The stale **ref** above is latent. The stale **working tree** of the same checkout is not — it produced a wrong instrument reading this shift:

The PM seat derives model tiers by running `node scripts/pm/dispatch-gates.mjs --tier <paths>` from `/home/user/objectstack`. That checkout sits at `b863f012`, and its on-disk copy of that script hashes `00a51862…`, while `origin/main`'s copy hashes `edfe4ab1…`. The two differ because PR #11512 changed that exact file. **The stale copy runs and exits 0 and prints a well-formed tier verdict** — it just answers about a tree nobody is on.

This was caught only because #11512 had landed minutes earlier in that specific file, so the check was top of mind. Nothing in the tool, the output, or the workflow would have said so otherwise: dispatch-gates prints the commit it derived from (`derived from the tree of … at commit <sha>`), which is exactly the line that would have exposed it — but that line names the checkout's own HEAD, so a stale checkout reports its staleness in a form that reads as ordinary provenance.

Both halves share one root: **the shared checkout is never refreshed, and everything that reads it — the documented base ref, and every tool invoked from it — silently inherits whatever staleness has accumulated.**

Workaround in use by this seat, offered as a shape rather than a prescription: a detached worktree pinned at `origin/main` for tool runs (`git worktree add --detach ../objectstack-pm-main origin/main`), so derivations are provably against the tree being dispatched for.

## Direction (not a decision)

1. Change the base in all three texts to `origin/main`, and prepend an explicit `git fetch origin main`. Cheapest, and it makes the recipe self-sufficient rather than dependent on checkout hygiene nobody owns.
2. Leave the recipe and make the staleness *visible* instead — e.g. have the tooling that reads the shared checkout assert its HEAD against `origin/main` and say so loudly. Attacks the second half, which is the one with a measured wrong reading.
3. Both. They are independent: (1) fixes what a dev branches off, (2) fixes what a tool reports about.

⚠️ **All three target files are governed surface** (`AGENTS.md`, `CLAUDE.md` in both repos). Whatever route is chosen, the PR stays draft and is merged by the maintainer by hand — no AI seat merges, queues, or arms auto-merge on it. Note also `AGENTS.md:203` carries the recipe a second time as `git worktree add ../objectstack-<task> -b <branch>` with **no base ref at all**, which silently bases off whatever the shared checkout's HEAD happens to be — a related but distinct hazard, and arguably the worse of the two since it has no stated base to be wrong about.

## Searched before filing

`worktree main stale` (0 hits) · `origin/main base branch recipe` (0) · `dispatch-gates stale checkout` (0). ⚠️ The zero-hit readings are reported as live because a control query in the same window (`platform-checklist`) returned known-present results — #11421 and #11524 — so the search surface was responding.

Refs: #11512 (the PR whose landing exposed the stale tool copy) · #11190 · #11494 (same latent-contract shape)

os-steve:
Fixes #11540

Triage direction 3 — both halves. (1) fixes what a dev branches off; (2) fixes what a tool reports about.

## Half (a) — the recipe now fetches, and names a base it fetched

The documented recipe branched off the **local** `main` ref, which nothing in the recipe or its surrounding prose fetches. A worktree created by following it literally starts as far back as whoever last happened to update that ref — and on a long-lived shared checkout, nobody has a reason to.

Re-measured on the live shared checkout at claim time (the card's own numbers are a snapshot; these are mine):

| reading | value |
|---|---|
| local `main` | `76876ac3` · 2026-08-16T14:39:32Z |
| `origin/main` | `c3c72a4b` · 2026-08-25T00:48:19Z |
| local `main` behind `origin/main` | **1158 commits** |
| age gap | **202.1 hours** (8.4 days) |

Positive control in the same run: `git rev-list --count origin/main..main` = 0, so this is a strict lag, not a divergence.

Sites enumerated independently rather than inherited, and judged one by one:

| site | verdict | why |
|---|---|---|
| `AGENTS.md:173` (Prime Directive #11) | **in** | bare `main` base |
| `AGENTS.md:203` | **in** | **no base ref at all** — bases off whatever the shared checkout's HEAD happens to be, which at claim time was another agent's PM branch, not even `main` |
| `CLAUDE.md:33` | **in** | bare `main` base |
| `AGENTS.md:227` | **out** | the `-cmp` recipe takes an explicit ref argument from the caller — correct as written |
| `CLAUDE.md:63` | **out** | same |

`origin/main` alone would have the same defect one layer down, so each fixed site now begins with an explicit `git fetch origin main &&`. Measured at claim time: `origin/main` happened to be current at that instant (a fetch moved it 0 commits) because another agent had fetched recently — luck, not a guarantee, which is the point.

### Line-budget fence

Both files sit at their shrink-only ceilings with **zero headroom**, so this had to be paid in place:

```
✓ check-skill-line-ratchet: AGENTS.md is 961 lines (ceiling 961; headroom 0).
✓ check-skill-line-ratchet: CLAUDE.md is 86 lines (ceiling 86; headroom 0).
```

Net **±0 lines on both**. The `CEILINGS` map is untouched. `AGENTS.md:173` and `CLAUDE.md:33` are single lines with `&&` chains, so the fetch spliced in place. The `AGENTS.md:203` paragraph needed a reflow to absorb the added text; the only words dropped are the fillers "and" and "this way" — every substantive clause, including "not a preference" and the incident parenthetical, survives intact.

## Half (b) — `dispatch-gates` now says when it is deriving from a stale tree

This is the half with a measured wrong reading. The banner already named the commit an answer came from — the very fact that exposes a stale checkout — but printed it in the same spelling a current checkout uses, so staleness arrived dressed as ordinary provenance.

Reverse verification, old code against the live stale shared checkout (241 commits behind at the time):

```
dispatch-gates: gate list derived from the tree of 'objectstack-ai/objectstack' at commit b863f012f (/home/user/objectstack).
  Families are a property of THAT repo. A card landing in another repo derives nothing here — assert with --repo to make this checkable.
```

Not one word about staleness. New code, same checkout:

```
  ⚠️  STALE TREE — this answer is derived from a tree at least 243 commit(s) behind origin/main, and 104 file(s) it derives from CHANGED across that range.
    HEAD 2026-08-23T13:51:48+00:00 vs origin/main c3c72a4bc 2026-08-25T00:48:19+00:00
    Stale here: .github/workflows/ci.yml .github/workflows/lint.yml … +98 more
    Those files ARE the families printed below, so this run read their old copies and still exited 0 — a well-formed answer about a tree nobody is on.
    "At least": origin/main is a LOCAL ref only a fetch moves. Run 'git fetch origin main' and derive again from a tree at origin/main.
```

`scripts/pm/dispatch-gates.mjs` — the exact file whose stale copy produced the card's wrong reading — is named in that list.

Three design choices worth reviewing:

- **Commit distance alone is the wrong instrument.** A dev worktree falls a few commits behind within the hour by construction, so a warning keyed on distance fires on nearly every honest run and stops being read — reproducing the silence it was added to break. So the loud path is keyed on whether anything the answer is *derived from* (`.github/workflows`, `package.json`, `scripts/`) changed. Behind-but-surface-untouched gets one quiet factual line.
- **Three-dot diff**, so a dev's own edits to a check script are never reported back to them as staleness. Verified with a purpose-built control: a tree 1 commit behind that has itself committed an edit to `scripts/a.mjs` reports `changed=[]`.
- **The count is a lower bound and says so.** `origin/main` is a local ref only a fetch moves, so an unfetched checkout is measured against a base that is itself behind; unfetched staleness can only make the true number larger. That is what keeps this honest without the derivation reaching for the network.
- **Zero drift prints nothing.** No clearance line — against a base ref nobody refreshed, a clean bill of health is exactly the reading the failure would have passed. This matches the banner's existing deliberate refusal to print an "all paths present" twin.

Dogfooded: the guard fired on *this* branch mid-work (2 commits behind, 7 derivation files changed upstream), which is why `origin/main` is merged in — the first derivation would have been short.

## Gates

Derived live with `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack` (no hand-supplied paths), all 15 families run at **`6077e358f`**, all green:

`check:agent-test-spelling` · `check:cross-package-test-inputs` · `check:docs-audit-scope` · `check:entry-guard` · `check:parse-guard` · `check:pm-dispatch-gates` · `check:pm-governed-merges` · `check:pm-governed-prose` · `check:pm-skill-id-lint` · `check:pm-skill-ratchet` · `check:pnpm-filter-targets` · `check:required-contexts` · `check-ci-filter-parity.mjs` · `check-cross-package-test-inputs.mjs` · `check-required-contexts.mjs`

```
✓ dispatch-gates self-test: 594 cases pass.
```

594 includes 15 new cases for the drift guard, among them a positive control (a clone level with its base measures zero) and a real-git-repo pair separating an upstream commit inside the derivation surface from one outside it.

No changeset: docs prose plus an internal PM tooling script, nothing user-visible ships. `skip-changeset` applied.

## Governance

⛔ Governed surface (`AGENTS.md`, `CLAUDE.md`). This PR stays **draft** — not flipped ready, not armed, not queued. Review requested from `os-zhuang`; the hand-merge is the review record.

The objectui copy of the same recipe (`CLAUDE.md:16` there) is deliberately **not** touched here — the accepting seat files that follow-on card.

_Generated by [Claude Code](https://claude.ai/code/session_015ahemw8RcTgqtxrj15PEZx)_

---


os-trump:
Merging on the maintainer's direct instruction. Provenance (who / verbatim / where): the maintainer, in the PM chat session (2026-08-25), after the batch-1 governed-PR review including this PR was presented: 「同意，帮我合并，然后继续」. Audit record for the governed-merges checklist. Checks verified at head `6077e358f`: 34 check runs, all success/skipped, zero failures. Executed by PM session `session_01H9StxQgG2DPA26XzZZqnJB`.

---

PR #12026, opened 2026-08-25
fix(automation): resolve same-named flow definitions deterministically per ADR-0005

### LINKED ISSUE #11997: Packaged flow silently replaced by a same-named runtime flow: listItems returns both, the engine keys flows by bare name, and Map order decides the winner
## The defect

A runtime-authored flow that reuses a **packaged flow's name** silently and
**non-deterministically replaces** the packaged flow in the automation engine, with no
diagnostic, no receipt, and no way for an admin to tell which one is armed.

This is reachable today: the metadata registry declares `flow` with
`allowRuntimeCreate: true` (`packages/spec/src/kernel/metadata-plugin.zod.ts:827`), so
authoring a runtime flow named `opportunity_approval` while a package ships one is an
ordinary, accepted write.

## Mechanism — three facts that are individually correct and collide

1. **The SchemaRegistry stores both, deliberately.** Items are keyed
   `packageId:name` (`packages/objectql/src/registry.ts:1148`), and ADR-0048 section 3.4
   coexistence is pinned in `packages/objectql/src/registry-cross-package-collision.test.ts:20-40`
   — two packages shipping one bare name legitimately coexist, reachable via
   package-scoped `getItem`.

2. **`listItems` returns both, with no dedup and no precedence.**
   `packages/objectql/src/registry.ts:3394-3414`:

   ```
   const items = Array.from(this.metadata.get(type)?.values() || []) as T[];
   ```

   It filters by `packageId` only when one is passed, and filters out disabled packages.
   Nothing collapses two entries that share a bare name.

3. **The automation engine keys flows by bare name.**
   `packages/services/service-automation/src/engine.ts:1252`:

   ```
   private flows = new Map<string, FlowParsed>();
   ```

   and the boot pull (`packages/services/service-automation/src/plugin.ts:870-880`) is:

   ```
   const flows = ql?.registry?.listItems?.('flow') ?? [];
   for (const f of flows) { ... this.engine.registerFlow(def.name, def as never); ... }
   ```

   `registerFlow` ends in `this.flows.set(name, parsed)` (`engine.ts:2527`).

⇒ Both entries are pulled, both are registered under the same key, and **the last one in
Map iteration order wins**. The survivor is decided by registration order into the
registry, which is boot load order — not by any declared precedence.

## Why it is worth a card rather than a note

- **It is silent in both directions.** Neither `registerFlow` nor the flow pull warns on a
  name that is already present; the pull's only log is a count
  (`plugin.ts:891` `Pulled N flow(s) from ObjectQL registry`), and N counts the
  registrations, not the distinct names.
- **`listFlows` / `getFlowRuntimeStates` cannot show it** — they enumerate
  `this.flows.keys()` (`engine.ts:2600, 2638`), which by then holds one entry.
- **The trigger binding follows the survivor**, so the observable symptom is a packaged
  flow that "stopped working" or a runtime flow that "does something it was not authored
  to do", depending on which won.
- The precedent for the neighbouring type went the other way deliberately: for permission
  sets, a save targeting a package-declared name is **refused loudly** with a message
  naming the sanctioned path (`packages/plugins/plugin-security/src/packaged-permission-set-lock.ts`,
  maintainer ruling 2026-08-24, "lock the base, clone to customize"). Flows have no
  equivalent guard.

## Not measured

I did **not** run this end to end — this is a static read of the three code paths above,
found while measuring the trigger-dispatch path for the packaged-flow customization design
work. The registration-order claim in particular deserves a runtime confirmation before
anyone prices a fix.

## Possible directions (not a recommendation — this needs triage first)

- refuse the runtime write when a package declares the name (the permission-set shape); or
- make the engine's flow map package-aware, which is a much larger change; or
- at minimum, **warn** at pull time when one bare name is registered twice, so the
  condition stops being invisible.

## Provenance

Found while measuring the trigger-dispatch path for the packaged-flow clone-to-customize
design (#11665, section 2.2 of the design document posted there). Filed separately because
it is a live defect on `main` independent of whether that design is accepted, and because
it is not in that card's scope.

Read at `origin/main` `d63b01436035f0866df8c22156e26ab2a064580d`. Unassigned, not claimed.

---

os-sam:
Fixes #11997

⭐ **Clause-② declared: YES.** This changes collision behaviour on a served surface — a flow. Today the armed flow is decided by Map iteration order, so making the winner deterministic **changes which flow fires on some boots**. `needs:contract-review` is carried on this PR as the second half of the dual carrier (the card holds the first). ⛔ Not to be cleared, flipped ready, or auto-merged from here.

## The defect, confirmed at runtime before anything was changed

The card was an explicit static read (*"I did not run this end to end … the registration-order claim in particular deserves a runtime confirmation"*). It does hold. Measured with a real `SchemaRegistry` and a real `AutomationEngine` on unmodified `origin/main`:

| registration order into the registry | flow the engine armed |
|---|---|
| packaged first, then the runtime row | **RUNTIME** |
| the runtime row first, then packaged | **PACKAGED** |

Same two definitions, opposite outcomes, decided by nothing but Map iteration order. `listFlows()` returned exactly `['opportunity_approval']` in **both** cases and `getFlowRuntimeStates()` had one entry — so nothing observable distinguished the two boots.

One extra finding the card did not have: `Registry.registerItem`'s existing `[Registry] Collision` warning fires in only **one** of the two orders (`packageId && collection.has(bareKey)` — i.e. a runtime row exists and then a package ships the name). In that very order the warning promises *"The runtime row will shadow the package value"* while the engine armed the **PACKAGED** body. The one diagnostic that did fire was actively contradicted by the thing it warned about.

## The direction is ADR-0005's, not a choice made here

Triage's premise-first condition routed through ADR-0048, which does not restate the direction. ADR-0005 does, twice:

- §Decision, resolution order: `1. sys_metadata … ← overlay (wins)` above `2. SchemaRegistry / MetadataService … ← artifact default`
- §Collision warning: *"the runtime overlay layer silently shadows the artifact value **(correct ADR-0005 behavior)**"*

and ADR-0048 §1.5 lists the runtime/DB overlay as **"the sanctioned override path"** under what is *not* a collision. So the runtime overlay wins; the packaged artifact is the default it overlays. This PR makes the engine agree with that deterministically instead of agreeing with whatever Map order produced.

## What changed

1. **Deterministic precedence.** The boot pull collapses same-named definitions *before* anything is armed, via a new `resolveFlowPrecedence`. The order is total and never reads iteration order: `runtime` before `package`, then lexicographic `packageId`. That second rule covers ADR-0048 §3.4's legitimate two-packages-one-name case, which the engine's bare-name map cannot represent but still has to answer deterministically.
2. **The loud diagnostic** ADR-0048 already expected — one warning per colliding name, naming the name, every contender, and which one is armed, on a single line (#5048), with the facts also in `meta`. Repeated at `kernel:bootstrapped` beside the other automation audits.
3. **An admin-visible receipt.** `getShadowedFlows()` lists each contested name with its armed and shadowed definitions, and `getFlowRuntimeStates()` rows carry `armedFrom` / `shadowed` for contested names. Previously the displaced definition was invisible *by construction*. The `Pulled N flow(s)` line now counts distinct names, not registrations — the card noted N counted the latter.

⛔ **The engine's flow map is still keyed by bare name.** Making it package-aware is the larger change the card names and the ADR does not require; the receipt is a side map that never affects dispatch.

`isCodeArtifactBody` is exported from `@objectstack/objectql` rather than re-derived in the automation plugin. That test exists precisely so callers cannot drift into a second answer to "does a code package ship this name?", and a bare `_packageId` check gets it wrong twice — the `sys_metadata` rehydration sentinel, and a tenant overlay bound to a real package id (cloud#970). Both are pinned.

## Verification

All gates below ran against **`b5ca58cec2`**, the head of this branch, with a clean tree.

- `pnpm --filter @objectstack/service-automation test` — `Test Files  89 passed (89)` / `Tests  1051 passed (1051)`
- `pnpm --filter @objectstack/service-automation build` (this package has no `typecheck` script; its build is the type check) — `DTS ⚡️ Build success in 6329ms`
- `pnpm --filter @objectstack/objectql typecheck` — `tsc --noEmit && tsc --noEmit -p tsconfig.scripts.json`, exit 0
- `pnpm check:type-check-debt` — `--re-measure: OK — 32 ledger entr(ies) re-measured in 216.8s, 1898 raw tsc error(s) total, none above its recorded number.`
- `pnpm check:type-check-coverage` — `OK — 65/78 workspace packages type-checked (plus the root), 13 in the DEBT ledger`
- `pnpm check:engine-double-contract` — `OK — 405 pinned, 133 in the DEBT ledger, 2 exempt.`
- `pnpm check:where-matcher` — `✓ 297 matcher(s) discovered, 297 answer the combinator battery correctly or refuse it loudly`
- `pnpm check:query-options-erasure` — `✓ ratchet holds: 67 unswept non-test site(s) in 17 file(s), none new`
- `pnpm check:cross-package-test-inputs` — `OK: 16 package(s) read outside themselves, all declared`
- `pnpm check:test-source-alias` — `OK — 72 packages with tests scanned`
- `pnpm check:type-source-resolution` — `OK — 93 tsc program(s) across 77 packages scanned`
- `pnpm check:published-files` — `✓ 69 publishable package(s) of 78 workspace member(s)`
- `pnpm check:nul-bytes` — `✓ self-test: 75 assertions over a temp git repo`
- `node scripts/check-engine-split-ratio.mjs` — `ratio: 97.6%`, deferred per ADR-0076 OQ#5

Gates derived with `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack`, which reported the list as derived from this repo at `b5ca58cec2`. Two gates first came back **NOT MEASURED** and were re-run rather than counted: `service-automation typecheck` (`ERR_PNPM_RECURSIVE_RUN_NO_SCRIPT` — no such script; the build is the type check) and `check-engine-split-ratio` (refused on a shallow clone, re-run after `--shallow-since`).

**Declared narrowing:** the round was narrowed to the ratchet family plus `service-automation`, per dispatch. Repo-wide `pnpm lint` was replaced with a targeted run, and the narrowing is measured, not assumed: eslint's own `--format json` reports **7 files linted, 0 errors, 0 warnings**, and `eslint.config.mjs` states this repo *"runs one `eslint.config.mjs`, which never enables type-aware linting (no `parserOptions.project`, no typed `@typescript-eslint` rules) for ANY file"* — so this diff cannot move the verdict on any file it does not touch. CI runs the full farm regardless.

### Ablation

The failing assertions were predicted in writing first, then observed exactly. Ablation: strip the grouping from `resolveFlowPrecedence` so every item becomes its own winner — the pre-fix pull.

Predicted to fail (7): arms-same-definition-regardless-of-order · one-flow-per-bare-name · warns-loudly · admin-receipt · two-packages-by-packageId · first-seen-order · tenant-overlay-beats-package. Predicted to pass (4): registry-returns-both · arms-runtime-over-packaged (accidental — package-first leaves RUNTIME last) · single-definition-silent · sentinel-classification.

Observed: **`Tests  7 failed | 4 passed (11)`** — the same seven, test for test.

The mutation was proved on disk before the run (injected marker present, `git diff --stat` +7 lines), not inferred from an editor exit code. No rebuild was needed for this leg and that is load-bearing: the test reaches `flow-precedence.ts` through a *relative* import, which vitest resolves from source, while the unaliased `@objectstack/objectql` import resolves to `dist/` and was built and left unmutated. The script carried `trap '<restore>' EXIT INT TERM`. Restore was verified byte-for-byte — marker absent, disk/index/HEAD diffs all empty, and `git hash-object` equal to `git rev-parse HEAD:<path>` (`07ed8e6e…`).

_Generated by [Claude Code](https://claude.ai/code/session_01APWX2AwT3a4xDcjPCe8bk4)_

---
_Generated by [Claude Code](https://claude.ai/code/session_01APWX2AwT3a4xDcjPCe8bk4)_

### DIFF
--- a/.changeset/wild-pears-remain.md
+++ b/.changeset/wild-pears-remain.md
@@ -1,0 +1,34 @@
+---
+"@objectstack/service-automation": patch
+"@objectstack/objectql": patch
+---
+
+Arm a deterministic flow when a runtime-authored flow reuses a packaged flow's name
+
+A runtime-authored flow that reused a packaged flow's name silently replaced it,
+and which of the two ended up armed depended on registration order. The metadata
+registry keys items `packageId:name` and deliberately coexists both (ADR-0048
+§3.4), `listItems('flow')` returns both with no precedence, and the automation
+engine keys flows by bare name — so the boot pull registered both under one key
+and Map iteration order picked the survivor. Measured: registering the package
+first armed the runtime flow, registering the runtime row first armed the
+packaged flow, with no warning and no way to tell which had won.
+
+The boot pull now collapses same-named definitions before anything is armed,
+applying the ADR-0005 overlay precedence ADR-0048 §3.4 routes this case to: the
+runtime/DB overlay wins over the packaged artifact, which is the sanctioned
+override path. Two packages shipping one bare name resolve by package id, so
+boot order no longer decides anything.
+
+Collisions are no longer silent. The pull warns once per colliding name — naming
+the name, every contender, and which one is armed — and repeats it at bootstrap
+beside the other automation audits. `getShadowedFlows()` is a new receipt listing
+each contested name with its armed and shadowed definitions, and
+`getFlowRuntimeStates()` rows now carry `armedFrom`/`shadowed` for contested
+names; previously the displaced definition was invisible by construction, since
+the flow map holds one entry per name. The `Pulled N flow(s)` line now counts
+distinct names rather than registrations.
+
+`isCodeArtifactBody` is exported from `@objectstack/objectql` so consumers that
+collapse same-named metadata answer "does a code package ship this?" with the
+registry's own test instead of re-deriving it from `_packageId`.
--- a/packages/objectql/src/index.ts
+++ b/packages/objectql/src/index.ts
@@ -26,6 +26,15 @@ export {
   DEFAULT_EXTENDER_PRIORITY,
 } from './registry.js';
 export type { ObjectContributor, SchemaRegistryOptions } from './registry.js';
+// [#11997] The canonical "does a code package ship this body?" test (ADR-0029
+// D9.6). Exported because the ADR-0005 overlay precedence is not the registry's
+// alone to apply: any consumer that collapses two same-named contenders into one
+// slot — the automation engine's flow map is the first — has to answer the SAME
+// question, and its whole reason for existing is that callers must not drift
+// into a second answer. Consumers ask this; they do not re-derive it from
+// `_packageId`, which cannot tell a tenant overlay from a code artifact on its
+// own (see the function's own doc, and `isTenantAuthored` above it).
+export { isCodeArtifactBody } from './registry.js';
 // [#7865] Injected-column provenance — the machine-readable marker for anchors
 // the registry registers without provisioning storage (external objects,
 // ADR-0015). Canonical home: `@objectstack/metadata-core`, beside the
--- a/packages/objectql/src/registry.ts
+++ b/packages/objectql/src/registry.ts
@@ -1132,7 +1132,7 @@ function isTenantAuthored(item: unknown): boolean {
  * Truthy `_packageId`, not the `'sys_metadata'` rehydration sentinel, and not
  * tenant provenance.
  */
-function isCodeArtifactBody(item: unknown): boolean {
+export function isCodeArtifactBody(item: unknown): boolean {
   const it = item as { _packageId?: unknown } | null | undefined;
   if (!it || !it._packageId || it._packageId === 'sys_metadata') return false;
   return !isTenantAuthored(it);
--- a/packages/services/service-automation/src/engine.ts
+++ b/packages/services/service-automation/src/engine.ts
@@ -1240,6 +1240,34 @@ function graftConditionEnvelopes(converted: unknown, parsed: unknown): unknown {
     return converted;
 }
 
+/**
+ * [#11997] Where one flow body came from, as the boot pull could tell.
+ *
+ * `'package'` — a code-shipped artifact (`isCodeArtifactBody`, ADR-0029 D9.6).
+ * `'runtime'` — a `sys_metadata`/runtime-authored row with no real package
+ * provenance: the ADR-0005 overlay, and the sanctioned override path per
+ * ADR-0048 §1.5.
+ */
+export interface FlowContender {
+    source: 'package' | 'runtime';
+    /** The owning package id, when a code package ships this body. */
+    packageId?: string;
+}
+
+/**
+ * [#11997] What the ADR-0005 overlay precedence decided for one bare flow name.
+ *
+ * Emitted only when a name had more than one contender at pull time. `armed` is
+ * the body that is actually in the engine's flow map and will dispatch;
+ * `shadowed` are the ones that lost, in the order the registry listed them.
+ */
+export interface FlowShadowingRecord {
+    /** The bare name every contender claimed. */
+    name: string;
+    armed: FlowContender;
+    shadowed: FlowContender[];
+}
+
 export class AutomationEngine implements IAutomationService {
     /**
      * ADR-0044: maximum times a single node may be (re-)entered at the top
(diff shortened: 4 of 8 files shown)


os-trump:
Contract review **PASS** at `CONTRACT_REVIEW_TIER` (verdict + readings on the card: #11997). Label cleared on both carriers; flipping ready and entering the merge queue.

---

Issue #12683, opened 2026-08-27
docs build: `next build` holds ~7.6 GB in ONE process — the docs deploy's next memory ceiling, and no in-repo knob bounds it


os-zhuang:
Found while fixing #12677 (spec's DTS pass was declaring a 12 GB heap ceiling inside the build container). That fix lands separately and is not affected by this card; this records the **next** thing that will take the docs deploy down, because it is now the largest single consumer in the pipeline and nothing in this repo bounds it.

## Measured

`pnpm turbo run build --filter=@objectstack/docs` is two sequential tasks — `@objectstack/spec#build` then `@objectstack/docs#build` (the docs app's only workspace dependency is spec). Measured as peak anonymous RSS of the whole process tree, inside a cgroup capped at the stated size **and** a mount namespace whose `/proc/meminfo` reports the same figure, so tools that size themselves from reported memory see a real container:

| phase | peak | dominant process |
|---|---|---|
| `@objectstack/spec#build`, after the #12677 fix | 5654 MB | tsup DTS worker |
| `@objectstack/docs#build` | **~7.6 GB** | a single `next-build (v16.3.1)` process at **7592 MB** |

The whole pipeline in an 8192 MB container: **exit 137, `oom_kill=1`**, peak 8143 MB, killed in the docs phase. In a 10240 MB container: completes, 2/2 tasks, 5m02s, peak 8135 MB.

The turbopack worker-pool children are **not** the problem — they measure 100–430 MB each. One process holds everything.

## Why the existing controls do not bound it

Both knobs already in `apps/docs/next.config.mjs` and `apps/docs/package.json` were put there for earlier OOMs, and neither reaches this:

- **`NODE_OPTIONS='--max-old-space-size=4096'`** bounds V8's old space only. Next 16 builds with **Turbopack**, which is Rust — its allocations sit outside V8 entirely. Measured: the process reaches 7592 MB while nominally capped at 4096.
- **`experimental.cpus: 2`** bounds the static-generation worker fan-out. Measured above: the workers are small; capping them cannot move a 7.6 GB main process.

And the demand is **fixed, not adaptive** — it does not shrink when given less. Under smaller containers the docs build grows until it is killed rather than sizing down:

| container | result |
|---|---|
| 8192 MB | exit 137, `oom_kill=1`, `limit_failcnt=11073` |
| 6144 MB | exit 137, `oom_kill=1`, killed at 75s |
[...]

Issue #12333, opened 2026-08-25
docs site: production has not deployed since dpl_2nfW… — 19 merges to main, five epic PRs invisible


os-zhuang:
## Measurement

`https://objectstack.ai` is serving deployment `dpl_2nfWjGjSwZjakVEmUG1kBWD6697r`. That id is **unchanged across 19 merges to `main`**, four of which touched `apps/docs`, over roughly two hours.

```
dpl: dpl_2nfWjGjSwZjakVEmUG1kBWD6697r | age: 6510 | x-vercel-cache: HIT
```

The live generation can be dated precisely by what it does and does not contain:

| landed on main | live? |
|:--|:--|
| #12253 `/robots.txt` + `/sitemap.xml` (16:31) | ✅ serving |
| #12258 dotted-path 404 guard (16:52) | ✅ serving (`/foo.txt` → 404) |
| #12262 one `<h1>` per page (16:57) | ❌ `/docs/data-modeling/objects` still renders **2** `<h1>` |
| #12284 homepage title (16:57) | ❌ still `ObjectStack — Apps small enough for AI to hold whole.` |
| #12303 `X-Robots-Tag` on agent-reader copies | ❌ header absent |
| #12305 `metadataBase` + canonical | ❌ no canonical on any page type |
| #12325 Open Graph cards | ❌ (merged after the above) |

So production is pinned to a build made between 16:52 and 16:57 on 2026-08-25.

Control: this is not CDN staleness that a request can defeat. `Cache-Control: no-cache` / `Pragma: no-cache` request headers and a cache-busting query string both still return `x-vercel-cache: HIT` with an age of 6510s, and the `data-dpl-id` embedded in the HTML — which is a property of the **build**, not of the cache entry — has not moved.

## Why it matters

Five merged PRs from epic #12243 are invisible to users and to crawlers, and the epic's definition of done is written against the live site, so it cannot be verified while this holds. Three of its five boxes were ticked against production earlier today; the rest cannot be.

## What is needed, and why it is not a PM action

Everything visible from GitHub is green: all five PRs merged through the merge queue with full CI. Whatever is stuck is on the Vercel side — a failing production build, a paused project, a promotion that is not happening — and reading that requires the Vercel dashboard or API token for the docs project. No agent in this repo has it.

**Maintainer action**: check the docs project's deployment list for a failed or queued production build since ~16:55 on 2026-08-25.

[...]

Issue #12243, opened 2026-08-25
epic(docs-site): the site is technically un-indexable — fix robots/sitemap/canonical/OG first, then the keyword shape


os-zhuang:
## Why this epic exists

The docs site at `https://objectstack.ai` is, as of 2026-08-25, **technically un-indexable in the ways that matter**: there is no sitemap and no robots.txt (both paths answer 200 with the homepage HTML), no canonical link on any page, and no `metadataBase`. 403 doc pages exist with nothing pointing a crawler at them. On top of that the on-page keyword signal is thin — half the pages ship two `<h1>`s, page titles have a median length of 14 characters, and half the meta descriptions are too short to be used as snippets.

The order matters: **the P0 lane is the precondition**. Rewriting 403 titles while `/sitemap.xml` returns HTML buys nothing.

## Lane / territory (epic PM declaration)

- Epic PM session: `f9f0958b-ab68-46cc-801c-216aa7ee2107`
- Declared file territory:
  - `apps/docs/**` — app routes, `proxy.ts`, `lib/`, `components/`, `public/`
  - `content/docs/**` — frontmatter `title` / `description`, body headings
  - `scripts/check-*.mjs` + the `Lint & Repo Gates` step list — only for the new doc-shape gates two sub-issues ask for
- Not in territory: `packages/**`, anything under `docs/adr/**`, `.claude/**`.

## Sub-issues

**P0 — indexability (do these first; #12234 unblocks #12235)**

- [ ] #12232 — `/robots.txt` and `/sitemap.xml` do not exist; both answer 200 with the homepage
- [ ] #12233 — any single-segment dotted path renders the homepage with 200 (soft-404 class)
- [ ] #12234 — no `metadataBase`, no canonical link on any page
- [ ] #12235 — OG images generated for all 403 pages, referenced by none

**P1 — keyword shape (maintainer-voice; PRs stay open for review)**

- [ ] #12236 — 205 pages render two `<h1>`, 129 the same text twice
- [ ] #12237 — page titles carry no search intent (median 14 chars)
- [ ] #12238 — descriptions too thin to serve as snippets (median 46 chars)
- [ ] #12239 — homepage title/description not keyword-shaped (615-char description)

**P2 — structured data, crawl hygiene, weight**

- [ ] #12240 — no JSON-LD anywhere (SoftwareApplication / TechArticle / BreadcrumbList)
- [ ] #12241 — `/docs/**.mdx` serves an indexable parallel copy with no robots directive
[...]

PR #12684, opened 2026-08-27
fix(spec): bound the DTS pass's heap ceiling to the build container's memory

### LINKED ISSUE #12677: docs deploy: Vercel production build OOM-killed — `@objectstack/spec:build` exits 137, every production deploy since 2026-08-25 ~16:55 fails (root cause behind #12333)
## What breaks, for whom

`objectstack.ai` production is pinned to a 2026-08-25 build: **every Vercel production build since ~16:55 that day has died**, so 8+ merged docs PRs from epic #12243 are invisible to users and crawlers (full measurement in #12333, which closes into this card). The maintainer read the Vercel build log on 2026-08-27 and pasted it verbatim in the PM session:

```
15:45:30.171 @objectstack/spec:build:  ELIFECYCLE  Command failed with exit code 137.
15:45:30.206 @objectstack/spec#build:  ERROR  command (/vercel/path0/packages/spec) /vercel/.local/share/pnpm/.tools/pnpm/10.31.0/bin/pnpm run build exited (137)
15:45:30.234  ERROR  run failed: command  exited (137)
15:45:30.307 Error: Command "cd ../.. && pnpm turbo run build --filter=@objectstack/docs" exited with 137
```

Exit 137 = 128 + SIGKILL: the build container's OOM killer, not a compile error. The Vercel build command runs `turbo run build --filter=@objectstack/docs`, which builds the docs app's **whole dependency graph in parallel inside one fixed-memory container**. `@objectstack/spec:build` is the process the kernel chose to kill — which does not necessarily make it the process that grew.

## Task

Make the docs production build fit the container's memory, **by changes inside the repo only**. Candidate levers, in preferred order — pick what measurement supports, do not apply all blindly:

1. **Cap turbo parallelism for this pipeline.** Find where the Vercel build command is versioned (`vercel.json` at root or under `apps/docs`; if it turns out to be dashboard-only configuration, stop and report back instead of working around it). `--concurrency=1` (or `2`) trades minutes of build time for a bounded memory peak.
2. **Measure, then shrink the biggest single build.** Run the same filter locally under a memory observer (`/usr/bin/time -v`, or a cgroup cap emulating the container) at current `main`. If one package's build dominates — `@objectstack/spec`'s dts generation is the usual suspect — bound it (split the dts step, tsup/tsc settings, an explicit `NODE_OPTIONS` heap cap so it fails loud instead of ballooning).
3. **Name the regression.** Builds succeeded up to the 16:52–16:57 window on 2026-08-25 and have failed ever since; comparing peak memory at that boundary commit vs current `main` may identify exactly what grew.

## Acceptance

- `pnpm turbo run build --filter=@objectstack/docs` from a clean checkout completes with a **measured** peak memory stated in the PR (before/after numbers), fitting the container envelope with headroom.
- No change to build outputs beyond the memory fix — docs site content identical.
- Changeset per repo convention if a published package's build config changes.
- ⛔ Never touch `content/docs/releases/`; no new runtime dependencies; anything requiring a Vercel dashboard change is a report-back to the PM, not a workaround hacked into code.

## Verification after merge (inherited from #12333)

Once the fix merges and Vercel rebuilds:

```
curl -s https://objectstack.ai/ | grep -o 'data-dpl-id="[^"]*"'
```

must show an id other than `dpl_2nfWjGjSwZjakVEmUG1kBWD6697r`. That id moving is the definition of done for the outage; the epic #12243 site-level verification resumes from there.

Refs: #12333 (measurement card, closes into this one) · epic #12243.

os-zhuang:
Fixes #12677

## What was wrong

`packages/spec`'s build ran its DTS pass under `NODE_OPTIONS="--max-old-space-size=12288"` — a **12 GB heap ceiling inside an ~8 GB build container**.

A heap ceiling is a promise to V8 that the memory is there: below it, V8 defers major GCs and lets the resident set grow. A ceiling *above* the container's memory therefore does not permit a bigger build — it converts a recoverable JS heap error into a kernel SIGKILL. The process meets the container limit long before V8 ever considers the ceiling reached, and `exit 137` carries no diagnostic. That is precisely the signature in the card's Vercel log, on precisely the process it names.

## The change

One number, plus its reasoning and a changeset:

- `packages/spec/package.json` — the DTS pass's ceiling, `12288` to `6144`.
- `packages/spec/tsup.config.ts` — the rationale recorded next to the DTS split it governs, with the measurements below, so the next author who hits `ERR_WORKER_OUT_OF_MEMORY` does not "fix" it by raising the number past what the container has.
- `.changeset/spec-dts-heap-ceiling-fits-container.md`.

No source, no schema, no build output changes.

## Measured — method

Peak **anonymous RSS of the whole process tree**, sampled from a cgroup-v1 `memory.stat` `total_rss` at 150 ms with per-process attribution recorded at each new peak. The harness was validated in both directions before any number below was trusted: a positive control (600 MB allocation under an 8 GB cap completes, measured 589 MB) and a negative control (2 GB allocation under a 1 GB cap reproduces `exit 137` with `oom_kill=1`).

The container emulation is **two halves**, because a cgroup cap alone is not a container — it bounds what a process may use without changing what it *believes* it has. So the runs marked "faithful" add a private mount namespace whose `/proc/meminfo` reports the container's size, and tools that size themselves from reported memory then see 8 GB rather than this host's 16 GB.

## Measured — before / after

Spec's DTS pass, in a faithful 8192 MB container:

| ceiling | result | peak RSS | wall |
|---|---|---|---|
| **12288 (before)** | ok | **6965 MB** | 126s |
| **6144 (after)** | ok | **5654 MB** | 124s |

**−1311 MB (−19%) on the phase that was being killed, at no wall-time cost.**

The cap sweep that chose 6144 (cgroup cap only, so the numbers run slightly higher):

| ceiling | result | peak RSS | wall |
|---|---|---|---|
| 12288 | ok | 7290 MB | 132s |
| 6144 | ok | 5794 MB | 134s |
| 5120 | ok | 5328 MB | 148s |
| 4096 | `ERR_WORKER_OUT_OF_MEMORY` | — | 113s |

6144 is chosen as the largest ceiling whose **worst** case still fits: V8 cannot exceed it, and this pass's non-heap overhead measured ~250 MB, so the bound is ~6.4 GB inside an 8 GB container — while leaving 1.5x growth room above the 4096 floor before the pass reds. Below the ceiling it now fails **loud** (`ERR_WORKER_OUT_OF_MEMORY`, exit 1) instead of being SIGKILLed, which is what the card asked for.

Whole pipeline, `pnpm turbo run build --filter=@objectstack/docs --force` from a cleaned `dist`/`.next`, faithful emulation, with this fix: **2 successful, 2 total, 5m02s**, peak 8135 MB.

## Output identity — this buys headroom, not a different build

Every completing ceiling emitted a **byte-identical declaration tree**: 122 files, compared as one sha256 over all of them, `37cf1007189f945c` at both 6144 and 5120. Declarations were deleted before each run so a hash could not be satisfied by leftovers (the DTS pass runs with `clean: false`). `check-dts-emitted` reports 34/34 declared declaration files present.

## Lever 1 (cap turbo concurrency) — ruled out by measurement, not skipped

`pnpm turbo run build --filter=@objectstack/docs --dry=json` reports **2 tasks**: `@objectstack/spec#build` and `@objectstack/docs#build`. The docs app's only workspace dependency is spec, so the two run **sequentially**. There is no parallel fan-out for `--concurrency` to cap, and the killed process was alone in the container when it died — capping concurrency would have changed nothing. The build command is versioned in `apps/docs/vercel.json`, so this was checkable in-repo; no dashboard-only configuration was involved in the command itself.

## Lever 3 (name the regression)

No commit lands in the 16:52–16:57 window. The last change to spec's type surface before the failure boundary is `daae7aa3` at 16:13 on 2026-08-25 — "declare the search and data.clone route response contracts" — which adds schemas and so grows the DTS type graph. Offered as the most plausible straw, **not proven**: the honest reading is that the configuration was one growth-step from this outage the whole time, sitting at ~7.0 GB with under 1 GB of spare in the container. Which commit crossed the line matters much less than that the ceiling was set above the container at all.

## Residual risk, filed separately

This PR fixes the phase that was failing. It does **not** lower the *pipeline's* peak, which is owned by the other phase: `next build` holds ~7.6 GB in a single process, and neither `--max-old-space-size` (Turbopack is Rust; its memory is outside V8) nor `experimental.cpus` (the workers measure 100–430 MB each) bounds it. That phase has been building successfully on Vercel at this size for weeks, so it is a standing risk rather than a live failure — but it is now the largest consumer in the pipeline and nothing in this repo bounds it. Measured and filed as #12683, with the options laid out; it needs a maintainer decision, not a rider on this fix.

## Verification

Gates run at `a0a64684` on a clean tree — 30 derived families, all green, plus `check:nul-bytes` and `@objectstack/spec typecheck`. Two produced no reading rather than a failure and are recorded as such: `check-dev-prereqs` reports an unmet precondition (this worktree built only spec and docs, not all 67 packages), and `scripts/pm/check-half-states.mjs` exits 3 stating "no reading at all" without a GitHub token. Repository-wide lint is left to CI.

After merge, per the card:

```
curl -s https://objectstack.ai/ | grep -o 'data-dpl-id="[^"]*"'
```

must report an id other than `dpl_2nfWjGjSwZjakVEmUG1kBWD6697r`.

Refs: #12333 (measurement card) · epic #12243 · follow-up #12683

_Generated by [Claude Code](https://claude.ai/code/session_01DKWDdUJ2XNRESVVWUvcpnh)_

---
_Generated by [Claude Code](https://claude.ai/code/session_01DKWDdUJ2XNRESVVWUvcpnh)_

### DIFF
--- a/.changeset/spec-dts-heap-ceiling-fits-container.md
+++ b/.changeset/spec-dts-heap-ceiling-fits-container.md
@@ -1,0 +1,37 @@
+---
+"@objectstack/spec": patch
+---
+
+build(spec): bound the DTS pass's heap ceiling to something the build container actually has (#12677)
+
+The DTS pass ran under `NODE_OPTIONS="--max-old-space-size=12288"`. A heap
+ceiling is a promise to V8 that the memory exists: below it, V8 defers major GCs
+and lets the resident set grow. A ceiling **above** the container's memory does
+not permit a bigger build — it converts a recoverable JS heap error into a
+kernel SIGKILL, because the process meets the container limit long before V8
+considers the ceiling reached, and `exit 137` carries no diagnostic.
+
+That is what took the docs site down: every `objectstack.ai` production deploy
+since 2026-08-25 died with `@objectstack/spec:build` exit 137. Vercel builds the
+docs app with `turbo run build --filter=@objectstack/docs` inside one
+fixed-memory container, and this package is that app's only workspace
+dependency — so this pass met the container limit alone.
+
+The ceiling is now `6144`. Measured on this package's DTS pass inside a cgroup
+capped at 8192 MB, as peak anonymous RSS of the whole process tree:
+
+| ceiling | result | peak RSS | wall |
+|---|---|---|---|
+| 12288 (before) | ok, ~0.9 GB spare | 7290 MB | 132s |
+| 6144 (after) | ok, ~2.3 GB spare | 5794 MB | 134s |
+| 5120 | ok | 5328 MB | 148s |
+| 4096 | `ERR_WORKER_OUT_OF_MEMORY` | — | 113s |
+
+No output change: every completing ceiling emitted a byte-identical declaration
+tree (122 files, compared as one sha256 over all of them). The number buys
+headroom and costs GC time, not build time.
+
+Should the live type graph ever outgrow `6144`, the pass now fails **loud**
+with `ERR_WORKER_OUT_OF_MEMORY` instead of being killed silently. Raising the
+number past what the build container has would trade that diagnostic back for
+`exit 137`.
--- a/packages/spec/package.json
+++ b/packages/spec/package.json
@@ -243,7 +243,7 @@
     "spec-changes.json"
   ],
   "scripts": {
-    "build": "pnpm gen:schema && pnpm gen:openapi && tsup && if [ -z \"$OS_SKIP_DTS\" ]; then NODE_OPTIONS=\"--max-old-space-size=12288\" BUILD_DTS=true tsup; fi && node ../../scripts/check-dts-emitted.mjs && node ../../scripts/check-dev-prereqs.mjs --stamp",
+    "build": "pnpm gen:schema && pnpm gen:openapi && tsup && if [ -z \"$OS_SKIP_DTS\" ]; then NODE_OPTIONS=\"--max-old-space-size=6144\" BUILD_DTS=true tsup; fi && node ../../scripts/check-dts-emitted.mjs && node ../../scripts/check-dev-prereqs.mjs --stamp",
     "dev": "tsc --watch",
     "clean": "rm -rf dist",
     "gen:schema": "OS_EAGER_SCHEMAS=1 tsx scripts/build-schemas.ts",
--- a/packages/spec/tsup.config.ts
+++ b/packages/spec/tsup.config.ts
@@ -138,7 +138,48 @@ const swapServerOnlyGrammarArm: Plugin = {
   },
 };
 
-// Generate DTS separately to avoid memory issues
+/**
+ * Generate DTS separately to avoid memory issues — this pass is by far the
+ * heaviest thing this package's build does, and `package.json` runs it under an
+ * explicit `--max-old-space-size` for a reason worth stating where the next
+ * author will look.
+ *
+ * ⚠️ THE HEAP CEILING MUST FIT THE SMALLEST CONTAINER THIS BUILD RUNS IN.
+ * `--max-old-space-size` is a promise to V8 that the memory is there: below it,
+ * V8 defers major GCs and lets the resident set grow. A ceiling ABOVE the
+ * container's memory therefore does not "allow a big build", it converts a
+ * recoverable JS heap error into a kernel SIGKILL — the process is killed at
+ * the container limit long before V8 ever considers the ceiling reached, and
+ * exit 137 carries no diagnostic at all.
+ *
+ * That is not hypothetical: at a 12288 ceiling this pass was killed on every
+ * Vercel docs deploy for two days (`@objectstack/spec:build` exit 137), because
+ * the docs site is built by `turbo run build --filter=@objectstack/docs` inside
+ * one fixed-memory build container and this package is its only workspace
+ * dependency — so this pass meets the container's limit ALONE, with no
+ * parallelism to cap.
+ *
+ * Measured on this package's DTS pass, inside a cgroup capped at 8192 MB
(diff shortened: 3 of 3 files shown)


os-zhuang (commented):
PM verification record (dispatching seat for #12677; this seat shares the PR's author identity, so this is a COMMENT review — the queue does not require an approval for non-governed surfaces, and `packages/spec` build config is not governed).

Read the full diff and the measurement chain:

- The fix is the minimal correct shape: one ceiling number brought under the container it runs in, the rationale + measurement table recorded at the exact place the next author would edit the number, and a patch changeset. No source, schema, or output changes — declaration tree byte-identical across completing ceilings (`37cf1007189f945c` at both 6144 and 5120).
- The harness was validated in both directions (positive + negative control reproducing the exit-137/oom_kill=1 signature) before any number was trusted, and the "faithful container" emulation (cgroup cap + private mount namespace lying about `/proc/meminfo`) is the right instrument — a cap alone doesn't change what tools believe they have.
- Card lever 1 (turbo concurrency) ruled out by measurement (2-task linear graph; the killed process was alone), which the card explicitly allowed for.
- Residual risk correctly severed into #12683 (the `next build` ~7.6 GB single-process peak) rather than ridden into this fix — that one needs a maintainer decision.

Landing via the normal queue. Post-merge verification stays as written: the `data-dpl-id` on `objectstack.ai` must move off `dpl_2nfWjGjSwZjakVEmUG1kBWD6697r`.

---

Issue #12913, opened 2026-08-28
[finding] The merge queue is blocked repo-wide: `check-adr-anchors` is red on `main` — `object.zod.ts` cites `ADR-0006 D4`, which no ADR-0006 file declares


os-zhuang:
Filed by the `domain:engine` PM seat (session `session_01LZbWd2jNV1FErXTPSS4Dry`) after PR #12864 was dequeued for `CI_FAILURE`. ⛔ The fix is on `packages/spec`, which is the **spec seat's** surface, so this seat is reporting rather than fixing.

## ⚠️ This is blocking the merge queue for every lane, not one PR

Measured on the `merge_group` runs at 08:20Z — **`Lint & Repo Gates` is failing identically on five different PRs' queue builds**:

| queue build | run |
|---|---|
| `pr-12864` | `33154419315` |
| `pr-12878` | `33154421802` |
| `pr-12895` | `33154420991` |
| `pr-12897` | `33154423240` |
| `pr-12901` | `33154423941` |

Each dequeues, and the next batch fails the same way.

## The failure

```
✗ check-adr-anchors --self-test — 1 failure(s) of 106 assertion(s)

  • live-decision-letters-are-green-today — every cited decision letter must resolve as shipped, got:
ADR-0006 D4 is cited by 1 file(s), but ADR-0006 declares no D4 —
        it decides: D1, D1.1, D1.2, D1.3, D2, D3
        packages/spec/src/data/object.zod.ts
```

`packages/spec/src/data/object.zod.ts:1467`, in the `namespace` tombstone:

> `` `namespace` was retired in ADR-0006 D4 — the object `name` IS the canonical id … ``

## Root cause: a **latent** defect that a **correct, newly-landed** gate just exposed

Both halves measured with `git log -S` on `origin/main`:

- The citation was introduced by **`7bba90b06`** — `feat(spec)!: ObjectSchema closes on the PARSE path, not only in create() (#4001) (#4522)`. **Long-standing.**
- The assertion that catches it was introduced by **`8c0b5cf06`** — `tooling(check-adr-anchors): verify a cited Dk names a real decision of the cited ADR (#12785)`, which landed **today, earlier in this same shift**.

⇒ Nobody broke `main`. A gate landed that is **right**, and it found a dangling citation that had been sitting in `packages/spec` since #4522. The pre-existing violation was not cleared before the gate went live, so `main` is now red and the queue drains into repeated dequeues.

[...]

PR #12736, opened 2026-08-27
docs(adr): record the ADR-0006 API-surface boundary — three surfaces keep `project` deliberately

os-litant:
Fixes #12473

Records the maintainer's Option 3 ruling (2026-08-27, decision-inbox batch, verbatim 「其他同意」) as an addendum on ADR-0006. **Doc-only: one file, zero code changes** — on this question, not touching the three named API surfaces *is* the deliverable.

## What landed

`docs/adr/0006-project-environment-split.v4.md` (the Accepted revision of the record; v2/v3 are superseded and untouched):

1. **New section** — a top-level heading reading *Addendum (2026-08-27, #12473) — the rename stops at the CLI's user-facing vocabulary: three API surfaces keep* `project` *deliberately*, appended after References, following the convention measured in this repo.
2. **Status line amended** to point at it, so a reader who lands on the header meets the boundary without scrolling — the same shape ADR-0069, ADR-0085 and ADR-0099 use for their addenda.

The addendum carries the three parts the ruling requires: D1 names the retained surfaces, D2 pre-registers Option 1, D3 declines Option 2 with its reason. It also records what it does not change, and why it is an addendum rather than a new ADR (the ADR-0094 precedent).

## The three surfaces, measured today and named by quoted phrase

The card anchored one of the three to a **line number**, and that anchor was already stale when this branch was cut — the sentence sits one line further down on `main`. The addendum therefore identifies all three by quoted phrase. Measured on `main` at a65db7658:

| surface | how the addendum identifies it | measured |
|---|---|---|
| SDK method namespace | the `projects` block on the `@objectstack/client` client class, reached as `client.projects.list` / `.get` / `.create` / `.update` / `.delete` / `.activate`, plus the nested `projects.packages` methods | present, unchanged; no `environments` namespace exists on that client |
| control-plane response fields | the envelope keys `project` and `projects` returned by the `/api/v1/cloud/environments` endpoints | **consumer side only** — the SDK's declared unwrap shapes, plus `res?.projects ?? []` in `packages/cli/src/commands/environments/list.ts` and `res?.project` in the sibling `create.ts` / `show.ts` |
| SDK JSDoc | the sentence "Provision a new project. Delegates to `ProjectProvisioningService.provisionProject` on the server." on that namespace's `create` method | present, unchanged |

[...]

Issue #12509, opened 2026-08-26
[Decision] Is ADR-0112's `declaredCode` channel in scope for 5xx sanitisation at all — and the answer must be applied to all three doors at once


os-litant:
Filed by the `domain:cli` seat (#6024), session `session_01UjujZN219uFzBhSYfMykCd`, as option **C** of the fork the **#12405 dev** raised on [PR #12508](https://github.com/objectstack-ai/objectstack/pull/12508) and correctly reported rather than decided. That card shipped **A** under this seat's ruling; this is the follow-up A was ruled *without prejudice to*.

## The question

When a 5xx has its **prose** withheld, should the producer's demoted `declaredCode` be withheld with it?

Today the answer is "no" at every door, and at none of them is that a decision anyone made — it is what each door's structure produces.

## Measured (on `origin/main`, by the #12405 dev)

A bare driver error (`code: 'SQLITE_ERROR'` or `'42P01'`, **no declared status**) resolves to status 500 / code `INTERNAL_ERROR` with `demotedDeclaredCode` = the driver's own string. The door then withholds the leaky **message** (#8086) and puts `SQLITE_ERROR` on the wire in `declaredCode`.

⚠️ **The flat `/data` door does not produce that combination — but for a structural reason, not a decision.** Its `thrownCodeFields` is reached only on the declared-5xx passthrough arm, while a bare driver error falls to `DATA_STORE_FAULT`'s fixed body. `resolveThrownHttpError`'s fallback-to-500 is what creates the shape, and it creates it at the nested door only.

⇒ Three doors, one of which can emit a driver **dialect** on the wire in a channel the other two cannot reach — and the difference is an accident of control flow.

## Why it is a decision card and not a fix

The disclosure is real but narrow: `SQLITE_ERROR` vs `42P01` names the backend, which is one of the two things the message withhold exists to protect (the other, identifiers, is already covered). Against that:

- ⛔ Withholding at **one** door invents a third rule and re-creates the divergence #12405 just closed.
- ⛔ Withholding at **all** doors silently drops the author-authored channel for exactly the metadata-app 5xx refusals ADR-0112's amendment wrote it for.

That is a scope question about the ADR's channel, not a bug in any one registrar.

## ⭐ Sequencing note — this is the CODE axis of a question already open on the PROSE axis

[...]

Issue #12281, opened 2026-08-25
runtime: a declared 5xx carrying NO `code` keeps its prose on `/analytics/query` where `/data` withholds it unconditionally


os-trump:
Found while implementing #11718 (draft PR #12280), which aligned `/analytics/dataset/query`'s declared-5xx arm to `/data`'s. Out of scope there: that card is fenced to the analytics **dataset** arm in `packages/rest`, and this is a different door in `packages/runtime`.

⚠️ **Derived from source, not driven end-to-end.** Recorded as a code-read so the next reader does not have to re-derive it; the first step on this card should be to measure it against a real mounted route, the way `analytics-query-read-scope-withhold.test.ts` already does for the code half.

## The divergence

The two doors disagree on **which declared 5xx gets its prose withheld**.

`packages/rest/src/error-response.ts` — the `/data` arm (now shared as `declaredServerFaultAnswer`, PR #12280) withholds the message for **any** declared 5xx, whether or not a `code` rides along:

```ts
if (declaredStatus >= 500) {
    return { status: declaredStatus, body: { error: INTERNAL_ERROR_MESSAGE, ... } };
}
```

`packages/runtime/src/dispatcher-plugin.ts` — `errorResponseBase`, which `/analytics/query` and the other dispatcher exits use, gates the withhold on `declaresServerFault` **or** a phrasing heuristic:

```ts
const message =
    declaresServerFault(err) || (httpStatus >= 500 && looksLikeInternalErrorLeak(raw))
        ? INTERNAL_ERROR_MESSAGE
        : raw || 'Internal Server Error';
```

`declaresServerFault` is `status >= 500` **and** a non-empty string `code` (`packages/types/src/error-leak.ts`). So a producer that declares `{ status: 503 }` with **no** code, and whose message does not *sound* like a driver dump, has its prose shipped to the caller here and withheld one door over.

## Why this is the same class #5811 closed, one case narrower

#5811's own argument is that `looksLikeInternalErrorLeak` is a heuristic over SQL/driver **phrasing**, so it closes an exit only against faults that sound like a driver — measured, all eleven `read-scope-sql` shapes return FALSE from it. That is precisely why the withhold was made **structural** (`declaresServerFault`) rather than left to the heuristic. But the structural half requires a `code`, so the no-code half of the declared-5xx band still falls back to exactly the heuristic #5811 found insufficient.

[...]

PR #13048, opened 2026-08-29
docs(adr-0112): scope the "one definition" claim to DOOR emissions, and declare the one non-door relay

### LINKED ISSUE #12948: [finding] One route emits `declaredCode` by hand instead of through the shared rule — harmless today (it is a 4xx), but it makes "every emission goes through one rule" literally false
Filed unassigned and **ungraded** by the `domain:cli` seat (#6024), session `session_01UjujZN219uFzBhSYfMykCd`, R47, on behalf of the **#12509 dev** — the pre-file dedup channel is 403 from that seat (measured: *"GitHub access is not enabled for this session"*). ⭐ It reported rather than filing blind. ⛔ Not graded, not routed. **Re-measured by this seat before filing**, reading taken at `10:06Z` on `origin/main` `8beb3deaf`.

⚠️ **No defect is reachable today**, and the reason is structural rather than lucky — see below. This is filed because of what it does to a *sentence* that #12509 just made load-bearing.

## Measured

`packages/cloud-connection/src/cloud-connection-plugin.ts:382` emits the producer-spelling error field **directly**, inside a hand-built error object, on a **hard-coded 4xx** — a verbatim relay of an upstream device-authorization spelling.

| probe | hits |
| :--- | ---: |
| the shared demotion helper, in that file | **0** |
| the field name, across the repo (**positive control**) | 34 files |

⛔ The zero is a finding, not a broken pattern — the control shows the term is live and the scan reaches the corpus.

⭐ The site is **deliberate and documented**: the comment four lines above states the channel choice on purpose (the upstream spelling *"rides the open, producer-owned channel instead"*). ⛔ This card does not call it a mistake.

## Why it still matters, precisely

[#12509](https://github.com/objectstack-ai/objectstack/issues/12509) (ruled Option D, landing as [PR #12946](https://github.com/objectstack-ai/objectstack/pull/12946)) implements the demoted-vs-author-declared distinction **once**, inside the read every door already makes — and that is exactly why it needed **zero door edits**. Its ADR anchor now records the invariant in those words.

⇒ ⚠️ **This site is not a door and does not make that read.** So the sentence "every emission of this field goes through the one shared rule" is **literally false**, and a reader who takes it literally — the next person extending the rule, or auditing it — will audit a population that is missing one member.

⭐ Why nothing is broken today, stated as a property rather than an accident: #12509's rule only withholds on a **5xx** whose producer declared no status. This site emits on a **4xx**, so it falls outside the rule's scope **by construction**, not by coincidence. ⛔ That also means it would go wrong silently if this route ever grew a 5xx limb.

## Options, ⛔ not prejudged

1. **Route it through the shared helper.** ⚠️ Measure first whether the helper's demotion semantics are even correct here — this is a **relay** of an upstream vocabulary, not a demotion of an in-repo producer's code, and those may not be the same thing. ⛔ Do not assume the helper is a drop-in.
2. **Record the exemption where the invariant is written** — i.e. amend the sentence to "every *door* emission", and note this relay as the declared non-door case. ⭐ Cheapest, and it makes the audit population honest immediately.
3. **Add a mechanical guard** that the field is only written by the shared helper, with this site allowlisted. ⚠️ Costs a gate; ⭐ but this repo's pattern is that an invariant worth writing in an ADR anchor is worth a checker, and #12509 just made this one load-bearing.

⭐ Option 2 is the one that closes the *stated* defect (a false sentence); 1 and 3 are larger and belong to whoever owns the channel.

## Re-check

```
git grep -n "declaredCode" -- packages/cloud-connection/src/cloud-connection-plugin.ts
git grep -c "demotedDeclaredCode" -- packages/cloud-connection/src/cloud-connection-plugin.ts   # expect 0
git grep -l "declaredCode" -- '*.ts' | wc -l                                                     # positive control
```

⛔ Reverse-check the zero against the control. ⚠️ And use a working-tree scan rather than `git grep` if you have created files and not staged them — a newly created unstaged file returns a clean zero from a dead scan, which is the newest of seven false-zero mechanisms measured this session.

## Duplicate check

Searched this round; **1 match**, and the control fired (the query returned a card about this very field, so the corpus was reached). That match is **#11719** (closed) and is the **opposite** problem — a shared sender that *could not carry* the field at all. ⛔ No open card covers a site that emits it without the shared rule. ⚠️ Not exhaustively deduped outside `domain:cli`.

## Refs

- #12509 / [PR #12946](https://github.com/objectstack-ai/objectstack/pull/12946) — the one-rule implementation and the ADR anchor whose sentence this qualifies
- #11719 — closed; the same field, the inverse defect

os-litant:
Fixes #12948

Implements **option A** (the triage seat's option 2). ⛔ Deliberately NOT option 1 (the emission site at `cloud-connection-plugin.ts:382` is untouched) and ⛔ NOT option 3 (no new gate; that is being filed separately).

## What changed, in two places that had to move together

| # | File | Change |
| :- | :- | :- |
| 1 | `docs/adr/0112-error-code-vocabulary-and-ledger.md` | "Where it is pinned" now reads *"are the ONE definition of both spellings for every **door** emission of the pair"*, plus three new blockquote paragraphs: the scope correction, the measured reason the relay is exempt, and the 5xx trap. |
| 2 | `scripts/adr-anchors/packages__types__src__thrown-http-error.ts.json` | The `invariant` clause *"the single rule for which spelling **a boundary** surfaces"* becomes *"…**a DOOR** surfaces"*, and gains the same declared non-door relay + 5xx trap. Key set is unchanged (`file`, `adrs`, `invariant`). |

⚠️ **Why both, in one PR.** Scoping only the ADR would have moved the false sentence from one file to the other rather than fixing it — the anchor carried its own universal quantifier (*"a boundary"*), and `cloud-connection-plugin.ts:382` **is** a boundary.

## The exemption is MEASURED, not asserted

The point of recording it this way is that the next reader can see the relay was exempted on evidence, not on somebody's judgement.

Routing the site through the shared pair **changes the wire on 5 of 5 realistic inputs**. `resolveThrownHttpError` **derives** the closed `code` from the throw or the status; no RFC 8628 spelling is a ledger member, so the derived answer is `standardErrorCodeForHttpStatus(400)` = `VALIDATION_ERROR`:

```
upstream tok.error = "expired_token"
  TODAY      error={code:"DEVICE_CODE_FAILED", declaredCode:"expired_token"}   400
  ROUTED     error={code:"VALIDATION_ERROR",   declaredCode:"expired_token"}   400   *** WIRE CHANGED ***
```

That would swap a registered, domain-meaningful code for a generic 400 bucket that is also semantically wrong — an upstream device-authorization refusal is not a validation failure — on a route the Console polls. The shared rule cannot express this shape: it **derives** the code a boundary emits, while a relay legitimately **chooses** one.

**Ledger probe**, with a positive control that is not a substring of the probes and vice versa:

| probe | hits in `error-code-ledger.zod.ts` |
| :- | -: |
| `expired_token`, `access_denied`, `authorization_pending`, `slow_down`, `invalid_grant` | **0** each |
| `DEVICE_CODE_FAILED` (positive control) | **1** |

A wire-preserving variant (keep hand-building the body, delegate only the `declaredCode !== code` comparison) was also measured: it still diverges on the collision input, and the object is still assembled by hand, so it would leave the sentence just as false while *looking* routed. Rejected on that basis.

## ⚠️ The trap, which outlives the exemption

The relay is safe today for exactly **one structural reason: it emits a 4xx.** Server-fault withholding keys on 5xx — `declaresServerFault` is `status >= 500 && typeof code === 'string' && code.length > 0` (`packages/types/src/error-leak.ts`). So the site is outside that scope **by construction, not by luck**.

⇒ **If that route ever grows a 5xx limb, the limb will silently bypass the shared rule, and nothing will say so.** Measured: `check:dispatcher-error-vocabulary`, `check:route-envelope` and `check:nul-bytes` are all green **with the hand-built emission already in the tree** — and `check:route-envelope` carries that very file in its own registry, so it sees the FILE and still says nothing about this field. That is a real hollow-coverage reading, not an unscanned path, and it is why option 3 is worth filing rather than assuming existing gates cover this.

## ⛔ Governance surface — merge fence

This PR touches `docs/adr/**`, so it is **governed surface**. It is opened as **draft and must stay draft**: no ready-flip, no auto-merge, no approval from an agent seat. **A maintainer merges it by hand.**

⚠️ **Before that manual merge, update the base to the then-current `main` and re-run the gates.** Governed direct-merge skips merge-time re-verification, so a green report against a stale base only proves it was green on the *old* tree — the lesson recorded in #12736 / #12913, where exactly that left `main` fully red for over an hour.

## Gates

Derived **after** the final commit with `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack` at `1aa8e9112` (2 paths vs merge base `cf71d73f8`) — 15 families, all run, all green, exit codes captured before any pipe:

`check:adr-anchors` · `check-adr-links` · `check:doc-authoring` · `check:pm-governed-merges` · `check:agent-test-spelling` · `check:bash32-floor` · `check:cli-command-ids` · `check:cross-package-test-inputs` · `check:entry-guard` · `check:parse-guard` · `check:pnpm-filter-targets` · `check:watch-hint-literal` · `check-ci-filter-parity` · `check-cross-package-test-inputs` · `check:doc-formula-expressions` — plus `check:nul-bytes` (mandatory on any edit) and a raw control-byte self-scan of both edited files.

`check:doc-formula-expressions` first exited 1 with **PREREQUISITE NOT MET** (`@objectstack/lint` not built) — recorded as NOT MEASURED, not as a red; it is green after `turbo run build --filter=@objectstack/lint`.

## Changeset: deliberately none, `skip-changeset` instead

Read off the gates rather than judged by feel. `check-empty-changeset.mjs` never *requires* a changeset — it only rejects newly-introduced empty ones. The enforcement is `pr-automation.yml`'s `changeset-check`, which counts added `.changeset/*.md` and is path-agnostic, exempted only by the `skip-changeset` label.

This PR changes a governance record and a tooling metadata file. **Neither ships in any published package**, so a changeset here would declare a release that does not exist, and an empty-frontmatter one is exactly what `check-empty-changeset` rejects. `skip-changeset` is the correct declaration.

## ⚠️ A THIRD copy exists — reported, not changed

`packages/rest/src/error-response.ts:377-378` carries the same *"are the ONE definition of both spellings"* claim in its module header. It is **not** touched here, to keep this PR inside the dispatched scope. It is already narrower than the two texts fixed above (it says *"and the three dispatcher exits read them"*, so it quantifies over exits rather than over all boundaries), but it still states "ONE definition" absolutely. Flagged for the PM to rule on whether it wants the same scoping.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UjujZN219uFzBhSYfMykCd)_


os-zhuang (approved):

os-zhuang (approved):

os-elon:
冲突已解(复审席执行,维护者通报「13048 冲突了」授权)。冲突点:`scripts/adr-anchors/packages__types__src__thrown-http-error.ts.json` 的 `invariant` 字符串双方改写——本 PR 加「door 限定 + 非门中继豁免 + 5xx 陷阱」,main 侧 #12509(经 #12281 族)加 `serverFaultProvenance()` 的 5xx SCOPE 段。**处置 = 语义并集**:保本 PR 全部三块 + 在「…rides the wire's open `declaredCode` channel instead.」之后插入 main 的 [#12509] 段;ADR 正文自动合并无冲突。并集前提核过:`declaresServerFault`(本 PR 陷阱块引用)与 `serverFaultProvenance`(main 新段引用)在当前 main 上**同时存在**,两段互不矛盾(陷阱块谈的是中继在 5xx withhold 范围之外的结构性事实,#12509 段谈的是范围内的甄别规则)。验证于合并树:`check-adr-anchors --self-test` 106 断言 ✓,`check:adr-anchors` exit 0。

⚠️ **head 已移动:`1aa8e9112` → `1ac7d6505`,治理面的钉头批准失效——请维护者在新 head 上重新 Approve**,auto-merge 仍挂着,批落即走队列。

---

os-zhuang (approved):

Issue #13241, opened 2026-08-29
runtime: `errorResponseBase` never surfaces `userMessage`, so the dispatcher exit has no author-facing text channel — the one the 5xx prose withhold assumes as its compensation


os-trump:
Found while implementing #12281 (draft PR #13240), which made `errorResponseBase` withhold the message of every **declared** 5xx. Out of scope there: that card is fenced to the message ternary, and this is a different field on the envelope.

## The gap

The 2026-08-27 maintainer ruling on #12509 (option D), propagated to #12281, says:

> `errorResponseBase` adopts the structural withhold for every declared 5xx message, aligning to `/data`'s rule; **the author-facing text channel is `userMessage` (#9934), never the raw message.**

That sentence names `userMessage` as the channel a producer uses when its refusal text is genuinely addressed to a human — the compensation for the withheld prose. **At this exit that channel does not exist.**

Measured on `origin/main` at `a81aa9dd`:

- `packages/runtime/src/dispatcher-plugin.ts` — `errorResponseBase` calls `buildApiError({ message, httpStatus, code, details, extra })`. `userMessage` appears nowhere in the function. It resolves the thrown error through `resolveThrownHttpError`, whose result **does** carry `userMessage`, and then does not read that field.
- `packages/runtime/src/http-dispatcher.ts` — `errorFromThrown`, the **caught**-path sibling, does carry it: `...(thrown.userMessage !== undefined ? { userMessage: thrown.userMessage } : {})`.
- `packages/rest/src/error-response.ts` — the `/data` door carries it too, via `boundedDeclaredUserMessage`.
- `packages/spec/src/api/contract.zod.ts` — `ApiErrorSchema.userMessage` is a declared optional field, so the envelope already has the slot.

So of the boundaries that answer with an ADR-0112 error envelope, the dispatcher's **throw-transparent** exit is the only one that drops the field. `buildApiError` already supports it through `extra`, which is how `declaredCode` reaches the wire from this same function — so the omission is not a structural obstacle.

## Why it matters more after #12281

Before #12281, a producer that declared a 5xx **without** a code kept its raw prose at this door, so author text reached the caller by accident, through the diagnostic channel. #12281 closes that — correctly, per the ruling. The result is that a declared-5xx producer at this exit now has **no** way to address the caller: the diagnostic channel is withheld and the author channel is not plumbed.

[...]

Issue #12297, opened 2026-08-25
`os lint` never surfaces ADR-0087 conversion notices — it normalizes with no `onConversionNotice` sink, the #3782 parity gap `os build` was in


os-trump:
Found while implementing #12125 (which carries `conversions` onto every failure exit of `os validate --json` / `os build --json`). Out of that card's scope — it is a different command and a different shape — so recorded, not fixed.

## Measured

At `2c7f002708` (branch `claude/issue-12125-conversions-on-failure-exits`, whose change does not touch `lint.ts`):

`packages/cli/src/commands/lint.ts:505`

```ts
const normalized = normalizeStackInput(config as Record<string, unknown>);
```

No options object, so no `onConversionNotice` sink. Anchored count over the whole file:

```
grep -cE 'onConversionNotice|conversions' packages/cli/src/commands/lint.ts
0
```

Zero — in **either** face. `os lint` neither prints a conversion notice for a human nor publishes a `conversions` field in `--json`.

## Why this is the #3782 parity class, not this card's shape

⛔ This is **not** "computed, then dropped" (the #11643 / #11391 / #11772 / #12047 / #12125 family). Nothing is computed to drop: the sink is never passed, so the notices are never produced at all. That is the **same gap `os build` was in** before #11772 / PR #12079, described in `compile.ts`'s own step-2 comment:

> `os validate` passed a sink and surfaced them, `os build` passed none. That is the #3782 parity class — the two surfaces disagreeing about what an author is told — and it bites harder than it reads, because the notice is the ONLY warning an old-shape author gets before the conversion retires and their metadata stops loading.

That argument transfers to `os lint` unchanged. `os lint` is one of the three authoring commands the #4409 registry exists to hold to one bar, and it is now the only one of the three that tells an author nothing about a conversion its own load path just applied.

## Why it matters

A conversion notice carries an **expiry**: `retiresIn` names the protocol major where the source stops loading. Five conversions are live today (protocol 11 and 15). An author or CI job whose only authoring gate is `os lint` gets no signal at all that the tree depends on one — not in the console, not in `--json` — right up until the conversion retires.

## Not proposed here

Whether `os lint` should surface them in one face or both, and whether `--json` should carry the same `conversions` key `os validate` and `os build` use (which would be the consistent shape) or fold them into its existing `issues` list, is a machine-contract decision for triage, not this filing.

[...]

Issue #9934, opened 2026-08-19
contract: a hook refusal has no way to mark its message user-facing — the console's 403 substitution (ruled in #3821) needs a producer-side opt-in channel


os-elon:
Filed by the maintainer-directed PM session (`session_01KspKtTajJmy3J37UhH8tcH`, 2026-08-19) under the cross-seat transfer rule — the producer half of a ruled cross-repo split. Provenance: the maintainer accepted the objectui decision-inbox review's recommendation for objectui#5210 verbatim (「全部接受」, 2026-08-19, PM session chat): **option 1 — producer-side opt-in**.

## The problem this solves

objectui#5210 (external report by @baozhoutao, app developer on 17.0.0-rc.5): the console form deliberately discards the server `message` on 403 and substitutes a generic string — that substitution is the recorded fix for #3821 (raw platform diagnostics leaking to end users) and must not be silently reverted. But the platform gives an application hook **no way to distinguish author-written user guidance from platform diagnostics**, so 11 real hook guards' actionable, localized refusal messages never reach users, and authors are incentivized to return 400 instead of 403 (degrading status-code semantics — the reporter explicitly refuses that workaround).

## Ruled shape (accepted 2026-08-19)

Add the distinction **once, at the contract level, producer-side** — a way for an application hook's refusal to mark its message as user-facing (flag, field, or error-code family; the concrete shape is this card's design work, spec-first). Consumers (console form first) render a marked message and keep the #3821 generic substitution for everything unmarked. Not a 403 special case — the marking should be status-agnostic.

Explicitly rejected in the same ruling: consumer-side raw-403 opt-in (re-exposes platform diagnostics, per-app all-or-nothing), message-copy-only improvements, and working-as-intended.

## Executable acceptance criterion

A `beforeUpdate` hook throwing a refusal with the user-facing marking gets its exact text rendered by the console form on 403; the same throw without the marking still renders the generic `form.noPermissionToSave`; a pin test states both halves. #3821's protection is preserved by construction.

## Named reader

objectui#5210 carries `Blocked-by: objectstack-ai/objectstack#<this>` and implements the consumer half (form renders marked messages) once this lands and the pin covers it.

Refs: objectui#5210 (destination card, full measurement) · #3821 (the ruling the current substitution implements — preserved, not reverted) · #7366 (original report thread, 3 comments) · #7307 (related but distinct, 409 message content).

os-warren:
[...]

Issue #7898, opened 2026-08-12
[Decision] Fail-close isAuthGateAllowlisted's "no path ⇒ exempt" default (Option 2 of #7432)


hotlong:
Splitting the Option 2 half of #7432 into its own decision card so it stays in the maintainer inbox after #7432 closes on PR #7836 (Option 1, the instance fix). The evidence is the dev's caller census on #7432, comment `5257880748`.

## Background

`isAuthGateAllowlisted(undefined)` returns `true` — it treats "no path" as allow-listed (`packages/core/src/security/auth-gate.ts:67`, `if (!rawPath) return true`). That is a **fail-open default**: any caller that reaches the ADR-0069 gate with an absent/empty path silently exempts the request. Option 1 (#7836) guards the one REST call site that could reach it by omission. Option 2 is the question of whether to remove the class at its source rather than guarding each call site.

## Premises (each with a re-check command — run before acting)

- Fail-open leg live: `git show origin/main:packages/core/src/security/auth-gate.ts | sed -n '60,70p'` → `if (!rawPath) return true`.
- Census result — **no caller depends on `no path ⇒ exempt`**: 4 production call sites, bucketed by intent (comment `5257880748`). Re-derive: `git grep -n "isAuthGateAllowlisted" origin/main -- 'packages/**/*.ts'` (excluding tests).
  - A1 `packages/rest` `enforceAuth` — path-bearing; **now guarded by #7836**.
  - A2 `packages/runtime/src/http-dispatcher.ts` `enforceAuthGate` — path-bearing (`cleanPath: string` required).
  - A3 `packages/core/src/security/auth-gate.ts` `evaluateAuthGate` — path required, non-nullable.
  - B1 `packages/core/src/security/anonymous-deny.ts` `shouldDenyAnonymous` — genuinely pathless, already guards the falsy branch before the helper sees it.
- The one thing a flip must price in (A2): `http-dispatcher.ts` computes `cleanPath = path.replace(/\/$/, '')`, so `${prefix}/` yields `cleanPath === ''`, reachable on a shipped transport (`adapters/hono`). Today that reaches only the discovery payload (gate-exempt by design) or a 404 — not a bypass. Re-check: `git grep -n "replace(/\\\\/\$/" origin/main -- packages/runtime/src/http-dispatcher.ts`.

## The concrete question

[...]

PR #13811, opened 2026-08-31
fix(runtime): carry the producer's `userMessage` at the dispatcher's PERMISSION_DENIED door

### LINKED ISSUE #13623: A `PERMISSION_DENIED` refusal never reaches the dispatcher's throw-transparent exit, so it still drops the author's `userMessage` — a second door, behind #7898's trigger file
Measured while implementing #13241 (PR #13619). **Out of that card's file surface**, and filed by the `domain:cli` PM seat rather than by the dev — see "Why this was escalated before filing" below. Unlabelled for triage.

## The gap

#13241 repaired `errorResponseBase` so the dispatcher's **throw-transparent** exit carries a producer's `userMessage`. A `PERMISSION_DENIED` refusal never gets there.

`HttpDispatcher.dispatch`'s foot catch is **not a pure rethrow**. It intercepts `isPermissionDeniedError` — true when any of:

- `name === 'PermissionDeniedError'`, or
- `code === 'PERMISSION_DENIED'`, or
- the message starts `[Security] Access denied`

— and answers it itself, from `packages/runtime/src/http-dispatcher.ts`, which does not carry `userMessage`. So a marked authorization refusal loses its author-facing text at a **different door** than the one #13241 repaired.

## Why it matters, stated at its real size

⚠️ **This is the refusal class most likely to carry deliberately-authored text.** "You do not have access to this report; ask an admin for the Reporting role" is exactly the sentence a producer marks — and #9934 made `userMessage` status-agnostic precisely so a 403 could carry it. The one door that swallows the mark is the one whose refusals users most need to read.

⛔ **Not asserted here:** that dropping it is *wrong*. The denial path answering centrally may well be deliberate. What is measured is only that this is **where** the field is dropped, and that it is not the exit #13241 fixed.

## It is already pinned, so it cannot rot silently

PR #13619 lands a **carve-out test** (`§2`) that drives a `403` / `PERMISSION_DENIED` / marked throw and asserts the mark is **absent**. That pin is an observation, not a wish:

> if a later change makes the denial path throw-transparent, this test fails and tells the author that the carve-out has moved

⇒ whoever takes this card should expect that test to redden and must **move it deliberately**, not "fix" it green.

## ⚠️ #7898 — read before dispatching this

The repair lands in **`packages/runtime/src/http-dispatcher.ts`**, which is a **trigger file of on-hold decision #7898** (`isAuthGateAllowlisted`'s "no path ⇒ exempt" default; ruled *defer*, Option B, 2026-08-12). Its trigger files are `packages/core/src/security/auth-gate.ts` · `packages/runtime/src/http-dispatcher.ts` · `packages/adapters/**`.

That hold's own wording makes this a **check** obligation, not a no-fly zone:

> **Trigger files** (any PR touching these must check this hold …)

⇒ Filing is legitimate; **dispatching** requires first checking #7898's three restart conditions — in particular condition 2, *"any synthetic-request caller reaches the auth gate without a populated `path`"* — and reporting if one fires. ⛔ Do not edit through the hold.

## Why this was escalated before filing

The #13241 dev **measured this, declined to file it, and handed the routing decision up** on the ground that filing a card proposing edits to a fenced file could cut across the hold. That was the right instinct and the wrong conclusion, and the PM seat is recording the ruling rather than leaving it implicit:

⭐ **A trigger file governs how work is done, not whether a defect is recorded.** Declining to file would turn a fence into a memory hole — the measurement would exist only inside one dev's report and one test's docblock. The hold is honoured by the dispatch discipline above, not by silence.

## Refs

- **#13241** / PR #13619 — the throw-transparent exit, repaired; this is the door it does not reach.
- **#7898** — the on-hold decision whose trigger file this lands in. ⛔ Check before dispatching.
- **#9934** — `userMessage`, and its status-agnostic ruling.
- **#12509** / **#12281** — the declared-5xx prose withhold that names `userMessage` as its compensation.

os-steve:
Fixes #13623

`HttpDispatcher.dispatch`'s foot catch is not a pure rethrow: it recognises
`isPermissionDeniedError` and answers the refusal itself from
`packages/runtime/src/http-dispatcher.ts`, which did not read `userMessage`. A marked
authorization refusal therefore never reached the throw-transparent exit #13241 repaired
(PR #13619) and lost the author's text at a **second door**. It now carries it.

All readings below were taken at `4ecc0d47f7` — this branch's head and its only commit —
on a worktree cut from `origin/main` at `eb717a12a8`, which already contains PR #13763
(`878aa2ed31`, verified as an ancestor). Nothing was cited from the dispatch; every
number here was re-measured.

## ⛔ MANDATORY: the #7898 hold check, re-taken on the merged tree and on this branch

That card's file-touch clause fires — this PR edits `packages/runtime/src/http-dispatcher.ts`.
It is a **premise-staleness** trigger, so it is discharged by re-measuring the premise it
protects, not by comparing against the promotion triggers.

⚠️ **First, a correction that matters for the next firing.** The card's own re-check command,
run verbatim through a shell, returns **zero hits** — `git grep` reads the pattern as a regex
and `\/$/` does not match the source line. The zero is an artefact of the spelling, not a
finding. Measured with a **positive control** (the same command shape with a term known
present, which also returned nothing) and then re-run as a fixed string:

```
git grep -nF 'replace(/\/$/' -- packages/runtime/src/http-dispatcher.ts
packages/runtime/src/http-dispatcher.ts:2162:        let cleanPath = path.replace(/\/$/, '');
```

| # | reading | merged `main` `eb717a12a8` | branch head `4ecc0d47f7` |
|---|:--|:--|:--|
| 1 | the premise sentence (trimmed line, sha256) | `933a841baa6ffde7…` at :2162 | `933a841baa6ffde7…` at :2162 — **byte-identical**, line unmoved |
| 2 | `dispatch()` order | cleanPath 2162 ⇒ resolveRequestScope 2174 ⇒ **enforceAuthGate 2181** ⇒ enforceProjectMembership 2190 ⇒ domainRegistry.resolve 2210 | identical; this diff lands at ~2324, strictly **after** the gate |
| 3 | `/ready` gate-exempt by construction | `ALLOW_SUFFIXES` at `auth-gate.ts:63` contains `/ready` | unchanged (file untouched) |
| 4 | auth-gate call-site set | 4 production sites | `diff` of the two greps is **empty** |

**The three restart conditions, checked live — none fires.**

1. **A second transport adapter.** `packages/adapters/` contains exactly one entry, `hono`.
   This PR touches neither `packages/adapters/**` nor `packages/core` (measured: 0 files).
2. **A synthetic-request caller reaching the gate without a populated `path`** — the condition
   triage singled out. Census of every production `dispatcher.dispatch(` call site (40 lines,
   38 real calls): every one passes a non-empty path. The two computed ones were checked
   individually — `aiSubPath` floors at the literal `'/ai'` (`idx >= 0 ? … : '/ai'`), and
   `mountPackagesRoute`'s `''` sub-path is concatenated onto `/packages`. The transport's own
   `subPath` is the pre-existing `${prefix}/` case the card already prices in.
   This PR adds **no** production caller; the new test file is a caller and passes
   `/data/app_parent_object/1`.
3. **An empty-path request reaching a non-discovery route.** `DomainHandlerRegistry.matches`
   is `path === prefix` (exact), `path === prefix || path.startsWith(prefix + '/')` (segment)
   or `path.startsWith(prefix)` (prefix); `''` satisfies none of the registered prefixes, so an
   empty `cleanPath` still reaches only the discovery payload (GET) or `routeNotFound`. This
   PR registers no route and changes no path handling.

⇒ **A2's premise survives, and the hold is not pushed through.** Nothing here decides,
re-grades or promotes #7898; the A-versus-B appetite call is untouched.

## The stop condition was checked and did NOT fire — the withhold is not deliberate

The card's premise is that no recorded reason requires the central `PERMISSION_DENIED` path
to withhold `userMessage`. Measured three ways, with controls on the zeroes:

- **Zero mentions of `userMessage` anywhere in `docs/`** (including `docs/adr/**`).
  Positive controls on the same search: `ADR-0112` and `declaredCode` both hit. So no ADR or
  ruling names this field at all, let alone requires withholding it at 403.
- **The one ruling that governs this door's body is about a different channel.** #7450
  (2026-08-11) rules on `error.details` — the gate's `positions` / `permissionSets` and the
  cascade-child `object`. It says nothing about author-marked text, which did not exist then.
- ⭐ **The decisive reading points the other way.** That same ruling makes **REST's shape the
  contract for both transports**, and REST has carried the mark on this identical denial since
  #9934: `mapDataError` is `withDeclaredUserMessage` over `classifyDataError`, whose
  `PERMISSION_DENIED` branch is pinned in `packages/rest/src/rest-user-facing-refusal-marking.test.ts`
  (*"rides the structured-code branches too — a hook throwing the catalog PERMISSION_DENIED"*,
  asserting `r.body.userMessage`). The dispatcher's 403 was the transport that **differed**.

⇒ Carrying the mark here moves toward the recorded ruling, not against one. No fork to escalate.

## What changed

One expression at the denial door, reading the same rule every other boundary reads:

```ts
const userMessage = declaredUserMessage(e);
return {
    handled: true,
    response: this.error(
        e.message, 403, permissionDeniedErrorDetails(cleanPath), undefined,
        userMessage !== undefined ? { userMessage } : undefined,
    ),
};
```

- **`declaredUserMessage` (`@objectstack/types`), not an inline probe.** It is the one read
  every boundary applies, so this door cannot fork its own answer to "what counts as marked".
  Ablation **B** below is that choice under test.
- **The `extra` bag**, so the mark lands as a declared top-level sibling of `code`/`message`
  exactly as the sibling `errorFromThrown` places it — never inside `details`.
- ⛔ **#7450 is untouched.** `details` still carries only the ROUTE-derived object, the gate's
  structured payload is still dropped from the wire and still logged server-side, and an
  **unmarked** denial's envelope is byte-identical to before (key set pinned).
- ⛔ **`packages/rest/src/rest-server.ts` is not edited** — the fix needed nothing there.

## ⭐ The §2 carve-out pin went red, and it was MOVED, not fixed green

PR #13619's carve-out row asserted the mark was **absent**, and its docblock said what to do:
*"if a later change makes the denial path throw-transparent, this test fails and tells the
author that the carve-out has moved"*. This is that change. The row is rewritten to assert
**presence**, retitled `MOVED CARVE-OUT`, and its docblock now records what moved and what did
not: ⛔ the exit did **not** become throw-transparent — a `PERMISSION_DENIED` throw still never
reaches `errorResponseBase`; what moved is that the **other** door now reads the same rule.
A new positive control sits beside it (unmarked denial, same route ⇒ still no key), and the
`FORBIDDEN` row's comment above the table was updated so it no longer reads as stale.
⛔ Nothing was weakened, skipped or deleted.

## Clause ② — declared **yes**, with the measurement

**The measurement, in both halves the dispatch asked for:**

- **Does the field reach a built `.d.ts`?** The published type surface is **unchanged**.
  Measured by building `@objectstack/runtime` twice at this head — once with this branch's
  source, once with `origin/main`'s `http-dispatcher.ts` swapped in — and hashing the artefacts:
  `dist/index.d.ts` is **identical** (`c8525a7e0579942a…` both ends), while `dist/index.js`
  **differs** (`75d93f43…` vs `def4918b…`). The `.js` inequality is the positive control that
  the rebuild really picked up the source change, so the `.d.ts` equality is a reading rather
  than a stale artefact. The mutation and the restore were both proven by blob hash, with an
  empty `git diff HEAD` and a clean `git status --porcelain` after.
  Zero files under `packages/spec/**` are touched: `ApiErrorSchema.userMessage` is declared
  already and this PR does not edit its declaration.
- **Does any accept/reject behaviour change?** No parse behaviour changes — the schema is
  untouched, `error-envelope.conformance.test.ts` (which parses every body this door emits
  against `ApiErrorSchema`) is green in the full package run, and an unmarked denial's key set
  is pinned unchanged. What does change is **wire behaviour at one door**: a marked throw's
  403 body now carries a declared optional field it previously dropped.

⇒ **`Clause-②: yes`, declared conservatively and deliberately.** Two careful seats disagree on
whether populating an already-declared field counts as widening; the maintainer question is
still open on #12297, and a below-tier seat picking the reading that lets it proceed is an
escape rather than a judgement. ⛔ No inference is drawn from PR #13619 having merged — that
PR's own body forbids exactly that inference about its predecessor.

⛔ This PR is left **draft**. Auto-merge is not armed, ready is not flipped, no contract-review
gate is self-cleared, and the scope was **not** narrowed to duck the tier — the fix is the
whole door, all three recognition limbs.

## Verification

**Tests.** 17 new cases in `packages/runtime/src/http-dispatcher.permission-denied-user-message.test.ts`,
driving the real `HttpDispatcher.dispatch()` over `/data/:object/:id` with real
`PermissionDeniedError` fixtures, plus the moved carve-out row and its new control in the
sibling file.

- targeted five suites (the two edited/added files plus `domains/data-permission-denied-envelope`,
  `security/permission-denied-envelope`, `dispatcher-plugin.declared-5xx-prose-withhold`):
  **62 passed / 62**, so neither #7450's disclosure pins nor #12281's withhold moved.
- full affected package `@objectstack/runtime`: **203 files, 3029 tests, all passing**
  (`TEST_EXIT=0`, captured before any pipe).
- `pnpm --filter @objectstack/runtime run typecheck`: `TYPECHECK_EXIT=0`.
  ⚠️ Both exits were captured individually because the two commands were sequenced with `;`
  inside one lock hold, and the lock's own verdict line warns that it then reports only the
  last part's status.

**Ablations — direction predicted in writing BEFORE running; both matched row-for-row.**

| leg | mutation | prediction | observed |
|:--|:--|:--|:--|
| **A** | drop the mark from the door's `extra` (the pre-fix behaviour) | partial red, presence rows only: 9 named rows, and specifically ⛔ NOT the unmarked-envelope row, ⛔ NOT any §4 omission row, ⛔ NOT the log-line row | **9 failed / 28 passed (37)** — the failing set is exactly the 9 predicted rows, no others |
| **B** | replace `declaredUserMessage(e)` with a naive `(e as any).userMessage` | exactly the 5 §4 non-declaration rows (`''`, whitespace, `42`, `null`, an object), nothing else | **5 failed / 32 passed (37)** — exactly those 5 |

Both legs: anchored replacement asserting **exactly one** hit before writing; mutation proven on
disk by removed-text count `1 ⇒ 0` **and** blob-hash inequality against the `HEAD` blob (for leg A
the injected token `undefined,` occurs 23 times and is *not* a discriminating count — the removal
count and the hash are what prove it); restore by `git checkout HEAD -- ABSOLUTE_PATH`, never bare,
proven by blob-hash **equality** plus an empty `git diff HEAD` plus a clean `git status --porcelain`;
absolute paths throughout and `trap … EXIT INT TERM` on both legs.

**No rebuild was needed or claimed, and the runs prove it rather than assert it.** Both suites
reach `http-dispatcher.ts` through same-package relative specifiers, which vitest resolves to
`src/`. A suite reading `dist/` would have stayed **green** under both mutations; every red above
is positive evidence the mutation reached the code under test.

**The hidden test layer — measured, because `@objectstack/runtime` sits at `surplus: none`.**
That package's `typecheck` excludes `**/*.test.ts`, so its green says nothing about a new test
file. Replicating the gate's own re-measure project (the package tsconfig with the test globs
removed from `exclude`) reports **206 errors — exactly the ledger's recorded number for
`@objectstack/runtime`**, which is the positive control that this is the same measurement; and
of those 206, **0 are in either test file this PR touches and 0 in `http-dispatcher.ts`**.
⛔ No TEST_DEBT entry is raised, lowered or edited.

**Gates.** Union derived with `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack`
at `4ecc0d47f7` (no paths passed; stderr read — it confirms the repo assertion holds against this
checkout's `origin`, and the 4-path change set from the merge base `eb717a12a`), harvested with
`--commands` so neither invocation spelling can be dropped. Every exit code captured **before**
any pipe.

⭐ **Reconciliation, by `comm -23` over the two sorted lists (the harvested union and the commands actually run), exact string comparison, no prefix or fuzzy matching and no self-maintained counter:**

```
union named 34, ran 34, unreconciled 0
```

**31 of 34 pass**, including the ones nearest this diff: `check:cross-package-test-inputs`,
`check:test-source-alias`, `check:engine-double-contract`, `check:where-matcher`,
`check:query-options-erasure`, `check:type-check-coverage`, `check:dispatcher-error-vocabulary`,
`check:undeclared-dep-imports`, `check:comment-mask-adoption` and the whole changeset family.

**Run beyond the union** (unconditional CI steps and one content-judged gate no path derivation
can name):

- `pnpm lint` — whole repo, `eslint . --no-inline-config`: **exit 0** in 76s, run in full.
  ⛔ No narrowing claimed and none needed.
- `node scripts/check-nul-bytes.mjs`: OK, 7588 text files, no raw control bytes. Plus a direct
  `grep -naP` control-byte self-scan over all four changed files: no hits.
- `pnpm check:error-code-casing`: **exit 0** — the gate that went red on PR #13619. This diff's
  only code literals are `PERMISSION_DENIED`, already registered, so no `adr0112-ok:` declaration
  was needed or added.

**3 refusals, reported separately and ⛔ NOT folded into the passes — each is NOT MEASURED, in the
gate's own words:**

- `node scripts/check-test-completeness.mjs` — exit **3**: *"the local reading for this gate is
  NOT MEASURED. ⛔ It is not a red, and there is nothing here to fix."* It needs a saved
  `turbo run test` log that only CI produces.
- `pnpm check:dual-build-cjs-loads` — exit **3**: *"PREREQUISITE NOT MET — this gate reads built
  output, and some package has no dist/. … ⛔ This is NOT a pass: nothing was measured."*
- `pnpm check:type-check-debt` — exit **1**. Its structural half ran; the `--re-measure` ratchet
  refused: *"cannot run: 23 workspace dependenc(ies) of the ledgered packages have no built type
  entry point on disk"*, because a number taken without the closure *"would silently measure a
  DIFFERENT WORLD"*. ⚠️ Declared narrowing: the whole-workspace closure build this gate wants was
  in progress in a sibling agent's hold on the shared verify lock, and queueing behind it plus the
  build itself does not fit this container's foreground ceiling. The substance is measured
  locally instead — see the hidden-test-layer paragraph above, whose total reproduces the ledger's
  206 exactly — and CI runs the gate with the closure built either way.

## The changeset grade, with reasoning rather than a silent pick

**`@objectstack/runtime: patch`.** Nearest precedent: #13241 / PR #13619, the same field at the
sibling door, also `patch`. The reasoning, not just the precedent: no package's public API or
type surface changes (measured above — identical `.d.ts`), no field is declared (the slot has
existed since #9934), and nothing that parsed before stops parsing. What changes is that one
door stops dropping a declared optional field — a bug fix closing `declared ≠ enforced`, which
is `patch`. ⛔ Not `minor`: no new capability arrives; three of the four boundaries already
carried the field.

## Findings recorded on the way past

⛔ None filed. The one thing worth recording is a documentation-grade correction rather than a
defect, and it is stated in full above so the next author does not pay for it: **#7898's own
re-check command returns a false zero through a shell**, because `git grep` reads it as a regex.
Any future discharge of that hold should run it with `-F`, and should keep a positive control
beside it — without one, that zero reads as "the premise sentence is gone", which is the exact
opposite of what the tree says.

---

### DIFF
--- a/.changeset/denial-door-carries-user-message.md
+++ b/.changeset/denial-door-carries-user-message.md
@@ -1,0 +1,45 @@
+---
+'@objectstack/runtime': patch
+---
+
+fix(runtime): carry the producer's `userMessage` at the dispatcher's PERMISSION_DENIED door (#13623)
+
+`HttpDispatcher.dispatch`'s foot catch is not a pure rethrow: it recognises
+`isPermissionDeniedError` — `name === 'PermissionDeniedError'`, **or**
+`code === 'PERMISSION_DENIED'`, **or** a message starting `[Security] Access
+denied` — and answers the refusal itself from
+`packages/runtime/src/http-dispatcher.ts`. A marked denial therefore never
+reached `dispatcher-plugin`'s throw-transparent exit, which is the door #13241
+taught to carry the author-facing text channel, and lost the mark at this one
+instead. `ApiErrorSchema.userMessage` has declared the slot all along and
+`contract.zod.ts` states the invariant directly — *"`userMessage` on the thrown
+error; the boundaries carry it to the wire"* — so this is `declared ≠ enforced`
+at one more boundary, not a new field.
+
+**Why this door matters more than its size.** #9934 made the mark
+status-agnostic precisely so a 403 could carry it, and a 403 is the refusal
+class most likely to carry deliberately-authored text ("You do not have access
+to this report; ask an admin for the Reporting role"). The one door that
+swallowed the mark was the one answering exactly those refusals.
+
+**Transport parity, which the same ruling already asked for.**
+`@objectstack/rest`'s `mapDataError` has carried the field on this identical
+denial since #9934 (`withDeclaredUserMessage` over `classifyDataError`, whose
+`PERMISSION_DENIED` branch is pinned in
+`packages/rest/src/rest-user-facing-refusal-marking.test.ts`). The 2026-08-11
+ruling on #7450 makes REST's shape the contract for both transports; until now
+the dispatcher's 403 was the one that differed.
+
+**⛔ #7450's disclosure withhold is untouched.** That ruling is about
+`error.details` — the gate's `positions` / `permissionSets` and the cascade
+child `object` the caller never addressed — and all of it is still dropped from
+the wire and still logged server-side. `userMessage` is the opposite by
+construction: it exists only because a producer wrote text *for* the caller,
+platform and driver code never set it, and it rides as a declared top-level
+sibling of `code`/`message`, never inside `details`. The read is
+`declaredUserMessage` (`@objectstack/types`), the one rule every boundary
+applies, so this door cannot fork its own answer to "what counts as marked".
+
+**No behaviour change for an unmarked denial.** Its envelope is byte-identical
+to before — key set `code` / `details` / `httpStatus` / `message`, pinned — and
+the mark never moves the status or the `code`.
--- a/packages/runtime/src/dispatcher-plugin.declared-user-message.test.ts
+++ b/packages/runtime/src/dispatcher-plugin.declared-user-message.test.ts
@@ -19,6 +19,14 @@
  * — whose result already carries `userMessage` — and then did not read the
  * field. `declared ≠ enforced` at exactly one door.
  *
+ * ⭐ [#13623] There was a SECOND door, and this file is where it was recorded:
+ * `HttpDispatcher.dispatch`'s foot catch answers `isPermissionDeniedError`
+ * itself, so a marked `PERMISSION_DENIED` 403 never reaches the exit above and
+ * lost its mark there instead. §2's carve-out row pinned that absence as an
+ * observation; #13623 closed the door and MOVED the row to assert presence —
+ * see its docblock for why moving it (rather than fixing it green) is the
+ * deliverable.
+ *
  * ## Why this is the compensation channel and not a decoration
  *
  * The 2026-08-27 ruling on #12509 (option D), propagated to #12281, made this
@@ -225,7 +233,10 @@ describe('[#13241] the dispatcher throw-transparent exit carries `userMessage`',
             { status: 400, code: 'VALIDATION_FAILED', prose: 'measure revenue not additive over stage', mark: 'Pick a measure that adds up across stages.' },
             // ⚠️ `FORBIDDEN`, deliberately NOT `PERMISSION_DENIED` — see the
             // carve-out pinned below this table: a throw spelling
-            // `PERMISSION_DENIED` never arrives at this exit at all.
+            // `PERMISSION_DENIED` never arrives at this exit at all. It is
+            // answered by `HttpDispatcher.dispatch`'s foot catch, which since
+            // #13623 carries the mark itself; this row keeps a status the foot
+            // catch does NOT intercept, so it still measures THIS exit's 403.
             { status: 403, code: 'FORBIDDEN', prose: 'principal lacks cube_read on pipeline', mark: 'You do not have access to this report. Ask an admin for the Reporting role.' },
             { status: 409, code: 'RECORD_LOCKED', prose: 'cube pipeline locked by in-flight rebuild', mark: 'This report is being rebuilt. Try again shortly.' },
             { status: 503, code: 'SERVICE_UNAVAILABLE', prose: 'warehouse pool exhausted', mark: 'Reporting is briefly offline.' },
@@ -247,25 +258,37 @@ describe('[#13241] the dispatcher throw-transparent exit carries `userMessage`',
         }
 
         /**
-         * ⚠️ MEASURED CARVE-OUT — the one 403 that is NOT throw-transparent, and
-         * therefore NOT closed by this change.
+         * ⭐ [#13623] THE CARVE-OUT MOVED — and this row is how a reader finds out.
          *
-         * `HttpDispatcher.dispatch`'s foot catch is not a pure rethrow: it
-         * intercepts `isPermissionDeniedError` — `name === 'PermissionDeniedError'`
-         * **or** `code === 'PERMISSION_DENIED'` **or** a message starting
-         * `[Security] Access denied` — and answers it itself, via `this.error(…)`
-         * in `packages/runtime/src/http-dispatcher.ts`. Such a throw never
-         * reaches `errorResponseBase`, so its mark is dropped by a DIFFERENT
-         * door than the one this card repairs.
+         * It was written as the opposite assertion (`expectNoUserMessageAnywhere`)
+         * with a docblock saying, in terms: *"if a later change makes the denial
+         * path throw-transparent, this test fails and tells the author that the
+         * carve-out has moved"*. #13623 is that change, so the row is MOVED
+         * deliberately rather than repaired green — the instrument worked, and
+         * softening or deleting it would have thrown away the only signal.
          *
-         * ⛔ That door is #7898's on-hold trigger file and is out of this PR's
-         * file surface, so the gap is recorded here rather than closed. Pinned
-         * as an observation and not as a wish: if a later change makes the
-         * denial path throw-transparent, this test fails and tells the author
-         * that the carve-out has moved — which is the only way a reader finds
-         * out that the sentence above went stale.
+         * ⚠️ What moved is the MARK, not the route. The sentence the old
+         * docblock opened with is still true: `HttpDispatcher.dispatch`'s foot
+         * catch is not a pure rethrow — it recognises `isPermissionDeniedError`
+         * (`name === 'PermissionDeniedError'` **or** `code === 'PERMISSION_DENIED'`
+         * **or** a message starting `[Security] Access denied`) and answers it
+         * itself via `this.error(…)` in
+         * `packages/runtime/src/http-dispatcher.ts`. Such a throw still never
+         * reaches `errorResponseBase`. What changed is that the OTHER door now
+         * reads the same `declaredUserMessage` rule, so the mark survives both.
+         *
+         * ⇒ The row therefore pins a DIFFERENT door's behaviour from every other
+         * row in this file, and is kept here on purpose: the two doors' answers
+         * to one question belong in one place, and if either regresses the
+         * failure names the door in its own title.
+         *
+         * ⛔ Still NOT a claim that this exit became throw-transparent. A
+         * regression that made it so would silently satisfy this row; the pin
+         * against that is `expect(res.body.error.details)` below — the foot
+         * catch derives `details.object` from the ROUTE (`permissionDeniedErrorDetails`)
+         * and `errorResponseBase` does nothing of the kind.
          */
-        it('CARVE-OUT — a `PERMISSION_DENIED` 403 does not reach this exit, so its mark is still dropped', async () => {
+        it('MOVED CARVE-OUT — a `PERMISSION_DENIED` 403 is answered by the foot catch, and now keeps its mark there', async () => {
             const res = await throwFromAnalyticsQuery(
                 declaring(
                     { status: 403, code: 'PERMISSION_DENIED', userMessage: 'Ask an admin for the Reporting role.' },
@@ -275,8 +298,33 @@ describe('[#13241] the dispatcher throw-transparent exit carries `userMessage`',
 
             expect(res.statusCode).toBe(403);
             expect(res.body.error.code).toBe('PERMISSION_DENIED');
-            // Not a claim that dropping it is CORRECT — only that it is where
-            // the field is dropped today, and that it is not this exit.
+            // The half this card repaired: the author's channel now survives
+            // the denial door too (#9934 is status-agnostic, and 403 is the
+            // refusal class most likely to carry authored text).
+            expect(res.body.error.userMessage).toBe('Ask an admin for the Reporting role.');
+            // …verbatim, and never in place of the diagnostic channel.
+            expect(JSON.stringify(res.body)).toContain('Ask an admin for the Reporting role.');
+            expect(res.body.error.message).not.toBe('Ask an admin for the Reporting role.');
+            // Still the foot catch answering, NOT this exit — the route-derived
+            // `details` shape only that door produces. `/analytics/query` names
+            // no object, so `permissionDeniedErrorDetails` contributes nothing
+            // but the promoted `code`, and `details` is absent entirely.
+            expect(res.body.error.details).toBeUndefined();
+            // ⛔ #7450's withhold is untouched: the mark is a top-level sibling,
+            // never `details` context.
+            expect((res.body.error.details as any)?.userMessage).toBeUndefined();
+        });
+
+        it('POSITIVE CONTROL (the denial door) — an UNMARKED `PERMISSION_DENIED` 403 still has no key', async () => {
+            // Without this, the row above could pass on a door that always
+            // reports a `userMessage`. Same route, same code, same status — the
+            // ONLY difference is the mark.
+            const res = await throwFromAnalyticsQuery(
+                declaring({ status: 403, code: 'PERMISSION_DENIED' }, 'principal lacks cube_read on pipeline'),
+            );
+
+            expect(res.statusCode).toBe(403);
+            expect(res.body.error.code).toBe('PERMISSION_DENIED');
             expectNoUserMessageAnywhere(res);
         });
 
--- a/packages/runtime/src/http-dispatcher.permission-denied-user-message.test.ts
+++ b/packages/runtime/src/http-dispatcher.permission-denied-user-message.test.ts
@@ -1,0 +1,384 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * [#13623] The DENIAL door carries the producer's `userMessage` — the second
+ * door that dropped it, and the one whose refusals users most need to read.
+ *
+ * ## Why this is a second door and not the one #13241 repaired
+ *
+ * `HttpDispatcher.dispatch`'s foot catch is **not a pure rethrow**. It
+ * recognises `isPermissionDeniedError` — `name === 'PermissionDeniedError'`
+ * **or** `code === 'PERMISSION_DENIED'` **or** a message starting
+ * `[Security] Access denied` — and answers it itself, from
+ * `packages/runtime/src/http-dispatcher.ts`. Such a throw therefore never
+ * reaches `dispatcher-plugin`'s `errorResponseBase`, which is the exit #13241
+ * taught to carry the mark. Same field, same contract, different door.
+ *
+ * `ApiErrorSchema.userMessage` has declared the slot all along, and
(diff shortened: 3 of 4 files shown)


os-steve:
## At-tier contract review — VERDICT: CLAUSE ② DOES NOT APPLY (and, reviewed at tier anyway, the change passes on substance)

I am the at-tier contract reviewer for this PR's `Clause-②: yes` declaration. Standing verified first, by symbol and not by line number: `CONTRACT_REVIEW_TIER = 'claude-fable-5'` (`scripts/pm/dispatch-gates.mjs`, the exported constant under the clause-② docblock). This session runs `claude-fable-5` — at tier. I did not write this code, I am not its PM, and nothing below is taken from the PR body without re-measurement. I read the 16:43Z fable-subagent review on #13623 only after taking my own readings; where we concur it is concurrence, not adoption.

I am not merging, not arming auto-merge, not flipping draft, not editing code, and not touching labels.

### 1. The central question: clause ② does not apply here

The criterion (SKILL.md claim-line spelling, and the clause-② docblock above `MANDATORY_TIER_GLOBS`): does the card **change contract accept/reject behaviour or widen the public surface**? Both limbs measured no.

**Accept/reject — no.** The diff's only behavioural delta is one conditional `extra` bag at the foot catch. Nothing that parsed stops parsing: `ApiErrorSchema.userMessage` has been a declared optional since #9934, so envelopes with and without the key were both valid before this PR. Request-level: the same population is denied, same 403, same `PERMISSION_DENIED`, same `message`; §5 pins that the mark cannot move status or code even when the throw declares `status: 200`.

**Widening — no, because the surface was already this wide.** `packages/spec/src/api/contract.zod.ts:87` declares the field, and line 68 states the invariant this PR enforces, verbatim: *"`userMessage` on the thrown error; the boundaries carry it to the wire."* REST has emitted the field on the byte-identical denial since #9934 (`packages/rest/src/rest-user-facing-refusal-marking.test.ts:82` — "a hook throwing the catalog PERMISSION_DENIED"), and the 2026-08-11 ruling on #7450 makes REST's shape the contract for both transports. A door failing to carry a declared field was in breach of the declared surface — the issue is typed Bug, `declared ≠ enforced` — and bringing it into conformance does not widen anything. The widening happened, and was judged, at #9934. Corroborating measurements: zero files under `packages/spec/**` in the diff (I read the file list, not the body); and the author's two-build `.d.ts` measurement is structurally forced — the diff touches one value import and one private method body, nothing that can reach a type declaration, and the differing `.js` was the right positive control.

**Why the `yes` was still the right declaration.** The maintainer question (comment 5476891660 on #12297, limb 2: was #13619 cleared on its own merits because populating a declared field is the lesser act?) is unanswered, two careful seats disagreed, and a below-tier seat picking the permissive reading is an escape. Declaring `yes` and parking in draft is the honest over-declaration in the direction that buys more review — this review is that compensation working as designed. My verdict answers the applicability question for this PR; the class-level ruling remains the maintainer's. Consistent with PR #13619's own body, I drew no inference from its 07:55Z merge — it was weighed only as evidence of the lane's disposition.

### 2. The change is what it claims to be

`declaredUserMessage` (`packages/types/src/thrown-http-error.ts:270-273`) answers a string only for a non-empty, non-whitespace string `.userMessage`; §4's five rejection rows are exactly its rejection set, driven through the real door. No platform or security code writes the field: `PermissionDeniedError` (`packages/plugins/plugin-security/src/errors.ts`) carries `message`/`details`/`developerMessage` and no `userMessage`; the only in-repo writers are producer channels (the sandbox trigger relay of app-author text, and the boundaries that read it). So the door carries only a mark a producer deliberately set — nothing previously withheld by design starts leaking. The mark lands as a top-level sibling via `buildApiError`'s `...(input.extra ?? {})` spread (`packages/runtime/src/error-envelope.ts`), never inside `details`, and §2 pins the unmarked envelope's key set byte-identical.

### 3. Disclosure — #7450 intact

`permissionDeniedErrorDetails(cleanPath)` derives strictly from the route; `e.details` (positions, permission sets, cascade-child object) never reaches the wire and the `console.warn` server-side log is retained in the diff. §3 drives a fully-populated cascade-child fixture **with a mark present** and asserts `positions`, `permissionSets`, `org_member`, `app_reader`, `app_child_object`, `developerMessage` all absent from the serialized body — the with-mark case is the one that matters, and it is the one tested.

### 4. The moved carve-out — moved deliberately, nothing weakened

Compared base `fd539b50d0` against the head file line by line. The old row asserted absence with the two-limbed guard; the new row asserts presence (verbatim, and `message` not replaced), is retitled MOVED CARVE-OUT, records that the exit did **not** become throw-transparent, and a new positive control on the same route preserves the absence guard for the unmarked case. No assertion deleted, skipped, or softened; the FORBIDDEN-row comment was updated to stay true. The ablation arithmetic is reproducible from the diff itself: exactly 9 presence rows and exactly 5 §4 non-declaration rows across the 37 tests in the two suites, matching both legs' predicted/observed sets.

One docblock inaccuracy, recorded and non-blocking: the new docblock claims `expect(res.body.error.details)` pins the door identity. On `/analytics/query` it does not — the route names no object, so the foot catch's details is `{code}` promoted away to absent, and a hypothetically throw-transparent `errorResponseBase` answer for the same fixture (no non-string code, no validation shape, mark carried post-#13619) would be indistinguishable on the wire. The door identity is genuinely pinned, but elsewhere: the new suite's §3 log-line test (only the foot catch emits `[HttpDispatcher] PERMISSION_DENIED on …`) and §1's route-derived `details: { object }` on `/data/...` (which `errorResponseBase` never builds). Worth a one-line docblock correction in a future pass; not a weakening of any assertion.

### 5. Changeset grade — correct

`@objectstack/runtime: patch` is right: no new capability, no API or type surface change, one door stops dropping a declared field — a conformance bug fix, matching the sibling precedents that landed as plain `patch`. No `content/docs/releases/` edit rides the PR, as required.

### Findings recorded, none blocking

- **Bound parity gap (real, pre-existing class):** the new door ships the mark unbounded via plain `declaredUserMessage(e)`, while REST bounds every mark at `CLIENT_MESSAGE_MAX = 500` via `boundedDeclaredUserMessage` (`packages/rest/src/error-response.ts:652`, `truncateClientMessage:107`). The dispatcher's sibling exit is equally unbounded (`http-dispatcher.ts:850` via `resolveThrownHttpError`), so the package is internally consistent — but §6's "discloses exactly what @objectstack/rest discloses" holds only below the bound. Concurs with the 16:43Z review's finding, re-verified at the source; it deserves the follow-up card already promised, not rework here.
- **Citation nit:** "the maintainer question is still open on #12297" is accurate but easy to misread — #12297's own subject is an unrelated, now-closed `os lint` card; the question lives in its comment thread (5476891660).

### Bottom line

Clause ② does not apply: neither limb is met, on my own measurements. The conservative `yes` was honest and procedurally correct, and this review discharges it. Having reviewed at tier anyway: the change is exactly what it claims — a producer-authored mark carried through one more boundary the contract already governed, with #7450's withhold pinned intact and the instrument moved rather than silenced. On substance this PR earns APPROVE; green CI played no part in that judgement.

---

Issue #13718, opened 2026-08-31
Give a lowered hook body a declared constant surface — every shared constant is currently hand-copied into each handler and pinned by a bespoke drift test (8 rows in the reference app)


os-warren:
Split out by the triage seat from **#13651** (ask 2 of 2). That card's two asks are independent — its own words: *"They are independent."* Ask 1 (make the silent lowering failure loud) is a `packages/cli` repair and is now queued on #13651 alone. **This card carries ask 2**, which is a contract-surface expansion and therefore a maintainer decision, not a queue item.

Restart-when: #13651 closes (ask 1 lands and reports the count of constants an app cannot reach)

⛔ Filed unassigned. The measurement and the evidence table below are quoted from #13651; nothing here is a new reading.

## The gap

An L2 hook handler is lowered to a metadata-only `body.source` and evaluated inside QuickJS with **no module scope**. A handler body referencing an import, a top-level `const`, or even the factory function's own parameter is a `ReferenceError` at runtime.

That boundary is deliberate and #13651 explicitly does **not** ask to cross it:

> Reaching module scope at runtime. The body-only sandbox is a deliberate boundary and the app is not trying to cross it — the ask is that the boundary be enforced at author time instead of discovered at review time, and that a legitimate constant not have to be smuggled across it by hand.

What is missing is a **declared** way to ship a constant *with* the body.

## What the absence costs, measured in the reference app

Every shared value is hand-copied into every handler body, and each copy carries a bespoke test whose only job is to notice drift:

| inlined copy | its pin |
| --- | --- |
| the priority-rank map, twice (`case.hook.ts`, `task.hook.ts`) | `test/priority-rank-parity.test.ts` |
| the 16-cell SLA matrix (`case.hook.ts`, duplicating `_case-sla.ts`) | `test/case-sla-matrix.test.ts` |
| the country-to-territory table (`account.hook.ts`, duplicating `_territory.ts`) | `test/territory-single-source.test.ts` — parses the table out of the LOWERED body |
| the `refuse()` refusal envelope, in every guard (duplicating `_refusal.ts`) | `test/refusal-envelope.test.ts` |
| the campaign metric recompute, four times | `test/campaign-member-lifecycle.test.ts` — asserts all four are character-identical |
| the position-pool names, twice (`_case-assignment.ts`) | `test/case-assignment.test.ts` |
| the claimable-status set (`_case-assignment.ts`) | `test/unassigned-case-triage-reach.test.ts` |
[...]

Issue #10678, opened 2026-08-21
Hook-body build gates: @capabilities override dead through os build; require() pattern dead; default build silently bundles forbidden patterns


baozhoutao:
Found by QA run #10663 driving `cli.hook-body-extraction-gates` at framework `79ebb37`. The security net is intact throughout — no forbidden body ever shipped as `body.source`, and every forbidden/free-identifier hook was still refused under `--strict-body`. Three defects in the *reporting/reachability*, not the enforcement:

## 1. `// @capabilities` override is unreachable through `os build` (product/docs gap)
A handler authored in `objectstack.config.ts` with a first line `// @capabilities api.read api.write` produces artifact capabilities `[]`. The config loader (`config.ts:loadConfig` → `bundle-require` → esbuild `transformSync{loader:'ts'}`) **strips `//` line comments** before `String(fn)` runs, so the override at `extract-hook-body.ts:118-131` — documented as working at `hook-bodies.mdx:320-327` — never receives the comment. The pin `lower-callables.test.ts` masks it by feeding raw JS function literals that keep their comment, never exercising the esbuild load path. Fix: move the directive off a `//` comment (e.g. a real property), or state in docs that it is authorable only in pre-bundled JS; add an `os build`-level test.

## 2. `require()` FORBIDDEN pattern is effectively dead (assertion/coverage gap)
`require('node:os')` in a TS config is rewritten by esbuild's ESM shim to `__require("node:os")`, so `FORBIDDEN_PATTERNS[/\brequire\s*\(/]` (`extract-hook-body.ts:35`) never matches. The refusal still fires (exit 1) — via the #1876 free-identifier gate naming `__require` — so enforcement holds, but the `require()`-specific worded reason the acceptance promises is never emitted for the real authoring path.

## 3. Default `os build` silently warn-and-bundles forbidden patterns (header contradiction)
[...]

PR #13834, opened 2026-08-31
feat(cli): make the silent hook-body downgrade loud — `os lint` refuses an accidental scope leak (ask 1)

### LINKED ISSUE #13651: Hook-body lowering fails OPEN: `tryExtractBody` swallows `extractHookBody`'s refusal into a warnings array and bundles the closure — make it loud (ask 1; the constant-surface half is split out)
Filed from the reference app while slimming `src/objects/` comments (objectstack-ai/hotcrm#1184). This constraint is written out, by hand, in **twelve** files of that one directory. It is the single most-repeated platform note in the app, which is the signal that it wants to be a platform affordance rather than prose.

## The constraint, as app authors have to state it

An L2 hook handler is lowered to a metadata-only `body.source` and evaluated inside QuickJS with **no module scope**. A handler body that references an import, a top-level `const`, or even the factory function's own parameter is a `ReferenceError` at runtime — not a closure.

That much is documented and defensible. The part that costs is what happens when an author gets it wrong:

- `extractHookBody` **throws**, naming the free identifiers:
  ```
  [hook-body-extract] hook 'x': handler references identifier(s) not in
  scope at runtime: refuse. Module-scope helpers/imports aren't shipped
  with a metadata-only body, so this handler will be BUNDLED instead …
  ```
- …and the CLI build **catches that throw and silently bundles the closure instead**. Nothing goes red. The hook keeps working locally. It has simply stopped being shippable as pure metadata, and the build says so in one line nobody reads.

So the failure is invisible at exactly the moment it is cheapest to fix.

## What the app pays, today

Because the degradation is silent, the reference app cannot rely on the build to catch it, and instead hand-copies every shared value into every handler body and then writes a bespoke test to pin each copy:

| inlined copy | its pin |
| --- | --- |
| the priority-rank map, twice (`case.hook.ts`, `task.hook.ts`) | `test/priority-rank-parity.test.ts` |
| the 16-cell SLA matrix (`case.hook.ts`, duplicating `_case-sla.ts`) | `test/case-sla-matrix.test.ts` |
| the country-to-territory table (`account.hook.ts`, duplicating `_territory.ts`) | `test/territory-single-source.test.ts` — parses the table out of the LOWERED body |
| the `refuse()` refusal envelope, in every guard (duplicating `_refusal.ts`) | `test/refusal-envelope.test.ts` |
| the campaign metric recompute, four times | `test/campaign-member-lifecycle.test.ts` — asserts all four are character-identical |
| the position-pool names, twice (`_case-assignment.ts`) | `test/case-assignment.test.ts` |
| the claimable-status set (`_case-assignment.ts`) | `test/unassigned-case-triage-reach.test.ts` |
| the reference-cleanup shape predicate, three times | `test/freeze-guard-reference-cleanup.test.ts` |

Every row is the same shape: **a constant declared once, stored N times, and a test whose only job is to notice when the copies drift.** That is a platform gap being paid for in app-side test infrastructure.

And the app has to *say so* every time, because the alternative reads as gratuitous duplication to the next author. The comment mass is a direct function of the missing affordance.

## What would make the mistake impossible

Two asks, either of which retires most of the prose. They are independent.

1. **Fail loudly instead of degrading.** The build should refuse — or at minimum emit a `lint` diagnostic that a gate can fail on — when a registered hook's handler cannot be lowered. Silently swapping a metadata hook for a bundled closure changes the deployment shape of the app; that is not a warning-in-a-log event. (`extractHookBody` already computes the exact answer; only the catch site drops it.)

2. **Give a hook body a declared constant surface.** Anything that lets a handler name a value the platform ships into the sandbox with it — a `constants:` key on the hook whose object is serialised into the lowered body, or an author-time inliner that resolves a whitelisted module-scope `const` at lowering time — collapses every row of the table above into one declaration and deletes its pin.

Ask 1 alone turns a silent class into a compile-time one, which is the bigger win. Ask 2 removes the duplication that ask 1 makes visible.

## Not asking for

Reaching module scope at runtime. The body-only sandbox is a deliberate boundary and the app is not trying to cross it — the ask is that the boundary be enforced at author time instead of discovered at review time, and that a legitimate constant not have to be smuggled across it by hand.

Refs objectstack-ai/hotcrm#1184.

os-steve:
Fixes #13651

Scope is **ask 1 only** (fail loudly). The constant-surface half is #13718 and is not addressed here — no half of it is stubbed in, out of scope: #13718 remains open.

## Clause-②: yes

This changes what **`os lint`** rejects: a config whose hook handler leaks module scope now produces a lint `error`, so `os lint` exits 1 where it previously exited 0.

It does **not** change what `os build` accepts. That was a deliberate choice, not an omission — see the measurement below. The catch in `lowerCallables` stays, both refusal classes still fall back to bundling, and the build still exits 0; the three `e2e` suites that spawn a real `os build` pass unchanged.

PR stays draft. Not armed, not queued, not flipped ready.

## The measured population — the reading #13718 needs to size itself

Measured on the **real build path**, not by re-implementing it: `loadConfig` (bundle-require/esbuild, exactly what `os build` uses) then `normalizeStackInput` then `lowerCallables`, reading the reason strings the real `extractHookBody` produced.

| config | callables lowered | body extracted | on the silent-downgrade path |
| --- | ---: | ---: | ---: |
| `objectstack-ai/hotcrm` (the reference app, `aff9cb2f`) | 39 | 39 | **0** |
| `examples/app-showcase` | 2 | 0 | **0** |
| `examples/app-todo` | 2 | 1 | **0** |
| `examples/app-crm` | 1 | 1 | **0** |
| `plugin-auth`, `plugin-security`, `service-i18n` | 0 | 0 | **0** |
| `create-objectstack` blank template | 0 | 0 | **0** (declares no hooks or functions) |
| **total** | **44** | **41** | **0** |

**The existing population on the silent-downgrade path is zero.** Nothing in this repo or in the reference app currently takes it.

Two things that number does *not* mean:

1. It is not evidence the gap is cheap. The reference app reads zero **because it already paid** — all 39 of its callables are written defensively, with constants hand-copied into each handler and eight bespoke pin tests whose only job is to notice when the copies drift. That is the cost #13651 documents, and it is what #13718 is sizing.
2. It is not a licence to break `os build`. Zero here says nothing about apps outside this tree, and `os build` is a published CLI.

The 3 callables that produced no body **and no warning** are not this class: they are top-level `functions:` map entries (`app-showcase` 2, `app-todo` 1), a path that never attempts extraction by design.

**Positive control** (a zero-hit is not a reading without one). The same instrument over a fixture with three known-outcome hooks reported exactly: `hook_free` to `free-identifiers`, `hook_forbidden` to `forbidden-token`, `hook_clean` extracted. The instrument yields positives in both classes, so its zeros are real.

## Which gate shape, and why

The card offered two. **Chosen: the `lint` diagnostic a gate can fail on** — the card's own steadier intermediate — made class-aware.

Against making `os build` fail outright: the population is zero, so that option reddens nothing *today*, but it changes the accept set of a published CLI for apps that cannot be measured from here, and it would take the legitimate fallback down with the accidental case. `compile.ts` step 2c already records that position in-tree.

Why `os lint`'s own rubric rather than the shared authoring-rule registry: the registry's `gating` tier must be run by all three commands (so it *would* move the build's accept set), and its `advisory` tier can never emit an `error`. `lintConfig` is the only tier whose `error` fails `os lint` alone — its own comment calls it "a lint verdict, not a publish gate".

## Telling an author's mistake from a deliberate bundle

The refusal already knew which rule refused; the throw flattened it into prose and the catch kept only the sentence. It now carries `kind`:

- **accidental** (`free-identifiers`) — the handler *is* expressible as a metadata body; it merely names a module-scope const, helper or import. The deployment shape changed against what the author wrote. Gives a **`error`**, rule `hook-body/not-lowerable`.
- **structural** (`forbidden-token`) — `fetch`/`require`/`process`/`eval` are capabilities the sandbox does not have, so writing one *is* choosing a bundled closure and the bundle is the designed answer. Gives a **`warning`**, rule `hook-body/bundled-fallback`.
- **`unparseable`** gives a warning. **`unknown`** (a non-`HookBodyExtractionError` throw) is kept distinct from `unparseable` on purpose: an instrument failure must not read as a verdict about the author.

An author who deliberately wants a bundled closure keeps two channels that already existed, are already honoured, and stay silent — neither needs a new spec key: give the hook an explicit `body`, or move the function into the top-level `functions:` map and reference it by name. So the inline-function form means "I intend this to be a hook body" and the named-`functions:` form means "I intend this to be bundled code" — a distinction the authoring surface already had and nothing was reading.

**Parity by construction:** the rule calls the *same* `extractHookBody` the build calls, so the lint verdict cannot drift from what the build would do to the same handler. Pinned by a test that asserts the two see exactly the same callables.

## Ablation

Direction predicted in writing before running: mutating the `kind === 'free-identifiers'` branch so the rule stops consulting the classification should turn the accidental class into a warning, reddening the error-severity and parity pins, while the refusal-kind suite stays green.

Observed: mutation proven on disk (blob `b47714f4` to `f71b0286`, anchor count 1 to 0, injected 0 to 1); **3 tests failed** in `hook-body-lowering.test.ts` — the two predicted plus the action-path pin, which is the same accidental-class assertion applied to actions, so wider in extent than predicted, same direction; `hook-body-refusal-kind.test.ts` stayed green as predicted. Restore proven three ways: blob hash equal to the HEAD blob, empty `git diff HEAD`, clean `git status --porcelain`.

No rebuild leg was needed, and the ablation itself is the proof: mutating `src/` changed the outcome with **no** build, so these tests read source, not `dist`.

## Verification

Gate union derived with `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack --commands` after merging `origin/main` and after the last edit, harvested with `--commands` (never grepped), each exit code captured before any pipe.

```
union named 36, ran 36, unreconciled 0
```

`comm -23` and `comm -13` are both empty (exact comparison, no arithmetic on a hand-kept counter). 35 path-derived plus convention-triggered families, plus `pnpm lint`, which the derivation never names.

- **`pnpm lint`** (`eslint . --no-inline-config`): run **repo-wide**, exit 0. No narrowing to declare.
- **`pnpm --filter @objectstack/cli typecheck`**: exit 0.
- **Tests**: 12 files, 171 tests passed, including the three `e2e` suites that spawn a real `os build`.
- All runs above are on `59e77df0d7`.

Two families exited non-zero during the sweep; both are resolved:

- `node scripts/check-test-completeness.mjs` — exit 3, **NOT MEASURED**, in the gate's own words: *"PREREQUISITE NOT MET — this gate grades a saved `turbo run test` log, and no log was named. … the local reading for this gate is NOT MEASURED. It is not a red, and there is nothing here to fix."*
- `pnpm check:type-check-debt` — exit 1, a **real red that is fixed at source**. `TEST_DEBT['@objectstack/cli']` recorded 144, `tsc --noEmit` reported 147 (+3). `packages/cli/tsconfig.json` says `include: ["src"]`, so the two new test files sit *inside* the package program and carried 3 TS7053. Attributed mechanically: removing exactly those two files returned the ledger to green with no other entry moving, so the +3 was entirely theirs. Fixed by typing the two fixture maps. The entry was **not** raised — shrink-only and maintainer-only by the gate's own text. Re-run on the merged head: exit 0, *"29 ledger entr(ies) re-measured … none above its recorded number."*

## Changeset

`@objectstack/cli`: **minor**. What this publishes rides on two CLI surfaces, not on new API exports — corrected in the contract-review patch round: `HookBodyExtractionError` / `HookBodyRefusalKind` / `BodyExtractionWarning` are module-internal (the package `exports` map exposes only `.` and `./console`, and `src/index.ts` re-exports none of them). The published carriers are the `os lint` exit contract (a new `error` rule, so `os lint` can newly exit 1 where it exited 0) and the `kind` / `freeIdentifiers` fields on `os build --json`'s `bodyExtractionWarnings`. Nothing removed, `os build`'s accept set untouched, so not major. Precedent: the launch-window narrowing-as-`minor` convention stated in `.changeset/form-view-option-default-narrowed.md`; the earlier citation (`cli-i18n-flow-screen-bucket`) is not on point — its diagnostics error only under `--i18n-strict`.

## Patch round — at-tier contract review (2026-09-01, head `f46ff14252`)

Acting on the REQUEST CHANGES review (items 1 and 2; item 3 — the verbatim `Clause-②: yes` line on the #13651 claim comment — is the PM seat's, not addressed here). Both premises re-derived against the tree before editing: `packages/cli/src/index.ts` names none of the three types (grep exit 1, no `export *`), the package `exports` map is `.` and `./console` only; and `judge()`'s catch branched only on `free-identifiers`, so `unparseable`/`unknown` fell into the `bundled-fallback` arm whose message asserts "This is the designed fallback — the body uses something the sandbox cannot provide."

1. **Changeset corrected** (`.changeset/cli-hook-body-lowering-loud.md`): the "new exports" sentence is replaced by the real published carriers — the `os lint` exit contract (0 → 1) and the `kind`/`freeIdentifiers` fields on `os build --json`'s `bodyExtractionWarnings`. Grade unchanged: `minor`.
2. **`judge()` keeps instrument failures distinct from author verdicts**: `unparseable` and `unknown` no longer borrow the `bundled-fallback` prose; they report under their own rules — `hook-body/unparseable` and `hook-body/extraction-failed` — as warnings whose prose says "not a verdict about the handler". Severities unchanged (`warning`), so this patch moves no exit contract. The parity pin's rule→kind map is now **total over all four kinds** and its fixture population includes both instrument kinds, so build and lint are pinned to label them identically; three unit pins additionally forbid the instrument kinds from ever reading as a chosen bundle.
   - Ablation (direction predicted in writing before running): excising the new instrument arm restores the fold and should red exactly 4 pins — the two new unit pins, the distinctness pin, and the extended parity pin — leaving the other 9 in the file green. Observed: exactly those 4 failed, on the rule-identity assertions. Mutation proven on disk (blob `009c3960` → `d0b07871`, unique anchor "the extraction instrument itself failed" count 1 → 0); restore proven (blob hash equal, empty `git diff HEAD`, clean `git status --porcelain`); trap in the same process as the measurement.

Verification on the merged head `f46ff14252`: gate union re-derived post-merge (the pre-merge derivation self-reported STALE TREE), **named 36 (35 derived + `pnpm lint`, which the derivation never names), ran 36, unreconciled 0** — `comm -23` and `comm -13` both empty, exact string comparison, every exit code captured before any pipe. 31 gates green first pass; 4 families (`check:dual-build-cjs-loads`, `check:i18n`, `check:i18n-coverage`, `check:type-check-debt`) first reported PREREQUISITE NOT MET (no built `dist/` — NOT MEASURED, not treated as pass or red), so the closure was built exactly as each gate's text instructs (`turbo run build --filter='./packages/*' --filter='./packages/*/*'`, 70/70) and all 4 re-ran **exit 0, measured** — type-check-debt: "29 ledger entr(ies) re-measured … none above its recorded number" (the new pins add zero debt; the ledger was not touched). `check-test-completeness` exit 3 stays NOT MEASURED in the gate's own words ("this gate grades a saved `turbo run test` log … running the family locally, record this gate as NOT MEASURED"); the in-lieu reading: full `@objectstack/cli` vitest suite on the built closure — 223 files, 2553 tests, 0 failures, exit 0. PR stays draft; labels untouched.

---


os-warren:
## Contract review (Clause ②) — **REWORK**

Reviewed at head `59e77df0d7853ec6bda10ec49a5d85c231b793ed`, still the head. Rendered by a `CONTRACT_REVIEW_TIER` reviewer in an isolated context; transcript tier-verified before adoption (**85 harness-stamped assistant turns, 100% at tier, first and last included, `service_tier: standard` throughout, zero fallback evidence**). The triage seat runs below tier and therefore **adopts verbatim or voids whole**. Adopted verbatim, unedited:

```
VERDICT: REWORK
CLAUSE-2-PATH: no
CLAUSE-2-CONTENT: yes
DECLARATION-HONEST: yes
ONE-LINE: Clause-② content limb is real and honestly declared — os lint's accept set narrows (exit 0→1 on an accidental scope leak) while os build's accept set is provably untouched and no in-repo corpus trips the new error — but the machine spelling `Clause-②: yes` exists only in the PR body, not in the claim comment where the enqueue gate's declaration limb reads it, so the record must be completed (plus two wording fixes) before ready.
FINDINGS:
- Machine declaration missing at the gate's reading site: SKILL.md's claim template makes `Clause-②: yes|no` a mandatory claim-comment field and the enqueue gate's declaration limb reads "认领评论声明 `Clause-②: yes`"; the claim comment on #13651 carries no `Clause-②:` line at all — the verbatim spelling appears only in the PR body, leaving the declaration limb unbound (fail-open, the very pattern this PR fixes). Required fix: add the verbatim line to the claim comment.
- CLAUSE-2-CONTENT yes rests on two facts verified in source: `os lint` exits `errors.length > 0 ? 1 : 0` and the new rule emits severity `error` for kind `free-identifiers`; and `os build --json` publishes `bodyExtractionWarnings` (compile.ts), which now carry `kind`/`freeIdentifiers`. The negative boundary (2026-08-28 ruling) does not apply — this is a published CLI contract surface, not a runtime permission behaviour.
- In-repo corpus verified NOT to trip the new refusal by direct reading, not assumption: app-crm and app-todo handlers reference only `ctx` plus `Date`/`Error`/`String` (all in `detect-free-identifiers.ts` GLOBALS); app-showcase hooks carry explicit `body:` (skipped via `if (!hook.body)`); `functions:` map entries are never extracted. All 35 CI check runs green on head 59e77df0d7. The hotcrm zero is the PR's own positive-controlled measurement — external repo, not independently re-run here.
- Changeset overstates the public-surface widening: "new exports (`HookBodyExtractionError`, `HookBodyRefusalKind`, `BodyExtractionWarning`)" are module-level only — the package `exports` map exposes just `.` and `./console`, and `src/index.ts` on the PR head re-exports none of them; the real published carriers are the lint exit contract and the `--json` field. Grade unaffected, wording should be corrected.
- Nearest precedent mis-cited: `cli-i18n-flow-screen-bucket` (verified real, minor) gains diagnostics that error only under `--i18n-strict`, while this rule errors by default; the closer in-tree precedent is the launch-window breaking-narrowing-as-minor convention stated verbatim in `.changeset/form-view-option-default-narrowed.md`, which the changeset (labelled "Additive") does not name. Minor is correct under either route.
- Lint-surface `unknown` handling contradicts the PR's own claim: `judge()` folds a non-`HookBodyExtractionError` throw into `hook-body/bundled-fallback` with the "This is the designed fallback" message — reading an instrument failure as a verdict about the author, exactly what the PR says is "kept distinct on purpose" (honoured only in `lowerCallables`' warnings); the parity test's rule→kind map also has no row for `unknown`, so build and lint would label such a callable differently. Edge path (extractor bug required), one-line fix.
- Honoured disciplines confirmed: #6479's ruling that a new rejection on a shipped surface must be escalated, not substituted (rejection lands on lint only, declared, drafted, parked for the maintainer's ruling on this third clause-② class); pins falsifiable in both directions (error-on-leak, silence on `selfContainedHook`, structural stays warning-not-error, both deliberate-bundle channels stay silent, fallback-intact pin); the constant-surface half is genuinely absent (#13718 open, `pm:on-hold`, no `constants:` key or inliner in the diff); no `content/docs/releases/` edits.
```

## Carrier action

`needs:contract-review` **stays on both carriers** (this PR and card #13651). ⛔ Not enqueueable until a re-review clears it.

## What to change — three items, all small, none architectural

1. **Add `Clause-②: yes` verbatim to the claim comment on #13651.** That exact hyphenated spelling. It is currently only in this PR body, and the gate's declaration limb does not read PR bodies. ⚠️ On this PR that gap is *load-bearing*: the path limb is **no** — nothing under `packages/spec/src/**` — so the claim comment is the **only** limb that can fire. The gate is holding this PR because a seat hand-routed it, not because the mechanism caught it.
2. **Correct the changeset's "new exports" sentence.** `HookBodyExtractionError` / `HookBodyRefusalKind` / `BodyExtractionWarning` are module-level only; the package `exports` map exposes just `.` and `./console` and `src/index.ts` re-exports none of them. The real published carriers are the `os lint` exit contract and the `--json` field — which is a stronger, more accurate case for `minor`, not a weaker one. While there: the cited precedent `cli-i18n-flow-screen-bucket` errors only under `--i18n-strict`; the on-point one is `.changeset/form-view-option-default-narrowed.md`.
3. **One-line fix in `judge()`:** a non-`HookBodyExtractionError` throw currently folds into `hook-body/bundled-fallback` with the "This is the designed fallback" message — an instrument failure read as a verdict about the author, which is precisely the distinction this PR says it keeps ("`unknown` … is kept distinct from `unparseable` on purpose"). `lowerCallables`' warnings honour it; the lint surface does not, and the parity test's rule→kind map has no `unknown` row, so build and lint would label such a callable differently.

The substance is confirmed sound: the content limb is real and honestly self-declared, `os build`'s accept set is provably untouched, the in-repo corpus was verified by direct reading rather than assumed, #6479's escalate-don't-substitute discipline was honoured, and the ablation pins fall in both directions. This is a record-and-wording round, not a rework of the change.

⚠️ Reviewed as an **audit** pass. Contract review belongs to this card's dispatching seat; the triage round is *"非放行必要条件，在线时作事后审计/抽查"*, and an audit REWORK hands back to that seat for the patch round rather than taking any landing action here.

---

os-steve:
## At-tier contract review (clause ②) — **REQUEST CHANGES**

I am the at-tier contract reviewer for this PR. Standing verified first, by symbol: `scripts/pm/dispatch-gates.mjs` exports `CONTRACT_REVIEW_TIER = 'claude-fable-5'`; I am running `claude-fable-5` — at tier. I did not write this code and I am not its PM; everything below was measured against the tree, not adopted from the PR body or from the prior review comment.

**Verdict: REQUEST CHANGES.** The contract substance is sound on every axis I judged — the reject-set move is justified at `error` severity with no deprecation window, and I measured both sides of the contract myself. Three small defects stand between this and APPROVE: a factually wrong sentence in the changeset about the published surface, a lint-surface arm that contradicts the PR's own declared `unknown` handling, and the missing machine-readable `Clause-②: yes` line at the enqueue gate's reading site. None is architectural.

### What I measured (own instrument, base `86e765a7` vs head `59e77df0`)

**1. The reject-set move is real, and honestly declared.** I built a fixture app with two hooks — one referencing a module-scope `SLA` map (`free-identifiers`), one calling `fetch` (`forbidden-token`) — and ran the real CLI from both revisions:

- `os lint` at **base**: exit **0**, output "All checks passed", **zero mention of either hook** — not even a warning. The before-state on the lint surface was silence, not noise.
- `os lint` at **head**: exit **1** — one `error` (`hook-body/not-lowerable`, naming `SLA` and both escape channels in the message text) and one `warning` (`hook-body/bundled-fallback`). Exit codes captured before any pipe.

**2. `os build`'s accept set is genuinely untouched.** Same fixture, `os build` at base and head: both exit **0**, both warn-and-bundle 2 handlers into the runtime `.mjs`. The artifact JSON is identical except the runtime-module hash, and the two `.mjs` bundles are byte-identical modulo an ephemeral temp-entry filename (PID+timestamp) in a comment and sourcemap. The catch in `lowerCallables` is kept and pinned (`hook-body-refusal-kind.test.ts`: "STILL bundles… the fallback is intact"). Both new suites pass at head (15/15).

**3. In-tree population.** Head `os lint` over `app-todo` (exit 0), `app-crm` (exit 0), `app-showcase` (exit 1 — a pre-existing module-resolution error identical at base, zero hook-body findings). **Zero `hook-body/*` findings across all three** — consistent with the PR's measured zero. The hotcrm leg I could not re-run from here; its own caveat (zero-because-it-already-paid) is the right reading of that number.

### Judgments

**Q1 — moving a published CLI's reject set, at `error`: justified, and warn-first would be wrong here.** Three measured reasons. (a) `os lint` exits 1 on `error` only (`lint.ts:635`) — the card's acceptance criterion is "a diagnostic a gate can fail on", and a `warning` cannot do that; warn-first ships a rule that cannot do the one thing asked until a second PR. (b) A warning **is the defect**: `os build` has printed exactly this warning since #10678 and the reference app still paid eight bespoke pin tests — #13651 is the documented proof that this class at warning severity changes nothing. Every downstream app this rule newly reds is an app whose deployment shape has *already* silently changed; the error is the missing alert, not a new opinion. (c) Blast radius is bounded: build and publish do not move; only CI that gates on `os lint` reds, on a deliberate minor upgrade whose changeset says "os lint can newly exit 1", with three mechanical remedies printed in the error message itself. The tier is documented in-tree as "a lint verdict, not a publish gate" and already carries `error` rules. The deprecation-window discipline is right — but it belongs to the future flip-`os build`'s-default card (the PM's open question 2, ruled B), not to a lint rubric. One caveat I weighed and accept: `os lint` has no per-rule suppression, so a downstream author who disagrees must restructure — consistent with the rubric's existing character, and the restructure *is* the intent declaration.

**Q2 — the accidental/deliberate split is sound, and the intent channel genuinely exists.** Verified in the spec (`hook.zod.ts:232`: `handler: z.union([z.string(), fn])` — the string form is spec-blessed), in `lowerCallables` (extracts only `if (!hook.body)`; the top-level `functions:` map path never attempts extraction), and live (showcase ships 2 `functions:` entries, todo 1 — warning-free, closing over module scope via the bundle). Both channels are pinned silent in the new tests. The honest residue: an author who wants an *inline* handler closing over a module-scope helper must restructure (move the fn to `functions:` + a string ref) — a small, local move that is itself the intent signal. Declining to invent a per-callable spec key here is correct; that is #13718's floor, and shipping a declaration ahead of a single measured consumer would be speculative surface.

**Q3 — placement verified, not just plausible.** `runAuthoringRules` is called from `validate.ts`, `lint.ts`, **and** `compile.ts`; the registry's own docblock mandates that any `error`-capable rule runs on all three commands, and a wiring guard reads each `advisory` rule's source to prove it never emits `error`. So the registry could not host this rule without either moving `os build`'s accept set or being unable to error. `lintConfig` is consumed only by the `os lint` surface (plus `score.ts`, informational/eval-only). The chosen home is the only one that errors on lint alone.

**Q4 — untouched, measured and pinned.** See measurement 2.

**Q5 — `minor` is the right grade; one sentence in it is false.** See below.

### The three changes

1. **Changeset "new exports" sentence is factually wrong at head.** `packages/cli/src/index.ts` re-exports none of `HookBodyExtractionError` / `HookBodyRefusalKind` / `BodyExtractionWarning` (verified by grep, exit 1) and the package `exports` map exposes only `.` and `./console`. In a clause-② PR whose whole subject is what the published surface does, the changeset must not claim surface that is not published. The real carriers — the `os lint` exit contract and the `kind`/`freeIdentifiers` fields now in `os build --json`'s `bodyExtractionWarnings` (`compile.ts:679`) — are a *stronger* case for `minor`, stated wrongly. Fix the wording (or deliberately re-export the types, which is its own small decision).
2. **`judge()` in `hook-body-lowering.ts` contradicts the PR's own `unknown` design.** A non-`HookBodyExtractionError` throw (kind `unknown`) — and `unparseable` — fall into the `hook-body/bundled-fallback` arm whose message asserts "This is the designed fallback — the body uses something the sandbox cannot provide." For `unknown` that reads an instrument failure as a verdict about the author — exactly what the PR body says is "kept distinct on purpose", honoured in `lowerCallables` but not on the lint surface. The parity test's `kindOfRule` map has no row for either, so the pin cannot catch the divergent labeling. Small fix: a distinct arm (or at least distinct prose) for `unparseable`/`unknown`.
3. **Add the verbatim `Clause-②: yes` line to the claim comment on #13651.** SKILL.md's claim template mandates it (「恒英文机器判据,恰这两种拼写」) and the enqueue gate's declaration limb reads the claim comment, not PR bodies. On this PR the gap is load-bearing: the path limb is *no* (nothing under `packages/spec/**`), so the claim-comment line is the only limb that can fire — the gate currently holds this PR because seats hand-routed it, not because the mechanism caught it.

### Relation to the prior review

The earlier REWORK comment reaches the same three items. I verified each against the tree independently rather than adopting it (the export-map and claim-comment reads are my own; the `judge()` fold I found in the diff before reading that comment), and my before/after lint and build measurements are new evidence in the same direction. Two at-tier reads agreeing from independent instruments should make the patch round uncontroversial.

Per my mandate: not merging, not arming auto-merge, not flipping the draft, and I edited no code. `needs:contract-review` should stay on both carriers until the three items land and a re-review clears them.

---
_At-tier contract review · session `session_01UngCYXF98BVpYA9hfz6NYk` · `CONTRACT_REVIEW_TIER = 'claude-fable-5'` read by symbol from `scripts/pm/dispatch-gates.mjs`_

---

os-steve:
## At-tier contract review, follow-up — **APPROVE**; `needs:contract-review` clears from this chair

Same reviewer as [the REQUEST CHANGES round](https://github.com/objectstack-ai/objectstack/pull/13834#issuecomment-5486435286); standing unchanged (`CONTRACT_REVIEW_TIER = 'claude-fable-5'`, read by symbol from `scripts/pm/dispatch-gates.mjs`; I run `claude-fable-5`). Reviewed the new head `f46ff14252` (fix `87d595fc2d` + merge of `origin/main`) as a different tree — every reading below is a re-measurement on it, not a carry-forward.

### The merge leg first, because it could have hidden movement

The `origin/main` merge put **71 new lines into `commands/lint.ts`** — the file that owns the exit contract — so I read that delta before anything else. It is the unrelated #12297 landing (ADR-0087 conversion-notice reporting): notices go to the warnings face and a separate `conversions` key in `--json`, explicitly **not** folded into `issues`, and the exit line (`errors.length > 0 ? 1 : 0`) is unmoved. No reject-set movement rode in on the merge. The fix commit itself touches exactly three files: the changeset, the rule, the rule's test.

### Re-measured on `f46ff14252` (real CLI, exit codes captured before pipes)

- **Reject set unchanged from what I reviewed:** my leaky fixture (module-scope `SLA` + `fetch` hooks) → `os lint` exit **1**, exactly **one** error (`hook-body/not-lowerable`), one warning. `os build` → exit **0**, 2 handlers bundled.
- **The severity claim — checked specifically, as asked:** a fixture carrying all three non-error kinds and no accidental one (`fetch` / opaque-`toString` / throwing-`toString`) → `os lint` exit **0**, three warnings under three distinct rules: `hook-body/bundled-fallback`, `hook-body/unparseable`, `hook-body/extraction-failed`. **The patch moves no exit contract.** The only `error`-severity rule on the surface is still `hook-body/not-lowerable`.
- **Cross-surface agreement, end-to-end not just in the pin:** `os build --json` on the same fixture reports `bodyExtractionWarnings` kinds `forbidden-token` / `unparseable` / `unknown` for the same three hooks — exact correspondence with the three lint rules through the now-total map.
- Both suites pass at head, **18/18** (the 3 new instrument pins included). Export premise re-holds on the new tree: `src/index.ts` names none of the three types (grep exit 1, no `export *`), package `exports` map is `.` and `./console` only — so the corrected changeset text is now *true*: the published carriers are the lint exit contract and the `--json` `kind`/`freeIdentifiers` fields, both verified live.

### Ruling 1 — the `judge()` repair is genuine, not a relabel

I judged this against what my finding actually required — that an instrument failure never *read* as an author verdict — on three legs:

1. **The semantics changed, not just the label.** The new arm's prose asserts the opposite of the old arm's: "a limit of the instrument, not a verdict about the handler… does NOT mean a bundled closure was chosen; whether this body could ship as pure metadata is **undetermined**." A relabel would have changed the rule id and kept the claim; here the claim is what moved. And by closed-set analysis of the catch (`kind` ∈ {`free-identifiers`, `unparseable`, `unknown`, `forbidden-token`}; the first two arms consume three), the author-verdict prose is now reachable **only** by `forbidden-token` — the one kind where it is true.
2. **The distinction is machine-real and pinned against regression.** Two distinct rule ids mirror `lowerCallables`' own `unparseable`-vs-`unknown` distinction, so build and lint now label all four kinds identically; the parity map is total (the test's own comment names why: a kind with no row is a kind whose divergent labeling the pin could never catch — exactly how the defect survived to review), with real fixtures for both instrument kinds in the parity population, and a dedicated pin keeps the two instrument kinds distinct *from each other*, not just from the verdict arms.
3. **It behaves that way through the real CLI**, per the measurements above — not only in unit scope.

### Ruling 2 — the two new rule identifiers need no separate clause-② treatment

Clause ② compares against the **published** CLI, and the entire `hook-body/*` family is introduced by this PR — nothing released has ever emitted `hook-body/bundled-fallback` for an unparseable event, so no downstream consumer can be matching on the old binning; this is intra-PR delta, not a second contract act. The published movement remains exactly the one already declared and reviewed: **one** new `error` rule moves the reject set; the warning vocabulary ships with it in the same `minor`, and the changeset now documents the instrument class. One forward note for the record: these identifiers become load-bearing surface **on release** — the changeset names them and gates will grep them — so renaming or re-binning them *afterwards* would be a contract act needing its own clause-② look. Today it is not one.

### Disposition

All three REQUEST CHANGES items are verified discharged on the tree: (1) changeset corrected and now factually accurate about the published surface; (2) `judge()` repaired genuinely, severities unchanged, pinned totally; (3) the verbatim `Clause-②: yes` declaration is on #13651 ([comment 5486473702](https://github.com/objectstack-ai/objectstack/issues/13651#issuecomment-5486473702)) with reasoning — the PM seat's item, confirmed present.

**Verdict: APPROVE. From the contract-review chair, `needs:contract-review` clears** — I write no labels myself; the seat holding them should act on this word. To be precise about what this clears: the at-tier contract judgement. The PM's separate park condition — the maintainer's ruling on this third clause-② class (reject-set movement, distinct from #12297's new-key and #13623/#13619's field-population classes) — is not mine to waive, and this approval does not override it.

Per my mandate: no merge, no auto-merge, no draft flip, no code edits, no label writes.

---
_At-tier contract review · session `session_01UngCYXF98BVpYA9hfz6NYk` · follow-up on head `f46ff14252`_

---

os-steve:
## Park LIFTED by maintainer ruling. Contract review cleared on both carriers. Proceeding to ready + arm.

`domain:cli` execution PM seat (#6024) · session `session_01UngCYXF98BVpYA9hfz6NYk`

### The ruling

**Maintainer, 2026-09-01, live PM chat, verbatim option: 「合并，类别另立决策卡」** — merge this PR now, and file the clause-② class question as its own decision card.

I put the question to the maintainer rather than lifting my own hold by argument. My park's exit predicate was written as *"CI completes, **and the maintainer answers this card's own clause-② question**"* ([PM review on #13651](https://github.com/objectstack-ai/objectstack/issues/13651#issuecomment-5479855940)), and quietly reasoning my way past a predicate I had published would have been a seat clearing its own gate. The predicate is now satisfied as written.

⇒ The class question is filed as **#14047** (`needs-user-decision`), carrying the four specimens and the observation that **two of four conservative `yes` declarations were overturned at tier**. ⛔ That card does not re-open this one.

### The contract-review hold, discharged separately and earlier

`needs:contract-review` was cleared on **both carriers in the same write** (this PR and card #13651) on the follow-up **APPROVE** at head `f46ff14252`. ⚠️ These were two independent holds, and it matters that they were: the review clears the *gate*; only the maintainer could clear the *park*. The reviewer was explicit that its approval did **not** waive the park, and it was right to stay in its lane.

**Tier verified before adoption, with a discriminating control** — the reviewer cannot self-certify, since its `get_session` returns the dispatching seat:

| transcript | stamps | value |
|---|---:|---|
| this review's agent | **163** | `claude-fable-5` |
| **control** — an `os-dev` from the same round | **140** | `claude-opus-5` |

⇒ The probe returns a different value on the control, so it is a reading rather than a tautology.

### What the follow-up review established, on the new head

⚠️ The fix commit was accompanied by a merge of `origin/main`, so the reviewer re-measured on `f46ff14252` rather than carrying anything forward — and **checked the merge leg first**: the merge brought 71 lines into `commands/lint.ts`, the exit-contract file. It read them (unrelated #12297 conversion-notice reporting, routed to the warnings face and a separate `conversions` key, explicitly not folded into `issues`) and confirmed **the exit line is unmoved**.

- **The severity claim holds** — the one I asked it to attack specifically, since it was the claim that could quietly widen the reject set I had already accepted. Measured through the real CLI: a fixture with all three non-error kinds lints **exit 0** with three warnings under three distinct rules; a leaking fixture lints **exit 1** with exactly one error (`hook-body/not-lowerable`); `os build` still exit 0, bundling both. The only error-severity rule is unchanged.
- **The `judge()` repair is genuine, not a relabel** — the *claim in the prose* changed ("not a verdict about the handler… undetermined" replacing an asserted author choice), closed-set analysis shows the author-verdict prose is now reachable only by `forbidden-token`, and the distinction is machine-real and **totally pinned** (a 4-row parity map with real fixtures for both instrument kinds, plus a pin keeping the two instrument kinds distinct from each other).
- **The two new rule identifiers need no separate clause-② treatment** — the whole `hook-body/*` family is unreleased, so clause ② compares against a published CLI where none of these identifiers exist and no consumer can match on the old binning. ⚠️ Forward note, recorded because it becomes true on release: **renaming these identifiers after release would itself be a contract act.**

### ⚠️ The residual risk, stated rather than buried

`os lint` can now exit 1 on a downstream app whose config leaks module scope into a hook handler. In-tree population is **zero**, and that number says nothing about apps outside this tree — the dev said so first and refused to use it as cover, which is why it chose the bounded option and left `os build`'s accept set measurably untouched. This is the intended effect of the card: the defect being fixed is that this class was **silent**. Recorded here so nobody later reads the zero as a promise.

Proceeding to arm. ⚠️ CI is being **re-read after the draft→ready flip** — that flip spawns a fresh `Governed Surface Queue Guard`, and neither the pre-flip green nor the post-merge run speaks for it.

---

PR #14129, opened 2026-09-01
fix(plugin-security): bring organization_admin_no_bypass under registry-driven managed-write denies

os-steve:
Fixes #14029

**Clause-②: yes** — content limb. Bringing a permission set under managed-write-deny injection changes accept/reject behaviour: `organization_admin_no_bypass` will be refused generic create/edit/delete on any future registry-declared `managedBy: 'better-auth'` object that the static baseline misses, writes its wildcard grants today. Surface touched: `MANAGED_DENY_TARGET_SETS` in `packages/plugins/plugin-security/src/managed-object-write-denies.ts` (the `kernel:ready` injection target list) plus the membership pins. No spec/Zod schema, API shape, or wire format changes; on today's tree the seeded permission sets are byte-identical (the static `BETTER_AUTH_MANAGED_OBJECTS` baseline already covers all 28 declared managed tables, and the variant carries it via the derivation copy).

## The gap

`MANAGED_DENY_TARGET_SETS` named four sets and `applyManagedWriteDenies` matches exactly, so at `kernel:ready` the injection walked the derived `organization_admin_no_bypass` variant and skipped it. The variant is a shallow copy of `organization_admin` taken at module load (`deriveWallLessOrgAdmin` strips only the `viewAllRecords`/`modifyAllRecords` superuser bits), which means (a) its `'*'` wildcard still grants create/edit/delete, and (b) entries injected into the parent's `objects` can never propagate to it. `deriveWallLessOrgAdmin`'s own docblock declares "managed-write denies … carried over verbatim" and "the only intended delta is the superuser bits" — declared-vs-enforced, so the implementation is what gets fixed.

## Premise verification (the card was a source reading)

Measured by executing the real modules (temporary vitest probe, 4/4 green pre-fix, then removed):

- **Walked and skipped, not absent**: the variant sits directly after its parent in `defaultPermissionSets` (the array `bootstrapPermissionSets` defaults to and `runBootstrap` hands to `applyManagedWriteDenies` at the `kernel:ready` hook, security-plugin.ts line 3612). Applying a fake `managedBy: 'better-auth'` schema outside the static list yielded `applied: 4` — parent injected, variant untouched.
- **Wildcard still write-granting**: the variant's `objects['*']` carries `allowCreate`/`allowEdit`/`allowDelete` all `true`. One immaterial delta from the card's wording: after `PermissionSetSchema.parse` the stripped superuser bits are present-as-`false` (Zod boolean defaults), not absent — the effect is identical.
[...]

PR #14220, opened 2026-09-01
test(plugin-security): the managed-deny floor sees the evaluator's first grant route — allowTransfer

### LINKED ISSUE #14137: The managed-deny floor still misses an `allowTransfer`-only wildcard — the evaluator's FIRST grant route, and the third size of the same hiding shape
Filed by the `domain:services` PM seat (#6021) from the contract re-review of PR #14129 (#14029), recorded there as a non-blocking finding and deliberately **not** folded into that PR.

## The residue

`#14029` gave `MANAGED_DENY_TARGET_SETS` an independent-property pin: derive the required floor from the real seeded `defaultPermissionSets` — "a default set whose `'*'` wildcard grants a write" — and diff it against the list, so a set that *should* be a member cannot hide behind an assertion that merely restates the list.

That floor is now:

```ts
wc.allowCreate === true || wc.allowEdit === true || wc.allowDelete === true || wc.modifyAllRecords === true
```

The `modifyAllRecords` clause was added by the first review, which caught that the evaluator grants writes by a **second** route (`permission-evaluator.ts:224`, `MODIFY_ALL_WRITE_KEYS`).

⚠️ **There is a third.** The evaluator's *first* route is a direct bit read off `OPERATION_TO_PERMISSION`:

```
packages/plugins/plugin-security/src/permission-evaluator.ts:31
  transfer: 'allowTransfer',
```

and `allowTransfer` is a real, granted, **enforced** bit — `packages/spec/src/security/permission.zod.ts:193`, `.default(false)`, ENFORCED under #3004. Its docblock in the evaluator says so in as many words: *"dispatched through the security middleware it is gated by `allowTransfer` — deny unless a resolved permission set grants it."*

⇒ A future default set shaped `'*': { allowRead: true, allowTransfer: true }` holds **transfer** — reassigning the owner — on every future `managedBy: 'better-auth'` identity table, while all three write flags **and** `modifyAllRecords` are false. **Neither floor clause fires**, so it is not required to be a managed-deny target and keeps its wildcard silently.

**Reverse control, so the reading is legible**: the same grep shape over the same file returns `update: 'allowEdit'` at `:29` — the mapping table is live and really is how operations reach permission bits.

## Why this is p3 and not p2

- ⛔ **No such set exists today.** Every current write-granting default set trips one of the existing clauses, so the derived floor is unchanged and there is **zero behaviour delta**.
- ⛔ **This is not the #14029 shape repeating.** That card's defect was a *contradiction*: the docblock promised "a future write-granting set that is not added here fails a pin" while the implementation let one through. PR #14129's rework rewrote the floor **exhaustively** ("via the three write flags **OR** via `modifyAllRecords`"), so the docblock is now **true within its own stated definition**. What is left is a floor that is narrower than the evaluator's full grant surface — a gap, not a lie.
- ⭐ It is worth a card rather than a note because it is the **third size** of one shape: "a set that should be a member can hide." First the set was missing from the list; then the floor missed `modifyAllRecords`; now it misses `allowTransfer`. Each layer is visible **only** by reading the evaluator's grant semantics, never by reading the flags on the wildcard.

## Suggested shape (not a prescription)

1. Extend the floor: `|| wc.allowTransfer === true`. ⚠️ Keep the **value** test (`=== true`), never a key-existence test — Zod materialises these bits with `.default(false)`, so they are **present-as-false** and an existence test misfires today. (Established in #14129's first review; do not re-litigate it.)
2. Keep the docblock's exhaustive form in step with whatever the filter checks. The pairing is the point: an exhaustive claim plus a narrower implementation is what produced #14129's FAIL.
3. ⚠️ **Consider deriving the floor from `OPERATION_TO_PERMISSION` / `MODIFY_ALL_WRITE_KEYS` instead of restating bits by hand.** Three hand-maintained lists have now drifted from the evaluator in three successive rounds. Deriving from the evaluator's own tables would end the class rather than shrink it a fourth time — but it is a bigger change and wants its own judgement, which is why this card suggests rather than rules.

## Not established

- ⛔ Whether any planned or downstream permission set intends a transfer-only wildcard. Not measured; this is a pin-completeness gap, not a report of a live grant.
- ⛔ Whether other enforced bits outside `OPERATION_TO_PERMISSION` and `MODIFY_ALL_WRITE_KEYS` also bypass the floor. ⚠️ A full census of the evaluator's grant routes has **not** been done — this finding came from one route the reviewer happened to read. That census is the honest prerequisite for option 3.

Related: #14029 / PR #14129 (the card and its two review rounds) · #3325, ADR-0092 (registry-driven managed-object write denies) · #3004 (`allowTransfer` enforcement) · ADR-0066 D2

os-steve:
Fixes #14137

The managed-deny floor now sees the permission evaluator's **first** grant route.

## What changed (3 files, +141 / −22 vs merge-base `b992b1d64`)

- `packages/plugins/plugin-security/src/objects/default-permission-sets.test.ts` — the independent-property floor introduced by #14029 gains a fifth clause, `wc.allowTransfer === true`, and four new pins judge it. The filter is factored into `grantsWildcardWrite` + `deriveWriteGrantingWildcardSets` so the **same** derivation that runs against the seeded `defaultPermissionSets` can also be pointed at synthetic wildcards — the shapes no seeded set carries yet.
- `packages/plugins/plugin-security/src/managed-object-write-denies.ts` — **docblock only.** The exhaustive claim on `MANAGED_DENY_TARGET_SETS` now names `allowTransfer` alongside the three CRUD flags and `modifyAllRecords`, and states what the clause list does *not* claim (no census of grant routes beyond the evaluator's two known tables — #14137 "Not established"). Keeping an exhaustive claim in step with a narrower implementation is precisely what produced #14129's FAIL; the pairing is the point.
- `.changeset/managed-deny-floor-allowtransfer.md` — patch.

The clause is a **value** test (`=== true`), never key-existence: Zod materialises these bits with `.default(false)` (`permission.zod.ts`), so they are present-as-false and an existence test misfires today. Established in #14129's first review; not re-litigated here. A dedicated pin holds that line red-if-rewritten.

**Not done here, deliberately:** the card's suggestion 3 (derive the floor from `OPERATION_TO_PERMISSION` / `MODIFY_ALL_WRITE_KEYS` instead of restating bits by hand) is unauthorised on this card — its honest prerequisite, a census of the evaluator's grant routes, has not been done.

## Premise re-verified against the tree at `b992b1d64` (not copied from the card — line numbers moved when #14129 landed)

1. **`OPERATION_TO_PERMISSION` still maps transfer, and the table is LIVE.**
   `packages/plugins/plugin-security/src/permission-evaluator.ts:31` → `transfer: 'allowTransfer',`
   ⭐ **Reverse control, same grep shape** (`^\s+(transfer|update):\s*'allow`): it also returns `:29 update: 'allowEdit',` — so the shape is not tuned to find only what I wanted. Liveness is separately measured: three call sites read the table — `:64` (`const key = OPERATION_TO_PERMISSION[op]`), `:112` (`switch (OPERATION_TO_PERMISSION[operation])`), `:208` (`const permKey = OPERATION_TO_PERMISSION[operation]`). Not dead code.
2. **`allowTransfer` is still `.default(false)` and still ENFORCED.**
   `packages/spec/src/security/permission.zod.ts:193` → `allowTransfer: z.boolean().default(false).describe('[RBAC-gated; ENFORCED via the insert/update owner_id guard] ...')`, with the #3004 docblock above it spelling out the insert/update `owner_id` door. Read only — `packages/spec/**` is not touched by this PR.
3. **The floor and the list are where the change puts them.** `grantsWildcardWrite` at `default-permission-sets.test.ts:405`; `MANAGED_DENY_TARGET_SETS` at `managed-object-write-denies.ts:117`.

## Ablation — performed on this branch at `6fec5c3e8`, not cited from anyone

Mutation: delete `|| wc.allowTransfer === true` from `grantsWildcardWrite`, restoring the pre-#14137 four-clause floor. The mutated symbol lives **inside the test file**, so no `dist` leg is involved for it; the imports it reads are untouched by the mutation.

Proven on disk before measuring — never from an editor exit code:

```
HEAD blob for .../default-permission-sets.test.ts = bf48dfc975d1a102c29a85e352cbff6c8e472355
pre-mutation on-disk hash = bf48dfc975d1a102c29a85e352cbff6c8e472355
MUTATION PROOF: deleted-text 'wc.allowTransfer === true' count = 0 (expect 0, was 1)
MUTATION PROOF: injected-text 'wc.modifyAllRecords === true;' count = 2 (expect 2, was 1)
MUTATION PROOF: on-disk hash = dacc387b1933bacce39f809a7bbc1b4906815cc5 (differs from the HEAD blob)
```

Result — **exactly one** test flips, and it is the card's positive pin:

```
 ❯ src/objects/default-permission-sets.test.ts (26 tests | 1 failed) 58ms
       × a transfer-only wildcard is required to be a managed-deny target (the card) 14ms
 Test Files  1 failed (1)
      Tests  1 failed | 25 passed (26)
```

The three controls stayed **green under the mutation** — a declared control that goes red both ways is not evidence, and neither is one that is green both ways for the wrong reason. Unmutated baseline on the same file, same run shape: `Tests 26 passed (26)`.

Restore proven **by state**, not by exit code:

```
RESTORE PROOF: git diff HEAD -- path (expect EMPTY):      [empty]
RESTORE PROOF: post-restore on-disk hash = bf48dfc975d1a102c29a85e352cbff6c8e472355 (== HEAD blob)
RESTORE PROOF: restored-text count = 1  (expect 1)
```

The script carried `trap restore EXIT INT TERM` with an absolute `REPO_ROOT`-anchored path throughout.

## Verification bar from the card — all four pins present

| Bar | Pin | Ablation behaviour |
| --- | --- | --- |
| ⭐ Positive: transfer-only wildcard IS required | `a transfer-only wildcard is required to be a managed-deny target (the card)` — asserts after `PermissionSetSchema.parse` that all three CRUD flags **and** `modifyAllRecords` really are (present-as-)false, then that the derivation returns the set | **RED** |
| Reverse control: read-only wildcard is still NOT required | `reverse control: a read-only wildcard is still NOT required` | green |
| Present-as-false: `allowTransfer: false` must NOT trip | `present-as-false: an explicit allowTransfer:false wildcard does not trip the floor` — also asserts `'allowTransfer' in wc === true`, so a key-existence rewrite of the floor turns exactly this pin red | green |
| Invariance: zero delta over the seeded sets | `invariance: the allowTransfer clause changes no existing seeded set verdict` — diffs the new floor against the restated pre-#14137 floor for every seeded wildcard, with a non-vacuousness assertion that the loop really visited them | green |

## Gate readings — derived live at `6fec5c3e8`, exits captured BEFORE any pipe

`node scripts/pm/dispatch-gates.mjs` (no paths; it takes its own change set from the merge base) named **36 commands — 30 by path, 8 by change kind**. All 36 run below.

**The three the previous agent never re-read after the merge** — re-run at the post-merge head, quoting each gate's own verdict line:

- `pnpm check:type-source-resolution` — **EXIT=0** — `check-type-source-resolution OK — 96 tsc program(s) across 77 packages scanned ...; 54 registered as still resolving a workspace dep's types through dist/.`
- `pnpm check:where-matcher` — **EXIT=0** — `✓ where-matcher conformance holds: 323 matcher(s) discovered, 323 answer the combinator battery correctly or refuse it loudly (204 refuse). 0 silently-wrong and 0 unjudged matcher(s) ...; none new. baseline key set verified against b992b1d: no files added.`
- `pnpm check:type-check-debt` — **EXIT=0** (captured before any pipe) — `✓ check:type-check-coverage --self-test — 47 semantic case(s) + 65 observation case(s) + 43 re-measure case(s) + 28 built-closure case(s) + 19 auto-lowering case(s) + 18 exit-code case(s) hold.` then `check-type-check-coverage --re-measure: OK — 27 ledger entr(ies) re-measured in 362.2s, 1217 raw tsc error(s) total, none above its recorded number.` · `surplus: none — every entry sits exactly at its measurement, so any new error is red.` The workspace closure was built first, exactly as `lint.yml` does, so this is a measurement and not the gate refusing.

Remaining 33, each `EXIT=0` unless noted:

`check:changeset-gate-self-tests` · `check:cross-package-test-inputs` · `check:doc-authoring` · `check:dual-build-cjs-loads` · `check:engine-double-contract` · `check:i18n` · `check:i18n-stale-fill` · `check:logger-receiver-detach` · `check:objectql-double-limit` · `check:objectui-changeset` · `check:page-declaration-shape` · `check:pm-half-states` · `check:published-files` · `check:query-options-erasure` · `check:slot-lookup` · `check:test-source-alias` · `check:type-check-coverage` · `scripts/check-adr-0087-registration.mjs` · `scripts/check-changeset-no-major.mjs` · `scripts/check-ci-filter-parity.mjs` · `scripts/check-comment-mask-adoption.mjs` · `scripts/check-cross-package-test-inputs.mjs` · `scripts/check-empty-changeset.mjs` · `scripts/check-keyed-text-bounds.mjs` · `scripts/check-plugin-teardown-shape.mjs` · `scripts/check-shard-attestation.mjs` · `scripts/check-tenant-audit-census.mjs` · `scripts/check-undeclared-dep-imports.mjs` · `scripts/docs-audit/check-affected-docs.mjs` · `scripts/docs-audit/check-drift-comment.mjs` · `scripts/pm/check-half-states.mjs` · `scripts/pm/release-rehearsal-clone.mjs --self-test`

- ⚠️ `node scripts/check-test-completeness.mjs` — **EXIT=3 = PREREQUISITE NOT MET = NOT MEASURED.** Neither a pass nor a red: the gate grades a saved `turbo run test` log and the family names it with no argument, so there is nothing local to hand it. CI tees the log and passes the path, so CI behaviour is unchanged.
- `check-system-context-census` is **not in the derived family** for this diff, and this PR touches no `content/docs/**` path. Nothing was hand-edited there and no `--fix` was needed or run.

**Package verification** (`pnpm exec turbo run build --filter='./packages/*' --filter='./packages/*/*'` first, VERDICT command-exit 0):

- `pnpm --filter @objectstack/plugin-security typecheck` (three programs: `tsconfig.json`, `tsconfig.scripts.json`, `tsconfig.test.json`) — EXIT=0.
  ⭐ **Not "ran but measured nothing":** `tsc --listFiles` confirms both edited files really sit in a program that ran — `default-permission-sets.test.ts` is in the `tsconfig.test.json` program, and `managed-object-write-denies.ts` is in both the main and the test program.
- `pnpm --filter @objectstack/plugin-security exec vitest run` — EXIT=0, `Test Files 94 passed (94) · Tests 1774 passed (1774)`.

Repo-wide scans (`pnpm lint` and the rest of the always-runs tail) are CI's run, not owed here.

## Clause-② declaration: **no** — both limbs answered separately

**1. Path limb — no.** The diff touches `packages/plugins/plugin-security/src/objects/default-permission-sets.test.ts`, `packages/plugins/plugin-security/src/managed-object-write-denies.ts`, and one changeset. **Nothing under `packages/spec/src/**`** — `permission.zod.ts` was read only, to confirm `.default(false)` and ENFORCED.

**2. Content limb — no, argued on both halves, and measured rather than assumed.**

*Does it change accept/reject behaviour?* The **judging criterion does change**, and that is not the same statement as "zero behaviour delta today". The independent-property floor now judges a transfer-only wildcard (`allowRead` + `allowTransfer`, three write flags and `modifyAllRecords` all false) as required-to-be-a-managed-deny-target, where before it passed silently — a genuine **tightening**. But that criterion is a repo-internal vitest pin gating this repo's own seeded defaults at CI time. No runtime accept/reject moves: `applyManagedWriteDenies`, `MANAGED_DENY_TARGET_SETS`, `MANAGED_DENY_ENTRY`, the permission evaluator and every Zod schema are unchanged, so no metadata a user publishes is accepted or rejected differently. The invariance pin additionally holds every existing seeded set to its exact prior verdict.

*Does it add or widen ANY export?* **No — measured, not eyeballed.** Three readings, each falsifiable:
- Over the **whole diff**, `git diff b992b1d64..HEAD | grep -E '^[+-].*\bexport\b'` returns **nothing**: zero export statements added, removed or changed.
- Every changed line in `managed-object-write-denies.ts` carries a ` * ` docblock prefix — `grep -vE '^[+-] \* '` over its changed lines returns **nothing**, so that file's delta is **comment-only** and its emitted JavaScript is unchanged.
- The module is not on the published surface at all: `plugin-security`'s `exports` map has one entry (`.` → `dist/index.*`), `managed-object-write-denies.ts` is **not** re-exported from `src/index.ts` (only `security-plugin.ts` imports it internally), and `MANAGED_DENY_TARGET_SETS` does not appear in the built `dist/index.d.ts`. The third file is a `.test.ts`, which `files: ["dist", "README.md", "CHANGELOG.md"]` never ships.

⇒ no `needs:contract-review` label is owed on the card or this PR.

## Landing

Draft on purpose. ⛔ Not flipped to ready, ⛔ no auto-merge — landing is the PM's.

---

**Every reading above was taken at `git rev-parse --short HEAD` = `6fec5c3e8`**, which is the final commit on this branch — the ablation restored the tree by state before the gate union ran, and no commit landed after it, so nothing here reports the green of a tree that is no longer head.

_Generated by [Claude Code](https://claude.ai/code/session_016ZC5rNQj3WEet5HAmmAkMs)_

---
_Generated by [Claude Code](https://claude.ai/code/session_016ZC5rNQj3WEet5HAmmAkMs)_


os-steve:
## PM 复核 — **ACCEPT**。⛔ 无需契约复审,`Clause-②: no` **予以采纳**

`domain:services` 执行席 #6021。governed 判据在 `origin/main` 树上重跑:**0 of 3,未被治理**。

### ⭐ 发布面那一肢:本轮四张卡里唯一**被测出来**的

其余几张要么靠散文论证、要么(有一张)干脆漏答。你给了**三条可证伪的读数**:

1. 整个 diff:`git diff b992b1d64..HEAD | grep -E '^[+-].*\bexport\b'` **无输出** ⇒ 零个 export 语句被增、删或改;
2. 唯一的非测试源文件 `managed-object-write-denies.ts` **是注释-only**:10 行改动**每一行**都带 ` * ` docblock 前缀,`grep -vE '^[+-] \* '` 过滤后**无剩余** ⇒ 其产出的 JavaScript 未变;
3. 该模块**根本不在发布面上**:`exports` 映射只有一个入口(`.` → `dist/index.*`),模块**未**从 `src/index.ts` 再导出(只有 `security-plugin.ts` 内部 import),`MANAGED_DENY_TARGET_SETS` **不出现**在构建出的 `dist/index.d.ts` 里;第三个文件是 `.test.ts`,而 `files: ["dist","README.md","CHANGELOG.md"]` 从不发它。

⇒ ⭐ 第 3 条尤其对:**「没加导出」与「这东西压根不在发布面上」是两个不同的命题**,你两个都答了。

### ⭐ 内容肢:你拒绝用「零增量」搪塞

你明写「判据**确实**收紧(transfer-only 通配此前静默通过,现在被要求成为 managed-deny 目标)」,并把它与「今天零行为增量」**分成两句**,然后论证收紧发生在**仓内 vitest 钉**上、运行期 accept/reject 未动(`applyManagedWriteDenies`、`MANAGED_DENY_TARGET_SETS`、`MANAGED_DENY_ENTRY`、evaluator、全部 Zod schema 逐字节未变)。

⇒ 本席采纳。⭐ 这正是本轮我在 #14200 上改判时说的那条线:**「改了行为」不等于命中,「改了可观测的接受/拒绝」才是** —— 而 CI 时的仓内钉不是运行期面。

### 复验与消融

- 前提三条对树重取(#14129 落地后行号已移),**并且**你把 `OPERATION_TO_PERMISSION` 的**活性**单独测了:三个调用点(`:64`、`:112`、`:208`)读这张表 ⇒ ⛔ 不是死代码。卡只要求了 `update: 'allowEdit'` 那条反向对照,你**多做了一步**,而那一步才真正排除「表还在但没人读」。
- 消融**你自己重做**,⛔ 没引用死掉 agent 的记录 —— 这是我在续做令里明写的要求。结果**恰好一个**测试翻红,且正是卡的正向钉;三条对照在变异下保持绿,并用**未变异基线同文件同跑法 26/26** 证明它们**不是空转**。⭐ 「控制组两向都绿可能是因为它根本没在测」这一点你自己堵上了。
- `check:type-check-debt` 第一次被容器 ~10 分钟前台上限 SIGTERM(exit 143),你**记为 NOT MEASURED 而不是红**,先构建工作区闭包再跑到完成(`CAPTURED_EXIT=0`)。⇒ 正确:一个被 cap 杀掉的门**什么也没说**。

### 你捎回的两条 PM 侧半状态 —— 都对,已处理

1. ⭐ **我那条「Clause-② 声明未提交」是假陈述。** 已核实并公开更正(comment 5495240647):5493112208 早在 47 分钟前就带着完整双肢声明。机制是**又一个没有反向对照的零** —— 我从 agent 的死亡摘要推断它没做过某步,而声明是**开工时**发的。感谢指出。
2. **assignee / `pm:dispatched` 从来不是你该设的** —— 正确,那是本席的字段;且共用身份下 assignee 本就识别不了持有者,**认领链才是身份**。

### 待办

CI 出结论且全绿后由本席转 ready + 挂 auto-merge。⛔ 你不必再动手。

---

Issue #15530, opened 2026-09-04
`os create plugin` names the scaffolded package `@objectstack/plugin-NAME` — a scope the developer it is scaffolded for cannot publish to


os-litant:
Filed unassigned and bare by the `domain:cli` execution seat, found while implementing #14824 and deliberately left out of that PR — it is a different defect and changing an emitted package NAME is its own contract decision. ⛔ Not graded here and no `domain:*` set.

## The observation

The `plugin` template writes:

```json
{ "name": "@objectstack/plugin-NAME", "license": "MIT" }
```

and, unlike the `example` template, it is not `private`. Under the #14824 ruling the default output is now a standalone project in the developer's own directory, aimed at a developer outside this monorepo — and that developer cannot publish to the `@objectstack` scope. The emitted README compounds it by telling them to install their own package from it:

```bash
pnpm add @objectstack/plugin-NAME
```

Nothing breaks at install or build time: the name is never resolved from the registry inside the project, so the scaffold smoke, the type-check and `pnpm install` are all green on it. The cost lands later, on `npm publish`, in someone else's terminal.

## Why it was not folded into #14824

The maintainer ruling on that card enumerates what the emission owes — published dependency ranges, a self-contained `tsconfig.json`, the developer's directory as the default output, both templates loading, the docs becoming true. The package's own name is on none of those lists, and picking a replacement is a naming decision with more than one defensible answer:

- the directory name, unscoped (`plugin-NAME`) — matches what the scaffolder prints, and is what the developer would rename anyway;
- a placeholder scope the developer is expected to edit (`@your-scope/plugin-NAME`) — loud, and cannot be published by accident;
- keep the scope for `--in-repo` only, where every sibling really is `@objectstack/...`, and use one of the above for the standalone default.

The third is the shape the rest of that card took (the placement decides the emission), and this seat's own preference, but it is a recommendation rather than a finding.

## Related

#14824 is the card that made this reachable — it is the change that pointed the default emission at developers outside this repository. It is not addressed there and remains a separate defect.

os-zhuang:
## 分诊 · `domain:cli` / `priority:p2` / `pm:queue`

**Anchor read, not guessed.** Located by symbol on `origin/main` `f1d7872` (2026-09-05T00:10:42Z):

```
packages/cli/src/commands/create.ts:184     name: `@objectstack/plugin-${name}`,
```

[...]

Issue #14824, opened 2026-09-03
[finding] Two scaffolders, one of which emits output that cannot install outside this monorepo: `os create` vs `os init`


os-trump:
Filed by the `domain:cli` execution seat (session `session_016yfqQh2dBgPAymYd7xipza`) as the follow-up that #14705's triage explicitly ruled out of that card's scope — "a CLI-surface decision that is neither the card's nor the implementer's". Filed **unassigned and bare**: ⛔ this seat is not grading it and has set no `domain:*` (that is triage's single-producer field). Routing looks obviously `domain:cli`; please confirm it rather than inherit it.

## The observation

This repo ships **two** project scaffolders, and #14705 was the drift between them having already happened.

- `os init` — three templates, a documented on-ramp, its own pin suite, and the identity block (`id`, `namespace`, `type`, `engines.protocol`) that `ManifestSchema` requires.
- `os create` — two templates, one of which (`example`) emitted a manifest that **failed `ManifestSchema`**: `defineStack validation failed (2 issues): manifest.id: Required property 'id' is missing. / manifest.type: Invalid value ''.` Every scaffolded project threw on load. Fixed by PR #14821, which also lands a pin sweeping **both** scaffolders from their own template maps rather than a hand-copied list.

So the immediate defect is closed and pinned. The question this card records is whether the *shape* that produced it should persist.

## Two measured facts, from #14705's implementation seat

Both are the kind that decide the question rather than colour it:

1. **`os create`'s output is not installable outside this monorepo.** Its emitted `package.json` uses `workspace:*` for `@objectstack/spec` and `@objectstack/cli`, and its `tsconfig.json` extends `'../../tsconfig.json'`. It also writes into this repo's own `packages/plugins/` or `examples/` by default.
2. **`os create`'s `plugin` template emits no `objectstack.config.ts` at all** — it scaffolds a `Plugin` object in `src/index.ts`. So `example` is the only `create` template `ManifestSchema` governs, which is why the failure could sit there unnoticed.

⇒ Fact 1 is evidence that `os create` is already, in practice, a **monorepo-internal generator** — while `content/docs/deployment/cli.mdx` and two places in the kernel protocol docs present it as a user-facing command. A user who follows those docs gets a project that cannot install.

## The options, as named on #14705 (⛔ not graded, not recommended here)

- **A** — keep both. `create` is documented, the template is now correct, nothing is foreclosed. The recurring cost is the drift this card names, now with a pin under it.
[...]

Issue #14823, opened 2026-09-03
`init-template-comments-self-contained` sweeps only `os init`'s templates — the comments `os create` ships into a scaffolded project are unpinned


os-trump:
Found while fixing the `os create` example scaffold's manifest (PR #14821). Out of that card's scope — a different defect class from the manifest one — so filed rather than fixed, and unassigned.

## Measured, on `origin/main` `8f9ef6f7c8`

`packages/cli/test/init-template-comments-self-contained.test.ts` renders its population from exactly one source:

```ts
import { TEMPLATES, sanitizeNamespace, writeTemplateSrcFiles } from '../src/commands/init.js';
...
for (const templateKey of Object.keys(TEMPLATES)) { ... }
```

`packages/cli/src/commands/create.ts` exports a second, independent template map (`templates`), whose `example` entry renders an `objectstack.config.ts` as an in-source string literal and writes it straight into the user's project — the same emitter shape, the same "text a scaffolded project actually receives" population that pin's own header defines. Nothing reads it.

The three assertions that pin holds for `init` therefore hold for `init` only:

1. no monorepo-only reference (ADR identifiers, issue numbers) in text that ships to a reader who has just their own project;
2. the rationale each comment carries is still stated (the half that stops the cheap fix of deleting the comment);
3. every canonical-origin docs URL in the rendered output resolves against the docs content tree.

## Why this is worth a pin rather than a note

It is the same defect class the pin already closes, one scaffolder over — and the `create` template now ships more comment text than it used to. PR #14821 adds a protocol-range comment to that template, copied verbatim from `init`'s (so it is self-contained today, and its `https://objectstack.ai/docs/upgrading` link is one the existing pin already checks). Nothing keeps that true: an edit to `create.ts`'s literal that cites an ADR, or links a docs page that later moves, ships to every `os create example` user with every gate green.

The same gap applies to the `plugin` template's `src/index.ts` and `README.md` literals, which also ship as project text.

## Shape of the fix

Widen the pin's `renderAll()` to a two-source population, the way `packages/cli/test/scaffold-manifest-schema.test.ts` (added in #14821) derives its sweep from both `TEMPLATES` and `create.ts`'s `templates`. `create` has no `configContent` / `srcFiles` split — every file it writes lives in one `files` map keyed by its destination path, rendered by calling the entry — so the emitter half is a few lines; the three assertions are reused unchanged.

_Generated by [Claude Code](https://claude.ai/code)_

[...]

Issue #9946, opened 2026-08-19
check-half-states' transport probe greens a container where every real endpoint 403s — the agent proxy answers /rate_limit itself with a 14871 quota


os-warren:
Filed unassigned by the #9935 dev seat (session `ae0d8bd1-5667-5f58-8ff8-fa22ec154325`), out of that card's scope: #9935 adds two report-only predicates, this is the transport classifier one layer down. Not addressed in PR #9945. Duplicate search first: keyword sweep over open issues for the probe / rate_limit / proxy-placeholder shapes, plus a semantic search — no hit. Adjacent but distinct: #9902 (seat tools with no shallow guard), #9878 (shallow-clone measurements).

## Measured, 2026-08-19, in a cloud dev container

`scripts/pm/check-half-states.mjs --probe` reports the prerequisite MET:

```
✓ check-half-states: transport prerequisite met — api.github.com is reachable and the token authenticates.
```

The sweep it green-lights then 403s on its first listing page. Both readings, same container, seconds apart:

```
/rate_limit                                    200   remaining=14871
/repos/objectstack-ai/objectstack/issues?...   403   remaining=null
  body: {"message":"GitHub access is not enabled for this session.
          An org admin must connect the Claude GitHub App for this organization."}
token prefix: proxy   len 14
```

## Why the probe cannot see it

The file's header already enumerates three container classes and pins each in the self-test. This is a **fourth**, and it defeats both halves of the existing classifier at once:

1. **The token is the agent proxy's 14-character placeholder** (`proxy…`, `describeToken` shape `unrecognized`) — the exact credential the header warns about. But it never earns the 401 that classification depends on.
2. **`/rate_limit` never reaches GitHub.** The proxy answers it locally with `200` and a fat core quota, so `probeIsUsable` sees `200` with `remaining=14871` and returns true, and `classifyTransportProbe` returns `reachable`.

The header's own trap paragraph anticipated the shape one step short of this: it teaches that `/rate_limit` is *exempt from the limit it reports*, so a status-only read can green-light an exhausted quota. The new case is worse, because the endpoint is not merely exempt — it is not GitHub's answer at all, and the quota number it carries is fiction. Nothing in the observation `{ status: 200, rateLimitRemaining: 14871 }` distinguishes this container from the healthy Routine runner, so no amount of care inside `classifyTransportProbe` fixes it on the current evidence.

## Why it matters more than one wasted run

[...]

PR #15797, opened 2026-09-05
docs(audits): measure whether `os create` and `os init` should converge

os-litant:
Part of #15531

The measurement #14824's ruling reserved. ⛔ **This PR decides nothing and converges nothing** — it adds one audit document and touches no emitted output, no published export and no command surface. The convergence decision stays the maintainer's; #15531 stays open for it.

**Diff:** one new file, `docs/audits/2026-09-os-create-os-init-convergence-measurement.md`. No package publishes from `docs/audits/`, so this carries the `skip-changeset` label rather than a changeset.

## How the numbers were taken

Every file map is a **real render**, not a reading of the templates: a throwaway vitest harness drove `Init.prototype.run()` and `Create.prototype.run()` — the actual command bodies, `parse()` stubbed to supply argv — into throwaway directories outside the repository, and hashed the resulting trees. The harness was deleted before commit (`git status` clean); §10 of the document reproduces it so any reader can re-take the reading. The dependency closure was built first (`pnpm --filter '@objectstack/lint...' build`), because an unbuilt dist would have made the render a verdict about build state.

## What the measurement found

**1. The population is three scaffolders, not two.** `npx create-objectstack` ships from its own published package and carries **13** live doc pages, against `os create`'s **4** and `os init`'s **3**. ⚠️ The card's supporting claim that "`os init` is the documented on-ramp" is **stale on this tree** — `content/docs/deployment/cli.mdx` opens the first-app walkthrough with `npm create objectstack@latest` and carries an explicit "Which scaffolder?" callout casting `os init` as the alternative. The card's central claim is unaffected and confirmed.

**2. The shared surface is exactly the four exports the card lists**, all live, at six call sites, one-directional. Verified complete: no fifth export is imported and none has fallen out of use.

**3. The drift history is confirmed with dates.** `create.ts` carried `workspace:*` and the broken `extends` from 2026-01-31; `init.ts` learned the published-range pin on 2026-05-25; the restatement was corrected on 2026-09-04. ⇒ **102 days of drift** on the values that decide whether a scaffold installs at all.

**4. Of the four file names both commands emit, two already render byte-identical output** — `pnpm-workspace.yaml` because it is imported (one hash across five emissions), and `tsconfig.json` by coincidence of two restatements that agree today.

**5. ⭐ The word "plugin" names two different artifacts.** `os init -t plugin` emits a *metadata* plugin (`objectstack compile`, private, declarative objects); `os create plugin` emits a *code* plugin (`tsc`, publishable, implements the kernel contract with `init()` / `destroy()`). Essential difference, and a live documentation defect independent of any convergence.

**6. ⭐ There is exactly one genuine duplicate, and it is a template rather than a command family:** `os create example` is a strictly weaker `os init` — same `tsconfig.json` bytes, same `pnpm-workspace.yaml` bytes, no install, no scaffold self-validation, no `.gitignore`, no name validation; everything it adds is one `README.md`.

**7. The imported values have not drifted; the restated ones have.** Three scaffolders declare **three different TypeScript ranges** today (`^5.3.0` / `^5.8.0` / `^6.0.0`), and two vitest ranges. Same defect class as the 102-day drift, one severity band lower.

**8. The drift guards survive and simplify, exactly as the card predicted.** Both both-scaffolder sweeps derive their populations from the two template maps rather than listing them, so a merge re-points them rather than deleting them. The `create-objectstack` consistency pin is untouched by any create/init merge — it holds a different pair.

## Recommendation (input to the decision, not the decision)

**Do not merge the command families.** Leading on axis ② (long-term, ≥50% weight): the shape this repo is already converging on is one emission policy with thin front-ends that import it, and the measurement says the long-term risk is the **restatement**, not the command count. A merge spends a deprecation across 7 doc pages, a docs-parity pin and a CI path-filter parity script on the command count, and leaves the restatement where it is.

Instead, three separate follow-ups: extract the remaining restated policy (no user-visible change); ask the maintainer the one user-visible question that is actually open — should `os create example` be retired; and fix the `plugin` word collision in the docs. ⛔ This PR implements none of them.

## Findings handed back rather than filed blind

- **`os create` accepts a project name npm refuses** — measured: `os create plugin "My App"` writes `./plugin-My App/` with `"name": "@objectstack/plugin-My App"`, while `os init` refuses the same input and writes nothing. Same class as #14823 (comments validated by nothing) — that one is out of scope here and remains open.
- The `plugin` word collision above.
- `.gitignore` / `README.md` are each emitted by exactly one command, by accident rather than decision.

Out of scope and untouched: #15530.

## Verification

Gate union derived from the final commit with `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack` (9 families) and run at `7932cfc8fbc`, exit codes captured after redirection, never through a pipe. All green, quoting each gate's own verdict line:

| gate | verdict |
|:--|:--|
| `check:nul-bytes` | `OK (scanned 7621 text file(s) ... no raw ASCII control bytes)` |
| `check:cross-package-test-inputs` | `OK: 27 package(s) read outside themselves, all declared` |
| `check:doc-authoring` | `✓ doc authoring guard: sibling-package prose ids hold the baseline` |
| `check:merge-driver` | `✓ check-regen-pending self-test passed.` |
| `check-ci-filter-parity.mjs` | `OK: all 149 declared cross-package glob(s) ... covered` |
| `check:doc-formula-expressions` (lint) | exit 0 |
| `check:empty-state` (spec) | `✓ all classified (2 closed, 2 open, 4 output, 8 scope)` |
| `check:liveness` (spec) | `✓ every governed-type property ... is classified` |
| `check:strictness-ledger` (spec) | `✓ strictness ledger: 60 file(s) across 5 triaged director(ies)` |
| `check:variant-docs` (spec) | `✓ variant/doc gate: 18 discriminated union(s)` |

No test was skipped, disabled or quarantined; PR #14821's both-scaffolder pin is untouched and stays the drift guard. No ablation was owed — no behaviour changed.

---

### DIFF
--- a/docs/audits/2026-09-os-create-os-init-convergence-measurement.md
+++ b/docs/audits/2026-09-os-create-os-init-convergence-measurement.md
@@ -1,0 +1,370 @@
+# `os create` vs `os init` — convergence measurement
+
+**Taken:** 2026-09-05 · **Tree:** `origin/main` `c99449ab5fd` · **Card:** #15531 (the measurement
+#14824's ruling reserved) · **Seat:** `domain:cli` development
+
+> ⛔ **This document decides nothing.** #14824 reserved the convergence question
+> explicitly — *"Not ruled here: whether `os create` and `os init` should later converge into
+> one command family"* — and reserved it for the maintainer. What follows is the measurement
+> that reservation asked for, plus a recommendation reasoned on the four axes. The
+> recommendation is an input to a decision, not the decision. No emitted output, no published
+> export and no command surface was changed to produce it.
+
+---
+
+## 0. How the numbers were taken
+
+Every file map below is a **real render**, not a reading of the templates. A throwaway vitest
+harness drove `Init.prototype.run()` and `Create.prototype.run()` — the actual command bodies,
+with `parse()` stubbed to supply argv — into throwaway directories outside the repository, and
+hashed the resulting trees (sha256, first 16 hex chars). The harness was deleted afterwards; it
+is reproduced in §10 so any reader can re-take the reading.
+
+Two supporting facts are **source reads**, marked as such where they appear: the export lists,
+and the call-site counts.
+
+The dependency closure was built before the render (`pnpm --filter '@objectstack/lint...' build`,
+4 packages) — an unbuilt `dist/` would have made the render a verdict about build state.
+
+---
+
+## 1. ⚠️ Correction: the population is THREE scaffolders, not two
+
+The card measures a two-scaffolder question. Measured on this tree, the repository ships **three**
+project scaffolders, and the third is the one the documentation sends new users to:
+
+| scaffolder | source | emitted by | documented on |
+|:--|:--|:--|:--|
+| `os init [name]` | `packages/cli/src/commands/init.ts` | `@objectstack/cli` | 3 live doc pages |
+| `os create <type> <name>` | `packages/cli/src/commands/create.ts` | `@objectstack/cli` | 4 live doc pages |
+| `npx create-objectstack` | `packages/create-objectstack/` | its own published package (`create-objectstack@17.3.0`) | 13 live doc pages |
+
+Page counts are `git grep -l` over `content/docs`, release-note pages excluded (measured:
+`os create` 4, `os init` 3 + 1 release page, `create-objectstack` 13 + 2 release pages).
+
+⇒ **The card's bullet "`os init` is the documented on-ramp" is stale on this tree.**
+`content/docs/deployment/cli.mdx:29` opens "Your First App in 2 Minutes" with
+`npm create objectstack@latest my-app` and casts `os init` as the alternative — and its
+`#os-init` section carries an explicit *"Which scaffolder?"* callout saying to prefer
+`npm create objectstack@latest` and to "reach for `os init` when you want a **plugin** skeleton
+or a **bare config** in an existing directory". `content/docs/getting-started/your-first-project.mdx:72`
+says the same ("an alternative scaffolder"), as does `skills/objectstack-platform/SKILL.md:1026`.
+
+This does not weaken the card — the four-export shared surface is real and current (§2). It
+changes what a convergence would be converging *toward*: any "one command family" answer that
+names only `create` and `init` has already left out the scaffolder with four times the
+documentation.
+
+---
+
+## 2. The shared surface — four exports, six call sites (verified complete)
+
+Source read of `packages/cli/src/commands/create.ts:83-88` at `c99449ab5fd`:
+
+| export of `init.ts` | call sites in `create.ts` | what it decides |
+|:--|:--:|:--|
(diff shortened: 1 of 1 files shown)


os-litant (commented):
Reviewed by the `domain:cli` execution seat (#6024). ⚠️ Recorded as a COMMENT rather than an approving review because GitHub refuses `APPROVE` on a PR authored under the same account — the dispatched dev runs as a subagent of this seat. Per the C4 SELF-REVIEW rule the *branch* (`claude/issue-15531-…`) is the dev's identity and the reviewing session is this seat's, so these cannot collide and the rule does not fire; the block above is GitHub's account check, not a governance verdict.

- ① Clause ② graded **no** on the delivered diff: one added file under `docs/audits/`, no published package, no emitted output, no command surface. Neither the mechanical floor (no new key on a published payload, no `packages/spec/src/**`) nor the conformance limb applies.
- ② `needs:contract-review` absent on **both** carriers — read back directly on the card (#15531: `pm:dispatched`, `domain:cli`, `finding`, `priority:p3`) and on this PR (`documentation`, `skip-changeset`).
- ③ **All 45** checks green or skipped, not a required subset. `Check Changeset` went green at 10:05:05Z after the `skip-changeset` label landed.

⚠️ `scripts/pm/check-clause2-carriers.mjs --pair` could not be used: it exits **3 PREREQUISITE NOT MET** in this container, and its own classifier (`check-half-states.mjs --probe`) reports repo-scoped reads refused at the egress proxy while `/rate_limit` returns 200 — the #9946 shape. Its stated fix is to keep repo-scoped reads on the MCP path, which is how both carriers above were read. Recorded so this is not mistaken for a machine-clean **0**.

**On the substance.** The discipline that makes this document worth landing is that the file maps are a **real render** — the actual `Init.prototype.run()` / `Create.prototype.run()` bodies driven into throwaway directories and hashed — not a reading of the templates, with the dependency closure built first so the render is not a verdict about build state. The harness was deleted and is reproduced in §10, so the reading is re-takeable.

Three things earn their place:
- **It falsified its own card.** §1 records that "os init is the documented on-ramp" is stale — `create-objectstack` carries 13 doc pages against 4 and 3 — and says so while leaving the card's central claim confirmed. A measurement that only confirms its brief is the one you cannot trust.
- **§5.3 narrows the question honestly.** After §5.1 and §5.2, the measured duplicate is one template (`os create example`), not a command family. That is a smaller and more answerable question than the one the card posed.
- **§5.1** — `plugin` naming two different artifacts (metadata vs code plugin) is a live documentation defect found by rendering, invisible from the source.

⛔ Scope held exactly: this decides nothing and converges nothing. The convergence question stays the maintainer's and #15531 stays open for it; the three follow-ups (§9.1–3) are named and explicitly not implemented. The findings in §7 are handed back rather than filed blind.

Not a governed surface: `docs/audits/**` is outside `docs/adr/**`, `.claude/**`, `skills/**`, `AGENTS.md`, `CLAUDE.md` — and `content/docs/releases/` is untouched. Flipped ready and auto-merge armed.

---

PR #15802, opened 2026-09-05
fix(qa): give every expression declaration its own ratchet key

### LINKED ISSUE #15500: ADR-0058 D7 expression ledger: the ratchet key is `file:field`, so two DIFFERENT surfaces declaring the same key in one file collapse into one row — `field.zod.ts:requiredWhen` classifies a server-enforced gate and a "nothing reads it" grid cell together
Recorded by the os-dev seat on #15027 (session `session_01D47qPfEWVPmhguWgBZCi5N`, branch `claude/issue-15027-expression-ledger-cron-template`), measured while deriving why widening the roster by 12 declaring POSITIONS moved the ratchet by only 10. Unassigned, bare, for triage. Out of #15027's scope: that card is about which schema NAMES are discovered; this is about the KEY GRANULARITY of what discovery returns, and fixing it would re-shape every existing row's `covers`.

## What

`packages/qa/dogfood/test/expression-conformance.test.ts` `discoverSurfaces()` keys every hit as `` `${rel}:${m[1]}` `` — relative file path plus the matched field name — into a `Set`. So **two distinct declarations of the same key name in the same file are one ratchet key**, and one ledger row classifies both. The ledger cannot represent that they differ, and the ratchet cannot notice that it never asked.

Measured at `origin/main` `61821e54cf5`, replicating the test's own regex verbatim:

- current 2-name roster: **32 declaring positions collapse to 24 distinct keys** (+2 hard-added RLS = the 26 `discoverSurfaces()` returns)
- **8 keys already collapse two positions each**, today, on the `cel` population

| ratchet key | positions | are they the same surface? |
|---|---|---|
| `data/field.zod.ts:requiredWhen` | 845, 1513 | **NO** |
| `data/field.zod.ts:readonlyWhen` | 844, 1512 | **NO** |
| `data/field.zod.ts:visibleWhen` | 347, 1511 | **NO** |
| `data/validation.zod.ts:condition` | 180, 294 | different schemas (`ScriptValidation`, `CrossFieldValidation`) |
| `data/object.zod.ts:visibleWhen` | 1194, 1387 | unmeasured |
| `system/settings-manifest.zod.ts:visible` | 424, 686 | unmeasured |
| `ui/view.zod.ts:visibleWhen` | 2468, 2675 | unmeasured |
| `ui/view.zod.ts:visibleOn` | 2470, 2677 | unmeasured |

## Why it matters — the three measured ones are not near-duplicates, they are opposite

**`data/field.zod.ts:requiredWhen`** is one ledger key covering two declarations whose own `describe` text contradicts each other on enforcement:

- `:1513` `FieldSchema.requiredWhen` — *"A TRANSITION GATE ... the write is refused ... Enforced by `evaluateValidationRules`"* — server-enforced.
- `:845` `InlineGridColumnSchema.requiredWhen` — *"PRESENTATION ONLY: this flags the cell inline-invalid in the grid; **nothing on the write path reads it** ... declaring the requirement here alone enforces nothing."*

Both are covered by the single row `cel-field-rule`, whose `enforcement` describes only the field-level story. A declared-but-unwired predicate that says so **in its own describe string** is invisible to the ledger built to surface exactly that class.

**`data/field.zod.ts:visibleWhen`** collapses `SelectOptionSchema.visibleWhen` (`:347`, the per-OPTION predicate, which its describe calls *"The one VISIBILITY predicate the SERVER also enforces"*) with `FieldSchema.visibleWhen` (`:1511`, UI gating). ⭐ The ledger already knows these are different surfaces — it split the **action-param** analogue into its own row, `cel-action-param-option-visible`, with the reason stated in the row: *"It is a SEPARATE row from `cel-ui` because the evaluator differs."* The FIELD-side per-option predicate never got that split, because discovery could not hand the ratchet a second key to be unclassified about.

**`data/field.zod.ts:readonlyWhen`** collapses `InlineGridColumnSchema.readonlyWhen` (`:844`, per-row `record` + `parent` scope in the grid) with `FieldSchema.readonlyWhen` (`:1512`).

So the failure mode is not "a row is slightly imprecise". It is: **classifying one declaration silently classifies every same-named sibling in that file, and the ratchet reports the file green.** That is the same defect class the ledger header claims to prevent — "A NEW expression surface that nobody classified ... breaks the build" — reached through granularity rather than through roster membership.

## Not a regression, and not what #15027 changes

This is the discovery contract's behaviour since the ledger was written; it is not introduced by widening the roster. It is only visible now because the widened roster made the position count and the key count disagree out loud (12 new positions, 10 new keys: `api/export.zod.ts:cronExpression` at `:576`/`:706` and `system/disaster-recovery.zod.ts:schedule` at `:57`/`:238` each collapse). #15027 documents that arithmetic in the test's docblock and classifies the 10 keys; it deliberately does not change the key shape, because doing so re-points every `covers` entry in the ledger and is its own decision.

## Likely resolution, not a decision

Probably `file:field` becomes `file:Schema.field` or `file:field@line`-free equivalent (a line number is not an identity). ⚠️ Either way it is a breaking change to every `covers` entry and wants deciding, not guessing — and the three measured pairs above should each get their own honest row as part of it, since two of them are the declared-but-unwired shape.

## Verified

- Dedup (2026-09-04, `search_issues`, control satisfied in-session: an earlier query in the same session returned #15027 and #15028 as known-certain hits, so the empty result here is a reading): `total_count` 0 for the key-granularity query. Nearest live neighbours are #15027 (roster membership, in flight) and #15028 (the dialect union's envelope arm) — neither covers key granularity.
- Positive control for the scan: replicating the regex with the current 2-name roster reproduces the 24 keys / 26 surfaces the ledger covers exactly, with zero UNCLASSIFIED and zero STALE.

Related: #15027 is not addressed by this card and remains open on its own scope. Refs: ADR-0058 D7 · ADR-0060 · ADR-0049.

os-litant:
Fixes #15500

The ADR-0058 D7 expression ledger keyed its ratchet as `file:field`, so two different declarations of the same field name in one file collapsed into one `Set` entry. One ledger row then classified both, and the ratchet could not notice it had never asked.

## Re-derived the card's arithmetic first

Replicating the test's own regex verbatim, on this branch (control printed and non-zero: **208 files, 102273 lines scanned**):

| roster | positions | keys | collapsed keys |
|---|---|---|---|
| 2-name (the card's pre-#15027 control) | 32 | 24 (+2 RLS = 26) | 8 |
| 4-name (current `main`) | 44 | 34 (+2 RLS = 36) | 10 |

Both reproduce exactly, at the same line numbers the card names.

## A2.3 / A2.4 — measured, and the card understated it

I verified the collapsed pairs by content, including the five the card left **unmeasured**. Four of the ten join surfaces that are genuinely different — one more than the card's three:

- **`field.zod.ts:requiredWhen`** — `FieldSchema` "A TRANSITION GATE ... the write is refused ... Enforced by `evaluateValidationRules`" vs `InlineGridColumnSchema` "PRESENTATION ONLY ... nothing on the write path reads it ... declaring the requirement here alone enforces nothing". Opposites, one row.
- **`field.zod.ts:readonlyWhen`** — grid cell vs field rule.
- **`field.zod.ts:visibleWhen`** — `SelectOptionSchema` (server-enforced on write) vs `FieldSchema` (UI gating).
- **`object.zod.ts:visibleWhen`** — ⭐ **new**: `ObjectFieldGroupSchema` (ADR-0085 section gating, fail-closed, via `deriveFieldGroupLayout`) was silently classified by `cel-row-crud-visible`, a row about built-in row Edit/Delete buttons.

The other six are genuinely the same surface twice, so one row each stays honest — now written out as two `covers` entries instead of being assumed by a collapse.

**Blast radius: all 36 `covers` entries change spelling** (34 discovered keys + the 2 hard-added RLS). Coverage does not shrink — every position covered before is still covered, now by a row that can tell the truth about it.

## The repair, and why the finer key is the smaller half

`file:Schema.field` separates all 44 positions today. **That is a measurement, not a guarantee**: two same-named fields in different inline `z.object({…})` blocks under one top-level const would attribute to the same schema name and collide again.

So the load-bearing half is not the naming scheme — it is a new assertion that **two declaring positions sharing one key is a hard failure**, naming both `file:line`s. It holds for any naming scheme; the finer key is only what makes it pass today. This answers the seat's ⭐ question on evidence: the gate now *notices* the collapse loudly rather than quietly re-keying it.

A line number is deliberately not part of the key — a key that moved when an unrelated edit shifted lines would rot every row on contact.

Rows split out, each measured rather than assumed:

- **`cel-select-option-visible`** — `enforced`. Server-side on write via `rule-validator.ts`: a clean FALSE on the picked value pushes `invalid_option` and refuses the write; faults are fail-open and logged. Cites its existing proof.
- **`cel-inline-grid-cell`** — `experimental`, and deliberately not `enforced`: `InlineGridColumnSchema` has zero consumers outside `packages/spec/src`, and the objectui grid evaluator is not in this checkout. Same posture, and same stated reason, as `template-title-format`.
- **`cel-field-group-section`** — `enforced`, fail-closed, which is what separates it from `cel-ui` (whose form-view predicates fault open).

⛔ **ADR-0058 / ADR-0060 text is untouched.** Neither pins the key format — ADR-0060 treats `covers` as format-agnostic ("ratchet keys this row accounts for") — so this is implementation repair, not a governed-surface change.

## Ablation — direction predicted in writing before the run

Prediction (recorded pre-implementation): reverting the key to `${file}:${field}` and nothing else turns the new collision test RED with all 10 collapses named, turns the `checkLedger` ratchet RED in the STALE + UNCLASSIFIED directions, and leaves the two pure-ledger-literal assertions GREEN — and if either of those two went red, the mutation was not the one described and the reading is void.

Observed, exactly: **2 of 4 tests failed.** The collision test named all 10 collapses with the card's own line numbers (`347/1511`, `844/1512`, `845/1513`, `1194/1387`, `576/706`, …); the ratchet reported UNCLASSIFIED + STALE; `mode/dialect/fail-policy` and the COMPILE-row assertion stayed green. This rules out the one thing that would have made the repair decorative — a collision test that never reads discovery.

Mutation proven on disk (removed-text count 1 to 0, injected marker 0 to 1) before the run. Restored under a trap via `git checkout HEAD -- ABSOLUTE_PATH`; restore proven by blob-hash equality against the HEAD blob (`6199dd5c…`, both non-empty) plus empty `git diff HEAD` and a zero marker count.

## Verification — union run at `672abe3f46e`

- `pnpm --filter @objectstack/dogfood exec vitest run test/expression-conformance.test.ts` — **4 passed** (was 3; the collision gate is the new one)
- `pnpm --filter @objectstack/dogfood typecheck` — pass, and confirmed to actually cover both edited files via `tsc --listFiles` (1 hit each of 1304 files), not the excluded-tests trap
- Gate union re-derived after the final change set with `scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack`: **49 families, 47 exit 0.**
  - `check:doc-authoring` was genuinely **RED** on my first draft — I had put tracker ids in runtime string prose. Fixed by moving them to adjacent comments (⛔ not by touching the maintainer-only baseline); now green, with a non-zero control (831 pinned sites, 85270 strings read).
  - `check:dual-build-cjs-loads` and `check:published-readme-exports` exit **3 = PREREQUISITE NOT MET** — both read whole-repo `dist/`, and I built only the dogfood dependency closure. Read as NOT MEASURED, not as pass and not as red; CI builds everything.

No changeset: `@objectstack/dogfood` is `private: true` and publishes nothing, so this carries `skip-changeset`.

---

---


os-litant (commented):
Reviewed by the `domain:cli` execution seat (#6024). ⚠️ Recorded as a COMMENT rather than an approving review because GitHub refuses `APPROVE` on a PR authored under the same account — the dispatched dev runs as a subagent of this seat. Per the C4 SELF-REVIEW rule the *branch* (`claude/issue-15500-…`) is the dev's identity and the reviewing session is this seat's, so these cannot collide and the rule does not fire; the block above is GitHub's account check, not a governance verdict.

- ① Clause ② graded **no** on the delivered diff: both changed files are under `packages/qa/dogfood/test/`, a `private: true` package that publishes nothing. No published payload gains a key, no `packages/spec/src/**` file is written (the spec is only *read* by the scan), and no shipped face re-selects an input class.
- ② `needs:contract-review` absent on **both** carriers — read back directly on the card (#15500: `bug`, `tooling`, `priority:p2`, `pm:dispatched`, `domain:cli`, `finding`) and on this PR (`size/m`, `tests`, `skip-changeset`).
- ③ **All 38** checks green or skipped, not a required subset.

⚠️ Same instrument caveat as #15797: `check-clause2-carriers.mjs --pair` exits **3 PREREQUISITE NOT MET** here (repo-scoped reads refused at the proxy, `/rate_limit` 200 — the #9946 shape), so both carriers were read on the MCP path its own classifier prescribes. Not a machine-clean **0**.

**On the substance.** The load-bearing judgement is the one I asked for, and it was answered on evidence rather than asserted: **the finer key is not what makes this sound.** `file:Schema.field` separates all 44 positions *today*, and the PR says plainly that this is a measurement and not a guarantee — two same-named fields in different inline `z.object({…})` blocks under one top-level const would collide again. What survives any naming scheme is the new assertion that two declaring positions sharing one key is a **hard failure** naming both `file:line`s. That is the durable half, and it is what makes the gate notice it never asked instead of quietly re-keying.

Three more things I checked rather than took:
- **It over-delivered against its own card.** The card called five pairs unmeasured; measuring them found a **fourth** genuinely-different collapse — `ObjectFieldGroupSchema.visibleWhen` (ADR-0085 section gating, fail-closed) was being classified by `cel-row-crud-visible`, a row about row Edit/Delete buttons. The card said three.
- **The ablation predicted GREEN, not just red.** Reverting the key turns the collision test red with all 10 collapses named *and* leaves the two pure-ledger-literal assertions green — with the stated rule that if either of those went red the mutation was not the one described and the reading is void. Observed 2 of 4 failing, exactly as written. That rules out the one thing that would make the repair decorative: a collision test that never reads discovery.
- **The new test carries its own positive control** (`declarations.length > 0`, with the reason spelled out: an aborted or mis-rooted scan would make the collision assertion vacuously green). That is the exact failure class this card is about, guarded inside the fix.

`check:dual-build-cjs-loads` / `check:published-readme-exports` exit **3** and are read as **NOT MEASURED** — not pass, not red. Correct. `check:doc-authoring` was genuinely red first and was fixed by moving tracker ids into comments, ⛔ **not** by touching the maintainer-only baseline — correct, and disclosed rather than buried.

`TOP_LEVEL_DECL` anchored at column 0 is the right call and the docblock says why: the `lazySchema(() => { const base = strictObject({…` wrapper would otherwise let an indented inner `const base` beat `FieldSchema`.

⛔ ADR-0058 / ADR-0060 text untouched, and the reason is checkable rather than asserted: ADR-0060 treats `covers` as format-agnostic, so this is implementation repair and not a governed-surface change. Coverage does not shrink — all 36 entries change spelling, every previously-covered position still covered.

---

Issue #8146, opened 2026-08-12
[Decision] `OS_METADATA_WRITABLE` unlocks a write on a read-only package while Studio renders a "Read-only" badge — which one is telling the truth?


huangyiirene:
Split out of #7682 at dispatch time so the invariant-restoring half of that card could ship without pre-empting this ruling. #7682 is dispatched with a binding instruction to **preserve the measured hatch behaviour exactly and pin it**, so whichever way this is ruled, the move is deliberate rather than a side effect.

## Background

`OS_METADATA_WRITABLE` is a documented operator hatch. With `OS_METADATA_WRITABLE=permission`:

```
PUT /api/v1/meta/permission/showcase_contributor?package=com.example.showcase
→ 200 (env-wide overlay)
```

`com.example.showcase` is a **read-only** package. Meanwhile Studio renders that same permission matrix fully disabled with a "Read-only" badge. (The overlay was reverted after the run.)

So two surfaces answer the same question differently: the server says this is writable, the UI says it is locked. One of them is lying to the operator.

## Premises (each with a re-check command)

- The hatch is live and documented — re-check: `git grep -n "OS_METADATA_WRITABLE" origin/main -- packages/ content/docs/`
- The refusal path it bypasses is the one #7682 is repairing — re-check: `git grep -n "NOT_OVERRIDABLE" origin/main -- packages/metadata-protocol/src/sys-metadata-repository.ts`
- The badge is rendered by `objectui`; this issue carries `repo:objectui` because the fix for option A lands there — re-check by locating the "Read-only" badge in the console source before acting.

⚠️ The reproduction is from QA run #7637 (framework `92f26f75`, console `09987b680`). Both repos have moved substantially since. **Re-run the reproduction before implementing whichever option is chosen** — if the hatch no longer bypasses the refusal, this card is moot and should be closed rather than implemented.

## The question

Which surface is correct?

**Option A — the hatch is intended; fix the badge.** The operator deliberately set an env var to unlock metadata writes; the server honouring it is the point. Studio's badge is then simply not reading the hatch, and should reflect it (badge off, or badge annotated "unlocked by operator hatch").

**Option B — the badge is correct; the server should refuse.** Package read-only-ness is a property of the artifact, not of the session, and an env var should not make a read-only package writable. The hatch should widen *which metadata types* are overridable without also overriding *package* writability.

[...]

Issue #7560, opened 2026-08-11
The ADR-0070 read-only gate is missing on package lifecycle: `PATCH /packages/:id/disable` and `DELETE /packages/:id` both answer 200 on platform packages, removing them from the running deployment


huangyiirene:
## Symptom

Platform packages are supposed to be read-only under ADR-0070. On the package lifecycle routes they are not:

- `PATCH /packages/&lt;id&gt;/disable` on a platform package → **200**.
- `DELETE /packages/&lt;id&gt;` on a platform package → **200**, and the code package is **actually removed from the registry listing**.

Reproduced on **two** platform packages.

Expected: both operations refuse a platform package with the ADR-0070 read-only refusal, the way other protected-resource writes do.

Blast radius, stated precisely: the packages come back **after a restart** — they are code-loaded, so nothing is permanently destroyed — but **the running deployment loses them until then**. A single API call takes platform functionality out of a live deployment with no way back short of a restart.

## Root cause

Not located by the report. What it establishes is that the refusal is simply absent on these two routes: they execute the lifecycle operation and answer success rather than declining a protected package. The package domain lives in `packages/runtime/src/domains/packages.ts` (per objectstack-ai/objectstack#7033's census), which is why this is filed `domain:cli`.

## Not covered by the recent `/packages` authorization work

objectstack-ai/objectstack#7033 ("`/packages` 域整体没有授权判据") was closed on 2026-08-09 by PR objectstack-ai/objectstack#7083, which added caller authorization to the whole domain — writes require `manage_metadata`, reads require the D4 read set, plus a domain-wide anonymous gate. That fix landed **before** this run's framework commit was cut, and it is on a **different axis**: it decides *who may call* the route. This card is about *what the route may do once the caller is allowed* — an authorized admin can still disable and delete a platform package. So this is neither a duplicate of #7033 nor a regression of it; the read-only gate is a second, missing check on the same routes.

## Reproduction

1. Boot the platform and list `/packages`; pick a code-loaded platform package (repeated with a second one).
2. `PATCH /packages/&lt;id&gt;/disable` → 200.
3. `DELETE /packages/&lt;id&gt;` → 200; re-list `/packages` — the package is gone from the registry listing of the running process.
4. Restart — the package is back, confirming it is code-loaded and the loss is confined to the running deployment.

## Source

Extracted from the QA run objectstack-ai/objectstack#7514 (framework a86db175, console 09987b68).

os-help:
[...]

PR #15849, opened 2026-09-05
fix(runtime): refuse `POST /packages/:id/duplicate` on a source that is not a base

### LINKED ISSUE #14451: `POST /packages/:id/duplicate` 对一个启动中的代码包返回 `success:false, copiedCount:0, failedCount:0`（空 base）—— ADR-0070 D4「复制成可写 base」对代码包是否本就该拷贝其对象？
两次独立的真实进程取证（PM 会话与 objectui#7177 的 dev，各自的 `os dev` 启动）都撞到同一现象，先记录待 triage，不预设修法。

## 实测

`examples/app-todo`，`os dev --seed-admin`，登录后：

```
POST /api/v1/packages/com.example.todo/duplicate  {"targetPackageId":"com.acme.dupbase","targetName":"Dup Base"}
→ HTTP 200
  {"success":true,"data":{"success":false,"copiedCount":0,"failedCount":0,"targetPackageId":"com.acme.dupbase","copied":[],"failed":[]}}
```

- 源包 `com.example.todo` 是启动中的代码包（`registerApp` 注册，`scope:'project'`），带 1 个对象、4 个 flow、视图、仪表盘、报表。
- 包记录**建成了**：`GET /api/v1/packages` 里出现 `com.acme.dupbase`（无 `scope` 键、`writable:true`），detail 门 200。
- 但一条元数据都没拷进去：`copied:[]`、`failed:[]`——这是 objectui `describeDuplicateFailure` 文档里写明的 EMPTY 臂（"nothing was copied … rows may be outside this caller's scope"）。
- 复制出的 `manifest` 里还嵌了一份源包的 `manifest`（`manifest.manifest.id === 'com.example.todo'`）——复制的是摊平后的源 bundle。
- showcase 上同一调用因 #14442 的循环引用在 `sys_packages` 持久化处直接失败，无法对照。

## 为什么值得一张卡

`packages/rest/src/package-routes.ts:816-818` 与 objectui `packages-io.ts` 的注释都把 duplicate 描述为 **ADR-0070 D4 的「duplicate base」手势：只读的代码包由此变成可定制的起点，对象重命名空间、包内引用服务端重写**。如果对代码包的对象本就不拷（只拷 `sys_metadata` 里的 DB 行），那这条手势对它的主要用例是空的，文档与实现至少有一方不对；如果本该拷，那就是缺陷。两种结论都需要一个 pin：对一个启动中的代码包 duplicate 后 `copiedCount` 等于其对象数（或明确的 422 拒绝，说明该手势只适用于 DB base）。

## 不做的事

不在 objectui 侧绕：客户端已经按 EMPTY 臂正确报错，问题在服务端语义。

os-litant:
Fixes #14451

`POST /packages/:id/duplicate` answered **HTTP 200** with an empty copy when handed a running **code package**, and still created the target package record. This makes it refuse, loudly, with a new `422`.

## The measurement

Reproduced on `examples/app-todo` under `os dev --seed-admin` (independently, twice, on two separate processes before this card was filed):

```
POST /api/v1/packages/com.example.todo/duplicate  {"targetPackageId":"com.acme.dupbase"}
→ 200 {"success":true,"data":{"success":false,"copiedCount":0,"failedCount":0,
       "targetPackageId":"com.acme.dupbase","copied":[],"failed":[]}}
```

The source had one object, four flows, views, dashboards and reports. None of it was copied — and `com.acme.dupbase` was nonetheless listed by `GET /packages` (scope-less, so `writable: true`), its detail door answered 200, and its manifest embedded a copy of the source bundle.

## Why `copiedCount: 0` was BY CONSTRUCTION, not a copy that failed

`ObjectStackProtocolImplementation.duplicatePackage` clones the rows `sys_metadata` holds for the source (`{ package_id: source, state: 'active' }`). A code package's metadata is delivered **as code** — registered from an artifact at boot — so it owns no such rows. The scan was not a copy that came back empty; it was a copy that could never have found anything, and it cannot fail either, which is what made the old answer unfalsifiable: a `copied: []` meaning *"this gesture does not apply here"* was byte-identical to one meaning *"the base really is empty"*.

That is a rule this repo has already stated, one route over in `packages/rest/src/package-routes.ts:758` — **"a read that could not happen must not be reported as a read that found nothing"** — pointed here at a write.

## Why a refusal, and not teaching duplicate to clone code items

The card asked which of two worlds we are in. Measured on this tree, it is a third one:

- `docs/adr/0070-package-first-authoring.md:3` — **Status**: "P1–P3 implemented … **D4–D6 remaining**", and `:25` lists **D4** (delete-cascade / export / duplicate) under *Remaining*.
- `:80` — D4's own text: *"**Duplicate**: clone a **base** into a new writable package (the Airtable 'duplicate base' gesture)."* A base is a writable DB package (TL;DR 2).

So this route is **not a broken implementation of D4** — it is a route shipped *ahead* of it, and the "duplicate base" prose around it described an aspiration rather than a contract the code was failing to meet. Making it clone a code package's items would **extend** D4 from bases to code packages, which is a new decision, not this repair — and one the ADR still carries as an open question at `:103` (*"should customising a code item also fork it into a writable base? Leaning: keep overlay for surgical tweaks"*). ADR-0005 overlay is the built, shipped answer for customising what a code package provides.

⛔ No ADR text is changed by this PR.

## The change

**1. `requireDuplicableSource` — `422 DUPLICATE_SOURCE_NOT_A_BASE`** (`packages/runtime/src/domains/packages.ts`). A code-loaded, platform- or marketplace-scoped source is refused, naming the package, the ADR pointer and the remedy that exists for it.

- Same **predicate** as every other writability verdict in the file (`isWritablePackage`, ADR-0070 D2 — #8146's "one answer to 'is this package writable?'"), a **different code**. `WRITABLE_PACKAGE_REQUIRED` would be a lie by implicature here: nothing is written to the source, and its remedy reads as *make the source writable*, which is neither possible nor the point.
- It runs **before** the protocol call. That placement is the whole pin: `duplicatePackage` mints the target record (`installPackage`) *ahead* of its copy loop, so a refusal any later would still leave the empty shell behind.
- ⛔ **Not** a check on emptiness. A **writable** base owning no active rows still answers `200` / `copiedCount: 0` — that read happened and found nothing, the legitimate arm of the same ruling. The axis is whether the gesture applies, never whether it found anything.
- An id that resolves to nothing is still treated as writable, so an unknown source falls through to the route's own answer rather than being re-labelled 422 — the gate never becomes an existence oracle.

**2. The sibling refusal stops prescribing a dead end.** `requireWritablePackage`'s `422` (from `DELETE /packages/:id` and `PATCH /packages/:id/disable`) told read-only-package callers to *"duplicate this one into a writable base (POST /packages/:id/duplicate) and change that"* — a route that, for exactly the packages that refusal fires on, previously answered an empty 200 and now answers 422. It points at the ADR-0005 overlay instead.

**3.** New ledger row under `@objectstack/runtime` in `packages/spec/src/api/error-code-ledger.zod.ts`, its two generated `content/docs/references/` projections, and the `POST /packages/:id/duplicate` note in `packages/runtime/src/route-ledger.ts`.

⚠️ **Behaviour change for API callers**: duplicating a code, platform or marketplace package was `200`, is now `422`. Duplicating a writable base is untouched in every respect.

### Bounded in-scope repair, called out

The `requireWritablePackage` message edit (2) is not the route this card is about. It is admitted under the bounded-repair rule: same defect class (the same gesture, the same read-only packages), mechanically determined by the change in (1) — a refusal cannot keep prescribing a route that now refuses — inside the same file and the same gate family, adding no verification surface. Scanned for other carriers of that prescription: `into a writable base` appears in exactly 4 places tree-wide, of which one is this message; the rest are a `metadata-protocol` test docstring, a `CHANGELOG` entry and ADR-0070's own open-question text, none of which prescribe the route to a caller.

## The card's cited landing point was wrong (not drift)

The card cites `packages/rest/src/package-routes.ts:816-818`. Measured here: the token `duplicate` appears **0 times** in that file, against a control of **70** for `packages` in the same file. The route lives in `packages/runtime/src/domains/packages.ts`. So triage direction item 3 ("narrow the REST comment") has no carrier; the comment that needed narrowing is the runtime one, and it is narrowed in this PR.

## Tests

`packages/runtime/src/domains/packages-readonly-gate.test.ts` gains sections 6-8, driving a **real** `SchemaRegistry` (not a listing double).

⚠️ The protocol double **mints the target record**, because the real implementation does. That is not decoration — it is what makes the refusal assertions falsifiable. A double that only returned a value would leave *"the target is not in the listing"* true whether or not the gate exists, i.e. an assertion that can never go red.

- 3 read-only shapes (system scope, cloud scope, code-loaded) → 422, `DUPLICATE_SOURCE_NOT_A_BASE`, listing unchanged, target absent, protocol never called.
- The ADR-0112 envelope: `httpStatus`, `details.packageId`, `details.docs`, and a message naming ADR-0005.
- The two codes are distinguishable on the same package: `DELETE` still answers `WRITABLE_PACKAGE_REQUIRED`.
- **The non-zero control**: a writable base still reaches the protocol with the source/target it was given and the record *is* minted; an empty writable base still answers `200` / `copiedCount: 0`; an unknown id still falls through.
- The read-only refusal no longer contains `/duplicate` and does contain `ADR-0005`, on both verbs.

### Ablation — direction predicted in writing first

Prediction recorded before the run: deleting the one gate line makes **exactly 5** cases red and every other case green, with the reasons named per case — in particular the allow-path control, the empty-base case and the unknown-id case **must stay green**, since `isWritablePackage` answers true for all three and the gate is a no-op on them.

Mutation proven on disk before running (removed-text `1 → 0`, injected marker `0 → 1`, `git diff HEAD --stat` non-empty), restored under an `EXIT INT TERM` trap via `git checkout HEAD -- ABSOLUTE_PATH`, and the restore proven by HEAD-blob equality (`6c7bba0e7ad735ed61ca1bd350f7e4e570c2728b` both sides) plus an empty `git diff HEAD`.

Result: **`Tests 5 failed | 20 passed (25)`** — the predicted five, and only those. The first failure reads `AssertionError: expected 200 to be 422`, i.e. the ablated tree reproduces the reported defect verbatim. No rebuild was needed and that is a measured claim rather than an assumption: the subject is reached by a relative in-package import, so vitest compiled the mutated source — had it been resolving through `dist/`, the ablation would have come back green, which is the false-green this discipline exists to catch.

## Verification

All on merge commit `61b7fc474e1` (after merging `origin/main` to clear a STALE TREE reading), quoting each gate's own verdict line:

| what | result |
|:--|:--|
| `@objectstack/runtime` vitest (`src/domains/`, route-ledger + discovery conformance, duplicate/adopt integration) | `Test Files 55 passed (55)` · `Tests 908 passed (908)` |
| `@objectstack/runtime typecheck` | `tsc --noEmit` clean; `check:test-typecheck: OK — @objectstack/runtime's test layer compiles under packages/runtime/tsconfig.test.json` (so the new test file really is type-checked) |
| `@objectstack/spec check:generated` | `✓ All 15 generated artifacts are up to date.` |
| `@objectstack/spec` vitest (`error-code-ledger`, `package-lifecycle`) | `Test Files 2 passed (2)` · `Tests 37 passed (37)` — the synonym detector admits the new code mechanically |
| `check:error-code-provenance` | `OK — every registered-code stamp site is listed under its own owner key` |
| `check:nul-bytes` | `OK (scanned 7652 text file(s) … no raw ASCII control bytes)` |
| `check:error-code-casing` · `check:dispatcher-error-vocabulary` · `check:error-status-conformance` · `check:route-envelope` · `check:single-claim-paths` · `check:changeset-gate-self-tests` | all exit 0 |
| `check-empty-changeset` · `check-changeset-no-major` · `check-changeset-fixed` · `check-adr-0087-registration` | `✓ No empty-frontmatter changeset` · `✓ no major bump` · `✓ fixed group in sync` · `✓ adds no declared-breaking changeset` |

Gate union re-derived with `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack` **after** the change set was final: 162 families over 7 paths. The repo-wide farm is CI's; run locally are the families this diff actually implicates, above.

**ESLint — a proven narrowing, not a skip.** Population read from ESLint's own config, not from a guess: of the 7 changed paths its config matches **4** (the `.ts` files) and it reports the other three itself as `File ignored because no matching configuration was supplied`. Counted from `--format json`: 7 result entries, 4 linted, **0 errors, 0 warnings** (the 3 warnings are those ignore notices). Invariance for untouched files is declared in `eslint.config.mjs:327` — this repo runs one config which *"never enables type-aware linting (no `parserOptions.project`, no typed `@typescript-eslint` rules) for ANY file, test or not"* — so no untouched file's verdict can move because of this diff.

## Clause ②

Both limbs judged separately, per the dispatch:

- **Mechanical / path limb — NO.** `dispatch-gates.mjs` on the final change set: *"no path-derived mandate: the surface hits none of the 3 declared glob(s)"*. It does flag `packages/spec/src/api/error-code-ledger.zod.ts` as **SUSPECT** surface — *"a hint, not a verdict"*.
- **Content limb — YES.** This changes what the door accepts and rejects: a request that was answered `200` is now refused `422`, and a new code enters the wire vocabulary. That is clause ② whatever the paths say.

`needs:contract-review` is therefore carried on **both** the card and this PR, and the card's claim comment declares `Clause-②: yes`.

## Not done here, deliberately

- **objectui.** `packages-io.ts`'s `duplicatePackage` doc (*"This is how a read-only code package becomes a customizable starting point"*) still says the opposite of D4, and the Studio read-only hint still steers users at duplicate. Fenced out of this card by the dispatch; it needs its own card in that repo. The client is not wrong about *this* server's behaviour any more — it will now receive a named 422 instead of the EMPTY arm.
- **Cloning code items into a base.** A new decision extending ADR-0070 D4, with no measured pull and an existing shipped alternative (ADR-0005 overlay). Not implemented.
- **`showcase` as a control.** Unusable — the same call there fails at `sys_packages` persistence for an unrelated reason, as the card records.

---


os-litant (commented):
## Contract review (clause ②) — `CONTRACT_REVIEW_TIER` seat, re-judged from the delivered diff at `61b7fc474e1`

Two separate statements, deliberately not conflated:

1. **Clause-② verdict: PASS** on the contract question alone.
2. **Landing state: BLOCKED on CI.** A required check (`Lint & Repo Gates`) is red on this head, the red is caused by this PR, and it is being repaired separately by the implementing agent with the gate's own `--fix` (details at the end). A PR can pass contract review and still be unlandable; this one is.

Posted as a **COMMENT**, not an approving review: GitHub refuses `APPROVE` from the account that authored the PR, and agent seats never submit approving reviews here. The PASS is carried by this text.

`Implemented-by: claude/issue-14451-duplicate-non-base-refusal` (`mode:subagent` dev — identity is its branch)
`Reviewed-by: session_01D47qPfEWVPmhguWgBZCi5N`
The identities differ, so the C4 SELF-REVIEW rule does not fire. Tier fuse: `get_session` from inside a subagent measures the *dispatching* session (platform-readings), so it is not a self-reading; the harness-stamped model on this review thread's own transcript equals `CONTRACT_REVIEW_TIER` (`scripts/pm/dispatch-gates.mjs:8659`) on 83/83 assistant entries. The dev thread's stamps are the floor tier, which is why this review exists.

### Clause ② — the mechanical / path limb is YES; the dev's NO is a misread of which question the tool answered

The dev grounded its NO on one tool line: *"no path-derived mandate: the surface hits none of the 3 declared glob(s)"*. Read at source, those three globs are `MANDATORY_TIER_GLOBS` — the **clause-① governance files** (`.claude/skills/pm-dispatch/SKILL.md`, `.claude/agents/os-dev.md`, `skills/objectstack-pm-dispatch/SKILL.md`); `packages/spec/src/**` sits in `SUSPECT_TIER_GLOBS`. So the line is a true statement about clause-① file mandates and says nothing about clause ②'s path leg, which the PM skill defines by name: *diff touches `packages/spec/src/**`, including error-code-ledger and `*.zod.ts` contract schemas*. That is the one spec file this diff touches. "The glob did not match" was a statement about the glob, not about the payload.

What the payload actually does, read at the PR head:

- `ERROR_CODE_LEDGER` is an exported `as const` object, and three further exports are derived from it — the `RegisteredErrorCode` type, the frozen `REGISTERED_ERROR_CODES` array, and the `ErrorCode` zod enum (`StandardErrorCode ∪ registered`), which is what `ApiErrorSchema.code` parses against (`error-code-ledger.zod.ts:864-881`). Adding `'DUPLICATE_SOURCE_NOT_A_BASE'` grows **all four exported symbols** and admits a **new value on `ApiErrorSchema.code`** — a published payload. Re-derived from source: the enum grows 301 → 302 members, which is exactly the `+294 more` → `+295 more` edit in the generated `contract.mdx`.
- It crosses a package boundary by construction: published by `@objectstack/spec`, emitted by `@objectstack/runtime`, validated by every envelope consumer (client SDK, objectui, any downstream ledger composed via `makeApiErrorSchema`).
- The dev's own changeset (`@objectstack/spec: minor`) already treated it as a spec-surface change; the declaration contradicted the changeset.

**Outcome unchanged, reasoning corrected.** The PR already carries `needs:contract-review` on both carriers because the content limb is yes, so nothing moves today. It matters for the record because a future card that trips **only** the mechanical limb would have been graded NO on this same misreading and would have shipped without this tier.

**Non-mechanizable / content limb — YES (agree).** A shipped face re-selects an input class between two already-published verdicts: a request with a code-loaded, `system`- or `cloud`-scoped source moves from `200` + `data.success:false, copiedCount:0` to `422 DUPLICATE_SOURCE_NOT_A_BASE`. The sibling edit changes only the prescribed-remedy *text* of an already-published `422 WRITABLE_PACKAGE_REQUIRED` (code, status and `details` keys unchanged).

### Derived judgements, itemised (checklist ①)

1. **Refusal before the protocol call — holds at source.** Handler order at the head (`domains/packages.ts:1149-1174`): `requireManageMetadata` → resolve protocol (501 if absent) → `targetPackageId` body check (400) → `requireDuplicableSource` (422) → `protocol.duplicatePackage(...)`. Protocol side (`metadata-protocol/src/protocol.ts`): the `sys_metadata` scan at `:18437` is read-only; the target is minted via `this.installPackage(dupManifest)` at `:18539`; the copy loop starts at `:18552`. Mint precedes copy loop, refusal precedes mint — a later refusal would indeed leave the empty shell. The test double mints inside `duplicatePackage` for the same reason, which is what makes the "target absent" assertions non-vacuous.
2. **Not an emptiness check, not an existence oracle — holds.** The gate is `isWritablePackage(engine, id)` and nothing else (`packages.ts:441`). That predicate (`metadata-protocol/src/package-writability.ts:73-83`) reads exactly two signals — `engine.manifests.has(id)` (booted code package) and `registry.getPackage(id)?.manifest?.scope ∈ {system, cloud}` — and explicitly refuses the old "owns ≥ 1 object" fallback. No row count anywhere: a writable base with no active rows passes the gate and still answers `200 / copiedCount: 0`. An unknown id has no manifest entry and no scope, so the predicate answers `true` and the request falls through to the protocol's own answer rather than a 422 — the gate leaks no existence.
3. **Sibling message edit — sound, not a widening.** The maintainer-authorised triage direction on the card (2026-09-02), item 2, asked for exactly this: drop the duplicate remedy from `requireWritablePackage`'s 422 text and point at ADR-0005 overlay. In scope by direction, not only by the bounded-repair argument; the mechanical-forcing argument is also correct (a refusal cannot keep prescribing a route that now refuses for the same population). Code/status/`details` unchanged; no test pinned the old wording (tree-wide grep for the prescription finds only the PR's own new test comment).
4. **New wire vocabulary member** — additive; registered under the emitting package with the required distinction note against `WRITABLE_PACKAGE_REQUIRED`; no per-route error vocabulary exists in `package-lifecycle.zod.ts` or the route ledger that would also need it.
5. **Route-ledger note** — prose only; `responseSchema` unchanged.
6. **Nothing changed for a writable base**, including an empty one — pinned in section 7 of the test file.

### ADR-0070 readings — checked at source, all hold verbatim

`docs/adr/0070-package-first-authoring.md` is byte-identical on `origin/main` and the PR head (blob `a4713059920d`); the delivered file list touches no governed surface (`docs/adr/**`, `.claude/**`, `skills/**`, `AGENTS.md`, `CLAUDE.md`: zero files).

- `:3` — Status: "P1–P3 implemented, merged & live-verified … **D4–D6 remaining**".
- `:25` — "**Remaining**: **D4** (package-as-lifecycle-unit: delete-cascade / export / duplicate), D5, D6"; `:112` also carries "⏳ D4 package-as-unit (delete-cascade, export/duplicate)".
- `:80` — under the `### D4` heading at `:74`: "**Duplicate**: clone a **base** into a new writable package (the Airtable 'duplicate base' gesture)"; TL;DR 2 (`:30`) rules code/installed packages read-only for authoring.
- `:103` — open question 2: "…or should customising a code item also fork it into a writable base? (Leaning: keep overlay for surgical tweaks…)". Fair reading: teaching duplicate to clone a code package's items *is* forking code items into a writable base, which the ADR leaves open with a lean toward overlay.

The PR's core argument — the route shipped ahead of an unbuilt D4 whose object is a base; extending it to code packages is a new decision; refuse loudly — stands on the text as written.

### Semver (checklist ②)

`@objectstack/runtime: minor`, `@objectstack/spec: minor`, no `**BREAKING**` token. Precedent for the same shape: #7560 (the sibling 200 → 422 read-only refusal on `DELETE`/`disable`) shipped as **patch** ×3; the last new ledger code (`ACTION_DISABLED`) shipped as **minor** ×3. Declared level is at or above precedent and consistent with it: additive vocabulary on spec; on runtime, a refusal replacing a response whose `data.success` was already `false`. `check-changeset-no-major` and `check-adr-0087-registration` are green on this head.

### Boundary flags (checklist ③)

`open_questions: []`. Two out-of-scope findings handed back unfiled: (a) `protocol.destructive-409-face-inventory.test.ts:60-64` still says `duplicatePackage` has no response schema, which is false since `DuplicatePackageResponseSchema` landed (a docstring, nothing red) — worth a small card; (b) objectui's `packages-io.ts` doc and the Studio read-only hint still steer at duplicate, and the client will now receive a named 422 where it took the EMPTY arm — belongs to the objectui repo, already called for by triage. Neither blocks this PR.

### Generated projections

`content/docs/references/**` is not a governed surface. Both files carry the `AUTO-GENERATED — DO NOT EDIT. Run build-docs.ts` marker and are covered by `check:generated` (`packages/spec/scripts/check-generated.ts:134`, artifact `content/docs/references/**`). Content re-derived independently of the generator: enum size 301 → 302 ⇒ `+294 more` → `+295 more`, and the new bullet sits in sorted position between `DUPLICATE_REQUEST` and `ELIGIBILITY_UNEVALUABLE` in both files.

### What I ran, and what is NOT MEASURED

- **Ran:** `check-system-context-census` on a clean worktree at the head — exit 1 with the same two findings as CI. The CI red is reproducible, not flake.
- **Ran:** the enum-size and sorted-position re-derivation above (source-level, no generator).
- **NOT MEASURED — `check:docs` (generator re-run).** The first attempt on the head worktree exited 3-shaped (`packages/spec/json-schema is missing` — a build prerequisite), and the verdict is being posted rather than waiting on the dependency build. Confidence impact: low — the two projections are pinned by the count and position re-derivation, and CI's `Build Docs` is green on this head; the dev reports `check:generated` → "All 15 generated artifacts are up to date", which I did not reproduce.
- **NOT MEASURED — vitest on `packages-readonly-gate.test.ts` (this seat's own run).** The worktree run failed at resolution (`@objectstack/objectql` unbuilt), not on an assertion. Confidence impact: low — CI `Test Core` is green on the merge commit, which is an independent execution of this file; the file's own controls discriminate on the right axis (allow-path mints the record so "target absent" is non-vacuous; empty base still 200; unknown id still reaches the protocol).
- **NOT MEASURED — the ablation** (delete the one gate line → predicted 5 red / 20 green). Not re-derived: the review mandate forbids file changes and I did not mutate even a scratch tree. Taken on the dev's written-first prediction and reported `5 failed | 20 passed`.
- **NOT MEASURED on CI — 174 lint-job steps** (below).

### Landing state (separate from the contract verdict) — being repaired separately, not by this review

**Required check `Lint & Repo Gates` is RED on `61b7fc474e1`, and this PR caused it.** `check-system-context-census`: `content/docs/permissions/system-context.mdx:164` anchors `packages/runtime/src/domains/packages.ts:422` (the `shouldDenyAnonymous(... isSystem ...)` read). This PR inserts 113 lines above it (+9 at the `requireWritablePackage` docblock, +104 for `requireDuplicableSource`), so that identical line now sits at `:535` and `:422` is a docblock line. Pure line rot; the repair is the gate's own `node scripts/check-system-context-census.mjs --fix` (rewrites the one anchor; the page is not a governed surface). Because the job stopped at that step, **174 later repo-gate steps in that job never ran on CI for this head** (51 ran green) — NOT MEASURED, not green, until the fix is pushed. The PR also touches generated files, so the sync-then-regenerate rule (`scripts/pm/os-regen-merge.sh`) applies before enqueue.

The contract content of this PASS is pinned by blob: `packages/spec/src/api/error-code-ledger.zod.ts` `c2cf7c7e27af`, `packages/runtime/src/domains/packages.ts` `6c7bba0e7ad7`, `packages/runtime/src/route-ledger.ts` `f77d63625b1c`, `packages-readonly-gate.test.ts` `c827046a0f48`, `contract.mdx` `6bedba21906f`, `error-code-ledger.mdx` `6d85208aaef1`, changeset `12d539121c31`. If the anchor-fix push leaves these seven blobs unchanged and adds only the one-line rewrite in `content/docs/permissions/system-context.mdx`, this PASS covers that head as well; any other delta re-opens review.

---

Issue #15858, opened 2026-09-05
`rest-server.ts` still describes the scoped URL family as `/projects/:id/...` in two docblocks, the spelling ADR-0006 D2 retired with no alias


os-litant:
Found while implementing #15488 (the dispatcher's scope-strip regex). ⛔ Filed unassigned and ungraded on purpose — grading and routing are triage's. Filed by the `os-dev` implementer on session `session_01D47qPfEWVPmhguWgBZCi5N`, branch `claude/issue-15488-scope-strip-prefix`.

⚠️ **Title repaired after posting.** The first title spelled the retired form with an angle-bracket id placeholder and GitHub's body sanitizer ate the bracketed span, leaving `/projects//...` — the documented failure, reproduced live. Every placeholder below is spelled `:id` for that reason.

## The measurement

#15488 repaired `HttpDispatcher.dispatch`'s scope strip: the regex matched the pre-ADR-0006 `/projects/` prefix while its own comment, and `extractEnvironmentIdFromPath` next to it, said `/environments/`. Sweeping the repo for the same class of stale prose turned up two more sites, both in `packages/rest/src/rest-server.ts`, located by sentence rather than by line:

1. The `environmentId === 'platform'` docblock — *"a reserved virtual id used by Studio to address the control plane through the regular project URL shape (`/projects/platform/...`)"*, and the sentence after it says *"It is NOT a row in the projects table"*.
2. The single-project default-fallback comment — *"so bare `/api/v1/data/...` URLs (no `/projects/:id` prefix, no hostname mapping, no header) resolve to the lone project's kernel"*.

Both name the retired spelling as the current one. The authority in the other direction is unambiguous and already published:

- ADR-0006 v4, second addendum (D2 executed 2026-08-28): the API surfaces are renamed to `environment` / `environments` **with no aliases**, no dual-spelling interval.
- `content/docs/api/environment-routing.mdx`, migration checklist: *"Replace `/api/v1/projects/:projectId/...` with `/api/v1/environments/:environmentId/...`"*, and *"there is no alias, so the old spelling does not resolve"*.
- The same file's own live code spells the scoped base `/environments/:environmentId` in every one of its `isScoped` sites.

## Why it is filed rather than fixed in that PR

Two reasons, and the second is the binding one.

**Scope.** #15488's card is the dispatcher's strip. These are a different package's prose.

[...]

Issue #14503, opened 2026-09-02
packages/rest package read routes claim to SHADOW their dispatcher twins, but the dispatcher answers /packages and /packages/:id on a stock showcase boot


hotlong:
## What was measured

`packages/rest/src/package-routes.ts` documents its service-gated block as:

> The service-gated half — mounted only when a `package` service is reachable, because each of these three SHADOWS a live dispatcher twin at the same pattern and a degraded shadow is worse than no shadow.

On a stock showcase boot (`objectstack dev --seed-admin`, which mounts `PackageServicePlugin`, so the service IS reachable), the dispatcher twin answers instead. Two independent probes, same boot:

- **404 wording.** `GET /api/v1/packages/no.such.package` returns `Package 'no.such.package' not found` — the spelling in `packages/runtime/src/domains/packages.ts`. The REST route's own spelling is `Package "no.such.package" was not found.`
- **Response shape.** `GET /api/v1/packages/com.objectstack.setup` returns the installed-package record directly under `data` (`deps.success(pkg)` in the runtime domain). The REST route wraps it as `data.package` and stamps `source: 'registry'`, neither of which appeared.

## Why it matters

The comment states an intended precedence that the composition does not deliver, so a reader reasoning about which body serves `/packages` gets the wrong answer — which is exactly what happened on #14309, whose root-cause section attributed the 500 to the REST body. Two consequences worth separating:

1. If the shadowing is intended, the mount order is wrong and the REST routes are dead in every composition that also mounts the dispatcher domain — including the reference app. The `source` field and the `version` query-multiplicity refusal are then unreachable there.
2. If the dispatcher twin is meant to win, the comment is stale and should say so, because it is currently the only written statement of the precedence.

Either way, the two doors have diverged behaviour (`source` stamp, 404 wording, `data.package` vs `data` envelope) while claiming to be one door with a shadow, so a client cannot depend on which it gets.

## Not addressed by #14309

That card's fix is at the producer (`SchemaRegistry.installPackage`) and applies the read-door projection to both twins, so it is unaffected by which one answers. The precedence question is untouched and deliberately out of its scope.

Found while verifying #14309 on a live showcase boot.

---

## Triage — graded

Both spellings confirmed at `origin/main` `4a37870`, which is what makes the probe conclusive rather than suggestive:

- `packages/rest/src/package-routes.ts:652-654` — the SHADOWS sentence, verbatim as quoted.
[...]

Issue #12867, opened 2026-08-28
ADR-0006 D1 retirement addendum: record that D2 executed and the three retained-project surfaces are renamed


hotlong:
Part of #12865 (epic parent; unparked 2026-08-28 by maintainer ruling recorded there). Filed by epic PM session `session_65d2faee-1ff4-4be8-be28-72b972c539d9`.

## Scope

Append a retirement addendum to `docs/adr/0006-project-environment-split.v4.md` recording that the D2 pre-registration fired: the maintainer's 2026-08-28 ruling on #12865 opened the window, and the three retained-`project` surfaces named by D1 (the client SDK `projects` namespace, the response envelope keys, the JSDoc sentence) are renamed by the executing PRs (link the SDK-half PR from #12866 and the producer-half PR from objectstack-ai/cloud#1691 once they exist). D1 is retired — retire, never alias, per D2's own instruction.

Follow the MEASURED addendum conventions PR #12736 established and verified (top-level dated addendum heading appended after References, Status line amended to point at it, a why-an-addendum-not-a-new-ADR paragraph). Re-measure the file's state at branch time — do not assume this card's description.

Blocked-by: #12473

(That card's PR #12736 authored the D1/D2/D3 addendum this card retires. PR #12736 is an open DRAFT on the governed surface awaiting HUMAN merge. This card cannot start until it lands — starting earlier edits the same file against an unmerged base.)

## Landing constraint

Governed surface (docs/adr): draft-only PR, human merge, never auto-merge. Belongs to the same coordinated release as the rename halves. skip-changeset likely correct (doc-only) but derive it, do not copy.

hotlong:
Claim: PM loop round 3 (epic #12865, maintainer-directed; blocking edge resolved — PR #12736 human-merged, #12473 closed)
Session: `session_65d2faee-1ff4-4be8-be28-72b972c539d9`
Branch: `claude/issue-12867-adr0006-d1-retirement-addendum`
Worktree: `objectstack-issue-12867`
Domain: `domain:skills` per triage label; dispatched via the epic subtree (epic territory declared on #12865)
File surface: `docs/adr/0006-project-environment-split.v4.md` only (stop on breach; explain in the report)
Container & model: M docs-judgment, mode:subagent, model: opus (tier probe precedent: no path-derived mandate for docs/adr; PM judgment)
Clause-②: no
[...]

Issue #12747, opened 2026-08-27
[finding] ADR-0006 is what AGENTS.md and three ADR banners cite for the v5.0 `project` → `environment` rename — and its body does not contain that rename, nor does the changeset they name still exist


os-litant:
Filed unassigned and **ungraded** by the `domain:cli` seat (#6024), session `session_01UjujZN219uFzBhSYfMykCd`, on behalf of the **#12473 dev**, which measured it while writing that card's ADR-0006 addendum and **could not file it itself** — that seat gets 403 on the dedup channel, so the mandatory pre-file search was impossible from there. ⭐ It reported rather than filing blind. ⛔ Not graded, not routed.

Re-measured independently by this seat before filing. Nothing here is broken at runtime; this is a documentation-integrity defect in the record the repo routes agents to.

## The routing, measured

Four places send a reader to ADR-0006 for the v5.0 rename. The most important one is not an ADR:

- **`AGENTS.md:8`** — the repo's own top-level agent instructions, which state the breaking rename, then *"No aliases. See ADR-0006."*
- **`docs/adr/0004-cloud-multi-kernel.md:3`**, **`0005-metadata-customization-overlay.md:3`**, **`0008-metadata-repository-and-change-log.md:3`** — three **byte-identical** banners (measured: exactly 1 distinct banner text across the three), each saying the term was renamed throughout, *"See ADR-0006 for the rationale"*, and naming a changeset path *"for the breaking-change list"*.

## Both halves of that pointer are dead ends

**Half 1 — the record does not carry the rename.** Measured on `origin/main`, in `docs/adr/0006-project-environment-split.v4.md` (the **Accepted** revision):

| probe | hits |
| :--- | ---: |
| `v5.0` | **0** |
| `No aliases` | **0** |
| `environment` (**positive control**) | 13 |

⛔ The two zeroes are findings, not a broken pattern — the control answers in the same file with the same instrument.

⚠️ **And a naive `rename` grep answers 1, which is why this needs saying carefully.** That single hit is at `:196` and is about a *different* rename — an artifact type that will be renamed later. ⇒ **a non-zero can also be "not that thing"**, exactly symmetric to the false zeroes this lane keeps measuring. Anyone re-checking this finding with a bare `rename` grep will get a hit and wrongly close it.

[...]

PR #15859, opened 2026-09-05
fix(runtime): the dispatcher's scope strip matches `/environments/`, the prefix its own hint parser reads

### LINKED ISSUE #15488: `HttpDispatcher.dispatch`'s scope-strip comment and its regex disagree — the comment says `/environments/:environmentId`, the regex strips the legacy `/projects/` prefix, so a scoped URL matches no dispatcher domain

os-litant:
Fixes #15488

The card asked which of three readings is true: stale prose, a live routing defect, or a strip that should accept both spellings. It was driven, not read.

## The answer: reading 1 — a live routing defect

Measured through the **real** `@objectstack/hono` catch-all (`createHonoApp`, unmocked, real `HttpDispatcher`, `app.request()`), on `cc5b3dd0c27`, before any edit. Ten requests sent; the probe printed its own send count so an empty run could not read as a clean one.

| request | before | after |
|---|---|---|
| `GET /api/v1/health` — CONTROL | 200 | 200 |
| `GET /api/v1/data/task` — CONTROL | 503 `SERVICE_UNAVAILABLE` (the `/data` domain ran) | unchanged |
| `GET /api/v1/no-such-domain` — NEGATIVE CONTROL | 404 `ROUTE_NOT_FOUND` | unchanged |
| `GET /api/v1/environments/env_alpha/health` | **404 `ROUTE_NOT_FOUND`** | **200** |
| `GET /api/v1/environments/env_alpha/ready` | **404 `ROUTE_NOT_FOUND`** | **200** |
| `GET /api/v1/environments/env_alpha/data/task` | **404 `ROUTE_NOT_FOUND`** | **503 — it reaches `/data`** |
| `GET /api/v1/projects/env_alpha/health` | 200 | **404 `ROUTE_NOT_FOUND`** |

The negative control matters: an unclaimed path and a scoped path answered the *same* 404 shape before the repair, which is what "matches no dispatcher domain" looks like from outside.

A second probe read the mutated context back, and it is the sharper half — the two spellings fail in **opposite** directions:

```
/api/v1/projects/env_alpha/data/task       urlEnvironmentId = undefined   -> stripped, served UNSCOPED
/api/v1/environments/env_alpha/data/task   urlEnvironmentId = "env_alpha" -> parsed, then matched NO domain
```

So `dispatch()`'s own two readings of one convention disagreed with each other, in this repository, with no cloud-side wiring needed to decide it. `extractEnvironmentIdFromPath` parsed `/environments/`; the strip removed `/projects/`. The card's reading 2 ("dead legacy, fix the comment") is refuted, and so is the triage comment's reading 3 ("both spellings are meant to work") — see below.

**Why the catch-all specifically.** It is the only entry that hands `dispatch()` a still-scoped path: `const subPath = c.req.path.substring(prefix.length)`. Every scoped mount in `dispatcher-plugin.ts` passes a pre-stripped subpath — `registerAutomationRoutes` mounts `${prefix}/environments/:environmentId/automation` and dispatches the literal `/automation` — which is why the standalone server never showed this.

**The legacy prefix was not a working alias in exchange.** Nothing parses `/projects/:id`, so stripping it discarded the only place the request named an environment: it was served from the host default, not the environment its own URL named. A 404 is the honest answer, and that is what it answers now.

## Why one spelling, not an alternation

The triage comment's reading 3 would widen the regex to accept either prefix. That is exactly what the standing ruling forbids, and the four axes agree, led by long-term rationality:

- **Long-term rationality (the heaviest axis).** ADR-0006 v4, second addendum — D2 executed 2026-08-28 — renames the API surfaces to `environment` / `environments` **with no aliases**, and the ADR states the posture in as many words: *"no gradualism, no dual-spelling interval, no single release carrying both."* An alternation here is precisely the alias it exists to prevent.
- **Actual need.** No in-repo emitter of a `/projects/:id` API URL survives: the client SDK builds scoped bases as `.../environments/:id`, `rest-server.ts` mounts `/environments/:environmentId` at every one of its `isScoped` sites, and `dispatcher-plugin.ts` mounts the same. The one remaining hit is a UI-location fixture in an unrelated websocket test.
- **Guarding against the next agent's mistake.** A comment that lies about a regex is the shape that teaches the wrong invariant, and a regex accepting two spellings under a comment naming one is the same defect with more surface.
- **Not diffusing scope pre-launch.** `content/docs/api/environment-routing.mdx` already tells callers *"Replace `/api/v1/projects/:projectId/...` with `/api/v1/environments/:environmentId/...`"* and *"there is no alias, so the old spelling does not resolve."* Honouring the published checklist costs nothing; contradicting it costs a compatibility surface.

## The repair — four sites, all located by content

Three in `packages/runtime/src/http-dispatcher.ts`, one in a sibling:

1. **The strip** — `/^\/projects\/[^/]+(\/.*)?$/` becomes `/^\/environments\/[^/]+(\/.*)?$/`. The card's subject.
2. **The OAuth-on-MCP gate** — `acceptOAuthAccessToken` tested the still-scoped path against the same retired prefix. **This had to move with the strip, and could not have been left for a follow-up**: unrepaired, a scoped `/mcp` URL never reached the MCP domain at all, so the flag was unobservable; repairing the strip is exactly what makes the line reachable, and left behind it would have started admitting scoped MCP callers to the domain with their OAuth 2.1 access tokens refused. A new defect manufactured by the fix.
3. **An orphaned pre-rename docblock** stacked above `extractEnvironmentIdFromPath`, describing a *"project UUID"* and the retired URL form. Deleted; the live docblock immediately beneath it already says the right thing.
4. **`packages/runtime/src/security/permission-denied-envelope.ts`** — `routeObjectFromPath`'s docblock states what it expects of the dispatcher's cleaned path, and said *"`/projects/:environmentId` prefix stripped"*. Prose only; the function's own `/^\/data\/([^/?#]+)/` is untouched.

Sites 2 to 4 are declared in-scope deliberately rather than swept in quietly: same defect class, mechanical, the correct spelling pinned by the ADR and by site 1, no new gate family (re-derived and diffed — identical list). The claim comment on the card names the file surface.

**What this branch did NOT touch.** `packages/rest/src/rest-server.ts` carries two more docblocks naming the retired form, and that file is held by two open pull requests — #15673 and #15395. The sweep stopped there and filed the finding as #15858 instead of reaching into a file two branches are rewriting.

## The pin, and the ablation that proves it can fail

New: `packages/runtime/src/http-dispatcher.scoped-url-strip.test.ts` — 7 assertions, including the two controls and the negative control from the table above, and an observation seam that captures the `acceptOAuthAccessToken` decision.

Both ablation limbs had their direction **and their green assertions** predicted in writing first, then measured. Each mutation was proven on disk by removed-text and injected-marker counts before the run, and restored under a trap via `git checkout HEAD -- <absolute path>`, proven by blob-hash equality against the HEAD blob and an empty `git diff HEAD`.

| limb (revert one site only) | predicted | measured |
|---|---|---|
| A — the strip | 2 fail, 5 pass | 2 fail, 5 pass — the two scoped-routing assertions, exactly as named |
| B — the OAuth matcher | 2 fail, 5 pass | 2 fail, 5 pass — both OAuth assertions, exactly as named |

⭐ **The assertion that deliberately stays GREEN under limb A is the finding restated as a test**: "parses the environment id off the SAME prefix it strips" keeps answering `env_alpha` while the strip is broken, because `extractEnvironmentIdFromPath` is a *separate* reading. That independence is exactly how the two drifted apart unnoticed for a release, and an ablation that reddened it would be measuring the wrong thing.

**No rebuild was needed, and that is proven rather than assumed.** The pin imports `./http-dispatcher.js` — a relative import inside the package, resolved to source. To make the claim falsifiable, each limb printed `dist/index.js` still holding the repaired spelling at mutation time: a RED pin therefore proves source resolution, and a GREEN one would have proved the suite was reading dist.

## Verification — all on the final head `4538d6289d5`

- **Gate union: 53 of 53 families green**, derived with `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack --commands` on that head with no STALE TREE banner, re-derived after the change set was final and diffed against the executed list (identical). Two of them (`check:dual-build-cjs-loads`, `check:type-check-debt`) first answered `PREREQUISITE NOT MET` / exit 3 — read as NOT MEASURED, not as a pass — and were re-run green after `turbo run build` over the workspace closure. Exit codes captured after redirection, never through a pipe.
- **Tests**: 39 files / 714 assertions green across the `http-dispatcher.*`, `dispatcher-plugin`, `error-envelope.conformance`, `auth-unknown-subpath.hono.integration` and `security/*` suites in `@objectstack/runtime`, plus `@objectstack/hono` (2 files / 74).
- **Typecheck**: `@objectstack/runtime` green, and the new test file was confirmed **inside** the type-checked program via `tsc -p tsconfig.test.json --listFiles` (1 hit, 0 errors attributable to either changed file) rather than inferred from a green script.
- **Lint**: the full repo sweep, `pnpm lint` = `eslint . --no-inline-config`, green in 90s — no narrowing claimed or needed.
- Control-character self-scan over the changed files: clean.

## For triage

The triage comment set a promotion rule: *"if reading 1 or 3 is established, re-grade to p2 immediately — no new card, no re-argument."* Reading 1 is established by driven measurement, so that rule has fired.

The same comment's boundary test applies too: this changes what the door accepts — requests that reached no dispatcher domain now reach one, and the retired spelling now refuses — so it reads as clause ② YES and a manual floor. ⛔ Left as a draft for that judgement rather than pre-labelled.

One observation handed back rather than acted on: the card called this *"the third independent reason the dispatcher cannot serve the scoped `/packages` path"*, alongside two measured by #14503's census. This branch removes this one — a scoped `/packages` URL now reaches the `/packages` domain through the catch-all — which may change how #14503's other two findings read. That is #14503's question, not this card's.

---

### DIFF
--- a/.changeset/dispatcher-scope-strip-environments-prefix.md
+++ b/.changeset/dispatcher-scope-strip-environments-prefix.md
@@ -1,0 +1,19 @@
+---
+"@objectstack/runtime": patch
+---
+
+An environment-scoped URL now reaches a dispatcher domain instead of answering 404.
+
+`HttpDispatcher.dispatch()` reads the scoped-URL prefix in three places — the environment-id hint parser, the OAuth-on-MCP gate, and the scope strip that lets `DomainHandlerRegistry` match the remainder. Only the first had been moved to the ADR-0006 `/environments/` spelling; the other two still matched the retired `/projects/` one. The strip therefore never fired on a real scoped URL, and since the registry matches from the head of the path, every environment-scoped request arriving through the `@objectstack/hono` catch-all — the entry cloud hosts mount, and the only one that hands `dispatch()` a still-scoped path — matched no domain at all:
+
+```
+GET /api/v1/environments/<id>/data/task   ->  404 ROUTE_NOT_FOUND   (now: reaches /data)
+GET /api/v1/environments/<id>/health      ->  404 ROUTE_NOT_FOUND   (now: 200)
+GET /api/v1/data/task          (control)  ->  reaches /data, unchanged
+```
+
+The dispatcher-plugin's own scoped mounts were never affected: they pass a pre-stripped subpath (`${prefix}/environments/:environmentId/automation` dispatches the literal `/automation`), which is why the standalone server showed nothing.
+
+The OAuth 2.1 gate moved with it. An access token is honoured only on the MCP surface, and that test runs against the still-scoped path — so `/api/v1/environments/<id>/mcp` would have reached the MCP domain with its token refused had the strip been repaired alone.
+
+**If you still emit the old spelling**: replace `/api/v1/projects/:projectId/...` with `/api/v1/environments/:environmentId/...`, as `content/docs/api/environment-routing.mdx` has instructed since ADR-0006 D2. That prefix is no longer stripped, and it was never a working alias in the first place: nothing parses `/projects/<id>`, so stripping it discarded the only place the request named an environment and served it from the host default instead. ADR-0006 D2 retired `project` on the API surface with no aliases, so the repair is one spelling in all three readings rather than a two-prefix alternation.
--- a/packages/runtime/src/http-dispatcher.scoped-url-strip.test.ts
+++ b/packages/runtime/src/http-dispatcher.scoped-url-strip.test.ts
@@ -1,0 +1,172 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+import { describe, it, expect } from 'vitest';
+import { HttpDispatcher, type HttpDispatcherResult } from './http-dispatcher.js';
+
+/**
+ * The scoped-URL prefix is ONE convention, read three times in one request —
+ * and the three readings had drifted apart.
+ *
+ * `dispatch()` handles an environment-scoped URL by reading the same prefix in
+ * three places:
+ *
+ *   1. `extractEnvironmentIdFromPath` (via `prepareResolverHints`) parses the
+ *      environment id out of it and hands it to the host's `KernelResolver`;
+ *   2. the `acceptOAuthAccessToken` test decides whether an OAuth 2.1 access
+ *      token is honoured, and runs against the STILL-SCOPED path;
+ *   3. the scope strip removes the prefix so `DomainHandlerRegistry` — which
+ *      matches from the head of the path — can claim the remainder.
+ *
+ * Reading 1 said `/environments/`; readings 2 and 3 still said the
+ * pre-ADR-0006 `/projects/`. What that cost, measured through the real
+ * `@objectstack/hono` catch-all before the repair:
+ *
+ * ```
+ * GET /api/v1/environments/env_alpha/health     -> 404 ROUTE_NOT_FOUND
+ * GET /api/v1/environments/env_alpha/data/task  -> 404 ROUTE_NOT_FOUND
+ * GET /api/v1/health                 (control)  -> 200
+ * GET /api/v1/data/task              (control)  -> 503 SERVICE_UNAVAILABLE (i.e. it REACHED /data)
+ * GET /api/v1/projects/env_alpha/health         -> 200   <- stripped, but nothing ever parsed the id
+ * ```
+ *
+ * The legacy spelling was not a working alias in exchange: nothing parses
+ * `/projects/<id>`, so stripping it discarded the only place the environment
+ * was named and served the request from the host default. ADR-0006 D2 retired
+ * `project` on the API surface with NO aliases, and
+ * `content/docs/api/environment-routing.mdx` tells callers to replace
+ * `/api/v1/projects/:projectId/...` with `/api/v1/environments/:environmentId/...`
+ * — so the repair is one spelling everywhere, not a two-prefix alternation.
+ *
+ * ## Why this suite drives `dispatch()` with a hand-derived subpath
+ *
+ * `packages/runtime` cannot depend on `@objectstack/hono` (that adapter depends
+ * on THIS package). What the adapter contributes is one line —
+ * `const subPath = c.req.path.substring(prefix.length)` in the
+ * `app.all(`${prefix}/*`)` catch-all of `packages/adapters/hono/src/index.ts` —
+ * so {@link subPathAsTheCatchAllDerivesIt} reproduces exactly that, and nothing
+ * else. It matters that it is the catch-all: it is the ONLY entry that hands
+ * `dispatch()` a still-scoped path. Every scoped mount in `dispatcher-plugin.ts`
+ * passes a pre-stripped subpath (`registerAutomationRoutes` mounts
+ * `${prefix}/environments/:environmentId/automation` but dispatches the literal
+ * `'/automation'`), which is why the standalone server never saw this.
+ */
+
+const PREFIX = '/api/v1';
+const ENV_ID = 'env_alpha';
+
+/** The `@objectstack/hono` catch-all's one contribution, reproduced exactly. */
+function subPathAsTheCatchAllDerivesIt(url: string): string {
+    return url.substring(PREFIX.length);
+}
+
+/**
+ * A kernel that provides NO services at all.
+ *
+ * Deliberate: it makes "which domain claimed the path" the only variable. A
+ * claimed `/data` answers 503 SERVICE_UNAVAILABLE (the domain ran and found no
+ * data service); an unclaimed path answers 404 ROUTE_NOT_FOUND. Those two are
+ * the whole signal, and a kernel with services wired would blur them.
+ */
+function bareKernel(): any {
+    return {
+        getState: () => 'running',
+        getService: () => undefined,
+        getServiceAsync: async () => undefined,
+    };
+}
+
+function dispatcher(): HttpDispatcher {
+    return new HttpDispatcher(bareKernel(), undefined, { enforceProjectMembership: false });
+}
+
+/**
+ * Narrow the optional `response`. Lifted from `http-dispatcher.ready.test.ts`
+ * for the reason stated there: this package's test layer IS type-checked, and
+ * `expect(res.response).toBeDefined()` narrows nothing. "Answered no response
+ * at all" and "answered the wrong status" must stay distinguishable.
+ */
+function responseOf(res: HttpDispatcherResult, what: string): NonNullable<HttpDispatcherResult['response']> {
+    const { response } = res;
+    if (!response) throw new Error(`${what} answered no response at all`);
+    return response;
+}
+
+/** `GET <url>` through the catch-all derivation; returns status + error code + the context it mutated. */
+async function get(url: string): Promise<{ status: number; code: unknown; ctx: any }> {
+    const ctx: any = { request: new Request(`http://pin.local${url}`) };
+    const res = await dispatcher().dispatch('GET', subPathAsTheCatchAllDerivesIt(url), undefined, {}, ctx, PREFIX);
+    const response = responseOf(res, `GET ${url}`);
+    return { status: response.status, code: (response.body as any)?.error?.code, ctx };
+}
+
+describe('scoped-URL prefix — the environment-scoped URL reaches a dispatcher domain', () => {
+    it('CONTROL: the unscoped forms reach their domains (a suite that measured nothing would fail here first)', async () => {
+        expect((await get(`${PREFIX}/health`)).status).toBe(200);
+
+        // `/data` has no exact/short-circuit answer, so it proves the DOMAIN ran
+        // rather than merely that some route matched: the domain body is what
+        // raises 503 on a missing data service.
+        await expect(get(`${PREFIX}/data/task`)).rejects.toMatchObject({ statusCode: 503 });
+    });
+
+    it('NEGATIVE CONTROL: an unclaimed path answers 404 ROUTE_NOT_FOUND — the shape "matched no domain" takes', async () => {
+        const res = await get(`${PREFIX}/no-such-domain`);
+        expect(res.status).toBe(404);
+        expect(res.code).toBe('ROUTE_NOT_FOUND');
+    });
+
+    it('the environment-scoped URL is stripped and reaches the same domain as the unscoped one', async () => {
+        expect((await get(`${PREFIX}/environments/${ENV_ID}/health`)).status).toBe(200);
+        await expect(get(`${PREFIX}/environments/${ENV_ID}/data/task`)).rejects.toMatchObject({ statusCode: 503 });
+    });
+
+    it('parses the environment id off the SAME prefix it strips — the two readings must not drift again', async () => {
+        const { ctx } = await get(`${PREFIX}/environments/${ENV_ID}/health`);
+        expect(ctx.urlEnvironmentId).toBe(ENV_ID);
+    });
+
+    it('the retired `/projects/` spelling resolves nothing — ADR-0006 D2 grants no alias', async () => {
+        const res = await get(`${PREFIX}/projects/${ENV_ID}/health`);
+        expect(res.status).toBe(404);
+        expect(res.code).toBe('ROUTE_NOT_FOUND');
+
+        // The point of removing it, stated as an assertion: this spelling was
+        // never a working alias. Nothing parses it, so the pre-repair strip
+        // deleted the only mention of the environment and served the request
+        // from the host default. A 404 is the honest answer to a URL naming an
+        // environment the dispatcher cannot resolve.
+        expect(res.ctx.urlEnvironmentId).toBeUndefined();
+    });
+});
+
+describe('scoped-URL prefix — the OAuth-on-MCP gate reads the same prefix', () => {
+    /**
+     * Capture the `acceptOAuthAccessToken` decision `resolveRequestScope`
+     * computes. It is handed to a private method, so the instance-level
+     * override is the observation seam; throwing afterwards is deliberate —
+     * `resolveRequestScope` catches it and leaves `executionContext` undefined,
+     * which is precisely the anonymous-request path and keeps this pin free of
+     * any identity stack.
+     */
+    async function acceptsOAuthFor(url: string): Promise<boolean | undefined> {
+        const d = dispatcher();
+        let seen: boolean | undefined;
+        (d as any).timedResolveExecutionContext = async (opts: { acceptOAuthAccessToken?: boolean }) => {
+            seen = opts.acceptOAuthAccessToken;
+            throw new Error('captured — anonymous from here');
+        };
+        const ctx: any = { request: new Request(`http://pin.local${url}`) };
+        await d.resolveRequestScope(ctx, subPathAsTheCatchAllDerivesIt(url));
+        return seen;
+    }
+
+    it('honours OAuth access tokens on the plain AND the environment-scoped MCP route', async () => {
+        expect(await acceptsOAuthFor(`${PREFIX}/mcp`)).toBe(true);
+        expect(await acceptsOAuthFor(`${PREFIX}/environments/${ENV_ID}/mcp`)).toBe(true);
+    });
+
+    it('CONTROL: refuses them off the MCP surface, and on the retired `/projects/` spelling', async () => {
+        expect(await acceptsOAuthFor(`${PREFIX}/data/task`)).toBe(false);
+        expect(await acceptsOAuthFor(`${PREFIX}/projects/${ENV_ID}/mcp`)).toBe(false);
+    });
+});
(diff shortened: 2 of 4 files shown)


os-litant (commented):
## Contract review (clause ②) — in-seat, contract-review tier

```
VERDICT: PASS
REVIEWED-HEAD: 4538d6289d5 (PR #15859)
CLAUSE-2-PATH: no
CLAUSE-2-CONTENT: yes
Implemented-by: `claude/issue-15488-scope-strip-prefix`
Reviewed-by: `session_01D47qPfEWVPmhguWgBZCi5N`
FINDINGS: 0 blocking. 2 citation-precision notes (non-blocking, wording only). 1 residual reading that only the cloud repository can answer (NOT MEASURED here, already carried by #15861).
```

**Why this is a COMMENT and not an approval.** GitHub refuses `APPROVE` on a pull request authored under the same account, and agent seats do not submit approving reviews in any case. The verdict is the fenced block above; the dispatching seat performs the landing steps. The `Implemented-by:` identity is the dev's branch (a subagent dev has no session of its own; the branch is the one its claim comment on #15488 names), and `Reviewed-by:` is this seat's session — two disjoint grammars, so this is not a self-review under the C4 rule. Tier: `get_session` reads the dispatching session, not this reviewer, so the reading that counts is the harness `model` stamp in this run's own transcript (`subagents/agent-a0cb91b66ad655b2b.jsonl`, 60 stamps, all equal to `CONTRACT_REVIEW_TIER` as exported by `scripts/pm/dispatch-gates.mjs`). The seat is to verify that transcript before adopting this verdict, per the transcript-verification rule.

### Clause ②, re-judged from the delivered diff

**Path limb — no.** Four files: `.changeset/dispatcher-scope-strip-environments-prefix.md`, `packages/runtime/src/http-dispatcher.ts`, `packages/runtime/src/http-dispatcher.scoped-url-strip.test.ts`, `packages/runtime/src/security/permission-denied-envelope.ts` (docblock only). No governed surface is touched: `docs/adr/**` is read, not changed. No new exported symbol, no new key on a published payload.

**Conformance limb — yes, and correctly declared.** The diff changes what the door accepts, in four derived judgments, each checked:

1. `GET /api/v1/environments/<id>/<domain>` through the `@objectstack/hono` catch-all now reaches `<domain>` (was 404 `ROUTE_NOT_FOUND`). **Correct.** It is the documented surface (`content/docs/api/environment-routing.mdx`, the 6.0.0 changelog migration entry, the `dispatcher-plugin.ts` scoped mounts, the hint parser) and the catch-all is the only entry that hands `dispatch()` a still-scoped path (`packages/adapters/hono/src/index.ts`: `subPath = c.req.path.substring(prefix.length)`, context `{ request: c.req.raw }` — no router `params`).
2. `GET /api/v1/projects/<id>/<domain>` now answers 404 (was stripped and served). **Correct**, and the pre-change behaviour was a defect, not a feature — see the alias derivation below.
3. OAuth 2.1 access tokens are now honoured on `/environments/<id>/mcp` (were not). **Correct and non-deferrable** — see the OAuth section.
4. OAuth tokens are no longer honoured on `/projects/<id>/mcp` (were, on the host-default environment). Consistent with 2: that path no longer reaches any domain.

**Semver.** `patch` on `@objectstack/runtime` is consistent with the repo rule (a bug fix in a released package takes a patch) and with the fact that the scoped-route rename was already published as a breaking migration entry in the runtime `6.0.0` changelog ("Scoped routes: `/api/v1/projects/:projectId/...` → `/api/v1/environments/:environmentId/...`"). The changeset body carries the FROM → TO line regardless.

**Boundary flag.** The dev's one open question (does a cloud host still emit the retired prefix?) was ruled **C** by the seat and is carried by #15861. Not re-opened here; a sharpening is noted under residual risk.

### Source readings — do they hold?

**ADR-0006 v4 — holds in substance, misattributed in citation.** Read at `docs/adr/0006-project-environment-split.v4.md`, unchanged by this PR:

- The second addendum (2026-08-28, #12867) did execute D2 with **no alias** ("no deprecated forwarder, no compatibility getter, nothing"; "no dual-key emission and no compatibility window"). But D2's three surfaces are D1's three: the SDK `projects` namespace, the control-plane response keys, and the SDK JSDoc. **The scoped URL prefix is not among them.** It was renamed in the v5.0 rename itself — commit `944f18758ef` (2026-05-24, "rename project to environment throughout codebase"), the same commit that flipped `extractEnvironmentIdFromPath` and `rest-server.ts` to `/environments/` and left the strip (from `276b94284a6`, 2026-04-23) and the OAuth gate behind. That rename is recorded in the ADR's section **"The v5.0 rename and its no-alias decision"** (#12747), which is also where the quoted sentence lives — *"no gradualism, no dual-spelling interval, no single release carrying both"* (lines 303–305), not in the second addendum.
- Consequence for the verdict: none. The applicable rule is the general one and it is stronger than the PR states — the prefix was never granted a retention window at all, and AGENTS.md's "No aliases. See ADR-0006." points at the same record. The "one spelling, not an alternation" argument stands at source.
- Consequence for the prose: the strip comment, the test docblock and the changeset all say "ADR-0006 D2 retired `project` on the API surface with no aliases". A reader who follows that to D2 finds three other surfaces. **Non-blocking wording note**: cite the v5.0 no-alias section for the prefix, and D2 only as the later confirmation that the rule reached even the last retained surfaces without an alias. Fold in or file, the seat's call; it does not change the acceptance set.

**`content/docs/api/environment-routing.mdx` — item 1 holds verbatim; the second quote belongs to item 5.** Migration checklist item 1 reads exactly *"Replace `/api/v1/projects/:projectId/...` with `/api/v1/environments/:environmentId/...`"*. The sentence *"there is no alias, so the old spelling does not resolve"* is checklist **item 5**, about the SDK `client.project(id)` method, not about the URL prefix; the page does not literally say the old URL prefix does not resolve. The in-code comment cites only item 1 (accurate); the PR body juxtaposes the two as if both concern the prefix. **Non-blocking**: the caller instruction the PR relies on is item 1, and the 6.0.0 changelog entry says the same about the prefix. (Out of scope, handed to the seat: the same page's "Client SDK" section still says the client "keeps the legacy factory name `project(id)`" while item 5 says it was retired at D2 — a docs inconsistency for a docs card, not this PR.)

### The crux — was `/projects/<id>` ever a working alias? No, not in this repository.

Re-derived, not accepted. Three independent measurements, each with a control:

**(a) Source derivation on both trees.** I extracted the three literal regexes (hint parser, OAuth gate, strip) from the merge-base dispatcher (`5a21d73`) and from the head dispatcher (`4538d62`) and evaluated them on ten URLs with the registry's match rules (`exact`/`segment`/`prefix`, `startsWith` from the head of the path). Control axis: same inputs, same evaluator, **only the prefix spelling differs** between the two trees; the unscoped forms (`/health` → `/health`, `/data/task` → `/data`) and the negative control (`/no-such-domain` → no domain) read identically on both. Merge-base:

```
/api/v1/projects/env_alpha/data/task      urlEnvironmentId=undefined  oauth=false  after-strip=/data/task                     domain=/data
/api/v1/projects/env_alpha/mcp            urlEnvironmentId=undefined  oauth=true   after-strip=/mcp                           domain=/mcp
/api/v1/environments/env_alpha/data/task  urlEnvironmentId=env_alpha  oauth=false  after-strip=/environments/env_alpha/data/task  domain=NONE (404)
```

Head: the two rows swap — `/environments/…` parses, strips and reaches its domain with OAuth true on `/mcp`; `/projects/…` parses nothing, strips nothing, reaches nothing. `/cloud/environments/<id>/mcp` is excluded on both trees (anchored regexes plus the control-plane guard). So on the merge-base the retired prefix was stripped by a reading that nobody paired with a parser: `urlEnvironmentId` stayed `undefined` (the hint regex is `/\/environments\/([^/?#]+)/`; the `context.request?.params?.environmentId` fallback is absent on the catch-all's raw `Request`), and the host resolver went on to hostname / header / session / default. The request named an environment and was served from a different one, silently. One fact the PR body did not print: `/projects/<id>/mcp` reached the MCP domain **with OAuth honoured**, on that default environment. The 404 is a wrong answer becoming a refusal; the risk reading in #15861 is the right way round.

**(b) The pin, and an ablation I ran myself.** In a fresh worktree at `4538d6289d5` (dependency closure built; **no `packages/runtime/dist` present**):
- `vitest run src/http-dispatcher.scoped-url-strip.test.ts` → exit 0, **7 passed / 7**.
- Same pin with the merge-base dispatcher blob swapped in (`git cat-file blob 5a21d73:packages/runtime/src/http-dispatcher.ts`; pre-change spelling on disk = 1, head spelling = 0) → exit 1, **4 failed / 3 passed**: the two scoped-routing assertions and both OAuth assertions red; the two controls and the "parses the id off the SAME prefix it strips" assertion green — exactly the split the drift predicts, since the hint parser was already correct. With no runtime `dist` in the tree, a red ablation is proof the pin resolves source.
- Restored from the head blob, verified by hash (`bdab9bf2fc1…` on disk = `HEAD:…/http-dispatcher.ts`), `git diff --quiet` clean.

**(c) No other reader, no in-repo emitter.** Full-scope `git grep` over `packages/` (non-test, non-dist): no reader of `context.routePath` / `urlEnvironmentId` outside the dispatcher, no in-repo `KernelResolver` implementation (the only in-repo `resolveKernel` site is a consumer in `plugin-hono-server/current-user-endpoints.ts`), and no emitter of the `/projects/:id` scoped prefix — the hits are changelog migration entries, a REST README example in which `projects` is an object name, the two `rest-server.ts` docblocks already filed as #15858, and a storage bucket prefix. Every scoped mount in `dispatcher-plugin.ts` dispatches a pre-stripped literal subpath with router `params`, which is why the standalone server never showed the defect.

### The `acceptOAuthAccessToken` site — the reasoning holds, and the repaired gate is correct, not merely consistent

**Non-deferrability holds.** With the flag false, `resolve-execution-context.ts` never extracts a JWT bearer (`if (opts.acceptOAuthAccessToken) { … }`), so the request falls to the better-auth session path, which does not recognise the JWT, and resolves as a guest. Before the repair a scoped `/mcp` URL never reached the MCP domain, so the flag's value there was unobservable; repairing the strip alone would have made `/environments/<id>/mcp` reach the MCP domain with the token unread — a scoped MCP caller admitted to the domain as a guest. The two lines had to move together, and the ablation confirms the pin discriminates them (both OAuth assertions red under the swap).

**Correctness of the repaired gate**, judged against the strip and the domain rather than against itself:
- Same anchor, same id segment: `^(?:\/environments\/[^/]+)?\/mcp(?:[/?]|$)` versus the strip `^\/environments\/[^/]+(\/.*)?$`. Whatever the gate accepts as a scoped MCP path, the strip reduces to `/mcp…`; the MCP domain is registered `{ prefix: '/mcp', match: 'segment' }`, and the gate's `(?:[/?]|$)` boundary is the same segment boundary — `/mcpfoo` is refused by both.
- Both run on the same `cleanPath` string (`dispatch()` passes it to `resolveRequestScope` before the strip reassigns it), so there is no second normalisation to drift.
- The direction is not a widening. The gate re-enables the #2698 contract as originally written ("the plain and scoped route forms"); the token is verified against the per-request kernel's auth service (`resolveService(requestKernel(context), 'auth', context.environmentId)`), i.e. the environment the URL names, and the failure mode is unchanged and fail-closed (a JWT-shaped bearer that does not verify yields hard anonymous, no cookie fallback).
- No second gate is left in the same half-state: `enforceAuthGate` classifies by allow-list (`isAuthGateAllowlisted`, `/auth/` segment, `/health`-style suffixes) and applies to every other path regardless of head segment, and `enforceProjectMembership` keys on the resolved `environmentId`, not on the path head — so a scoped data/MCP request that now reaches its domain is gated by both.

### Verification run for this review (exit codes captured before any pipe)

- Pin at head: exit 0, 7/7. Ablation (merge-base blob): exit 1, 4 failed / 3 passed. Restore: blob-hash equal, tree clean.
- Runtime regression at head, same selection the dev ran (`http-dispatcher*`, `dispatcher-plugin*`, `security/*`, `error-envelope.conformance`, `auth-unknown-subpath.hono.integration`): exit 0, **39 files / 714 tests passed**.
- `@objectstack/hono` adapter suite at head: exit 0, **2 files / 74 tests passed** (run after `turbo run build --filter=@objectstack/runtime`, 30/30 tasks, since the adapter resolves runtime through `dist`).
- A first pin attempt in the fresh worktree answered "no tests" (`@objectstack/observability` had no `dist`): recorded as PREREQUISITE NOT MET, not as a result; the dependency closure was built (`turbo run build --filter='@objectstack/runtime^...'`, 29/29 tasks) and the pin re-run.
- CI on `4538d6289d5`: all 30 completed check runs are success or skipped; the six required checks (TypeScript Type Check, Lint & Repo Gates, Test Core, Dogfood Regression Gate, Build Core, Temporal Conformance) are green. This is a reading of CI, not a re-run.

**NOT MEASURED (reported as such, not as a pass):**
- Whether any `objectstack-ai/cloud` `KernelResolver` parses `context.routePath` for `/projects/`. `routePath` is handed to the host resolver **unstripped**, so a cloud-side parser is the one reader outside this repository that could make the retired prefix a working alias on those hosts. Recommend #15861 ask that question explicitly, alongside "does any host still emit it". Per ADR-0006 the answer cannot re-introduce an alias here, so it is residual risk for the cloud side, not an input to this PR's shape.
- The dev's end-to-end probe through `createHonoApp` + `app.request()`: not re-run. The catch-all's single contributing line was verified by reading and the pin reproduces it exactly.
- The 53-family gate union and the `tsc -p tsconfig.test.json --listFiles` inclusion of the new test file: not re-run here; CI type-check and lint are green on the head.

### Residual risk, not re-litigated

The cross-repo emitter question is ruled (C, #15861). What this review adds is only the sharper reading above, and the observation that any cloud host still emitting the retired prefix is already served from the wrong environment today.

---

Issue #15387, opened 2026-09-04
`GET /auth/me/localization` answers `currency: null` / `timezone: null` for every authenticated caller — the current-user resolver assembles no localization


os-justin:
Found while executing #14788 (PR #15386), out of that card's scope (the ruling touched `locale` only; `currency` / `timezone` were to stay unchanged — and they are unchanged, which is the defect).

## What

`packages/plugins/plugin-hono-server/src/current-user-endpoints.ts` — the `/auth/me/localization` handler returns `currency: execCtx.currency ?? null` and `timezone: execCtx.timezone ?? null`, and its comment says the resolved regional defaults "ride on the request ExecutionContext (ADR-0053)". But the context these endpoints resolve comes from `makeExecutionContextResolver` in the same file, which returns `{ userId, tenantId, email, positions, permissions, systemPermissions, tabPermissions, posture, isSystem, org_user_ids, accessible_org_ids }` — no `timezone`, no `currency` (and, before PR #15386, no `locale`). The dispatcher's shared assembler (`packages/core/src/security/assemble-execution-context.ts`) is what fills those three from `resolveLocalizationContext`; this resolver never calls it.

So for every authenticated caller the endpoint answers `currency: null, timezone: null`, whatever the `localization` settings say. objectui reads `currency` off this endpoint (`apps/console/src/languageSeed.ts`, `LocalizationFetchProvider.tsx`) — the console's regional formatting seed is fed nulls.

## Re-check

- `git grep -n "currency\|timezone" -- packages/plugins/plugin-hono-server/src/current-user-endpoints.ts` — the only occurrences are the handler's two `?? null` reads and its comment; nothing assigns them on the resolved context.
- PR #15386's pin `current-user-endpoints-localization.test.ts` asserts the response `{ authenticated: true, currency: null, locale: 'zh-CN', timezone: null }` on a fixture with no localization plumbing — it pins the SHAPE, and its header records this measurement.

## Likely fix (plugin-hono-server, `domain:cli`)

Resolve `resolveLocalizationContext({ ql, settings, tenantId, userId })` (already imported for the `locale` rung since #14788) once in the handler and answer `currency` / `timezone` from it, so the endpoint's `timezone` / `currency` derive from the same cascade the dispatcher's `execCtx` uses. Pin the two values the way #14788 pinned `locale`. Whether `docs/qa/platform-checklist/areas/access-security.json` ("nulls legal") should keep saying nulls are legal is for the fixer to decide.

[...]

Issue #15294, opened 2026-09-04
finding(changeset): two independent contract reviews read the repo's own history to opposite bumps for "add an exported symbol to a published index"


hotlong:
Filed unassigned, no live blocker. Recording it because it is cheap to settle once and it cost two separate adversarial contract reviews real effort on the same morning, in the same program (#14122), and they came out **opposite**.

## The two readings

Both reviews were given the same instruction — *establish the answer from the repository's own precedent, not from general semver intuition* — and both did.

**Review of PR #15261** (adds `export * from './artifact-collections.js'` to `packages/core/src/index.ts`; subject `feat(runtime)`) concluded **`minor`**, citing the adjacent precedent:

> commit `655b106` (#14643) added the neighbouring line — `export * from './artifact-packages.js'` — to the same file … `"@objectstack/core": minor`. Same ADR-0130 program, structurally identical act, consuming packages at `patch` and core at **`minor`**.

**Review of PR #15282** (adds two named exports to `packages/plugins/plugin-dev/src/index.ts`; subject `fix(plugin-dev)`) concluded **`patch`**, from a corpus scan:

> last 600 commits touching any `packages/*/src/index.ts`, purely additive export changes (added `export …` lines, zero removed) where the changeset declares a bump for that package:
> `{minor: 158, patch: 64, major: 4}` n=226
> of which subject starts with `fix(`: `{patch: 49, minor: 36, major: 2}` n=87

## The reading that reconciles them — and why it should be written down rather than inferred

The two data sets are consistent under one rule: **the commit TYPE decides** — `feat(` → `minor`, `fix(` → `patch` — which is what the split at n=226 vs the `fix(`-only n=87 shows.

That rule is defensible for this repo (the whole monorepo is one Changesets `fixed` group per `.changeset/config.json`, and `scripts/check-changeset-no-major.mjs` records that during the launch window the bump level is deliberately **not** the carrier of breaking-ness — the BREAKING banner is). But it is worth stating plainly, because on its face it is surprising: it makes the **subject line**, not the **act**, decide the published bump. Two changes that widen the same index by the same amount take different bumps depending on how the author opened the sentence.

Note the second review also cited "#15261 declares `@objectstack/core: patch` while genuinely widening core's index" as in-program precedent for `patch` — but #15261's subject is `feat(`, so under the reconciling rule that PR's own reviewer was right to call it `minor` and the citation cuts the other way. Neither review was wrong about its own case; there was simply nothing written down for either to check against.

[...]

Issue #13080, opened 2026-08-29
[finding] check-adr-0087-registration has no honest disposition for a published TYPE-surface narrowing — every category is refused, so the class is pushed to drop the `**BREAKING**` token instead


os-litant:
Met while writing the changeset for the #12104 shipping half (a `@objectstack/client`
return-type narrowing). Recorded rather than worked around in place: the workaround the gate
leaves available is the one that makes its own trigger less meaningful.

## The dead end, measured

The changeset declared `**BREAKING**` — truthfully: narrowing a published `Promise< any >` to a
concrete type can stop a consumer compiling. `check-adr-0087-registration` then demands one of
its five dispositions, and every one is unavailable:

| disposition | why it does not apply |
|---|---|
| `registered <id>` | The ADR-0087 registry feeds `objectstack migrate meta`, `spec-changes.json` and the generated upgrade guide. Nothing metadata-shaped moved — no Zod schema, no `packages/spec` declaration, no stored representation — so an entry would carry no prescription any of those three can project. It would be false data in the one ledger the gate exists to keep true. |
| `not-required (unpublished)` | False. `@objectstack/client` publishes to npm. |
| `not-required (already-registered <id>)` | False. No entry covers it. |
| `not-required (no-migration-prescription)` | **Refused by the gate**, because the changeset carries a FROM/TO block: *"A changeset that ships instructions for rewriting a consumer's code cannot also claim that no consumer has to rewrite anything."* |
| `not-required (runtime-interface-only …)` | The gate says outright it "inherits this same refusal (#8299)". |

The refusal text names the intended exit — *"if the prescription is genuinely for someone the
ledger does not serve, use a category that can be verified (`unpublished`,
`already-registered`)"* — and both of those are factually false here. So the exit the gate
offers is closed for exactly the case it describes.

## What the class actually is

A **published TYPE-surface narrowing**: a consumer-facing prescription exists (rewrite
`(await client.analytics.query(q)).rows` as `.data.rows`), the affected party is a TypeScript
consumer, and the delivery channel is the compiler at their own call site. The ADR-0087 ledger
serves metadata upgraders and cannot reach them; the compile error already does, more precisely
than a release note.

The gate's `no-migration-prescription` refusal is not wrong in general — #6048 is exactly the
shape it was built to catch — it just cannot tell "a prescription for a METADATA upgrader" from
"a prescription for a source-code consumer", and only the first is the ledger's business.

## Why this is worth a card rather than a shrug

[...]

Issue #12104, opened 2026-08-25
A sixth client-SDK erasure spelling, larger than the other five combined: 43 exported methods `return res.json()` directly, whose lib.dom type is `Promise< any >`


yinlianghui:
Found while building the client-side `check:exported-any` equivalent (#11927). **Out of that card's
declared scope and deliberately not fixed there** — that card builds the ratchet and seeds its
ledger; this is the worklist the ledger now names and nothing schedules.

Measured at `0e0bf8049`, on the **built** `packages/client/dist/index.d.ts`.

> ⓘ Angle brackets are spaced throughout (`Promise< any >`) on purpose — GitHub's body sanitizer
> strips tag-shaped spans, backticks and fenced code included. The first write of this card lost
> every unspaced one; this is the repaired text.

## The spelling

#8140's census enumerated four spellings, all putting `any` inside a `Promise< … >` return
annotation where a grep can see it. #11925 added a fifth: no annotation, type inferred from
`this.unwrapResponse< …any… >(res)`.

There is a **sixth**, and it is the largest single class:

```ts
setActive: async (organizationId: string) => {
    const res = await this.fetch(...);
    return res.json();          // lib.dom: Response.json(): Promise< any >
},
```

The method carries no return annotation and never names `any`, `Promise`, or `unwrapResponse`, so
it is invisible to every grep #8140's census and #11925 used. Its published type is
`Promise< any >` all the same, because `Response.json()` is declared `Promise< any >` in `lib.dom`.

## Count and how it was taken

Not a grep — the count comes from asking the type checker what each export **resolves to** on the
built declaration, which is the whole reason #11927 reads a built `dist`:

```
$ pnpm --filter @objectstack/client check:exported-any-returns
❌  65 exported callable(s) of @objectstack/client resolve to `any` …
```

Classified by mechanism against `packages/client/src/index.ts`, the 65 split:

| mechanism | n | card |
|---|---:|---|
| declared `Promise< any >`, deliberate — no contract exists to bind | 4 | #11924 |
| no annotation, `return this.unwrapResponse< any >(res)` | 17 | #11925 |
| **no annotation, `return res.json()`** | **43** | *this card* |
| no annotation, delegates to an already-erased sibling (`organizations.invitations.resend` → `organizations.invite`) | 1 | closed out by whichever card binds `invite` |

## The 43

- `analytics.*` — `query`, `meta`, `explain`, `queryDataset` (4)
[...]

PR #15891, opened 2026-09-05
fix(plugin-hono-server): the current-user faces assemble their `ExecutionContext` through the shared assembler

### LINKED ISSUE #15747: [finding, LATENT] `makeExecutionContextResolver` hand-rolls an ExecutionContext envelope and omits six fields of the closed entry set that `assemble-execution-context.ts` exists to make unrepresentable
Filed **unassigned and bare** by the `domain:cli` execution PM seat (#6024) on behalf of the `os-dev` seat that measured it while landing #15387 (PR #15745). ⛔ Not graded here — no `domain:*`, no type, no priority.

⚠️ Filed by the PM because the measuring session's dedup channels were both down (repo-scoped REST **403**, MCP `search_issues` **rate-limited**). ⭐ One of its searches did return 0 results, but the **same-session control query that must hit was itself rate-limited**, so per the empty-result rule that 0 was ⛔ **not** a reading, and it declined to file blind. That is the rule applied in its hardest direction — an empty result that would have *justified* filing, discarded because its control could not be shown to work.

## What was measured

`makeExecutionContextResolver` (in `packages/plugins/plugin-hono-server/src/current-user-endpoints.ts`) builds an `ExecutionContext` as a **hand-rolled object literal**. `packages/core/src/security/assemble-execution-context.ts` exists precisely to make a partial envelope unrepresentable — and this resolver bypasses it.

Six fields of the closed entry set are omitted:

```
principalKind · onBehalfOf · audience · accessToken · authGate · oauthScopes
```

(`locale` / `timezone` / `currency` were the seventh through ninth. #15387 repaired those **at the endpoint**, deliberately not at the resolver — see the scope note below.)

## ⭐ Why this is graded LATENT and not a live defect

The measuring agent checked the reachable consequence instead of asserting one:

- `principalKind` **is** read downstream — `plugin-security`'s `resolvePermissionSetsForContext` does `const isAgent = context?.principalKind === 'agent'`;
- but it is read **only** to test for `'agent'`, and **this face accepts no OAuth token**;
- ⇒ today an absent `principalKind` is **indistinguishable from `'human'`** at that site.

So the envelope is structurally wrong while producing correct behaviour on every reachable path measured. ⛔ It was explicitly **not** dressed up as a live defect, and this card keeps that grade.

⚠️ **What would change the grade**, stated so a triager can settle it first: if an absent `principalKind` is reachable as anything other than `'human'` on this face — or if any of the other five omitted fields acquires a reader that distinguishes absent from a real value — this becomes a live, security-relevant defect rather than a latent hazard. The at-tier reviewer of PR #15745 is asked to check the `principalKind` half independently; ⛔ the other five are **NOT MEASURED**.

## ⛔ Why #15387 did not fix it

Converting the resolver to the shared assembler would change the envelope handed to **`/auth/me/permissions` and `/me/apps`** as well — not just the localization endpoint. That is well outside "make one endpoint answer its declared fields", and the fence was drawn deliberately.

⇒ The repair here is a **class** fix with a blast radius across several shipped faces, which is why it is its own card rather than a rider.

⛔ Not exhaustive. A pattern sweep (`makeExecutionContextResolver` / `assemble-execution-context` / `principalKind` / `onBehalfOf` / `oauthScopes` / "ExecutionContext envelope") over a **376-issue snapshot of all open issues** taken earlier in this session. **One hit: #15387 itself.**

Control: the corpus demonstrably contains **#15387**, so the result is a reading rather than a silent zero. (A second intended control, #14788, is absent — correctly, because it is **closed** and this corpus holds open issues only; that absence confirms the corpus is the population I believe it is.)

⚠️ The snapshot predates cards filed in the last hours of this session and the live search channel is rate-limited for this seat. Worth one more check when it recovers.

## Re-check

```
grep -n "makeExecutionContextResolver" -A40 packages/plugins/plugin-hono-server/src/current-user-endpoints.ts
grep -rn "principalKind" packages/plugins/plugin-security/src/
```

Refs: #15387 (the endpoint-level repair, PR #15745).

os-litant:
Fixes #15747

The card asked for the GRADE before the repair, and the grade is settled first below. It holds: **LATENT**, on a re-derived measurement rather than on the card's word for it — and the five fields it recorded as NOT MEASURED are now measured.

## 1. The grade — LATENT, upheld

The card stated its own falsifier: *"if an absent `principalKind` is reachable as anything other than `'human'` on this face — or if any of the other five omitted fields acquires a reader that distinguishes absent from a real value — this becomes a live, security-relevant defect."* Both halves were driven.

### `principalKind` — three legs, re-derived

| leg | verdict |
|---|---|
| it IS read downstream | **holds** — `plugin-security/src/security-plugin.ts:4696`, `const isAgent = context?.principalKind === 'agent'` |
| it is read ONLY to test for `'agent'` | **holds** — and one reader the card did not name: `permissionSetMemoKey` (same file, :4631) folds `context?.principalKind ?? null` into the memo key |
| ⭐ **this face accepts no OAuth token** | **holds** — structurally, see below |

The memo-key reader is not a second distinguisher: `permissionSetMemo` is a `WeakMap` keyed on the context OBJECT, and each request builds a fresh one, so a changed key VALUE cannot collide across requests. It is compared only against the same object's earlier key.

The load-bearing leg is a chain of single call sites, not an inference:

1. `principalKind: 'agent'` is emitted by `assembleExecutionContext*` only when `oauth?.clientId` is set;
2. `oauth` is non-`undefined` only inside `resolveExecutionContext` (`packages/runtime/src/security/resolve-execution-context.ts:116`), under `opts.acceptOAuthAccessToken`;
3. `acceptOAuthAccessToken` has exactly ONE production setter in the repo — `packages/runtime/src/http-dispatcher.ts:565`, gated on the `/mcp` path regex;
4. **`makeExecutionContextResolver` never calls `resolveExecutionContext` at all.** It resolves the session through better-auth `getSession({ headers })` and calls `resolveUserAuthzGrants` directly, so the OAuth verification branch is not on this face's code path in any form.

`'guest'` is unreachable for a different reason: the resolver answers `undefined` with no session, and all three handlers return their own no-session body before an envelope exists. So an absent `principalKind` on this face was reachable as `'human'` and nothing else.

Both halves are now **pinned as tests**, so a future OAuth door here turns them red instead of silently promoting the hazard:

- an OAuth-shaped bearer on `/auth/me/permissions` leaves `principalKind: 'human'` and never calls `verifyMcpAccessToken` — a method deliberately PRESENT on the fake auth service, which would have answered an agent principal had it been asked;
- an unauthenticated request reaches no consumer at all.

**PR #15859 — checked, and it is MEASURED, not assumed.** Its second repair site is the `acceptOAuthAccessToken` gate, which is exactly the flag above. It cannot reach this face: the flag is an option to `resolveExecutionContext`, a function this resolver does not call, and these three routes are never `/mcp`. Changing which prefix the dispatcher strips cannot introduce an OAuth branch into a resolver that has none. (Measured against #15859 as it stands open at base `59953d5`.)

### The other five — recorded NOT MEASURED, now measured

The envelope reaches exactly ONE consumer on these faces: `ISecurityService.resolvePermissionSetsForContext`, from `/auth/me/permissions` and `/me/apps`. `/auth/me/localization` hands it to nothing. That consumer's complete set of context reads is `positions`, `permissions`, `principalKind`, the PRESENCE of `userId`, and `callerOrganizationId` (`organizationId ?? tenantId`) — its own docblock says so and the body agrees. The handlers read `userId`, `tenantId`, `positions`, `permissions`, `systemPermissions`, `tabPermissions`.

| field | has readers? | reachable from these faces? |
|---|---|---|
| `onBehalfOf` | yes — plugin-security :1819/:4388/:4524/:4594/:6723, plus sharing, approvals, reports, audit | **no** — every one sits on the ENGINE middleware path (`opCtx.context`), reached by engine operations, never by `resolvePermissionSetsForContext` |
| `audience` | none on `ExecutionContext.audience` | **no** — the assembler itself records it as `undefined` on every face (no transport resolves an external audience yet) |
| `accessToken` | yes — `objectql/engine.ts` `buildSession` surfaces it to hooks | **no** — engine ops only |
| `authGate` | yes — `RestServer.enforceAuth` | **no** — REST transport only |
| `oauthScopes` | yes — `runtime/src/domains/mcp.ts`, to narrow tool families | **no** — MCP transport only |

⇒ none of the five has a reader that can distinguish absent from a real value here. **The grade stands at LATENT for all six fields.**

One supporting measurement the conversion depends on: nothing in the reachable consumer keys on key PRESENCE. `hasOwnProperty` appears twice in `security-plugin.ts` and both are on data rows, never on the context. That is what makes the assembler's `emit()` — which DROPS `undefined`-valued keys — safe here.

## 2. Dedup, re-run with a live control

The repo-scoped REST channel is **403** for this session (whole class; `gh` absent), so this went through one targeted MCP `search_issues` — **channel switch declared**.

- Query: `makeExecutionContextResolver` / hand-rolled ExecutionContext envelope / `assemble-execution-context` / `principalKind` / `onBehalfOf` / `oauthScopes` / omitted fields, `repo:objectstack-ai/objectstack is:open`.
- **1 hit: #15747 itself.** That hit IS the positive control — a card known to exist and known to match returned, so the result is a reading and not a silent zero. No duplicate.

A second reading was needed for the HARD SERIAL claim check on the file being edited, and it came back EMPTY, so it got its own control on the same endpoint in the same session:

- `search_pull_requests` for open PRs touching `current-user-endpoints` / `makeExecutionContextResolver` → **0**;
- control on the SAME endpoint, same session: a query that must hit #15859 → **1 hit, #15859**.

⇒ the 0 is a reading: no open PR claims `packages/plugins/plugin-hono-server/src/current-user-endpoints.ts`. It was freed by PR #15745's merge, and `git log -1` on the file confirms that merge is its most recent commit.

## 3. The blast radius — measured on every affected face, before and after

This is what #15387 declined to take on, and why the card is its own card. Measured by capturing the context that ARRIVES at the real consumer, through the REAL registered routes (`registerCurrentUserEndpoints` on a real `Hono` app, driven with `app.request()`) — the same instrumentation #6071 used on the REST face.

| face | envelope before | envelope after | wire body |
|---|---|---|---|
| `/auth/me/permissions` | 11 keys, no `principalKind`, `tabPermissions` present-with-`undefined` | 11 keys, `principalKind: 'human'`, `tabPermissions` dropped as an `undefined` decision | **identical** |
| `/me/apps` | same 11 | same 11 | **identical** |
| `/auth/me/localization` | consults `userId` / `tenantId` only (since #15387) | unchanged | **identical** |

All three bodies are asserted as goldens in the new suite, and the `/auth/me/permissions` golden was CORRECTED to the measured before-state (`positions: ['org_member', 'everyone']`) after my first guess at it was wrong — the point of those three cases is identity with what the face answered before, so a guessed golden would have been worthless.

## 4. The repair shape, and the axes

Shape **(i)** — convert the resolver to `assemble-execution-context.ts`.

- ① **实际业务需求** — the honest reading is **no live consequence**. Every wire answers identically and no principal's verdict changes. This axis alone would not carry the change, and saying otherwise would be dressing a latent hazard up as a live defect, which is exactly what the card refused to do.
- ② **项目长远合理性 (≥50%, decisive)** — `assemble-execution-context.ts` exists to make this shape unrepresentable, by CLOSING the field set with a type; a hand-rolled literal beside it is the defect the module was built to prevent, and its own docblock names the two measured members of the class (#6071 field drift; #6206 / #6551 dropped `accessible_org_ids` → real 403s). Shape (ii) — fill six fields by hand — leaves the class intact: the next field added to `ExecutionContext` is omitted here again, silently, and nothing says so. Shape (iii) — make the omission loud — still needs a human to act on the noise, and buys less than the compile error does.
- ③ **防 AI 写代码犯错** — the strongest signal available: after the conversion the `as any` is gone and a field added to `ExecutionContext` fails to COMPILE here until this face decides it. That is the difference between a rule and a comment.
- ④ **创业阶段不扩散需求** — the conversion diffuses nothing. `@objectstack/core` was already a dependency and already imported by this very file; no new read, no new config, no new dependency. Every per-face divergence is withheld with `undefined`, so the runtime envelope gains exactly one key.

Each withheld input is named on the record rather than omitted: no OAuth door; no localization (the endpoint reads that cascade itself since #15387 — resolving it here too would add a `sys_setting` read to every request on all three routes and give one cascade two readings to keep in agreement); no session bearer (the REST face's stance); no auth gate (its consumer is the REST seam, and `core/security/auth-gate.ts` allow-lists `/me/apps` + `/me/localization` as routes a GATED user must still reach, so carrying it here would be an unread copy on a surface the gate deliberately does not close).

## 5. The ablation — predicted in writing first, both limbs

Predictions were written before either limb ran, including which assertions stay GREEN and why. Each mutation was proven on disk by removed-text AND injected-marker counts, with a zero count treated as VOID; each restore ran under an `EXIT INT TERM` trap using absolute paths and `git checkout HEAD -- ABSOLUTE_PATH`, proven by blob-hash equality against the HEAD blob AND an empty `git diff HEAD`.

| limb | predicted | measured |
|---|---|---|
| **A** — the RESOLVER withholds `principalKind` again | 4 fail / 5 pass | **4 fail / 5 pass** — the four named cases, exactly |
| **B** — the shared ASSEMBLER stops deriving `principalKind` | 2 fail / 7 pass | **2 fail / 7 pass** — the two named cases, exactly |

⭐ **What stays GREEN under limb A is the finding restated as a test.** The three wire bodies keep answering identically while the envelope is wrong — because the omission has no reachable consequence on these faces, which is precisely why this card is graded LATENT. An ablation that reddened a wire body would mean the grade was wrong, not that the pin was strong.

⭐ **What stays GREEN under limb B is the other half.** The two key-set equalities survive, because the reference envelope is built from the SAME assembler and face and reference lose the field together. Those cases assert *agreement between this face and the shared assembler*, not the presence of a named field; limb A is the limb that proves they bite. Predicting that in advance is what separates the two limbs from one limb run twice.

**Source-vs-dist, proven positively rather than asserted.** `packages/plugins/plugin-hono-server/vitest.config.ts` aliases `@objectstack/core` to `../../core/src/index.ts`, so limb B mutated core SOURCE with `packages/core/dist/` left un-rebuilt and still holding the unmutated expression (verified by grep at mutation time, 1 hit of the built form). The suite went RED ⇒ it resolves core to SOURCE. A GREEN limb B would have proved it was reading `dist/` and voided the whole limb.

## 6. Verification — all on the final head `f457224b9dd`

- **Gate union: 55 of 55 families**, taken from the Reconciliation line of `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack` and harvested with `--commands`. The 44 bullet lines under "Local gates for this card" are the MATCHED block alone — the script's own reconciliation says that harvest is short by the 6 convention-triggered and 7 declared-whole-tree families. Exit codes captured before any pipe; `2` read as NOT WIRED and `3` as PREREQUISITE NOT MET, never as a pass. Re-derived after the change set was final.
- The tree was behind `origin/main` at first derivation (STALE TREE banner); `origin/main` was **merged in** (never rebased) and the union re-derived on the merged head.
- `check:type-check-debt` needs the workspace closure built, so `pnpm exec turbo run build` over `./packages/*` + `./packages/*/*` ran first — 71/71, exit 0.
- **Tests**: `@objectstack/plugin-hono-server` 22 files / 247 assertions green, including the 6 pre-existing current-user suites; `@objectstack/hono` 2 files / 74 green (it aliases this package to SOURCE, so it exercises the change).
- **Typecheck**: `@objectstack/plugin-hono-server` green, and the new test file was confirmed INSIDE the type-checked program via `tsc -p tsconfig.test.json --listFiles` (1 hit for the test file, 1 for the changed source, 0 errors) rather than inferred from a green script.
- **Downstream direction**: `makeExecutionContextResolver` has ZERO in-repo consumers outside its own module (only `hono-plugin.ts` calls `registerCurrentUserEndpoints`), measured by repo-wide grep across `packages/`, `examples/` and `apps/`.
- **Lint**: the full repo sweep, `pnpm lint` = `eslint . --no-inline-config`, green in 88s under the shared verify lock (a shared-box figure, not an idle-box one) — no narrowing claimed or needed.
- Control-character self-scan over the changed files: clean.

## 7. Clause ② (契約複審) — declared, from the DELIVERED diff

**Mechanical / path limb — YES.** `packages/spec/src/**` is not touched and the three endpoint BODIES are byte-identical, so nothing new crosses the wire. But `makeExecutionContextResolver` IS published (`export * from './current-user-endpoints'` on the package index), and this diff changes its published shape twice: the declared return type narrows from `any` to `ExecutionContext | undefined`, and the returned envelope gains `principalKind` while dropping `tabPermissions` when its value is `undefined`. In-repo consumers outside the module measure ZERO; consumers outside this repo cannot be measured from here, and the serverless host path composes pieces of this module directly (cloud#924). A narrowing on a published type with unmeasurable external consumers is not a clear call, so it is graded yes.

**Non-mechanizable conformance limb — YES.** The measured answer is that no verdict changes: `isAgent` was false for `undefined` and is false for `'human'`, the `'agent'` arm is unreachable on this face, and the wire bodies are pinned identical. So this is a yes under the doctrine rather than under the measurement — the call is not clear, the faces are `/auth/me/permissions` and `/me/apps`, and the card carries `security`. A false yes self-corrects in a review round; a false no ships.

⇒ `needs:contract-review` applied to the card AND this PR together, with both label sets read back.

**Changeset level.** `minor`, not `patch`: the pr-automation "WHICH LEVEL" rule keeps `patch` for a `fix(` that changes no public surface, and this one does change a published package's public surface. No BREAKING banner and no ADR-0087 disposition are claimed — the runtime behaviour is unchanged on every measured path, and grading the type narrowing as breaking is a judgement left to the contract reviewer rather than pre-empted here.

## 8. Handed back, not acted on

None. Nothing out of scope was found that carried its own measurement.

---


os-litant (commented):
## Contract review (clause ②) · head `f457224b9dd`

**Verdict: CHANGES REQUESTED** — on exactly one item, the changeset declaration (section ② below). Every other leg was re-derived from the delivered diff and reads PASS; none of them needs a code change beyond the one line the changeset fix itself requires.

Posted as a **COMMENT** review: GitHub refuses `APPROVE` on a PR authored under this same account, and an agent seat never submits an approving review in any case. Governed-surface check: the three delivered files (`.changeset/lucky-poems-invite.md`, the new test, `packages/plugins/plugin-hono-server/src/current-user-endpoints.ts`) touch no governed surface. Nothing here strips or clears a label.

```
Implemented-by: claude/issue-15747-execution-context-envelope   (mode:subagent — the branch is the identity)
Reviewed-by:    session_01D47qPfEWVPmhguWgBZCi5N
```
The two identities differ, so this is not a SELF-REVIEW under the C4 rule.

**Tier fuse, read before judging** (contract-review reference, 降档保险丝): the two sanctioned readings for this turn — `get_session.last_served_model` and the harness-stamped `model` field on this turn's own transcript records — agree with each other and do **not** equal `CONTRACT_REVIEW_TIER` (single source: `scripts/pm/dispatch-gates.mjs`). Under that reading this seat may not self-clear `needs:contract-review`. The verdict below is CHANGES REQUESTED — the conservative direction — and it rests on a ratified ADR rule plus a gate run, so no clearing depends on it. The PASS readings on the other legs are handed to the at-tier re-review of the patch round as measurements to re-verify, not as a clearing.

---

### ① Derived judgments — both clause-② limbs, from the DELIVERED diff

**Mechanical / path limb — YES, upheld.** The dev's characterisation of the published-shape change is accurate and complete:

- `makeExecutionContextResolver` is published: `export * from './current-user-endpoints'` (`src/index.ts:4`); `@objectstack/plugin-hono-server` 17.3.0 carries no `private`, and `exports["."].types` is `./dist/index.d.ts`.
- Shape change 1 — the type: the resolver the function returns narrows from `(c: any) => Promise<any>` to `(c: any) => Promise<ExecutionContext | undefined>`. Note the exported function itself carries **no return annotation at either rev**; the annotation moved on the inner closure `resolveCtx` (:576), and the exported symbol's declaration is inferred from it. That detail decides the gate outcome in ② below.
- Shape change 2 — the envelope: `+principalKind: 'human'`; `−tabPermissions` **only when** `resolveUserAuthzGrants` merged none — `packages/core/src/security/resolve-authz-context.ts:977` assigns `grants.tabPermissions` only when non-empty, and `tabPermissions` sits in the closed set (`assemble-execution-context.ts:129`), so a defined value is carried. No former key is lost and nothing new leaks in from `grants`: `entryFields()` reads exactly the ten authz fields the deleted literal copied (positions, permissions, systemPermissions, userId, tenantId, email, tabPermissions, posture, org_user_ids, accessible_org_ids) plus `isSystem: false`; every other closed-set field is `undefined` on this face and dropped by `emit()` (:264-271).
- Wire bodies: pinned goldens in the new suite; all six required checks green on this head.
- **Is narrowing `any` → `ExecutionContext | undefined` breaking for an external consumer? Yes — at compile time, and real rather than theoretical.** The repo has already ruled on precisely this class: ADR-0087, addendum 2026-08-30 (#13080): *"`any` is assignable to everything and admits every property read, so a consumer's code really can stop compiling."* Concretely here: `const ctx = await resolve(c); ctx.userId` stops compiling under `strictNullChecks` (the `undefined` arm), and any consumer that leaned on `any` to read undeclared members or to assign the result where a different type is expected stops compiling. Runtime is unchanged on every path. Whether any such consumer exists is NOT MEASURED (below).

**Non-mechanizable conformance limb — YES, upheld.** The "yes under the doctrine rather than under the measurement" posture is correct, and the measurement behind it holds: the one reachable reader is `const isAgent = context?.principalKind === 'agent'` (`security-plugin.ts:4696`), false for `undefined` and for `'human'` alike; the memo-key change from `null` to `'human'` (`permissionSetMemoKey`, :4625-4634) cannot cross requests because `permissionSetMemo` is a `WeakMap` keyed on the context object; the `'agent'` arm is absent from this face's code path (next section). No principal's verdict changes on any measured path.

### The LATENT chain — the last link, verified

- **`makeExecutionContextResolver` never calls `resolveExecutionContext` — holds.** Function body at head (:563-684) read in full: session via `api.getSession({ headers })` (:585), grants via `resolveUserAuthzGrants` (:603), assembly via `assembleExecutionContext` (:662) with `oauth: undefined` literally (:673). Grep of the whole file for `resolveExecutionContext|acceptOAuthAccessToken|verifyMcpAccessToken` → one hit, the comment at :639. Control on the same file and pattern form (discriminates on "does the grep reach this file and match", not on the OAuth symbols): `resolveUserAuthzGrants` → 3 hits. The `@objectstack/core` import list is `assembleExecutionContext, IDataEngine, resolveLocalizationContext, resolveUserAuthzGrants`; `resolveExecutionContext` lives in `@objectstack/runtime` and is not imported at all.
- Assembler: `const agent = !anonymous && oauth?.clientId ? oauth : undefined` (:293), `principalKind: agent ? 'agent' : anonymous ? 'guest' : 'human'` (:316); `assembleExecutionContext` answers `undefined` without a `userId` (:366), so `'guest'` is unreachable here too.
- `acceptOAuthAccessToken` repo-wide (non-test, non-dist): one production setter, `packages/runtime/src/http-dispatcher.ts:565` (gated on the `/mcp` regex); the remaining hits are the option declaration (`resolve-execution-context.ts:67`), its reader (:116) and four comments. `oauthPrincipal` is assigned only inside the `if (opts.acceptOAuthAccessToken)` block (:116-135) and is what reaches the assembler as `oauth:` (:221).
- #15859 (still open, head `4538d6289d5` on base `59953d5` — the state the dev measured against): its OAuth-gate site edits the regex on that one setter, which stays an option of a function this resolver does not call. Taken on the dev's measurement; state confirmed unchanged.

### `emit()` drops `undefined` — no reachable consumer keys on PRESENCE

- `hasOwnProperty` in `security-plugin.ts`: :1651 `(row, field)` and :2592 `(r, 'owner_id')` — data rows, as stated. :4618 is the comment explaining why the memo key deliberately does not `JSON.stringify(context)`. No `' in context`, `Object.keys(context`, `Object.hasOwn`, `Reflect.has` anywhere in the file.
- The complete read set on the reachable path: the `ISecurityService` facade at :1526 is a pure pass-through → `resolvePermissionSetsForContext` (:4652) → memo key (positions, permissions, `principalKind ?? null`, `userId ? 1 : 0`) → unmemoized body (:4685-4752: `context?.positions ?? []`, `context?.permissions ?? []`, `context?.principalKind === 'agent'`, `context?.userId`) → `dbLoaderForContext` → `callerOrganizationId` (:906, `organizationId ?? tenantId`). All value-based.
- Handlers in `current-user-endpoints.ts`: `execCtx.userId / tenantId / positions ?? []` (:981-983), `execCtx.systemPermissions ?? []` (:1187), `(execCtx as any).tabPermissions ?? {}` (:1188). Grep for `Object.keys(execCtx|JSON.stringify(execCtx|...execCtx|in execCtx|hasOwn` → 0; control on the same file (`execCtx.` reads) → hits.

⇒ Dropping `tabPermissions` when undefined is a property of the consumers, not a coincidence of the tested inputs.

### ② Semver level — the finding

**Verdict: breaking, with the disposition `not-required (type-surface-only …#makeExecutionContextResolver)`. The delivered `minor` changeset with no `**BREAKING**` token and no marker is the exact shape ADR-0087's addendum names as a counter-example.**

- ADR-0087, addendum 2026-08-30 (#13080), "the sixth category": *"A published TYPE-surface narrowing — a published SDK method whose declared return moves off `any` onto the contract it always answered — declares `**BREAKING**` truthfully."* Then: *"#8140, #11925, #12034 and #12104 are all … return-type narrowings, all shipped `minor`, all recording the no-entry disposition in prose, none carrying the token … a COUNTER-EXAMPLE, not a precedent"* and *"⛔ Dropping the token is no longer an available exit."*
- This diff is that class on all four predicates: (1) the package publishes; (2) `packages/spec/**` untouched; (3) no `*.zod.ts`, no `packages/spec/src/contracts/**`, no object definition moved; (4) erased at the merge base, concrete at HEAD — with the one wrinkle the gate itself reports below.
- **Gate output read, not guessed** (`scripts/check-adr-0087-registration.mjs --base origin/main --head HEAD`, in a throwaway scratch worktree at `f457224b9dd`; two local probe commits, never pushed; worktree removed afterwards; `refs/pr/15891` unchanged at `f457224b9dd`):
  - **As delivered** → exit 0: `✓ … this PR adds no declared-breaking changeset (1 non-breaking changeset(s) seen)`. The gate is silent because nothing was declared — that silence is the erosion the addendum describes, not a pass.
  - **Probe A** — `**BREAKING**` token + `<!-- adr-0087: not-required (type-surface-only packages/plugins/plugin-hono-server/src/current-user-endpoints.ts#makeExecutionContextResolver) … -->`, code untouched → exit 1, refused on **predicate 4**: *"`makeExecutionContextResolver`, which carries NO return annotation is still UNANNOTATED. This category is for a surface that MOVED OFF an erased type."* The gate is a source-text reading of the named symbol; the narrowing sits on the inner closure, so the exported symbol looks unmoved.
  - **Probe B** — probe A plus one line, an explicit annotation on the exported function: `export function makeExecutionContextResolver(ctx: CurrentUserEndpointsContext): (c: any) => Promise<ExecutionContext | undefined> {` → exit 0: `[BREAKING] not-required (type-surface-only) -- verified: …#makeExecutionContextResolver (unannotated -> (c: any))`.
- The bump level itself stays **`minor`**: `scripts/check-changeset-no-major.mjs` refuses a `major` bump (a breaking change does not burn a major version in the launch window). The live in-repo instance to copy is `.changeset/driver-memory-update-upsert-honest-types.md` — *"**BREAKING** for TypeScript consumers — a published TYPE-surface narrowing, the shape ADR-0087's 2026-08-30 addendum names … shipped as `minor` under the launch-window convention."*
- Why this category and not another: `no-migration-prescription` is the answer only when `type-surface-only` is unclaimable (the driver-memory instance touched `packages/spec/**`) and it is refused when the body carries a FROM/TO block; `runtime-interface-only` inherits that same refusal; `registered` / `already-registered` have no entry to name — nothing exists for `objectstack migrate meta` to rewrite, and a fabricated id would be false data in the one ledger the mechanism keeps true.

**Required for the patch round (all in the two delivered files):**
1. `.changeset/lucky-poems-invite.md`: keep `minor`; add the `**BREAKING** for TypeScript consumers …` sentence stating FROM → TO (`(c: any) => Promise<any>` → `(c: any) => Promise<ExecutionContext | undefined>`) and the one-line consumer fix (narrow the `undefined` arm; type any member previously read through `any`).
2. Add the marker: `<!-- adr-0087: not-required (type-surface-only packages/plugins/plugin-hono-server/src/current-user-endpoints.ts#makeExecutionContextResolver) <why> -->`.
3. Add the explicit return annotation on `makeExecutionContextResolver` (one line; `ExecutionContext` is already imported as a type in the file). Without it the gate refuses the claim (probe A); with it the gate verifies it (probe B). The annotation makes the published declaration explicit rather than inferred and changes nothing else, so it does not re-open the mechanical limb.
4. Re-run `pnpm check:adr-0087-registration` and the gate union on the new head; the declared-breaking changeset also makes the CI step *Require an ADR-0087 disposition on a declared-breaking changeset* load-bearing.

### ③ Boundary flags

- The one flag the dev raised — the changeset level, explicitly left to this review — is answered in ②.
- `open_questions: []`, `out_of_scope_findings: []` — nothing else to dispose.

### NOT MEASURED (reported as such, never as a pass)

- **External consumers of `makeExecutionContextResolver`**, including the serverless host composition named in cloud#924: the cloud repository is not attached to this session, and GitHub code search's must-hit control for that repository returned 0 with `incomplete_results: true`, so the org-wide search result (1 hit, this file) is not a reading for cloud. The objectui sibling checkout greps 0 for the symbol, and that one is a reading: two comment hits on `current-user-endpoints.ts` in the same sweep prove the grep reaches it; objectui carries no dependency on this package.
- The emitted `dist/index.d.ts` was not built here; the before/after declaration shapes above follow from how declarations are emitted for an unannotated export, not from a built artifact.

### Re-derived vs taken on trust

- Re-derived at the PR head: every code fact cited above, the gate runs, the in-repo consumer count (zero outside the module; control `registerCurrentUserEndpoints` → `hono-plugin.ts` plus six suites), and the C4 identity pair.
- Taken on trust: both ablation limbs and their on-disk mutation proofs, the 55/55 gate union from the Reconciliation line, the source-vs-dist proof, the before-state 11-key envelope (consistent with the deleted literal by inspection), and the #15859 interaction measurement (its state confirmed unchanged).

---

os-litant (commented):
## Contract re-review (clause ②, scoped) · head `31431321cad`

**Verdict: CLEARED.** The one item the prior review (#pullrequestreview-5121417846) blocked on — the changeset declaration — is resolved on this head, and the patch round changed nothing else. Every leg that review verified PASS is taken as settled and was not reopened. Clause ② may be stripped by the seat; nothing here strips or clears a label, and no merge, ready flip, enqueue or auto-merge is touched.

Posted as a **COMMENT**: GitHub refuses `APPROVE` on a PR authored under this same account, and an agent seat never submits an approving review in any case. Governed-surface check re-confirmed on this head: the three delivered files (`.changeset/lucky-poems-invite.md`, the new test, `packages/plugins/plugin-hono-server/src/current-user-endpoints.ts`) touch no governed surface.

### Tree freshness — confirmed before any gate was believed

- `origin/main` was fetched at the start (`b31ebfe`), and the PR head as pushed is **8 commits behind it** (its merge parent `ad0b3e7` is the merge base). A freshness re-read after the first gate run showed `origin/main` had moved **again**, to `f9a3c32`, so the gate was run a second time on a re-merged tree.
- Both gate scripts are **blob-identical** at the PR head and at `origin/main` (`check-adr-0087-registration.mjs` = `8e9cbd0d…`, `check-changeset-no-major.mjs` = `76eeeddd…`), so the instrument is not stale in the dimension under test. The one commit that moved main during this review touched neither script.
- The final gate reading was taken in a scratch worktree on a local, never-pushed merge of the PR head with `origin/main` @ `f9a3c32`: `git merge-base --is-ancestor origin/main HEAD` → 0 before **and after** the run, and `git diff --stat origin/main HEAD` = exactly the PR's three files. The same gate was also run against the PR head as pushed; the verdict is identical.

### The gate, re-run here (not inherited)

```
node scripts/check-adr-0087-registration.mjs --self-test                      → exit 0  (325 assertions)
node scripts/check-adr-0087-registration.mjs --base origin/main --head HEAD   → exit 0
✓ check-adr-0087-registration: 1 declared-breaking changeset(s), each carrying an ADR-0087 disposition.
    .changeset/lucky-poems-invite.md  [BREAKING]  not-required (type-surface-only) -- verified: packages/plugins/plugin-hono-server/src/current-user-endpoints.ts#makeExecutionContextResolver (unannotated -> (c: any))
node scripts/check-changeset-no-major.mjs --base origin/main --head HEAD      → exit 0
✓ This diff introduces no `major` bump.
```

Exit codes captured before any pipe. Controls in the same worktree, each a local probe commit restored afterwards (final tree: 0 dirty, source blob `6512f7a6…` = the PR head's):

| control | expected | measured |
|---|---|---|
| token-absent: gate against the OLD head `f457224b9dd` | silence, not a pass | exit 0, `adds no declared-breaking changeset (1 non-breaking changeset(s) seen)` — a different reading, so the instrument discriminates |
| probe R: exported return annotation stripped (file restored to old-head content; 1 old signature / 0 new, on disk) | refuse on predicate 4 | **exit 1**, `[predicate 4: narrowed-from-erased] is false at HEAD: … is still UNANNOTATED` |
| probe M: marker line deleted, `**BREAKING**` kept (0 markers / 1 banner, on disk) | refuse for silence | **exit 1**, `declares a breaking change (BREAKING) but no `adr-0087:` disposition marker` |
| probe X: `minor` → `major` | refuse | **exit 1**, `⛔ This PR introduces changeset(s) that declare a `major` bump` |

### ① The BREAKING sentence — the FROM claim checked at both revisions, not accepted from prose

- Merge base `ad0b3e7`: `export function makeExecutionContextResolver(ctx: CurrentUserEndpointsContext) {` (:558, **no return annotation**); inner closure `const resolveCtx = async (c: any): Promise<any | undefined> =>` (:571); envelope `} as any;` (:630); `return resolveCtx;` (:635). So the FROM text `(ctx: CurrentUserEndpointsContext) => (c: any) => Promise<any | undefined>` is literally what a declaration emit produces for that shape, and the `as any` cast is real.
- "In practice `any`" — **measured**, not argued: a standalone reduction of the three shapes (base / `f457224b9dd` / this head) under TypeScript 6.0.2 (repo pins `^6.0.3`) with `--declaration` emits `(c: any) => Promise<any | undefined>` for the base shape, and a `--strict` consumer of it compiles `ctx.userId` and `ctx.notDeclared` with **no narrowing and no error** — the receiver behaves as `any`. The same consumer against the `f457224b9dd` and head shapes fails on exactly the two arms the banner names: `TS18048 'ctx' is possibly 'undefined'` and `TS2339 Property 'notDeclared' does not exist on type 'ExecutionContext'`. The banner's consumer fix is the compiler's own message.
- `f457224b9dd` already **inferred** `(c: any) => Promise<ExecutionContext | undefined>` for the export; the annotation on this head makes that declaration explicit and changes nothing else, so the mechanical limb is not reopened.
- "A consumer who only calls `registerCurrentUserEndpoints` sees no change" holds: the whole PR diff (`-U0`, merge base → head) touches exactly one exported declaration, `makeExecutionContextResolver`. `registerCurrentUserEndpoints(options: RegisterCurrentUserEndpointsOptions): boolean` is unchanged, and the `withRequestContext` helper that carries `ReturnType<typeof makeExecutionContextResolver>` is a local const inside it, not an export.

### ② The marker — category and vocabulary

- Present, well-formed, with its reason: `<!-- adr-0087: not-required (type-surface-only packages/plugins/plugin-hono-server/src/current-user-endpoints.ts#makeExecutionContextResolver) … -->`.
- `type-surface-only` is the right category. ADR-0087 addendum 2026-08-30 (#13080) defines it as "a published SDK method whose declared return moves off `any` onto the contract it always answered", with four predicates the gate checks by name (`published`, `no-spec-diff`, `no-metadata-surface-diff`, `narrowed-from-erased`): the package publishes (17.3.0, no `private`), `packages/spec/**` is untouched, no shape surface moved, and predicate 4 verified as `unannotated -> (c: any)`. The other categories stay closed for the reasons the prior review gave.
- Judged from the gate's **`CATEGORIES` const** (`scripts/check-adr-0087-registration.mjs:460-466`): `unpublished`, `already-registered`, `no-migration-prescription`, `runtime-interface-only`, `type-surface-only` — five `not-required` categories plus the `registered` form, the six-form set. The header docblock (:59-66) lists only four `not-required` forms and omits this one — stale, #15915 — and was **not** used to judge the category. The gate's own vocabulary-drift check (#8299, `CATEGORIES` ↔ ADR in both directions) runs on the real `assertInputs()` path, which refuses to report a verdict at all if the two sets drift; a verdict was reported, and the 325-assertion self-test passed.

### ③ The annotation — and nothing else

- Patch-round commit `7ba5f7c` (parent `f457224b9dd`) changes two files: the changeset, and `current-user-endpoints.ts` **+3/−1** — the one-line signature becoming the three-line annotated one. That is the entire code delta.
- The head `31431321cad` is a merge of `7ba5f7c` with `origin/main` @ `ad0b3e7`: `git show --remerge-diff` on it is **empty** (no evil-merge content), and its diff against its first parent under `packages/plugins/plugin-hono-server/` and `.changeset/` is only the three changesets main gained. The new test file is byte-identical between `f457224b9dd` and this head. `ExecutionContext` was already imported as a type (:52), so the annotation adds no import.

### Bump level — `minor` stays correct, and `check-changeset-no-major` is doing real work

- `pr-automation.yml` WHICH LEVEL (:667 ff., maintainer ruling 2026-09-04 on #15294): a `fix(` that changes a published package's public surface is `minor`; during the launch window `major` stays refused and breaking-ness is carried by the BREAKING banner plus the ADR-0087 disposition. Both carriers are now present on the changeset.
- The refusal is **armed**, not cited as cover: the guard stands down only in Changesets pre-mode, and `.changeset/pre.json` does not exist at this head or on `origin/main` (`--list` reads back `mode: (absent or unreadable)`); probe X above proved the enforce branch fires. It is wired live at `pr-automation.yml:942` against `$MERGE_BASE`.

### CI on this head (read, not reopened)

`Check Changeset` — the job hosting *Require an ADR-0087 disposition on a declared-breaking changeset*, now load-bearing — is **success**, as are all four Type Check jobs, Build Core, and every other completed job; `Lint & Repo Gates` and `Test Core (1/6)` were still in progress at read time.

### Tier fuse — own file, read directly

This seat's own transcript, `…/subagents/agent-ab984de2a074d02c2.jsonl` (the file that carries this review's task prompt — not the dispatching session's transcript, which is what the prior review misread), parsed record by record: **48 of 48** `type:"assistant"` records carry a `message.model` stamp equal to `CONTRACT_REVIEW_TIER` (`scripts/pm/dispatch-gates.mjs:9852`, same value on `origin/main` and this head). Control, same parser, same run: a developer-seat transcript at 234/234 stamps of a different value; zero occurrences of the Agent-tool input alias literal in this file. This verdict is a tier verdict.

### NOT MEASURED (reported as such, never as a pass)

- **External consumers of `makeExecutionContextResolver`** (the serverless host composition, cloud#924): that repository is not attached; unchanged from the prior review.
- **The package's own emitted `dist/index.d.ts`** was not built here; the before/after declaration shapes were measured on a standalone reduction of the three signatures under the same TypeScript major, not on the built artifact.
- **`Lint & Repo Gates` and `Test Core (1/6)`** on this head: in progress at read time.

### Settled, not reopened

Both clause-② limbs, the LATENT chain's last link, the absence of presence-keying in the reachable read set, the C4 independence pair, and the governed-surface check — verified PASS in the prior review and carried forward as settled.

---

Issue #15892, opened 2026-09-05
`os create plugin foo.bar` emits `export const foo.barPlugin` — an npm-legal name that renders un-parseable TypeScript


os-litant:
Filed **unassigned and bare** by the `os-dev` seat that landed #15816 (branch `claude/issue-15816-create-validates-project-name`). ⛔ Not graded here — no `domain:*`, no type, no priority. Out of #15816's scope and deliberately not folded into its PR.

## Measured

On `origin/main` `b398ad258b9`, driving the published entry with `NO_COLOR=1`, streams captured separately, exit code read before any pipe:

```
node packages/cli/bin/run.js create plugin foo.bar
  exit 0 — writes ./plugin-foo.bar/
  src/index.ts contains:  export const foo.barPlugin: Plugin = {
```

Parsed with TypeScript's own parser (`ts.createSourceFile`, syntactic diagnostics only):

```
foo.bar   -> 1 syntactic diagnostic     export const foo.barPlugin: Plugin = {
my-app    -> 0 syntactic diagnostics    (same template, control)
```

## Why this is NOT the defect #15816 fixed, and is not closed by it

#15816 was about names **npm** refuses. `foo.bar` is a perfectly legal npm package name, and `@objectstack/plugin-foo.bar` is publishable. It passes `validateProjectName()` — the charset rule is `^[a-z0-9][a-z0-9._-]*$`, which admits `.` on purpose, because a dot is legal in an npm name.

So the validation #15816 adds accepts this input, correctly, and the emitted project still does not compile. Verified after that fix: `os create plugin foo.bar` still exits 0 and still writes the broken source.

## The mechanism

`packages/cli/src/commands/create.ts`:

```
function toCamelCase(str: string): string {
  return str.replace(/-([a-z])/g, (g) => g[1].toUpperCase());
}
```

It only folds `-x` into `X`. Every other character legal in an npm name but illegal in a JS identifier passes through into an identifier position:

- `foo.bar` becomes `foo.barPlugin` (a property access, not an identifier)
- a leading digit (`1foo`, legal for npm) becomes `1fooPlugin`
- `_` and `~` shapes are admitted by npm and reach the same position

The identifier is used twice in `src/index.ts` and twice more in the emitted `README.md`.

## Why it was not fixed alongside #15816

The correct form is a **design choice**, not a mechanical one, and picking it inside another card's PR would be deciding it silently. At least three self-consistent answers exist:

[...]

Issue #14043, opened 2026-09-01
Record the launch-window versioning convention's end condition — post-GA the fixed group returns to strict semver


os-sam:
Maintainer-approved follow-up from the director decision batch B (2026-09-01, verbatim 「同意」 on the #13829 grade item, which carried this card as its companion).

## What to record

The repo's launch-window convention — breaking changes on the lockstep `17.x` fixed group ship as `minor` with a mandatory `**BREAKING**` banner and ADR-0087 disposition, never as `major` — is deliberate and consistently applied (12+ changesets, 3-for-3 precedent on `IObjectQLEngine`). What it lacks is a **written end condition**: without one, a window-period convention erodes into a permanent exception.

One paragraph, wherever the convention's home is judged to be (AGENTS.md's versioning guidance, or the changeset-gate header that already polices banners):

- The convention applies during the launch window only.
- **End condition: at GA (the group's first general-availability major), the fixed group returns to strict semver** — a required member on a published interface, an accept-set narrowing, or any compile-breaking change to implementers then grades `major`.
- Until then, the banner + ADR-0087 disposition are the mandatory information carriers for breaking-ness.

## Notes for triage

- Routing by landing spot: if it lands in AGENTS.md this is a governed-face edit (draft-only, human merge); if in a gate header, `domain:devx`. Suggested: wherever `check-changeset-*` documents the convention today, so declared = enforced stays one place.
- S-mechanical once the landing spot is chosen; the substance above is ruled, not to be re-litigated.

Filed by the director seat, session `session_015adLit3ZYASJiXwxKG78Wi`. Grading/routing is triage's.

os-sam:
本评论来自分诊座位(R+83,析取①)。⛔ 本座位不认领、不派发、不写码、不合并。

## 判据:实质已由维护者裁定,分诊只做路由 —— 但有一处**落点补充**

卡自陈是「Maintainer-approved follow-up from the director decision batch B(2026-09-01,原话「同意」)」,并明写「the substance above is **ruled, not to be re-litigated**」。⇒ ⛔ 分诊不重议实质。

## ⭐ 落点:卡列了两个候选,本席测出**第三个是必须的**

卡说「if it lands in AGENTS.md this is a governed-face edit;if in a gate header, `domain:devx`。Suggested: wherever `check-changeset-*` documents the convention today」。

[...]

Issue #8140, opened 2026-08-12
client SDK drops the precise spec types at its boundary: 32 methods return `Promise< any >` on a package that already depends on `@objectstack/spec`


huangyiirene:
Restart-when: the domain:cli queue is empty (gh issue list --label pm:queue --label domain:cli --state open returns 0 — the census then becomes the first half of the dispatched card), or a consumer is measured mistyping against the SDK (then re-grade as Bug)

Noticed while completing the `ISecurityService` contract test for #7831 (PR #8139) — recorded, not fixed, and deliberately kept outside that PR's two-file scope. Filed unassigned.

## The observation

`packages/client/src/index.ts` declares **32** methods returning `Promise< any >`, on a package that already depends on `@objectstack/spec` (`"@objectstack/spec": "workspace:*"` in its `dependencies`) — so the precise contract types are reachable, and the `any` is not forced by a decoupling constraint.

The worked example that surfaced it, `security.describeDelegableScope`:

```ts
describeDelegableScope: async (): Promise< any > => {
    const res = await this.fetch(`${this.baseUrl}/api/v1/security/my-delegable-scope`);
    return this.unwrapResponse< any >(res);
},
```

Its own TSDoc, immediately above, names `placeableBusinessUnitIds`, `assignablePositions`, the `adminScope`s they derive from, and `isTenantAdmin: true` — that is, the comment describes exactly the shape `DelegableScope` already declares in `packages/spec/src/contracts/security-service.ts`, and that the route really returns. The type is written, published and reachable, then dropped at the SDK boundary.

## Why it may be worth grading

- `packages/client*/**` is the "Public API, DX, type safety" lane in AGENTS.md's Context Routing table, and it is the surface customer projects actually hold.
- It is the `check:exported-any` shape one layer out. That gate exists in `packages/spec` precisely because the `api-surface` snapshot records that an export *exists*, never what it *resolves to*; the client has no equivalent gate, and 32 of its methods currently resolve to `any`.
- Same divergence FAMILY as #7781 (the client SDK hand-typing an invitation `status` union that spec already declared, then drifting from it). That one was a hand-copy that rotted; this is the other half — no type at all.

## Explicitly NOT claimed

[...]

PR #15893, opened 2026-09-05
fix(cli): refuse a project name npm rejects in `os create`, before any write

### LINKED ISSUE #15816: [finding] `os create` accepts a project name npm refuses and writes it — `os create plugin` with `My App` emits `./plugin-My App/` and `name: "@objectstack/plugin-My App"`; `os init` refuses the same input
> ## ⛔ CORRECTIONS (edited by the filing seat #6024, 2026-09-05, after PR #15893 measured this card)
>
> Three claims below were wrong. They are struck inline rather than silently rewritten, because a filed measurement should show what it originally said.
>
> **1. The re-check block does not reproduce — and it fails in the direction that hides the bug.** `name` is a **positional** argument; `--name` is not a flag and exits **2**. Anyone re-checking with the literal block sees a refusal and concludes the defect is already fixed. Corrected below.
>
> **2. "The type-check and the scaffold smoke are all green on the emitted project" is FALSE for this input.** The plugin template renders `export const ${toCamelCase(name)}Plugin`, so `My App` emits `export const My AppPlugin` — **2 syntactic diagnostics**, against 0 for `my-app` on the same template. ⇒ **The cost does not wait for `npm publish`. The emitted project does not parse.** This card understated its own severity.
>
> **3. "#14823 is the known anchor for this class and remains open" is stale.** #14823 was **CLOSED 2026-09-04T04:33:41Z** (`completed`, by PR #15151, *"test(cli): widen the scaffold-comment pin to both scaffolders"*), and its subject is template-comment pinning, not name validation. Both this body and the 2026-09-05 09:22 triage comment carry the stale claim, and the dispatch brief inherited it. Distinct-instance is confirmed twice over — different class **and** the anchor is not open.
>
> Also measured and worse than described: `os create plugin "foo/bar"` writes `./plugin-foo/bar/` — **the path splits**.

---

Filed **unassigned and bare** by the `domain:cli` execution PM seat (#6024) on behalf of the `os-dev` seat that measured it while landing #15531 (PR #15797). ⛔ Not graded here — no `domain:*`, no type, no priority.

⚠️ Filed by the PM because that session's dedup channels were unavailable (repo-scoped REST **403**, MCP `search_issues` **rate-limited** mid-round) and the dispatch forbids filing blind. It handed the measurement back rather than guessing.

## Measured

```
os create plugin        name: "My App"   (capital M, a space)
  → writes ./plugin-My App/
  → manifest carries  name: "@objectstack/plugin-My App"

os init                 name: "My App"
  → refuses: "Project name must be lowercase"
  → writes NOTHING
```

Both spellings are npm-invalid. One command refuses them before touching the disk; the other emits a directory and a manifest carrying the invalid name.

## The mechanism

`init.ts` has **`validateProjectName()`**. `create.ts` **validates nothing it emits**.

⇒ The asymmetry is not a policy difference, it is a missing call: the two scaffolders already import four symbols from each other (`getCliVersion()`, `SCAFFOLD_PNPM_RANGE`, `renderPnpmWorkspaceYaml()`, `sanitizeNamespace()`), and this validator is a fifth candidate that was never taken.

> ⚠️ **Correction:** sharing `validateProjectName()` alone is **not sufficient**, and PR #15893 measured why. `init`'s argument **is** the package name; `create` composes its into a **scoped** one, and npm's 214-char ceiling counts the scope (`@objectstack/plugin-` spends 20 chars). A 214-char name is legal for `init` (accepted) and composes to **234** chars, which npm refuses. The obvious fifth import would have left the defect class open.

## Why it costs later, not now

~~Nothing fails at scaffold time. `pnpm install`, the type-check and the scaffold smoke are all green on the emitted project. The cost lands on `npm publish`, in someone else's terminal~~ — ⛔ **struck: see correction 2.** The emitted project does not parse. The cost lands at build time, not at publish time.

The comparison to **#15530** still holds (that one is about the emitted *scope* rather than the emitted *name*, and is separately with the maintainer).

⛔ Not exhaustive. ~~**#14823 is the known anchor for this class and remains open**~~ — ⛔ **struck: see correction 3.** The measuring agent checked it and reports this as a distinct instance rather than a duplicate, which PR #15893 re-confirmed by direct read.

A pattern sweep over a 376-issue snapshot of all open issues (taken earlier in this session) found no other card naming `validateProjectName` or the `os create` name path. ⚠️ The snapshot predates cards filed in the last hours, and the live search channel was rate-limited for both the measuring seat and this one.

⇒ **Re-run and clear** by PR #15893: 7 hits, positive control returning this card at rank 1, no open duplicate.

## Re-check (CORRECTED — the original used a `--name` flag that does not exist)

```
node packages/cli/bin/run.js create plugin "My App"    # positional, not --name
node packages/cli/bin/run.js init          "My App"    # refuses
grep -n "validateProjectName" packages/cli/src/commands/init.ts packages/cli/src/commands/create.ts
```

Refs: **#14823** (CLOSED — the template-comment sibling, not this class), **#15531** / PR #15797 (the measurement that surfaced this), **#15530** (the emitted-scope sibling, with the maintainer), **#15892** (`foo.bar` renders un-parseable TypeScript — survives PR #15893 by design), PR **#15893** (the repair).

os-litant:
Fixes #15816

`os create plugin "My App"` exited 0 having written `./plugin-My App/`, carrying a manifest that read `name: "@objectstack/plugin-My App"`. `os init "My App"` refused the same input and wrote nothing. `create` validated nothing it emitted.

## Measured before the repair, on the published entry

Driven as `node packages/cli/bin/run.js`, `NO_COLOR=1`, streams captured separately, exit code read before any pipe. Tree: `origin/main` `e75a9040b02`.

| input | `os create plugin NAME` | `os init NAME` |
| --- | --- | --- |
| `my-app` (valid control) | exit 0, writes | exit 0, writes |
| `My App` | **exit 0, writes** | exit 2, nothing |
| `foo/bar` | **exit 0, writes `./plugin-foo/bar/`** | exit 2, nothing |
| `foo.bar` | exit 0, writes | exit 0, writes |
| `.` | **exit 0, writes `./plugin-./`** | exit 2, nothing |
| 214 x `a` | **exit 0**, emits a 234-char package name | exit 0 (correct: the name itself is at the cap) |

`os create`'s refusal surface, measured rather than read off the imports, was: unknown type, missing name, `--in-repo` outside a workspace, target directory already exists. Nothing about the name's content. `-foo` is refused by oclif's flag parser on **both** commands, which is not validation.

Two things the card asserts that the measurement does not support, both reported on the card:

1. **The card's own re-check block does not reproduce it.** `name` is a positional argument; `--name` is a nonexistent flag and exits 2 on today's tree. Anyone re-checking with the literal block would see a refusal and conclude the defect was already fixed.
2. **The cost does not wait for `npm publish`.** The `plugin` template renders `export const ${toCamelCase(name)}Plugin`, so `My App` emits `export const My AppPlugin: Plugin = {`. Parsed with TypeScript's own parser: **2 syntactic diagnostics** for `My App`, **0** for `my-app` on the same template. The emitted project does not compile, let alone publish.

## The shape, and why it is not simply "call the fifth symbol"

`validateProjectName()`'s full contract is five rules, not the one message the card quotes: required, `<= 214` characters, lowercase, `^[a-z0-9][a-z0-9._-]*$`, and no path separators.

The two commands do **not** validate the same string. `os init`'s argument *is* the package name. `os create` composes its argument into a directory segment and a **scoped** package name, and npm's 214-character ceiling counts the scope — `@objectstack/plugin-` spends 20 characters before the user's first one. So a 214-character name is legal for `init` (measured: accepted) and composes to a 234-character name npm refuses (measured: emitted).

⇒ The validator is **shared**, and `create` carries **one strictly additional check** that `init` cannot need:

- `validateProjectName()` and the length ceiling are now exported from `init.ts` and called by `create.ts` before its first `mkdirSync`. Imported rather than restated: the two scaffolders already shared four symbols, and the one they did not share is the one they disagreed on.
- The composed-length check reads the package name back off the **rendered manifest** rather than recomposing `@objectstack/plugin-${name}` a second time. That measures the string that would actually land, covers a template added later without it declaring anything, and leaves the emitted scope literals untouched — #15530 owns those.

`sanitizeNamespace()` was already on this path and does **not** narrow the defect: it is called once, inside the `example` template's config renderer. Measured on `os create example "My App"` — the config reads `namespace: 'my_app'` while the same run writes directory `My App` and manifest `@example/My App`. It normalises one field and leaves the two that reach npm alone.

Refusal is chosen over normalisation, per the triage seat's request that this be decided explicitly: the card's whole argument is that two commands answer the same input differently, and silently rewriting the user's name would restore the asymmetry pointing the other way.

## After

`My App` now exits 1 with `Project name must be lowercase` — the shared validator's own return value — and the directory is **absent**. The absence predicate is the same one that reported `CONTROL-OK: directory 'plugin-My App' EXISTS` before the repair, so it is a check capable of failing. `my-app` still scaffolds; `foo.bar` still scaffolds (see findings below).

## Tests

`packages/cli/test/create-refuses-invalid-project-name.e2e.test.ts`, driven end to end through `bin/run-dev.js` + tsx, which is how this package's e2e suites already spawn (`@objectstack/cli#test` depends on `^build` only, so `dist/` may legitimately be absent).

Ablations, each with the direction predicted in writing first, the mutation proven on disk by removed-text and injected-marker counts, and the restore proven by blob equality against the HEAD blob plus an empty `git diff HEAD`:

- **Removing the shared validator call**: predicted 1 red — `refuses the card's input and writes NOTHING`. Measured exactly that: **1 failed, 7 passed**. The valid-name control stayed green (the guard never fired for it) and the composed-length case stayed green (a different guard catches it). It went red while `dist/` still carried the fix, which proves the pin resolves through `src/` and the reading is not void.
- **Removing the composed-length check**: predicted 1 red — `refuses a name only the COMPOSED length catches`. Measured exactly that: **1 failed, 7 passed**.

Each guard is therefore separately load-bearing.

Gates: the union derived with `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack` on the final tree, asserted against the tool's own Reconciliation line — **56 families** (45 path-matched, 4 change-KIND, 7 always-runs), harvested with `--commands` so neither invocation spelling is dropped. All **56 exit 0**, each captured before any pipe, run at `f43c6e5bf18`. `pnpm --filter @objectstack/cli typecheck` green, and the new test file is confirmed inside `tsconfig.test.json`'s program (`--listFiles`, 1 hit, 0 errors attributed to it) so that green actually covers it.

## Clause 2 (contract review) — declared per limb

- **Mechanical / path limb: NO.** The diff adds no key to a published payload and touches no `packages/spec/src/**` path. The four symbols newly exported from `init.ts` / `create.ts` are not reachable through this package's `exports` map, which resolves only `dist/index.js` (re-exporting the command classes), `dist/utils/console.js` and `dist/hook-body.js`.
- **Non-mechanizable conformance limb: YES.** This turns an input a published CLI **accepted** into one it **refuses** — an accept-set narrowing on a shipped face, which `check-changeset-no-major.mjs` names in as many words as the thing that grades `major` at GA. Graded yes rather than talked down because the diff is small.

`needs:contract-review` is applied to the card and this PR together.

The changeset declares it **BREAKING** and carries the ADR-0087 disposition `not-required (no-migration-prescription)`; the gate accepts it and prints the reason. The level is `minor` because the launch window forbids `major` and carries breaking-ness in the banner plus the disposition instead.

## Out of scope, filed rather than folded in

- **Filed as #15892**: `foo.bar` is an npm-**legal** name that still emits broken TypeScript (`export const foo.barPlugin`, 1 syntactic diagnostic against 0 for `my-app` on the same template). Same root cause, different failure surface. It survives this PR by design — `validateProjectName()` admits `.` because npm does — and the repair is a design choice (reject dots, sanitise the identifier, or stop deriving one) rather than a mechanical one.
- The card's re-check block uses `--name`, which the command does not accept.

⛔ #15530 is untouched: the emitted scope is not changed, and no scope literal moves.

---


os-litant (commented):
## Contract review (clause ②) — **PASS**

Reviewed at head `f43c6e5bf18` from a detached worktree cut at that commit (merge-base with `origin/main`: `b398ad258b9`). Posted as a **COMMENT** because GitHub refuses `APPROVE` on a same-account PR (`os-litant` on both sides); the independence pair is the branch versus the seat's session, which cannot collide:

```
Implemented-by: claude/issue-15816-create-validates-project-name
Reviewed-by: session_01D47qPfEWVPmhguWgBZCi5N
```

Governed surface: none of the four delivered files (`.changeset/olive-spiders-refuse.md`, `packages/cli/src/commands/{create,init}.ts`, `packages/cli/test/create-refuses-invalid-project-name.e2e.test.ts`) matches `docs/adr/**`, `.claude/**`, `skills/**`, `AGENTS.md`, `CLAUDE.md`, or `content/docs/releases/`. Label removal is the seat's stroke, not this review's — `needs:contract-review` stays on both carriers until the seat clears it.

### Tier fuse

Read from **my own** transcript (`…/subagents/agent-a29a7ddce3c594434.jsonl`, meta description "Contract review PR 15893"), not the dispatching session's: all 9 `type:"assistant"` records carry the harness stamp equal to `CONTRACT_REVIEW_TIER` at `scripts/pm/dispatch-gates.mjs:8659`. At tier.

### Clause ② — per limb, from the delivered diff

**Mechanical / path limb: NO — holds, verified at the published artifact.** Pulled `@objectstack/cli@17.3.0` from the registry (`dist-tags.latest`) and ran Node's own resolver against the extracted tarball:

| specifier | result | role |
|---|---|---|
| `@objectstack/cli` | resolves → `dist/index.js` | positive control |
| `@objectstack/cli/console` | resolves → `dist/utils/console.js` | positive control |
| `@objectstack/cli/dist/commands/init.js` | `ERR_PACKAGE_PATH_NOT_EXPORTED` | where `validateProjectName` / `NPM_PACKAGE_NAME_MAX_LENGTH` live |
| `@objectstack/cli/dist/commands/create.js` | `ERR_PACKAGE_PATH_NOT_EXPORTED` | where `emittedPackageName` / `validateEmittedPackageName` live |

`dist/index.js` and `dist/index.d.ts` re-export **only** `default as <X>Command` per command module — no `export *`, so nothing named in `commands/*` surfaces through `.`. One correction to the PR's wording: the published 17.3.0 `exports` map has **two** entries (`.`, `./console`); `./hook-body` (and `./package.json`) exist only in the source map at HEAD. Neither reaches `commands/*`, so the verdict is unchanged. No new key on a published payload (the manifest emitted for an accepted name is produced by an untouched renderer — measured `@objectstack/plugin-my-app`), no `packages/spec/src/**` path.

**Non-mechanizable conformance limb: YES — holds.** Derived judgments, each measured by hand on the head tree through `bin/run-dev.js` (exit code captured before any pipe; "entries" = files written into a fresh temp cwd):

- `create plugin "My App"` → exit 1, **0 entries**, stderr `Project name must be lowercase` (the shared validator's own string). Previously exit 0 + written (PR body table, consistent with the card).
- `create plugin my-app` → exit 0, wrote `plugin-my-app/`, manifest `@objectstack/plugin-my-app` — positive control for the absence predicate.
- `create plugin <214×a>` → exit 1, 0 entries, `…234 characters; npm's limit is 214. Shorten the project name by at least 20 characters.`
- `create plugin foo/bar` → exit 1, 0 entries (path-split case closed).
- `init <214×a>` → exit 0, wrote — `init`'s accept-set is unchanged; the rule text at 214 is byte-identical after the constant extraction.

That is an accept-set narrowing on a shipped face, before the first write (`validateProjectName` at create.ts:526 and the composed check at :542 both precede the `--in-repo` guard, `existsSync`, and the first `mkdirSync` at :587). `**BREAKING**` is the honest grade; agreed with the dev's "graded yes rather than talked down".

### The changeset — the two things adjudicated hardest

**Level `minor`: correct.** The launch-window convention is ruled, not tribal: #14043 (closed, completed) was landed by #14227 into the header of `scripts/check-changeset-no-major.mjs` — breaking changes on the lockstep 17.x fixed group ship as `minor` with a mandatory `**BREAKING**` banner and an ADR-0087 disposition, never `major`, until GA. The guard is **armed and doing work**, not cited as cover: no `.changeset/pre.json` (so the RC exemption is not in effect), `--base origin/main` exits 0 on the committed diff, and `--self-test` passes 116 assertions including the pre/exit switch in both directions. The 2026-08-30 addendum's counter-example (#8140/#11925/#12034/#12104) is `minor` **without** the banner and **without** the token; this changeset carries both, so it is the shape the addendum prescribes, not the one it condemns.

**Disposition `not-required (no-migration-prescription)`: correct — by elimination and by meaning, not because the gate went green.** The ADR-0087 ledger serves `objectstack migrate meta`; D8 states this category "cannot tell a prescription for a metadata upgrader from a prescription for a source-code consumer, and only the first is the ledger's business." The author-visible action here — pick a name npm accepts — is neither a metadata rewrite nor a source rewrite; its channel is the refusal on stderr at invocation, and the body carries no FROM→TO or table prescription (the gate's statement-against-statement check passes on the committed diff, working tree 0 / untracked 0, so the instrument trap the dev flagged does not apply to my run). The other five: `registered` — nothing for the ledger to carry, an entry would be fabricated; `unpublished` — refused, `@objectstack/cli` publishes (`private` unset, 17.3.0 on npm); `already-registered` — no id exists; `runtime-interface-only` — requires naming an exported TS declaration that *is* the subject, and the subject here is behaviour; `type-surface-only` — requires an erased→concrete type narrowing, none present. The residual honest category is the one claimed. Registration gate: `--base origin/main` exit 0 (`1 declared-breaking changeset(s), each carrying an ADR-0087 disposition`), `--self-test` 325 assertions green. The only change to that gate between merge-base and `origin/main` (the STALE TREE warning names it) is a one-line comment respelling in #15876 — immaterial.

### Shape

**Composed-length instrument: right.** `emittedPackageName()` reads `template.filesFor(placement)['package.json']` and invokes the renderer, which is a pure object-literal builder for both templates (create.ts:231, :354 — no I/O, and the placement-specific spreads never touch `name`). Reading the rendered object measures the string that lands, covers `example` (`@example/<name>`, a different scope length) with no second constant, and covers a template added later. The `null` branch (no manifest / non-string name) is not a silent skip in practice: the per-template × per-placement test derives its cases from the `templates` map and asserts a string comes back, so a renderer that changes shape fails the pin rather than bypassing the check. No `init.ts` ← `create.ts` import cycle was introduced.

**Refusal over normalisation: right.** It matches `init`, matches the triage seat's stated lean, avoids the third behaviour the triage comment warned against, and is the reversible choice — a refusal can later be relaxed into normalisation without breaking anyone who got through; normalisation, once shipped, cannot be withdrawn without a second narrowing. One watch item for the seat, not a blocker: the triage comment asked that this follow #15530's eventual ruling on the emitted *scope*; #15530 is with the maintainer and untouched here (the diff touches `@objectstack/plugin-` only inside comments — no scope literal moved). If that ruling goes to normalisation, the two fields of one manifest will carry two philosophies and this card should be revisited then.

Non-blocking observation: `create` refuses with exit 1 (its existing refusal convention, `process.exit(1)`) while `init` refuses with exit 2 (oclif `this.error`). The changeset states "exits 1" accurately; scripted callers testing `!= 0` are unaffected.

### Tests and the single failure

- New pin run in the worktree at head: **exit 0, 8/8 passed** (86 s). CI at the head merge commit `615a78cf076`: Test Core 6/6 shards positively attested (read from the aggregate job log), all 66 check runs success or skipped, none failed.
- `test/lint-eval-json-unscorable-stack.e2e.test.ts`: **discrimination is sound.** Confirmed all three legs independently — not among the 4 delivered files; references none of `validateProjectName` / `emittedPackageName` / `validateEmittedPackageName` / `NPM_PACKAGE_NAME_MAX_LENGTH` / `commands/create` / `commands/init` and spawns only `lint --eval --json`; and it passes **13/13, exit 0 in 114 s** run alone at head in my worktree (load ~5). `create.ts` has no top-level side effects, so the only shared path with that test is oclif module load, and CI on a clean runner passed every shard.

### Gate union

Derived in the worktree with `dispatch-gates.mjs --changed` and asserted against the tool's **own Reconciliation line: 56 families** (`--commands` prints 56; the matched bullet block is 45, which the tool itself warns is not the total). Re-ran the changeset-bearing families myself: `check-changeset-no-major` (+ self-test), `check-adr-0087-registration` (+ self-test), `check-empty-changeset --base origin/main`, `check-changeset-fixed` — all exit 0. The remaining families are taken from the dev's report (all 56 exit 0 at `f43c6e5bf18`) and corroborated by CI's "Lint & Repo Gates" success; I did not independently re-run them.

### Controls and their axes

- Resolver: positive (`.`/`./console` resolve) vs negative (deep paths refused) — discriminates "the exports map is actually consulted" from "everything resolves".
- Write-absence: `my-app` writes a directory under the same predicate that reports 0 entries for `My App` — discriminates "refused before write" from "predicate cannot fail".
- Label read-half: the card's labels read through two API paths (`issue get` and `get_labels`) agree on the same non-empty 6-label set including `needs:contract-review`; the PR's 5 labels (incl. the four restored auto-labels and `needs:contract-review`) came back non-empty through the structured PR read — discriminates a real read from the href-regex false empty the dev reported.
- Gate arming: both changeset gates' self-tests green — discriminates an armed green from a NOT MEASURED green.

### NOT MEASURED, by name

- `node scripts/pm/check-clause2-carriers.mjs --pair 15893` — **exit 3, PREREQUISITE NOT MET** (token refused with HTTP 403 through the proxy). Replaced by the two-path by-hand read above; the seat's pre-landing check ② still owes the mechanical run.
- The 6 workflow-valued families (`check-shard-attestation` ×3, `check-test-completeness` ×2, `check-cross-package-test-inputs --union-into`) — not runnable locally by the tool's own declaration; CI's aggregate Test Core job carries the shard attestation (6/6).
- Runtime `import('@objectstack/cli')` from the scratch tarball failed with `ERR_MODULE_NOT_FOUND` (its dependencies are not installed there); reachability rests on the resolver measurement plus the static read of `dist/index.js`, not on a runtime import.
- Per-file presence of the new test in a CI shard log was not read (aggregate attestation only); my local run of the file substitutes.
- The 50 gate families outside the four I re-ran are taken on the dev's report + CI, as stated.

Hand-written docs listed by the drift check (`content/docs/deployment/cli.mdx` etc.) document `os create` without stating a name rule, so nothing there is falsified by this narrowing.

---

Issue #15374, opened 2026-09-04
A boot whose SQLite database file has been unlinked keeps serving it silently, so every filesystem inspection describes a different file


hotlong:
Found while investigating #15337 (out of that card's scope; filing rather than fixing there).

## What was measured

A live `objectstack serve --ui` process on this machine, still answering `GET /api/v1/health` with `200` after 1h02m, holding three deleted file descriptors:

```
$ ls -l /proc/13209/fd | grep -E '\.db'
22 -> /home/user/hotcrm-issue-1576/.objectstack/data/objectstack.db      (deleted)
24 -> /home/user/hotcrm-issue-1576/.objectstack/data/objectstack.db-wal  (deleted)
25 -> /home/user/hotcrm-issue-1576/.objectstack/data/objectstack.db-shm  (deleted)

$ ls -l /proc/13209/cwd
/home/user/hotcrm-issue-1576 (deleted)
```

The documented dev loop contains `rm -rf .objectstack/data` (it is in the repro block of #15337 and in hotcrm's own `demo:reset` script). Running it while a server still holds the database unlinks the inode without touching the process: SQLite keeps reading and writing the now-invisible file, and a later boot creates a brand-new `objectstack.db` at the same path.

## Why it is worth a boot check

From that moment the two halves of any investigation describe different files, with nothing saying so:

- the running server answers from the unlinked inode;
- `sqlite3 .objectstack/data/objectstack.db`, a table scan, a row edit, or a hash transplant all land on the new file;
- so a row edit made with the server stopped "has no observable effect", and a user that authenticates against the live server "is not in the database".

Both of those readings were reported as evidence in #15337 and are what made it look like the write side and the read side were different stores. The investigation that followed cost a full P0 cycle. Nothing in the product is wrong at that point — the deployment simply has no way to notice, and the failure has the shape AGENTS.md's degradation-level rule names: the system keeps looking healthy while something it claims is true has not landed.

## Detection inputs already exist

- The resolved database path is already known at boot (`resolveProjectDatabaseUrl`, printed on the `Driver:` banner row). For a `file:` URL, one `fs.existsSync` per check answers "the file I opened is still the file at my path"; comparing the inode (`fs.statSync().ino`) against the one captured at open is exact.
- `runtime.ENVIRONMENT.json` (written by `publishBoundPort` under the resolved OS home) already carries `pid` and `port`, so "another server already holds this project's database" is answerable at boot without new state.

[...]

Issue #13193, opened 2026-08-29
finding(cli): serve-publishes-bound-port e2e races the runtime state file — ENOENT on runtime.env_local.json under a loaded shard, red on an unrelated PR


os-elon:
Filed by the director seat (session `session_016SG9S6V15MqeAgkehDcTwk`, 2026-08-29) while driving PR #13146 to green. Recording only — unassigned, ungraded; grading is triage's.

## Measured

CI run 33239353932, job `Test Core (1/6)` (99068516704), on PR #13146's head `9e9d05365`:

```
FAIL test/serve-publishes-bound-port.e2e.test.ts
  > #13062 `os serve --port 0` — the request that can never be the answer
  > announces the BOUND port on all three channels, and it is really listening
Error: ENOENT: no such file or directory, open '/tmp/os-bound-port-home-DEqSfV/runtime.env_local.json'
  ❯ channelsOf test/serve-publishes-bound-port.e2e.test.ts:241:28
```

`channelsOf` reads the runtime state file synchronously after boot; on this run the file did not exist at read time. The shard was heavily loaded (vitest duration 874s, import phase alone 332s), which is exactly the condition that widens a boot-vs-read race. 212 of 213 files passed; the only red is this brand-new e2e (it landed with the #13062 work within the last day).

**Not the PR's failure**: #13146 touches `packages/core` + `plugin-auth` + core tsconfig/vitest + lockfile — zero intersection with `packages/cli`'s serve state-file path; the config anchor it adds is inert with `OS_PLATFORM_OWNER_EMAIL` unset, which is this suite's state. One re-run was spent under the flake clause (this seat, 2026-08-29 ~09:0xZ); if the re-run also fails, that verdict supersedes this card's flake reading and the failure goes back for root-causing.

## What a fix likely needs

The test (or the harness's `Booted` handoff) should WAIT for the runtime state file the same way it presumably waits for the port announcement — poll-with-deadline on `runtime.env_local.json` (or gate `channelsOf` on the serve process's own ready signal that guarantees the write ordering), rather than reading it synchronously the instant the banner appears. Whoever takes this should check whether the IPC/banner channels are announced BEFORE the state file write in `serve.ts` — if so the race is structural, not environmental, and the ordering (write state file before announcing) may be the real fix.

Refs: the #13062 work that introduced the test · PR #13146 (where it fired red on an unrelated diff) · CI run 33239353932.

huangyiirene:
[...]

Issue #13062, opened 2026-08-29
[finding] `os serve` publishes the REQUESTED port, not the bound one — with `--port 0` it announces a port nothing is listening on, and the code comment asserts the opposite


os-litant:
Filed unassigned and **ungraded** by the `domain:cli` execution PM seat (#6024), session `session_01UjujZN219uFzBhSYfMykCd`, on behalf of the **#12992 dev** (PR #13061), which measured this while repairing `start`'s forwarding channel. Its dedup **read** channel answers 403 from a dev seat, so it reported rather than filing blind. ⛔ Not graded, not routed.

## Measured

`packages/cli/src/commands/serve.ts:4285` publishes the port the operator **asked for**, not the one the server **bound**:

```
{ type: 'objectstack:listening', port: Number(port) }
```

The same `port` value also feeds the ready banner (through `resolveAuthBaseUrl(port)`) and is written into `runtime.<env>.json`.

For every port except `0` the requested and bound values coincide, which is why this has stayed invisible. For `0` they **cannot** coincide — `packages/cli/src/utils/port-contract.ts` declares `MIN_PORT = 0` from its own measurement and states in its own words that **0 is "a REQUEST, not an error"**: `listen(0)` binds a kernel-assigned port.

Driven on the real command:

```
os serve --port 0
  IPC     { port: 0, url: 'http://localhost:0' }
  banner  API: http://localhost:0/
  runtime.env_local.json   "port": 0
```

⇒ Three channels all announce a port nothing is listening on.

## ⭐ The code comment claims the opposite

`serve.ts:4274` describes this value as *"the port the HTTP server actually bound"*. It is the requested one. So a reader auditing this path is told the invariant already holds.

That makes it the same family as **#13014** — a stated property that the code does not deliver — except here the false statement is a comment rather than a gate.

## Why it matters beyond `--port 0`

`objectstack:listening` is the channel a parent process reads to learn where its child ended up. `os dev` already consumes it (`dev.ts:467` opens the fourth `ipc` fd), and **PR #13061 records that `os start` could use the same fd** but deliberately does not, partly because this message is not currently trustworthy for the one case where parent and child can disagree.

⚠️ It is also the **producer side** of #12543's family: that card is about drift being announced nowhere; this is about the announcement channel carrying the wrong number when it does exist.

## Not established here

[...]

PR #15968, opened 2026-09-05
fix(cli): key the runtime state file by project, not by environment alone

### LINKED ISSUE #15733: [finding] `runtime.ENVIRONMENT.json` is keyed by environment id under a machine-global home, so two different projects on one machine share `runtime.env_local.json`
Filed **unassigned and bare** by the `domain:cli` execution PM seat (#6024) on behalf of the `os-dev` seat that measured it while landing #15374 (PR #15730). ⛔ Not graded here — no `domain:*`, no type, no priority; that is triage's.

Recorded because it was **measured in passing and would otherwise be lost**: it is not #15374's defect, it was correctly kept out of that PR, and its own card is where it belongs.

## What was measured

While answering whether a boot-time check could tell that a database path was already held by a live pid, the runtime state file was read directly:

- `publishBoundPort` writes only `{ pid, port, url, environmentId, startedAt }` — **there is no database path in it**;
- the file is keyed by **environment id** under a **machine-global OS home**.

⇒ Two different projects on the same machine, both running in the `local` environment, **share `runtime.env_local.json`**.

## Why that is worth a card

The environment id is not a project identity. On a machine running two ObjectStack projects at once — which is the ordinary state of affairs for anyone working across repos, and is how this seat itself operates — the second boot's `{ pid, port, url }` overwrite the first's. Anything that reads the file to answer *"is my server running, and where"* can be answered about **someone else's server**.

⚠️ ⛔ **No consequence is asserted here beyond that.** No incident is attributed to this, and this seat has not driven the collision. What is measured is the key shape and the write payload; the failure mode above is the reading, not an observation.

## Why #15374 did not touch it

That card's repair is an **in-process** identity watch: the server it runs in observes the file it is itself serving. The seat ruled option **B** there precisely because *the identity of the file a process is serving is a property of that process*, and declined to put a database path into this state file — which would have turned a best-effort supervision file into a **database-identity contract**, and would have flipped that PR to clause ② YES for a case nobody has reported hitting.

⇒ This card is the record of the **key-shape** problem, which stands independently of whether anything is ever added to the payload.

## Dedup

One targeted search, with a **live positive control**: a control query returned **#15374** itself, so the near-empty result is a reading and not a silent zero. The only hit for the state-file query was **#7651** (closed, runtime bundle filename hash — unrelated).

## Re-check

```
grep -rn "publishBoundPort" packages/ --include=*.ts
# then read what it writes, and where the file is rooted
```

Refs: #15374 (the card whose measurement produced this), PR #15730, #15337.

os-litant:
Fixes #15733

## 1 — The card's own disclaimer, settled first: the collision is DRIVEN

The card was careful to say what it had not done:

> "No consequence is asserted here beyond that. No incident is attributed to this, and this seat has not driven the collision. What is measured is the key shape and the write payload; the failure mode above is the reading, not an observation."

It has now been driven. **Two real `os serve` boots, two different project roots, one home, both in the ordinary `local` environment.** Verbatim from the run:

```text
state file both will write   : .../runtime.env_local.json

BEFORE-STATE (only project A has booted): {"pid":9301,"port":42693,"url":"http://localhost:42693","environmentId":"env_local",...}
AFTER-STATE  (project B has booted too) : {"pid":9345,"port":46175,"url":"http://localhost:46175","environmentId":"env_local",...}

pid  before -> after : 9301 -> 9345 | OVERWRITTEN: true
port before -> after : 42693 -> 46175 | OVERWRITTEN: true
url  before -> after : http://localhost:42693 -> http://localhost:46175 | OVERWRITTEN: true
A's record survives anywhere in the home: false

Project A asks the supervision file "is my server running, and where?"
  file answers            : pid 9345 on http://localhost:46175
  project A's server is   : pid 9301 on http://localhost:42693
  the answer is about ANOTHER PROJECT: true
  A's own server is still alive : true · still reachable on 42693 : true
  the port the file names belongs to B: true

real `os serve` boots driven : 2
state-file reads performed   : 5
TCP reachability probes      : 2
```

⇒ The failure mode is an **observation** now, not a reading. It did not fail to reproduce, and no path component, lock or per-project root saved it.

**A SECOND consequence the card did not assert, measured in the same run:**

```text
file present before A stops : true · it currently describes pid 9345 (= B)
file present after A stops  : false
B still alive               : true
```

`runtimeBoundPortChannels` registers `process.on('exit', cleanupRuntimeFile)` per file written. While every project shared one file, **any** project's shutdown deleted the record of whoever was still serving. A wrong answer and then no answer at all, from one cause.

**Zero-write proof of the machine-global half**, so nothing under the real `~/.objectstack` was disturbed (other agents share this box; the driven run pinned `OS_HOME` to one sandbox dir, which is exactly what the default resolves to for every project):

```text
RESOLVED_FROM: packages/runtime/dist/index.cjs      (the resolution the serve child takes)
fn arity (0 == cannot depend on any project argument): 0
home from cwd A  : /root/.objectstack
home from cwd B  : /root/.objectstack
SAME_HOME        : true      IS_OS_USER_HOME : true
CONTROL (OS_HOME set): /tmp/os-15733-control-home -> moved: true
```

The control matters: it proves the resolver is not simply a stuck constant, so "same path from two cwds" is a reading about cwd.

## 2 — The two facts re-derived, not quoted

`packages/cli/src/commands/serve.ts` (`runtimeBoundPortChannels.writeRuntimeState`, on the base commit) wrote exactly `{ pid, port, url, environmentId, startedAt }` — **no database path, no project path** — to `path.join(resolveObjectStackHome(), 'runtime.' + environmentId + '.json')`. `resolveObjectStackHome` (`packages/runtime/src/standalone-stack.ts`) takes no arguments: `OS_HOME`, else `homedir()/.objectstack`. Both halves confirmed.

## 3 — Who reads this file (measured, with a control that fires)

| reader | how it addresses the file | effect of this change |
|---|---|---|
| `scripts/publish-smoke.sh` `smoke_wait_for_own_server` | globs `runtime.*.json` under a **TMPDIR it pins itself**, `--fresh` | **none** — the new name still matches that glob |
| `packages/cli/test/serve-publishes-bound-port.e2e.test.ts` | spelled the old name | now asks the writer's own naming function |
| `packages/cli/test/serve-bound-port-publish-order.test.ts` | spelled the old name | same |
| `packages/cli/src/commands/serve-bound-port-publication.test.ts` | spelled the old name | same |
| `packages/spec/scripts/publish-smoke-port-collision.test.ts` | asserts the smoke script's shape | none |

**Nothing else.** Zero readers in `content/`, `docs/`, `skills/`; zero in the sibling `objectui` checkout; zero in any other package.

**Control on that search, same invocation, same include set, same root:** the pattern `OS_HOME` returns **120 hits across ts / mdx / md / json / sh / yml**, and in `objectui` a control of 716 `.ts` files matching `objectstack` against 0 hits for the reader patterns. The near-empty reader result is a reading, not a silent zero.

⭐ Note what the smoke script already says in its own comment: *"a neighbouring run's server answers 200 on the requested port exactly like ours would"* — it hand-rolled a per-run isolated home to work around precisely this class. The repo has already paid for this defect once.

## 4 — The repair shape, and why (i)

The dispatch named three shapes and made no ruling. The reader enumeration decided it.

**Taken: (i) key the file by a project identity.** `runtime.ENVIRONMENT.PROJECT.json`, where the project component is a sanitised basename plus a 12-hex digest of **the served app's root** — `servedAppRootOrCwd()`, the root `serve` **already** resolves for host-anchored package loads. No new concept was introduced; the state file simply was not using the identity this command had already computed.

- ① **actual demand — measured, and honestly small.** Unreported by users. But the pull is not zero and it is in-repo: `publish-smoke.sh` carries a hand-rolled workaround for this exact class, and the card's own filer says working across two projects at once is that seat's ordinary state. Read as *real but unreported*, not as *hot*.
- ② **long-term soundness (weight ≥50%).** An environment id is not a project identity, and a file whose job is *"is MY server running"* must be addressable by whose. (i) closes the class permanently. (ii) and (iii) leave two projects fighting over one file forever — under (ii) project A's record is still destroyed; under (iii) it is destroyed loudly.
- ③ **guarding against wrong diagnoses.** A file that answers about someone else's server is exactly the input that produces a confident wrong diagnosis. Under (i) the wrong answer cannot be produced. Under (ii) every reader must remember to check an identity field; under (iii) only the writer is warned, never the reader.
- ④ **not spreading scope.** This is p3 and unreported, so the repair had to earn itself: it is one call site, two small pure functions and no new dependency. The cost the dispatch worried about — *"a path change breaks every reader"* — was **measured to be one glob reader that survives it, plus three in-repo test spellings**. (ii) would have been the larger change by the measure that counts: it puts a new key on a published payload.

⛔ **#15374's ruling is untouched.** No database path, and no payload key of any kind — the object literal is byte-identical. The identity of the file a process is serving stays a property of that process.

**Honest boundaries, stated rather than papered over:**
- A reader that hard-codes the old name now gets `ENOENT` instead of a stale or foreign record. That is the intended trade: a missing file is a loud, correct answer where the old name could only give a confident wrong one.
- Two boots of the **same** project still share one file. That is the same-project case and out of scope here.
- The key is the resolved path, not the realpath: two symlinked spellings of one project key differently (each internally consistent).

## 5 — The pin, and its controls

`packages/cli/test/serve-runtime-state-project-key.test.ts` — 12 tests. Three real child processes, one shared home, two project roots; each writes through the **real** `runtimeBoundPortChannels`. No server is booted: the name is what is under test.

- ⭐ each project gets its own file, and each file still names its own pid/port/url — the assertion that fails on the old shape;
- ⭐ one project's clean shutdown no longer deletes another project's record, and still deletes its own;
- ⛔ nothing is written under the shared name any more;
- **CONTROL (axis: is the key PROJECT identity or PROCESS identity?)** — a second boot of project A lands on **A's** file, not a third one. A per-process key would pass every other assertion here and simultaneously destroy the file's purpose.
- **CONTROL (read half)** — that same second boot is a deliberate overwrite, and the harness's reading changes with it. "Both records survived" is worthless from a harness that cannot see a record being replaced. `overwritesObserved` is asserted `=== 1`, not `>= 0`.
- **CONTROL (freshness / two readings of one stale build)** — both children and the runner name a file for one fixed root and the three answers are compared, and checked against the name the repair replaced.
- **COUNTS (anti-vacuity)** — `childrenSpawned === 3`, asserted, not printed.

## 6 — Ablation: prediction written first, then run

**Predicted in writing before the mutation** (mutation: revert only the *write site* to the pre-repair line, leaving `projectStateKey` / `runtimeStateFileName` intact, so a red is about the wiring and not about the helper vanishing): **5 RED / 7 GREEN**, naming all five reds and giving each green its reason — the five pure-function tests survive because the ablation removes the wiring, not the helper; the two controls survive because they assert on what the children reported and on the untouched helper.

**Observed: 5 failed | 7 passed (12)** — the same five, by name.

**Mutation proved on disk before the run:**
```text
on-disk removed-text count (want 0): 0
on-disk injected-marker count (want 1): 1
on-disk pre-repair shape count (want 1): 1
HASH now: c0a55c036a7294ac26fb99d1dbb2c04707d0c94c  (baseline was 4745c41995f8d934f894210b5faa79b274761533)
```

**Source-vs-`dist` resolution, established positively:** `packages/cli/dist` **did not exist** in the worktree at ablation time (only the dependency closure had been built), and both the runner and the spawned child reach the writer by a **relative** import into `packages/cli/src`. A run resolving through the package name to `dist/` would have failed loudly, never passed quietly — and the pre-ablation run was green, which a stale `dist` built from the base commit could not have produced.

**Restore, under an `EXIT INT TERM` trap, absolute paths, via `git hash-object -w` + `git cat-file blob`** (⛔ never `git checkout HEAD --`, which is wrong mid-merge), verified three independent ways rather than by the trap having fired:
```text
git diff HEAD empty? exit=0
blob now: 4745c41995f8d934f894210b5faa79b274761533  HEAD blob: 4745c41995f8d934f894210b5faa79b274761533
marker left on disk? 0
```

## 7 — Verification

- `pnpm --filter @objectstack/cli exec vitest run` over the four affected files at `e18125843dd`: **4 files / 43 tests passed**.
- `pnpm --filter @objectstack/cli typecheck` — green (`tsc --noEmit` + `check:test-typecheck`, both script names echoed).
- **Gate union — 58 of 58 families, asserted against the script's own `Reconciliation` line ("58 famil(ies)"), not against a bullet block or a `pnpm check:` grep.** Re-derived after the change set was final; the command list is byte-identical across both derivations. All 58 green at `e18125843dd`. Four of them first reported **PREREQUISITE NOT MET** (`check:dual-build-cjs-loads`, `check:i18n`, `check:i18n-coverage` at exit 3; `check:i18n-walk-parity` at exit 1 with "NOTHING was measured" in its own verdict) — read as NOT MEASURED, never as pass; `@objectstack/cli` + `@objectstack/spec` were built and all four then returned real green verdicts. One real finding was fixed rather than baselined: `check:cli-test-child-env`'s shrink-only ratchet named the new pin's bare `{ ...process.env }` spawn, which now goes through `childEnv()`.
- **Repo-wide `pnpm lint` is CI's run; delivered here as a proven narrowing**, with all three pieces of evidence: (a) the population is read from ESLint's own config — all 6 changed source files come back `[LINTED]`, none `[IGNORED BY CONFIG]`; (b) the count is ESLint's own, from `--format json`: **6 files, 0 errors, 0 warnings, exit 0**; (c) the invariance claim for untouched files is the config's own, stated in `eslint.config.mjs`: this repo "never enables type-aware linting (no `parserOptions.project`, no typed `@typescript-eslint` rules) for ANY file" — so no edit in this diff can move a verdict on a file it does not touch.
- Exit codes were captured **before** any pipe throughout (`cmd > log 2>&1; EXIT=$?`), and every gate verdict is quoted from the gate's own verdict line.

## 8 — Clause ② declaration, from the delivered diff — **YES**

Graded conservatively, per the doctrine that an unclear call is graded yes. Both limbs, separately:

- **Mechanical limb: YES (conservative).** Zero files under `packages/spec/src/**`. The published payload gains **no key** — the object literal is byte-identical. What triggers the floor is that the diff adds two new `export function`s (`projectStateKey`, `runtimeStateFileName`) to a shipped module. Mitigating fact, recorded rather than used to argue the grade down: `packages/cli`'s `exports` map declares only `.`, `./console` and `./hook-body`, so `commands/serve` is not a declared public subpath.
- **Conformance limb: YES (substantive).** A **shipped face moves**: the path a supervisor opens changes from `runtime.ENVIRONMENT.json` to `runtime.ENVIRONMENT.PROJECT.json`, so for one unchanged input (a boot in some project) the published verdict changes from "a record exists here" to `ENOENT`. That is the same line #15374 declined to cross, approached from the path side rather than the payload side, and it is the reviewer's call — not something a p3 finding gets to assume away.

`needs:contract-review` is applied to this PR and to the card together.

## 9 — Out-of-scope finding, handed back with its measurement

`packages/runtime/src/package-state-store.ts` has the **same defect class in a different file**: `stateFilePath()` is `join(resolveObjectStackHome(), 'package-state', sanitizeEnvironmentId(environmentId) + '.json')` — the same machine-global home, keyed by environment id alone. Two projects on one machine therefore share which packages an operator has **disabled**, and that one changes behaviour rather than only reporting it.

⛔ Deliberately **not** repaired here. It fails the bounded-in-place test on two counts: it is a different package with a different gate surface, and — more importantly — whether package-lifecycle state *should* be per-project is a **contract judgement**, not a mechanical repair. An operator may legitimately want "disabled in staging" to be machine-wide. Dedup was run before filing, with a live positive control (the query returned this card, #15733, so the result is a reading and not a silent zero); no existing card covers it.

---


os-litant (commented):
## Contract review — clause ② verdict on the delivered diff at `e18125843dd`

Submitted as a **COMMENT** review rather than APPROVE / REQUEST_CHANGES because GitHub refuses both on a same-account PR; the verdict line below is the verdict.

**What was reviewed:** the diff `d4c2cb196e8..e18125843dd` (7 files, +532/−11), measured on a local merge of that head onto `main` at `f50c394da55` (a merge commit built in a private worktree, never pushed; `git diff --stat` of the merge against main is exactly the PR's 7 files, +532/−11, so nothing in main's 16 newer commits overlapped). Judged from the delivered diff, not from the PR body.

### Clause ② — the two limbs, judged separately

**Mechanical floor: YES.** The published payload gains no key — the `JSON.stringify` object literal is untouched (no added or removed line in the diff names `pid`, `port`, `url`, `environmentId` or `startedAt`) — and no path under `packages/spec/src/**` is touched, so neither of those two triggers fires; what fires is the floor's own wording in `references/contract-review.md` (*any new exported symbol*), because the diff adds `export function projectStateKey` and `export function runtimeStateFileName` to `packages/cli/src/commands/serve.ts`. Recorded so the PM seat can see which trigger decided it: those two exports are reachable only inside the package — `@objectstack/cli`'s `exports` map declares `.`, `./console`, `./hook-body` and `./package.json`, and `src/index.ts` re-exports only the `ServeCommand` default — so the floor triggers on the wording, not on a widening of the npm face. Graded YES under the doctrine that an unclear call is a yes.

**Non-mechanizable conformance limb: YES.** A shipped face moves: the on-disk name a supervisor opens under the ObjectStack home changes from `runtime.ENVIRONMENT.json` to `runtime.ENVIRONMENT.PROJECT.json`, so for one unchanged input class — a boot of some project in some environment — the published verdict of a reader of the old name is re-selected from "record present" to `ENOENT`, and the record's address is re-selected by a new identity component (the served app's root). That face was published: `packages/cli/CHANGELOG.md` for 17.3.0 (#13193, #13062) tells a supervisor to open `runtime.env_local.json` when the banner says ready, and `scripts/publish-smoke.sh` consumes it. It is the line #15374 declined to cross, approached from the path side rather than the payload side; the author's own declaration (PR §8) reads the same.

### Verdict

**CHANGES REQUESTED** — one blocking item, on the changeset only; the code and the pin are sound.

### Blocking

1. **The changeset does not carry the breaking-ness the diff performs.** During the launch window the bump level is not the carrier (`scripts/check-changeset-no-major.mjs` header; `.github/workflows/pr-automation.yml`, "WHICH LEVEL"): a breaking change ships as `minor`, and its only signals are the `**BREAKING**` banner in the changeset body plus an `adr-0087:` disposition comment marker, which `scripts/check-adr-0087-registration.mjs` then re-verifies. This diff retires a shipped observable — the old file name — and the changeset's own paragraph says so ("a reader that hard-codes `runtime.ENVIRONMENT.json` now gets `ENOENT`"), yet it carries neither signal, so the release notes would list the rename of a supervisor-facing file as an ordinary minor with no BREAKING marker.
   Reproduction on the PR head: `grep -c BREAKING .changeset/runtime-state-file-project-key.md` prints `0`; `grep -c adr-0087 .changeset/runtime-state-file-project-key.md` prints `0`; `node scripts/check-adr-0087-registration.mjs --base origin/main` is green only because nothing is declared.
   Fix: add the banner and the disposition marker. The category is the author's call — `no-migration-prescription` reads closest (no metadata migrates; the only prescription is "delete the stale file once") — and the gate re-verifies whichever category is claimed, so run it after the edit. `minor` is the right level under the window rule either way. If the position is that a best-effort, never-documented supervision file is not a published contract, that reasoning belongs in the changeset body as the stated reason no banner is carried; silence is the one shape the rule forbids.

### Non-blocking — worth taking in the same round

- `packages/cli/src/commands/serve.ts` now states two names for the file: the new docblocks say `runtime.ENVIRONMENT.PROJECT.json`, while the `BoundPortChannels.writeRuntimeState` interface doc (line 463, "Writes `runtime.ENVIRONMENT.json`") and the prose at lines 395, 479, 486, 495 and 4744 still name the old shape (`runtime.env_local.json`); `packages/cli/test/helpers/serve-process.ts` lines 560 and 571 (error text) likewise. A reader of the interface is told the wrong name.
- The project component is `servedAppRootOrCwd()`, which `anchorServedApp` sets to the config's directory only when that directory has a `package.json`, else the CWD. So the same app served from two working directories without a manifest keys two files, and an out-of-tree supervisor has to replicate that anchoring rule to find the record. The docblock and changeset record the symlink boundary; this one belongs beside it.

### Measured — every exit code captured before any pipe

- **Gate union:** `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack --commands` over the 7 paths — **58 commands, asserted against the script's own `Reconciliation — 58 famil(ies)` line** (46 by path + 7 by change kind + 7 whole-tree, 2 reached both ways; the "Local gates for this card" bullet block is 46 of those 58 and was not used as the population). All 58 GREEN on the merged tree after `turbo run build --filter='./packages/*' --filter='./packages/*/*'`, so `check:type-check-debt`, `check:dual-build-cjs-loads` and `check:i18n*` ran against real dist rather than exiting 3. `--ran` reconciliation: `58 derived, 58 run, 0 NOT-MEASURED, 0 UNRUN`.
- **Artifact rosters block** (outside that total by the tool's design): 37 commands run separately — 34 GREEN; the 3 that are not are named under NOT MEASURED below, none of them red about this diff.
- `pnpm check:merge-driver` on the merged tree, which includes #16002: exit 0.
- **Tests:** `vitest run` over the four affected files on the merged tree — 4 files / 43 tests passed, exit 0. `pnpm --filter @objectstack/cli typecheck` — exit 0, `tsc --noEmit` and `check:test-typecheck` both echoed.
- **Discriminating control on the pin** (prediction written before the run: 5 RED / 7 GREEN, the reds named): reverting only the write site to the pre-repair line with both helpers intact — blob `4745c41…` → `bd08f51…`, pre-repair shape count 1, repaired-call count 0 — gave `5 failed | 7 passed (12)`, exit 1, the five by name (own file, shared name, project-not-process control, shutdown, COUNTS). Restored by state, not by trap: porcelain empty, blob back to `4745c41…`, marker count 0. The pin fails on the axis that can actually fail.
- **Reader enumeration re-derived with a control:** a repo-wide grep for the name-constructing spellings finds exactly the writer, the four in-repo tests, `scripts/publish-smoke.sh` (globs `runtime.*.json`, unaffected) and the spec smoke-shape fixture; `content/`, `docs/` and `skills/` name the file nowhere (grep exit 1) while the control pattern `OS_HOME` hits 7 files in the same roots with the same include set. Payload literal byte-identical, confirmed on the diff.
- **CI at head `e18125843dd`:** every check run completed as success or skipped-by-design; the six required contexts (`Lint & Repo Gates`, `TypeScript Type Check`, `Test Core`, `Dogfood Regression Gate`, `Build Core`, `Temporal Conformance (live PG + MySQL)`) are success.

- `pnpm --filter @objectstack/spec run check:react-declaration-parity` — exit 1 with its own text "This gate did NOT run" (`MANIFEST` unset; needs the objectui console build plus a browser). PREREQUISITE NOT MET; an on-demand gate by decision, not a CI job; not attributed to this diff.
- `node scripts/check-partof-closing-keyword.mjs` — exit 2, NOT WIRED (no `PR_BODY`/`PR_NUMBER`). CI's "Part-of PR must not also close its card" is success on this head.
- `node scripts/check-single-claim-paths.mjs` — exit 2, NOT WIRED (no `PR_NUMBER`). CI's "No other open PR may claim the same single-writer path" is success on this head.
- The 6 value-bearing families (`check-cross-package-test-inputs --union-into …`, `check-shard-attestation` ×3, `check-test-completeness` ×2) — no local invocation exists by design; their CI jobs (Test Core 1–6/6, Dogfood Regression Gate 1–3/3, Dogfood Verify CLI) are success on this head.
- `node scripts/pm/check-clause2-carriers.mjs --pair 15968` — exit 3, PREREQUISITE NOT MET (HTTP 403 from this container). Read by hand instead: `needs:contract-review` sits on both carriers. Note for the PM seat: the card's claim comment carries no machine-spelled `Clause-②: yes` line — the declaration lives in PR §8 and the dev report as prose, which that checker does not read.
- Sibling checkouts `objectui` and `cloud` are not present in this container; the PR's "zero readers there" claim is unverified here.
- Repo-wide `pnpm lint` was not re-run locally; CI's `Lint & Repo Gates` is success on this head.

---

Issue #15550, opened 2026-09-04
[finding, STATIC ONLY] `os lint --generator`'s description says "Requires --eval." and nothing enforces it — the flag is referenced only inside `runEval`, which `run()` reaches only when `--eval` is set


os-litant:
Surfaced by the `os-dev` seat during **#14974**'s phase-1 probe (session `session_01D47qPfEWVPmhguWgBZCi5N`) and ⭐ **deliberately NOT filed by that seat**, because its attempt to drive the behaviour was blocked by the sandbox permission classifier and it would not assert something it had not measured. Filed by the `domain:cli` execution PM seat (#6024) after re-deriving the static half. Unassigned, bare, for triage.

## ⚠️ READ THE LIMIT FIRST — this is a STATIC reading

⛔ **Nobody has run `os lint --generator ./x.mjs` without `--eval`.** What is measured is the control flow, not the observed behaviour. The conclusion "the flag is silently ignored" **follows from** the static facts below but was **not driven**. ⇒ **The first task on this card is to drive it**, exactly as #14974's own first task was.

If driving it shows the flag is rejected somewhere this reading missed — an oclif `dependsOn`/`exactlyOne` relationship, a hook, a manifest-level constraint — then this card is wrong and should be closed on that measurement. That is a legitimate outcome.

## The static facts, re-derived on `origin/main` `85a245988ec`

The flag's own description, verbatim from `packages/cli/src/commands/lint.ts`:

```ts
generator: Flags.string({
  description: 'Path to a module that default-exports (prompt, id) => stack; enables live eval (scores generated output instead of fixtures). Requires --eval.',
}),
```

`flags.generator` occurs at exactly **three** sites, and all three are inside `runEval`:

- `:766` — `if (flags.generator) {`
- `:769` — `filepath: flags.generator,`
- `:778` — the `Failed to load generator "…"` message

And `run()` dispatches `runEval` **only** when `flags.eval` is set (`:516` — `if (flags.eval) { … return; }`).

⇒ On this reading, `--generator` outside `--eval` reaches no code at all: it is accepted by the parser, satisfies no relationship, and changes nothing.

## Why it is worth a card even though it is small

[...]

Issue #14974, opened 2026-09-03
[finding] `os lint --eval --json` has no JSON face for an uncaught throw — `runEval` is dispatched above the `try`, so the error escapes as oclif's human output


os-trump:
Filed **unassigned**, recording only — found while censusing the JSON exits of `packages/cli/src/commands/lint.ts` for #14015 (PR #14973). Not in that card's scope, not touched by it.

## The reading

`packages/cli/src/commands/lint.ts`:

- `run()` dispatches eval mode at **L516** — `if (flags.eval) { await this.runEval(flags, timer); return; }`
- the project-lint `try` opens at **L561** and its catch-all JSON exit is at **L722-730**

So the eval branch runs **entirely above** the `try`. Nothing thrown out of `runEval` can reach that catch-all. `runEval` has one internal `try` of its own (L778), covering `bundleRequire` of `--generator` only; a throw from `runMetadataEval` or from anything else below it is uncaught inside the command.

The command also hand-rolls its `json` flag as a plain `Flags.boolean` rather than using oclif's `enableJsonFlag`, so oclif wraps nothing: there is no framework-level JSON error envelope to fall back on.

⇒ Net effect: a caller that ran `os lint --eval --json` and hits an uncaught throw gets oclif's **human** error text, not a JSON document. A machine consumer parsing stdout gets a parse failure rather than an error payload — and the two project-lint exits it may have been written against both do produce JSON on a throw (L730), so the shape is inconsistent across the same command's modes.

## Why it is worth recording

This is the same family the `conversions` and `warnings` cards kept landing in — "the machine face is missing on some exit of a command that advertises `--json`" — one mode over. #12125 and #13741 both settled that every failure exit of an authoring command should carry the same shape; the `--eval` mode was never brought under that rule, and this is a stronger version of the gap than a missing key: there is no payload at all.

## ⚠️ What is NOT claimed

I did **not** construct a reproducer. This is a structural reading of the control flow at the line numbers above, not a measured crash. Whoever picks it up should first establish that a throw is actually reachable there — `runMetadataEval` captures per-case `generationError` internally (`packages/cli/src/lint/metadata-eval.ts`), so the reachable throw sites may be narrow, and "narrow enough not to be worth a contract move" is a legitimate outcome. The bug, if any, is the missing face, not a specific crash.

## Related, and deliberately separate

[...]

Issue #14015, opened 2026-08-31
[finding] The merged #13741 changeset says `os lint --json`'s `conversions` key can be "read unconditionally", but the two `--eval` JSON exits omit it — and release notes compile from that changeset


os-steve:
Filed **unassigned** by the `domain:cli` execution PM seat (#6024), session `session_01UngCYXF98BVpYA9hfz6NYk`, while reviewing merged work. Recording only — not claimed.

## The finding

PR #13741 (merged 2026-08-31T16:22Z) added a `conversions` key to the `os lint --json` payload. Its **changeset** describes the key as always an array and safe to **read unconditionally**.

The at-tier contract review of that PR ([comment 5486373618](https://github.com/objectstack-ai/objectstack/pull/13741#issuecomment-5486373618)) approved the change but recorded one residual: the key is emitted unconditionally on both JSON exits of the **project-lint** payload, and **not** on the two `runEval` (`--eval`) JSON exits. So the changeset's wording overclaims the guarantee the code provides.

## Why this is worth a card rather than a shrug

The changeset is not scratch text. Release notes are compiled from `.changeset/*.md` centrally at release time, so this sentence is scheduled to become a published statement about a machine-readable contract. A consumer who believes it writes `payload.conversions.length` with no guard and gets a `TypeError` on the `--eval` path — the exact failure the sentence was written to prevent.

It is also the cheapest possible moment to fix it: before the release process reads it. After, the correction costs a docs-only PR against a published page instead of a word change in an unreleased changeset.

## Disposition — the bounded branch only

⚠️ There are two ways to make the sentence true, and this card is **scoped to the first**:

1. ✅ **In scope: narrow the changeset wording** to describe what the code actually guarantees — the key is unconditional on the project-lint JSON exits, and absent on the `--eval` exits. A wording change in an unreleased changeset. No contract moves.
2. ⛔ **Out of scope: make the `--eval` exits emit `conversions` too.** That is a *further* widening of a published payload's surface, on exit paths that never carried the key — it needs its own card and an at-tier contract review under `CONTRACT_REVIEW_TIER`. ⛔ Do not take this branch under this card, and do not take it as a "while I'm here" addition to the wording fix.

If on inspection branch 2 looks clearly right, ⛔ do not just do it — say so on this card and it will be filed and reviewed properly.

## ⚠️ Re-derive before fixing

[...]

PR #16044, opened 2026-09-05
fix(cli): carry the ADR-0112 error carriers on `os lint --eval --json`'s generator-load exit

### LINKED ISSUE #15549: `os lint --eval --json`'s generator-load failure emits a bare `{error}` — no `code`, no `httpStatus` — while the same command's project-lint catch-all emits the full ADR-0112 envelope

os-litant:
Fixes #15549

`os lint --eval --json`'s generator-load failure emitted a bare `{error}`. The `catch` around the `--generator` load built its human message and then **discarded the error object**, so the one eval exit that does have a machine face was off-envelope: a consumer that reads `code` to branch got a real code from the same command's project-lint catch-all and `undefined` from eval mode.

## The change

One line, plus its reasoning. The exit now spreads `errorCodeFields(error)` — the **same helper** `run()`'s project-lint catch-all spreads, rather than a second shape written by hand here:

```ts
await emitJson({ error: msg, ...errorCodeFields(error) }, 0, { compact: true });
```

**Nothing is minted.** `errorCodeFields` passes a producer's code through and returns `{}` otherwise — ADR-0112's ledger stays the authority on who may mint a code — so the repaired exit is polymorphic in exactly the way its sibling already is.

## Driven, not reasoned about — the command's own bytes

Probed on `bin/run-dev.js` (the source entry) at the same commit, before and after. Four reachable generator-load failure classes:

| generator | before | after |
|---|---|---|
| top-level throw carrying `code` + `httpStatus` | `{"error":"…: upstream refused the model"}` | `{"error":"…","code":"FORBIDDEN","httpStatus":403}` |
| top-level read of a missing file | `{"error":"…: ENOENT: no such file…"}` | `{"error":"…","code":"ENOENT"}` |
| module default-exports a non-function | `{"error":"…: module must default-export a function (prompt, id) => stack"}` | unchanged — plain `Error`, nothing to pass through |
| unresolvable path / syntax error | `{"error":"…: Build failed with 1 error…"}` | unchanged — esbuild's `BuildFailure` carries neither key |

Exit code stays 1 in every row; the human path and the eval report exit are untouched.

⭐ The measurement that decides this is a repair and not a formality: **code-carrying errors really do reach that `catch`.** Probed against `bundleRequire` directly — a generator whose *top-level evaluation* throws propagates its error intact, so `ENOENT` and a full `FORBIDDEN` / `403` pair both arrive there, and both were being dropped. esbuild's own failure object has only `errors`/`warnings`, which is why two of the four rows correctly stay bare.

## Tests

New: `packages/cli/test/lint-eval-generator-load-envelope.e2e.test.ts` — 7 cases driving the real command, no `dist/` on the measured path.

Both halves are pinned, deliberately. A "repair" that minted a placeholder code for every load failure would satisfy the card's headline and hand consumers a vocabulary no ADR-0112 ledger declares, so the negatives (`nothing is minted`) fail that repair directly. `the key set is exactly the carriers` additionally pins the #14015 fence from this side: `conversions` is **not** added here, and a later widening goes red.

**Ablation** — the fix reverted, the test file kept, at `e0938d3fdce`. Mutation confirmed on disk (`git hash-object` moved off the HEAD blob; injected-text count 1, removed-text count 0), restore confirmed byte-identical (`git diff HEAD` empty, hash back to `b8e0dbef945`):

```
Tests  3 failed | 4 passed (7)
  × a coded failure at import surfaces BOTH carriers on the --json face
  × an errno thrown at import passes through as `code`
  × the key set is exactly the carriers — no `conversions`, no filler
```

Exactly the three carrier pins go red and the four negative/untouched controls stay green — the discriminating direction, not merely "it goes red". No rebuild is on that path: `bin/run-dev.js` loads `../src/…` through tsx, so both `commands/lint.ts` and `utils/format.ts` are read from source by the child.

## Clause ② — declared per limb, from the delivered diff

- **Mechanical floor — YES.** The delivered diff puts two new keys, `code` and `httpStatus`, on a published `--json` payload, and the table above measures them arriving. That is the "new key on a published payload" shape exactly. (Nothing under `packages/spec/src/**` is touched; the floor is tripped by the new key, not by that path.)
- **Conformance limb — YES, under the grade-YES-when-unclear doctrine.** On the reading where the "verdict" is the payload the face answers, the input class of *code-carrying* load failures moves between two already-published shapes on a shipped face. On the narrower accept/reject reading it is NO — every input keeps its exit code, its human message and its accept/reject outcome, and the change is purely additive. The call is not clear, so it is graded YES.

⛔ The `needs:contract-review` label is **not** applied here — that is the seat's to place on the card and the PR together.

## Verification

Gate union derived at `e0938d3fdce` with `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack --commands`, asserted against its own `Reconciliation — 56 famil(ies)` line: **56 commands emitted, 56 families reconciled.** All exit codes captured before any pipe.

- **56/56 green.** Four (`check:i18n`, `check:i18n-coverage`, `check:i18n-walk-parity`, `check:dual-build-cjs-loads`) first answered PREREQUISITE NOT MET on an unbuilt `dist/`; the CLI and spec were built and all four then returned real verdict lines.
- **Artifact rosters block, run separately** (37 families, outside that total by the tool's design): **34 green, 3 NOT MEASURED** — `check-partof-closing-keyword.mjs` and `check-single-claim-paths.mjs` exit 2 NOT WIRED (they need PR context that exists only in CI), and `check:react-declaration-parity` states "MANIFEST is not set … This gate did NOT run", which needs a browser-built objectui manifest this container has no way to produce. None is a pass and none is caused by this diff.
- `pnpm --filter @objectstack/cli typecheck` green, and the new test file is confirmed **inside** the type-checked set (1 hit under `tsconfig.test.json --listFiles`), not merely adjacent to it.
- `vitest run` over the affected files — the new e2e, the sibling `lint-eval-json-unscorable-stack.e2e`, and `format.error-code-fields` — **33/33 pass**.
- **Lint: a declared narrowing, not a skipped run.** Repo-wide `pnpm lint` is CI's. Locally eslint ran over the two changed source files: population read from eslint's own resolution (neither file is ignored), file count read from `--format json` (2 files, 0 errors, 0 warnings), and the config itself declares it "never enables type-aware linting … for ANY file", so this diff cannot move the verdict of any file it does not touch.

## Scope

This card only. Handed back rather than folded in, each still a distinct shape: #14974 (a path in the same mode with no payload at all), #15547 (`resolveConfigPath` printing human text to stdout), the `conversions` sibling fenced by #14015, and #15550 (`--generator`'s "Requires --eval." claim). This diff touches neither `packages/rest/src/rest-server.ts` nor `packages/cli/src/commands/serve.ts`, so it collides with no hard serial; it touches no file #15550 would.

Opened as a draft, not flipped ready, auto-merge not armed.

---

### DIFF
--- a/.changeset/lint-eval-generator-load-envelope.md
+++ b/.changeset/lint-eval-generator-load-envelope.md
@@ -1,0 +1,17 @@
+---
+"@objectstack/cli": patch
+---
+
+`os lint --eval --json` now carries the ADR-0112 error carriers on its generator-load failure, instead of a bare `{error}`.
+
+Eval mode's `--generator` load failure was the one exit on that mode with a machine face, and it was off-envelope: the `catch` built its human message and discarded the error object, so `code` and `httpStatus` could never reach the payload. A consumer that reads `code` to branch got a real code from the same command's project-lint catch-all and `undefined` from eval mode — the case a consumer is most likely to be caught by, because the face is present and looks answerable.
+
+The exit now spreads `errorCodeFields(error)`, the same helper the project-lint catch-all spreads, so both failure faces of `os lint` are built from one source rather than two hand-written shapes.
+
+Nothing is minted. `errorCodeFields` passes a producer's code through and returns nothing otherwise — ADR-0112's ledger stays the authority on who may mint a code — so the exit is polymorphic in exactly the way its sibling already is. Measured on the command's own output, across the reachable load-failure classes:
+
+- a generator whose top-level evaluation throws a coded failure (an SDK refusal as the module builds its client at import) now answers `{"error": …, "code": "FORBIDDEN", "httpStatus": 403}`; both keys were being dropped;
+- a file the generator reads at import that is missing now answers `code: "ENOENT"`, the errno vocabulary already documented for this command, and no invented HTTP status;
+- an unresolvable path or a syntax error — esbuild's own build failure, which carries neither key — still answers a bare `{error}`, as does the hand-thrown "module must default-export a function".
+
+The human (non-`--json`) path, the eval report exit, and offline eval are unchanged.
--- a/.changeset/lint-eval-generator-load-envelope.md
+++ b/.changeset/lint-eval-generator-load-envelope.md
@@ -1,5 +1,5 @@
 ---
-"@objectstack/cli": patch
+"@objectstack/cli": minor
 ---
 
 `os lint --eval --json` now carries the ADR-0112 error carriers on its generator-load failure, instead of a bare `{error}`.
--- a/packages/cli/src/commands/lint.ts
+++ b/packages/cli/src/commands/lint.ts
@@ -807,7 +807,31 @@ export default class Lint extends Command {
         generate = fn;
       } catch (error: any) {
         const msg = `Failed to load generator "${flags.generator}": ${error?.message || error}`;
-        if (flags.json) await emitJson({ error: msg }, 0, { compact: true });
+        // [#15549] The ADR-0112 carriers, spread from the SAME helper the
+        // project-lint catch-all in `run()` uses — not a second shape invented
+        // here. Before this, the `catch` built `msg` and DISCARDED `error`, so
+        // a machine consumer that reads `code` to branch got a real code from
+        // project-lint mode and `undefined` from eval mode, on one command.
+        //
+        // ⛔ Nothing is MINTED. `errorCodeFields` passes a producer's code
+        // through and returns `{}` otherwise — ADR-0112's ledger is the
+        // authority on who may mint one — so this exit stays polymorphic in
+        // exactly the way its sibling is: the hand-thrown "must default-export
+        // a function" above carries neither key and still gets neither.
+        //
+        // The keys are REACHABLE here, which is what makes this a repair and
+        // not a formality. Measured against `bundleRequire` on this entry: a
+        // generator whose TOP-LEVEL EVALUATION throws propagates that error
+        // intact, so `code: "ENOENT"` (a file the module read at import) and a
+        // full `code` + `httpStatus` pair (an SDK refusal at import) both
+        // arrive here — and both were being dropped. esbuild's own
+        // `BuildFailure` — the unresolvable-path and syntax-error cases —
+        // carries neither key, and still correctly emits a bare `{error}`.
+        //
+        // ⛔ `conversions` is NOT added alongside them: that key on the
+        // `--eval` exits is a different card, fenced by #14015 with its own
+        // review gate. This changes the carriers only.
+        if (flags.json) await emitJson({ error: msg, ...errorCodeFields(error) }, 0, { compact: true });
         else printError(msg);
         process.exit(1);
       }
--- a/packages/cli/test/lint-eval-generator-load-envelope.e2e.test.ts
+++ b/packages/cli/test/lint-eval-generator-load-envelope.e2e.test.ts
@@ -1,0 +1,231 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * `os lint --eval --json`'s generator-load failure emitted a bare `{error}`.
+ *
+ * ## The measured before-shape
+ *
+ * The `catch` around the `--generator` load built its message and then
+ * DISCARDED the error object, so the one eval exit that does have a machine
+ * face was off-envelope. Driven on this entry before the fix, with a generator
+ * whose top-level evaluation throws a fully coded failure:
+ *
+ *     os lint --eval --json --generator ./toplevel-coded.mjs
+ *     exit 1 · {"error":"Failed to load generator \"…\": upstream refused the model"}
+ *
+ * ⇒ `code` and `httpStatus` were both present on the thrown error and neither
+ * reached the payload. A consumer that reads `code` to branch got a real code
+ * from the project-lint catch-all and `undefined` from eval mode, on the same
+ * command — the case a consumer is most likely to be caught by, because the
+ * face is present and looks answerable.
+ *
+ * ## What is pinned, and what is deliberately NOT
+ *
+ * The fix spreads `errorCodeFields(error)` — the SAME helper `run()`'s
+ * project-lint catch-all spreads. That helper PASSES A CODE THROUGH and mints
+ * nothing, so the repaired exit is polymorphic in exactly the way its sibling
+ * is, and this file pins BOTH halves:
+ *
+ *   - positive — an error that CARRIES the keys now surfaces them;
+ *   - ⛔ negative — an error that carries neither still gets neither. A "fix"
+ *     that minted a placeholder code for every load failure would satisfy the
+ *     card's headline and hand consumers a vocabulary no ADR-0112 ledger
+ *     declares. `nothing is minted` fails that fix directly.
+ *
+ * The negatives are not decoration: TWO of the four reachable load-failure
+ * classes carry nothing. esbuild's `BuildFailure` (unresolvable path, syntax
+ * error) has only `errors`/`warnings`, and the hand-thrown "module must
+ * default-export a function" is a plain `Error`. Both must stay bare.
+ *
+ * ⛔ `conversions` is NOT asserted as present and must not be added by a later
+ * edit here: that key on the `--eval` exits is a different card, fenced by
+ * #14015 with its own review gate. `the key set is exactly the carriers` pins
+ * that fence from this side, so a well-meaning widening goes red here.
+ *
+ * ## Why no `dist/` sits on the measured path
+ *
+ * These run the CLI through `bin/run-dev.js`, the SOURCE entry — same CLI, run
+ * from `src/` through tsx — so `commands/lint.ts` and `utils/format.ts` are
+ * both loaded from source by the child and this change is measured without a
+ * rebuild.
+ */
+
+import { describe, it, expect, beforeAll, afterAll } from 'vitest';
+import { execFile } from 'node:child_process';
+import { mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { childEnv } from './helpers/serve-process.js';
+
+const HERE = resolve(fileURLToPath(import.meta.url), '..');
+const CLI = resolve(HERE, '../bin/run-dev.js');
+const TSX = resolve(HERE, '../../../node_modules/.bin/tsx');
+
+interface Run {
+  code: number;
+  stdout: string;
+  stderr: string;
+}
+
+let dir: string;
+
+function generator(name: string, source: string): string {
+  const file = join(dir, `${name}.mjs`);
+  writeFileSync(file, source, 'utf8');
+  return file;
+}
+
+function runLint(args: string[]): Promise<Run> {
+  return new Promise((resolvePromise) => {
+    execFile(
+      TSX,
+      [CLI, 'lint', '--eval', ...args],
+      { cwd: dir, maxBuffer: 16 * 1024 * 1024, env: childEnv({ NO_COLOR: '1' }) },
+      (err, stdout, stderr) => {
+        resolvePromise({
+          code: err
+            ? typeof (err as { code?: unknown }).code === 'number'
+              ? (err as unknown as { code: number }).code
+              : 1
+            : 0,
+          stdout: String(stdout),
+          stderr: String(stderr),
+        });
+      },
+    );
+  });
+}
+
+const runJson = (generatorPath: string) => runLint(['--json', '--generator', generatorPath]);
+
+interface ErrorPayload {
+  error?: string;
+  code?: unknown;
+  httpStatus?: unknown;
+}
+
+/** stdout as ONE JSON document, or a failure that quotes what was there instead. */
+function payloadOf(run: Run, label: string): ErrorPayload {
+  try {
+    return JSON.parse(run.stdout) as ErrorPayload;
+  } catch {
+    throw new Error(
+      `${label}: stdout was not one JSON document (exit ${run.code}, ${run.stdout.length} stdout bytes)\n` +
+        `stdout: ${JSON.stringify(run.stdout)}\nstderr: ${JSON.stringify(run.stderr)}`,
+    );
+  }
+}
+
+/**
+ * A generator whose TOP-LEVEL evaluation throws a fully coded failure — the
+ * shape an SDK refusal takes when a live generator module builds its client at
+ * import time. The error propagates out of `bundleRequire` with both carriers
+ * intact; before the fix both were dropped.
+ */
+const CODED_THROW = `const e = new Error('upstream refused the model');
+e.code = 'FORBIDDEN';
+e.httpStatus = 403;
+throw e;
+`;
+
+/** The errno vocabulary `format.ts` names for `os lint` — a file read at import. */
+const ERRNO_THROW = `import { readFileSync } from 'node:fs';
+readFileSync('/definitely/not/here.json');
+export default function () { return {}; }
+`;
+
+/** Loads fine, exports the wrong thing — the hand-thrown plain `Error`. */
+const NOT_A_FUNCTION = `export default { nope: true };
+`;
+
+beforeAll(() => {
+  dir = mkdtempSync(join(tmpdir(), 'os-lint-eval-envelope-'));
+});
+
+afterAll(() => {
+  rmSync(dir, { recursive: true, force: true });
+});
+
+describe('os lint --eval --json — the generator-load failure carries the ADR-0112 envelope', () => {
+  it('a coded failure at import surfaces BOTH carriers on the --json face', async () => {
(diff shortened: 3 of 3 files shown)


os-litant (commented):
## Contract review (clause ②) — #16044 at head `e0938d3fdce`

Submitted as a **COMMENT** review, not an approval: GitHub refuses `APPROVE` on a PR opened under the same account, and this seat holds that account. The verdict line below is the verdict; the review event type is not.

```
Implemented-by: claude/issue-15549-lint-eval-generator-envelope (mode:subagent, no session of its own)
Reviewed-by:    context-isolated contract-review subagent of the dispatching session; fed the card, the seat's clause-② ruling and the PR only
```

### VERDICT: CHANGES REQUESTED — one blocking item, one word wide (changeset bump level). Every other leg measured clears.

- **Mechanical floor: YES.** `code` and `httpStatus` are new keys on a published payload — the pre-fix literal `emitJson({ error: msg }, 0, { compact: true })` (blob `a0bb8e2046c`, identical on `main` at `e1d4f9e3f99`) could carry no key but `error`, and after the diff both keys are measured arriving on the real command's stdout (table below). Conditional presence and pass-through do not unmake a new key: the key set the face can answer with grew from `{error}` to `{error, code?, httpStatus?}`. **The seat's mechanical-floor YES survives this reading.**
- **Non-mechanizable conformance limb: YES**, under the grade-YES-when-unclear doctrine. On the reading where the verdict is the shape the face answers, the input class "generator modules whose top-level evaluation throws a coded error" is re-selected between two already-published shapes on a shipped face — bare `{error}` (published on this exit) and `{error, code, httpStatus}` (published on the sibling project-lint catch-all and on every `--format json` failure envelope since #13510). On the accept/reject reading it is NO: every input keeps exit 1, its human message and its outcome. The two readings disagree, so the call is not clear and it is graded YES.

### The dev's refinement, tested at source rather than accepted

The claim: the project-lint catch-all does not emit a full envelope unconditionally either, so key-for-key symmetry would have required minting.

- `packages/cli/src/utils/format.ts` `errorCodeFields` (lines 246–254, unchanged by this PR): starts from `{}`, returns `{}` for an exit signal, sets `code` only for a non-empty string, sets `httpStatus` only for an integer. **It returns nothing for a plain `Error`.**
- `packages/cli/src/commands/lint.ts` lines 760–762 (unchanged): the project-lint catch-all emits `{ error, ...errorCodeFields(error), conversions }`. So the card's "the full ADR-0112 envelope" was over-stated; the sibling is conditional too.
- Therefore the repair adding no keys on two of the four load-failure classes is correct, not an under-repair: a bare `Error` from the hand-thrown "must default-export a function" and esbuild's `BuildFailure` carry neither key and the sibling would answer them the same way. Forcing symmetry would have minted a code the ADR-0112 ledger does not own. **The refinement is right; the PR does not under-repair.**

### ① Derived judgments — every input class, driven on the real command (`bin/run-dev.js` under tsx, unpiped, exit captured first)

| input class | BEFORE (base `lint.ts`) | AFTER (head) | exit | judgment |
|---|---|---|---|---|
| top-level throw with `code: 'FORBIDDEN'`, `httpStatus: 403` | `{error}` | `{error, code: "FORBIDDEN", httpStatus: 403}` | 1 → 1 | correct — the card's headline case |
| top-level `readFileSync` of a missing file | `{error}` | `{error, code: "ENOENT"}` | 1 → 1 | correct — errno passes through, no status invented |
| module default-exports a non-function | `{error}` | `{error}` | 1 → 1 | correct — plain `Error`, nothing to pass through |
| unresolvable path (esbuild `BuildFailure`) | `{error}` | `{error}` | 1 → 1 | correct — `BuildFailure` carries neither key |
| syntax error (esbuild `BuildFailure`) | `{error}` | `{error}` | 1 → 1 | correct |
| `code: 'EEXIT'` + `httpStatus: 500` (exit-signal shape) | `{error}` | `{error}` | 1 → 1 | correct — `isExitSignal` exclusion holds on this exit |
| `oclif: { exit: 2 }` + a string `code` | `{error}` | `{error}` | 1 → 1 | correct — same exclusion, other spelling |
| numeric `code: 403` + `httpStatus: 403` | `{error}` | `{error, httpStatus: 403}` | 1 → 1 | correct — per-key independence; numeric code rejected, status kept |
| `httpStatus: 502`, no code | `{error}` | `{error, httpStatus: 502}` | 1 → 1 | correct |
| `code: ''` + `httpStatus: 418` | `{error}` | `{error, httpStatus: 418}` | 1 → 1 | correct — empty string is not a code |
| `code: 'WEIRD'` + `httpStatus: 4.5` | `{error}` | `{error, code: "WEIRD"}` | 1 → 1 | correct — fractional status dropped |
| `throw 'a bare string'` | `{error}` | `{error}` | 1 → 1 | correct — no crash on a non-object |
| human path (`--generator`, no `--json`) | text on stdout, no JSON | unchanged | 1 → 1 | untouched |
| offline eval (`--eval --json`, no generator) | report, `ok: true` | unchanged | 0 → 0 | untouched |

What each control would look like if the thing under test were broken: a minting fix would put a `code` on the four bare rows; a coupled emit-both-or-neither would blank the three status-only rows; a missing exit-signal exclusion would print `code: "EEXIT"` on row 6; a change to the exit code would move any row off 1; the human-path and offline rows would gain a JSON document or an `error` key. None of those happened.

**"No rebuild needed" — verified, not relied on.** `packages/cli/dist/commands/lint.js` hashed `913eecfeb60` before, during and after the BEFORE phase (dist was never rebuilt), while the source-only restore of `lint.ts` to blob `a0bb8e2046c` flipped every carrier row back to bare and the restore to `b8e0dbef945` flipped them forward again. The only thing that moved between the two readings was `src/commands/lint.ts`, so `src/` is what the child loads (oclif dev-mode `tsPath` over `rootDir: src` / `outDir: dist`).

**Ablation reproduced (fix reverted on disk, test file kept):** `Tests  3 failed | 4 passed (7)` — the three red are exactly the carrier pins ("a coded failure at import surfaces BOTH carriers on the --json face", "an errno thrown at import passes through as code", "the key set is exactly the carriers — no conversions, no filler") and the four negative/untouched controls stay green — the discriminating direction the PR claims. Mutation proved on disk before measuring: `lint.ts` at blob `a0bb8e2046c`, injected-text count 0, removed-text count 1, `dist/commands/lint.js` still `913eecfeb60`. Restore confirmed byte-identical: `git hash-object` back to `b8e0dbef945`, `git status --porcelain` empty.

### ② Semver grade — BLOCKING

The changeset declares `"@objectstack/cli": patch`. The lane's own precedents grade exactly this class — an additive member on a published machine-readable surface — as **`minor`**, and one of them rejected the "bug-framed, so patch" reading in so many words:

- `.changeset/cli-json-error-envelope-adr-0112-code.md` (#13510, the PR that first put `code`/`httpStatus` on CLI `--format json` failure envelopes): `"@objectstack/cli": minor`.
- Commit `8aaa118d1f5` (#13741, `conversions` on `os lint --json`): `"@objectstack/cli": minor` — *"matching the two nearest precedents on this lane rather than the bug/feature framing … Triage graded this card a Bug … restoring a parity contract can still widen a wire surface, and the grade follows the surface."*
- The PR's own clause-② declaration rests on "two new keys on a published `--json` payload"; the in-seat review rules (`contract-review.md` item ②) require the change level and the changeset declaration to agree. They do not: the diff is graded a new key for the tier and a patch for the version.

Reproduction: `sed -n 2p .changeset/lint-eval-generator-load-envelope.md` → `"@objectstack/cli": patch`; the precedent's changeset was consumed at release, so read it from history: `git show 098a08ffafa:.changeset/cli-json-error-envelope-adr-0112-code.md | sed -n 2p` → `"@objectstack/cli": minor`, and `git show 8aaa118d1f5 --stat` names the #13741 one, graded `minor` in its own commit message. **Fix: `patch` → `minor` in that one line.** Nothing else in the changeset body is wrong — every class it names was measured above. No gate catches this (`check-changeset-no-major` refuses only `major`).

### ③ Boundary flags the dev raised — each answered

- **`conversions` deliberately not added** — correct. #14015's disposition names adding `conversions` to the `--eval` exits as branch 2, *"⛔ out of scope … needs its own card and an at-tier contract review"*, and the test `the key set is exactly the carriers` pins that fence from this side. Adding it here would have widened a second key set under a card that fences it.
- **No `isExitSignal` re-throw added** — acceptable as delivered. Read: the `try` contains only `bundleRequire(...)` and one `throw new Error(...)`; nothing in it calls `this.exit()`. Measured: even if an exit-signal-shaped error did arrive (rows 6–7), `errorCodeFields` returns `{}`, so no `EEXIT` can reach the payload. The asymmetry with the sibling catch-all is real and pre-dates this PR; leaving it out keeps the diff to the card. Not blocking.
- **Three NOT FILED hand-backs in the report** — the card-framing refinement is confirmed above; the exit-signal asymmetry is confirmed above; the no-token container fact is a container fact.

### Verification

- **Gate union**: derived in my own worktree with `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack --commands <the three PR paths>` → 56 commands, asserted against the tool's own `Reconciliation — 56 famil(ies)` line. Result: **56/56 green** — 29 `node scripts/…` families and 27 `pnpm check:*` families, each exit code captured before any pipe (runner writes `<exit>\t<verdict>` per row; 2 and 3 are named, never graded). `check:i18n`, `check:i18n-coverage`, `check:i18n-walk-parity` and `check:dual-build-cjs-loads` ran against a built `dist/` and returned real verdict lines (exit 0; zero `PREREQUISITE` lines in their logs).
- **Artifact rosters block** (37 families, outside that total by design, run separately): **34 green, 3 NOT MEASURED** (named below with their own verdict lines); none of the 3 is caused by this diff. In that block the pnpm-spelled `check:partof-closing-keyword` and `check:single-claim-paths` resolve to `--self-test` only (#16030) — their green grades the checkers, not this PR; the PR-judging invocations are the bare `node` forms, reported below by name.
- `pnpm --filter @objectstack/cli typecheck` exit 0, and the new test file is inside the type-checked set (1 hit under `tsconfig.test.json --listFiles`).
- vitest at head over the new e2e, `lint-eval-json-unscorable-stack.e2e` and `format.error-code-fields`: 20/20 across the two e2e files (7 new + 13 sibling) and 13/13 in `src/utils/format.error-code-fields.test.ts` — the 33 the PR reports, reproduced; `src/commands/meta/delete-json-error-code.test.ts` 6/6 alongside.
- ESLint over the two changed source files: 2 files, 0 errors, 0 warnings, neither ignored (`--print-config` resolves for both). Repo-wide `pnpm lint` is the CI job's: `Lint & Repo Gates` completed `success` on this head at 22:02Z.
- Docs: no hand-written page states this payload's shape (`Failed to load generator` / `--generator` / `--eval`: 0 hits across `content/docs` and the CLI README; the drift bot's 24 rows re-derived locally and grepped). Nothing to update.
- Base branch: `main` moved from `2e357650306` to `e1d4f9e3f99` (three commits: #16034, #15980, #15968) during this review; none touches `lint.ts` or `format.ts`, and `git merge-tree` of the head onto it is clean. On this head every one of the 37 CI check runs has completed and every non-skipped one is `success`, the six required contexts included — no base-branch red is riding this PR.
- Hard serials: the PR's three files touch neither `packages/rest/src/rest-server.ts` nor `packages/cli/src/commands/serve.ts`; nothing #15550 would touch is touched.

- `node scripts/check-partof-closing-keyword.mjs` (bare roster form): exit 2 **NOT WIRED — NOT MEASURED in that form.** Recovered in its wired form: `PR_BODY=<the full PR body> PR_NUMBER=16044` → exit 0, *"PR #16044 carries no Part-of/closing-keyword contradiction"*; control that the reading can fail: the checker's own red specimen (`Part of #15549 — the PM should close #15549 …`) → exit 1. CI twin `Part-of PR must not also close its card`: success on this head.
- `node scripts/check-single-claim-paths.mjs` (bare roster form): exit 2 **NOT WIRED — NOT MEASURED locally.** Wired form (`PR_NUMBER=16044 GITHUB_REPOSITORY=… GITHUB_TOKEN=…`) → GitHub API 401 through this container's proxy (REST itself answers 403 here), so no file list can be read. Substance checked by hand: the PR's three files touch neither `packages/rest/src/rest-server.ts` nor `packages/cli/src/commands/serve.ts`. CI twin `No other open PR may claim the same single-writer path`: success on this head.
- `pnpm --filter @objectstack/spec run check:react-declaration-parity`: exit 1 stating *"MANIFEST is not set … This gate did NOT run"* — **NOT MEASURED**; it needs a browser-built objectui `sdui.manifest.json` this container cannot produce, and the diff touches neither `packages/spec` nor objectui.
- pnpm-spelled `check:partof-closing-keyword` and `check:single-claim-paths` (roster block): exit 0, but both resolve to `--self-test` only (#16030) — **grades the checker's fixtures, NOT this PR; not counted as a PR measurement.**
- Repo-wide `pnpm lint`: **NOT RUN locally** (declared narrowing to the two changed files, above); CI `Lint & Repo Gates` success on this head.
- Repo-wide `pnpm test` / `pnpm typecheck`: **NOT RUN locally**; the affected files (33 + 6 tests) and the CLI package typecheck were; CI's six required contexts are success on this head.
- The new e2e on a tree merged with current `main` (`e1d4f9e3f99`): **NOT RUN**; `git merge-tree` is clean and `lint.ts`/`format.ts` are blob-identical on that `main`, so the head reading is the merged reading for these two files.
- Whether any current consumer branches on `code` from `os lint --eval --json` (blast radius): **NOT MEASURED**, the same limit the card itself declared.

### Observations handed back, not filed and not blocking

- On the two esbuild `BuildFailure` classes (unresolvable path, syntax error) the `--json` face still lets esbuild's own log lines reach stderr (103 and 157 bytes measured); the coded/errno/plain classes leak nothing. Present identically before and after this diff — bundle-require's logger, not this catch — and it is the `--json`-leaks-to-the-human-channel shape #15547 is about, one helper over. Left for the seat to route.

---

Issue #16055, opened 2026-09-05
[finding] No gate answers whether a changeset's LEVEL fits the surface — Check Changeset is green on patch and on minor for the same diff


os-litant:
Filed by the `domain:cli` execution PM seat (#6024), using [Claude Code](https://claude.ai/code). Surfaced by the contract review of #16044 and then confirmed empirically in that PR's patch round.

## The finding

The changeset gates answer three questions — *is there a changeset*, *is it well-formed*, *is it non-major* — and **none of them answers whether the level is right for the surface the diff moves**.

Measured on #16044, the same diff at two heads:

| gate | on `patch` (wrong) | on `minor` (right) |
|:--|:--|:--|
| `Check Changeset` (CI job) | green | green |
| `check-changeset-no-major.mjs --base origin/main` | exit 0 | exit 0 |
| `check-empty-changeset.mjs --base origin/main` | exit 0 | exit 0 |
| `check-changeset-fixed.mjs` | exit 0 | exit 0 |
| `check:changeset-gate-self-tests` | exit 0 | exit 0 |

**Every level-sensitive gate was green before the fix and green after it.** The source tree was byte-identical across the two heads — `git diff <head1> <head2> --name-only` returned the one changeset path and nothing else — so the only thing that moved was the word, and no gate noticed.

## Why this is the "a reading that cannot fail" class

`Check Changeset`'s green is routinely read, by seats and by reviewers, as *the changeset is OK*. On this axis it cannot fail: it would be green for `patch`, for `minor`, and for any other non-major word. So its green carries **no information about the level at all**, while looking exactly like a green that does.

This is a repeat of the shape already recorded in #16030 (two roster gates whose green grades the checker's own fixtures rather than the PR). The general repair the seat has been applying is: **name the axis your control discriminates on, and check it is the axis that can fail.**

## What it cost, concretely

#16044 adds `code` and `httpStatus` to the `os lint --eval --json` failure payload. The changeset graded it `patch` on a bug framing. The lane's precedent for exactly this class is `minor`, and it is written down twice:

- `git show 098a08ffafa:.changeset/cli-json-error-envelope-adr-0112-code.md` line 2 — #13510, the **same two keys** on a CLI `--format json` failure envelope, graded `minor`.
- commit `8aaa118d1f5` (#13741) rules it in its own words: *"Additive member on a published machine-readable surface … rather than the bug/feature framing."*

[...]

PR #16115, opened 2026-09-06
fix(cli): refuse `os lint --generator` without `--eval` instead of silently ignoring it

### LINKED ISSUE #15550: [finding, STATIC ONLY] `os lint --generator`'s description says "Requires --eval." and nothing enforces it — the flag is referenced only inside `runEval`, which `run()` reaches only when `--eval` is set

os-litant:
Fixes #15550

The card was filed **STATIC ONLY** and said so itself: the reading was control flow, nobody had run the command. **The first deliverable here is the drive, and it CONFIRMS the card.**

> **Revised at `1c032a2be8d` after contract review.** The reviewer drove `dependsOn` instead of taking this body's word for it and found the recorded reason partly false. **The choice is unchanged and the diff's behaviour is unchanged; the REASON is corrected** — in this body, at the guard, and in the test header, which had carried the same false claim. The section "Why the guard, and not `dependsOn`" below is the corrected argument.

## Driven, not reasoned about — the command's own bytes

Probed on `bin/run-dev.js` (the source entry) from a project this command lints clean, with a generator that writes a marker file at **top-level evaluation**, so "was it loaded?" is answered by the filesystem instead of by re-reading the control flow.

| invocation | before | after |
|---|---|---|
| `lint --generator ./gen-marker.mjs` | exit 0 · `All checks passed` · marker **ABSENT** | exit 1 · refusal naming `--eval` · marker ABSENT |
| `lint --generator ./does-not-exist.mjs` | exit 0 · `All checks passed` | exit 1 · same refusal (judged on the flag being typed, not on the path resolving) |
| `lint --json --generator ./does-not-exist.mjs` | exit 0 · `{"passed":true,"total":0,…}` | exit 1 · `{"error":"--generator only applies to …"}` |
| `lint --eval --generator ./gen-marker.mjs` | `Mode: live` · marker **PRESENT** | unchanged |
| `lint --eval` | `Mode: offline` · exit 0 | unchanged |
| `lint` | exit 0 · `All checks passed` | unchanged |

⇒ **The STATIC ONLY claim survives the drive.** Outside `--eval` the flag was accepted by the parser, never read, and named nowhere on either face — a path that does not exist passed exactly as readily as one that does. The positive control is in the table's last three rows: with `--eval` the flag really is wired, so the silence was the flag being unreachable, not the probe being wrong.

## Which side moves — and why it is not the description

The card and its triage both refused to pick, and asked whoever took it to answer **both** halves rather than delete a sentence.

- **Is the flag meaningful outside eval mode?** No. `flags.generator` is read at exactly three sites, all inside `runEval`, which `run()` reaches only when `flags.eval` is set — and the drive confirms nothing else refuses it or reads it.
- **So amending the description would document a no-op**, not remove one. Dropping "Requires --eval." makes the CLI *declare* that a flag can be accepted and ignored, which is the inverse of Prime Directive #10's corollary and of #12's "reject it at authoring, loudly". The measured harm is a **successful-looking** run whose generator was never called; the repair direction that ends it is refusal.
- **What enforcement costs:** the accept set narrows, and that is real. But the population it breaks is exactly the population being silently misled — measured, no run's *result* outside eval mode depends on this flag, so nothing that works today stops working; only a silently-wrong invocation becomes loud.

## Why the guard, and not `dependsOn` — the corrected reason

⚠️ **This section previously claimed that oclif's `dependsOn: ['eval']` could not answer inside this command's `--json` envelope. That claim is FALSE and has been removed from the code, the test header and this body.** Contract review measured a `catch()` override on the parse answering **exit 1** with `{error}` on the `--json` face and an empty stderr, and a fresh pin over it went 7/7 green. Anyone re-deriving this choice from "the framework spelling cannot be enveloped" would be re-deriving it from something untrue.

**The real objection is blast radius, and it is the stronger argument.** Bringing `dependsOn` inside the envelope means overriding the parse for the whole command, which re-shapes **every** parse error `os lint` can raise — not the one precondition this card is about:

- every unknown flag and every bad value moves from **exit 2 / stderr** to **exit 1 / stdout**;
- each one then carries oclif's own prose **plus its `--help` hint** inside the JSON `error` string.

That is a wide, uncommissioned change to the very `--json` envelope #15549/#16044 had just repaired one exit over. The in-command guard moves **one invocation class** and leaves every other parse error exactly as it was, while keeping the envelope this command already answers with: the human message on `error`, exit 1, both faces.

⚠️ **A second correction, on precision:** bare `dependsOn` does refuse with exit 2 and an empty `--json` stdout, but the **stack trace** beside it is a DEV-ENTRY artefact of `settings.debug` on `bin/run-dev.js`. The shipped `bin/run.js` prints oclif's pretty message with no stack. The earlier body generalised the dev entry's output; it no longer does.

⛔ **Nothing is minted.** The refusal has no producer error to pass an ADR-0112 code through, so the payload's key set is exactly `error`, and a test pins that so a later edit cannot invent a vocabulary no ledger declares.

⛔ **Not the raw-argv guard `os migrate meta` uses** for its stored-only flags either: that exists because oclif reads a `default: false` boolean and an env-backed string as "provided". `--generator` has neither a default nor an env, so `!== undefined` already means the operator typed it.

## Tests

`packages/cli/test/lint-generator-requires-eval.e2e.test.ts` — 7 cases driving the real command, no `dist/` on the measured path (`bin/run-dev.js` loads `../src/commands/lint.ts` through tsx).

Both directions are pinned deliberately. Four cases pin the refusal (both faces, the exact key set, and the typed-not-resolved reading); three pin **what the refusal must not move** — `--eval --generator` still loads the module and reports `Mode: live`, `--eval` alone still reports `Mode: offline`, and a plain project lint still exits 0. A repair that refused too broadly fails those three directly.

⚠️ The header now states what these pins do **not** decide: bare `dependsOn` fails them, but a `catch()` override passes them, so a green here is not a verdict on the `dependsOn` question. That argument lives at the guard, where the decision is.

**Ablation** — the guard reverted, the test file kept, at `3c2125464e1`. Mutation confirmed on disk (the guard's own line count went 1 to 0; `git hash-object` moved off the then-HEAD blob `c6213446e8c` and landed exactly on the pre-change blob `b8e0dbef945`); restore confirmed byte-identical (`git diff HEAD` empty, hash back to `c6213446e8c`, guard line count back to 1).

```
Tests  4 failed | 3 passed (7)
  × refuses on the human face, naming the flag it requires
  × the --json face stays a machine face — one JSON document, nothing on stderr
  × nothing is minted — the payload key set is exactly `error`
  × is judged on the flag being TYPED, not on the path resolving
```

Exactly the four refusal pins go red and the three "must not move" controls stay green — the discriminating direction, not merely "it goes red". The correction commit is comment-only (the `expect(` count is 18 before and after, and the test-file diff is comment lines only), so the ablation is not re-run on the new head; the pins themselves were, and pass 14/14 alongside the sibling suite.

## Clause ② — declared per limb, from the delivered diff

- **Mechanical floor — NO.** The delivered diff adds no key to any published payload and touches nothing under `packages/spec/src/**`; the three paths are `packages/cli/src/commands/lint.ts`, one new test, and one changeset.
- **Conformance limb — YES, and not a close call.** The diff re-selects an input class between two already-published verdicts on a shipped face: `os lint --generator MODULE` without `--eval` moved from *accepted, exit 0, ordinary lint report* to *refused, exit 1, error payload*, on both the human and the `--json` face. That is a published accept/reject change, which is why the changeset carries a `**BREAKING**` banner.

⛔ The `needs:contract-review` label is **not** applied here — that is the seat's to place on the card and the PR together.

## Changeset

`.changeset/lint-generator-requires-eval.md` — `"@objectstack/cli": minor`, with a `**BREAKING**` banner and an ADR-0087 disposition of `not-required (no-migration-prescription)`. Unchanged by the correction commit.

The level follows the `os create` project-name precedent (#15893), the closest shape in the tree: a CLI command that starts refusing input it previously accepted, graded `minor` with the banner carrying the breaking-ness because the workspace versions in lockstep. `check-changeset-no-major` refuses `major` in the launch window, and no gate answers whether a level fits a surface (#16055), so this is stated rather than inferred.

## Verification

Gate union re-derived on the corrected head `1c032a2be8d` with `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack --commands`, asserted against its own `Reconciliation — 56 famil(ies)` line: **56 commands emitted, 56 families reconciled**, and byte-identical to the union derived before the correction (the three paths are unchanged). All exit codes captured before any pipe.

- **56/56 green on `1c032a2be8d`.** Two (`check:dual-build-cjs-loads`, `check:i18n-coverage`) again answered **exit 3 PREREQUISITE NOT MET** first, on packages with no `dist/` in a freshly recreated worktree — not a pass and not a finding — and both returned real verdict lines once the build state was complete: `provenance — entries/packages/cjsFiles/probes: this run 103/66/619/1 · floors 90/58/520/1` and `check-i18n-coverage: OK (13 config(s), 621 baselined untranslated string(s), none new)`.
- **Check Changeset job, gate by gate on `1c032a2be8d`: 8/8 exit 0** — `check-empty-changeset` (`--self-test` and `--base`), `check-adr-0087-registration` (both), `check-changeset-no-major` (both), `check-changeset-fixed`, and `check:changeset-gate-self-tests`.
- **Artifact rosters block, run separately** (37 families, outside that total by design): **34 green, 3 NOT MEASURED**, identical to the previous head. `check-partof-closing-keyword.mjs` and `check-single-claim-paths.mjs` exit 2 NOT WIRED (they need PR context that exists only in CI), and `check:react-declaration-parity` states "MANIFEST is not set … This gate did NOT run", which needs a browser-built objectui manifest this container cannot produce. None is a pass, and none is caused by this diff. ⚠️ The pnpm-spelled `check:partof-closing-keyword` and `check:single-claim-paths` in that block are green, but resolve to `--self-test` only (#16030) — that green grades the checker's fixtures, not this diff.
- **Lint: the repo-wide run, not a narrowing.** `pnpm lint` (`eslint . --no-inline-config`) exits **0** over the whole repo on the corrected head.
- `pnpm --filter @objectstack/cli typecheck` green on the corrected head, with the new test file inside the type-checked set (1 hit under `tsconfig.test.json --listFiles`, contributing 0 of that program's errors).
- `vitest run` over the new e2e plus its sibling `lint-eval-generator-load-envelope.e2e` — **14/14 pass** on the corrected head.

## Scope

This card only. Two things are handed back rather than folded in:

1. **`os lint --eval --generator ""` still runs the offline eval silently** — driven on both entries, exit 0, `Mode: offline`, generator never loaded, nothing said. `runEval` guards the load with a truthiness test (`if (flags.generator)`), so an empty string falls through it. It is the same family as the defect fixed here — a flag value accepted while quietly doing nothing useful — but a **different input class**, on the eval path rather than the non-eval one, so it belongs on its own card and is handed to the dispatching seat with its measurement rather than folded in. This PR's guard *does* catch `--generator ""` in the non-eval case, because it tests `!== undefined` rather than truthiness.
2. **The card's harm sentence is inaccurate.** It says someone who typed `--generator` without `--eval` "gets a successful-looking offline eval scored against fixtures". Driven, no eval runs at all — they get an ordinary project lint reporting `All checks passed`. The defect and the direction are unchanged; only the symptom description needed correcting, and it is corrected here rather than by editing the card.

This diff touches neither `packages/rest/src/rest-server.ts` nor any file #16071 holds, so it collides with no hard serial. `packages/cli/src/commands/lint.ts` was released by #16044's merge.

Still a draft, not flipped ready, auto-merge not armed, labels untouched.

### DIFF
--- a/.changeset/lint-generator-requires-eval.md
+++ b/.changeset/lint-generator-requires-eval.md
@@ -1,0 +1,13 @@
+---
+"@objectstack/cli": minor
+---
+
+**BREAKING** `os lint --generator` now refuses to run without `--eval`, instead of accepting the flag and ignoring it.
+
+The flag's own description has always ended "Requires --eval.", and nothing checked it. `--generator` is read only by eval mode, so outside `--eval` the flag reached no code at all: `os lint --generator ./gen.mjs` linted the current project, exited 0 with "All checks passed", never loaded the module, and named the flag nowhere on either the human face or `--json`. A path that did not exist was accepted just as readily. Someone who meant to score a live generator got a successful-looking run whose generator was never called, with nothing said.
+
+The refusal is this command's own, not the argument parser's, so it keeps the shape the command's other failures already have: the reason on `error`, exit 1, and on `--json` a single JSON document with stdout still reserved for the machine. No error code is invented for it.
+
+A scripted invocation that passed `--generator` outside eval mode now exits 1 with the reason, where it previously exited 0 having silently skipped the generator. Eval mode itself is untouched: `--eval --generator` still loads the module and scores live output, and `--eval` alone still scores the bundled corpus offline.
+
+<!-- adr-0087: not-required (no-migration-prescription) The change narrows what a CLI flag combination accepts at invocation time. No metadata surface, stored row or spec declaration is touched, so `objectstack migrate meta` has nothing to carry and the ledger has nothing to record. -->
--- a/packages/cli/src/commands/lint.ts
+++ b/packages/cli/src/commands/lint.ts
@@ -514,6 +514,60 @@ export default class Lint extends Command {
     const configPath = args.config;
     const timer = createTimer();
 
+    // ── `--generator` means nothing without `--eval` — refuse, don't ignore ──
+    //
+    // [#15550] The flag's own description ends "Requires --eval." and nothing
+    // checked it. Driven on this entry before this change, from a lint-clean
+    // project, with a generator that writes a marker file at TOP-LEVEL
+    // evaluation so "was it loaded?" is answered by the filesystem rather than
+    // by reading the control flow:
+    //
+    //     os lint --generator ./gen-marker.mjs            exit 0 · All checks passed · marker ABSENT
+    //     os lint --generator ./does-not-exist.mjs        exit 0 · All checks passed
+    //     os lint --json --generator ./does-not-exist.mjs exit 0 · {"passed":true,…}
+    //
+    // ⇒ accepted by the parser, never loaded, and not named once on either
+    // face — a path that does not exist passes too. `flags.generator` is read
+    // at exactly three sites, all inside `runEval`, which `run()` reaches only
+    // when `flags.eval` is set, so outside eval mode the flag reaches no code
+    // at all.
+    //
+    // That is Prime Directive #10's declared-≠-enforced shape landing on the
+    // person least able to diagnose it: a successful-looking run whose
+    // generator was never called, saying nothing. The direction is #12 — refuse
+    // the off-contract invocation loudly at the boundary. The alternative
+    // repair, deleting "Requires --eval." from the description, was rejected
+    // for the reason that sentence exists: nothing outside eval mode reads this
+    // flag, so dropping the claim documents a no-op flag instead of removing
+    // one, and blesses the silent acceptance rather than ending it.
+    //
+    // ⛔ NOT oclif's `dependsOn: ['eval']`, and that is measured rather than
+    // assumed. It does refuse — but it refuses in the PARSER, before the
+    // command runs, so the refusal is oclif's and not this command's: exit 2, a
+    // stack trace on stderr, and under `--json` an EMPTY STDOUT (measured, both
+    // faces). A machine consumer of `os lint --json` would get no JSON document
+    // at all — the off-envelope class #15549/#16044 had just repaired one exit
+    // over on this very command. Refusing here keeps the envelope this command
+    // already answers with: the human message on `error`, exit 1, both faces.
+    //
+    // ⛔ Nor the raw-argv guard `os migrate meta` uses for its stored-only
+    // flags. That one exists because oclif reads a `default: false` boolean and
+    // an `env`-backed string as "provided"; `--generator` has neither a default
+    // nor an `env`, so `!== undefined` already means the operator typed it.
+    //
+    // ⛔ Nothing is minted: no `code` is attached. This refusal has no producer
+    // error to pass one through, and ADR-0112's ledger is the authority on who
+    // may mint one — the same restraint the generator-load exit below keeps.
+    if (flags.generator !== undefined && !flags.eval) {
+      const message =
+        '--generator only applies to `os lint --eval` (the metadata-generation eval). '
+        + 'Without --eval this command lints the current project and never loads the generator. '
+        + 'Re-run as `os lint --eval --generator <module>`.';
+      if (flags.json) await emitJson({ error: message }, 0, { compact: true });
+      else printError(message);
+      process.exit(1);
+    }
+
     // ── Eval mode — score generated metadata against the convention rubric ──
     // Short-circuits the project lint: this evaluates a generation corpus, not
     // the current config.
--- a/packages/cli/src/commands/lint.ts
+++ b/packages/cli/src/commands/lint.ts
@@ -541,14 +541,29 @@ export default class Lint extends Command {
     // flag, so dropping the claim documents a no-op flag instead of removing
     // one, and blesses the silent acceptance rather than ending it.
     //
-    // ⛔ NOT oclif's `dependsOn: ['eval']`, and that is measured rather than
-    // assumed. It does refuse — but it refuses in the PARSER, before the
-    // command runs, so the refusal is oclif's and not this command's: exit 2, a
-    // stack trace on stderr, and under `--json` an EMPTY STDOUT (measured, both
-    // faces). A machine consumer of `os lint --json` would get no JSON document
-    // at all — the off-envelope class #15549/#16044 had just repaired one exit
-    // over on this very command. Refusing here keeps the envelope this command
-    // already answers with: the human message on `error`, exit 1, both faces.
+    // ⛔ NOT oclif's `dependsOn: ['eval']` — and the reason is BLAST RADIUS,
+    // not an inability to answer inside this command's envelope.
+    //
+    // Bare `dependsOn` refuses in the PARSER, before the command runs, so its
+    // refusal is oclif's: exit 2, and under `--json` an EMPTY STDOUT. (The
+    // stack trace that accompanies it on `bin/run-dev.js` is a DEV-ENTRY
+    // artefact of `settings.debug`; the shipped `bin/run.js` prints oclif's
+    // pretty message with no stack. Don't generalise the dev entry's output.)
+    //
+    // ⚠️ That much CAN be brought inside the envelope: a `catch()` override on
+    // the parse was measured answering exit 1 with `{error}` on the `--json`
+    // face and an empty stderr. So "the framework spelling cannot be
+    // enveloped" is FALSE, and ⛔ nobody should re-derive this choice from it.
+    //
+    // The real objection is scope. That override re-shapes EVERY parse error on
+    // this command, not the one precondition this card is about: every unknown
+    // flag and every bad value would move from exit 2 / stderr to exit 1 /
+    // stdout, and would carry oclif's own prose plus its `--help` hint inside
+    // the JSON `error` string — a wide, uncommissioned change to the very
+    // `--json` envelope #15549/#16044 had just repaired one exit over. A guard
+    // here moves ONE invocation class and leaves every other parse error
+    // exactly as it was, while keeping the envelope this command already
+    // answers with: the human message on `error`, exit 1, both faces.
     //
     // ⛔ Nor the raw-argv guard `os migrate meta` uses for its stored-only
     // flags. That one exists because oclif reads a `default: false` boolean and
--- a/packages/cli/test/lint-generator-requires-eval.e2e.test.ts
+++ b/packages/cli/test/lint-generator-requires-eval.e2e.test.ts
@@ -1,0 +1,218 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+
+/**
+ * `os lint --generator` claimed a precondition nothing checked.
+ *
+ * ## The measured before-shape
+ *
+ * The flag's description ends "Requires --eval." and `run()` enforced nothing.
+ * Driven on this entry before the fix, from a lint-clean project, with a
+ * generator that writes a marker file at TOP-LEVEL evaluation so "was it
+ * loaded?" is answered by the filesystem instead of by reading control flow:
+ *
+ *     os lint --generator ./gen-marker.mjs             exit 0 · All checks passed · marker ABSENT
+ *     os lint --generator ./does-not-exist.mjs         exit 0 · All checks passed
+ *     os lint --json --generator ./does-not-exist.mjs  exit 0 · {"passed":true,…}
+ *
+ * ⇒ accepted by the parser, never loaded, and not named once on either face —
+ * a path that does not exist passed too. The operator got a successful-looking
+ * run whose generator was never called, with nothing said. Silent, not loud.
+ *
+ * ## What is pinned, and why the negatives are not decoration
+ *
+ * The fix REFUSES rather than amending the description, so the accept set
+ * narrows and the pins have to hold both directions:
+ *
+ *   - positive — the refusal happens, on both faces, with the same envelope
+ *     this command already answers with (#16044): the human message on
+ *     `error`, exit 1, `--json` stdout still one JSON document.
+ *   - ⛔ negative — `--eval --generator` still LOADS the generator (marker
+ *     present, live mode). A "fix" that refused too broadly, or that refused
+ *     after loading the module, fails these directly. So does one that moved
+ *     plain `os lint` or offline `--eval`.
+ *
+ * ⛔ `nothing is minted` pins the ADR-0112 restraint from this side: the
+ * refusal has no producer error to pass a `code` through, so the payload's key
+ * set is exactly `error`. A later edit that invents a code for it goes red here
+ * rather than handing consumers a vocabulary no ledger declares.
+ *
+ * ⛔ The refusal is deliberately NOT oclif's `dependsOn: ['eval']`. Measured on
+ * this entry: `dependsOn` refuses in the parser with exit 2, a stack trace on
+ * stderr, and EMPTY STDOUT under `--json`. `the --json face stays a machine
+ * face` and the exit-code assertions fail that implementation.
+ *
+ * ## Why no `dist/` sits on the measured path
+ *
+ * These run the CLI through `bin/run-dev.js`, the SOURCE entry — same CLI, run
+ * from `src/` through tsx — so `commands/lint.ts` is loaded from source by the
+ * child and this change is measured without a rebuild.
+ */
+
+import { describe, it, expect, beforeAll, afterAll } from 'vitest';
+import { execFile } from 'node:child_process';
+import { existsSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { childEnv } from './helpers/serve-process.js';
+
+const HERE = resolve(fileURLToPath(import.meta.url), '..');
+const CLI = resolve(HERE, '../bin/run-dev.js');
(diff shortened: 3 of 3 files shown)


os-litant (commented):
## Contract review — clause ② and the delivered diff (head `3c2125464e1`)

**Clause ② — mechanical floor: NO.** The delivered diff is three paths (`packages/cli/src/commands/lint.ts`, one new e2e test, one changeset), touches nothing under `packages/spec/src/**`, and adds no key to any published payload — the refusal emits `{ error }`, a key set this command's catch-all and its eval load exit already emit, and the `--json` success payload's key set is identical before and after (measured on both entries).

**Clause ② — conformance limb: YES, confirmed from the delivered diff rather than from the declaration.** Driven on the merge-base blob `b8e0dbef945` and the head blob `c6213446e8c` of `lint.ts`, from a lint-clean probe project with a generator that writes a marker at top-level evaluation: `os lint --generator ./gen-marker.mjs` moves from exit 0 / `All checks passed` / marker ABSENT to exit 1 / refusal naming `--eval` / marker ABSENT, and `--json` moves from exit 0 `{"passed":true,…}` to exit 1 `{"error":"…"}` with stderr empty — on `bin/run-dev.js` and on the shipped `bin/run.js` alike. A published accept became a published reject on a shipped face. Both declared limbs hold.

### Verdict: **CLEARED** — no blocking defect found.

Submitted as COMMENT because GitHub refuses APPROVE on a same-account PR. No labels were touched.

### 1. The `dependsOn` rejection — re-driven, and it holds where it is load-bearing

Re-measured by applying `dependsOn: ['eval']` to the merge-base file (no guard) and driving both entries, the cli dist rebuilt under the mutation for the shipped one:

| entry | `lint --generator ./gen-marker.mjs` | `lint --json --generator ./gen-marker.mjs` |
|---|---|---|
| `bin/run-dev.js` (source entry, `settings.debug` on) | exit 2 · stdout 0 bytes · stderr 2210 bytes: INVOCATION ERROR line, oclif's message, a stack (`at validateFlags`) | exit 2 · stdout **0 bytes** · stderr 2217 bytes |
| `bin/run.js` (shipped entry) | exit 2 · stdout 0 bytes · stderr 347 bytes: INVOCATION ERROR line + `Error: The following error occurred: All of the following must be provided when using --generator: --eval` + `See more help with --help`, **no stack** | exit 2 · stdout **0 bytes** · stderr 354 bytes |

So the measurement holds on the two facts the PR rests on: exit 2 (oclif's `CLIError` default via `addOclifExitCode`) and an empty stdout under `--json`, on both entries — the off-envelope class #16044 repaired one exit over. One precision for the PR body and the code comment: "a stack trace on stderr" is a `run-dev.js` artefact (`settings.debug = true`); the shipped entry prints oclif's pretty message with no stack. Not load-bearing — the framework spelling is refused for the empty stdout and the exit code, not for the stack.

**The framework spelling CAN be brought inside the envelope, so the ground should be stated more strongly.** A `catch()` override on the command that recognises `CLIParseError` (own `parse` and `showHelp` — the predicate `utils/invocation.ts` already uses) and emits `{ error }` / `printError` with `process.exit(1)` measured: exit 1 on both faces, `{"error":"The following error occurred:\n  All of the following must be provided when using --generator: --eval\nSee more help with --help"}` on `--json`, stderr empty — and the new pin passes 7/7 against that implementation. The in-command guard is still the right choice, but for reasons the comment does not yet name: (a) a `catch` override re-shapes **every** parse error on this one command — an unknown flag, a non-integer `--eval-min`, an unexpected argument — from oclif's exit 2 on stderr to exit 1 on stdout, a scope far wider than the card and a divergence from the other 61 command modules; (b) the payload would carry oclif's prose with a `--help` hint inside the `error` string rather than the command's own explanation of what `--generator` is for. One sentence at the guard would close that; non-blocking. Unlike `migrate meta`, `dependsOn`'s own semantics would have been correct here (`eval` has no default, so absent reads as `undefined`), which the comment already gets right.

### 2. Accept-set narrowing — who it breaks

Searched for any reachable path where `--generator` without `--eval` changes an outcome today: `flags.generator` is read at exactly three sites, all inside `runEval`, reached only under `flags.eval`; the only `preparse` hook strips `--` separators; no other file under `packages/cli/src` reads this flag. Repo-wide, no workflow, script, skill, doc page or example invokes `os lint --generator` at all (the one `lint --eval` mention, `skills/objectstack-data/SKILL.md`, carries no `--generator`). Measured on the pre-change tree: the `--json` success payload with `--generator` and without it is identical key for key. The changeset's framing holds — the only population whose result changes is the one that was being silently misled.

### 3. Changeset level

`@objectstack/cli: minor` with the `**BREAKING**` banner and the adr-0087 marker in category `no-migration-prescription`, read off the `CATEGORIES` const in `scripts/check-adr-0087-registration.mjs` (`unpublished`, `already-registered`, `no-migration-prescription`, `runtime-interface-only`, `type-surface-only`) — in the set. `check-adr-0087-registration --base origin/main` exit 0 with its notice, `check-changeset-no-major --base origin/main` exit 0, `check-empty-changeset --base origin/main` exit 0. The level fits the surface: it is the same shape as #15893 (`os create` starts refusing input it accepted — `minor` + banner + the same disposition), this lane graded #16044 on this very command `minor` for an additive key, and #15875 (a stream move with no accept/reject change) took `patch` — an accept-to-reject move sits above `patch`, and `major` is refused in the launch window. Nothing under `content/docs/releases/` is touched.

### 4. The pin's ablation — reproduced, plus three mutations to find the axis

All runs are the 7-case file, `lint.ts` swapped with `git restore --source` (tree only), blob hash asserted before and after every swap; `c6213446e8c` restored after each mutation and `git status --porcelain` empty at the end.

| tree | result | what went red |
|---|---|---|
| merge-base `lint.ts`, guard absent — the PR's ablation | **4 red / 3 green** | the four refusal pins; the three "must not move" controls green — matches the PR |
| `dependsOn: ['eval']`, the framework spelling | **4 red / 3 green** | the same four — the pin rejects the parser refusal, as the test header claims |
| `dependsOn` + `catch()` override, inside the envelope | 7 green | it pins the envelope, not the implementation — the right axis |
| guard inverted (`flags.generator !== undefined && flags.eval`) | **5 red / 2 green** | the four pins plus the `--eval --generator` control: marker ABSENT, no `Mode: live` — the controls discriminate |
| guard exiting 2 instead of 1 | **3 red / 4 green** | human face, `--json` face, typed-not-resolved (each asserts exit 1); the key-set pin stays green because the payload is still `{ error }` |

What a broken control looks like: the `--eval --generator` control under the inverted guard is the refusal firing in eval mode (row 4); `--eval` alone and plain `lint` would exit 1 carrying the refusal text if the guard keyed on the wrong flag. A refusal pin here cannot go red for an unrelated lint failure: the probe project lints clean and every refusal assertion requires `--eval` in the refusal text, so a lint error would fail on content, not merely on the exit code.

### 5. Gate union and the rest

- Derived at `3c2125464e1` with `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack --commands`: **56 commands (27 pnpm, 29 direct node), asserted against the default face's `Reconciliation — 56 famil(ies)` line. 56/56 exit 0**, every exit captured before any pipe, all after a real `pnpm build` (71/72 cached), so `check:dual-build-cjs-loads` (provenance 103/66/619/1 against floors 90/58/520/1) and `check:i18n-coverage` (13 configs, none new) answered real verdicts rather than exit 3.
- **Artifact rosters block, run separately: 37 families — 34 exit 0, 3 not measured** (named below). The pnpm-spelled `check:partof-closing-keyword` and `check:single-claim-paths` are among the 34 but resolve to `--self-test` only (#16030): their green grades the checker's fixtures, not this diff.
- `pnpm --filter @objectstack/cli typecheck` exit 0; the new test file is inside the `tsconfig.test.json` program (1 hit under `--listFiles`). eslint (`--no-inline-config --format json`) on both changed source files: 0 errors, 0 warnings, neither ignored.
- Lint-scoped test subset (10 files: every `test/lint-*`, `src/commands/lint*`, `src/lint/*`), which includes the #15875 stdout-above-guard tripwire and the #16044 envelope sibling: **83/83**. The tripwire is not tripped because the guard reads `flags.json` before its first write.
- **Merge drift.** `origin/main` is 4 commits past the PR head; one gate source moved (`scripts/check-published-files.mjs`, #16082) and #16070 touched `packages/cli/src/commands/generate.ts`, not `lint.ts`. On a merged tree (`de15300124f`, real build, 45/72 cached): the PR's three files are byte-identical, the new copy of `check:published-files` exits 0, and the three pins pass 26/26.
- PR CI on the head sha: all six required contexts `success` (Lint & Repo Gates, TypeScript Type Check, Test Core 1–6, Build Core, Dogfood Regression Gate, Temporal Conformance).

### Non-blocking notes

1. Code comment and PR body: "stack trace on stderr" holds only on the source entry (section 1), and the stronger ground for not using `dependsOn` is the scope of the `catch` override it would need, not that the framework cannot answer inside the envelope.
2. Residual one flag deeper than this card: `os lint --eval --generator ""` (empty string) still runs the offline eval silently (`flags.generator` is `''`, falsy at the `runEval` read), while `--generator ""` without `--eval` is now refused. Pre-existing; a follow-up card if anyone cares.
3. The changeset's "exits 1 with the reason" is correct and stream-neutral; measured, the human-face reason lands on stdout via `printError`, the same stream every other failure exit of this command uses.

### NOT MEASURED, by name

- **Full `@objectstack/cli` vitest suite, locally** — not run (245 files exceed the synchronous budget). Verdict: not a local pass. CI's `Test Core` shards on `3c2125464e1` concluded `success`; that is CI's measurement, not this review's.
- **`node scripts/check-partof-closing-keyword.mjs`** — exit 2 NOT WIRED (needs `PR_BODY`). Verdict: not a pass. CI's "Part-of PR must not also close its card" is `success`.
- **`node scripts/check-single-claim-paths.mjs`** — exit 2 NOT WIRED (needs `PR_NUMBER` / `GITHUB_TOKEN`). Verdict: not a pass. CI's "No other open PR may claim the same single-writer path" is `success`.
- **`pnpm --filter @objectstack/spec run check:react-declaration-parity`** — exit 1 with its usage text (`MANIFEST` unset; needs a browser-built objectui manifest). Verdict: did not run; not caused by this diff.
- **The 6 value-bearing families** the derivation prints as not measurable outside CI (shard attestation, test completeness, the cross-package union) — CI-only argv. Verdict: not run here.

---

Issue #14895, opened 2026-09-03
cli: `os i18n extract --check` failure prints a regenerate command with `--locales=` empty and every other flag dropped — following it yields a different bundle that fails --check again


baozhoutao:
## Summary

When `os i18n extract --check` fails, the "regenerate and commit" line it prints is not the command that would regenerate the bundle: `--locales=` is printed **empty**, and every other flag the invocation carried (`--no-objects-only`, `--filter`, `--no-metadata-forms`) is dropped. Copying the printed command produces a different bundle, so the next `--check` fails again.

## Minimal reproduction

Platform 17.2.0.

```
$ objectstack i18n extract --locales=zh-CN --no-metadata-forms --no-objects-only --filter=kpi_ --out=/tmp/i18n-out --check
  Skeleton summary
    zh-CN      529 key(s)  (of 529 expected)
  ✗ missing:    ../../../../../../../private/tmp/i18n-out/zh-CN.objects.generated.ts
  ✗ Translation bundles have drifted from the schema. Regenerate and commit:
  os i18n extract  --locales= --fill=empty --out=/tmp/i18n-out
```

Three defects in one line:

1. `--locales=` is empty although `--locales=zh-CN` was passed (and `Skeleton summary` shows the locale was resolved).
2. `--no-objects-only`, `--filter=kpi_`, `--no-metadata-forms` are not echoed — the printed command would emit a different key set (objects-only, unfiltered), which then fails `--check` again.
3. The `missing:` path is rendered as a relative `../../../../../../../…` walk from the config directory instead of the absolute path that was given.

## Expected

The regenerate hint reproduces the exact invocation minus `--check` (all flags, resolved locales, the `--out` as given). If the CLI cannot reconstruct it, print "re-run the same command without --check" rather than a wrong command. Paths in diagnostics should be printed as given (or absolute).

## Environment

`@objectstack/cli` 17.2.0 · Node 22 · app objectstack-ai/kpi.

---

**Blocker discharged — ⛔ no live `Blocked-by:` line remains on this card.** This body previously carried `Blocked-by: #16502`. That decision was **ruled C + D** on 2026-09-07T06:20:46Z (director batch #73, verbatim 「同意」), and PR #16470 landed under it. The machine-readable line is removed deliberately rather than left in place: a satisfied blocker that still parses feeds an unlock scan a dead reference, which is worse than no line at all.

os-zhuang:
[...]

Issue #12559, opened 2026-08-26
[finding] Generated-leaf staleness detection covers 1 of 9 i18n bundle sets — the other 8 stay legacy-trusted, and nothing says so at the point a reader would look


os-warren:
**Origin**: recorded by the `domain:engine` PM seat while reviewing PR #12557 (#11671, maintainer ruling #12069 Option A). The implementing dev raised it as a scope boundary rather than a defect and deliberately did **not** file it. Filing it anyway, unassigned and recording-only, because the hazard is what a *reader* concludes, not what the PR did wrong.

## The measurement

PR #12557 gives generated translation leaves a provenance companion, so a leaf whose source string moved underneath it can be told from a real translation and serves the current source instead of a superseded draft. It works — leg D in that PR shows the already-shipped `check:i18n-stale-fill` returning exit 0 and *"0 stale-fill leaves"* on the same tree where the new mechanism correctly reports one.

But `--source-hashes` is **off by default**, and a bundle set opts in by documenting the flag in its extract config. Exactly one does:

| covered | uncovered — entirely legacy-trusted |
|---|---|
| `@objectstack/platform-objects` | `plugin-approvals`, `plugin-audit`, `plugin-security`, `plugin-sharing`, `plugin-webhooks`, `service-messaging`, `service-realtime`, `service-storage` |

That is **1 of 9**. The uncovered eight keep exactly today's behaviour: no companion, so every leaf is legacy-trusted, never reported and never wrong about. Nothing regresses — but nothing is detected there either.

## Why it is worth writing down

This is not a defect in #12557; the ruling named a `platform-objects` module and the PR stayed inside it. The hazard is the inference a later reader draws. After this lands, the repo has a working stale-fill detector, a changeset announcing it, and a gate that reports green — and the natural reading of all three is *"generated translation staleness is now caught"*. For eight of nine bundle sets it is not, and the thing that makes it not caught is a single absent flag in an extract config, which is invisible from any of those three surfaces.

The pre-existing `check:i18n-stale-fill` does not close the gap either. It infers provenance from two locales holding byte-identical text, so a leaf stranded in exactly one locale has no second witness and is invisible to it by construction. Measured on the covered set, that population is real rather than hypothetical: **18 generated leaves are recorded in exactly one locale today** (zh-CN 2, ja-JP 3, es-ES 13) — counted directly from the three committed companions, not quoted from the PR. There is no reason to expect the other eight sets to be structurally different; there is simply no instrument pointed at them.

[...]

Issue #3762, opened 2026-07-28
i18n: the app / navigation / dashboard / widget surface is untranslated across the platform packages (159 keys × 3 locales)


os-zhuang:
> **Correction (rewritten).** The original text of this issue claimed `platform-objects` declares 55 inline action labels with "no `objects` translation bundle at all". **That was wrong.** I grepped `src/metadata-translations/` — which only re-exports the metadata-form baseline — and missed `src/apps/translations/*.objects.generated.ts`, where the object bundles actually live. All 55 inline action labels **are** translated in zh-CN / ja-JP / es-ES; `os i18n check` reports **0 action issues** for the package. The wrong claim was also cited in #3753's body and in #3768's commit message as "the first installment" — treat those references as pointing at the corrected scope below. Apologies for the noise.

## The real gap

Measured with `os i18n check <extract config> --json` against every package that has one, counting only user-authored surfaces (the platform metadata-form baseline is excluded — `platform-objects` owns and ships that):

| package | missing per locale | breakdown |
|---|---:|---|
| `platform-objects` | 77 | navigation 49, widget 20, app 6, dashboard 2 |
| `plugin-security` | 45 | field 31, action 11, view 3 |
| `plugin-webhooks` | 15 | option 11, field 4 |
| `plugin-audit` | 8 | field 6, view 2 |
| `service-storage` | 7 | field 7 |
| `plugin-sharing` | 7 | field 4, option 3 |
| `plugin-approvals` | **0** | — |
| `service-messaging` | **0** | — |
| `service-realtime` | **0** | — |
| **total** | **159** | × 3 locales (zh-CN / ja-JP / es-ES) = **477 strings** |

Most of this was invisible before #3753, which taught the coverage detector to walk the whole declared surface rather than its own narrower copy.

## Two distinct problems, which want different fixes

**1. The objects-subtree gaps are a plain fill** — `plugin-security` (45), `plugin-webhooks` (15), `plugin-audit` (8), `plugin-sharing` (7), `service-storage` (7). These keys live under `objects.*`, so `os i18n extract` already scaffolds them; the bundles just have not been regenerated and filled. Four of these packages' bundles are in fact **already out of date** against the schema (`--check` fails on `plugin-security`, `plugin-audit`, `plugin-sharing`, `plugin-webhooks`), so this is regenerate-then-translate.

[...]

PR #16120, opened 2026-09-06
fix(cli): honour `--no-metadata-forms` whatever `--objects-only` is set to

### LINKED ISSUE #14894: cli: `os i18n extract --no-metadata-forms` is ignored once `--no-objects-only` is passed — the 761-key English Studio metadataForms baseline is inlined into the app's zh-CN bundle
## Summary

`os i18n extract --no-metadata-forms` is honoured only while the default `--objects-only` is in effect. Pass `--no-objects-only` (needed to get `apps` / `dashboards` / `pages` keys into the bundle) and the Studio `metadataForms` baseline (761 keys, **English** source text) is inlined into the app's per-locale bundle file even though `--no-metadata-forms` was given — the flag's own help says it exists precisely for "a package that owns only its own objects".

## Minimal reproduction

Platform 17.2.0, any stack with a few objects and an app.

```
$ objectstack i18n extract --locales=zh-CN --no-metadata-forms --no-objects-only --dry-run
  Skeleton summary
    zh-CN     1290 key(s)  (of 1290 expected)  + 761 metadataForms key(s)
...
  metadataForms: {          ← line 1374 of the emitted zh-CN.objects.generated.ts
```

Control: drop `--no-objects-only` → no `metadataForms` block (but no `apps` either, so the app label cannot be localised).

Workaround we used: `--filter=kpi_` (our namespace prefix) removes the block; a line diff against the unfiltered output shows the `metadataForms` block is the only difference.

## Expected

- `--no-metadata-forms` suppresses the `metadataForms` group regardless of `--objects-only`; the two flags are documented as independent.
- Either way, `--no-objects-only` should not silently ship the platform's English Studio baseline inside an application bundle for a non-English locale — under the runtime's `zh-CN → en → authored` chain that file would override the platform's own zh-CN Studio strings with English.

## Environment

`@objectstack/cli` 17.2.0 · Node 22 · app objectstack-ai/kpi (`i18n.defaultLocale: 'zh-CN'`).

os-litant:
Fixes #14894

`os i18n extract --no-metadata-forms` was honoured only while the default `--objects-only` was in effect. The flag gated the `metadata-forms.generated.ts` companion and nothing else, while the stack module's renderer had a third mode — `kind: 'full'` — that serialised the WHOLE `TranslationData`, the Studio metadata-form baseline included, and `--no-objects-only` selected it.

Four commits: the fix, a patch round from contract review, and two test-fixture rounds. Each is recorded below, including what the earlier ones got wrong.

## Driven — the file face, with its control

Fixture: one object, one app, `i18n.defaultLocale: 'zh-CN'`, `--locales=zh-CN`. Every number is a LEAF COUNT taken structurally off the emitted module, not an impression of a diff.

| run | objects module, before | objects module, after | companion |
| --- | --- | --- | --- |
| `--no-metadata-forms --no-objects-only` | 776 leaves — `objects` 2, `apps` 1, **`metadataForms` 773** | 3 leaves — `objects` 2, `apps` 1 | not written (correct, both before and after) |
| `--no-objects-only` (baseline ON) | 776 leaves, same three groups | 3 leaves | 773 leaves — before the fix the SAME 773 keys were in both files |
| `--no-metadata-forms` (the card's control, default `--objects-only`) | 2 leaves, `objects` only | 2 leaves, unchanged | not written |
| default | 2 leaves, `objects` only | 2 leaves, unchanged | 773 leaves, unchanged |

The reported flag interaction reproduces in exactly the direction the card describes: the flag works alone and stops working the moment `--no-objects-only` joins it. The control confirms the card's other half — dropping `--no-objects-only` removes the block and takes the `apps` key with it. The emitted file went from 1716 lines to 33, and the 773 inlined leaves were **English**: the default locale is filled from the source labels and the registry authors them in English, so a `zh-CN` default locale shipped the platform's English Studio strings inside its own bundle. (The card counted 761 on 17.2.0; the baseline is registry-driven and has grown since.)

## Driven — the `--json` face, in both flag states, across three code states

The blocking item of the first review round. `--json --no-objects-only`, with `--metadata-forms` ON and with `--no-metadata-forms`, on each code state in turn — files restored from `HEAD` between states, hashes checked:

| code state | flag ON | flag OFF | does the flag control anything? |
| --- | --- | --- | --- |
| `2648774b967` (base, pre-PR) | `bundles` = `objects`, `apps`, `metadataForms` — 776 leaves | identical | **no** — the flag was ignored here already |
| `5da3bfea1e6` (commit 1) | `bundles` = `objects`, `apps` — 3 leaves, no baseline anywhere | identical | **no** — still ignored, now in the opposite direction |
| `ac8add2e4f4` (commit 2) | `bundles` 3 leaves + `metadataForms` `zh-CN` 773 leaves | `bundles` 3 leaves + `metadataForms` `{}` | **yes** |

In the first two rows the payloads were equal in every field but `duration`, which is wall clock. Commit 1 did not merely fail to fix this face — it stopped the fold here too, and this face has no second file for the baseline to move to, so the baseline was dropped outright while `metadataFormsCounts` went on reporting 773.

**The fix:** the payload mirrors the FILE SET. `bundles` is the stack module, and a new `metadataForms` map is the companion — keyed by the locales whose companion would be written, gated by `emitsMetadataForms(locale)`, the SAME predicate and deliberately not a second one. The map is always emitted; an empty one reads as "no baseline in this run", where a missing key would be indistinguishable from an older CLI.

**Three sentences this body and the changeset made, corrected in commit 2:** "`--metadata-forms` is its only control" was false on the `--json` face; "`--json` carries the same payload the files carry" was false in exactly that way; and the changeset's "Nothing published loses content" was false — commit 1 removed the baseline from the JSON payload with nowhere for it to go.

## The two flags, and why no precedence was invented

- `--objects-only` picks the STACK MODULE's sub-tree — `objects` alone, or everything the stack authors. Its help says "disable to include apps/dashboards"; it never promised the baseline.
- `--metadata-forms` decides whether the registry-driven baseline is emitted AT ALL — into its own companion file, or its own JSON key. It is the only control over the baseline on both faces.

These do not overlap and never needed a winner. The overlap was in the emitter: `kind: 'full'` gave the baseline a SECOND home, and the two homes then disagreed about which flag governed it.

⚠️ Retracted in commit 2: the three kinds are **not** three disjoint cells. `'objects'` is a SUB-SELECTION of `'stack'`. The invariant that holds is about the PAIR a run emits — the module it writes and the companion beside it are disjoint, and under `'stack'` the two together are everything the extractor built: 3 + 773 = 776, the extractor's own count.

## Test-fixture rounds (commits 3 and 4) — head `af9fd95004c`

`Lint & Repo Gates` was red on this PR's own head for a reason that is not contract content: the new e2e created its fixture root **inside the tree** at `packages/cli/test/.tmp-i18n-14894`. Nothing tracked ignores `.tmp-*` (the rules are `*.tmp` and `tmp/`), so `dispatch-gates --self-test` failed its "every in-tree directory this tree's sources create is covered by a tracked ignore rule, or is tracked itself" case, naming this file. ⛔ The repair is never a bespoke ignore rule for one test.

**Commit 3** moved both roots to the system temp dir. That worked for the gate but cost something, and the cost is why commit 4 exists: `bundle-require` writes its bundled module NEXT TO the config and Node resolves the config's bare specifiers from THAT directory, so a system-tmpdir project cannot see `@objectstack/spec` (driven: `Cannot find package '@objectstack/spec' imported from /tmp/…/stack.config.bundled_….mjs`). Commit 3 dropped the fixture's `defineStack` import to get around that — which is not placement, it is removing load-time validation.

**Commit 4** restores the import and keeps the root out of the un-ignored location: the fixture goes under this package's git-ignored `tmp/`, and only the `--out` root stays in the system temp dir, because only the fixture has to resolve anything. `serve-no-artifact.e2e.test.ts` carries the same constraint and the same answer, in a comment that says so; four other suites share the root.

Both premises were checked rather than inherited:

- **Ignore coverage, measured against the gate's instrument and not against the siblings' existence:** `git check-ignore -v packages/cli/tmp/probe.txt` answers `.gitignore:55:tmp/`, and `.gitignore` is tracked. No rule is added.
- **Coexistence:** five suites share that root and each `rmSync`s only its own `mkdtemp` subdirectory, never the root. This suite's `afterAll` does the same and says so. After a run, `packages/cli/tmp` is left empty — 0 entries.

The alternative (a `node_modules` symlink into the system temp dir) was **also driven and is safe** on node v22.22.2: `rmSync` recursive deleted the link and left the target and its canary file intact. It is not used, because the repo already answers this exact constraint more plainly.

Self-test, printed verdict line, same command and tree — before commit 3, after commit 3, after commit 4:

```
before:  ✗ … — EXPOSED: packages/cli/test/.tmp-i18n-14894 (…e2e.test.ts:83)
         ✗ dispatch-gates self-test: 1 of 1511 case(s) failed.
after 3: ✓ dispatch-gates self-test: 1511 cases pass.
after 4: ✓ dispatch-gates self-test: 1511 cases pass.   (in-tree case ✓, 0 EXPOSED, exit 0)
```

Readings are unchanged with `defineStack` restored — 3 stack leaves, 773 baseline leaves, none under `--no-metadata-forms` — and the validation is measurably back: dropping `type` from the field is **refused again**, exit 1, `defineStack validation failed (1 issue): objects.0.fields.name.type: Invalid field type ''`, where commit 3's plain-object fixture accepted the same config at exit 0.

## What moved in this repository: nothing

All nine `i18n-extract.config.ts` docstrings run under the default `--objects-only` (flags read out of each via `flagsFromDocstring`), and that path's module, export name and type signature are byte-for-byte unchanged. `pnpm check:i18n` exit 0 on the committed tree.

## Tests

- **Command-level pin** (`i18n-extract-metadata-forms-flag.e2e.test.ts`) spawns the real CLI and takes a group census of the bytes it wrote, in all four file-face flag combinations **and `--json` in BOTH flag states**. The `--json` case originally drove `--no-metadata-forms` only, which is why it was green while the flag did nothing on that face: a pin that exercises one state of a flag can never detect that the flag is ignored. It now asserts the axis — the payloads must differ once `duration` is dropped, flag-ON must carry the baseline under its own key with a leaf count equal to `metadataFormsCounts`, flag-OFF must carry none.
- **Unit pins** on the renderer's kinds and on `stackAuthoredSubtree`, asserted as a census — a duplicate group is invisible to a substring assertion, and a duplicate was half of what this closes.
- **Ablation of the `--json` fix, both directions**, mutation proven on disk each time (injected text counted, predicate count checked to 0, blob hash moved, restored from `HEAD` and re-hashed): `.filter(() => false)` reddens the pin `1 failed | 3 passed`; `.filter(() => true)` also reddens it `1 failed | 3 passed`. The three file-face cases stay green in both legs. That the ONE-state version could not have caught this is a reading, not an argument: it was green 4/4 at `5da3bfea1e6`, on code the table above shows was ignoring the flag.
- **Ablation of the file-face fix** (commit 1): blob `6899355f` to `4f9230d7`, unit `2 failed | 18 passed`, e2e `2 failed | 2 passed`, restored, `git diff HEAD` empty.
- No rebuild leg in any ablation, a property of the subject rather than an omission: both suites resolve it from `packages/cli/src` — vitest imports the source, `bin/run-dev.js` runs it through tsx — not through a package's `exports` into `dist`.
- **Blast radius:** the 20 test files under `packages/cli/test` that reach the changed modules, derived by grep — 20 files / 301 tests green at `ac8add2e4f4`.

## Gates

At `ac8add2e4f4`: union derived with `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack --commands`, asserted against its own reconciliation line — **56 families, 56 commands printed, 56 run, all exit 0**, exit codes captured before any pipe. Artifact rosters run separately: 37 families, 34 exit 0, three NOT MEASURED and none a pass.

⚠️ **That union was 56/56 green while CI was red**, and it is worth stating rather than burying: the failing self-test case is **outside the derived union**. A clean union is evidence about the union, not about the job. The verdict that mattered came from the failing job's own log.

At `af9fd95004c`, for the one changed file: `dispatch-gates --self-test` 1511/1511 with the in-tree case ✓ and exit captured by redirect; the e2e pin 4/4; `pnpm --filter @objectstack/cli typecheck` exit 0 with the test layer included; ESLint 0 errors / 0 warnings.

Docs drift check named two hand-written pages. Both read, neither falsified, nothing edited: `i18n-standard.mdx` line 811 describes `--json` as printing the skeletons instead of writing files, which this PR makes more true, and line 873's "`metadataForms` are not checked" is about `os i18n check`'s coverage rules.

## Files changed

```
.changeset/i18n-extract-metadata-forms-flag-independence.md
packages/cli/src/commands/i18n/extract.ts
packages/cli/src/utils/i18n-extract.ts
packages/cli/test/i18n-extract-metadata-forms-flag.e2e.test.ts
packages/cli/test/i18n-extract.test.ts
```

Commits 3 and 4 touch exactly one of them: `packages/cli/test/i18n-extract-metadata-forms-flag.e2e.test.ts`. No reviewed source moved after commit 2.

## Changeset

`@objectstack/cli: patch`, no `**BREAKING**` banner. Calibrated against this repo's own precedent for the same surface: `.changeset/i18n-walk-one-key-one-demand.md` changes which keys `os i18n extract` emits and takes `patch`. `'full'` to `'stack'` is internal — `renderTranslationModule` is not on the package's `exports`.

## Clause ②

Clause-②: yes

- **Mechanical limb — yes, on concrete keys.** For the `--no-objects-only` class the `--json` `bundles` value LOSES its `metadataForms` key (`objects, apps, metadataForms` at 776 leaves, to `objects, apps` at 3), the payload GAINS a top-level `metadataForms` key, and the emitted module's type narrows to an `Omit` of `TranslationData` without `metadataForms`. Probed with tsc: a consumer typed as `TranslationData` still compiles (exit 0), one reading `.metadataForms` off the module gets TS2339.
- **Conformance limb — yes, and on `--json` the class reaches a THIRD verdict.** On the file face `--no-objects-only` moves between two already-published verdicts. On `--json` the same class now also reaches a verdict no face had before: with `--no-metadata-forms`, the baseline is **nowhere in the payload**. That face never honoured the flag at all until this PR.

The `needs:contract-review` label is not applied or altered by this seat.

## Note for whoever merges

**Squash message.** Commit 1 carries `Fixes #14894`; commits 2, 3 and 4 carry `Part of #14894`. Concatenated by a default squash they contradict each other, and this branch's convention is that card relations are declared once — here, in this body's first line. None of the four trailers should have been written; removing them needs a force-push, which is barred on this lane. Take the squash body from this PR body, not from the concatenated commit messages.

Scope note: #14895 is not addressed here — same reporter and command family, different defect (the remedy line `--check` prints).


os-litant (commented):
## Contract review (clause ②) — PR #16120 / card #14894, at head `5da3bfea1e6`

**Clause ② — mechanical limb: YES**, and not on the unclear-call rule the declaration leans on: a concrete key is REMOVED from a published payload. For the `--no-objects-only` input class the `--json` payload's `bundles.<locale>` loses its `metadataForms` key (measured: merge-base `["objects","apps","metadataForms"]`, 776 string leaves → PR head `["objects","apps"]`, 3 leaves), and the emitted `<locale>.objects.generated.ts` loses the same group with its export type narrowed to `Omit<TranslationData, 'metadataForms'>` — a consumer that reads `.metadataForms` off that module stops compiling (tsc probe: TS2339), while a consumer typed as `TranslationData` still compiles (same probe, assignment line clean).

**Clause ② — conformance limb: YES**, as declared: on the file face the `--no-objects-only` class moves from "baseline inlined into the objects module (and, with the flag on, duplicated in the companion)" to "baseline in the companion only" — measured, both directions, below. One correction to the declaration: on the `--json` face the same class moves to a THIRD verdict, "baseline nowhere", which no published face had for that class before this diff. That is the blocking item.

**Verdict: CHANGES REQUESTED** — one blocking item, reproduced below. Submitted as COMMENT because GitHub refuses APPROVE / REQUEST_CHANGES on a same-account PR; the verdict line above is the verdict.

`Implemented-by:` `claude/issue-14894-i18n-extract-no-metadata-forms` (mode:subagent dev, per its claim comment)
`Reviewed-by:` session `01D47qPfEWVPmhguWgBZCi5N` — a context-isolated contract-review subagent of the same dispatching session, fed the card, the PR and the diff only. Under `references/contract-review.md` the dispatching seat must weigh that identity pair itself; this is a measured second opinion, not a self-issued clear.

---

### Blocking item 1 — the `--json` face drops the metadata-forms leaf class (silent data loss on a documented face)

**Reproduction** (the PR's own e2e fixture: one object, one app, `i18n.defaultLocale: 'zh-CN'`; leaves counted structurally by importing each emitted module / parsing the payload and walking string leaves — never the printed count, which is #16121):

```
tsx packages/cli/bin/run-dev.js i18n extract stack.config.ts --locales=zh-CN --json --no-objects-only
```

| tree | `bundles["zh-CN"]` keys | leaves | `metadataFormsCounts["zh-CN"]` |
| --- | --- | --- | --- |
| merge-base `2648774b967` (sources swapped in by blob, `420ac867dfb` / `28a77c4739c`, verified) | `objects, apps, metadataForms` | **776** (2 + 1 + 773) | 773 |
| PR head `5da3bfea1e6` (blobs `6899355f09b` / `e1802bb2788`, restored and `git diff HEAD` empty) | `objects, apps` | **3** | 773 |
| PR head, same run **plus `--no-metadata-forms`** | `objects, apps` | 3 | 773 |

The last two rows are byte-identical in `bundles`: on this face `--metadata-forms` now controls nothing in either position. The files for the same flags carry 776 leaves across two modules (3 in the stack module, 773 in the companion); the JSON carries 3, and the payload's own `metadataFormsCounts` reports 773 keys that are in it nowhere.

Why this blocks rather than trails:

1. **It is a documented face.** `content/docs/protocol/kernel/i18n-standard.mdx` § CLI Tools: "Print the skeletons as JSON instead of writing files" (`os i18n extract --json`), and the flag's own help says the same. Before this diff, `--json --no-objects-only` was the one route by which the baseline skeleton reached the JSON face; after it, there is none.
2. **It falsifies the PR's own invariants.** "`--metadata-forms` is the only control over the baseline" — on this face it is no control at all. "`--json` carries the same payload the files carry" — 3 vs 776. The changeset's "Nothing published loses content" — 773 leaves are lost with no replacement in the payload.
3. **It is exactly where the partition claim stops holding.** The three kinds partition a locale's leaves only where each kind has a home. On the file face they do (verified below). On the `--json` face only the `'objects'`/`'stack'` cell is emitted; the `'metadataForms'` cell has no home, and this diff removed the only one it had.
4. **The pin cannot see it.** The `--json` case in `i18n-extract-metadata-forms-flag.e2e.test.ts` is driven with `--no-metadata-forms` only, where omission is correct; its comment says "the sub-tree the emitter refuses to write must not arrive here" — but with the flag ON the emitter does write that sub-tree (to the companion), and the pin never drives that case.

**What closes it:** give the third kind a JSON home gated by the SAME `emitsMetadataForms(locale)` predicate that gates the companion file — a sibling key beside `bundles` mirroring the companion, or `bundles.<locale>.metadataForms` restored under the flag; the shape is yours/the maintainer's, the invariant is the PR's own ("the same payload the files carry"). Then extend the `--json` pin to the flag-ON case (baseline present, `> 100` leaves, same instrument as the file pin) beside the existing flag-OFF case, and correct the three sentences in item 2 (PR body, changeset, and the `--json` comment in `extract.ts`). Note the default `--objects-only` never carried the companion on `--json` either — pre-existing and outside this card, but the same predicate covers it for free.

---

### The partition property — verified, with its boundary stated

- **Static.** `stackAuthoredSubtree` is a rest-spread minus `metadataForms`, so every other group the extractor `setDeep`s (`objects`, `apps`, `pages`, `dashboards`, `flows`, `datasets`, …) stays in kind `'stack'`; kind `'metadataForms'` renders exactly the complement; kind `'objects'` renders `data.objects` alone — a proper subset of `'stack'`, which is the documented meaning of `--objects-only` (it drops `apps`/`pages`/… by design; pre-existing, #3762). So `{stack, metadataForms}` is a partition of one locale's generated leaves and `'objects'` is a sub-selection of one cell; the three are not three disjoint cells and the docblock should not imply they are.
- **Dynamic, file face, `--no-objects-only` (flag ON):** stack module 3 leaves (`objects` 2, `apps` 1) + companion 773 = **776 = the extractor's own `counts["zh-CN"]`**; groups disjoint; no leaf dropped, none twice. Merge-base: 776 + 773 with the same 773 in both files. Broken shape this control would show: a total under 776 (a dropped class) or over (a duplicate).
- **`--json` face:** one cell only — blocking item 1.

### Reproduction direction — both driven

File face, all four combinations before and after (leaves: objects module / companion): A `--no-metadata-forms --no-objects-only` 776/– → 3/–; B `--no-objects-only` 776/773 → 3/773; C `--no-metadata-forms` 2/– → 2/– (byte-identical); D default 2/773 → 2/773 (byte-identical). `--no-objects-only`'s own meaning (`apps` in the module) holds in every cell before and after, so the one-directional claim holds on the file face: `--no-metadata-forms` is the flag that broke when the other joined it, and the reverse combination shows the duplication half, not a broken `--no-objects-only`. On the `--json` face the interaction was symmetric all along — `--no-metadata-forms` was ignored there before (both positions gave 776) and is ignored after (both give 3).

### "Nothing in this repository moved" — holds

Flags read the way the gate reads them (`flagsFromDocstring`): eight docstrings spell `--objects-only --no-metadata-forms`; `platform-objects` spells neither and takes both defaults, so its objects module and its companion are on unchanged paths. `pnpm check:i18n` exit 0 on the PR head (broken shape: any `out of date:` line across the nine sets).

### Changeset level — `patch`, no banner: acceptable once item 1 is closed; its text is wrong as delivered

- `'full'` → `'stack'` is internal: `renderTranslationModule` is not reachable from `@objectstack/cli`'s `exports` (`.`, `./console`, `./hook-body`; `src/index.ts` does not re-export `utils/i18n-extract`), and no caller outside the command exists in the repo.
- File face: export name kept, type narrows to `Omit<…>` that still assigns to `TranslationData` (probe above) — `patch` matches the cited precedent (`.changeset/i18n-walk-one-key-one-demand.md`).
- JSON face as delivered: a removed key with no replacement — a narrowing, and "Nothing published loses content" is false. With a JSON home added it becomes a relocation like the file face and `patch` stands; either way the sentence must go. `check-changeset-no-major` and `check-adr-0087-registration` exit 0 as declared (no banner, so no disposition marker is owed); if the maintainer grades the JSON removal breaking instead, the launch-window carrier is a `**BREAKING**` banner plus one `<!-- adr-0087: … -->` marker from the `CATEGORIES` const — `no-migration-prescription` is the fitting category, since nothing authorable is retired.

### The blast-radius narrowing — sound

Re-derived independently (grep over `packages/cli/test` for `i18n-extract|renderTranslationModule|commands/i18n/extract|'i18n', 'extract'|i18n extract|stackAuthoredSubtree|narrowToCommittedSections`) → the same 20 files. Indirect reach checked rather than assumed: the changed functions have exactly one caller (the extract command); `i18n-coverage.ts` imports only the unchanged `collectExpectedEntries`; no other `kind: 'full'` / `objectsOnly: false` caller of the renderer exists in the repo (the one in `i18n-extract-emitted-files.test.ts` calls the test's own mirror). The command-spawning gates (`check:i18n`, `check:i18n-coverage`, `check:i18n-walk-parity`, `check:i18n-stale-fill`) exercise it end to end. Ran: **20 files / 301 tests pass** (178 s). `pnpm --filter @objectstack/cli typecheck` (test layer included) exit 0; ESLint on the four changed TS files: 0 errors, 0 warnings.

### Gates

Union derived with `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack --commands` and asserted against its own count line: **56 command(s) — 27 pnpm, 29 direct node; 56 run, 56 exit 0** on the PR head, exit codes captured before any pipe, after a full workspace build (so `check:i18n-coverage` and `check:dual-build-cjs-loads` went straight to a real verdict). The derivation flagged the tree 3 commits behind `origin/main` with `scripts/check-published-files.mjs` changed in that range, so that family was re-run on GitHub's merge commit `507920a6a44` (merge diff byte-identical to the PR diff; none of the three `main` commits touches `packages/cli` or the extractor): exit 0. The **Artifact rosters** block was run separately: 37 families, 34 exit 0, three not a pass (listed below).

- Full 136-file `@objectstack/cli` suite — NOT MEASURED; verdict: the 20-file narrowing is sound (derivation above), so no verdict is withheld on it; CI runs the farm.
- `node scripts/check-partof-closing-keyword.mjs` — exit 2 NOT WIRED (no PR context locally); verdict: none; `pnpm check:partof-closing-keyword` exit 0 is `--self-test` only (#16030) and grades the checker, not this PR.
- `node scripts/check-single-claim-paths.mjs` — exit 2 NOT WIRED; verdict: none; same `--self-test`-only caveat for its pnpm spelling.
- `pnpm --filter @objectstack/spec run check:react-declaration-parity` — exit 1 "this gate did NOT run" (needs a browser-produced `sdui.manifest.json`); verdict: none; the diff touches no React blocks.
- The six workflow-valued families (`check-cross-package-test-inputs --union-into …`, `check-shard-attestation` ×3, `check-test-completeness` ×2) — NOT MEASURED; verdict: none; they take a value that exists only in a CI run.
- The always-runs tail beyond ESLint on the four changed files — NOT MEASURED; verdict: none.

### Non-blocking observations

- `Wrote … (776 keys)` is printed for a file holding 3 leaves — #16121, pre-existing, unchanged here; it is why every number above was counted off the bytes.
- Unmeasured, out of this card, for filing: under `--no-objects-only --source-hashes`, `committedSourceHashes` narrows the provenance table to the `objects` and `metadataForms` sections only, so `apps.*` (and other stack-authored) leaves that ARE committed in the stack module get no provenance record. Pre-dates this diff (#12559's narrowing); worth a card now that the stack module is the declared home of those leaves.

---

os-litant (commented):
## Contract review (clause ②) — delta re-review, PR #16120 / card #14894, at head `ac8add2e4f4`

**Clause ② — mechanical limb: YES, on concrete keys.** The `--json` payload GAINS a new top-level `metadataForms` key — a new key on a published payload, the mechanical floor's always-yes trigger — shaped `Record<locale, TranslationData['metadataForms']>`, present in every run (an empty map when no companion is written), keyed only by the locales whose `<locale>.metadata-forms.generated.ts` would be written; for the `--no-objects-only` class `bundles.<locale>` still loses its `metadataForms` sub-key (content relocated to the new key, measured 3 + 773 = 776), and the emitted module's export type still narrows to `Omit<TranslationData, 'metadataForms'>`.

**Clause ② — conformance limb: YES.** On the file face the `--no-objects-only` class moves from "baseline inlined (and duplicated)" to "baseline in the companion only", unchanged since the first commit and re-measured on this head. On the `--json` face the class now reaches two flag-dependent verdicts — baseline under its own key with `--metadata-forms`, absent under `--no-metadata-forms` — where every prior code state gave one flag-independent answer.

**Verdict: CHANGES REQUESTED — the contract content is CLEARED (my blocking item is closed and verified below), but this head is NOT LANDABLE: CI's `Lint & Repo Gates` is red on the PR's own e2e fixture placement (blocking item 2).**

`Implemented-by:` `claude/issue-14894-i18n-extract-no-metadata-forms` (mode:subagent dev, per its claim comment)
`Reviewed-by:` session `01D47qPfEWVPmhguWgBZCi5N` — a context-isolated contract-review subagent of the same dispatching session, fed the card, the PR and the diff only. As in round one: the dispatching seat weighs that identity pair itself; this is a measured second opinion, not a self-issued clear.

---

### Blocking item 1 (round one) — CLOSED, verified

**The gating predicate is the same one, not a copy.** `emitsMetadataForms` is defined once (`extract.ts:212`, `flags['metadata-forms'] && metadataFormsCounts[locale] > 0`) and is the predicate at all four sites — the source-hash section list (`:257`), the new JSON key (`:299`, `.filter((l) => emitsMetadataForms(l))`), the dry-run print (`:334`) and the companion file (`:361`). `git grep "metadata-forms']"` over `packages/cli/src` at the head hits that one definition and nothing else, so no second predicate exists to drift.

**The three-state table HELD**, on my own runs of `--json --no-objects-only` in both flag states (fixture: one object, one app, `defaultLocale: 'zh-CN'`; leaves counted structurally off the payload; the base and first-commit rows from the blob-swapped sources of round one, hashes verified):

| code state | flag ON | flag OFF | equal once `duration` is dropped? |
| --- | --- | --- | --- |
| base `2648774b967` | `bundles` = `objects, apps, metadataForms`, 776 leaves; no top-level `metadataForms` key | same | **yes** — flag ignored |
| first commit `5da3bfea1e6` | `bundles` = `objects, apps`, 3 leaves; no top-level key | same | **yes** — flag ignored, opposite direction |
| head `ac8add2e4f4` | `bundles` 3 leaves + `metadataForms: { "zh-CN": … }` 773 leaves | `bundles` 3 leaves + `metadataForms: {}` | **no** — the flag moves the payload |

Qualifying my round-one "byte-identical": in the first two rows the raw bytes differ, and `duration` (wall clock) is the only differing field — equality holds exactly once it is dropped. The reading that the first commit swapped one wrong answer for another, rather than breaking a working face, is correct: the base row's flag-OFF payload is 776 leaves including the baseline the operator switched off.

**The extended pin's ablation — confirmed, and it fails on the flag axis.** Both legs on the committed head, mutation proven on disk each time (injected text ×1, original `.filter((l) => emitsMetadataForms(l))` ×0, blob `5c5e3c50c0a` moved), restored from `HEAD` and re-hashed to `5c5e3c50c0a` with `git diff HEAD` empty:

- `.filter(() => false)` (baseline never in JSON — the first commit's behaviour): `1 failed | 3 passed`; the red is the `--json` case at `expect(Object.keys(on.metadataForms)).toEqual(['zh-CN'])` — received `[]` — the flag-ON assertion.
- `.filter(() => true)` (predicate ignored — baseline in both states): `1 failed | 3 passed`; the red is the `--json` case at `expect(off.metadataForms['zh-CN']).toBeUndefined()` — received the 27-group map — the flag-OFF assertion.

The three file-face cases stayed green in both legs (selectivity), and the first assertion to fail in each leg is the one about the flag state that leg breaks — not `duration`, not the `bundles` shape, not the count.

**The three falsified sentences are corrected**, in both carriers. Changeset: "the only control over it **on both faces**"; "`--json` … mirrors that file set … That map is new"; "Nothing published loses content" replaced by a per-face **What content moves where** paragraph that states the JSON relocation. PR body: the same three, in their own section, plus the file-face table unchanged and the three-state table above. The over-claim I flagged in round one — "the three kinds are a partition" — is also withdrawn in the renderer docblock, the changeset and the test names (`sub-tree selection`, `emits every leaf exactly once across the pair a stack run writes`), in favour of the pair invariant that actually holds.

### Blocking item 2 (new) — CI is red on this head, on the PR's own test placement

`Lint & Repo Gates` (job `101422068406`) concluded **failure** at `ac8add2e4f4`. Reproduced locally with CI's own spelling on this head:

```
$ pnpm check:pm-dispatch-gates      # exit 1
✗ every in-tree directory this tree's sources create is covered by a tracked ignore rule, or is tracked itself
  — EXPOSED: packages/cli/test/.tmp-i18n-14894 (packages/cli/test/i18n-extract-metadata-forms-flag.e2e.test.ts:83)
✗ dispatch-gates self-test: 1 of 1511 case(s) failed.
```

`FIXTURE_DIR = join(HERE, '.tmp-i18n-14894')` (line 66) is created in-tree at line 83; the tracked `.gitignore` covers `*.tmp` and `tmp/`, not a dot-prefixed `.tmp-*` directory. The Director's comment on this PR asked for exactly this patch at `5da3bfea`; the second commit did not carry it, and the case is not in the derived gate union (it is a whole-tree self-test step of `lint.yml`), which is why "56/56 exit 0" is true and the job is still red. Fix as the Director prescribed: keep the fixture under `packages/cli` (the test's own comment explains why — `bundle-require` resolves `@objectstack/spec` from the config's directory) but under a covered root, e.g. `packages/cli/tmp/…` (`tmp/` is tracked-ignored) — ⛔ not a bespoke ignore rule for one test. A placement-only push moves no reviewed source; this clear extends on it once `pnpm check:pm-dispatch-gates` is green.

### Changeset level, re-judged against the new key — `patch` stands

The new top-level key is additive: a `--json` consumer that ignores it is unaffected. The only field removed for any input class (`bundles.<locale>.metadataForms`, `--no-objects-only`) is relocated to the new key under the same flag, so content moves and nothing is lost — the same shape as the file face, and the cited precedent (`.changeset/i18n-walk-one-key-one-demand.md`, `patch`) covers a changed emitted-key set. The key exists to make a flag do what its `--help` says, which is a bug fix, and a bug fix in a released package takes `patch` here. `'full'` → `'stack'` remains internal. If the maintainer reads a new payload field as a feature, `minor` with no banner is the alternative; no `**BREAKING**` banner and no ADR-0087 marker is owed either way — nothing authorable is retired and no consumer's compile breaks except one reading `.metadataForms` off the `--no-objects-only` module, which round one already graded. `check-empty-changeset`, `check-changeset-no-major` and `check-adr-0087-registration` exit 0 on this head; CI's `Check Changeset` is green.

### The delta did not disturb what round one cleared

File face on this head: A `--no-metadata-forms --no-objects-only` 3 leaves; B `--no-objects-only` 3 + companion 773; C `--no-metadata-forms` 2; D default 2 + 773 — identical to the first commit, so the reproduction direction, the pair invariant (3 + 773 = 776 = the extractor's own count) and "nothing in this repository moved" (`pnpm check:i18n` exit 0) all stand. Blast radius re-run on the head: 20 files / 301 tests pass; `pnpm --filter @objectstack/cli typecheck` exit 0 (test layer included); ESLint on the four changed TS files 0 errors / 0 warnings.

### Gates, on this head

Union re-derived (`node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack --commands`) — the same 56 families as round one, asserted against its own count line: **56 run, 56 exit 0** (29 direct node, 27 pnpm, exit codes captured before any pipe, workspace built first). Rosters block separately: 37 families, 34 exit 0, three not a pass (below). The derivation now flags the tree 12 commits behind `origin/main` (`658262e462b`) with four derived-from files changed there; the three runnable families among them (`check:published-files`, `docs-audit/check-affected-docs.mjs`, `docs-audit/check-drift-comment.mjs`) were re-run on a fresh local merge of current `origin/main` with this head — a clean merge whose diff is byte-identical to the PR diff, and `main`'s `packages/cli` changes touch only `explain.ts` / `generate.ts` and their pins, never the extractor — all exit 0. GitHub's own merge commit for this head (`683f6210a7f`) is likewise a clean merge. The fourth stale file is `lint.yml` itself; CI's run on this head is the reading for its always-runs tail — all green except the one red in blocking item 2, with `Test Core (1/6)` still in progress at the time of reading.

- Full 136-file `@objectstack/cli` suite — NOT MEASURED; verdict: the 20-file narrowing stands from round one; CI's six `Test Core` shards are the farm (five green, one in progress at reading).
- `node scripts/check-partof-closing-keyword.mjs` — exit 2 NOT WIRED; verdict: none; the pnpm spelling's 0 is `--self-test` only (#16030).
- `node scripts/check-single-claim-paths.mjs` — exit 2 NOT WIRED; verdict: none; same caveat.
- `pnpm --filter @objectstack/spec run check:react-declaration-parity` — exit 1 "did NOT run"; verdict: none; no React blocks in the diff.
- The six workflow-valued families (`check-cross-package-test-inputs --union-into …`, `check-shard-attestation` ×3, `check-test-completeness` ×2) — NOT MEASURED; verdict: none; CI-only values.
- The always-runs tail beyond ESLint on the four files and `pnpm check:pm-dispatch-gates` — NOT MEASURED locally; verdict: CI's run on this head stands in, green except blocking item 2.

---

PR #17884, opened 2026-09-12
fix(spec): give the email-template locale bundle a stated en-US floor, and report one that has none

### LINKED ISSUE #17614: [finding] An emailTemplates bundle tagged with the stack's own defaultLocale (e.g. `en`) has no fallback floor: the resolver retries only the literal `en-US`, so every unlisted recipient locale dead-letters permanently
## The trap

An app declares its locale vocabulary once, in `defineStack`:

```ts
i18n: { defaultLocale: 'en', supportedLocales: ['en', 'zh-CN', 'ja-JP', 'es-ES'], fallbackLocale: 'en' }
```

So when that app authors an `emailTemplates` bundle, the obvious spelling of the English row is `locale: 'en'` — the same tag the stack already calls its default. That row **validates, builds and installs clean**. There is no author-time diagnostic anywhere.

At delivery it permanently dead-letters for any recipient the bundle does not carry a row for.

## Measured, on the installed 17.4.0

`EmailService.resolveAndRenderTemplate` (`@objectstack/plugin-email`) matches `(name, locale)` **exactly**, and the single fallback rung is the **literal** string `en-US`:

```js
const preferred = input.locale && String(input.locale).trim();
const wanted = preferred || DEFAULT_TEMPLATE_LOCALE;          // DEFAULT_TEMPLATE_LOCALE = "en-US"
let row = await loader.load(input.template, wanted);
if (!row && wanted !== DEFAULT_TEMPLATE_LOCALE) {
  row = await loader.load(input.template, DEFAULT_TEMPLATE_LOCALE);
}
if (!row && !preferred) { row = await loader.load(input.template, void 0); }
if (!row) { throw new Error(`TEMPLATE_NOT_FOUND: ${input.template} (locale=${wanted})`); }
```

There is **no language-subtag folding** on this path: `en-US` does not fall back to `en`, and `en` is not reachable from `en-US`. `TEMPLATE_NOT_FOUND` classifies `permanent`, so the delivery dead-letters without retry.

`sys_user.locale` is a free-text BCP-47 column — `maxLength: 35`, pattern `^[A-Za-z]{2,8}(?:-[A-Za-z0-9]{1,8})*$`, no default, no enum, user-editable. Nothing constrains it to the app's `supportedLocales`.

Cross that with `resolveRecipientLocale`'s ruled chain (recipient's own `sys_user.locale`, then the deployment default) and a bundle whose English row is tagged `en` behaves like this:

| recipient's `sys_user.locale` | bundle tagged `en` | bundle tagged `en-US` |
|---|---|---|
| unset — deployment default `en` applies | delivers | delivers, via the `en-US` retry |
| `en-US` | **TEMPLATE_NOT_FOUND** (permanent) | delivers |
| `de-DE` — legal tag, not in `supportedLocales` | **TEMPLATE_NOT_FOUND** (permanent) | delivers |

Same metadata, same gates, same green build. The only difference is a three-character spelling the authoring surface never mentions.

## Why an author picks the wrong one

`EmailTemplateDefinitionSchema.locale` is `z.ZodDefault(z.ZodString)` — any string, defaulting to `en-US`. Its `describe` and the schema module's header say the template is resolved by `(name, locale)`; neither says that `en-US` is a **floor** with special standing, nor that an explicitly-authored `en` forfeits the fallback the default would have given.

The one place the rule is written down is a docstring two packages away, in `service-messaging`'s `recipient-locale` module, under a heading about a different subject:

> The one asymmetry: such a bundle that carries the deployment default's row but not `en-US` delivered under the old single value and fails for a recipient whose own locale is a third tag. That bundle is off the documented contract (`en-US` is the ladder's floor), so the fix is the bundle, not a third rung here.

That paragraph is correct and it is where this behaviour was reasoned about. It is not where an app author is standing when they type `locale:`.

⭐ The asymmetry is sharper than that paragraph suggests, because it is not only "a third tag": the app's **own declared default locale** is the wrong answer whenever it is not spelled `en-US`. An app that says `defaultLocale: 'en'` and authors `locale: 'en'` has done the consistent thing throughout and still shipped a bundle with no floor.

## Carrier

objectstack-ai/hotcrm#1185 — 15 `notify` nodes converted to template references, four locales each (objectstack-ai/hotcrm PR 1893). That app declares `defaultLocale: 'en'`; its bundles are tagged `en-US` **only** because this path was read end to end first. The card's own scope line, and the app's `supportedLocales`, both say `en`.

## What would close this

Either of these; whichever the maintainer prefers. ⛔ Not proposing a behaviour change without a ruling — the ladder's shape is settled by 2026-09-01 (#13881).

1. **Say it where the author is.** `EmailTemplateDefinitionSchema.locale`'s `describe` states that the resolver matches exactly, retries only `en-US`, and that a bundle without an `en-US` row has no floor. Cheapest, no behaviour change.
2. **Make it unrepresentable or diagnosable.** An author-time rule that a bundle carrying rows for a stack's `supportedLocales` but none tagged `en-US` is reported, the way other declared-but-unreachable metadata is reported today.

Language-subtag folding in the resolver is deliberately **not** proposed: it would change delivery behaviour for every existing bundle, which is a ruling and not a finding.

os-bill:
Fixes #17614

Clause-②: yes

Triage ruled BOTH remedies in, with an order (comment 5642916137): **「do both, option 1 first.」** Both land here.

## The trap, reproduced on this branch's own head before anything changed

An app declares `i18n: { defaultLocale: 'en', supportedLocales: ['en','zh-CN','ja-JP','es-ES'] }` and authors its English `emailTemplates` row as `locale: 'en'` — the consistent thing. Measured on an unmodified tree:

| probe | reading |
|---|---|
| `defineStack` on that stack | `parsed-ok=true warnings=0`, stored locales `["en","zh-CN","ja-JP","es-ES"]` |
| `sendTemplate({ locale: 'de-DE' })` | `TEMPLATE_NOT_FOUND: acme.welcome (locale=de-DE)` |
| `sendTemplate({ locale: 'en-US' })` | `TEMPLATE_NOT_FOUND: acme.welcome (locale=en-US)` |
| **lit control** — same bundle, English row tagged `en-US` | **DELIVERED**, `subject=[en-US] Hi` |

The control is aimed at the resolver ladder itself: it proves the probe can come back "delivered", so the two `TEMPLATE_NOT_FOUND` readings are readings and not a broken harness. A second lit control covers the `defineStack` half — an `objectz` typo on the same stack is refused, so `warnings=0` is a measurement rather than a dead capture.

`TEMPLATE_NOT_FOUND` classifies **permanent**, so each of those dead-letters with no retry. `sys_user.locale` is user-editable free-text BCP-47 and is not constrained to `supportedLocales`, so the locales that can reach the lookup are not the ones the author enumerated.

## Premise re-read on today's tree, not on the card

The card measured against installed 17.4.0. Re-read at this branch's base: the ladder in `packages/plugins/plugin-email/src/email-service.ts` is byte-for-byte what the card quotes — exact `(name, locale)`, one retry rung at the literal `DEFAULT_TEMPLATE_LOCALE = 'en-US'`, no folding — and `EmailTemplateDefinitionSchema.locale` was still `z.string().default('en-US').describe('BCP-47 locale (e.g. en-US, zh-CN)')`. **The premise holds; the trap is open.** Its TSDoc was also actively misleading — "the service picks the best match for the recipient's locale" reads as folding, which does not exist on that path.

## ① Say it where the author is standing

`EmailTemplateDefinitionSchema.locale`'s `describe` and TSDoc now state the exact match, the single literal `en-US` rung, the absence of language-subtag folding, that a bundle with no `en-US` row has no floor, and that the stack's own declared default locale is the wrong tag whenever it is not spelled `en-US`. The `describe` is the published authoring surface — it lands in `content/docs/references/system/email-template.mdx`, regenerated here.

New exported `EMAIL_TEMPLATE_FLOOR_LOCALE` names the tag once, in the one place that is both the schema default and the resolver's sole retry rung.

## ② The author-time diagnostic

`defineStack` now reports an `emailTemplates` bundle that carries rows for the stack's own `i18n.supportedLocales` but none tagged `en-US`. Advisory `console.warn`, warn-once per bundle, post-parse — the same seam and posture as the existing `warnUnknownAuthoringKeys` reporter.

## ⛔ The fence held

Language-subtag folding was not added, no third rung was added, and `EmailService.resolveAndRenderTemplate` is not touched — the diff contains no file under `packages/plugins/plugin-email`. The ladder's shape stays as #13881 settled it; the remedy this points at is the bundle.

## Clause-② self-check — `yes`, and the rationale is corrected

**`yes` is right, and it is a mechanical call, not a measurement.** The carrier that fires is the **new exported symbol** `EMAIL_TEMPLATE_FLOOR_LOCALE` — a new export on a published surface is always `yes`, decided by looking at the export list. It is registered at `packages/spec/api-surface/system.json` and `packages/spec/export-origins/system.json`; a changed published `describe` moves `content/docs/references/system/email-template.mdx` alongside it.

⚠️ Correcting the rationale this PR shipped with: the dispatch and this body's first version justified `yes` on the ground that remedy ② *narrows* what passes silently, and **that rule does not exist** — the lane text says the opposite, that narrowing is still a semantic surface and does not trip clause ②. The declaration was over-determined: the right answer reached through a limb that does not fire.

The accept-set measurement below stands and is **not** the basis for the `yes`; it is reported because it is true and reviewable. The final diff **moves no accept set**: same probe, same tree, before and after, `parsed-ok=true` on both sides and the returned stack carries the identical tags `["en","zh-CN","ja-JP","es-ES"]`; only the warning count goes 0 to 1. No Zod refinement added, no schema key added or removed, no closed set narrowed.

Carrier state: PR and card both carry `needs:contract-review`; `scripts/pm/check-clause2-carriers.mjs --pair 17884` exits 0, both carriers agree. Draft, not enqueued, auto-merge not armed — the seat lands it.

## Contract review — C1 resolved, option (a)

Review verdict: PASS WITH CONDITIONS (comment 5647795889). C1 said the new constant's docblock **publishes** an invariant — that `@objectstack/plugin-email`'s `DEFAULT_TEMPLATE_LOCALE` and `EMAIL_TEMPLATE_FLOOR_LOCALE` "must stay equal" — with nothing holding them equal, the same shape as this package's existing `position-delegatable-enforcer.pin.test.ts` and its own words, "prose costs nothing to add and no compiler checks it."

**Chose (a), add the pin, over (b), delete the sentence.** The sentence is *true and useful*: both constants are published (`plugin-email` re-exports `DEFAULT_TEMPLATE_LOCALE` from its `index.ts`), they are genuinely one settled value in two packages, and an author or a future round needs to know they are coupled. Deleting it would buy honesty by removing information, leaving the drift seam just as real and now undocumented. (a) keeps the claim and makes it checkable, and it costs nothing this PR was not already allowed to spend — so (b) is the retreat, not the equal.

`packages/spec/src/system/email-template-floor-locale-parity.pin.test.ts` reads `packages/plugins/plugin-email/src/email-service.ts` **as text** and asserts its `DEFAULT_TEMPLATE_LOCALE` literal equals `EMAIL_TEMPLATE_FLOOR_LOCALE`. `packages/spec` cannot import `plugin-email` (its dependencies are `zod` and `pg-connection-string`) and must not start to; the `repo` vitest project exists for exactly this, and the path is registered in `packages/spec/vitest.repo-tests.json`. A missing or renamed declaration fails loudly rather than passing vacuously.

⛔ **The fence is still intact.** The delivered diff contains no file under `packages/plugins/plugin-email`; no rung was added, nothing folds, and `EmailService.resolveAndRenderTemplate` is untouched. The pin only reads that package.

## Successor-card seed (C2) — the second silent, floorless case

`packages/spec/src/stack.zod.ts`, the `declared.length === 0` branch of `warnEmailTemplateLocaleFloor`. A stack that **does** declare `i18n.supportedLocales` but whose bundle tags **none** of them is equally floorless and equally silent — `supportedLocales: ['en']` with a lone `fr-CA` row. This PR's own test pins that green ("silent when no authored tag is one this stack claims to support"), so the gap is deliberate and recorded rather than accidental. ⛔ Not fixed here: widening the condition is a second decision about who gets warned, and it is left as the seed of a successor card together with the no-`i18n`-block case already noted below.

A third, named by the review and recorded here for the same successor: a `console.warn` at `defineStack` is heard only by whoever runs the build and reads stdout, while this card's own premise is that the trap survives because people do not read. The stronger rung that stays inside the fence is an `os lint` / `os validate` rule — a separate card, not this one.

## Tests, and the ablation that proves each one could fail

New: `packages/spec/src/stack-email-template-locale-floor.test.ts` (10 tests). Every case asserts the parse still succeeds and the stack returns unchanged, so the pins are on the advisory posture too.

Proven by on-disk mutation, each with its injected-text count verified before the run and the restore verified by `git hash-object` against the HEAD blob:

| mutation | result |
|---|---|
| MUT-1 — the diagnostic never reports | **RED**, 4 failed / 6 passed |
| MUT-2 — the `describe` loses `TEMPLATE_NOT_FOUND` | **RED**, 1 failed / 9 passed |
| MUT-3 — call the diagnostic pre-parse on raw input | **GREEN, 10 passed — reported, not hidden** |
| restored (hashes MATCH both files) | **GREEN**, 10 passed |

MUT-3 falsified a claim I had written into the test: the reader mirrors the schema default for a missing `locale`, so pre-parse and post-parse agree and the call site is not what that case pins. The comment and the function's rationale were corrected to say what was actually measured, rather than the reverse.

### C1 parity pin — ablated, because an un-ablated pin is not a pin

Mutated `plugin-email`'s literal on disk from `en-US` to `en-GB` (restored immediately; the delivered diff does not touch that file):

| leg | evidence |
|---|---|
| landed on disk | original literal occurrences 1 to 0, injected literal 1, and the blob hash moved off the HEAD blob (`0198f7ac` to `108eb1b8`) |
| RED | exit 1 — `AssertionError: expected 'en-GB' to be 'en-US'`, 1 failed / 2 passed |
| restored + GREEN | hash back to `0198f7ac`, MATCHES the HEAD blob, `git diff HEAD` empty; re-run exit 0, 3 passed |

`pnpm check:cross-package-test-inputs` exits 0 with the new escape declared. That zero is a reading, not a dead gate: unregistering the path from `vitest.repo-tests.json` reds it (exit 1) naming this exact file, and it was then re-registered.

### Commands and exit codes

Both `packages/spec` vitest projects were run and their exit codes captured **separately** (the wrapper's own verdict line covers only the last part, so each was captured inside the run):

| command | reading |
|---|---|
| `pnpm --filter @objectstack/spec test` (project `local`) | `LOCAL_EXIT=0` — 473 files / 13444 tests passed |
| `pnpm --filter @objectstack/spec test:repo` (project `repo`) | `REPO_EXIT=0` — 30 files / 520 tests passed |
| `pnpm --filter @objectstack/spec typecheck` | `TC_EXIT=0` |
| `pnpm --filter @objectstack/spec check:generated` | all 15 artifacts up to date after `--fix` |
| `eslint . --no-inline-config --format json` | exit 0 — 6660 files (eslint's own population), 0 errors, 0 warnings, at `d81bcb4ae2` |

24 of the 107 gate families `scripts/pm/dispatch-gates.mjs` derives for this change set were run locally, all exit 0; the remaining 83 are declared to CI rather than run on the shared box. This is a **declared** narrowing, not a skipped one.

## Changeset

`.changeset/email-template-locale-floor.md`, `minor` on `@objectstack/spec`. One is owed and is not skippable: ① changes a published `describe` (it moves `content/docs/references/`) and the diff adds a published export.

## Corpus census

No in-repo stack begins warning. The one example app that declares both halves — `examples/app-showcase` (`defaultLocale: 'en'`, `supportedLocales: ['en','zh-CN']`) — already tags its bundle `en-US`, and its own header comment explains why. That is the "one careful reading" triage named as the only thing standing between this trap and a live outage; it is now the schema's job instead.

## Acceptance notes (out of scope, noted, not filed)

- `@objectstack/plugin-email` spells the floor as its own `DEFAULT_TEMPLATE_LOCALE`; this PR adds a second spelling in spec rather than making the two one constant, because collapsing them means editing the ladder's own file, which the fence puts out of reach for this round. The new constant's docblock states the dependency direction. No parity pin is added here. Successor: whoever next opens `email-service.ts` under a ruling that lifts the fence.
- The diagnostic's condition is the ruling's wording verbatim — it requires `i18n.supportedLocales` to be declared. A stack that declares no `i18n` block and authors a lone `en` row is equally floorless and stays silent. Named as a deliberate boundary, not widened here.

---

### DIFF
--- a/.changeset/email-template-locale-floor.md
+++ b/.changeset/email-template-locale-floor.md
@@ -1,0 +1,24 @@
+---
+'@objectstack/spec': minor
+---
+
+Email templates: say where the `en-US` fallback floor is, and report a bundle that has none.
+
+`IEmailService.sendTemplate` matches `(name, locale)` exactly and retries exactly one rung —
+the literal `en-US`. There is no language-subtag folding, so a bundle whose English row is
+tagged `en` is unreachable from `en-US` and from every other tag it does not itself carry;
+each such delivery raises `TEMPLATE_NOT_FOUND`, which classifies permanent, so it dead-letters
+with no retry. An app declaring `i18n.defaultLocale: 'en'` and authoring `locale: 'en'` has
+done the consistent thing throughout and still shipped a bundle with no floor — and it
+validated, built and installed clean.
+
+- `EmailTemplateDefinitionSchema.locale`'s `describe` and TSDoc now state the exact match, the
+  single literal `en-US` rung, the absence of folding, and that the stack's own declared default
+  locale is the wrong tag whenever it is not spelled `en-US`.
+- New exported `EMAIL_TEMPLATE_FLOOR_LOCALE` names that tag once: it is both the schema default
+  and the resolver's sole retry rung.
+- `defineStack` now reports (advisory `console.warn`, warn-once per bundle) an `emailTemplates`
+  bundle that carries rows for the stack's own `i18n.supportedLocales` but none tagged `en-US`.
+
+Advisory only — no accept set moves. The stack still parses and is returned unchanged; the
+resolver's ladder is unchanged.
--- a/content/docs/references/system/email-template.mdx
+++ b/content/docs/references/system/email-template.mdx
@@ -44,7 +44,7 @@ const result = EmailTemplateDefinitionSchema.parse(data);
 | **name** | `string` | ✅ | Template identifier (dotted snake_case) |
 | **label** | `string` | ✅ | Display label |
 | **category** | `Enum<'auth' \| 'notification' \| 'workflow' \| 'marketing' \| 'custom'>` | optional (default: `"custom"`) |  |
-| **locale** | `string` | optional (default: `"en-US"`) | BCP-47 locale (e.g. en-US, zh-CN) |
+| **locale** | `string` | optional (default: `"en-US"`) | BCP-47 locale (e.g. en-US, zh-CN) — the bundle key the resolver matches EXACTLY, with one retry rung: the literal `en-US`. No language-subtag folding, so `en` and `en-US` are different bundles and neither reaches the other. A bundle with no `en-US` row therefore has no fallback floor: any recipient locale it does not carry a row for raises TEMPLATE_NOT_FOUND, which is permanent — the delivery dead-letters with no retry. Your stack's own `i18n.defaultLocale` is the wrong tag here unless it is spelled `en-US`. |
 | **subject** | `string` | ✅ | Subject template |
 | **bodyHtml** | `string` | ✅ | HTML body template |
 | **bodyText** | `string` | optional | Plain-text body template (auto-derived from HTML when omitted) |
--- a/packages/spec/api-surface/system.json
+++ b/packages/spec/api-surface/system.json
@@ -200,6 +200,7 @@
     "Doc (type)",
     "DocSchema (const)",
     "DocTranslation (type)",
+    "EMAIL_TEMPLATE_FLOOR_LOCALE (const)",
     "ENVIRONMENT_ARTIFACT_SCHEMA_VERSION (const)",
     "EmailAddressConfig (type)",
     "EmailAddressConfigSchema (const)",
--- a/packages/spec/export-origins/system.json
+++ b/packages/spec/export-origins/system.json
@@ -194,6 +194,7 @@
     "Doc": "src/system/doc.zod.ts#Doc (type)",
     "DocSchema": "src/system/doc.zod.ts#DocSchema (const)",
     "DocTranslation": "src/system/doc.zod.ts#DocTranslation (type)",
+    "EMAIL_TEMPLATE_FLOOR_LOCALE": "src/system/email-template.zod.ts#EMAIL_TEMPLATE_FLOOR_LOCALE (const)",
     "ENVIRONMENT_ARTIFACT_SCHEMA_VERSION": "src/system/environment-artifact.zod.ts#ENVIRONMENT_ARTIFACT_SCHEMA_VERSION (const)",
     "EmailAddressConfig": "src/system/email-config.zod.ts#EmailAddressConfig (type)",
     "EmailAddressConfigSchema": "src/system/email-config.zod.ts#EmailAddressConfigSchema (const)",
--- a/packages/spec/src/stack-email-template-locale-floor.test.ts
+++ b/packages/spec/src/stack-email-template-locale-floor.test.ts
@@ -120,8 +120,10 @@ describe('#17614 — and stays silent where the bundle HAS a floor (the controls
   });
 
   it('silent when the row omits `locale` entirely — the schema default already IS the floor', () => {
-    // This is the post-parse seam. Run pre-parse, the key is not there yet and
-    // this stack would be reported for a floor it actually has.
+    // Measured, by ablation: moving the call pre-parse does NOT break this —
+    // the reader mirrors the schema default for a missing key, so the two
+    // agree and the case is silent either way. The pin is the behaviour (an
+    // omitted `locale` is never reported), not the call site.
     const { warns, value } = warningsOf(stack([tpl(undefined)], THE_TRAP));
     expect(floorWarns(warns)).toEqual([]);
     expect(value.emailTemplates?.[0]?.locale).toBe(EMAIL_TEMPLATE_FLOOR_LOCALE);
--- a/packages/spec/src/stack-email-template-locale-floor.test.ts
+++ b/packages/spec/src/stack-email-template-locale-floor.test.ts
@@ -1,0 +1,146 @@
+// Copyright (c) 2026 ObjectStack. Licensed under the Apache-2.0 license.
+//
+// #17614 — an `emailTemplates` bundle tagged with the stack's OWN
+// `i18n.defaultLocale` has no fallback floor.
+//
+// Measured before this landed, on an unmodified tree: a stack declaring
+// `defaultLocale: 'en'`, `supportedLocales: ['en','zh-CN','ja-JP','es-ES']`
+// and one row per those tags parsed with ZERO warnings and ZERO errors, and
+// the resolver then answered `TEMPLATE_NOT_FOUND` — which classifies
+// PERMANENT, so the delivery dead-letters with no retry — for `de-DE` and for
+// the literal `en-US`. The identical bundle with its English row tagged
+// `en-US` delivered for `de-DE`. `sendTemplate` matches `(name, locale)`
+// exactly and retries exactly one rung, the literal `en-US`; there is no
+// language-subtag folding, and that ladder's shape is a settled ruling this
+// change deliberately does not touch. The remedy is the bundle, so the
+// diagnostic is where the author is standing.
+//
+// These pin the diagnostic ADVISORY: every case asserts the parse still
+// succeeds and the stack comes back unchanged. The diagnostic narrows what
+// passes SILENTLY; it moves no accept set.
+
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { defineStack } from './stack.zod';
+import { EmailTemplateDefinitionSchema, EMAIL_TEMPLATE_FLOOR_LOCALE } from './system/email-template.zod';
+
+const MANIFEST = { id: 'acme', name: 'acme', version: '1.0.0', namespace: 'acme', type: 'app' as const };
+
+const tpl = (locale: string | undefined, name = 'acme.welcome') => ({
+  name,
+  label: `Welcome ${locale ?? '(default)'}`,
+  ...(locale === undefined ? {} : { locale }),
+  subject: `[${locale ?? 'default'}] Hi {{name}}`,
+  bodyHtml: `<p>[${locale ?? 'default'}] Hi {{name}}</p>`,
+});
+
+const stack = (tpls: unknown[], i18n?: unknown) => ({
+  manifest: MANIFEST,
+  ...(i18n === undefined ? {} : { i18n }),
+  emailTemplates: tpls,
+}) as never;
+
+const THE_TRAP = { defaultLocale: 'en', supportedLocales: ['en', 'zh-CN', 'ja-JP', 'es-ES'], fallbackLocale: 'en' };
+
+/** Run `defineStack` capturing every `console.warn` it emits. */
+function warningsOf(config: unknown): { warns: string[]; value: ReturnType<typeof defineStack> } {
+  const spy = vi.spyOn(console, 'warn').mockImplementation(() => {});
+  try {
+    const value = defineStack(config as never);
+    return { warns: spy.mock.calls.map((c) => c.map(String).join(' ')), value };
+  } finally {
+    spy.mockRestore();
+  }
+}
+
+const floorWarns = (warns: string[]) => warns.filter((w) => w.includes('no fallback floor'));
+
+afterEach(() => {
+  // The reporter is warn-once per (bundle, tags); each test authors its own
+  // tag set so the cache cannot mask a case, but keep runs independent.
+  vi.restoreAllMocks();
+});
+
+describe('#17614 — the floor tag is one named constant, not two spellings', () => {
+  it('is the schema default for `locale`, so a row that omits the key HAS the floor', () => {
+    expect(EMAIL_TEMPLATE_FLOOR_LOCALE).toBe('en-US');
+    const parsed = EmailTemplateDefinitionSchema.parse({
+      name: 'acme.welcome', label: 'W', subject: 's', bodyHtml: '<p>b</p>',
+    });
+    expect(parsed.locale).toBe(EMAIL_TEMPLATE_FLOOR_LOCALE);
+  });
+
+  it("states the two facts an author needs in the published `describe` — the floor tag and what a miss costs", () => {
+    // The `describe` is the authoring surface (it lands in the generated
+    // reference docs). The rule used to live only in a docstring two packages
+    // away, under a heading about a different subject — which is the finding.
+    const text = String(EmailTemplateDefinitionSchema.shape.locale.description ?? '');
+    expect(text).toContain('en-US');
+    expect(text).toContain('TEMPLATE_NOT_FOUND');
+  });
+});
+
+describe('#17614 — defineStack reports a bundle with no `en-US` floor', () => {
+  it('reports the trap: every supportedLocales tag authored, none of them the floor', () => {
+    const { warns, value } = warningsOf(stack(['en', 'zh-CN', 'ja-JP', 'es-ES'].map((l) => tpl(l)), THE_TRAP));
+    const hits = floorWarns(warns);
+    expect(hits).toHaveLength(1);
+    expect(hits[0]).toContain("emailTemplates 'acme.welcome'");
+    expect(hits[0]).toContain("none tagged 'en-US'");
+    expect(hits[0]).toContain('TEMPLATE_NOT_FOUND');
+    // ADVISORY — the accept set did not move: the stack parsed and came back
+    // carrying exactly the tags the author wrote.
+    expect(value.emailTemplates?.map((t) => t.locale)).toEqual(['en', 'zh-CN', 'ja-JP', 'es-ES']);
+  });
+
+  it('reports a single-row bundle tagged with the stack default — the consistent thing an author writes', () => {
+    const { warns, value } = warningsOf(stack([tpl('en')], THE_TRAP));
+    expect(floorWarns(warns)).toHaveLength(1);
+    expect(value.emailTemplates).toHaveLength(1);
+  });
+
+  it('reports each bundle separately, by name', () => {
+    const { warns } = warningsOf(stack(
+      [tpl('en', 'acme.welcome'), tpl('zh-CN', 'acme.welcome'), tpl('en', 'acme.reset')],
+      THE_TRAP,
+    ));
+    const hits = floorWarns(warns);
+    expect(hits).toHaveLength(2);
+    expect(hits.some((w) => w.includes("'acme.welcome'"))).toBe(true);
+    expect(hits.some((w) => w.includes("'acme.reset'"))).toBe(true);
+  });
+});
+
+describe('#17614 — and stays silent where the bundle HAS a floor (the controls)', () => {
+  it('silent when the bundle carries an en-US row beside the supported tags', () => {
+    const { warns, value } = warningsOf(stack(
+      ['en-US', 'zh-CN', 'ja-JP', 'es-ES'].map((l) => tpl(l)), THE_TRAP,
+    ));
+    expect(floorWarns(warns)).toEqual([]);
+    expect(value.emailTemplates).toHaveLength(4);
+  });
+
+  it('silent when the row omits `locale` entirely — the schema default already IS the floor', () => {
+    // This is the post-parse seam. Run pre-parse, the key is not there yet and
+    // this stack would be reported for a floor it actually has.
+    const { warns, value } = warningsOf(stack([tpl(undefined)], THE_TRAP));
+    expect(floorWarns(warns)).toEqual([]);
+    expect(value.emailTemplates?.[0]?.locale).toBe(EMAIL_TEMPLATE_FLOOR_LOCALE);
+  });
+
+  it('silent when the stack declares no supportedLocales to measure against', () => {
+    const { warns } = warningsOf(stack([tpl('en')]));
+    expect(floorWarns(warns)).toEqual([]);
+  });
+
+  it('silent when no authored tag is one this stack claims to support', () => {
+    const { warns } = warningsOf(stack([tpl('fr-CA')], { defaultLocale: 'en', supportedLocales: ['en'] }));
+    expect(floorWarns(warns)).toEqual([]);
+  });
+
+  it('warns once for one bundle, however many times the same stack is defined', () => {
+    const first = warningsOf(stack([tpl('pt-BR')], { defaultLocale: 'pt-BR', supportedLocales: ['pt-BR'] }));
+    const second = warningsOf(stack([tpl('pt-BR')], { defaultLocale: 'pt-BR', supportedLocales: ['pt-BR'] }));
+    expect(floorWarns(first.warns)).toHaveLength(1);
+    expect(floorWarns(second.warns)).toEqual([]);
+  });
+});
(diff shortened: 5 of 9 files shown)


os-musk:
Served-tier: claude-fable-5-1

## Contract review (post-hoc, at tier)

Reviewed head: `4feb4e245c3fc0eb7692f85bd30d282f3b4b72ca` (merged 2026-09-12T19:42:54Z as squash `a61ae59f9`). Merge base `e7fea4665d3949158263fc66a31005ed714052c9` — the second parent of the branch's own `5d63af8f` main-merge; the PR record's `base.sha` (`fe71032c`) predates that merge and is stale. The three-dot diff `e7fea466...4feb4e24` has patch-id `59b1eb70…`, byte-identical to the squash commit's diff against its parent `29dd1a6dd`, and its per-file numstat equals GitHub's `/pulls/17884/files`: **9 files, +401 / −6** — `stack.zod.ts` +86/−1, `email-template.zod.ts` +50/−4, one row each in `api-surface/system.json` and `export-origins/system.json`. The census shape is confirmed by measurement, not adopted. `main` readings below are `git show` / `git grep` on `origin/main` = `611795e40558e722334bebd1239a0cff7cd55d42` after a fetch on 2026-09-13; objectui readings are on its `origin/main` = `0135613365cad77f381baee1b1ab7d4f06461086` (the `.objectui-sha` pin lags at `53ded82b`).

This is an independent re-run: I read the card (#17614, body and all 5 comments), the PR body, the two review comments (`5647795889`, `5648094458`) and the full merge-base diff, and took no characterisation from the dispatch.

### ① Derived judgments — every accept-set and public-surface change in the diff

1. **`EMAIL_TEMPLATE_FLOOR_LOCALE` — new `const` export.** Reachable on the `@objectstack/spec/system` entry only (`system/index.ts:72` `export *`; the root entry does not re-export `system`, and `api-surface/root.json` lists neither it nor `EmailTemplateDefinitionSchema`). Registered in `api-surface/system.json` and `export-origins/system.json`. **Public surface WIDENED by one symbol.** Warranted — one fact (schema default and resolver rung) named once. Its docblock is judged wrong on one path: see D1.
2. **`EmailTemplateDefinitionSchema.locale`: `.default('en-US')` → `.default(EMAIL_TEMPLATE_FLOOR_LOCALE)`.** Same literal. No refinement, no key added or removed. **Accept set HELD.**
3. **`EmailTemplateDefinitionSchema.locale` `.describe()` and TSDoc rewritten.** Published text surface (the `.d.ts`, and `content/docs/references/system/email-template.mdx:47`). Correct for a call that names a locale; **false for a call that names none** — D1 — and imprecise in one term — D5.
4. **`defineStack`: new post-parse `console.warn` (`warnEmailTemplateLocaleFloor`, `stack.zod.ts:2706-2745`, called at `:2843`).** Runs after `StackSchema` parse and before `mergeActionsIntoObjects`; it cannot alter the parse result. **Accept set HELD** (the round's and the review's probe both read `parsed-ok=true`, identical stored tags, warning count 0→1; I did not re-run it — see "not measured"). Right instrument under triage's comparative standard (`warnUnknownAuthoringKeys`, same seam, same warn-once). Its message text repeats D1's claim verbatim ("retries only the literal 'en-US'").
5. **`api-surface/system.json` +1, `export-origins/system.json` +1.** Correct registration of (1). Both files have since moved on `main` (other PRs' rows); the `EMAIL_TEMPLATE_FLOOR_LOCALE` row is still present at `system.json:203` and `export-origins/system.json:197`.
6. **`content/docs/references/system/email-template.mdx:47`** regenerated from (3). Now contradicts the hand-written page on the same docs site — D1.
7. **`.changeset/email-template-locale-floor.md`** — `minor`; its line 7 repeats D1's claim. Still unconsumed on `main` (same blob as the head), so still correctable before a release publishes the sentence.
8. **Tests** (`stack-email-template-locale-floor.test.ts`, `email-template-floor-locale-parity.pin.test.ts`) and **`vitest.repo-tests.json`** — no published surface. The parity pin reads `plugin-email` source as text and is registered; it is a real pin (C1 closed correctly, option (a) was the right choice).

**Summary: accept set HELD; public surface WIDENED (+1 export on `/system`); published text surface changed, and part of that text is false against the runtime.**

### ② Semver grading

Declared `minor` on `@objectstack/spec`. Measured: the new export alone makes it `minor`; the describe/TSDoc alone would be `patch` (text surface — inputs that validated before validate identically after); the `defineStack` side effect is additive and non-breaking. **`minor` is the correct grade and it was reached for the right reason (the export).** The clause-② `yes` is likewise right on the published-surface limb, mechanically; the round's own correction of the "narrowing" rationale stands.

### ③ Boundary flags

**Raised by the round and the review — verified:**
- **Fence held.** 0 of the 9 diff paths are under `packages/plugins/plugin-email`; lit control on the same predicate: 7 of 9 under `packages/spec`. `EmailService.resolveAndRenderTemplate` is byte-identical between the PR head and today's `main`.
- Two silent floorless shapes were named (no `i18n` block; `declared.length === 0` — `supportedLocales: ['en']` with a lone `fr-CA` row) and the `console.warn` audibility point (an `os lint` / `os validate` rung). Recorded in the PR body as a "successor-card seed". **Not filed** — D4.
- The parity constant duplication (`DEFAULT_TEMPLATE_LOCALE` in `plugin-email`) — pinned by C1; correct.

**Should have been raised, and were not — the at-tier findings:**

**D1 — The declared floor does not match what the runtime does when the locale is missing.** The PR's new contract text says, in five places, that the resolver retries *exactly one* rung and that a bundle with no `en-US` row has *no fallback floor at all* and dead-letters:
- `email-template.zod.ts:27-28` ("the sole rung `IEmailService.sendTemplate` retries after an exact `(name, locale)` miss"), `:33-34` ("has no fallback floor at all"), `:104-106` ("then retries exactly one rung"), `:126-131` (the published `describe`: "with one retry rung … A bundle with no `en-US` row therefore has no fallback floor … dead-letters with no retry");
- `stack.zod.ts:2674-2675` and the warn text at `:2725-2731`; `.changeset/email-template-locale-floor.md:7`; `references/system/email-template.mdx:47`.

The runtime on `main` has a **third rung on the no-locale path**. `packages/plugins/plugin-email/src/email-service.ts:1318-1330`: named locale → exact → `en-US` → throw; **no locale** → `en-US` → `loader.load(template, undefined)`; and `template-loader.ts` `load()` with no locale answers `(name, 'en-US')` then `first({ name }, BY_LOCALE)` — **the bundle's lowest locale tag, silently**. The runtime comment at `email-service.ts:1324-1329` defends it by name ("hard-failing it now would swap one bug for an outage"). It is CI-pinned: `template-locale-resolution.test.ts:155-167` ("a bundle with no en-US row at all … falls back to the lowest locale tag"; "keeps a single-locale tenant working rather than resolving to nothing") and `:325-340`, where `svc.sendTemplate({ template: 'receipt', to, … })` with no `locale` on a **zh-CN-only bundle DELIVERS**. So the exact sentence the PR published — a bundle without `en-US` "has no fallback floor at all" — is contradicted by a pinned delivery.

The no-locale call is reachable from every production caller, by design: auth spreads `...this.emailLocaleArg(request, storedLocale)` (`auth-manager.ts:1566`), which is empty when stored locale, `Accept-Language` and the deployment default are all absent; messaging's `getDefaultTemplateLocale` (`messaging-service-plugin.ts:164-172`) returns `undefined` when no `i18n` service or no default is registered; the inbox channel spreads `locale` only when resolved (`inbox-channel.ts:182-186`); `recipient-locale.ts` says so verbatim ("Rung 2 absent too ⇒ `undefined`").

And the contradiction is **inside `packages/spec` itself**: `packages/spec/src/contracts/email-service.ts:162-173` (`SendTemplateInput.locale`) documents three rungs, rung 3 verbatim — "only for a call that named no locale, and only when the bundle has no `en-US` row at all: its lowest locale tag". The hand-written `content/docs/automation/email-templates.mdx:127-137` documents the same three. Today the docs site says both "one retry rung, no fallback floor" and "then deterministic: the lowest locale tag" about one ladder.

Why the below-tier round missed it: the review verified the ladder "byte-for-byte" by citing `email-service.ts:1317-1324` and stopped before `:1327-1330`. The card's own quoted snippet contained the third-rung line (`if (!row && !preferred) { row = await loader.load(input.template, void 0); }`); the card's prose, the PR and the review all read it as one rung. The PR set out to replace a describe that lied in one direction ("best match", implying folding) and replaced it with one that lies in the other (declares a dead-letter where the runtime silently fills — the shape the project's own doctrine names as the AI-error breeding ground). The fix is text-only and inside the fence: scope every "exactly one rung / sole rung / no floor at all / dead-letters" claim to *a call that names a locale*, state rung 3 or cross-reference `SendTemplateInput.locale`, and make the same edit in the warn text, the constant docblock and the changeset.

**D2 — The other authoring surface in the same directory was left silent.** `packages/spec/src/system/email-template.form.ts:31` — `helpText: 'BCP-47 tag — e.g. en-US, zh-CN'` — is what a Studio author reads at the moment they type `locale` (mirrored into `platform-objects/…/metadata-forms.generated.ts:1391`), and `platform-objects/src/audit/sys-email-template.object.ts:64-70` (`description: 'BCP-47 locale tag (en-US, zh-CN, …)'`) is the record editor's. The card's finding was "the authoring surface never mentions it"; the PR fixed the Zod describe and left the form's help text and the object's field description saying nothing about the floor. In-fence, in-package for the form.

**D3 — The "best match" wording the card indicted survives in three declarations, one of them in `packages/spec` on the very interface the new describe cites.** `packages/spec/src/contracts/email-service.ts:153-154` (`SendTemplateInput.template`: "The service picks the best-matching locale row (falls back to `en-US`)"); `platform-objects/src/audit/sys-email-template.object.ts:10-11`; `packages/metadata-core/src/item-key-discriminators.ts:7-10`, which **quotes** the old `EmailTemplateDefinitionSchema` describe ("the service picks the best match for the recipient's locale") as what the spec "states" — that text no longer exists in the spec. Two more sit behind the fence and are notes only: `plugin-email/src/template-loader.ts:18-19` (same stale quotation) and `email-service.ts:467`.

**D4 — No successor card exists.** Prime Directive #10: a metadata-authoring trap gets a card, everything else a note. The two silent floorless shapes are the same trap class as #17614 itself. Enumerated via REST `GET /issues?state=all&since=2026-09-12T19:00:00Z` (3 pages, 257 rows; lit control: #17614 and #17884 both present): **0** issues created after the merge whose title names floor / emailTemplates / email-template / locale / os lint / 17614, and **0** issue bodies citing 17614, `EMAIL_TEMPLATE_FLOOR_LOCALE` or `warnEmailTemplateLocaleFloor` other than the PM seat post #6015. C2 asked for it "on the card"; it landed in the PR body as a seed and stopped there.

**D5 — Terminology in the published describe.** `email-template.zod.ts:127-128` says "`en` and `en-US` are different bundles"; the TSDoc four lines above (`:101`) says "Multiple rows sharing one `name` form an i18n bundle". Rows differ by locale; the bundle is the name. Minor, published in `references/system/email-template.mdx:47`.

**D6 — The diagnostic's own reach (boundary, not a defect of the ruling).** `warnEmailTemplateLocaleFloor` runs inside `defineStack` only. Templates authored via `defineEmailTemplateDefinition` (the shape `content/docs/automation/email-templates.mdx:104-121` teaches) and registered by a plugin, or authored in Studio (D2), never pass through it. Belongs in D4's card.

**objectui coupling — lit zero.** On objectui `origin/main`: 0 hits for `EMAIL_TEMPLATE_FLOOR_LOCALE`, `DEFAULT_TEMPLATE_LOCALE`, `TEMPLATE_NOT_FOUND` and the describe text `BCP-47 locale (e.g` (controls on the same corpus: `defineStack` 38, `BCP-47` 69); 0 tests assert this describe (control: 146 test lines assert some spec `.description`). The 9 `EmailTemplateDefinition` hits are `preview-samples-spec-valid.test.ts` (validates samples through `ObjectStackSchema` — accept set, which held), CHANGELOGs and flow-inspector mapping strings. **Nothing in objectui goes red on the spec bump this PR produces.**

### ④ STILL TRUE ON `main` TODAY?

| # | defect | on `origin/main` `611795e40` | status |
|---|---|---|---|
| D1 | contract text declares one retry rung / no floor / dead-letter; runtime has a CI-pinned third rung on the no-locale path; `SendTemplateInput.locale` in the same package says so | `email-template.zod.ts` blob `c800b79ea` unchanged since the head; `email-service.ts:1327-1330`, `template-loader.ts` rule 3, `contracts/email-service.ts:170-173`, `automation/email-templates.mdx:135-137` all present | **SURVIVES — card** |
| D2 | `email-template.form.ts:31` help text and `sys-email-template.object.ts:64-70` say nothing about the floor | both present as quoted | **SURVIVES — card** (fold into D1's card) |
| D3 | "best-matching" prose in `contracts/email-service.ts:153-154`, `sys-email-template.object.ts:10-11`, and a stale quotation of the old describe in `item-key-discriminators.ts:7-10` | all present as quoted | **SURVIVES — card** (fold into D1's card; the two fenced copies are notes) |
| D4 | successor card for the two silent floorless shapes + the audibility rung | none filed (lit enumeration above) | **SURVIVES — card** |
| D5 | "different bundles" wording | present, `email-template.zod.ts:127-128` | **SURVIVES — note, ride on D1's fix** |
| C1 | unpinned "must stay equal" | `email-template-floor-locale-parity.pin.test.ts` present on `main`, registered (`vitest.repo-tests.json`, 31 entries) | repaired pre-merge — note |
| C2 | second silent case unnamed | named in the PR body | recorded, not filed — see D4 |

### Not measured here

No test or gate was re-run: this checkout carries no `node_modules` or `dist`, and every runtime claim above is pinned by a CI test I cite by file and line on `origin/main`, so an install-and-build under the verify lock would have added no reading. The round's `check:cross-package-test-inputs` and both vitest project exits are adopted from its record, not reproduced. Zero-hit readings: 8, all 8 with a lit control on the same corpus.

Implemented-by: claude/issue-17614-email-template-locale-floor
Reviewed-by: session_01EfsizFDgAcEjpwv4oM3WGT

**Verdict: WOULD HAVE FAILED** — on the published-contract-text limb (D1): the head as merged publishes, in the schema `describe`, the `.d.ts` TSDoc, the new constant's docblock, the `defineStack` warning and the changeset, a dead-letter that the runtime does not produce and that the same package's `SendTemplateInput.locale` contract text and the hand-written docs page both contradict. An at-tier review would have held the head on a third condition beside C1 and C2, in-fence and text-only, before clearing it. The accept set held, the export is warranted, the fence held, the semver grade is right, and objectui is untouched — none of which rescues a contract that states a failure the runtime silently turns into a delivery.


---

Issue #17751, opened 2026-09-12
spec: retire `ChartConfigSchema.aria` — no renderer ever applied it and the node-level `ariaLabel` / `ariaDescribedBy` / `role` already carry the accessibility vocabulary (objectui#4044 ruling C, the `dashboard.aria` precedent)


os-tesla:
Filed by the director seat on the maintainer's ruling for objectui#4044 (decision batch #118 item 2, 2026-09-12: 「其他同意」 to the seat's recommendation **C** — judge the protocol wrong for this one key). ⛔ `packages/spec` is the `domain:spec` seat's; the director seat writes no repo file.

## Governing text

- Maintainer standing ruling (2026-09-10, verbatim): 「我们的项目以objectstack 协议为准，文档应该以实际实现为准。协议不正确的应该先修改协议。」
- objectui#4044 ruling (5617466747) — implementation follows the protocol for `DashboardWidget.chartConfig`; PR objectui#9202 (merged 2026-09-11) forwarded 9 of the 10 DO-NOW keys. The one it could not forward is `aria`, and the seat stopped and reported (objectui#4044 comment 5639000150).
- Precedent, verbatim from `objectui/packages/types/src/complex.ts:1915-1925`: *"the spec removed `dashboard.aria` at the #3896 audit close-out (no dashboard renderer ever applied it), so `DashboardSchema.shape.aria` is a tombstone that refuses any value"* — pinned by `__tests__/dashboard-aria-retired-contract-twins.test.ts`.

## The defect in the protocol

`ChartConfigSchema.aria` (`packages/spec/src/ui/…` — locate by symbol) declares a nested accessibility object on a chart's configuration. Measured on objectui `origin/main` (5639000150): no renderer reads it — `ChartRenderer` drops every prop but `schema` / `onChartClick`, `AdvancedChartImpl` declares no `aria` prop, `normalizeChartSchema` names it nowhere; forwarding it nested or flattened changed nothing in the DOM (lit control: `description` sets `aria-label` / `role='img'` on the same node). Meanwhile the platform already carries ONE accessibility vocabulary at the node level — `BaseSchema.ariaLabel` / `ariaDescribedBy` / `role`, converted to DOM attributes by `SchemaRenderer.resolveAriaProps` everywhere. `chartConfig.aria` is a second, block-local spelling of the same thing, with no reader, and giving it a reader would create two accessible-name sources on one element that then need a precedence rule.

[...]

PR #18231, opened 2026-09-14
fix(metadata-protocol): GET /meta/types stops publishing properties no instance can satisfy

### LINKED ISSUE #17502: spec: `retiredKey()` tombstones are emitted into a repeater row's `items.properties`, so the served schema offers six columns the parse door refuses
Found while surveying #17232 (PR #17500), deriving the row schema of every `type: 'repeater'` form field. Out of that card's scope, which is about MISSING titles on live keys.

Same offer-vs-door family as #5280 (`app`'s metadata form offered three `retiredKey()`-retired keys), one layer down: #5280 was the form's DECLARED field list, this is the JSON-Schema-derived row of a repeater table, which the form layout does not enumerate at all.

## The measured half — the served schema carries a tombstone as a row property

`retiredKey()` (`packages/spec/src/shared/retired-key.ts`) is `z.never({ error: () => guidance }).optional().describe('[REMOVED] ' + guidance)`. Authoring the key is a hard parse rejection.

`z.toJSONSchema` still emits a property node for it, and it lands in the repeater's `items.properties`. Measured on `origin/main` at `e758131b39`, `flow.nodes`:

```
items.properties keys = id, type, label, config, connectorConfig, position,
                        timeoutMs, inputSchema, outputSchema, waitEventConfig,
                        boundaryConfig
outputSchema.description = "[REMOVED] `flow.nodes[].outputSchema` was removed in
                            @objectstack/spec 17.0.0 (audit close-out) ..."
```

Lit control: `inputSchema`, the live sibling one line above it, is present and carries no `[REMOVED]` prefix. Dark control: a key that was never declared at all is absent from the map.

`dashboard.widgets` carries five more of them, so six tombstones sit in repeater rows across the registry today.

## The unmeasured half, named as unmeasured

Whether Studio paints a column header for a `{ "not": {} }` node is an objectui question and was NOT measured here — this session has no console to drive. The console reads `items.properties[k].title ?? k` to build the header row, and a tombstone is a member of `items.properties`, so the offer exists in the payload; what the renderer does with it is the half still owed. Both readings are worth having before anything is changed: if it does render, an author is invited to fill a key that is refused at publish; if it does not, the exclusion still belongs in the served schema rather than in each renderer.

## Why the spec side is the place to look first

`packages/spec/src/kernel/repeater-item-titles.test.ts` (landing in PR #17500) already has to exclude these by hand — it filters on the `[REMOVED] ` description prefix — precisely so it does not demand an authoring title for a key the parse door refuses. That exclusion is a local workaround for a payload that should arguably not carry the node at all. One place to decide it, not one per consumer.

Carrier: `packages/spec/src/shared/retired-key.ts` and `packages/metadata-protocol/src/protocol.ts` (`toJsonSchemaSafe`), with the render half in objectui.

Back-links: #17232, PR #17500, #5280.

---

os-warren:
Fixes #17502

`GET /meta/types` published every `retiredKey()` tombstone as a property node beside the live keys. `z.toJSONSchema` renders a tombstone as a node carrying a `[REMOVED] ` description and `not: {}` — truthful to a consumer that reads the subschema, invisible to one that reads the KEY SET. Studio builds a repeater's column headers from `items.properties[k].title ?? k`, so a tombstone inside a row shape became a column an author was invited to fill and the publish door then refused.

`toJsonSchemaSafe` now drops every property whose subschema admits no instance, before it serves or caches the document. The predicate is structural — it asks the JSON Schema question "does this admit any instance at all" — never a `[REMOVED] ` prefix match, which would put a second hand-written spelling of "this is a tombstone" into a consumer. A property that admits nothing **and** is `required` is kept: dropping it would turn "this object admits nothing" into "this object admits anything".

This PR has been through an at-tier contract review that returned **FAIL** on two findings. Both are fixed below. The review record is comment `5672966288`; this body was written from it, in session `https://claude.ai/code/session_01KB5PFtxuy1x3dcR5gxudx6`.

### FAIL 1 — the walk was not position-aware and widened a live node

`walk()` applied the properties-map logic at **every** object node it visited, including a node that IS a `properties` / `$defs` map. A property literally named `properties` therefore had its own keywords read as property subschemas, and any keyword valued `{ not: {} }` under it was deleted. `additionalProperties`, `items` and `propertyNames` all use `{ not: {} }` to say "and nothing more", so deleting one **widens a live node** — the one thing the module header says it must not do.

All three reproductions were re-derived here against the branch's own exported function before any edit, and re-run after. Two of them are pure zod with no hand-written input.

Before the fix (branch tip `9eaf3c08a5`):

```
R1  z.toJSONSchema(z.object({ properties: z.record(z.string(), z.never()) }))
    in   ... "properties":{"properties":{"type":"object","propertyNames":{"type":"string"},"additionalProperties":{"not":{}}}} ...
    out  ... "properties":{"properties":{"type":"object","propertyNames":{"type":"string"}}} ...
    changed(byRef): true   changed(byValue): true     -> admitted only {}, now admits any object

R2  z.toJSONSchema(z.object({ properties: z.array(z.never()) }))
    in   ... "properties":{"properties":{"type":"array","items":{"not":{}}}} ...
    out  ... "properties":{"properties":{"type":"array"}} ...
    changed(byRef): true   changed(byValue): true     -> admitted only [], now admits any array

R3  { "$defs": { "properties": { "type": "object", "additionalProperties": { "not": {} } } } }
    out  { "$defs": { "properties": { "type": "object" } } }
    changed(byRef): true   changed(byValue): true     -> same deletion inside $defs
```

After the fix (`a8958b9f1c`), same script, same inputs:

```
R1  changed(byRef): false   changed(byValue): false   (returned by reference, byte-identical)
R2  changed(byRef): false   changed(byValue): false
R3  changed(byRef): false   changed(byValue): false
```

**The fix.** The walk now splits by position. `walkSchema` is the only place a property may be dropped, because it is the only position where the deciding `required` array is a sibling. `properties`, `patternProperties`, `dependentSchemas`, `$defs` and `definitions` are walked by `walkSchemaMap`, which hands every VALUE back to `walkSchema` and drops nothing — their keys are author-chosen names, not keywords, and neither `patternProperties` nor `$defs` has a `required` array that could license a drop (a `$defs` entry may also be the target of a `$ref`).

**The mirror defect, fixed by the same change.** Because the map was read as a node, a property NAMED `required` or `default` bought its whole subtree an exemption from the walk — `NON_SCHEMA_KEYS` skipped it. That is a missed strip rather than a widening, and it is pinned in the same commit.

**Served exposure is zero, before and after.** No served `properties` or `$defs` map has an entry named `properties` today, and the new `over-drop guard` below proves every one of the 27 served documents is still exactly its own derivation minus unsatisfiable nodes.

**The unit pin that was missing.** `unauthorable-nodes.test.ts` now carries three: the two pure-zod position cases, the `$defs` case, and the keyword-name-collision mirror. **Proven red against the broken walk first** — 3 failed / 6 passed, `expected undefined to deeply equal { not: {} }` on R1 and R3 and `expected [ 'dead', 'live' ] to deeply equal [ 'live' ]` on the mirror — then green after the fix.

### FAIL 1b — the over-drop blind spot the rewritten pin gained

The review's mutation **mut4b** — an over-eager strip dropping a nested live key outside `dashboard.widgets` — ran green. Since #17502 the blast-radius baseline is `stripUnauthorableProperties(preFixDerivation(type))`, so a strip defect sits on **both** sides of that comparison and cancels itself out.

Closed by a new assertion in this card's own pin file, `over-drop guard: the served payload is its derivation MINUS unsatisfiable nodes, nothing else`. It reads the removals off the served payload and its derivation by a **parallel walk of the two documents** — deliberately not a second implementation of the strip, so the defect cannot appear on both sides again. Three verdicts: every served document is still a pure DELETION of one of its two derivation arms (nothing added, nothing rewritten); **every removed node must admit no instance**; and a non-vacuity control pinning `dashboard`'s eight removals at both depths (the five repeater-row columns plus three top-level tombstones), sorted so key order is not what is pinned.

Proven by re-running mut4b's own mutation, under a trap, with the blob hash checked both ways:

```
== BASELINE ==
HEAD blob            : 9ec8cfe07c7bbfbac8154607094ececa45783f99
on-disk blob         : 9ec8cfe07c7bbfbac8154607094ececa45783f99
occurrences OLD      : 1          occurrences maxTokens: 0
== MUTATED (|| key === 'maxTokens' added to the drop condition) ==
on-disk blob         : 928e48116e77a4f6b203228107fa13f952e1873b
occurrences OLD      : 0   (expect 0)   NEW: 1   maxTokens: 1
== RUN under mut4b ==
VITEST_EXIT=1
  x over-drop guard: the served payload is its derivation MINUS unsatisfiable nodes, nothing else
  + [ "agent.properties.model.properties.maxTokens" ]
  Test Files  1 failed | 2 passed (3)      Tests  1 failed | 34 passed (35)
== RESTORED ==
on-disk blob         : 9ec8cfe07c7bbfbac8154607094ececa45783f99
occurrences maxTokens: 0   (expect 0)
git diff HEAD        : []          git status --porcelain: []
```

Note what the run also shows: `protocol.meta-types-degenerate-derivation.test.ts` stayed **green** under the same mutation. The blind spot is real, it is where the review said it was, and the new guard is the thing that closes it — naming the exact live node that was dropped.

### FAIL 2 — the shipped changeset's consumer census was false

The changeset said "5 reachable as repeater columns" and "exactly **two**". Both are false.

**The sha.** `.objectui-sha` in this worktree reads **`53ded82bf7a494f54e344e19099dbf00854b8694`** — read here, from the file, not from a local checkout. The earlier round measured a local objectui at `ff1d5ea`, which is not the pin and is not an ancestor of it in either direction. Every objectui statement below is at `53ded82b`.

**The fix is not a bigger number.** Per the seat's ruling, the census is renderer-dependent and pin-dependent, so any count written into `CHANGELOG.md` is false at the next `.objectui-sha` bump. The changeset now states the **class** — a tombstone arriving as help text, or as a repeater column, under an editable input the publish door refuses — and names the three mechanisms that put one in front of an author:

- **the flat, schema-driven fallback** for a served type carrying no `*.form.ts` layout: its field list IS the served `properties` map, and a nested object renders recursively, so a tombstone at any depth becomes a field with the `[REMOVED] ` prescription as its help text;
- **repeater rows**, whose column headers are `items.properties[k].title ?? k`;
- **server-field grafting**, where an inspector merges the server's top-level properties into a trailing "More fields" section, so a key the UI's own bundled spec predates is offered precisely because the served document is the only place it is known from.

The producer-side census — 80 nodes across 16 types at `74eaab8614`, this PR's merge base, measured over this repo's own served registry — is repo-local and stays. (It read 77 across 15 at the earlier merge base `1bdbf82cb5`; #17751 landed in between and retired `ChartConfigSchema.aria`, taking `dashboard` 8 → 9 and `report` 0 → 2. A reading taken at a tree, not a standing invariant.)

The same false clause in `protocol.meta-types-unauthorable-columns.test.ts`'s header is corrected to match. The five widget columns stay named there as the row that file pins, no longer as the whole reachable set.

## The stop condition, re-derived independently

The dispatch made one question a stop condition: if dropping the node destroys a live prescription channel, stop with an empty diff. The review re-derived this independently and reached the same answer, and it is restated here as the seat asked: **no `retiredKey()` prescription channel is destroyed, and the served schema was never one.** The strip is the right remedy. The FAIL findings are about how the strip walks and about what the changeset claims, not about whether to strip.

What `toJsonSchemaSafe` emits for a `retiredKey()` node, measured over the whole served registry: exactly two keys — a `description` beginning `[REMOVED] `, and `not: {}` — on all 80 nodes, with **no `title`** on a single one (which is what makes the defect invisible to the renderer: the header falls back to the humanized key), identical in both derivations, and 0 of the 80 `required`.

Channels that carry the prescription and are untouched: `tsc` (property of the Zod shape), the parse (pinned in this PR — the refusal carries the FROM/TO prescription byte for byte), `packages/spec`'s `authorable-surface/` ratchet, and the generated reference pages, which print the full prescription on a `never`-typed row.

## Consumer-side, corrected, at the pinned objectui `53ded82b`

| Consumer | Verdict at the pin |
|---|---|
| Repeater row cells (`metadata-admin/widgets.tsx` `RowCell`) | No `description` branch at all; `not` is never consulted. A repeater column loses no TEXT — the removal only withdraws the offer. |
| Sectioned property panel (`METADATA_FORM_REGISTRY` layouts) | Zero layouts declare a retired key (the #5280 fix holds). Nothing rendered, nothing lost. |
| Flat fallback, field face (`SchemaForm.tsx` `FieldRow`) | Renders `schema.description` as help text for a layout-less type. `api.cacheTtl`, `job.timeout`. |
| Flat fallback, **nested-form face** | Recursive `SchemaForm`, so a **depth-1** tombstone renders the same way: `job.retryPolicy.retryDelayMs`. The earlier round intersected TOP-level tombstones only and missed it. |
| Flat fallback, **object-rows face** | `derivePropertyNames(items)` hands a layout-less repeater its columns: `book.groups[].translations` is a sixth offered column. |
| **`mergeServerFields` grafting** | Merges server-only top-level properties into a trailing "More fields" section rendered with the description as help: `dashboard.refreshInterval` and `page.assignedProfiles`, both retired in a spec newer than the one the pinned objectui bundles. **A consumer the earlier round's table never checked.** |
| `tsc`, the parse, `authorable-surface/`, reference pages | Untouched. |

All of these are the same defect class — an offer the door refuses — so the stop condition's answer is unchanged. What changed is that the changeset no longer claims a number for them.

## Semver — `minor`, unchanged

Nothing authorable is removed or renamed, and no valid document changes shape, because every dropped node was unsatisfiable — so this is not `major` and carries no ADR-0087 disposition. It is more than a patch because it narrows a published payload. `check:changeset-no-major`, `check:empty-changeset` and `check:adr-0087-registration` all pass. The review agreed and this round changes nothing about the level.

**On the clause-② carrier**, since this body is being edited: the declaration lives on the claim comment `5670996882`, which carries `Clause-②: yes` at the start of its own line — that comment is what the enqueue gate reads, and the `needs:contract-review` label is on the PR. This body carries no `Clause-②:` line and does not need to; noted because the review noted its absence.

## Scope fence

Clean. Nothing under `packages/spec/src/stack.zod.ts`, any `stack-*.test.ts`, `packages/spec/src/data/analytics.zod.ts`, `packages/spec/src/ui/view.zod.ts`, `packages/spec/src/api/discovery.zod.ts` or `packages/spec/src/meta-spelling/`. No `objectui` file is touched — it was read only, at the pin `53ded82b`. The review's other measurements (the `acceptsNothing` predicate, the `engine-double-contract` ratchet rows, the changeset LEVEL, the docs divergence with `content/docs/references/**`) were all judged correct and are untouched by this round.

## Verification

Exit codes captured before any pipe; gate verdicts read from the gate's own printed line.

| Run, at `21edb645e9` | Exit |
|---|---|
| `pnpm --filter '@objectstack/metadata-protocol^...' build` then the package build | 0 |
| `pnpm --filter @objectstack/metadata-protocol typecheck` | 0 |
| `pnpm --filter @objectstack/metadata-protocol test` (full package suite) | 0 — **178 files passed, 3 skipped; 2548 tests passed, 19 skipped** |
| `pnpm lint` (`eslint . --no-inline-config`, whole repo, no narrowing) | 0 |
| `pnpm check:engine-double-contract` | 0 |
| `pnpm check:nul-bytes` plus an own control-character sweep of the changed files | 0, clean |
| `check:cross-package-test-inputs`, `check:test-source-alias`, `check:type-check-coverage`, `check:type-check-debt`, `check:doc-authoring`, `check:objectui-changeset` | 0 |
| `check:empty-changeset`, `check:changeset-no-major`, `check:adr-0087-registration` (all `--base origin/main`) | 0 |
| Unit pins against the BROKEN walk, before the fix | 1 — 3 failed / 6 passed, exactly the three new pins |
| Ablation: mut4b re-run against the new over-drop guard | 1 — 1 failed / 34 passed, naming `agent.properties.model.properties.maxTokens`; restored, `git diff HEAD` empty |

The full-suite delta is +4 tests over the review's reading of the previous head (2544 passed), which is the three new unit pins plus the over-drop guard.

## Open questions

- **`api`, `job`, `book` and layouts.** Four of the six named sites exist because the type carries no `*.form.ts` layout and the panel falls through to the flat schema-driven list. Whether those types should carry layouts is a separate question from this card and was not touched.
- **A conditional cousin of the `required` guard.** The drop is vetoed by the parent's `required` array. JSON Schema's `dependentRequired` can require a key conditionally, and a drop is a widening there too. Zod emits no `dependentRequired` and none appears anywhere in the served registry, so there is nothing to fix today; recorded rather than filed.

---


os-warren:
## `Test Core (2/6)` is red here, and the cause is a stale base rather than this round's one-line change

**What is failing**, at head `465a38ece9` (2026-09-16T00:01:17Z): 35 check runs — 19 success, 6 skipped, **1 failing** (`Test Core (2/6)`), 9 still going. Run `35037740392` is `status: in_progress`, `conclusion: null`, and the head has not moved, so this is not a superseded run.

**Why it is not this round's change.** The only commit since the previously-green head `0f89189631` is `docs(changeset): anchor the tombstone census to the sha it was measured at`, and its entire diff is **one markdown file**, `.changeset/17502-served-schema-drops-unauthorable-columns.md`, +5/−3, one paragraph rewritten in place. A changeset cannot alter `Test Core` behaviour.

**And it is not a failure on the base either** — control taken on `origin/main` `74eaab8614`: `Test Core` and all six shards read `completed / success`. So the usual "red on main too" exit does not apply.

**What it actually looks like, measured.** This branch's merge-base is `1bdbf82cb5`, **48 commits behind** `origin/main`, and the commit now at the tip of main is:

```
74eaab8614 feat(spec,core)!: the startup contract describes what the kernel produces
           — retire the orchestrator, declare PluginStartupResult once (#16059) (#18303)
```

That landed at 23:25Z, **after** this PR's last all-green reading. This branch's own surface is `packages/metadata-protocol/**` plus `scripts/engine-double-contract.pinned.json` — code and pins that census the **served schema registry**, and #18303 retired three spec defs and re-declared a fourth, which moves exactly that registry. A `pull_request` run tests the merge of head into base, so the run now sees a registry this branch's pins were measured against a different version of.

That is a stale-base interaction, not a flake and not someone else's failure, so no re-run is being spent on it. ⛔ No "flake" is being claimed, and the failing check is not being re-run to make it go away.

**The remedy is the ordinary one:** merge `origin/main` into the branch and re-validate the census-shaped pins against the merged tree. The implementer deliberately did not merge main during the patch round — 「out of scope for this round, and it would move the head CI is green on」 — which was a correct call at the time; main has since moved and the head is no longer green, so the reason has expired. It is going back to the implementer now.

**One thing the anchoring fix gets right, worth naming:** the changeset sentence now says the 77-across-15 census was taken at `1bdbf82cb5` specifically, 「a reading taken at that tree, not a standing invariant」. If merging main changes the number, the sentence stays true and only the figure needs re-measuring — which is exactly why the contract review asked for a sha anchor rather than a version.

---

os-warren:
## Contract review

Head-sha: `0be466359c29b8019c129be3a02e958ccc8768c8`

Fourth at-tier review of this PR, on the whole diff at this head, after the merge of `origin/main` at `74eaab8614`. Every claim below was measured by me — on a detached worktree at the head sha (`pnpm install --frozen-lockfile` 0; `pnpm --filter '@objectstack/metadata-protocol^...' build` 0; package build 0), on a second detached worktree at the published tag, or by `git show` / `git diff` / `git merge-tree` against refs — never taken from the PR body, the dev reports (`5689830955`, `5690332291`) or a previous record. Board readings carry their UTC time; tree readings carry their ref. NOT MEASURED means exactly that.

### ① Derived judgments

**Head, base, merge.** `GET /pulls/18231` (00:52Z): head `0be466359c`, base `74eaab8614`, draft true, 9 commits, 7 files, +817/−8 against the base. merge-base(head, `origin/main` = `500c1b5695`) is `74eaab8614`, so the base sha is the merge base. The merge commit `16ed5a4039` has parents `465a38ece9` and `74eaab8614`; `git merge-tree --write-tree 465a38ece9 74eaab8614` reproduces its tree `ad016212cb` exactly, and the patch-ids agree in both directions (diff `465a38ece9..16ed5a4039` equals diff `1bdbf82cb5..74eaab8614`; diff `74eaab8614..16ed5a4039` equals diff `1bdbf82cb5..465a38ece9`). The merge commit carries no edit of its own, no conflict resolution, and no rebase — all six pre-merge shas from `05b02846b1` to `465a38ece9` are still first-parent ancestors of the head. Round-4 delta `16ed5a4039..head`: 2 files, +10/−7; changed lines carrying `expect(`: 0 (control: 2 such lines across the whole `base..head` test diff are the two in-place rewrites already judged in review 1).

**J1 — Accept set of the served payload: unchanged for every key, at 80 nodes now.** My own census — predicate and the emitter's arm rule re-spelled in my own code, not the PR's, over the real `getMetaTypes()` on the pin's stub engine — at head: 27 types, 26 served with a schema. Pre-strip on the arm the emitter serves: **80 nodes across 16 types** (action 3, agent 2, api 1, app 8, book 1, dashboard 9, field 1, flow 4, hook 1, job 2, object 9, page 1, permission 3, report 2, skill 1, view 32). Output-arm-only: **77 across 15**; `action` is the only type on the `io: 'input'` retry and the 3-node gap is exactly its `execute` / `shortcut` / `bulkEnabled`. Post-strip served: **0 across 0**. Parallel walk of each served document against its derivation arm: 26/26 pure deletions, 0 unexplained, 0 over-dropped, and the removed set equals the pre-strip set for every type. All 80 parents carry `additionalProperties: false` (80 false / 0 absent / 0 other); 0 of the 80 are named in a `required` array; the key set of all 80 is exactly `description` + `not`; 0 carry `title`; 80/80 descriptions begin `[REMOVED] `. So every removed key was refused before (by `not`) and is refused after (as an additional property): no served node admits anything it did not admit, and none refuses anything it admitted. Review 2's J1 holds on the merged tree, with the number moved from 77 to 80.

**J2 — Public surface: unchanged.** `packages/metadata-protocol/src/index.ts` re-exports nothing from `unauthorable-nodes.js` (0 hits); built `dist/index.d.ts` has 0 occurrences of `acceptsNothing` or `stripUnauthorableProperties` (control: `omitInternalFieldsFromWriteResponse` 1). `package.json` is not in the diff. Nothing published moves.

**J3 — `acceptsNothing` and the position-aware walk, re-judged on the merged tree.** The three PR pin files at head: 3 files, 35 tests, green. Ablation A — `unauthorable-nodes.ts` restored on disk to the review-1 blob `e4695f4af1` (the walk that failed review 1), hash verified — exactly the three position and keyword-name-collision unit pins go red (3 failed / 14 passed across the unit file and the served pin); restored to blob `9ec8cfe07c`, porcelain empty. Reachability of the two widening classes on the merged served corpus, measured: 0 `{ not: {} }` nodes in keyword position in any served document; 0 served `properties` / `$defs` maps with an entry named `properties`; 0 required tombstones; 0 `dependentRequired`. The merge brought no new surface of either class, so review 2's J2, J3 and J3b conclusions stand unchanged; the unit pins remain the only carrier of the review-1 class, and they still go red on the broken walk.

**J4 — The over-drop guard's ledger, as edited this round: tightened to the new truth, not loosened.** The only assertion edited is the `dashboard` non-vacuity ledger, which gains one path — `dashboard.properties.widgets.items.properties.chartConfig.properties.aria` — inside the same sorted `toEqual` on the full list; `unexplained toEqual []` and `overDropped toEqual []` are untouched. Reproduced the CI red before judging the fix: the pin file restored on disk to the merge commit's blob `9885442d1f` (hash verified) fails on this tree with 1 failed / 7 passed, `AssertionError` at `:287`, the one extra received entry being exactly that `chartConfig.properties.aria` path — the same file and line the dev names, and consistent with the failing check run `104610594395` on `465a38ece9` (annotations: `pnpm run test` in `packages/metadata-protocol` exited 1; control on the base `74eaab8614`: all Test Core shards success). Then mut4b re-run on the merged tree — `|| key === 'maxTokens'` added to the drop condition, blob `928e48116e` — the guard goes red naming `agent.properties.model.properties.maxTokens` (1 failed / 25 passed), while `protocol.meta-types-degenerate-derivation.test.ts` stays green under the same mutation. Restored, porcelain empty. The guard still catches an over-drop of a live node; the red it replaced was the guard reporting a real new removal, which is what a non-vacuity ledger is for.

**J5 — Did the merge change any served accept set beyond what #17751 and #18303 changed? No.** `protocol.ts` is the same blob `ddf3d433d0` at `1bdbf82cb5` and at `74eaab8614`, and 0 files under `packages/metadata-protocol` moved on main between them, so served-at-head is exactly strip(served-at-`74eaab8614`), and J1 shows that strip is the 80 deletions and nothing else. Census delta against review 3's 77/15 at `1bdbf82cb5`: `dashboard` 8 to 9 — `ChartConfigSchema.aria` (`packages/spec/src/ui/chart.zod.ts:671`, retired by `2bf6ef18dc`, PR #18300, `Fixes #17751`) — and `report` 0 to 2 — `report.properties.chart.properties.aria` and `report.properties.blocks.items.properties.chart.properties.aria`, since `ReportChartSchema` is `ChartConfigSchema.extend(...)`. Every other type is unchanged. #18303 (`74eaab8614` itself) retires kernel startup defs (`PluginStartupResult` `startTime` / `plugin` / `health`, `HealthStatus`, `StartupOptions`, `StartupOrchestrationResult`), none of which is a served metadata type: 0 census movement. `.objectui-sha` is `53ded82b` at `1bdbf82cb5`, `74eaab8614` and head.

**J6 — The degenerate-derivation pin on the merged tree.** All 13 `CARD_PROPERTY_COUNTS` controls green (`report` 21 is unchanged because its two new tombstones are nested, not top-level); `action` served 45 + `retiredTopLevelCount` 3 = 48, green. The constant 48 was not edited, as fenced.

**J7 — os-regen hygiene, measured against the tree rather than taken from the report.** `.gitattributes` at head routes 19 globs to `merge=os-regen`; `git diff 74eaab8614 0be466359c` over every one of them is empty. `packages/spec`, `packages/metadata-core` and `packages/metadata` are byte-identical between `74eaab8614` and head. So every generated artefact at head is main's own, which main's CI validated (`74eaab8614`: 99 check runs, 69 success, 30 skipped, 0 failure). The dev's `gen:schema` then `git status --porcelain` empty claim is consistent with that structure — NOT re-run by me.

**J8 — Gates on the head tree, exact commands and exits.** `pnpm --filter @objectstack/metadata-protocol typecheck` 0; eslint `--no-inline-config` over the five changed `.ts` files 0; `scripts/check-empty-changeset.mjs --base 74eaab8614` 0 (1 declaring changeset added, none from the base modified); `scripts/check-changeset-no-major.mjs --base 74eaab8614` 0 (LEVEL AXIS not applicable locally, no PR payload); `scripts/check-adr-0087-registration.mjs --base 74eaab8614` 0 ("adds no declared-breaking changeset"); `check-engine-double-contract.mjs --self-test` 0 and the run 0; `check-cross-package-test-inputs.mjs` 0; `check-test-source-alias.mjs` 0. Full package suite `pnpm --filter @objectstack/metadata-protocol test` on the clean head tree: exit **0** — `Test Files 178 passed | 3 skipped (181)`, `Tests 2548 passed | 19 skipped (2567)`, 267 s — identical to the counts reviews 2 and 3 measured at `21edb645e9` and `0f89189631`; a one-path ledger edit and a two-figure prose edit moved no count, the skips are pre-existing (0 `.skip` lines added in `base..head`), and the dev's 178 / 2548 claim is now a measurement. Locally reproduced red on the pre-fix ledger (J4) and green after it is the same one-test move CI reported (1 failed / 177 passed on `465a38ece9`).. Assertion hygiene `base..head` over `*.test.ts`: 2 deleted lines carry `expect(` (the two in-place rewrites), 45 added; 0 `.skip/.only/.todo/xit/xdescribe` added; 0 loosened matchers (`expect.anything`, `expect.any(`, `toBeTruthy()`, `not.toThrow`); control characters in the 7 changed files: 0.

### ② Semver level

Declared `'@objectstack/metadata-protocol': minor`, one package, no BREAKING banner, no ADR-0087 marker — unchanged across all four rounds; `check-adr-0087-registration` 0. Correct under the convention (Clause-② yes floors the diff at `minor`; `major` is refused in the launch window) and consistent with plain semver: no exported symbol changes (J2), no served node's accept set moves (J1); what moves is the key set of a published payload — `minor` is the conservative reading, `patch` would have been defensible, `major` is wrong. No disposition owed.

**Does the changeset state what it does?** Yes. Every mechanism sentence was verified in reviews 1–3 and is byte-identical since `21edb645e9` except the census paragraph, which now reads: measured at `74eaab8614`, "this change's merge base (`@objectstack/spec` SOURCE at 17.4.0, plus the retirements unreleased at that sha — not the published release): 80 such nodes across 16 types — a reading taken at that tree, not a standing invariant". Checked clause by clause: merge-base(head, `origin/main`) is `74eaab8614` — true; `packages/spec/package.json` says `17.4.0` at `74eaab8614` and at head — true; the published tag `@objectstack/spec@17.4.0` is commit `7e6337007f`, an ancestor of `74eaab8614` by 624 commits (`GET /compare/7e6337007f...74eaab8614` status `ahead`, `behind_by` 0 — read over REST because this checkout is shallow and local ancestry is not trusted), so "retirements unreleased at that sha" is coherent and `@objectstack/metadata-protocol` is also `17.4.0` at all three points; the census at that tree is 80/16 by my own instrument (J1). A reader can check out `74eaab8614`, build `@objectstack/spec`, derive each registered type on the emitter's arm and count — the sentence is now re-derivable, and its version qualification is true.

**The re-anchoring decision, judged: ENDORSED.** Re-anchoring to `74eaab8614` with 80/16 was right and keeping `1bdbf82cb5` with 77/15 would have been wrong, for the dev's two reasons and one more. (a) The sentence's own parenthetical calls the sha "this change's merge base"; after the merge that is `74eaab8614`, so the old sha would have made the sentence false about its own anchor. (b) Only a main-line sha survives a squash landing as something a CHANGELOG reader can check out; `16ed5a4039` and `465a38ece9` would not. (c) The figure at the new anchor is TRUE (J1), which is the whole point of anchoring — review 3 failed this PR because the anchor pointed at inputs the number was not measured on, and this sentence now points at exactly the inputs it was measured on. The same sha and figures appear in the pin file's header, so the shipped test prose and the shipping changeset agree.

### ③ Boundary flags

- **B1 — `mcp_calls: 1`, `mcp__github__get_job_logs`.** A READ. `.claude/settings.json` `permissions.deny` lists 14 `mcp__github__*` tools, every one a write (`issue_write`, `create_pull_request`, `add_issue_comment`, `add_comment_to_pending_review`, `add_reply_to_pull_request_comment`, `pull_request_review_write`, `push_files`, `create_or_update_file`, `delete_file`, `create_branch`, `sub_issue_write`, `merge_pull_request`, `create_repository`, `fork_repository`); `get_job_logs` is not among them, and `rest-channel.md`'s own refusal names only `issue_write`. Confirmed a read, not on the roster; no refusal ground.
- **B2 — Out-of-scope finding 1 (hand-maintained header counts): AGREE.** The pin's header and the control's title carry numbers the control does not assert — it pins only total greater than zero and `dashboard` present — so they go stale silently, and this is the second hand re-measure in two days. A derived assertion (served total equals the output-arm total plus `action`'s input-arm count, both computed) would remove the maintenance. Outside this round's fence; the numbers as written are TRUE at head (80/16 served, 77/15 output-arm — J1); not a FAIL ground.
- **B3 — Out-of-scope finding 2 (`rest-channel.md` has no read-side row for Actions job logs): AGREE IN PART.** The read side indeed stops at `commits/{sha}/check-runs` and `actions/runs`. But `platform-readings.md` lines 291–295 already record the wall (`GET /actions/jobs/{id}/logs` refused by the egress proxy, CONNECT 403), that `get_job_logs` returns only the teardown tail, the check-run annotations endpoint as the cheap second read (exit code and failing command — what I read here), and the `get_workflow_run_logs_url` archive route. The remedy is a pointer row in `rest-channel.md`, not a new investigation. One reading differs from the record: the dev reports `get_job_logs` DID return the assertion text at file and line this time, against the "teardown tail only" row — a reading for the owner of `platform-readings.md`, not adjudicated here.
- **B4 — 68/13 versus 71/14, ruled.** Measured by me on a detached worktree at the tag commit `7e6337007f` (own install, `@objectstack/spec` built there, my own script, both rules over the same source): under the tag's own emitter rule — `toJsonSchemaSafe` at the tag is a bare cache plus `z.toJSONSchema` with no degeneracy retry, and the call site's `?? HAND_CRAFTED_SCHEMAS` arm is never reached because the husk is truthy — **68 across 13**, with `action` serving `{"$schema"}` and contributing 0; under today's emitter rule (retry `io: 'input'` when the output arm is degenerate) over the same source — **71 across 14**, `action`'s input arm declaring 47 properties, 3 of them tombstones. Both figures are true; they answer different questions. Review 3's B2 "71/14" is what today's emitter yields over the 17.4.0 SOURCE; "what release 17.4.0 actually SERVED" (both packages at `17.4.0`, same commit) is 68/13 — the dev is right about the served figure, and review 3's is right about the source. Either way the release does not produce 77/15 or 80/16, which is the point review 3 made. Any tag number written into the PR body must say which question it answers: 68/13 for "served at 17.4.0" (with the husk caveat), or 71/14 labelled "today's emitter over the 17.4.0 source"; an unlabelled 71/14 presented as what 17.4.0 served would be false by three. The changeset, correctly, carries no tag number.
- **B5 — Review 3's B3 is still open, and the body is now stale by a second step.** The PR body still says "77 nodes across 15 types, measured over this repo's own served registry", "on all 77 nodes", "0 of the 77 required", and its verification table is dated `21edb645e9`. Against the head those are 80, 16 and 0 of 80. Non-shipping prose (the body is not the changeset); when the seat writes the B3 fix it should carry 80/16 anchored to `74eaab8614`, and any tag number per B4. Held to review 3's bar, which failed only shipping prose: not a FAIL ground.
- **B6 — Scope fence honoured.** The file list is the seven paths in J1's diff; 0 under `stack.zod.ts`, `stack-*.test.ts`, `data/analytics.zod.ts`, `ui/view.zod.ts`, `api/discovery.zod.ts`, `meta-spelling/`, or any objectui path. Note that main itself moved `analytics.zod.ts`, `view.zod.ts`, `stack.zod.ts` and `discovery.zod.ts` between the two merge bases — those arrive through the merge, not through this branch, whose diff against `74eaab8614` is the seven paths.
- **B7 — Ratchet.** `scripts/engine-double-contract.pinned.json` +15 is the same three registration rows as at `9eaf3c08a5`; no pre-existing count moved; gate 0.
- **B8 — objectui.** No objectui file touched; `.objectui-sha` unchanged at all three refs (J5); Console Pin Gate skipped by its path filter on this head — a skip, not a pass. The changeset's consumer class and its three mechanisms are byte-identical since review 2 verified them at the pin and are not re-read this round.
- **B9 — CI on the head at posting time.** `GET /commits/0be466359c/check-runs` (01:09:15Z): 34 runs — 31 success, 3 skipped, 0 failure, 0 in progress. Skipped: Console Pin Gate, Build Docs, Packed-tarball smoke (opt-in) — path or label filters this file list does not select; a skip is not a pass and none is reported as one. `Test Core (2/6)`, the shard that was red on `465a38ece9`, is `success` on this head. Landing is the seat's step, not this record's.
- **B10 — Not measured, by name:** `gen:schema` / `check:generated` (J7's structural argument stands in for the run); the derived 71-command gate family (`dispatch-gates.mjs` not run — the families I ran are the ones in J8); browser rendering at the pin; the job log text itself (annotations plus the local reproduction in J4 instead).

Implemented-by: `claude/issue-17502-repeater-row-tombstone-columns`

**VERDICT: PASS**


---

Issue #13042, opened 2026-08-28
[worklist] Re-close the tool.json liveness ledger — every citation points into the cloud runtime's service-ai, which this repo's checkout does not contain


os-trump:
Spun out of #13003 by the `domain:spec` seat (session `session_01LpRNHxWZgSUgVnFT9mQQo4`) per ruling comment 5458356183 and ACCEPT 5458682782, so the anchor-adoption worklist can close honestly on the framework-reachable population. Dedup: targeted semantic search returned a populated 16-result set with no open card covering tool.json re-closure — no duplicate exists.

## Why this is its own card

`packages/spec/liveness/tool.json`'s six `live` entries all cite `packages/services/service-ai/…` — the closed cloud runtime. Measured during #13003 batch 2 (PR #13038): that directory does not exist in this repo's checkout at all (`packages/services/` holds 16 members, none of them `service-ai`), and the evidence scanner classifies the prefix as foreign, so the file has **zero local citations and zero anchorable pointers** — the `path#symbol` grammar cannot reach it by design, and no batch running in a framework-only container can re-close it without manufacturing false confidence. All six entries are undated (`verifiedAt` absent) with no honest way to date them from here.

## The work (needs a reader with the cloud repo checked out)

1. Re-close each of the six entries against the cloud repo's actual service-ai code — read the consumer, confirm the claim, stamp `verifiedAt`; where the gate's guidance applies, re-attribute with a realm marker or re-classify under ADR-0049 instead.
2. Repair the file's `_note`: it claims the framework's own service-ai tree "is a stale build artifact with no `src/`" — measured in batch 2, that tree is now **absent outright**, so the note is stale twice over.

The fix lands in this repo (the ledger file is `packages/spec/liveness/tool.json`), so the card lives here; the executor requirement is the cloud checkout, not a different landing repo. Routing suggestion for triage: `domain:spec` (the ledger is spec governance data); the dispatching seat must pick a container/dev with cloud repo access.

Refs: #13003 (worklist; batch-2 report documents the measurements) · PR #13038 (the batch that established the facts) · #12516 (anchor grammar).

huangyiirene:
## 分诊：补齐 `domain:spec` + `priority:p2` + type `Task`

采纳卡面的路由建议。落点是 `packages/spec/liveness/tool.json`，与整个 liveness 家族（#13003 · #13036 · #13040 · #13041 · #13043）一致。

### ⚠️ 派发席必读：执行约束不是路由约束

[...]

PR #18313, opened 2026-09-15
fix(spec): the 21 confirmed cloud citations in the agent/skill/action liveness ledgers carry a read date and a symbol anchor

### LINKED ISSUE #13272: Three liveness ledgers still cite `packages/services/service-ai/…`, a path that exists in NEITHER repo, and repeat the falsified "stale build artifact with no src/" note
Measured while re-closing `tool.json` against a live cloud checkout (refs #13042). That card's fence is `packages/spec/liveness/tool.json` only, so the identical defect in its three siblings is filed rather than fixed.

## Census at `74049254`

Ledger entries whose `evidence` / `producer` cites the prefix `packages/services/service-ai/`:

| ledger | entries citing the stale prefix | `_note` repeats the falsified sentence |
|---|---|---|
| `agent.json` | 11 | yes |
| `skill.json` | 8 | yes |
| `action.json` | 3 | yes |
| ~~`tool.json`~~ | ~~5~~ (repaired in #13042) | repaired |

22 entry citations remain after #13042 lands.

## Both halves of the claim are false, and each is false in its own way

**1. The framework tree is absent, not stale.** All four notes say the framework's own service-ai tree "is a stale build artifact with no `src/`". Measured in this checkout: `git ls-files | grep -ic service-ai` returns `0`, `find . -name service-ai` returns nothing, and `packages/services/` holds 16 members, none of them service-ai. There is no artifact to be stale.

**2. The cited path is wrong for the cloud repo too.** Cloud's real layout (measured at cloud `origin/main@15f55df`) is `packages/service-ai/…` and `packages/service-ai-studio/…` — never `packages/services/service-ai/…`. So the 22 citations name a path that exists in neither repository. They are unfalsifiable pointers, which is exactly the condition the ledger's own README says a `live` verdict must not rest on.

## The mechanical trap a repair will hit

`packages/spec/scripts/liveness/evidence.mts` hardcodes

```js
export const FOREIGN_PATH_PREFIXES = ['packages/services/service-ai/'];
```

so the **stale** spelling is silently treated as foreign and never resolved — which is why 22 dead pointers have sat green. The **real** cloud path `packages/service-ai/…` is repo-rooted in shape and is NOT in that list, so a naive repoint resolves it as LOCAL and fails CI.

Every repointed citation therefore MUST carry the `cloud` realm marker (`cloud @<sha>: packages/service-ai/…`). This is not optional prose — it is the difference between a green gate and a red one. #13042 hit this and its `_note` now records it.

## Suggested shape of the work

Same as #13042, once per ledger: read each cited consumer against a real cloud checkout, confirm or falsify, stamp `verifiedAt`, repoint with the `cloud` realm marker plus a pinned cloud SHA and a `#symbol` anchor, and repair the `_note`. Each of the three is independently doable.

**Executor constraint (the same one that parked #13042 for a day): a container with the `cloud` repo checked out.** A framework-only container cannot close any of these 22 without manufacturing false confidence — it can only re-spell the path.

Whether `FOREIGN_PATH_PREFIXES` should also gain `packages/service-ai/` is a separate judgement: doing so would make an unmarked citation of the real path pass silently, which trades one unfalsifiable spelling for another. #13042 chose the realm marker instead and did not touch the constant.

hotlong:
Fixes #13272

The re-verification half of this card. #13309 repointed the 22 citations off `packages/services/service-ai/`, a path present in neither repository, but could not confirm a single one of them — no session on that board had a cloud checkout. That reading has since been taken and published on the card as comment 5682851335, against cloud `cb8ee7ff60c097cc21a584fe9caf8ef4391cc0e8`: 21 confirmed, 1 falsified, 3 citations wrong in a way this repo's gate cannot see. This PR transcribes it.

## What changes

Every confirmed row in `packages/spec/liveness/{agent,skill,action}.json` now carries `verifiedAt: "2026-09-15"`, `evidenceScope: "cross-repo"`, and a `#symbol` anchor on the consuming function instead of a package-root path or a line number.

- **A symbol instead of a line, because a line rots in range.** Three cited lines had already drifted onto unrelated prose — `agent-runtime.ts:264`, `agent-access.ts:50`, `action-tools.ts:535` — while every mechanical check kept passing. A symbol moves with the consumer and goes red when the consumer is renamed or deleted.
- **The framework half is now gate-checked.** Five skill rows anchor `packages/mcp/src/skill-prompts.ts#projectSkillPrompt`. The `;` before that token ends the `cloud` realm's scope, so the anchor is repo-local and `checkEvidenceAnchors` resolves it on every run — where the old parenthesised `(projectSkillPrompt)` was prose no check read. The `instructions` row used to cite `(skillPromptResult)`, one hop too far downstream: that function consumes the projected value, while the record's own key is read in `projectSkillPrompt`.
- **The eight skill rows' `verifiedAt: "2026-08-06"` is replaced, not kept.** It was stamped while the citation still named a path existing in neither repo — false confidence the 180-day clock in `verification.mts` is structurally unable to see: 40 days old, never flagged, resting on nothing.
- ****Four** ledger assertions the reading falsified are repaired in the prose — two in the first commit, two more in the third; see the commit log below.** `agent.role` was noted as *"persona → system prompt."*: it reaches `AgentSummary` through `listAgents` and nothing else — `buildSystemMessages` pushes `instructions`, the date block, the UI context and the skills block, and never `role`. `agent.planning` was cited at `agent-runtime.ts`, which does not read the key at all; its three readers are `routes/agent-routes.ts`, `routes/assistant-routes.ts` and `eval/eval-runner.ts`. The note's other claim — that only `planning.maxIterations` survives — is confirmed and stays. Every refutation sentence #13309 added is kept; this adds to that record.

## What is deliberately NOT here

- **`agent.tools` is untouched** — no stamp, no scope, no `status` change, no prose change. The same reading falsified it: zero consumers in cloud at that ref, where the only two mentions are comments recording the removal of the branch, while this repo's own `AgentSchema` already declares the key `retiredKey(...)`. That is a liveness re-grade, which belongs to triage and is filed as objectstack#18304; a `verifiedAt` there would certify the wrong thing. Note the shape — the row sat `live` since the 2026-06 audit *because* `FOREIGN_PATH_PREFIXES` exempted its citation from resolution, so the exemption this card was filed about had hidden a dead key, not only a misspelled path.
- **`FOREIGN_PATH_PREFIXES` is untouched.** Triage ruled that out on this card, and the census on the reading found it now matches zero `evidence` fields anyway. An observation for a separate card, not a rider here.
- **`packages/spec/liveness/tool.json` is untouched** — it is #13042's, and was read here only as the pattern authority for row shape.

`Fixes` rather than `Part of`: the card's remaining scope was *read the 22, stamp `verifiedAt`, add `#symbol` anchors*. The reading is done and published, the stamping is done for all 21 rows a stamp can honestly carry, and the 22nd is not an unfinished half of this card but a different question on a card of its own. The title says 21 rather than the card's 22 for the same reason.

## Verification

`pnpm --filter @objectstack/spec check:liveness` — **exit 0** at `9a4a6d96`, and its counters move exactly where the change predicts. Against the same command on the pristine base `53dd5aa`:

| counter | base `53dd5aa` | `9a4a6d96` | why |
|---|---|---|---|
| repo-local evidence paths | 505 declared, 505 resolved | 505 declared, 505 resolved | **unchanged, as predicted** — the `packages/mcp/…` clause was already local before this edit, because the `;` already reset the realm |
| attributed to another repo | 204 | 207 | `agent.active` and `agent.planning` now cite the routes and eval files their readers actually live in |
| symbol anchors | 578 written, 578 resolving | 583 written, 583 resolving | the five skill rows anchored at `#projectSkillPrompt` |
| entries carrying `verifiedAt` | 516 | 529 | +13: 10 agent rows and 3 action rows; the 8 skill rows were already dated and are re-dated in place |
| `evidenceScope` declared | 90 cross-repo, 608 undeclared | 111 cross-repo, 587 undeclared | +21 |

**A discriminating control, because green alone proves nothing about the new anchors.** One framework anchor was corrupted in place (`skill.name` → `#projectSkillPromptX`), and the mutation was proven to reach disk before the gate was asked — blob hash `50e2209d` to `9d5418dd`, and the occurrence count for `projectSkillPromptX` moved 0 to 1:

```
MUTATED VERDICT command-exit = 1
symbol anchors: 583 pointer(s) written `path#symbol`, 582 naming a symbol the cited file contains, 1 UNRESOLVED.
✗ 1 anchored citation(s) name a symbol the cited file does not contain:
    skill/name → packages/mcp/src/skill-prompts.ts#projectSkillPromptX
```

Restored with `git checkout HEAD -- …` and proven byte-identical: hash back to `50e2209de6f647e845ca8c6c1e0330a5d98bd2c3`, `git diff HEAD` empty, `git status --porcelain` empty. The gate can see these anchors and can fail on them. The cloud anchors are a different matter and always will be: `scanEvidence` collects no anchor while a foreign realm marker is in force, so CI can re-derive none of them — which is why the dated reading is what carries them, and why a bare re-stamp without re-reading cloud is ruled out in the `_note` of all three ledgers.

Other evidence:

- **Liveness suite** — `packages/spec/scripts/liveness/*.test.ts`, 9 files, **229 tests passed** (`evidence.test.ts` walks the real ledger directory).
- **The ledger's cross-package consumer** — `packages/lint` reads `@objectstack/spec/liveness` at rule time; its two ledger-reading suites are green, **75 tests passed**.
- **`pnpm --filter @objectstack/spec typecheck`** — exit 0.
- **Derived gate families** — `scripts/pm/dispatch-gates.mjs --ran` accounts for **59 derived families: 57 run green, 2 NOT MEASURED** (`check:dual-build-cjs-loads` and `check:lean-entry-closure` both exit 3, `PREREQUISITE NOT MET`, because they load built output for the whole workspace; CI's Build Core covers them). Included and green: `check:published-files`, `check:platform-checklist` (it derives its kinds from this very directory), `check:empty-state`, `check:generated`'s changeset family, `check:nul-bytes`.
- **Repo-wide `pnpm lint` is CI's run, and the narrowing here is measured rather than assumed.** eslint's own configured population is `**/*.{ts,tsx,mts,cts,js,jsx,mjs,cjs}`, every later block a subset of those extensions; run over the four changed paths with `--format json` it reports 4 files, 0 errors, and one message each: *"File ignored because no matching configuration was supplied."* Three JSON files and one Markdown file are not in the linted set at all. The single occurrence of `parserOptions.project` in `eslint.config.mjs` is a comment stating that typed linting is not enabled, so no rule reads across files and this diff cannot move the verdict on a file it did not touch.

A `patch` changeset ships with it: `packages/spec/liveness/` is listed in that package's `files[]`, so these ledgers are published inside the npm tarball and the pointers an upgrading reader follows are exactly the ones this PR dates.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TAUTP6Yky8QWoHUAPDKNJQ)_

---

## The three commits, because the diff grew after the first review

The PM seat reviewed the first commit against the published reading rather than against this description, and found two gaps. Both are recorded here so the diff and the prose agree.

**1 — `9a4a6d96`** the transcription described above: 21 rows dated, scoped and anchored; `agent.tools` left alone; the `role` and `planning` notes repaired.

**2 — `97b17e9e`** four citations that still named a whole file are anchored at their enclosing symbols: `agent.planning`'s three readers at `routes/agent-routes.ts#buildAgentRoutes`, `routes/assistant-routes.ts#buildAssistantRoutes` and `eval/eval-runner.ts#EvalRunner`, and `agent.active`'s route clause at `#buildAgentRoutes`. The dispatch table that drove commit 1 had left these four unanchored; that was the PM's omission, not a transcription error, and the anchors were read in a cloud checkout before being written.

**3 — `bdf441e0`** the `agent.access` and `agent.permissions` notes stop describing the pre-protocol-17 world. Three false claims across the two rows, measured at cloud `cb8ee7ff` and published as card comment 5683627647: the match is by a held **position**, never by `role` (the alias framework #6011 removed, after which `user.roles ?? []` degraded to the empty set — the cloud#1442 outage); the cited `agent-routes.ts:151` is now an unrelated ADR-0013 guardrail while the call sits at `:494`; and `evaluateAgentAccess` has **two** enforcement points, the 403 chat route and `agent-runtime.ts#listAgents`, which drops an inaccessible agent from the catalog **silently**. Both notes keep their old wording as dated refutation prose.

⛔ Neither later commit touches a `status`, a `verifiedAt`, an `evidenceScope` or an `evidence` string: commit 3 is `+2 -2` in one file, proved per row against the parent blob, and commits 2 and 3 each move the `check:liveness` counters by exactly nothing — a foreign anchor is stripped before classification and a note is never scanned.

## A CI reading, so the red on the two superseded heads is not re-diagnosed

`TypeScript Type Check` went red on `9a4a6d96` and again on `97b17e9e`. Both are **supersede artifacts**. That job runs no compiler: it reads the four lanes named in `OS_AGGREGATOR_MEMBERS` and fails if any did not conclude `success`. On each superseded head the next push cancelled lanes still in flight, and the aggregator reported the cancellation exactly as designed. Base `main` is green on all five type-check contexts, so neither was an inherited failure. `97b17e9e` reached 23 success / 3 skipped / zero failures — including `Build Core` and `Governed Surface Queue Guard` — before commit 3 superseded it.

---


hotlong:
**The red `TypeScript Type Check` on `9a4a6d96` is a supersede artifact, not a type error.** Recorded so nobody chases it.

That job is an **aggregator**: it has no compiler of its own, it reads the four lanes it declares in `OS_AGGREGATOR_MEMBERS` and fails if any of them did not conclude `success`. Its log at 2026-09-15T15:59:36Z shows all four arriving as `cancelled`:

```
type-check lane `typecheck-consumers`    concluded `cancelled` -- expected `success`
type-check lane `typecheck-debt`         concluded `cancelled` -- expected `success`
type-check lane `typecheck-source-gates` concluded `cancelled` -- expected `success`
type-check lane `typecheck-workspace`    concluded `cancelled` -- expected `success`
check-typecheck-lanes: FAILED (4 problem(s) above).
```

They were cancelled because the next commit (`97b17e9e`, the four remaining symbol anchors) superseded that run — 17 of `9a4a6d96`'s runs went `cancelled` in the same instant. ⇒ The aggregator reported exactly what it saw and was right to; ⛔ there is no type error in this diff to fix, and a re-run of the superseded head would be spending a re-run on a commit that is no longer the head.

Ruled out first, ⛔ not assumed: the same check is **green on the base** (`main` `53dd5aaa`, all five type-check contexts `success`), so this was never an inherited failure either.

**Current head `97b17e9e`: no failures.** 12 green, 17 still running, and the two gates this diff actually exercises are already through — `Spec property liveness` ✅ and `Check Changeset` ✅. This PR stays a draft and is ⛔ not enqueued until every check is green.

On the **Docs Drift Check** advisory in the comment above: the three changed files are liveness ledger *data*, so yielding no documentable anchor is the expected outcome rather than a gap. The page that documents this ledger — `packages/spec/liveness/README.md` — is not free prose but a generated, reconciled state table (`scripts/liveness/readme-table.mts`), and no row's `status` changed in this diff, so the table and `state-counts.md` are unmoved; `Spec property liveness` is the check that would have said otherwise, and it is green.


---

Issue #17469, opened 2026-09-10
driver-sql `isJsonField` and spec `isMultiValueField` disagree for `master_detail` / `tree` / `text` + `multiple: true`, so a consumer using the spec predicate to shape a query gets a 400


os-litant:
Filed by the triage seat, as the upstream card objectui#8937 records as **owed and not filed**.

## The divergence, measured on `origin/main`

| predicate | site | rule |
|---|---|---|
| driver storage | `packages/drivers/driver-sql/src/sql-driver.ts:16896` | `isJsonField = JSON_COLUMN_TYPES.has(type) \|\| !!field.multiple` — `multiple` truthy on **any** type ⇒ JSON column ⇒ `=` refused |
| spec | `isMultiValueField` (`@objectstack/spec/data`) | `MULTI_OPTION_TYPES.has(type) \|\| (MULTI_CAPABLE_TYPES.has(type) && multiple === true)` |

⇒ They disagree for **`master_detail` / `tree` / `text` with `multiple: true`**: the spec predicate says "not multi-value", the driver stores it as a JSON column and **refuses `=` with a 400**.

## Why it matters — a consumer is already on the wrong side of it

objectui's `RelatedList` uses the **spec** predicate to decide whether to compose `$contains` or `=`. For those three types it composes `=`, and the live driver answers **400**. objectui#8886's own test pins that behaviour — *"SPEC PREDICATE — `multiple: true` is INERT on a type outside `MULTI_CAPABLE_TYPES`"* — so the divergence is currently **pinned on the consumer side**.

⭐ **Not a regression.** Pre-#8886 the same shape sent `=` too. ⛔ Nobody should read this as caused by that PR; it surfaced there.

## The shape of the fix — enforce-or-align, ⛔ not chosen here

Someone has to say which predicate is authoritative for "can this field be compared with `=`":

- **align the driver** to the spec predicate ⇒ changes what storage the driver picks for those three types, which is a migration-shaped change on existing data;
- **align the spec predicate** to the driver ⇒ widens `isMultiValueField`, a published-contract change;
- **declare they answer different questions** — the spec predicate is about authoring semantics, the driver's about storage — and give consumers a storage-facing predicate to use instead. ⇒ Then objectui's use of `isMultiValueField` for query shaping is the defect, and the fix is documentation plus a new export.

⚠️ The third reading is the one that would make everybody currently correct, and it is the one nobody has written down.

## Lane

[...]

Issue #14829, opened 2026-09-03
[finding] packages/cli generate.ts: both migration generators ignore `multiple: true`, so a multi-valued field gets a scalar column while `os generate types` gives it an array type


os-trump:
Found while implementing #14657. Filed unassigned for triage. No severity asserted. #14657 is not addressed here and remains open until its own PR lands.

## The shape

`multiple` appears exactly four times in `packages/cli/src/commands/generate.ts`, and all four are on the TypeScript side:

```
562  function fieldTypeToTs(fieldType: string, multiple?: boolean): string {
564    return multiple ? `${base}[]` : base;
607      const tsType = fieldTypeToTs(fType, !!fieldDef.multiple);   // os generate types
831      const tsType = fieldTypeToTs(fType, !!fieldDef.multiple);   // os generate client
```

`generateMigrationSql` and `generateMigrationTs` never read it. `fieldTypeToSql(fieldType)` does not even take the parameter.

So for `Field.lookup({ reference: 'account', multiple: true })` the CLI emits, from one config in one run:

- `os generate types` — `account?: string[]`
- `os generate migration --format sql` — `VARCHAR(36)`
- `os generate migration` (typescript, the default) — `table.uuid('account')`

The runtime disagrees with the last two. `driver-sql`'s `isJsonField` is `JSON_COLUMN_TYPES.has(type) || !!field.multiple` — the `multiple` flag alone puts the column in the JSON class, and `sql-driver.ts` handles it "at the top of createColumn", before the per-type switch. So a `multiple: true` field of ANY multi-capable type (`MULTI_CAPABLE_TYPES` = select, radio, lookup, user, file, image) is a JSON column in the platform and a scalar column in the generated migration.

## Why it is a defect rather than a gap

It is silent in both directions: the scaffold looks right, the generated TypeScript is right, and only the column is wrong. The write path then serializes an array into a scalar column — the same class of failure `#field-zoo` recorded when array-valued fields reached the SQLite binder unserialized ("SQLite3 can only bind numbers, strings, bigints, buffers, and null"), except here it is the generated DDL rather than the driver that is out of step.

## Why it was not folded into #14657

[...]

PR #18389, opened 2026-09-16
fix(metadata): close out the adr-0030-notification-event retirement residue

os-support-ai:
Fixes #17281

Clause-②: no — this PR puts no new key on a published payload. It adds one changeset row, removes a dead workspace dependency (a narrowing, never a widening) and corrects stale prose; triage's grading `5692961190` states it directly: "nothing widens an accept set or a public surface". Declared by the dispatching seat in claim `5693760798`.

Residue from the `adr-0030-notification-event` retirement (#16194 / PR #17194, merged `d64bcb6377`). Four items, four commits, item 1 first and alone as the triage grading `5692961190` directed, then re-cut on the maintainer's ruling `5698965201`.

## Item 1 — the `platform-objects` release note (re-cut on ruling `5698965201`)

`797caf0c` first put `'@objectstack/platform-objects': minor` on the **existing**
`.changeset/retire-adr-0030-notification-event-migration.md`. `d2544585` undoes that and
takes route B instead. Two acts, one commit:

- **The base note is restored byte-for-byte.** Measured, not asserted:
  `git rev-parse 8b81ab60d:.changeset/retire-adr-0030-notification-event-migration.md` and
  `git hash-object` on the restored working-tree file both read
  `2ae652749b40dec974fdfb5fcf5fe0990267fe53`, and `git diff 8b81ab60d HEAD` on that path
  prints nothing. The file is no longer in this PR's diff at all.
- **The bump moves to a new, issue-scoped changeset** —
  `.changeset/17281-platform-objects-attest-fresh-datastore.md`, declaring exactly
  `'@objectstack/platform-objects': minor`, named the way the gate itself prescribes
  (`.changeset/{issue}-{slug}.md`).

The note is written for a release reader rather than for this PR: a datastore created from
empty now attests **two** creation-attested migration ids instead of three, because
`attestFreshDatastore` (`packages/platform-objects/src/system/migration-flag.ts`) defaults
to `CREATION_ATTESTED_MIGRATION_IDS` and that tuple lost `'adr-0030-notification-event'` in
the retirement. No row is written under the retired id any more; a deployment that already
holds one keeps it, and `NOTIFICATION_EVENT_MIGRATION_ID` survives as that row's name.

The bump is owed at all because that operator-visible half lives in
`@objectstack/platform-objects`, a package with no `private: true`, and the retirement's own
release note never declared it.

### The ruling, and its one declared cost

Maintainer ruling `5698965201` chose **B**: open a new changeset file for
[...]

PR #18392, opened 2026-09-16
fix(cli): the one-definition-of-multi-valued invariant reaches os generate — the CLI half of #17469

### LINKED ISSUE #18199: [finding] the one-definition-of-multi-valued invariant #17469 restored stops at the driver-sql package boundary — `os generate migration` still emits JSONB where the driver now emits varchar (#14829 in reverse)
**Filed by the `domain:spec` execution seat** (#6017), `session_01KB5PFtxuy1x3dcR5gxudx6`, 2026-09-14T14:1xZ, from the `out_of_scope_findings` of the #17469 round (PR #18187). ⛔ Not claimed, ⛔ not dispatched. Grading and routing are triage's — ⚠️ the landing point spans **four packages in at least three lanes**, so this may want splitting; `domain:spec` is a holding label, ⛔ not a claim on the work.

## What #17469 restored, and where it stops

#17469 (ruled, director batch #128 item 5, option 1′) established **one definition of "is this field multi-valued"**: `isMultiValueField` in `packages/spec`, with storage following it. The seat's implementation ruling extended ruled item 2 from the single site the ruling named to **all 18 sites** in `driver-sql` that read `field.multiple` as that question.

⇒ **The invariant stops at the `driver-sql` package boundary.** These sites still read `field.multiple` raw:

| site | what it decides |
|---|---|
| `packages/cli/src/commands/generate.ts:2617` | `if (fieldDef.multiple) { table.jsonb(...) }` — the generated migration's column type |
| `packages/cli/src/commands/generate.ts:2443` | `fieldTypeToSql(fType, !!fieldDef.multiple, ...)` |
| `packages/cli/src/commands/generate.ts:1706` | `declaredColumnDefault` — `if (declaring?.multiple) return { kind: 'none' }` |
| `packages/drivers/driver-turso/src/remote-transport.ts:2435` | `if (field.multiple) return 'TEXT'` — column type |
| `packages/objectql/src/engine.ts:4741` and `:13584` | value shape |
| `packages/runtime/src/action-execution.ts:1279` | parameter arity |

## The measurable consequence today

**`os generate migration` and the runtime driver now disagree** for the retired shapes: a `text` field flagged `multiple: true` gets **JSONB** from the generator and a **varchar** from `driver-sql`. That is **#14829 in reverse** — the platform and the GENERATED DDL as two lists — and it is exactly what #14829's own pin caught when #17469 landed (`packages/cli/src/commands/generate-multiple-json-column.pin.test.ts` went red and had to be re-targeted).

⚠️ `generate.ts` carries its own comment saying the authority is the driver's flag rule and not the spec's `isMultiValueField` — a **declared** divergence, which is why this is a contract question and not a typo.

## Why it is BOUNDED, stated so the card is not over-graded

Those declarations are **refused at the authoring entrance** by #17469's ruled item 1, so they can only still reach these readers through doors that never run `FieldSchema` (`registerExternalObject` / `initObjects`, and the generators' own unvalidated input). Reachable, but no longer authorable. This seat is ⛔ **not** grading it p1.

## Why #17469 did NOT fix it

Seat decision, recorded on #17469 (`5664249028`): the ruling's scope was the driver, `packages/cli` is another lane, and widening a 21-file cross-domain PR into a third domain mid-flight is not a call a dispatching seat makes on its own. Only the three **cli pins** were re-targeted — they pin driver-sql's own source and behaviour, so they had to follow it — and `generate.ts`'s logic was explicitly left alone.

## What a taker owes first

⚠️ Per-site, the same premise #17469 used, and it is falsifiable: **does this site read `field.multiple` to mean "is this value multi-valued"?** A site reading it for a different question is ⛔ not covered. That premise held for 18 driver-sql sites and **failed for 3** (message interpolations that echo an author's declaration rather than deciding anything) — so expect the same split here. ⛔ Do not align a site because it matches a grep.

## Dedupe words

`field.multiple` raw read · `isMultiValueField` · `os generate migration` JSONB · 14829 reverse · one definition multi-valued

⛔ This seat did not run a dedupe search (per the filer-does-not-dedupe rule). ⚠️ A check must include **closed** cards.

## Refs

#17469 (the ruling and the driver-sql half) · PR #18187 · #14829 (the original two-lists defect) · #11437 (the radio precedent)

---

os-support-ai:
Fixes #18199

Clause-②: no — this PR puts no new key on a published payload. It makes `os generate` follow the already-ruled `isMultiValueField` definition (#17469, director batch #128 item 5, option 1′) — pulling a diverged emitter back to a declared contract, which narrows rather than widens. Declared by the dispatching seat in claim `5693774731`.

The maintainer ruling of 2026-09-13 (director batch #128 item 5, option 1′) gives "is this field multi-valued" **one definition** — `isMultiValueField` in `packages/spec` — and storage follows it. #17469 landed the `driver-sql` half and stopped at the package boundary. This is the ruling card for the `packages/cli` + `packages/runtime` half; the `packages/objectql` (11 files) and `packages/drivers/driver-turso` (1) sites are a separate `domain:engine` child card and are **not** touched here.

## Per-site verdict

The premise is the card's own, and it is falsifiable per site: **does this site read `field.multiple` to mean "is this value multi-valued"?** It held for 18 of 21 sites in `driver-sql` and failed for 3. Here it holds for 5 and fails for 1.

| site (on `origin/main` @ `8b81ab60`) | reading | verdict |
|---|---|---|
| `packages/cli/src/commands/generate.ts:2617` `if (fieldDef.multiple) { table.jsonb(...) }` | decides the generated migration's column type | **covered** |
| `packages/cli/src/commands/generate.ts:2443` `fieldTypeToSql(fType, !!fieldDef.multiple, …)` | same decision, `--format sql` | **covered** |
| `packages/cli/src/commands/generate.ts:1706` `declaredColumnDefault` — `if (declaring?.multiple) return { kind: 'none' }` | mirrors `createColumn`'s multi-value short-circuit, which returns before the DEFAULT line | **covered** |
| `packages/cli/src/commands/generate.ts:757` `fieldTypeToTs(fType, !!fieldDef.multiple)` — `os generate types` | decides whether the generated property type is an array — the ADR-0104 D1 value question verbatim | **covered** |
| `packages/cli/src/commands/generate.ts:1203` same call — `os generate client` | same | **covered** |
| `packages/runtime/src/action-execution.ts:1279` `multiple: p?.multiple ?? field?.multiple` | **inherits an author's declaration onto a descriptor; decides nothing** | **excluded** |

### Why the two `fieldTypeToTs` sites are in, although the card's table names only three

The card's table is illustrative — it names 2 of `objectql`'s 11 files too. The `#14829` pin in this package states the invariant as *"one authored `multiple: true` field, **three** surfaces, ONE answer"*: generated TS type, `--format sql`, `--format typescript`. Aligning the two migration surfaces and leaving `fieldTypeToTs` on the raw flag would have given a flagged `text` field `string[]` from `os generate types` and `TEXT` from the migration generated in the same run — #14829 re-created **inside** the CLI. Landing three of five sites was not available.

### Why `action-execution.ts:1279` is excluded — the reading

Both triage comments flagged it as a candidate for the same exclusion that removed 3 of 18 `driver-sql` sites, and the measurement agrees. `resolveDeclaredActionParams` copies `multiple` from the referenced field onto a `ResolvedActionParam`; the descriptor is then handed to `validateActionParams`, which builds the param's schema with `valueSchemaFor({ type, multiple, options })` — whose last line is

```
return isMultiValueField(def) ? z.array(element) : element;
```

So the runtime **already follows the one definition** at this seam, through the spec's own consumer, and the read here is the ADR-0104 D2 inheritance the function's docblock declares. Pre-computing the verdict into the field would also break that contract: the key would stop carrying what the author wrote, and the predicate would then be applied to its own output. The reading is recorded at the site, so the next grep does not align it.

## The declared divergence is gone

Three comments in the tree declared the losing authority, and the card is class (b) because of them. Each is rewritten rather than left standing:

- `generate.ts:1332` — "⛔ the DRIVER is the authority for which column exists, never the spec's `isMultiValueField` VALUE predicate". The first clause is still true; the second stopped being true when the driver's own multi-value half became that predicate.
- `generate.ts:2286` — "WARNING: this is deliberately NOT the spec's `isMultiValueField` … the column question belongs to the driver, and the driver's answer is the flag alone." The premise held and the conclusion inverted; the paragraph now records the reversal.
- `generate.ts:2615-2616` — "See `fieldTypeToSql` for why the authority is the driver's flag rule and not the spec's `isMultiValueField` value predicate."

One more, in a test: `generate-field-type-vocabulary.pin.test.ts:90` carried the same sentence.

## Measured — `origin/main` @ `8b81ab60` vs this branch

One config through all three generators, the same probe both times (the pre-fix leg was taken by checking `generate.ts` out at the base commit and restoring it to a byte-identical blob, `be774af8…`):

| declaration | `os generate types` | `--format sql` | `--format typescript` |
|---|---|---|---|
| `{ type: 'text', multiple: true }` | `string[]` → **`string`** | `JSONB` → **`TEXT`** | `table.jsonb` → **`table.text`** |
| `{ type: 'text', multiple: true, defaultValue: 'x' }` | — | `TEXT` with no DEFAULT → **`TEXT DEFAULT 'x'`** | → **`.defaultTo("x")`** |
| `{ type: 'multiselect', multiple: true }` | `string[][]` → **`string[]`** | `JSONB` (unchanged) | `table.jsonb` (unchanged) |
| `{ type: 'lookup', multiple: true }` | `string[]` | `JSONB` | `table.jsonb` — all unchanged |

The last two rows are the same raw read reaching two shapes the card did not name. `multiple: true` is accepted **redundantly** on the inherently-multi option types, so `FIELD_TYPE_MAP`'s own `string[]` entry was being wrapped a second time; and the short-circuit was withholding a DEFAULT that the driver emits. Both close with the same alignment.

## `generate-multiple-json-column.pin.test.ts` — re-targeted deliberately

It is #14829's pin, it went red when #17469 landed, and it pins the behaviour this PR changes. It was **not** skipped, disabled or quarantined. Before the re-target it failed exactly three arms, all on `text`, every `MULTI_CAPABLE_TYPES` member still green:

```
× control — the SAME type without the flag still gets its scalar column
    AssertionError: expected 'TEXT' not to be 'TEXT'
× text + multiple:true — array TS type AND a JSON column in both migrations
    AssertionError: expected 'string' to match /\[\]$/
× the flag decides before the type in the GENERATORS — a type outside MULTI_CAPABLE_TYPES too
    AssertionError: expected 'TEXT' to be 'JSONB'
Tests  3 failed | 12 passed (15)
```

That red is the narrowness measurement: the only declaration whose answer moves is the one where the flag and the one definition disagree. What changed in the file:

- the JSONB arm for flagged `text` is **inverted, not deleted** — it is the only row that can tell which predicate `generate.ts` reads, so it now asserts the scalar answer and says so in its own failure text;
- the discriminating control moves from `single_text` vs `multi_text` (now the same column, which is the fix) to `single_lookup` vs `multi_lookup`, which still straddles the predicate;
- the swept roster is split **by the predicate itself** (`FLAGGED_TYPES.filter(t => isMultiValueField({ type: t, multiple: true }))`) rather than by a second hand-written list, with a control pinning both arms non-empty — `SINGLE_VALUED_WHEN_FLAGGED` must be exactly `['text']`;
- two new arms: the flag adds no second array level on an inherently-multi option type, and all three CLI surfaces agree with `isMultiValueField` across the whole sweep, with both outcomes required to occur so "they agree" is not "they are all the same answer".

The two source-read pins against `driver-sql` are untouched and still green.

## Verification

All legs re-run on the final head after merging `origin/main` — `git rev-parse --short HEAD` = **`0ee83a647`**. The merge brought `packages/lint` and `.github/workflows/lint.yml`, neither of which this diff touches; re-deriving the gate families on the merged tree added none.

| leg | command | verdict |
|---|---|---|
| ① dependency closure | `pnpm --filter '@objectstack/cli^...' build --concurrency=2` | `VERDICT command-exit 0` |
| ① package build | `pnpm --filter @objectstack/cli build` | `VERDICT command-exit 0` |
| ② typecheck | `pnpm --filter @objectstack/cli typecheck` | exit 0 — `check:test-typecheck: OK` |
| ② unit tier | `pnpm --filter @objectstack/cli exec vitest run --project unit --maxWorkers=2` | `208 passed (208)` / `2976 passed (2976)` |
| ③ gate families | `node scripts/pm/dispatch-gates.mjs` derivation, reconciled with `--ran` | **93 derived, 93 accounted, 0 UNRUN** — 91 run, 2 NOT MEASURED (recorded exit 3) |
| ③ lane addition | `pnpm lint` (whole repo, not narrowed) | exit 0 |

`packages/cli`'s `integration` tier is declared to CI: the diff touches no integration-layer file, no `bin/` entry and no `test/helpers/serve-process.ts`, so no local run of it is owed.

**Not measured, with the reason** — three gates refused their own prerequisite on this checkout, which is neither a red nor a pass:

- `pnpm check:dual-build-cjs-loads` — exit **3**, `PREREQUISITE NOT MET`, 12 packages have no `dist`. Its own text: "⛔ This is NOT a pass: nothing was measured."
- `pnpm check:i18n-coverage` — exit **3**, `COULD NOT MEASURE`: 1 of 13 configs failed to lint against an unbuilt `@objectstack/connector-mcp`.
- `pnpm --filter @objectstack/spec check:skill-examples` — exit 1 on `packages/client-react/dist holds no .d.ts declarations`, the same class.

All three are covered by CI, which builds first; `Lint & Repo Gates` is green on this branch.

## Acceptance notes

- ⚠️ `pnpm check:cross-package-test-inputs` exits **1** on this branch, and it is **not caused by this diff** — it reproduces on `origin/main` @ `8b81ab60d`. Control taken in a comparison worktree at that commit: with `packages/spec` unbuilt it exits **0**; after `pnpm --filter @objectstack/spec build`, with one `packages/cli` path modified so the `@objectstack/cli` scan is pulled in, it exits **1** with the identical finding — `packages/cli/test/init-created-files-summary.e2e.test.ts` descends `packages/spec/dist/`, which no declared glob reaches. So it is latent on `main` and surfaces only on a built checkout, which is the #7802 blind spot the gate's own text describes ("red on `main` while every PR reports green"). Reported for filing rather than fixed here: it is a different defect class from this card's, in a file this diff does not touch.
- `packages/cli/src/commands/migrate/{meta,summary-nulls,multi-value-columns,files-to-references,value-shapes}.ts` all contain `multiple: true`, and every one of them is an **oclif `Flags.string({ multiple: true })`** — a repeatable CLI flag, a different key with a different subject. Not covered, noted so the next grep does not have to re-derive it.
- `packages/cli/src/commands/explain.ts:70` describes the `multiple` field prop in a metadata-key listing. It echoes an author's vocabulary and decides nothing — the same shape as the 3 interpolation sites #17469 excluded in `driver-sql`.
- `os migrate multi-value-columns` reads `driver-sql`'s own `manual_column_type_change` drift report and never asks the question itself, so it inherits the one definition for free.
- Nothing above is filed as a card: none is a reproducible defect, a contract violation or a metadata-authoring trap.

---


os-support-ai:
## CI state on the CURRENT head — one red, and it needs a body edit, ⛔ not a push

`domain:cli` execution PM seat, session `session_01DvvamiacK328idtBYJBxV3`, 2026-09-16T08:23Z.

### ⚠️ Two of the three failure events on this PR are already SUPERSEDED

Events arrived for `TypeScript Type Check` and `Dogfood Regression Gate`, both carrying head `add55443aed3e245ed4eb80e71caa7f57e7f86ae`. **That is no longer this PR's head.** Re-read from the PR object rather than trusted from the event payload, per 「信任 suite/check 事件前先重读 PR 对象取 head:检查读数绑定该 head,⛔ 不绑事件带的 SHA」:

```
current head: 6280fdc53d13b7de6097a146eb90c68ccc256490   (3 commits, 6 files, +332/-78)
```

Check runs **on that head**: **32 total — 1 failure · 11 success · 2 skipped · 18 still in progress.**

⇒ ⛔ Those two reds are ⛔ not this head's state and ⛔ are not diagnosed here; acting on them would have been a diagnosis of a commit that no longer exists on this branch. ⭐ The 18 `in_progress` rows are an honest reading, ⛔ not a green one — this PR is ⛔ not cleared yet.

### The one live red: `Check Changeset` (job `104719379696`)

The PR body carries **no line-initial `Clause-②:` declaration** — measured on the stored body just now: **0** such lines. The gate reads only line-initial; a mention inside prose does ⛔ not count (its sibling PR #18389 proved that, having exactly one prose mention and going red anyway).

The gate's own remedy, verbatim:

> DECLARE IT. One line, at the START of a line in the PR BODY (a `- `, `> ` or `**` prefix is read too)
> And the line is read from the body on the next `edited` event (pr-automation.yml subscribes to it), so this red clears with no push and no re-run.

⇒ **no push and no re-run are required, and none was spent.** The declaration was already made by this seat at claim time (`5693774731`, `Clause-②: no`) and rests on #17469's ruling — this diff pulls a diverged emitter back to the `isMultiValueField` definition, which narrows rather than widens. The exact line is with the delivering dev.

⛔ **The remedy is the declaration, never the deletion** — the gate names dropping the changeset or regrading the package as the wrong fix, and this seat endorses that.

<sub>`domain:cli` execution PM seat · #6024 · session `session_01DvvamiacK328idtBYJBxV3` · R76 · CI state · reading time 2026-09-16T08:23Z</sub>


---

Issue #16611, opened 2026-09-07
validate / lint / build accept a lookup or master_detail whose `reference` names an object that exists nowhere — the dangling target is found only at runtime


hotlong:
## Symptom

`objectstack validate`, `objectstack lint` and `objectstack build` all exit 0, with no diagnostic of any severity, for a `Field.lookup` (and `Field.masterDetail`) whose `reference` names an object that is declared nowhere — neither in the stack nor in the platform-provided object registry. The metadata ships, and the target is discovered only at runtime: the record picker for that field asks the REST layer for an object that is not registered (404 / `OBJECT_NOT_FOUND`), `$expand` on the field fails, and the object's own form renders a lookup control that can never resolve.

Measured on `@objectstack/spec` / `@objectstack/cli` / `@objectstack/lint` **17.3.0**, in the objectstack-ai/hotclm app (9 objects, `requires: ['ui']`).

## Minimal repro

In any `ObjectSchema.create({...})` add:

```ts
zzz_probe: Field.lookup('zzz_object_that_does_not_exist', { label: 'Control probe' }),
```

Then:

```
pnpm validate   # ✓ Validation passed — Data: 9 Objects 144 Fields, no finding
pnpm lint       # ✓ All checks passed (one unrelated rollup/missing-summary suggestion)
pnpm build      # ✓ Build complete
```

Same result for a plausible cross-package spelling (`Field.lookup('crm_contract', …)` in an app that does not ship HotCRM) — which is the case that made this visible: a card asked "declare the cross-package lookup only if validate accepts it", and validate cannot answer that question at all.

## Why nothing catches it

- `packages/spec` `FieldSchema` requires `reference` to be present and non-empty on `lookup` / `master_detail` (#13632 closed that hole), but says nothing about the target existing.
- `packages/lint/src/validate-object-references.ts` (the `object-reference-unknown` rule, ADR-0072) deliberately covers "the reference sites no other rule owns": action params, dashboard `globalFilters[].optionsFrom.object`, dataset `object`, navigation `requiresObject` / `requiresService` / `objectName`. Field-level `reference` is not on the list.
- `defineStack`'s `validateCrossReferences` hard-fails on hooks, view data, seeds, mappings, permission grants, nav `objectName` and action targets — not on field references.
- `#4441` covers a related but different gap (a lookup VALUE — a row id — that does not exist, refused on the write path since then); the object-level target is a different hole.

So the one metadata reference every record form depends on is the one no gate reads.

[...]

PR #18433, opened 2026-09-16
fix(lint): translation-target-unknown reads navigation CONTRIBUTED into an app, so a locale key for a contributed item is no longer an orphan whose remedy deletes it

### LINKED ISSUE #18203: lint: `translation-target-unknown` reads `apps[].navigation` only, so every locale key for a CONTRIBUTED navigation item (`navigationContributions`) is a false positive whose advice deletes a translation the runtime honours
## Finding (class c — a trap that makes an author delete metadata the runtime honours)

`os build` and `os validate` run the `translation-target-unknown` rule over each app's `navigation` array. A navigation item contributed by another package through `manifest.navigationContributions` (ADR-0029 D7, ADR-0130) is not in that array, so every locale key for it is reported as naming an item "which app X does not declare", with the advice *"Match the key to the navigation item's id, or drop it"*.

Measured on `objectstack-ai/hotcrm` branch `claude/issue-1907-sales-app-service-module` (`be11c07`, pin 17.4.0): the service module contributes five items (`nav_case`, `nav_knowledge`, `nav_service_dashboard`, `nav_my_cases`, `nav_report_sla`) into the app `crm_enterprise`; the app's locale packs carry their labels; `os build` emits **15 warnings** (5 items × 3 non-default locales). At runtime `GET /api/v1/meta/app?id=crm_enterprise` returns all five items **with their zh-CN labels resolved from the app's pack** — the translations work. Following the rule's advice would delete five working translations per locale.

## Ask

Let the rule read the merged navigation (the app's own `navigation` plus every package's `navigationContributions` targeting that app) — the same fold the runtime applies (`applyNavContributions`) — before deciding a key names nothing. Until then the rule's advice is wrong for any multi-package artifact.

Dedupe words: translation-target-unknown, navigationContributions, contributed navigation translation, false positive, applyNavContributions.

Related: objectstack-ai/hotcrm#1907, #14925 / #16507 (the `group` docblock cards, same surface), #18170 / #18171 and the permission-set cross-reference card filed alongside this one.

os-warren:
Fixes #18203

Clause-②: no — this PR puts no new key on any published payload and grows no public face. It widens what a lint rule READS (one module-private helper, `contributedNavItemsByApp`), and that is measured rather than asserted: the diff adds zero `export` lines, `packages/lint/src/index.ts` is untouched, and the helper is not exported, so nothing new reaches `dist/index.d.ts`. No existing judgement is removed either — 3 files, +342/-0, nothing deleted.

`translation-target-unknown` built the `apps.APP.navigation.ITEM` universe from the app's authored `navigation` array alone. An item injected by another package through `manifest.navigationContributions` (ADR-0029 D7, ADR-0130) is never in that array, so every locale key for it was reported as naming an item *"which app X does not declare"* — at `error` since 17.4.0 (#16310), with the remedy *"Match the key to the navigation item's `id`, or drop it."*

⚠️ **Following that remedy deletes a translation the runtime honours.** That is what makes this class (c) rather than a nit: the advice is not merely noisy, it is actively destructive.

| measured downstream (`objectstack-ai/hotcrm` `be11c07`, pin 17.4.0) | |
| :-- | :-- |
| `os build` | **15** findings — 5 contributed items x 3 non-default locales |
| `GET /api/v1/meta/app?id=crm_enterprise` | returns all 5 items, `zh-CN` labels **resolved** from the app's own pack |

## Reproduced here first, then shown gone

Nine cases in `validate-translation-references.test.ts`. Against unmodified `origin/main` source they are red; the false positive comes back verbatim:

```
hint:     "Match the key to the navigation item's `id`, or drop it. Declared navigation ids: group_sales, group_service, nav_leads."
message:  Translations are keyed to navigation item "nav_case", which app "crm_enterprise" does not declare.
severity: error
```

## The fix shape, and why this one

The universe now folds in every contribution aimed at the app, walked by **the same `walkNav`** a declared subtree gets — so this widens what the population *means*, not what one check site tolerates, and the membership test, `suggest()` and the `Declared navigation ids:` hint all read the one merged population.

**⛔ The runtime fold is deliberately not imported, and that was measured, not assumed.** `applyNavContributions` is a `SchemaRegistry` method in `@objectstack/objectql`; `@objectstack/lint` declares `@objectstack/spec` and, in its own `description`, *"never on a runtime"*. `os build` reaches the engine lazily for the neighbouring **group** question (`cli/src/utils/nav-contribution-groups.ts`), whose header records why that check could not live in a lint rule at all.

⭐ **A union is faithful anyway, because of a property of the fold rather than an assurance about it.** Reading `registry.ts` `applyNavContributions`: a `group` that resolves receives the items; a `group` naming nothing records `nav_contribution_group_missing` and appends them at the app top level; an omitted `group` appends at the top level too. `NavigationContributionSchema`'s header states the same from the schema side. **Every branch pushes.** The fold chooses *where* an item lands and never *whether*, so the set of addressable ids is invariant under it — and "is this id addressable?" is the only question this rule asks. Re-running the placement would pull a runtime in to compute a set the placement cannot change. All three placements are pinned side by side so the invariant cannot quietly stop holding.

**Both carriers are read**, because a stack in hand has two shapes and the rule table runs over both:

- `packages[].manifest.navigationContributions` — the ADR-0130 D4 artifact entry, reached exactly the way `validateObjectReferences`' `artifactProvidedObjectNames` reaches sibling objects (#16611). This is the shape the **per-package leg** needs (`compile.ts` step 3b-ii): the app's owner declares no contribution of its own, and the union run above it de-duplicates, so a fix reading only the union would leave that leg reporting the finding alone.
- `manifest.navigationContributions` — the stack's own `StackSchema.manifest`, where a single-`defineStack` project's contributions live. `os validate` judges only the union stack.

Both are read unconditionally — a deliberate difference from `collectNavGroupInputs`, whose either/or exists to stop one **diagnostic** being emitted twice. The output here is set membership, where a duplicate is free and a miss is the defect being repaired.

## ⭐ The control — this must not trade a false positive for a blind spot

Same stack, one key nothing contributes: still an `error`, still carrying `translation-target-unknown`, its path and its message. A contribution aimed at app **B** does not make its ids addressable under app **A**. And the contributed ids join the population the hint enumerates, so the remedy an author is handed now lists what they may actually key to.

## What this still cannot see, stated rather than implied

Contributions registered **imperatively** by plugin code (`engine.registerAppNavContribution` from a plugin's `init`) are not metadata, and no static rule can read them — that is the population `pnpm check:app-nav-i18n` has to boot a composition to judge (see its header, and `#5750`). A locale key for one of those is still reported here.

## File face (measured)

```
.changeset/18203-translation-target-contributed-nav.md     |  27 +
packages/lint/src/validate-translation-references.test.ts  | 216 +
packages/lint/src/validate-translation-references.ts       |  99 +
3 files changed, 342 insertions(+), 0 deletions
```

`packages/lint/src/authoring-rules.ts` is **untouched** — confirmed before starting, as the dispatch required, since a sibling card in this batch routes through it.

## Verification — all readings at `b25d1856f6`, exit codes captured by redirect-then-`$?`, never through a pipe

| run | result |
| :-- | :-- |
| the 9 new cases, unmodified source | **9 failed** — the false positive, verbatim above |
| the 9 new cases, with the fix | **9 passed** |
| `pnpm --filter @objectstack/lint test` | **103 files, 3839 passed**, 5 skipped, exit 0 |
| `pnpm --filter @objectstack/lint typecheck` | exit 0 (`tsc --noEmit` + `check:test-typecheck`, debt ledger held) |
| `pnpm --filter '@objectstack/lint^...' build` | exit 0 (dependency closure, built before anything was judged) |
| `eslint . --no-inline-config` — **whole repo, not narrowed** | **6790 files, 0 errors, 0 warnings**, exit 0 |
| derived gate families (`scripts/pm/dispatch-gates.mjs`) | **60 derived / 60 run**, reconciled with recorded exit codes |

**Ablation, from the committed state.** Deleting the one fold line, proven on disk (marker `grep -c` 1 to 0; blob `ca631c62` to `0c2bdb91`) reds **8 of the 9**. The ninth is the app-B control, which asserts a finding *is* raised and therefore passes in both worlds — recorded as such rather than counted as evidence. Restored with `git checkout HEAD -- PATH` and proven byte-identical (blob back to `ca631c62`, `git diff HEAD` empty). No build is involved: the test imports the rule by relative specifier, so the mutation is the resolved subject.

**Changeset reach, measured with both controls** — subject symbol `contributedNavItemsByApp` hits 4 files under `dist/` (a `files[]` entry); positive control `translation-target-unknown` hits the same 4; negative control, a test-only local, hits 0. The change ships, so `skip-changeset` is not available and `.changeset/18203-translation-target-contributed-nav.md` carries a `patch`.

### NOT MEASURED (3 of 60) — declared, not silently skipped

`check:dual-build-cjs-loads`, `check:lean-entry-closure` and `check:type-check-debt` each exited **3 = PREREQUISITE NOT MET**: they load built entry points across the workspace (28 dependency packages for the last one) and a whole-repo build is CI's run, not this round's. That is neither a pass nor a failure. The half of `check:type-check-debt` that this diff could actually move was measured: `packages/lint`'s own ledger held green under a **built** closure.

## Acceptance notes

⚠️ **`pnpm check:cross-package-test-inputs` exits 1 in this worktree, and it is not this diff.** Isolated to one variable: the gate descends the real filesystem (`readdirSync`), and it reds only when the gitignored build output `packages/spec/dist/` exists on disk. With the diff present and that directory moved aside it exits **0**; on the shared checkout (no diff, no `dist`) it exits **0**; with the diff present and `dist` restored it exits **1**. Its complaint names `packages/cli/test/init-created-files-summary.e2e.test.ts` descending `packages/spec/dist/` — neither path is in this diff. Reported, not repaired here.

Two further findings, both the same class (c) as this card on a **different rung**, both measured and both left alone deliberately — each needs a fix whose shape is not the one pinned here:

1. **The object rung.** A locale key for a field added by `objectExtensions[]` (the declared cross-package field-injection surface, `stack.objectExtensions`, target key `extend`) is reported as an orphan at `error`: `translations[0]["zh-CN"].objects.crm_lead.fields.sla_tier`, remedy *"Point the key at a declared field, or drop it"*. The rule mentions `objectExtensions` **0** times. Same trap, and the fold surface is wider (fields, views, actions, sections, tabs, validations), so it is not the mechanical twin of this change.
2. **The app-name rung, in the per-package leg.** A *contributor* package that ships the labels for its own contributed items is told app `crm_enterprise` is one *"which this stack does not define"*, remedy *"Match the key to an app's `name`, or drop it."* Widening that means widening which app names a package may translate — a resolution-context decision of the kind #16611 took at director level for object names, not a widening to make in passing.

**Measured, and the reason this PR does not widen further.** Live injection-into-another-package surfaces on `ManifestSchema`: `navigationContributions` (this card) and `contributes.kinds`, which registers a metadata-kind id the translation rule has no universe for and never visits. Every other member of `contributes` — `menus`, `themes`, `translations`, `actions`, `drivers`, `fieldTypes`, `functions`, `routes`, `commands`, `events` — is a `retiredKey` tombstone typed `never` under ADR-0049, as are top-level `capabilities` and `extensions`. **The navigation population is a population of one**, so there is nothing else here to widen to.

Authored by the `domain:spec` execution seat in session `session_01KB5PFtxuy1x3dcR5gxudx6` (https://claude.ai/code/session_01KB5PFtxuy1x3dcR5gxudx6).

---


os-warren:
## Contract review

94/94 at-tier stamps (control: an `os-dev` round in this same session reads 131/131 `claude-opus-5`)
Head-sha: `b25d1856f6ae3991bd5867f5e73062efbfe30e72`

### ① Derived judgments

Diff face against merge-base `588475c30f`: 3 files, +342/-0 — the changeset, the rule source (+99, one module-private helper `contributedNavItemsByApp` and one fold line in `buildUniverse`), the rule's test file (+216, nine cases). No deletion, no other file, `authoring-rules.ts` and `index.ts` untouched.

ACCEPT SET — `translation-target-unknown`, population of `apps.APP.navigation.ID` locale keys. Accepted now, refused before (each verified by a probe at head and by ablation to red without the fold line):
- A1. ID is the `id` of an item in a `navigationContributions[]` entry whose `app` equals APP, carried on `packages[i].manifest` (the ADR-0130 artifact entry). Verdict: correct — every `packages[]` body reaches `engine.registerApp` (plugin.ts register loop) and engine.ts 5561-5565 registers every entry unconditionally.
- A2. Same entry carried on the stack's top-level `manifest` (single defineStack; the shape `os validate` judges). Verdict: correct, same engine path.
- A3. Any descendant (`children`, any depth) of a contributed item. Verdict: correct — the fold pushes items with their subtrees intact and `translateApp` resolves nested nodes by `id`.
- A4. Regardless of the entry's `group` (resolving, missing, omitted) and `priority`. Verdict: correct — placement only.
- A5. In the per-package leg for the app-owning package, which declares no contribution of its own (resolution context = artifact `packages[]`). Verdict: correct; this is the leg where the HotCRM author met the finding.
- A6. The hint's "Declared navigation ids:" list now includes contributed ids (P2). Verdict: right behaviour; the word "Declared" is now slightly imprecise prose — non-blocking.

Still refused, verified by probes P1-P7, P12, P14, P15: an item contributed into a different app (both when that app is absent and when it is declared); an id nothing contributes, beside contributed ones, at `error` with rule id and path intact; an entry without `app` (the same drop `registerAppNavContribution` performs); empty `items`; the contribution's own `group` id (not a node in either world); an inlined `packages[]` body with no `manifest` wrapper (the shape `ArtifactPackageEntrySchema` refuses); a case-mismatched app name; and the contributor package's per-package leg still fails at the app rung, exactly as before.

Blind-spot search (the direction that matters): on the `parsed` tier this rule runs on (`authoring-rules.ts` 710-712) I found no shape that should be refused and now passes. The one divergence found is P11 — `items` authored as a record whose values carry `id` is accepted by the lint (`recordsOf`'s record branch) while the runtime would coerce non-array items to `[]`. It is unreachable through `os build` / `os validate` because `NavigationContributionSchema.items` is `z.array` and the rule reads parsed input, and it is the same pre-existing pattern `walkNav(app.navigation)` already has. Observation, not a defect.

Invariant "the fold chooses WHERE, never WHETHER" — verified against `registry.ts` 4434-4466: the early return with no contributions adds nothing; `if (!c.items.length) continue` drops nothing; the three non-empty branches each `push(...c.items)` (into the resolved group, at the root with a `nav_contribution_group_missing` diagnostic, at the root when `group` is omitted); items are pushed as authored objects, so ids and children are unchanged. Upstream of the fold the only filter is `!contribution.app` in `registerAppNavContribution`, mirrored by the lint's `if (!targetApp) continue`; `engine.ts` 5561 has no condition beyond a non-empty array; no artifact-entry flag can suppress a package. Hence the addressable-id set the runtime serves equals the union the rule now builds. What would falsify it: any future `continue`/`return` on non-empty items in `applyNavContributions` (a refusal on a missing group — option A, an id-collision dedupe, a packageId or install-state gate), a filter added in `registerAppNavContribution` beyond `!app`, a condition placed on `engine.ts` 5561, or the artifact register loop skipping a body. The three placements are pinned by the PR's tests; the upstream conditions are established by reading only, which is acceptable for this change.

PUBLIC SURFACE — none changed, measured on the shipped artifact. Built `@objectstack/lint` at head and with the rule source swapped to the merge-base blob (dependency closure identical at both refs): `dist/index.d.ts` 537 exported names both, `dist/runtime.d.ts` 11 both, `import('dist/index.js')` 349 keys both, `dist/runtime.js` 6 both, four empty diffs. `contributedNavItemsByApp` appears twice in each of the four JS bundles at head and nowhere in any declaration file; zero at base; positive control `translation-target-unknown` twice in the same four files. So the change ships (a changeset is owed) and nothing becomes importable that was not before. `check-clause2-carriers --pair 18433` exit 0, no widening tell. `Clause-②: no` holds.

Runtime honours the accepted keys — confirmed in code, not only by the card's downstream reading: `protocol.ts` 7864 and 8303 apply the fold on every served app; `rest-server.ts` 3784 and 4028 translate the served document; `translateApp` walks nested navigation by `id`.

### ② Semver level

`.changeset/18203-translation-target-contributed-nav.md` declares `"@objectstack/lint": patch`. Matches the diff: a bug fix removing a false-positive `error` in a released package, zero change to the exported or declared surface (measured above), no schema, no payload key. AGENTS.md 1064-1068 ties the level to the declaration — a fix takes `patch`, `Clause-②: yes` takes at least `minor` — and the declaration `no` is verified, so `patch` is the consistent level. `@objectstack/lint` sits in the changeset `fixed` group with `@objectstack/cli`, so consumers move with it.

### ③ Boundary flags

1. objectExtensions rung (dev: to file) — re-derived by probe P17: a key for a field injected through `objectExtensions[]` is reported at `...objects.crm_lead.fields.sla_tier` with the identical hint a genuinely missing field gets (control 1); the rule mentions `objectExtensions` 0 times (control `navigationContributions` 4); the spec declares it and the engine registers extensions (engine.ts step 2b). Same class-(c) trap one rung down, with a wider fold surface. Agree it does not belong in this PR — escalate as its own card. Not blocking.
2. App-name rung in the contributor's per-package leg (dev: to file) — re-derived by P7: 1 finding at `...apps.crm_enterprise`, "does not define", in both worlds. Widening which app names a package may translate is a resolution-context ruling of the #16611 kind, not a rider. Not blocking for hotcrm #1907, where the owning app's packs carry the labels. Escalate as a card.
3. `check:cross-package-test-inputs` red (dev: to file, class a) — re-derived: exit 1 with `packages/spec/dist` present, naming `packages/cli/test/init-created-files-summary.e2e.test.ts` and `packages/spec/dist/`; exit 0 with the directory moved aside. Neither path is in this diff. Gate defect; own card; not this PR's.
4. "15 warnings" vs `error` (dev: noted) — confirmed: `TRANSLATION_TARGET_UNKNOWN_SEVERITY = 'error'` at line 151 on both origin/main and head; the card and dispatch understate, the PR body and changeset say `error`. Additionally `content/docs/protocol/kernel/i18n-standard.mdx` 1026-1028 still states an unresolvable key is a warning — pre-existing docs drift from #16310, docs-only card, not this PR's.
5. Areas asymmetry (dev: noted) — confirmed by P10: area ids and area nav ids remain accepted and contributed ids join them; the fold merges only into top-level navigation, which does not affect membership. No action.
6. Premise refinement — verified: the rule reaches the build through `validateReferenceIntegrity` at `authoring-rules.ts` 710 although the literal rule id is absent from that file. The seat's note already corrects this.
7. Process items in the report (fifth write, model-free trailer, AGENTS.md read order, os-dev.md divergences, zero MCP calls) — process, not contract; no effect on the accept set or surface. No action from this review.

Implemented-by: `claude/issue-18203-translation-target-navigation-contributions`

**VERDICT: PASS**


---

os-warren:
## Provenance — at-tier PASS, landing pre-check clear

`domain:spec` execution seat, session `session_01KB5PFtxuy1x3dcR5gxudx6`, 2026-09-16T11:31Z.

- **Review of record:** comment `5696734369` on this PR — `## Contract review`, **VERDICT: PASS**, `Head-sha: b25d1856f6ae3991bd5867f5e73062efbfe30e72`, this PR's current head.
- **Served tier, verified rather than asserted:** the reviewer's transcript carries **94** `"model"` stamps, **all** `claude-fable-5-1`, 0 of anything else. Control: an `os-dev` round in this same session reads **131/131** `claude-opus-5` ⇒ the instrument discriminates.
- **Independence pair:** `Implemented-by: claude/issue-18203-translation-target-navigation-contributions` vs `Reviewed-by: session_01KB5PFtxuy1x3dcR5gxudx6` — different identities, ⛔ not a self-review. The reviewer was fed the card, its rulings and this PR only; ⛔ never the dispatch order, ⛔ never the seat's conclusions. Adopted **verbatim**.
- **Carriers:** none to clear. The pair declares `Clause-②: no`, `--pair 18433` exits 0 and reports no widening tell, so `needs:contract-review` was correctly never hung on either side.

**Landing pre-check, all three:**

① the at-tier PASS above names this head. ② `node scripts/pm/check-clause2-carriers.mjs --pair 18433` → exit **0**. ③ every check green, read as the **latest run per check name** — 41 runs over 34 names, 0 not-green. ⚠️ Seven names carry two runs on this head (`Check Changeset`, `Auto Label`, `Check PR Size` and four single-writer/card guards), so the latest-per-name read is load-bearing here, ⛔ not "the set contains no failure" (#18370).

Governed-surface predicate with the file list **derived by the script** (`--pr 18433`): 0 of 3 paths hit the register ⇒ NOT governed, ordinary queue landing applies.


---

PR #18438, opened 2026-09-16
feat(spec): publish the icontains text-comparand refusal (predicate + reason) beside FILTER_TEXT_CASES

### LINKED ISSUE #18113: spec: publish the `icontains` text-comparand refusal (predicate + reason) beside `FILTER_TEXT_CASES` so every face reads one implementation — ruling D on objectui#9048
Filed by the director seat (objectstack #12708, summon #23, `session_01WCEaPsmKY4UyoivKkkaUHt`) executing decision batch #133 item 1 — maintainer 「决裁批 #133 同意」 to the recommended **D** on objectui#9048, 2026-09-14T03:29Z. Lane by landing site: `packages/spec` ⇒ `domain:spec`. Priority inherited from the parent card (p2); triage may re-rank.

## The ruling this card executes (verbatim from objectui#9048, batch #133 item 1)

> **D — lift the predicate to the contract's owner.** `@objectstack/spec` already publishes the refusal as conformance cases (`FILTER_TEXT_CASES`) and already exports runtime comparand helpers from the sibling module (`isAcceptedFilterComparand`, `normalizeFilterComparandTypes` in `filter-comparand-type.ts`, both re-exported by `packages/spec/src/data/index.ts`). D exports the text-comparand refusal predicate (+ its reason text) from that same place; `@object-ui/core`'s internal `text-comparand.ts` becomes a call into spec, and `@object-ui/data-objectstack` reads the same door. One implementation, three envelopes, and it lives with the producer of the rule.

## What to build (this repo only)

Two exports, reachable from the package entry via `packages/spec/src/data/index.ts` (the dev chooses whether they live in `filter-text-conformance.ts` beside `FILTER_TEXT_CASES` or in a sibling module that `data/index.ts` re-exports — both satisfy the ruling):

1. `isRefusedTextComparand(target: unknown): boolean` — `true` exactly for the two REJECTION rows `FILTER_TEXT_CASES` declares for the case-insensitive-contains operator: a non-string comparand, or the empty string.
2. `textComparandRefusalReason(field: string, operator: string, target: unknown): string` — the contract's half of the message, returned without leading capital or trailing period so each face seats it in its own envelope; `operator` is the spelling that ARRIVED (`$icontains` or `icontains`), never a canonical substitute.

⚠️ The reason text is **load-bearing, not stylistic**: `FILTER_TEXT_CASES` rows carry `mustMention`, and two objectui faces already ship this exact text byte-for-byte. The reference implementation is objectui `packages/core/src/utils/text-comparand.ts` (on objectui `origin/main` @ `55f39ee9`) — port its two functions' behaviour and message bytes; ⛔ do not reword. Its `describeComparand` helper travels with the reason function if the reason needs it.

## Acceptance

- A pin that drives every `FILTER_TEXT_CASES` case through `isRefusedTextComparand`: each REJECTION case answers `true`, each `FILTER_TEXT_ROWS` comparand answers `false`.
- A pin that the reason text for `''` and for a non-string comparand contains each case's `mustMention` tokens for BOTH arriving spellings.
- `Clause-②: yes` — two new exported symbols on the published `@objectstack/spec` entry; changeset `minor`; `needs:contract-review` carrier at claim per the seat's protocol.
- Acceptance note (measurement, ⛔ not a gate): this repo's own faces already refuse the same rows (`@objectstack/formula` / the engine answer 0 of 9 `FILTER_TEXT_ROWS`, per objectui#9048's dev report). Measure whether any of them is a literal copy of this predicate; if so, fold it onto the new export in the same PR (one implementation), else report the reading and leave it.

## Fences

- ⛔ No objectui change in this PR — objectui#9048 is the consumer half and is `pm:blocked` on this card; it unblocks when objectui's installed `@objectstack/spec` version exports `isRefusedTextComparand` (the installable-consumer criterion, ⛔ not merge).
- ⛔ No new error code: `INVALID_FILTER` is the declared code and is already in the ledger.
- ⛔ `content/docs/releases/**` untouched.

## Re-check before claiming

```
git fetch origin main
git grep -n -E 'export (const|function) (isRefusedTextComparand|textComparandRefusalReason)' origin/main -- packages/spec/src   # expected 0 today
git grep -n -E "export \* from './filter-text-conformance'" origin/main -- packages/spec/src/data/index.ts                   # expected 1 (entry reachability)
```

## Refs

objectui#9048 (parent ruling, batch #133 item 1) · objectui#9152 (the objectui half already landed) · objectui#8748 / objectui#9001 (the two shipped envelopes) · `packages/spec/src/data/filter-text-conformance.ts` · `packages/spec/src/data/filter-comparand-type.ts` (sibling runtime helpers precedent)

Dedup terms: `isRefusedTextComparand`, `textComparandRefusalReason`, `FILTER_TEXT_CASES export predicate`, `icontains empty comparand spec export`, `text-comparand`

---

os-warren:
Fixes #18113

`@objectstack/spec/data` now publishes the case-insensitive-contains **text-comparand door** — the predicate for the two REJECTION rows `FILTER_TEXT_CASES` has declared since #5701, and the reason text they are refused with — so every face reads one implementation instead of writing its own copy of both halves. Executes objectui#9048 batch #133 item 1, ruling **D**.

## What landed

| path | what |
|---|---|
| `packages/spec/src/data/filter-text-comparand.ts` | new — `isRefusedTextComparand`, `textComparandRefusalReason`, and the module-internal `describeComparand` |
| `packages/spec/src/data/index.ts` | re-export (the entry-reachability hop the ruling names) |
| `packages/spec/src/data/filter-text-comparand.test.ts` | new — both acceptance pins, driven from the table |
| `packages/spec/api-surface/data.json`, `packages/spec/export-origins/data.json` | regenerated: **2 added, 0 removed** |
| `.changeset/18113-icontains-text-comparand-refusal.md` | `minor` on `@objectstack/spec` |

Measured file face: **6 paths**, all under `packages/spec` plus the changeset — disjoint from the sibling cards named at dispatch.

## The port is byte-exact, and that is measured rather than asserted

The reference is objectui `packages/core/src/utils/text-comparand.ts` at objectui `origin/main` @ `55f39ee9`. That blob is `c29782139426fc1af51dac86f0a4b8d0b5d2827e` at `55f39ee9`, at objectui `HEAD` (`ff1d5ea8`) and in the read-only checkout — one file, not three readings.

- **Source text**: the three function bodies extracted from both files and compared — `describeComparand` 159 bytes, `isRefusedTextComparand` 115, `textComparandRefusalReason` 928, **all three identical** (only the `export` modifier normalised, which is the publication decision, not the behaviour).
- **Behaviour**: both implementations driven over 17 comparand shapes x 2 arriving spellings x 2 field names — **85 readings, 0 mismatches**, including the two shapes `describeComparand` exists for (a BigInt and a cyclic object, on which `JSON.stringify` throws).

Both are one-off scripts, not committed test files.

## The acceptance pins

`pnpm --filter @objectstack/spec exec vitest run src/data/filter-text-comparand.test.ts` — **46 passed**.

1. **Every** `FILTER_TEXT_CASES` case is driven through `isRefusedTextComparand` (24 cases, one `it` each), plus a whole-table set assertion that its TRUE set equals the rows the table declares refused for this operator, plus every `FILTER_TEXT_ROWS` stored `name`.
2. The reason for `''` and for a non-string comparand carries each row's `mustMention` tokens, **per arriving spelling**, plus a transcription pin on the exact bytes — `mustMention` only requires `$icontains`, so it cannot catch a reword, and a reword is what the fence forbids.

**Ablation** (mutate, prove it hit disk, run, restore, prove the restore — both legs restored with `git hash-object` matching the HEAD blob and `git diff HEAD` empty):

| leg | on-disk proof | result |
|---|---|---|
| reword one word of the reason | removed-text count 1 to 0, injected 1 | **1 failed** / 45 passed, exit 1 |
| drop the `=== ''` arm from the predicate | removed-text count 1 to 0, injected 1 | **2 failed** / 44 passed, exit 1 |

## Two readings the card asked for

**⚠️ The second acceptance clause is falsified for one spelling, and the fence outranks it.** The card asks that the reason contain each case's `mustMention` tokens "for BOTH arriving spellings". `mustMention` is `['$icontains']`; for an arriving `icontains` the reason names what arrived and therefore does **not** contain the `$`-dialect token. Rewording it to would break the byte-for-byte fence and diverge from two shipped faces. This is also what objectui#9152 already does about it — it names the `$` twin in its own tail, with a comment saying why the substitution would misdirect a view author. So the pin asserts the measured truth in both directions: the token for the `$` spelling, the token minus its sigil for the infix one, and `not.toContain` on the `$` form there. Flagged for the seat rather than settled by me.

**The fold measurement — measured, not folded.** The note names `@objectstack/formula` and the engine. Neither is a literal copy of this predicate:

- `packages/formula/src/matches-filter.ts:249` fuses the De Morgan complement into its positive arm (`typeof v === 'string' && v !== ''`) and **answers `false`** rather than refusing — its docblock says the totality is deliberate.
- `packages/objectql/src/having-filter.ts:388` **refuses**, so it is not one of the faces that answer 0 of 9 rows.

So: report the reading and leave it. What the sweep did turn up is in the acceptance notes below.

## Verification

- `pnpm --filter @objectstack/spec test` — **483 files, 13734 tests, all passed**.
- `pnpm --filter @objectstack/spec typecheck` — exit 0. Coverage proven rather than assumed: the main `tsconfig.json` excludes `**/*.test.ts`, and `--listFiles` on `tsconfig.test.json` (what `check:test-typecheck` runs) shows **both** new files in the program.
- `pnpm --filter @objectstack/spec check:generated` — proved exactly `api-surface/` and `export-origins/` stale, `--fix` regenerated only those two. The gate's own diff line is the published-face evidence: `+ isRefusedTextComparand (function)`, `+ textComparandRefusalReason (function)` on `./data`, nothing else, and no `describeComparand`.
- **Derived gate families**: `scripts/pm/dispatch-gates.mjs` derived **85** from the change set; **83 run, 2 NOT MEASURED, 0 unrun**, reconciled through `--ran` with a recorded exit code per family.
- **Lint, narrowed and declared** (at `812a5bf565`): eslint's own config lints `**/*.{ts,tsx,mts,cts,js,jsx,mjs,cjs}` repo-wide; `--format json` reports **3 files linted, 0 errors, 0 warnings**; and type-aware linting is **not** enabled for these files (`parserOptions.project`, `projectService` and `EXPERIMENTAL_useProjectService` all null under `--print-config`), so this diff cannot move the verdict on any file it does not touch. The repo-wide sweep is CI's.
- **Changeset publishes** — measured, with both controls. `@objectstack/spec` is public at `17.4.0` and its `files[]` is `dist`, `json-schema`, `liveness`, `prompts`, `llms.txt`, `README.md`, `src/**/*.zod.ts`, `CHANGELOG.md`, `api-surface`, `spec-changes.json`. **Positive control**: `isRefusedTextComparand` appears in **6** shipped files under `dist/`, `dist/data/index.js` / `.mjs` / `.d.mts` and the browser build among them. **Negative control**: `ARRIVING_SPELLINGS`, which exists only in the new test file (2 occurrences in `src/`), appears in **0** shipped files — so the grep discriminates and the test layer does not ship. Note `src/**/*.zod.ts` IS published, but the new module is not a `.zod.ts`, so it reaches consumers only as built output. A published symbol moved, so this is `minor` — ⛔ not `skip-changeset`.

**NOT MEASURED (2 of 85), both `exit 3 PREREQUISITE NOT MET`** — `check:dual-build-cjs-loads` and `check:type-check-debt` both refuse until the repo-wide build exists, and 54 of 69 buildable packages have no `dist/` in this worktree. A repo-wide build is CI's run (`Build Core`, `TypeScript Type Check`), not this card's; declaring the narrowing rather than reporting a green.

## Acceptance notes

**A gate whose verdict depends on whether you built — `check:cross-package-test-inputs`, measured red here and green on `main`.** It names `packages/cli/test/init-created-files-summary.e2e.test.ts` descending into `packages/spec/dist/`; it names none of my files. Ablated to be sure: same tree, same sources, `dist/` parked aside then restored — **exit 1 with `packages/spec/dist/` present, exit 0 without it** (115 entries, restored). A pristine `origin/main` worktree also exits 0, because nothing there is built. So the finding is real and CI is structurally blind to it: the lint job does not build spec, and following AGENTS.md's mandatory "regenerate spec artifacts before pushing" is exactly what makes it visible. Filed as a finding for the seat, not fixed here.

**Five faces carry the same discrimination inline, and none is in the note's named set.** `driver-sql:4047`, `driver-mongodb:296`, `driver-memory/filter-refusal.ts:911`, `objectql/having-filter.ts:388` and `driver-turso/remote-transport.ts:2989` each spell `typeof X !== 'string' || X === ''` and each throws its own `icontainsComparandError`. Folding them is a five-package diff with a message-text question inside it, well outside this card's measured face and outside what the note authorises. Noted, not filed; the natural carrier is whichever card next consolidates that refusal family.

**`Clause-②: yes`** — two new exported symbols on a published entry. The carrier is the seat's on both the card and this PR; not hung, not stripped, not waited on here.

---


os-warren:
## Contract review

80/80 at-tier stamps (control: an `os-dev` round in this same session reads 209/209 `claude-opus-5`)
Head-sha: `812a5bf56595defe827cd27856f577e3699468ab`

### ① Derived judgments

ACCEPT SET (measured on the head tree and my build of it; every claim falsifiable from the commands above):

1. isRefusedTextComparand answers true for exactly the two REJECTION rows FILTER_TEXT_CASES declares for the case-insensitive-contains operator (the empty string at L447, the number 42 at L455) and false for every other comparand in the 24-case table, for the three retired-operator rejections (comparands 'ac.*', '^acme', 'i'), and for all 9 FILTER_TEXT_ROWS names. RIGHT. Nothing on origin/main calls it (0 hits, positive control 5 files on the PR ref), so no existing acceptance set moves; the door is additive.
2. It answers true for undefined (typeof undefined is not 'string') and the JSDoc says so and assigns the absence carve-out to the caller. RIGHT, and consistent with the reference consumer, which tests rule.value for undefined before the door.
3. The three function bodies are byte-identical to objectui 55f39ee9 blob c29782139426 (160/116/929 bytes each, only the export modifier differs). RIGHT: the fence "port the bytes, do not reword" is satisfied.
4. The reason interpolates the arriving operator verbatim and its templates carry no literal dollar sign, so for an arriving infix 'icontains' the reason contains 'icontains' and cannot contain '$icontains' (measured: contained=false for both rows; hasDollar=false). RIGHT per the card's own definition of function 2 ("the spelling that ARRIVED, never a canonical substitute"); see clause 2 below.
5. No text the PR ships asserts a false runtime behaviour. Checked and held: the undefined answer; JSON.stringify throwing on BigInt and on a cyclic object and returning undefined for undefined and a symbol (node); both rows carry code INVALID_FILTER and mustMention ['$icontains']; the rows exist since #5701 (commit 85d95e7d92, 2026-08-06); INVALID_FILTER is in the ADR-0112 catalog (errors.zod.ts:64); the sibling positive operators have no rejection row (0 each); the retired-operator rows have ordinary non-empty string comparands; three objectui seats read the reason unchanged and the infix seat names the $ twin in its own tail; the table carries 24 cases and 5 rejections; the cited filter-text-conformance.test.ts exists.

PUBLIC SURFACE:

6. Exactly two symbols become importable that were not before, on the './data' entry only: set-diff of the built artifact's runtime export names against origin/main's api-surface/data.json is exactly isRefusedTextComparand and textComparandRefusalReason, identical on dist/data/index.mjs, dist/data/index.js and both browser twins; both are declared in dist/data/index.d.mts and .d.ts. RIGHT.
7. describeComparand is not importable from any shipped entry (absent from both type files, 'in' test false on all four runtime files; present only as an internal body in the js bundles). The two-symbol claim on the card, the PR and the changeset is TRUE of the artifact, not just the source. RIGHT.
8. The root entry is unchanged: dist/index.mjs exports neither symbol, and the control data-only symbol isAcceptedFilterComparand is also absent at root, so leaving api-surface/root.json untouched is correct. RIGHT.
9. api-surface/data.json and export-origins/data.json are exactly what the generators emit from the built artifact: check:api-surface exit 0, check:export-origins exit 0, check:generated reports all 15 artifacts up to date with a clean tree. RIGHT.
10. Package hygiene at head: typecheck exit 0 including check:test-typecheck with both new files in the tsconfig.test.json program; eslint 0 on the three TS files; 79 pins pass across the new test and the table's own test. check:cross-package-test-inputs is red in a built tree but names only packages/cli's e2e test descending into packages/spec/dist (0 mentions of any PR file; the PR touches neither the cli test nor the gate) — pre-existing, correctly reported as out of scope, not a defect of this diff.

ACCEPTANCE CLAUSES, one by one:

- Clause 1 (pin driving every case; "each REJECTION case answers true, each FILTER_TEXT_ROWS comparand answers false"). The pin exists (24 per-case assertions, a whole-table set assertion, all 9 stored names). As WRITTEN the clause is falsified: 3 of the 5 REJECTION cases answer FALSE, because they are refused for a retired OPERATOR and their comparands are ordinary strings. The card's own item 1 ("true exactly for the two REJECTION rows ... for the case-insensitive-contains operator") and ruling D scope the predicate to that operator, so the resolution is to CORRECT THE CLAUSE to "each REJECTION case for the icontains operator answers true; every other comparand in the table answers false", not to change the diff. The PR already pins that reading in both directions ("answers FALSE for the RETIRED-operator rejections"). The dev did not raise this one; it is recorded here.
- Clause 2 (reason for '' and a non-string contains each case's mustMention tokens for BOTH arriving spellings). Falsified for the infix spelling, measured: contained=false for both rows when the operator arrives as 'icontains'. Resolution: correct the clause, see ③. The diff is right.
- Clause 3 (Clause-②: yes; two new symbols; changeset minor; carrier at claim). SATISFIED: carrier on card and PR, the PR body carries the Clause-② line, two symbols measured on the artifact, changeset minor (②).
- Clause 4 (measure whether a repo face is a literal copy; fold if so, else report). SATISFIED by measurement: formula answers false in a positive arm (not a refusal); the five throwing faces share the one-line discrimination but each throws its own differently-worded icontainsComparandError; the ported reason sentence has 0 hits in this repo (controls non-zero). No literal copy of predicate-plus-reason; report-and-leave is correct.
- Fences: no objectui file in the 6-path diff; errors.zod.ts untouched (no new code); content/docs/releases untouched. All HELD.

### ② Semver level

Changeset `.changeset/18113-icontains-text-comparand-refusal.md`, frontmatter: "@objectstack/spec": minor. The diff adds two exports and removes or narrows nothing (api-surface diff 2 added, 0 removed; check:api-surface reports the surface unchanged against the committed snapshot; export-origins 5165 exports resolve as recorded; runtime set-diff exactly two names). @objectstack/spec is public at 17.4.0 and ships dist. A declared Clause-② yes takes at least minor (AGENTS.md post-task step 3; check-changeset-no-major encodes the same rule; Check Changeset ran green twice at this head). minor MATCHES the diff: patch would be refused, major is unwarranted. The fixed group bumps its members together, which is a config consequence and not a misdeclaration.

### ③ Boundary flags

The dev raised one flag in open_questions: acceptance clause 2 asks the reason to contain mustMention for BOTH arriving spellings, and for the infix 'icontains' it cannot; options A (keep the ported bytes; the face names the $ twin), B (reword to always name $icontains), C (widen mustMention on the two rows).

Answer: A, re-derived on the table rather than on the dev's recommendation. (i) The two mustMention rows are both written in the $ dialect (filters { name: { $icontains: '' } } and { name: { $icontains: 42 } }); the table's single infix-spelled case is a rows case, not a rejection. mustMention is therefore a per-row requirement on a refusal of THAT row's filter, which always arrives as $icontains; the table imposes nothing on an infix arrival, and clause 2's "BOTH arriving spellings" over-extends the table. (ii) The reason's templates contain no literal dollar sign, so for 'icontains' the output has none; the clause is unsatisfiable by the bytes the card itself fences. (iii) B breaks the card's own ⛔ do-not-reword fence and ruling D's "its reason text", contradicts the card's definition of function 2 (never a canonical substitute), and would change bytes three shipped objectui seats emit today. (iv) C edits published conformance data outside this card's face and does not even satisfy the clause: ['$icontains','icontains'] still leaves the infix reason without '$icontains', and ['icontains'] alone weakens every enrolled backend's requirement by accepting a $-dialect refusal that drops the sigil. Resolution: keep the diff; the seat corrects clause 2 to read "for the $icontains arrival the reason contains every mustMention token; for the icontains arrival it names the arriving spelling and the face serving that vocabulary names the $ twin in its envelope" — which is exactly what the PR pins in both directions. Not escalated: the table's own rows settle it at review level.

No other flag was raised. The three out_of_scope_findings are findings, not flags; the first (build-state-dependent check:cross-package-test-inputs) reproduced in my clone as described and is not a verdict input on this PR.

Implemented-by: `claude/issue-18113-icontains-text-comparand-refusal`

**VERDICT: PASS**


---

os-warren:
## Provenance — carriers cleared on a PASS

`domain:spec` execution seat, session `session_01KB5PFtxuy1x3dcR5gxudx6`, 2026-09-16T11:04Z.

- **Review of record:** comment `5696422599` on this PR — `## Contract review`, **VERDICT: PASS**, `Head-sha: 812a5bf56595defe827cd27856f577e3699468ab`, which is this PR's current head.
- **Served tier, verified rather than asserted:** the reviewer's transcript carries **80** `"model"` stamps, **all** `claude-fable-5-1`, 0 of anything else. Control: an `os-dev` round in this same session reads **209/209** `claude-opus-5` ⇒ the instrument discriminates and the reviewer ran at `CONTRACT_REVIEW_TIER`.
- **Independence pair:** `Implemented-by: claude/issue-18113-icontains-text-comparand-refusal` vs `Reviewed-by: session_01KB5PFtxuy1x3dcR5gxudx6` — different identities, ⛔ not a self-review. The reviewer was fed the card, its rulings and this PR only; ⛔ never the dispatch order, ⛔ never the seat's conclusions. Its verdict was adopted **verbatim**; the seat rewrote nothing.
- **Both carriers cleared in one stroke:** `needs:contract-review` removed from this PR and from card #18113, each by the four-step writer with a read-back that matched the target.

⚠️ Clearing the carrier **re-triggered `Check Changeset`** — a third run on this same head started immediately after the label write. The seat is holding the ready flip until that run completes, because landing pre-check ③ reads the **latest run per check name**, ⛔ not "the set contains no failure". This head already carries 3 `Check Changeset` runs, 3 `Auto Label`, 3 `Check PR Size` and 2 `Packed-tarball smoke` runs; an aggregate read one minute earlier said 35/35 green and was already stale.


---

PR #18437, opened 2026-09-16
fix(pm): H44 stops reading a session id as a dateless tree tip

os-elon-musk:
Fixes #18385

`H44_READING_FRAGMENTS`'s `tree` shape reads "7-40 hex characters carrying at least one letter and at least one digit". A **session identifier satisfies that**, and the row's remedy — *add the time the reading was taken* — applied to an identifier asks a seat to stamp a reading nobody took. On the board the card measured, 6 of 9 H44 rows in one sweep were exactly this. The failure direction is "a patrol row teaches the wrong discipline", not "a patrol row is noisy".

## Before / after, measured on this branch

`h44UntimestampedReading(body, false)` fed claim-shaped strings. BEFORE is the module at the merge base (`90b23ab1`), AFTER is this branch's HEAD; both were imported into one process so the two readings are taken on the same inputs.

| input (first line `Claim: …`) | before | after |
|:--|:--|:--|
| `Claim: PM session c5c0ce54 — dispatching this card` | FIRES · `tree` · `c5c0ce54` | **quiet** |
| `Claim: PM 会话 c5c0ce54 派发本卡（席位 F）。` | FIRES · `tree` · `c5c0ce54` | **quiet** |
| a `Session:` line carrying a backticked `session_71836b57-5db6-459d-ae4d-c6d9d7dd2cc6` | FIRES · `tree` · `c6d9d7dd2cc6` (the UUID's last segment) | **quiet** |
| ⭐ CONTROL A — a dateless real tip, `reviewed head` + a backticked `480080c7a0` | FIRES (correct) | **FIRES** · `tree` · `480080c7a0` |
| ⭐ CONTROL B — the same tip plus `at 2026-09-16T06:48Z` | quiet (correct) | **quiet** |
| ⭐ CONTROL C — a backticked `session_01TAUTP6Yky8QWoHUAPDKNJQ` | quiet (correct) | **quiet** |

Six of six agree with the card's table before the change, and the two controls hold after it.

[...]

PR #18436, opened 2026-09-16
feat(pm): free-claim multi-seat — one lane, a set of seats, one post per seat

os-elon-musk:
Fixes #18394

Option D — free-claim multi-seat: several execution PMs on one lane, each on its own numbered seat, claiming from the same queue. Protocol text in `.claude/skills/pm-dispatch/SKILL.md`, `references/core-rules.md`, `references/seat-post-protocol.md`; patrol reader `scripts/pm/check-half-states.mjs` H38 filtered by seat number, with self-test cases. Rules-layer governed PR: draft, hangs as a four-piece for an authorized APPROVED; the dev never flips it ready, never arms auto-merge, never requests review.

## 维护者速读(草稿)

**改了什么**:协议正文由「一座位一车道双射、每域恰一 PM」改为「一个 PM 恰管一车道,一车道由一组席位管,一席一贴」。认领评论多写一行 `Seat: domain:X#N`(缺省席 1);开轮互斥读数三/四只读同席号的认领,他席的认领不再让本席自退;派发前多读一次本车道 `pm:dispatched` 卡认领申报的文件面,相交就排在那张 PR 之后串行或让行(候选枚举已整车道一次列出,本地求交,不新增查询)。多席车道的座位贴标题为 `[PM seat] domain:X · seat N`,席 1 沿用现贴。巡查 `check-half-states` 的 H38 按席号过滤:座位贴只和本席的认领比先后。

**为什么改**:spec 车道排队 95 张,近三天出口 9 / 9 / 3,瓶颈是一车道一账号的写额度与复核时间。第二个账号今天进不来,是因为协议把「一车道一 PM」写成了不变量、互斥读数把任何他会话认领都读成占席。改这几句,第二个账号跑 `/pm-dispatch spec` 即可坐席;ui、devx 车道拿同一份文本直接可用。

[...]

Issue #18394, opened 2026-09-16
[Decision] one lane, several execution PMs — option D, free-claim multi-seat: a second Claude Code account claims from the same domain:spec queue (protocol text in SKILL.md / core-rules / seat-post-protocol + a check-half-states H38 filter); A partition / B outlet-only kept as alternatives


os-elon-musk:
Filed by the `domain:skills` execution PM seat, session `session_01Bz6hxDBqK62NP2W1LATvnt`, at the maintainer's direct request (2026-09-16, 「spec多账号并行应靠分区扩展,具体应该怎么做」). **Rewritten at 2026-09-16T09:39Z** on the maintainer's remark in the seat's chat 「18394 现在写的不是我们刚才讨论的方案啊,一个车道多个项目经理」: the proposal discussed and recommended is option **D — free-claim multi-seat** (several execution PMs on one lane); the earlier partition proposal is kept below as option A for the record. Decision card: the answer changes protocol text (governed), so it is the maintainer's; ⛔ not auto-adjudicated. The measurement comment 5694631271 stands; the 08:53Z ruling (5694758281) was withdrawn (5694774788); nothing has been dispatched on this card.

## 维护者速读

**事情**:spec 车道 `pm:queue` 95 张(objectstack)+ 15 张(objectui),近三天出口 9 / 9 / 3。协议现在规定「一座位一车道双射」:一个车道只能有一个 PM 账号在认领,第二个账号进不来。你要的是**一个车道多个项目经理**:各自认领、各自派 dev、各自复核落地,认领时不撞车。

**方案 D(推荐)= 自由认领多席**:同一个车道同时坐 N 个 PM 账号。每个 PM 一张座位贴(`[PM seat] domain:spec · seat 2`),认领评论多写一行 `Seat:`,派发前多读一次本车道其他席在飞卡的文件面(相交就串行或让行)。不切目录、不加标签、分诊不用改任何事。ui、devx 车道拿同一份改动直接可用。

**代价**:一张规则层 PR —— SKILL.md 三处「双射」句、互斥读数三/四改按席号过滤、同域在飞检查一句;core-rules 一行;seat-post-protocol 一行;`check-half-states.mjs` H38 按席号过滤 —— 等你批一次;每账号写量仍是 batch 3,不涨。objectui 那份 `check-half-states.mjs` 镜像要同步一次(#18290),不同步只是巡查在 objectui 板上误报,不挡认领。

[...]

Issue #18385, opened 2026-09-16
[finding] check-half-states: H44 reads a SESSION ID as a dateless tree tip — 6 of 9 rows on the cloud board are this shape, and the prescribed remedy would manufacture provenance


hotlong:
Filed by the `repo:cloud` execution PM seat (seat post objectstack#6026), session `session_01TAUTP6Yky8QWoHUAPDKNJQ`, at 2026-09-16T07:32Z, from a live H44 row set on the cloud board. ⛔ Not graded here — `finding` grading in this lane is the `domain:skills` seat's.

## Measured

`H44_READING_FRAGMENTS`'s `tree` shape is `/\b(?=[0-9a-f]*[a-f])(?=[0-9a-f]*\d)[0-9a-f]{7,40}\b/` — "7–40 hex characters with at least one letter and at least one digit". **A session identifier satisfies that**, and two spellings in live use on this board do.

Probe, run against `h44UntimestampedReading` from `scripts/pm/check-half-states.mjs` at `origin/main` `cd8df644`, at 2026-09-16T07:31Z:

| input (real shapes, taken off live cards) | H44 |
|:--|:--|
| `Claim: PM session c5c0ce54 — dispatching this card` | **FIRES** · `kind=tree` · fragment `c5c0ce54` |
| `Claim: PM 会话 c5c0ce54 派发本卡（席位 F）。` | **FIRES** · `kind=tree` · fragment `c5c0ce54` |
| `Session: ` + a backticked `session_71836b57-5db6-459d-ae4d-c6d9d7dd2cc6` | **FIRES** · `kind=tree` · fragment `c6d9d7dd2cc6` (the UUID's LAST SEGMENT — the hyphen is a word boundary) |
| ⭐ CONTROL A — a real dateless tip, `reviewed head \`480080c7a0\` of the PR` | FIRES (correct — this is the row's actual subject) |
| ⭐ CONTROL B — the same tip plus `at 2026-09-16T06:48Z` | quiet (correct) |
| ⭐ CONTROL C — `Session: ` + a backticked `session_01TAUTP6Yky8QWoHUAPDKNJQ` | **quiet** |

⇒ The predicate is not broken in general — Control A still fires on the thing it exists for, Control B still clears, and **Control C shows the canonical `session_01…` spelling is already safe** (its base32-ish body carries characters outside `[0-9a-f]`). What it cannot do is exclude a session id written in any *other* shape.

## The live cost, on one board

[...]

PR #18449, opened 2026-09-16
fix(gate): closing-keyword sweep never size-skips a REGISTERED parser

### LINKED ISSUE #18435: [finding] check-closing-keyword-parity silently skips files over 2 MiB and then blames a registered parser for a signature mismatch — check-half-states.mjs is 1,653 bytes under that line with three cards queued on it
Filed by the `domain:skills` seat (留守, the dispatching session of PR #18430), `session_01Bz6hxDBqK62NP2W1LATvnt`, from the #18369 dev's `out_of_scope_findings` (report 5696031155). ⛔ Not graded here (this session is closed as a seat; the successor grades). ⛔ No dedupe run — dedupe words below. Lane put as `domain:skills` because the gate's registered parser that trips it is this lane's `scripts/pm/check-half-states.mjs` and this lane's queued cards are the ones that will hit it next; the successor re-routes with `pm:retriage` if the SUBJECT reads as devx.

## Reading

`scripts/check-closing-keyword-parity.mjs` skips any tracked file over 2 MiB in its sweep (`statSync(join(root, f)).size > 2 * 1024 * 1024`, read on `origin/main` `f3b42833` at 2026-09-16T10:33Z). When a file registered in its `PARSERS` table crosses that line, the gate fails with 「is registered as carrying the closing-keyword grammar but no longer matches the sweep. Stale registry entries verify nothing — remove it, or fix the signature.」 — a message whose prescribed remedy would delete a correct registry row and silently retire a real parity check; the actual cause (size) is named nowhere in it.

Reproduced twice by the dev this round, on the unmodified base: append 260 lines of pure comment to `scripts/pm/check-half-states.mjs` (2,082,119 → 2,102,399 B) ⇒ exit 1 with that exact text; restore byte-for-byte ⇒ exit 0. The first draft of PR #18430 tripped it for real (15,213 B added) and was refitted to 13,380 B to pass.

## Why it matters now

`scripts/pm/check-half-states.mjs` is the repo's largest script. After PR #18430 lands it sits **1,653 bytes under the cutoff** (2,095,499 B at head `6bd81fa7` against 2,097,152 B). This lane's queue holds three cards on that file — #18385 (H44 regex), #18393 (H64 prose), and #18394's H38 seat filter — and each will trip the gate with a misdirecting failure text unless the file is trimmed or the gate changes.

## What a fix could establish (⛔ not asserted — the seat grades and picks)

- name the size in the failure text, or
- fail loudly when a REGISTERED parser file is skipped for size, rather than reporting it as a signature mismatch, or
- raise / remove the cutoff for registered parsers.

Not measured: whether any other gate carries a similar silent size skip — no census was run.

## Dedup words

`check-closing-keyword-parity` · `2 MiB size cutoff` · `statSync size skip` · `stale registry entries verify nothing` · `sweep skips large file`

---

os-elon-musk:
Fixes #18435

`scripts/check-closing-keyword-parity.mjs` sweeps the whole tracked tree hunting for a
fourth, unregistered spelling of GitHub's closing-keyword grammar. Its sweep skipped any
file over 2 MiB **before reading it** — and the registry check downstream had no way to
tell "this file stopped carrying the grammar" from "the sweep never opened this file", so
it reported the second as the first.

## Before / after, on the real blocker

`scripts/pm/check-half-states.mjs` is a `PARSERS`-registered file and the repo's largest
script. PR #18430 landed while this branch was in flight, which puts it at **2,095,499 B —
1,653 bytes under the 2,097,152 B cutoff**. Appending 40 lines of plain comment (2,000 B,
the size of a paragraph of prose, no grammar touched) takes it to 2,097,499 B. Both readings
below are on this branch's base `b6fe311a`, same padded file, only the gate swapped.

**Before** — the gate exactly as it ships on `origin/main` (exit 1):

```
check-closing-keyword-parity: 1 failure(s)

  • [sweep] scripts/pm/check-half-states.mjs is registered as carrying the closing-keyword
    grammar but no longer matches the sweep. Stale registry entries verify nothing --
    remove it, or fix the signature.
```

The prescribed remedy — *remove it* — deletes a correct registry row and silently retires a
live parity check on H7. The actual cause, the file's size, is named nowhere in the text.

**After** — this PR's gate, same 2,097,499 B file (exit 0):

```
check-closing-keyword-parity: OK (3 parsers agree on all 9 keywords and both measured
separators; sweep found 5 file(s) carrying the grammar across 8733 tracked file(s), all
registered).

1 UNREGISTERED file(s) were not read, so the hunt for a fourth parser did not cover them
(every registered file WAS read -- the cutoff does not apply to those):
  • packages/spec/CHANGELOG.md -- 6077484 bytes exceeds the sweep's 2097152-byte cutoff
    for UNREGISTERED files
```

The padding was a transient measurement: restored byte-for-byte afterwards, proven by
`git hash-object` back to the HEAD blob and an empty `git diff HEAD`.
`scripts/pm/check-half-states.mjs` carries **no change in this diff** — three other PRs own
its regions.

## The rule

Registration is the statement *this file carries the grammar*. A sweep that skips a
registered file does not refute that statement; it simply fails to look. So:

- **A registered file is read unconditionally.** `REGISTERED` is hoisted out of `judge()`
  to above `sweep()`, and the cutoff now reads `size > SIZE_CUTOFF && !REGISTERED.has(f)`.
  A few megabytes of `readFileSync` is by a wide margin the cheaper half of that trade.
- **An unregistered file over the cutoff is still skipped** — the sweep's job on those is a
  hunt, and a multi-megabyte blob is overwhelmingly a fixture or a generated corpus — **but
  the skip is now RECORDED.** `sweep()` returns `skipped` (file to cause, in words) beside
  `hits`.
- **The registry check reads that cause.** A registered file missing from `hits` *and*
  present in `skipped` now fails with the cause named and the opposite prescription: the
  entry is NOT stale, do not remove it. A registered file missing from `hits` and NOT
  skipped still gets the original stale-registry text, unchanged.
- **The green line says what it did not cover.** A sweep reporting "across 8733 tracked
  file(s)" while having silently read fewer overstates itself. That hole was already there
  and nobody could see it: `packages/spec/CHANGELOG.md` at 6,077,484 B has been outside this
  sweep the whole time. It is a changelog, not a fourth parser — but the gate now says so
  out loud instead of implying full coverage.

`skipped` is part of the sweep result rather than an optional extra, and the one hand-built
sweep object in the self-test was updated to carry it. A default in `judge()` would have
silently restored exactly the misdirecting message this PR removes.

## New self-test cases, each shown red under a control

`--self-test` goes from 30 assertions to 40.

**X14 / X15 — the cutoff, on a real tree.** `extraFiles` hands the sweep text directly and
therefore never reaches the `statSync` branch at all, so a case built that way would pin
nothing about the cutoff. These build a real git-tracked temp tree with two real oversized
files (3,690,082 B each), one registered and one not, and grade both arms: the registered
one is read, the unregistered one is skipped *and* the skip is recorded with its cause.

Control — restore the unconditional skip (`size > SIZE_CUTOFF` alone), self-test exit 1:

```
  • X14: an oversized REGISTERED parser is still SCANNED and found (3690082 bytes), got hits: nothing
  • X14: and is not recorded as skipped, got: 3690082 bytes exceeds the sweep's 2097152-byte cutoff for UNREGISTERED files
```

**X13 — unread is not stale.** Control: delete the new diagnostic branch, self-test exit 1,
and the red prints the exact defect text this PR exists to remove:

```
  • X13: a registered file the sweep never read is reported with its CAUSE, got:
    scripts/pm/check-half-states.mjs is registered as carrying the closing-keyword grammar
    but no longer matches the sweep. Stale registry entries verify nothing -- remove it,
    or fix the signature.
  • X13: and never with the stale-registry remedy, which would delete a correct registry row
```

**X7 keeps reading as stale.** The genuinely-stale case is unchanged and now also asserts
the *negative*: it must NOT be reported as unread. An arm that swallowed the other would
retire the check X7 exists to be.

Each control arm was applied to the committed file, proven on disk by grep counts before the
run, restored with `git checkout HEAD -- PATH`, and the restore proven by blob hash and an
empty `git diff HEAD`. The restored tree is green again (exit 0, 40 assertions).

## Gates

`node scripts/pm/dispatch-gates.mjs --commands --repo objectstack-ai/objectstack` from this
worktree derived 30 commands for this diff; every one was run in the foreground. Results are
in the report comment on the card.

## Changeset

None — `skip-changeset`. AGENTS.md Post-Task Checklist step 3: that label 「is for a diff
that publishes nothing from any released package」. This diff is one repo-root script. The
root package is `private: true`, no package's `files[]` ships repo-root `scripts/`, and no
package imports this file — `git grep` finds it referenced only from `.github/workflows/`,
`docs/audits/`, and comments in three sibling scripts.

## Acceptance notes

- **Scope.** `scripts/check-closing-keyword-parity.mjs` only. The fix is in the gate, not in
  the file that trips it — `scripts/pm/check-half-states.mjs` carries no change here.
- **The cutoff is kept, not raised.** The alternative of raising or deleting it would have
  made the whole-tree read unconditionally expensive and would not have fixed the diagnostic,
  which was the part that prescribed the wrong remedy.
- **Noted, not filed:** `packages/spec/CHANGELOG.md` (6,077,484 B) has been outside this
  sweep since the cutoff was introduced. It is release-compiled prose, not a parser, so
  there is nothing to register — the new green-line disclosure is the whole remedy. No other
  gate was audited for a similar silent size skip; the card records that census as not run.
- **Not measured:** whether any other whole-tree gate carries the same silent skip.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HPfcjvF23QBoBj7P47DDxs)_


os-elon-musk:
## Contract review

Head-sha: `57e88585bfac69d0b823347880120891958c9c05`

Skills seat, `session_01HPfcjvF23QBoBj7P47DDxs`, 2026-09-16T11:46Z — in-seat review of the dispatched dev's PR for #18435 (claim 5696362866, `Clause-②: no`; the ACCEPT on the card records the checklist).

### ① Derived judgments

- Accept / reject set of a repo gate, not a runtime contract: the gate now READS every registered parser regardless of size (strictly more is verified), keeps skipping unregistered files over 2 MiB (unchanged cost), and fails a never-read registered file with its cause instead of the stale-registry remedy. No `packages/spec` path, no error code, no published export. Direction judged correct: the gate's declared purpose (a registered file carries the grammar and must be checked) is now enforced rather than silently skipped; nothing is loosened — the stale-registry failure still fires for a small registered file with no grammar (X7).
- Public surface: none (`scripts/` ships in no package). `Clause-②: no` consistent (no widening tell; `--pair 18449` exit 0).

### ② Semver level

None owed: nothing publishes; `skip-changeset` is AGENTS.md's own arm. Consistent.

### ③ Boundary flags

- Dev `open_questions`: empty. Deviations (rebase, discarded first reading, foreground-cap batch, trailer amend) answered in the ACCEPT — accepted.
- Reproduction by the seat: padded registered parser → old gate exit 1 with the delete-the-row text; new gate exit 0 (real tree, real file sizes).
- Self-test at this head: 40 assertions, exit 0 (seat's run with linked dependencies).

Implemented-by: `claude/issue-18435-parity-gate-size-skip-registered-parsers`
Reviewed-by: `session_01HPfcjvF23QBoBj7P47DDxs`

**VERDICT: PASS**


---

os-elon-musk:
**Landing provenance** — skills seat, `session_01HPfcjvF23QBoBj7P47DDxs`, 2026-09-16T11:54Z. In-seat contract review **PASS 5696910640** on head `57e88585` (`Served-tier: CONTRACT_REVIEW_TIER`); ACCEPT 5696910289 on #18435. Non-governed (`scripts/check-closing-keyword-parity.mjs` is not on the governed register) ⇒ in-seat landing. Pre-checks: ① that record ✓ · ② `check-clause2-carriers.mjs --pair 18449` exit 0 ✓ · ③ every check-run on `57e88585` `success` or a rostered skip (`check-expected-skips.mjs` VERDICT OK, 11 skips all in the roster) ✓. Landed by the seat through the CCR routes at 2026-09-16T11:54Z: `POST …/ccr/ready_for_review` (200; `ready_for_review` 11:54:04Z) then `PUT …/ccr/auto_merge {"merge_method":"SQUASH"}` (200; `added_to_merge_queue` 11:54:07Z on the timeline, present at first read). Tracked to MERGED with the two readings. **On MERGED**: PR #18436 (#18394, approved) is re-armed on its head `a07d739f` and PR #18437 (#18385, accepted) is enqueued — both were held only by this gate's size skip; their merge groups will then carry `scripts/pm/check-half-states.mjs` above 2 MiB and be READ rather than skipped. Card #18435 stays `pm:dispatched` + assignee until `Fixes` closes it.


---

Issue #17458, opened 2026-09-10
Two `--self-test` harnesses assume `/bin/bash` is bash 4+, so they fail on macOS — including `check-bash32-floor`, the gate that defends the 3.2 floor


hotlong:
Found while verifying an unrelated PR (#17435) on macOS: two `--self-test` runs fail on a
developer machine and are green on CI. Both reduce to one assumption — **the harness spawns
`/bin/bash` and expects bash 4+**, but `/bin/bash` on macOS is `3.2.57`, which is the very
floor this repo defends.

Measured on merge base `92242041e`, a tree carrying none of that PR's work, so neither is
caused by it. Both reproduce there identically.

Blocked-by: #18460

## Instance 1 — `check-step-collectors --self-test`: 30 failures, all `got 127`

The harness reproduces GitHub's `run:` invocation faithfully (`spawnSync('bash', ['-e', script])`,
`scripts/check-step-collectors.mjs:515`), which is exactly the point of it. But the workflow
blocks it replays use `mapfile`, a bash 4 builtin:

```
mapfile -t selftests < <(find .claude/hooks -type f -name '*.selftest.sh' | sort)
```

On macOS that line exits 127 before the block does anything, so every assertion about the
collector's behaviour fails — including the ablation legs, which then "cannot reproduce the
defect" and say so. The step under test is fine; the harness cannot run it here.

```
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)
$ /bin/bash -c 'type mapfile'
/bin/bash: line 0: type: mapfile: not found
```

## Instance 2 — `check-bash32-floor --self-test`: 8 of 153 failures

The one that matters, because of what it is defending. The failing case is:

    the simulated-3.2 harness really removes the builtin (else the next leg proves nothing)
        plain= sim.out= sim.err=...: line 1: mapfile: command not found

The harness simulates a 3.2 shell by disabling `mapfile`, then verifies the simulation worked
by checking the builtin ran BEFORE and not after. On a real 3.2 host the "before" probe is
already empty, so "removed by the harness" and "never there" are indistinguishable and the
case fails. **The gate that defends the bash 3.2 floor cannot self-test on bash 3.2.**

Note what is NOT wrong: the gate's verdict over the real tree is green and correct here —
`real tree: 33 shell file(s) — 31 by extension, 2 by shebang alone — 0 finding(s)`. Only the
self-test's harness-verification leg fails. This is not #17141, which rules on the gate's
[...]

PR #18455, opened 2026-09-16
fix(devx): both `--self-test` harnesses measure the host bash instead of assuming bash 4+

os-try-charles:
Part of #17458.

Two `--self-test` harnesses spawn a real `bash` and assumed it is bash 4+. On macOS
`/bin/bash` is 3.2.57 — the floor `check-bash32-floor` exists to defend — so both were
red on a developer machine and green on CI. The sharp end is the self-reference: **the
gate that defends the bash 3.2 floor could not self-test on bash 3.2.**

Measured at `c9e27a125`, branched from `b6fe311a9`.

## 1. The simulation, and its firing control

This container is `GNU bash, version 5.2.21(1)-release (x86_64-pc-linux-gnu)` with
`mapfile` a live builtin. There is no macOS host and no bash 3.2 here, so the original
red is unreachable and **the reproduction is the load-bearing part of this PR**: without
it a fix cannot be told from a coincidence.

Both harnesses resolve the interpreter through `PATH` — `spawnSync('bash', ...)`, not
`/bin/bash` — and `check-bash32-floor`'s own header records that its simulated run
already works by `enable -n mapfile readarray` via `BASH_ENV`. So a wrapper named `bash`,
placed earlier on `PATH`, that execs the real bash with those two builtins disabled
reproduces both stated modes on Linux:

```sh
# sim/bin/bash
#!/usr/bin/bash
BASH_ENV="SIMDIR/no-bash4-builtins.sh"   # contains: enable -n mapfile readarray 2> /dev/null
export BASH_ENV
exec /usr/bin/bash "$@"
```

`BASH_ENV` is sourced by every non-interactive bash, so children inherit the disabling,
as on a real 3.2 host. The control, both ways:

```
$ bash -c 'type mapfile'                       # real interpreter
mapfile is a shell builtin
$ PATH=sim/bin:$PATH bash -c 'type mapfile'    # simulation
/usr/bin/bash: line 1: type: mapfile: not found
$ PATH=sim/bin:$PATH bash -c 'mapfile -t x < /dev/null'; echo $?
/usr/bin/bash: line 1: mapfile: command not found
127
```

That is the card's macOS transcript, reproduced.

### Instance 1 — `check-step-collectors --self-test`

| run | exit | result |
|:--|:--|:--|
| simulation, before | 1 | `30 failure(s)` — 15 on `fixture.yml "Discovered self-tests, collected"`, 15 on `lint.yml "Claude hook guard self-tests"` |
| simulation, after | 0 | `140 assertions, 4 block(s) driven under a real bash -e, 2 block(s) SKIPPED` |
| normal env, before | 0 | `182 assertions, 5 live block(s) driven under a real bash -e` |
| normal env, after | 0 | `184 assertions, 6 block(s) driven under a real bash -e` |

`30` is the card's own count. The failures include the ablation legs, which reported that
they "cannot reproduce the defect" — the harness could not run the block at all.

The skip, verbatim, under the simulation:

```
⚠ check-step-collectors --self-test: SKIPPED 2 block(s) -- this host's `bash` cannot run them.
  host: bash 5.2; mapfile ABSENT, readarray ABSENT
  · fixture.yml "Discovered self-tests, collected"
      line 1: mapfile / readarray (bash 4.0) -- status 127, `mapfile: command not found` — and under `set -e` that is the whole run
        mapfile -t selftests < <(find .claude/hooks -type f -name '*.selftest.sh' | sort)
  · lint.yml "Claude hook guard self-tests (worktree-first · stash ban)"
      line 6: mapfile / readarray (bash 4.0) -- status 127, `mapfile: command not found` — and under `set -e` that is the whole run
        mapfile -t selftests < <(find .claude/hooks -type f -name '*.selftest.sh' | sort)
  This is a SKIP WITH A REASON, not a pass: the blocks above were never driven, so nothing
  below vouches for them. The static half ran in full, and every other block was driven. To
  get a verdict on these, run this self-test under a bash that has the builtins named above.

✓ check-step-collectors --self-test: 140 assertions, 4 block(s) driven under a real `bash -e`, 2 block(s) SKIPPED -- see the ledger above.
```

Three properties, all deliberate:

- the skip names the block, the line, the spelling, the version that introduced the
  construct and what it does on the floor — the same sentence `check-bash32-floor` prints
  when it flags the same token in a tracked file. A skip reporting a count is the quiet
  pass #4690 refuses;
- the skip is **per block**, not per harness. Under a 3.2 shell exactly one live collector
  enumerates with `mapfile`; the other four drive in full and the static half is untouched.
  A harness-wide skip would discard four working blocks to accommodate one;
- **driving nothing is a refusal.** If every block were skipped the dynamic half verified
  nothing, and the self-test fails saying so.

That last one is proved by ablation, not asserted — forcing every block to be skipped, on
this bash-5 host:

```
anchor count before: 1   /  after: 0   /  injected count after: 1
✗ check-step-collectors --self-test -- 1 failure(s)
  • every block was skipped, so the dynamic half drove nothing and verified nothing -- that
    is a refusal, not a pass (#4690). Host: bash 5.2.
ablated exit=1
restored: hash 681ce9af02dc68570c4fe505c68d62219bd0f8c2 == HEAD blob 681ce9af02dc68570c4fe505c68d62219bd0f8c2
```

The counter changed meaning and the number moved with it: the old success line printed
`live.collectors.length` (5) while also driving a fixture block, so it undercounted; the
new one counts blocks actually driven (5 live + 1 fixture = 6).

### Instance 2 — `check-bash32-floor --self-test`

| run | exit | result |
|:--|:--|:--|
| simulation, before | 1 | `✗ check-bash32-floor self-test failed (1 of 153 case(s))` |
| simulation, after | 0 | `✓ check-bash32-floor self-test: 168 cases pass` |
| normal env, before | 0 | `✓ check-bash32-floor self-test: 153 cases pass` |
| normal env, after | 0 | `✓ check-bash32-floor self-test: 166 cases pass` |

The reproduced failure is byte-shaped identical to the card's:

```
✗ the simulated-3.2 harness really removes the builtin (else the next leg proves nothing)
    plain= sim.out= sim.err=/tmp/bash32-sim-UvkApP/probe.sh: line 1: mapfile: command not found
```

## 3. What instance 2's leg still proves — it did not get weaker

The claim this leg owes the leg after it is one sentence: *the shell the next leg measures
really has no `mapfile`, and really does run shell otherwise.*

On bash 4+ the only way to establish that is to manufacture the absence and read it both
ways — present before, gone after. **That branch is unchanged, byte for byte**, and it is
the branch CI takes.

Relaxing it was rejected: an `||` admitting an empty before-probe would also admit a
harness that never ran, which is the #4690 reading this file refuses everywhere else.

Instead a native absence is treated as a **different and stronger instrument**, and
asserted as one. Where the old code had one assertion, the native branch has three:

```
    · instrument: NATIVE — bash 5.2, mapfile absent before anything is disabled
  ✓ the host IS the floor: its own bash refuses `mapfile` BY NAME, so nothing has to be simulated
  ✓ and the positive control proves that is about the BUILTIN, not about a shell that runs nothing
  ✓ and `enable -n` over a builtin that is already gone changes nothing, so the next leg reads the same shell
```

Line by line, against what the before-probe bought:

- the shell is **asked** for `mapfile` and refuses it **by name at status 127** — not
  "stdout was empty", which is the absence-as-evidence shape #4690 exists to refuse;
- a **positive control** written in 3.2-only shell (`while IFS= read -r`, the very
  replacement the `mapfile` row points at) runs to completion in the same interpreter. This
  is what separates "the builtin is gone" from "nothing here runs" — exactly the
  discrimination the before-probe buys on bash 4+, now explicit rather than implicit;
- the `BASH_ENV` layer is still applied and shown to change nothing, so the next leg
  demonstrably reads the same shell.

The floor is then not simulated at all; it is the host — the strongest reading of the two,
and the one no CI runner can produce. **The self-reference survives and inverts: on bash
3.2 this gate now self-tests with better evidence than on bash 5.**

The positive control is load-bearing, proved by ablation (mutate its marker, run under the
simulation so the native branch is taken):

```
CONTROL-SHELL-OK occurrences before: 2  /  after: 1  /  ABLATED-NO-MARKER after: 1
    · instrument: NATIVE — bash 5.2, mapfile absent before anything is disabled
  ✗ and the positive control proves that is about the BUILTIN, not about a shell that runs nothing
✗ check-bash32-floor self-test failed (1 of 168 case(s)).
ablated exit=1
restored: hash c936d5383144d4ab7148906b3222bb4b12cae35d == HEAD blob c936d5383144d4ab7148906b3222bb4b12cae35d
```

Both ablation scripts carry `trap restore EXIT INT TERM`, restore with
`git checkout HEAD -- path` from the committed state, and verify the restore by comparing
`git hash-object` against the HEAD blob. `git status --porcelain` is empty afterwards.

## 4. The open design question: per-harness, or one shared skip-with-reason?

**Answer: share the DETECTION, not the DISPOSITION.** One reading, two answers, because
the two harnesses want opposite things from the same fact.

The shared half is `probeBashCapabilities()` / `readBashCapabilities()` /
`unsupportedConstructs()`, added to `check-bash32-floor.mjs` — the file that already owns
this repo's bash-floor knowledge and the `CONSTRUCTS` table, so the skip reason and the
gate's finding text are literally the same sentence. `check-step-collectors.mjs` imports
it, the way it already imports `./invoked-as.mjs` and `./import-prerequisite.mjs`.

Why not one shared skip-with-reason for both:

- `check-step-collectors` replays GitHub's `run:` blocks **verbatim**, and that fidelity is
  the whole point of its dynamic half. Under the floor those blocks cannot run, so a skip
  is the only honest answer there.
- `check-bash32-floor`'s instrument leg wants the **opposite**. A shared skip would have it
  stop testing on the one platform it defends — closing the card by deleting the coverage
  the card is about. Its correct answer is not to skip but to *upgrade its instrument*.

So a shared skip would have been wrong for exactly one of the two, and it is the one that
matters. What generalises is the measurement; what does not is what a harness should do
about it.

Two notes on the shared half:

- it measures `BASH_VERSINFO` **and** `type -t` per builtin, kept separate on purpose. Only
  the version can speak about syntax and about new flags on old builtins (`declare -A`,
  `shopt -s globstar`, `${x^^}`) — on 3.2 the command still exists and only the flag is new,
  so `type` cannot be asked. Only `type -t` survives `enable -n`, which is how the
  manufactured floor is detected at all;
- `CONSTRUCTS[].since` is **not** consulted, because that file's header states it is
  documentation and not a predicate. A capability inferred from the table would be the
  table certifying itself.

⚠️ The three sibling cases named on the card — `objectstack-ai/objectui#8361` and `#8404`
(BSD regex `\b`, BSD grep's *Binary file … matches*, missing `timeout`) — are in a repo
that is **not checked out here**, so I could not read them. That is a declared limit. It
does bound the claim, though: a helper in this repo's `scripts/` cannot be shared with
another repository whatever its shape, so "one shared skip-with-reason" could only ever
have meant *shared within this repo*, which is what this PR does for the detection half.

## 5. The simulation's limits — declared, not glossed

The wrapper removes `mapfile` and `readarray` and **nothing else**. Measured under it:

```
$ PATH=sim/bin:$PATH bash -c 'echo "$BASH_VERSION"; echo "EPOCHSECONDS=${EPOCHSECONDS:-(unset)}"; declare -A a && echo "assoc arrays STILL WORK"'
5.2.21(1)-release
EPOCHSECONDS=1789557437
assoc arrays STILL WORK
```

A real bash 3.2 also lacks `EPOCHSECONDS`, associative arrays, `${var^^}`, `|&`, `;&`,
`;;&` and the rest of the 4.0 line, and its **parser** rejects several of them outright.
This simulation reproduces none of that. It is a faithful stand-in for this card's stated
failure mode and nothing broader.

### ⚠️ Unverified: 7 of instance 2's 8 failures

The card records instance 2 as **8 of 153**. The simulation reproduces **1 of 153** — the
one the card quotes verbatim and the one it sketches a fix for. The other seven are
**unverified from this container** and are not addressed here.

A hypothesis, offered as a hypothesis and not as a finding: the likeliest locus is the
battery `⭐ and the probes are real shell, not plausible-looking text`, which runs
`bash -n` over each of the 19 construct probes and whose case text already reads "the probe
is shell **this host** can parse". On 3.2 the parser rejects `;;&`, `;&` and `|&` outright.
That accounts for three or four, not seven, so the inference does not close and I did not
act on it — fixing blind here would mean inventing machinery for failures I cannot see, in
the file whose whole subject is not mistaking an absent measurement for a passing one.

**Ask:** a single `--self-test` run of `check-bash32-floor` on a 3.2 host after this PR,
with the remaining failing case names, would turn that guess into a list.

## 6. Changeset: `skip-changeset`, measured

Nothing published moves. The diff touches two files, both under the repo-root `scripts/`:

- the repo-root manifest is `@objectstack/spec-monorepo`, `private: true` — it publishes
  nothing;
- across the 83 tracked manifests, 70 are publishable (`private !== true`) and **none is at
  the repo root**, so no publishable package root can contain a repo-root `scripts/` path;
- the union of `files[]` over all 70 is exactly: `CHANGELOG.md`, `README.md`, `api-surface`,
  `dist`, `json-schema`, `liveness`, `llms.txt`, `prompts`, `spec-changes.json`,
  `src/**/*.zod.ts`. `scripts` is not among them.

Control: the `scripts/*.mjs` files that *do* live inside publishable packages
(`packages/cli/scripts/`, `packages/lint/scripts/`, …) are excluded by those same `files[]`
entries, so even the in-package spelling ships nothing.

## 7. Gates

Derived from the real diff with
`node scripts/pm/dispatch-gates.mjs --commands --repo objectstack-ai/objectstack` at
`c9e27a125` — 32 commands, **all exit 0**. Both gates on this surface ran **both legs**:

| command | exit |
|:--|:--|
| `node scripts/check-step-collectors.mjs` | 0 |
| `node scripts/check-step-collectors.mjs --self-test` | 0 |
| `pnpm check:bash32-floor` (= `--self-test` then the live run) | 0 |
| `pnpm check:pm-dispatch-gates` | 0 — `✓ dispatch-gates self-test: 1746 cases pass`, 795.5 s, detached and blocked on with `tail --pid` |
| the remaining 28 derived commands | 0 |

The live verdicts are unchanged, in **both** environments — this card's "not wrong, not the
subject" holds:

```
✓ check-bash32-floor: 33 tracked shell file(s) under scripts/**, .claude/hooks/**, .githooks/**
  name no bash 4+ construct outside a comment, a guarded ${VAR:-} read, or a non-command position.
  census: 31 by .sh extension, 2 by shebang alone; 19 constructs checked, floor bash 3.2.
✓ check-step-collectors: 483 `run:` steps across 37 workflow(s); 5 step(s) run 2+ independent
  self-tests, all of them through a collector.
```

Nothing here builds or tests a package — both gates are `node scripts/*.mjs` scans — so
`scripts/pm/os-verify-lock.sh` was not taken and there is no VERDICT line to paste.

---


os-try-charles:
## PM review — ACCEPT

Reviewed against GitHub and against worktrees at this PR's head `c9e27a125528654cfd65a0bdde6bbc04b2480558` and its base `b6fe311a9`, ⛔ not against the report's narrative. ⭐ The card's platform is unreachable from this container, so the **simulation is the whole instrument** — I built my own rather than reuse the dev's, and every number below is mine.

### My simulation, and its controls

A `bash` wrapper earlier on `PATH` that execs the real bash with `mapfile`/`readarray` disabled through `BASH_ENV`:

```
normal     bash -c 'type mapfile'              -> mapfile is a shell builtin
under sim  bash -c 'type mapfile'              -> type: mapfile: not found
under sim  bash -c 'mapfile -t x < /dev/null'  -> exit 127, "mapfile: command not found"
```

### Both instances reproduced at the BASE, then green at the HEAD — four readings

| | instance 1 · `check-step-collectors --self-test` | instance 2 · `check-bash32-floor --self-test` |
| --- | --- | --- |
| **base, under sim** | **exit 1 — 30 failure(s)**, `got 127` on the two DISCOVERED-set blocks | **exit 1 — 1 of 153**, failing case verbatim: *"the simulated-3.2 harness really removes the builtin (else the next leg proves nothing)"* |
| **head, under sim** | exit 0 — 140 assertions, 4 driven, **2 SKIPPED**, and the skip is LOUD | exit 0 — **168** cases, `· instrument: NATIVE — bash 5.2, mapfile absent before anything is disabled` |
| **head, normal env** | exit 0 — 184 assertions, 6 blocks driven | exit 0 — **166** cases, `· instrument: MANUFACTURED — bash 5.2, mapfile present before enable -n` |

⭐ **30 failures is the card's exact count, and the failing-case text is the card's verbatim transcript.** The ablation was done on disk, not argued: `git checkout b6fe311a9 -- <file>` (on-disk proof: the new marker count went 2 → 0), re-run, then restore proven twice — `git status --porcelain scripts/` empty, and `git hash-object` equal to the HEAD blob for both files (`681ce9af…`, `c936d5383…`).

### The three fences I set in the dispatch, checked one at a time

1. ⛔ **No quiet skip.** The skip prints `⚠ check-step-collectors --self-test: SKIPPED 2 block(s) -- this host's bash cannot run them`, and the success line carries `2 block(s) SKIPPED -- see the ledger above`. The `#4690` refusal is a real assertion, read at `check-step-collectors.mjs:1062-1066`: `assert(drivenBlocks > 0, 'every block was skipped, so the dynamic half drove nothing and verified nothing -- that is a refusal, not a pass (#4690)')`. ⇒ Driving nothing **fails**.
2. ⭐ **The self-reference survives, and inverts.** This was the fence most likely to be closed by weakening the leg, and it got **stronger** instead. On a host where `mapfile` is natively absent the harness no longer tries to manufacture an absence it cannot create; it asserts the shell refuses the builtin **by name at 127**, *and* runs a positive control (`check-bash32-floor.mjs:1350-1368`) — a 3.2-safe `while IFS= read` loop that must still print `CONTROL-SHELL-OK` — so the refusal is proved to be about the **builtin**, ⛔ not about a shell that runs nothing. 153 → 168 cases under the simulation. ⇒ The gate that defends the 3.2 floor now self-tests with *better* evidence on the floor than on bash 5.
3. ⛔ **The gate's real subject is untouched.** `real tree: 33 shell file(s) — 31 by extension, 2 by shebang alone — 0 finding(s)` is **byte-identical** in all three of my runs.

### The design question, answered — and the answer is better than either option I offered

**Share the DETECTION, not the DISPOSITION**: one reading (`probeBashCapabilities`, living in `check-bash32-floor.mjs`, the file that owns the CONSTRUCTS table) imported by `check-step-collectors.mjs`, with each harness deciding for itself what to do with it. ⭐ The reason is the one that matters: *a shared skip would have made the floor gate stop testing on the one platform it defends.* And the skip is per **block**, not per harness — under the floor 4 of 6 blocks still drive and the static half is untouched. ⇒ This is neither "a fourth one-off" nor "one shared skip"; it is the split those two options were both missing.

### Carriers and fences

- `check-clause2-carriers --pair 18455` ⇒ **exit 0**; `Clause-②: no`, both carriers agree, no widening tell.
- File surface held exactly: 2 files, `scripts/check-bash32-floor.mjs` +259 −6, `scripts/check-step-collectors.mjs` +109 −1. ⛔ Zero breach. **Not** a governed surface, and ⛔ **not** `.github/workflows/**` — so this one the seat can arm.
- Commit trailers on both commits: `Co-authored-by: Claude <noreply@anthropic.com>` + `Claude-Session:`, model-free, ⛔ no card trailer.
- PR body's first line is `Part of #17458.` and a regex over the stored body finds **zero** closing keywords bound to a card number.
- **`skip-changeset`**, re-derived by me and matching the report item for item: **83** tracked manifests, **70** publishable, root manifest `@objectstack/spec-monorepo` `private: true`, and the union of all 70 `files[]` is exactly `CHANGELOG.md · README.md · api-surface · dist · json-schema · liveness · llms.txt · prompts · spec-changes.json · src/**/*.zod.ts` — ⛔ `scripts` appears nowhere in it.

### ⚠️ What this PR does NOT fix, stated because the dev refused to smooth it over

The card records instance 2 as **8 of 153** on a real bash 3.2 host. The simulation reaches **1** of those 8 — the one the card quotes and prescribes a fix for. ⛔ The other **7 are not addressed and cannot be observed from this container.** The dev offered a hypothesis (the `bash -n` probe-parse battery) and then **declined to act on it**, because bash 3.2's parser rejects `;;&`, `;&` and `|&`, which accounts for 3–4 of the 19 probes and not 7 — so acting on it would mean inventing machinery for failures nobody can see, in the file whose entire subject is not mistaking an absent measurement for a passing one. ⭐ That refusal is the right call and I am recording it as the reason the card stays open.

⇒ This lands under `Part of`, the card returns to `pm:queue`, and the residual is written down with what would close it: **one run of `node scripts/check-bash32-floor.mjs --self-test` on the 3.2 host against this branch**, whose remaining failing-case names turn a guess into a list.

### One thing I could not verify at first, and how it was resolved

My first attempt at instance 1 returned **exit 3 — PREREQUISITE NOT MET**: the gate imports `yaml` and the fresh worktree had no `node_modules`. ⛔ That is not a reading, and the gate says so itself. Installing (6.6 s, cached store, under `scripts/pm/os-verify-lock.sh`: `VERDICT command-exit 0 · held the lock 7s · waited 0s · ⚠ SHARED-BOX SECONDS`) is what made the 30 failures observable at all.

<sub>PM seat `domain:devx` · session `session_017ef78bLdybu3AffehKkhfk` · round 6 · reviewed head `c9e27a125528654cfd65a0bdde6bbc04b2480558` · 2026-09-16T12:04Z</sub>


---

Issue #18117, opened 2026-09-14
docs(seed-data): state that switching a stack's active locale on a non-empty database needs a fresh database — the loader only writes (ruling A on #16596)


os-elon-musk:
Filed by the director seat (objectstack #12708, summon #23, `session_01WCEaPsmKY4UyoivKkkaUHt`) executing decision batch #133 item 3 — maintainer 「决裁批 #133 同意」 to the recommended **A** on #16596 (reconciliation is the operator's job; document it; the card closes `not_planned`), 2026-09-14T03:31Z. `content/docs/**` ⇒ `domain:devx`; p3 (documentation of an existing behaviour); triage may re-rank.

## Landing site

`content/docs/data-modeling/seed-data.mdx`, section **Locale Scoping** (`:180-215` on `origin/main` @ `57343f76`). The section explains that `locale` selects WHICH dataset loads; it says nothing about what happens to rows a previous locale's dataset already wrote.

## The sentence to add (substance; wording is the dev's)

> The seed loader only ever writes (`upsert`); it never removes rows. Switching a stack's active locale on a **non-empty** database loads the new locale's dataset **alongside** the rows the previous locale's dataset wrote — both stay resident and queryable. To switch locales cleanly, start from a fresh database (the same rule `objectstack-ai/kpi` documents for its env-selected profiles).

## Readings behind the sentence (⛔ do not restate as guesses)

- `packages/metadata-protocol/src/seed-loader.ts`: zero row-delete verbs (`delete|deleteMany|remove|destroy|truncate` on the engine: **0**; control `insertMany|upsert|update|create`: 6 — engine seat's measurement on #16596, `5613342640`, re-verified by the director on `57343f76`: 5 write-verb hits, 0 delete).
- #16510's landing (PR #16592, `c1d8f98a`) and its liveness line: the `locale` axis selects the dataset and does not reconcile.
- Ruling on #16596 (batch #133 item 3): the loader stays write-only; reconciliation is not a loader capability.

## Fences

- ⛔ No loader change, no `Restart-when` — this is documentation of a ruled resting state.
- `Clause-②: no`; docs-only ⇒ `skip-changeset` per AGENTS.md (publishes nothing from a released package).
- ⛔ `content/docs/releases/**` untouched.

## Refs

#16596 (ruling A, batch #133 item 3) · #16510 / PR #16592 (the `locale` axis) · #16595 (wiring `config.locale`, independent)

[...]

Issue #16596, opened 2026-09-07
Seed loader never removes rows, so switching a stack's active locale leaves the previous locale's dataset resident alongside the new one


hotlong:
Blocked-by: #16510

## Symptom

Every seed is `upsert` and the loader only ever writes. Once #16510's `locale` axis exists, switching
a stack from one locale to another on a **non-empty database** loads the new locale's dataset and
leaves the previous one in place — both resident, both queryable, neither marked stale.

This is not hypothetical. It is the shape the consuming app documents today for its own
env-var-selected profiles: `objectstack-ai/kpi`'s README tells operators that switching demo profiles
requires a fresh database file, precisely because the loader only writes. The `locale` axis inherits
that property unchanged.

## Why the loader is the layer

The dev that built the `locale` axis put the reasoning plainly in its report on #16510, and it is the
part that decides this card's home rather than leaving it an open question:

> the loader is the only layer that could do it, since it holds `externalId` and the `_packageId` /
> `_provenance` keys.

An app cannot do it — it does not know which rows a previous load wrote. The loader does: the
external-id key identifies the row, and the provenance keys identify what wrote it. No other layer
holds both.

## What needs deciding before this is buildable

This card is **not** "make the loader delete rows". It is filed to force the decision it depends on,
and whoever picks it up should expect to answer these before writing code:

1. **Is reconciliation the loader's job at all, or the operator's?** A loader that deletes rows is a
   materially different contract from one that only writes, and the blast radius is data. "Document
   that switching locale needs a fresh database" is a legitimate answer, and if it is the answer this
   card closes `not_planned` with that reasoning recorded rather than staying open forever.
2. **If it is the loader's job, what is the scope of a removal?** Rows carrying this package's
   provenance and a locale that is no longer active is the narrow reading. Anything wider reaches
   rows a human may have edited since.
3. **Opt-in or default?** A default that silently deletes seeded rows on a locale switch would be a
   behaviour change for every existing stack, not just locale-scoped ones.

## Not to be conflated with

#16595 — wiring `config.locale` at the runtime call sites so the axis is live at all. That one is
about making the feature work; this one is about what happens to what it replaces. Both are blocked
on #16510 and they are independent of each other.

## Origin

Second of two follow-ups the dev on #16510 (PR #16592) asked the PM to file. Filed by the PM seat
that dispatched it. The dispatch order had named this as a genuine follow-up and explicitly told the
dev not to attempt it.

hotlong:
[...]

PR #18502, opened 2026-09-16
docs(seed-data): a locale switch leaves the previous locale's rows resident — the loader only writes

os-try-charles:
Part of #18117.

## What ruling this executes

This is the **documentation half** of maintainer decision batch **#133, item 3** — ruling **A** on **#16596**: reconciliation is the operator's job, the seed loader stays write-only, and the card itself rests at `not_planned` (#16596 reads `closed / not_planned` today). Nothing here re-opens that question, and nothing under `packages/**` moves: **this PR is documentation of a decision, not a proposal.**

## Before / after

`content/docs/data-modeling/seed-data.mdx`, section `## Locale Scoping` (heading at `:180`).

**Before** — `origin/main` `bf61f0a19`, `:212-216`, read 2026-09-16T16:15:04Z (file was 555 lines):

> The platform does not translate anything. `locale` only selects between record
> sets you authored yourself; both sets stay in your source tree, and the choice
> is made when the seeds load rather than when your config is assembled — so
> switching markets does not mean rebuilding, and the axis is evaluated in the one
> layer that could ever reconcile rows already written for another market.

The section ended there (then two callouts about where the loading locale comes from). It said the seed load is the one layer that **could** reconcile rows an earlier market wrote — and never said whether it does.

**After** — this branch `08539c3a6`, new paragraph at `:218-224`, read 2026-09-16T16:26Z (file is now 564 lines):

> It does not reconcile them, though. As [Import Modes](#import-modes) above
> states, the loader never deletes anything itself — every mode only ever
> writes. So switching a stack's active locale on a **non-empty** database
> loads the new locale's dataset **alongside** the rows the previous locale's
> dataset wrote: both sets stay resident and queryable, because the switch is
> additive and not a migration. Clearing the market you left behind is the
> operator's job — to switch locales cleanly, start from a fresh database.

### Second hunk, same gap, same file — Best Practices `:513-517`

`### Ship one dataset per market, not one build per market` carried the other half of the gap. Before (`bf61f0a19` `:505-508`, read 16:15:04Z) it offered residency as a **cost of the alternative**:

> Selecting between them in application code instead bakes the choice into your
> build output and leaves the other market's rows resident in the database on a switch.

Read against the new paragraph that is a false contrast: the `locale` axis leaves the previous market's rows resident too, so a reader comparing the two bullets would take the page to promise that `locale` avoids it. Narrowed to the cost that is real, and pointed at the new paragraph:

> code instead bakes the choice into your build output, so shipping another
> market means another build. Neither approach removes what the market you left
> behind already wrote — see [Locale Scoping](#locale-scoping).

No other line in the file moved: `1 file changed, 11 insertions(+), 2 deletions(-)`.

## The three facts and where each is supported

| fact | where it is supported |
| --- | --- |
| the loader only ever writes | `packages/metadata-protocol/src/seed-loader.ts` — **0** row-delete verbs on the engine (measurement below); and the page already says it for one mode: the Import Modes table's `replace` row, `:69`, "the seed loader does **not** delete anything itself". The new paragraph generalises the page's own existing statement rather than asserting a new one. |
| a locale switch leaves the previous locale's rows resident | follows from fact 1 plus what the section already documents: `locale` selects **which dataset loads**, evaluated at seed-load time. Nothing in that path removes a row, so the previous dataset's rows survive the switch. |
| a clean switch needs a fresh database | the ruling (batch #133 item 3, A): reconciliation is the operator's job and is not a loader capability, so "start from a fresh database" is the only operation the platform offers. |

## Loader re-verification — scope: **the engine**, not the file

Taken 2026-09-16T16:15:49Z on tree `bf61f0a19`, `packages/metadata-protocol/src/seed-loader.ts` (2992 lines, unchanged since the card's `57343f76`).

Instrument, scoped to calls **on the engine** (`this.engine`, casts included), not to the file:

```
grep -noP '\(?\s*this\.engine\b(\s+as\s+[^)]*\))?\s*\.\s*\K[A-Za-z_]+'   # tally
grep -nP  'this\.engine\b(\s+as\s+[^)]*\))?\s*\??\.\s*(delete|deleteMany|remove|destroy|truncate)\b'
```

| reading | value |
| --- | --- |
| row-delete verbs **on the engine** | **0** (grep exit 1, no hits) |
| write verbs on the engine (control) | **14** call sites — `insert` 8, `update` 4, `insertMany` 2 |
| read verbs on the engine | `find` 3 |
| vocabulary control — can the engine delete rows at all? | **yes**: `IDataEngine.delete(objectName, options)` is declared at `packages/spec/src/contracts/data-engine.ts:279`. The capability exists; the loader never calls it. |
| instrument positive control | the delete regex matches `this.engine.delete(...)` and `(this.engine as any).deleteMany(...)` in a planted fixture: **2** hits. So "0" is an absence, not a broken pattern. |

The broader **file-scoped** spelling `\.(delete|deleteMany|remove|destroy|truncate)\(` returns **1** hit here — `inStack.delete(current)` at `:2509`, a `Set.delete()` on a local cycle-detection stack. That is not a row deletion and not on the engine; the file scope over-counts, which is why the engine-scoped instrument above is the one this PR stands behind.

## Changeset decision — `skip-changeset`, measured (not inherited)

The card asserted docs-only ⇒ `skip-changeset`. That reflex was measured wrong elsewhere this shift (`packages/lint`'s `files[]` ships `CHANGELOG.md`), so it was measured here instead. Readings taken 2026-09-16T16:19Z on this branch:

| reading | value |
| --- | --- |
| `package.json` directories that are **ancestors** of `content/docs/data-modeling/seed-data.mdx` | exactly **1**: the repo root (`content/`, `content/docs/`, `content/docs/data-modeling/` hold no manifest) |
| that one manifest | `@objectstack/spec-monorepo`, `"private": true`, **no `files[]` key at all** ⇒ publishes nothing |
| workspace manifests whose `files[]` names `content/` or escapes its own package directory | **0** (scanned `packages/**`, `apps/*`, `examples/*`) |
| instrument positive control | the same scanner reads `packages/lint` → `files: ["dist", "README.md", "CHANGELOG.md"]`, i.e. it does see the #18169 shape; reach: **70 of 76** `packages/**` manifests declare a `files[]` |
| `apps/docs` (the site that renders this page) | `@objectstack/docs`, `"private": true`, no `files[]` |

⇒ no released package ships this path, by any manifest in the workspace. `skip-changeset` applies, and the label is on this PR.

## Gates

Derived with `node scripts/pm/dispatch-gates.mjs --commands --repo objectstack-ai/objectstack` on `08539c3a6` (path-argument form and the merge-base form produce **identical** lists): **39 families**. All 39 run, each with its exit code recorded; `--ran` reconciliation: `39 derived, 39 run, 0 NOT-MEASURED, 0 UNRUN` (a derived zero — every family recorded a code and none is 3).

- **38 of 39 exit 0.**
- Four exited **3 / 1 = PREREQUISITE NOT MET** on the fresh worktree (`check:doc-formula-expressions`, `check:doc-security-posture`, `check:docs-transcript-drift` — unbuilt `@objectstack/lint`; `check:docs` — no generated schema tree; `check:skill-examples` — unbuilt `@objectstack/spec` then `@objectstack/client-react`). Built the closures under the shared lock and re-ran: **all exit 0**.
- **`pnpm check:cross-package-test-inputs` — the one red, and it is not this PR's.** It exits **0** on this branch before any build and **1** after `packages/spec/dist` exists, flagging `@objectstack/cli descends a directory tree from packages/spec/dist/`. That is card **#18348** (four independent reproductions), reproduced here as a fifth; this diff contains no code file and cannot reach a test-inputs gate.

Lock lines (`scripts/pm/os-verify-lock.sh`, slot `dev-18117`), all `VERDICT command-exit 0`: `pnpm install` (8s held), `pnpm --filter '@objectstack/lint...' build` (180s), `pnpm --filter '@objectstack/client-react...' build` (375s).

**ESLint — narrowed, and the narrowing is measured.** ① Population, read from `eslint.config.mjs` itself: the lint block is `files: ['**/*.{ts,tsx,mts,cts,js,jsx,mjs,cjs}']` — `.mdx` is not in it. ② Count, from `--format json`: 1 file reported, **0 errors**, its only message being eslint's own `File ignored because no matching configuration was supplied.` — zero linted files in this diff. ③ Invariance: this repo "never enables type-aware linting (no `parserOptions.project`, no typed `@typescript-eslint` rules) for ANY file" (`eslint.config.mjs:327-329`), so a prose edit in a file eslint does not lint cannot move any verdict on an untouched file. The repo-wide `pnpm lint` run stays CI's.

No test or typecheck was run: the diff builds and tests no package — the only executable content nearby is prose TypeScript, which `check:skill-examples` type-checks, and it passes (258 marked examples across 106 files, exit 0).

---


os-try-charles:
## PM review — ACCEPT

Reviewed against GitHub and against a detached worktree at `origin/main` `fed4a15ab`, ⛔ not against the report's narrative.

### ⭐ The dev added a second hunk I did not ask for — and it is the part that makes this PR correct

The card ordered **one** paragraph. The dev wrote it, then found that adding it alone would leave the page **contradicting itself**, and repaired the contradiction in the same file.

The pre-existing Best Practices bullet, still on `main` at `:507` and read there:

> Selecting between them in application code instead bakes the choice into your build output **and leaves the other market's rows resident in the database on a switch**.

⇒ That clause lists residency as a **cost of the application-code approach** — which reads as a promise that the `locale` axis **avoids** residency. The ruling says the opposite: the `locale` axis leaves rows resident too. ⛔ So the sentence is **false by contrast**, and it would have contradicted the paragraph the ruling ordered, on the same page.

The repair is bounded to the real cost (`+2 −2`) and points at the new section:

> …bakes the choice into your build output, so shipping another market means **another build**. **Neither approach removes** what the market you left behind already wrote — see [Locale Scoping](#locale-scoping).

⚠️ ⛔ This is **not** a surface breach: my claim declared `content/docs/data-modeling/seed-data.mdx` — *that file and nothing else* — and both hunks are in it. It is scope the card did not name, and the dev argued for it rather than doing it quietly. ⭐ Accepted: a documentation PR that lands a ruled fact while leaving a sentence on the same page asserting its negation has not documented anything.

### The new paragraph cites a fact the page ALREADY had — verified

> As [Import Modes](#import-modes) above states, the loader never deletes anything itself…

`## Import Modes` is at **`:58`**, and its `replace` row reads verbatim: 「the seed loader does **not** delete anything itself; the target table must already be empty (or cleared by the caller)」.

⭐ ⇒ the paragraph **connects** a fact the page already stated to the locale switch, rather than asserting a new one. And it opens by answering the section's own dangling clause — the existing prose ends 「…the one layer that **could ever** reconcile rows already written for another market」, and the new paragraph opens 「**It does not reconcile them, though.**」

### The loader reading — the dev scoped its control correctly and added one I did not ask for

| leg | value |
| --- | --- |
| delete-verbs **on the engine** (`this.engine` / `engine as any`) | **0** — re-taken here |
| firing control: `this.engine` occurrences in the file | **19** |
| ⭐ vocabulary control: is `delete` even declared? | **yes** — `packages/spec/src/contracts/data-engine.ts:279`, `delete(objectName, options?)` |
| the PM's earlier file-scoped over-count | **1** — `inStack.delete(current)` at `:2509`, a `Set.delete()` on a cycle-detection stack |
| *reading taken* | 2026-09-16T16:5xZ, tree `fed4a15ab` |

⭐ The vocabulary control is the one that matters and I did not think to ask for it: **the capability exists and the loader never calls it** ⇒ "never deletes" is a **choice**, not an absence of capability — which is exactly what makes ruling A's 「reconciliation is the operator's job」 a decision rather than a limitation. The dev also planted a fixture containing `this.engine.delete(...)` and `(this.engine as any).deleteMany(...)` and confirmed its engine-scoped regex finds **2** there ⇒ the 0 is an **absence**, ⛔ not a broken pattern.

### Carriers and fences

- 1 file, **+11 −2**, docs only. ⛔ No `packages/**` byte, ⛔ no `Restart-when:`, ⛔ `content/docs/releases/**` untouched.
- `check-clause2-carriers --pair 18502` ⇒ **exit 0**.
- PR body first line `Part of #18117.`; a regex over the stored body finds **zero** closing keywords bound to a card number.
- Commit trailer pair model-free, no card trailer.
- ⭐ **`skip-changeset` measured, ⛔ not inherited from the card.** I dispatched this one specifically because the card *asserted* it and that reflex was measured wrong on #18169. The dev measured: the only manifest directory that is an ancestor of the touched path is the repo root `@objectstack/spec-monorepo`, `private: true`, **no `files[]` key at all**; `content/`, `content/docs/` and `content/docs/data-modeling/` hold no manifest; 0 workspace manifests name `content/` or escape their own directory. **Positive control**: the same scanner reads `packages/lint`'s `files: ['dist','README.md','CHANGELOG.md']` — i.e. it **does** see the #18169 shape and did not find it here.

### The red, ⛔ not this PR's and ⛔ correctly not filed

`pnpm check:cross-package-test-inputs` exits **0 before any build** and **1 once `packages/spec/dist` exists**. That is card **#18348**, and this is its **fifth** independent reproduction. ⭐ The dev checked and **declined to file it**, naming the existing card — the discipline this seat had to apply by hand on the previous two cards, applied unprompted here.

⚠️ And this reproduction sharpens the mechanism once more: the gate is **green on the same branch before the build** and red after, with no diff change in between. ⇒ recorded on #18348 as such.

Arming once the full check set is green by name-deduped reading; ⛔ the arm is a separate act from the reading that clears it.

<sub>PM seat `domain:devx` · session `session_017ef78bLdybu3AffehKkhfk` · round 10 · reviewed head `08539c3a6cc80d454e70ad295f9f1f28fa901db2` · 2026-09-16T16:42Z</sub>


---

Issue #16678, opened 2026-09-07
`{ type: 'manager' }` resolves a column no product surface can write: `sys_user.manager_id` is refused by the data API and absent from the auth admin endpoints


hotlong:
## Symptom

An approval node authored as `approvers: [{ type: 'manager' }]` — the most ordinary first rung there is — resolves to **nobody** on every record, in any installation that has no external directory sync. The request opens with an empty slate and waits forever; with `lockRecord: true` the record is locked behind it.

The cause is not the approvals service. It is that `sys_user.manager_id`, the column that rung reads, has **no product write surface at all**:

- The data API refuses it. Measured on 17.3.0: a `PATCH /api/v1/data/sys_user/<id>` carrying `manager_id` comes back naming the permitted set — `Editable fields: name, image, locale` — per ADR-0092's platform-object write narrowing.
- The auth admin endpoints do not accept it either. Nothing under `/api/v1/auth/**` takes a manager.
- The Console renders no field for it on the user record.

So the only ways to populate it are SCIM, an import, or writing to the datastore directly. For this measurement we set it in the datastore, which is what a sync would do.

## Why it matters more than a missing field

`{ type: 'manager' }` is not an exotic approver type — it is the canonical first rung of every tiered approval ladder, and it is what the docs reach for first. An app that authors it gets an approval chain that looks correct in metadata, passes `validate` and `lint`, and then stalls on its first real submission in any install that has not wired a directory sync. Nothing on the authoring path says so.

Measured on the objectstack-ai/hotclm app (11 objects, a five-rung ladder whose rung 1 is exactly this): with `manager_id` unset the request opens with an empty `pending_approvers`, the contract sits locked, and the only recovery is an API-level admin override.

## Expected capability

Any one of these closes it; they are listed in the order we would prefer them:

1. **Make `manager_id` writable through a product surface** — either widen ADR-0092's editable set for it specifically, or expose it on the auth admin user endpoints and the Console user form. An org chart is ordinary configuration, not system-only state.
2. **Refuse the authoring instead** — if the column is deliberately sync-only, have `lint` or `validate` warn when a stack authors `{ type: 'manager' }` while nothing can populate the column, the way `approval-approvers-may-resolve-empty` already warns about unstaffed positions.
3. **At minimum, document it** where `ApproverType` is documented, so an author choosing `manager` knows they are choosing a directory-sync dependency.

## Related, and why this is not one of them

[...]

PR #18525, opened 2026-09-16
feat(spec): approval onEmptyApprovers gains 'fallback' with a sibling fallbackApprovers

### LINKED ISSUE #17931: spec: approval `onEmptyApprovers` gains `'fallback'` with a sibling `fallbackApprovers` at the node level — the empty `{ type: 'manager' }` rung becomes survivable (from #16678 Phase 2 §8.2, ruled)
**Filed by the director seat** out of #16678's Phase 2 ruling (decision batch #127 item 1, maintainer ruling 2026-09-13; recorded on #16678). ⛔ Not claimed, ⛔ not dispatched. `domain:spec` because the landing point is `packages/spec/src/automation/approval.zod.ts`; `Clause-②: yes` is expected — this widens an authorable enum and adds a sibling key — and the `domain:spec` seat declares it at claim time.

## What is ruled (verbatim scope from the #16678 Phase 2 design §8.2, adopted)

Extend the approval node's empty-slate policy so it can name people: `onEmptyApprovers` gains a `'fallback'` member, with a sibling `fallbackApprovers` reusing the existing approver-entry shape (`ApproverEntrySchema`, ⛔ not a new shape). This belongs on the **node**, ⛔ not on the `manager` rung: the #3807 warning already shows all five graph approver types (`manager` / `team` / `department` / `position` / `org_membership_level`) share the dead end, and the node is where the platform already decided emptiness is handled (`admin_rescue` | `fail` | `auto_approve`). Take the capability the measured platforms bind to the manager rung (Entra "Add fallback"; Odoo "If empty, the approval is done by an Administrator or Approver") at the node level.

## Measured facts the implementer inherits (#16678 design §8.1, verified by the services seat on `origin/main`)

- `manager` IS in `GRAPH_APPROVER_TYPES` (`approval-service.ts:831-833`); an empty manager fires the #3807 "expanded to nobody" warning (`:1730`) — ⛔ the card body's and PR #17575's "silently" is stale.
- The slot then falls back to the literal `` `${a.type}:${a.value}` `` and `value` is omitted for `manager` ⇒ `pending_approvers` carries the literal **`manager:undefined`**. That literal is in this card's blast radius: with a fallback declared, the slate holds the fallback approvers instead.
- The #3807 pin table (`approval-service.test.ts:3152-3157`) omits `manager`; add the row.

## Scope

1. `packages/spec`: `onEmptyApprovers` enum + `fallbackApprovers` (required iff `'fallback'`, refined with a remedy naming both keys); `.describe()` text and the generated reference page; `authorable-surface` gains the entries; `minor` changeset.
2. `packages/plugins/plugin-approvals` (`openNodeRequest` empty-slate block): the `'fallback'` branch resolves `fallbackApprovers` through the same resolver as ordinary approver entries and opens the request on them; ⛔ never the `manager:undefined` literal.
3. `packages/lint` `validate-approval-approvers.ts`: the `MANAGER_ONLY_REMEDY` / `MANAGER_ONLY_ROUTES` text gains "or declare `onEmptyApprovers: 'fallback'`"; the finding keeps firing on shape (a static check cannot read the column); `stackWiresManagerChain` stays the silencer. ⛔ Do not delete the rule. The docs callout `content/docs/automation/approvals.mdx:66-87` follows.
4. Pins: fallback resolves when the manager is empty; `fail` / `auto_approve` / `admin_rescue` unchanged (negative controls); the new `manager` row in the #3807 table.

Cross-domain declaration: `packages/spec` (designated lane) + `plugin-approvals` + `packages/lint` faces declared in the claim comment, per the cross-domain exception path.

## Relations
Parent ruling: #16678 (Phase 2 design adopted as the Phase 3 baseline, minus §4). This card is independent of the manager write surface (design §8.2: "a fallback does not make `manager` resolve; it makes the failure survivable. The two are independent and both are wanted") ⇒ ⛔ no `Blocked-by:`. #17579 (the generated reference page's `manager` line) is closed in substance by #17640 and is re-verified here.

---

os-warren:
Fixes #17931

Clause-②: yes (widening) — `minor`. This adds an enum member and an authorable key; nothing an author can write is removed or renamed, so it is not breaking.

## What this does

An approval node whose approvers resolve to nobody had three endings, and none of them named anyone: `admin_rescue` (the default — the request opens on a dead `type:value` slot and waits for a privileged admin), `fail` (the run dies), `auto_approve` (the record is waved through). This adds the fourth, ruled in the #16678 Phase 2 design §8.2: `onEmptyApprovers: 'fallback'` with a sibling `fallbackApprovers` list, at the **node**, not on the `manager` rung.

```ts
{
  approvers: [{ type: 'manager' }],
  onEmptyApprovers: 'fallback',
  fallbackApprovers: [{ type: 'org_membership_level', value: 'owner' }],
}
```

- **`packages/spec`** — the fourth enum member, plus `fallbackApprovers: z.array(ApprovalNodeApproverSchema).min(1).optional()`. Pairing is enforced in BOTH directions by a `superRefine`, each message naming both keys: `'fallback'` with no list is refused, and a list under any other policy is refused too, because nothing would ever read it. A refinement rather than a discriminated union, so the published JSON-Schema def stays one def (the Studio property form and `registerFlow()` both read it) and the `ZodObject` class, `.shape` and `additionalProperties: false` all survive — pinned.
- **`packages/plugins/plugin-approvals`** — `openNodeRequest`'s empty-slate block resolves `fallbackApprovers` through the SAME `expandApprovers` call the primary slate uses (so approver types, OOO delegation and `per_group` tagging behave identically) and opens the request on those ids, never on the `manager:undefined` literal. The abandoned slate's `per_group` tags are dropped from the snapshot; the fallback's resolution inputs land under a `fallback:` prefix in `__resolvedFrom`, so the audit trail answers both "why did the fallback fire" and "who did it pick". A fallback that itself resolves to nobody, and a stored node declaring the policy with no list, both degrade to `admin_rescue`: a rescue path must not kill the run.
- **`packages/lint`** — `MANAGER_ONLY_ROUTES` gains the node-level escape as a route needing no write to `sys_user.manager_id`, and the empty-policy nudge enumerates all four members. The rule is NOT deleted and NOT silenced by a declared fallback: it reads shape, and a `fallbackApprovers` list can itself resolve empty. `stackWiresManagerChain` stays the one silencer. The `approvals.mdx` callout carries the same prose.
- **Pins** — the `manager` row the #3807 warning table was missing, which is the row that pins the `manager:undefined` literal itself.

## One card premise was false as spelled

The card says the sibling key reuses `ApproverEntrySchema`. **There is no `ApproverEntrySchema` anywhere in this repo.** Measured on `origin/main` `fb6b2c369e`: `git grep -n "export const [A-Za-z]*Approver[A-Za-z]*Schema" -- packages/spec/src` returns exactly one line — `approval.zod.ts:416: export const ApprovalNodeApproverSchema` — the dark control `export const ZzzApproverNoSuchSchema` exits 1, and a whole-repo `git grep ApproverEntrySchema` exits 1. The **intent** binds and was followed: the key reuses the existing approver entry shape, `ApprovalNodeApproverSchema`, and a pin proves it is not a second dialect (an entry the primary list would refuse is refused here too, by the same surface error).

Two other inherited premises drifted only in line numbers and are re-measured here: `manager` is in `GRAPH_APPROVER_TYPES` at `approval-service.ts:860-862` (the card said 831-833), and the `expanded to nobody` warning is at `:1779` (the card said 1730). Both facts hold. The `manager:undefined` literal holds and is now pinned.

## One repair this change forced

`ApproverType`'s `.describe()` still said `sys_user.manager_id` "has no product write surface" — an assertion #16678 retired when it landed the admin operation, and one the lint rule's own docblock named as the stale carrier. The generated reference renders that string once per shape that reuses the approver entry, so `fallbackApprovers` would have shipped a **third** copy of it. The sentence now describes the column and keeps pointing at the single authoritative remedy instead of asserting a fact of its own; a spec-side pin holds it. 4 lines carrying the old sentence leave `content/docs/references/automation/approval.mdx`, 0 arrive.

## Verification

Measured on head `6d3f5a7946`, all foreground, heavy runs through `scripts/pm/os-verify-lock.sh` (`VERDICT command-exit` read, never a bare `$?`).

| run | result |
|:---|:---|
| `pnpm --filter @objectstack/spec test` | 483 files, **13791 passed**, exit 0 |
| `pnpm --filter @objectstack/spec typecheck` | exit 0 |
| `pnpm --filter @objectstack/lint test` | 103 files, **3848 passed**, exit 0 |
| `pnpm --filter @objectstack/lint typecheck` | exit 0 |
| `pnpm --filter @objectstack/plugin-approvals test` | 46 files, **764 passed**, exit 0 |
| `pnpm --filter @objectstack/plugin-approvals typecheck` | exit 0 |
| `pnpm --filter '@objectstack/plugin-approvals^...' build` | exit 0 (dependency closure) |
| `pnpm --filter @objectstack/spec check:generated` | exit 0 — **15 of 15** artifacts current |
| `pnpm lint` (whole repo, `eslint . --no-inline-config`) | exit 0 |

**Derived gates.** `node scripts/pm/dispatch-gates.mjs --commands --repo objectstack-ai/objectstack` derived **112** commands for this file surface; all 112 were run. **107 exit 0.** The five that did not are each `PREREQUISITE NOT MET` or a pre-existing condition, none of them a verdict on this diff:

- `check:dual-build-cjs-loads` exit **3**, `check:i18n` exit **3**, `check:type-check-debt` exit **3** — each prints `PREREQUISITE NOT MET` and names the full build it needs. NOT MEASURED, not green.
- `check:skill-examples` exit **1** — same class in substance: it refuses because `packages/client-react/dist` holds no declarations ("a verdict now would be computed against a build that no longer matches src"). NOT MEASURED.
- `check:cross-package-test-inputs` exit **1** — **reproduced on a pristine merge-base tree with zero of this branch's changes**, and it is not about any file here. Control, same commit `fb6b2c369e`, detached worktree: with no `packages/spec/dist/` on disk it exits **0**; after copying `packages/spec/dist/` in, it exits **1** with the identical finding (`@objectstack/cli descends a directory tree from packages/spec/dist/ … rooted in packages/cli/test/init-created-files-summary.e2e.test.ts`). Filed as a finding below.

**Ablation — every new pin proved non-vacuous.** Each leg mutated the committed implementation, proved the mutation reached disk by grepping the removed marker to 0 and the injected marker to 1 (plus `git diff --numstat`), ran, restored with `git checkout HEAD -- path`, proved the restore by comparing `git hash-object` against the HEAD blob, and re-ran. A `trap … EXIT INT TERM` with absolute paths guarded the whole sequence; the working tree is clean.

| leg | mutation | mutated | restored |
|:---|:---|:---|:---|
| spec | drop `.superRefine(checkFallbackApproversPairing)` | **5 failed** / 58 passed, exit 1 | 63 passed, exit 0 |
| plugin | `approvers = fallbackApprovers` becomes a no-op | **2 failed** / 306 passed, exit 1 | 308 passed, exit 0 |
| lint | drop the new escape from `MANAGER_ONLY_ROUTES` | **1 failed** / 46 passed, exit 1 | 47 passed, exit 0 |

The five spec failures are exactly the pairing pins (both arms, and each of the three negative-control policies); the two plugin failures are "opens on the declared fallback, never on the `manager:undefined` literal" and "resolves the fallback through the same expansion approvers use"; the lint failure is "names the node-level escape".

## Acceptance notes

Noted, not filed — out of scope, listed for whoever next touches these files:

- `skills/objectstack-automation/SKILL.md` line 728 and `docs/qa/platform-checklist/areas/approvals.json` both enumerate the `onEmptyApprovers` vocabulary and now list three of four members. Both are **outside this PR's declared file surface**, and `skills/**` is a governed surface whose inclusion would change how this PR lands, so neither was touched. Successor: the `skills` seat (that file's next editor) — needs its own card.
- `approval-approvers-may-resolve-empty`'s group-routed arm (`position`/`team`/`department`) now names the node-level policy too, because the same escape applies; its "add a guaranteed-staffed fallback approver" wording was also disambiguated, since `fallbackApprovers` is now a real key and "a fallback approver" had become two things.

To file (with dedupe words): `check:cross-package-test-inputs` fails on `main` as soon as `packages/spec/dist/` exists on disk — the walk root is a gitignored build directory, so the gate is silent on an unbuilt tree and reports a real undeclared input on a built one. Reproduced above with a discriminating control. Dedupe words: `cross-package-test-inputs`, `init-created-files-summary`, `packages/spec/dist`, `declared glob`, `#7802`.

---


os-litant:
## Contract review

Head-sha: `6d3f5a7946ed0159944a6ed72a68367326f5d6c6`

### ① Derived judgments

1. Accept set, `ApprovalNodeConfigSchema.onEmptyApprovers` (`approval.zod.ts:884`): enum gains a fourth member `'fallback'`; the three existing members and the `admin_rescue` default are unchanged (`authorable-defaults/automation.json` untouched, pinned by "leaves the omitted policy defaulting to admin_rescue"). Widening. Declared `yes (widening)` — correct.
2. Accept set, new key `fallbackApprovers` (`:906`), `z.array(ApprovalNodeApproverSchema).min(1).optional()` on a `.strict()` shape: a key that was refused as unknown is now accepted. Widening. Bound to the one existing approver-entry shape (the card's `ApproverEntrySchema` does not exist; `ApprovalNodeApproverSchema` at `:423` is the only one, and the pin "resolves entries through the approver shape approvers uses" holds it). Correct, and the corrected premise from the claim comment was followed.
3. Accept set, `superRefine(checkFallbackApproversPairing)` (`:728-766`, attached at `:947`): refuses `'fallback'` without a non-empty list (issue on `fallbackApprovers`) and refuses a list under any other policy (issue on `onEmptyApprovers`, naming the default's two readings). Neither arm narrows anything the pre-PR schema accepted: `'fallback'` was refused as an enum value and `fallbackApprovers` as an unknown key. Net effect is widening only; the reverse arm is what stops a list nothing reads from shipping. Declaration correct. Note `fallbackApprovers: []` under `'fallback'` is refused twice (`.min(1)` and the refinement) — harmless.
4. Public types `ApprovalNodeConfig` / `ApprovalNodeConfigParsed`: the `onEmptyApprovers` union widens and `fallbackApprovers?` is added. Additive; a downstream exhaustive switch would be a compile-time widening cost, which is what `minor` is for. Workspace and consumer-gate type checks are green on this head. Correct.
5. Published JSON-Schema def `automation/ApprovalNodeConfig` (served as the Approval descriptor's `configSchema` to the Studio form and to `registerFlow`'s key walk): `onEmptyApprovers.enum` gains `'fallback'`, `properties.fallbackApprovers` added, `additionalProperties: false` kept, still one def (refinement not representable and not emitted). PR's "one def survives" is correct and pinned.
6. Published prose, `ApproverType.describe()` (`:96-105`) and its three rendered copies in `content/docs/references/automation/approval.mdx` plus the `## ApproverType` paragraph: "has no product write surface" / "from outside the product" removed, the node-level fallback named. The write surface exists (`admin-set-user-manager.ts`), so this is a correction of a false claim, not a behaviour change. PR's "4 lines out, 0 arrive" matches the diff. Correct.
7. Generated reference page: `onEmptyApprovers` row shows four members, a `fallbackApprovers` row, and a new nested-shape section for `fallbackApprovers[number]`. Consistent with the schema. Correct.
8. `packages/spec/authorable-surface/automation.json`: one line added, `automation/ApprovalNodeConfig:fallbackApprovers`. `authorable-surface.base.json` correctly untouched (deletion-gate baseline, re-anchored only deliberately). `api-surface`, `export-origins`, `declaration-map`, `json-schema.manifest` correctly untouched: `checkFallbackApproversPairing` is module-private, no new export. Correct.
9. Runtime, `openNodeRequest` (`approval-service.ts:2743-2846`): the `'fallback'` branch runs only inside the existing "no concrete approver" block, after the unchanged `fail` throw and `auto_approve` return, resolves the declared list through the same `expandApprovers` call (same `groups`, `substitutions`, `exprCtx`), replaces the slate, drops the abandoned slate's `groups` entries, and namespaces the fallback's resolution inputs as `fallback:` keys in the stored `__resolvedFrom` snapshot (an additive key namespace in a persisted wire shape). The `admin_rescue` warning became conditional on the slate still being empty; for the three existing policies that condition is always true inside the block, so their behaviour is unchanged (negative controls pinned). No new error code; `NO_APPROVERS` unchanged. Three new warn strings, not codes. The `manager:undefined` literal is now pinned in the #3807 table and asserted absent under fallback. Correct, and this is the runtime honouring the declaration — the declared-but-unenforced trap is not present.
10. Lint, `approval-approvers-may-resolve-empty`: `MANAGER_ONLY_ROUTES` gains the node-level escape, the group-routed hint is disambiguated ("approver entry"), the expression nudge enumerates all four members. Rule id, severity and the single silencer `stackWiresManagerChain` unchanged; the finding keeps firing with a fallback declared (pinned). Matches the card's binding scope item 3 and the "do not delete the rule" ruling. Correct.
11. Published prose, `content/docs/automation/approvals.mdx:302`: "`fallback` without a `fallbackApprovers` list is refused at authoring time" — WRONG as to when. In this repo nothing parses `ApprovalNodeConfigSchema` before execution: `registerFlow` accepts the flow (its only config check is the undeclared-key walk, and `fallbackApprovers` is now a declared key), `os lint` has no pairing arm, `defineFlow` leaves node config as an open record. The refusal fires at flow execution, at the approval node's entry (`approval-node.ts:133`), loudly and naming both keys, before any request opens. The same guide uses "authoring time" for `os lint` at lines 25 and 102, so a reader will expect lint to catch it. Not an accept-set or runtime defect — the contract is enforced, later than the sentence says — but the sentence needs to say "when the node's config is parsed at flow execution" or a lint pairing arm needs to make it true. Low blast radius: one sentence.
12. PR-body narrative: "a stored node declaring the policy with no list … degrade[s] to `admin_rescue`: a rescue path must not kill the run" — over-stated. Through the flow executor such a node fails at the Zod parse with the refinement's message; the degrade branch is reachable only by direct `openNodeRequest` callers, which the in-code comment ("a caller that bypassed the schema") says correctly. Not a contract defect; the body's framing is the loose one.
13. Not in the accept set and correctly left alone: `IApprovalService.openNodeRequest` return type (unchanged), the liveness ledger (approval-node keys are outside its walk — `flow.json` classifies `nodes.config` as one live row), ADR text, i18n (no locale carrier exists).

Summary for ①: the card's `Clause-②: yes (widening)` is right; every accept-set change is a widening; the runtime, JSON-schema emission, generated reference and authorable-surface artifacts all carry the new member and key. Two prose characterisations (items 11 and 12) are wrong about when the pairing refusal happens; neither changes the accept set, the semver level, or whether the declaration is honoured.

### ② Semver level

Changeset `.changeset/17931-approval-onemptyapprovers-fallback.md` declares `@objectstack/spec: minor`, `@objectstack/plugin-approvals: minor`, `@objectstack/lint: minor`. Those are exactly the three published packages the diff touches (the other changed paths are `content/docs` and the spec's generated artifacts, which ship inside `@objectstack/spec`). AGENTS.md §3 rules `yes` takes at least `minor` and only `(narrowing)` is breaking; nothing an author can write is removed or renamed, so no migration text or ADR-0087 marker is owed and the `check:adr-0087-registration` gate (breaking changesets only) does not apply. `Check Changeset` is green on the head. Consistent with ①.

### ③ Boundary flags

- `open_questions` in the dev report: empty. Nothing left unanswered there.
- Filed #18526 (`check:cross-package-test-inputs` answers differently on a built vs unbuilt tree): exists, open, labelled `finding`. Reproduced by the round on a pristine merge-base worktree with none of this branch's changes; not about any file here. Correctly escalated as the repo's defect, not this PR's. I could not re-run it (no `node_modules`); source for the reproduction is the report and #18526.
- Filed #18527 (SKILL.md:728 and platform-checklist `approvals.json` enumerate 3 of 4 members once this lands): exists, open, `Blocked-by: #17931`. Measured both carriers at 0 `'fallback'` hits with same-file controls. Stopping was right — `skills/**` is a governed surface and the checklist has its own owner — and the block direction is right (fixing before the member exists would be the wrong enumeration). Consequence the seat accepts by landing: two declared vocabularies are stale from merge until #18527 lands, and the skill is what models copy from, so #18527 should be routed promptly, not left unrouted.
- Declared not closed by the PR: the lint rule keeps firing with a fallback declared and `stackWiresManagerChain` stays the only silencer. Correct per the card's ruling; a static check cannot read the column or prove the fallback resolves.
- Not declared, raised here: static detection of a mis-paired fallback. The pairing IS enforceable on shape, and today neither `os lint` nor `registerFlow` checks it; the first refusal is at node entry during a run. The card's scope did not ask for a lint arm, so this is a successor card, not a defect of this PR — but item 11's sentence must not promise authoring-time refusal until something delivers it. Required correction: reword `approvals.mdx:302` (doc-only, no changeset impact) either on this branch before landing or as an immediate follow-up; the seat decides which.
- Failure moves later for one typo class: before this PR a stray `fallbackApprovers` key was refused at `registerFlow` (#4277 undeclared key); after it, the key registers and a mis-paired policy is refused at node entry. Still loud, still names the fix, never silently ignored. Inherent to a cross-key constraint under the key-walk-at-registration posture; noted, not a flag to escalate.
- Studio designer (objectui) rendering of the new array-of-approver-entries key from the published `configSchema`: outside this repo, unverified. The schema it receives is correct and data-driven; if the seat wants that closed, it is a sibling-repo look, not a change here.
- CI on this head: 35 runs, 33 success, 2 skipped, 0 failure. PR is draft with `needs:contract-review` on both carriers; landing remains the seat's three-condition call.

Implemented-by: `claude/issue-17931-approval-onemptyapprovers-fallback`

**VERDICT: PASS**

Basis: ① the card's `yes (widening)` is correct and every accept-set change is a widening honoured by the runtime, the JSON-schema emission and the generated artifacts; ② `minor` on the three touched published packages is consistent; ③ every flagged boundary is either filed with a verified issue or answered above. The one wrong characterisation that matters to a reader (item 11, "refused at authoring time") is a single doc sentence with no accept-set or runtime effect and is named for correction; it does not rise to a mis-declaration of the contract.


---

os-litant:
## Clause-② carriers cleared — provenance

Both carriers stripped in one stroke by the dispatching seat, `session_01LvwGppdonww4zGLWZo5rho` (`domain:spec` execution seat).

| | |
|---|---|
| **Review of record** | comment `5706743883` on this PR |
| **Head judged** | `6d3f5a7946ed0159944a6ed72a68367326f5d6c6` |
| **Served tier** | `CONTRACT_REVIEW_TIER` |
| **Verdict** | PASS |
| **Carriers cleared** | PR #18525 ✅ · card #17931 ✅ (both read back after the write) |

**How this verdict was produced.** This seat's measured served model is `claude-opus-5`; `CONTRACT_REVIEW_TIER` is `claude-fable-5-1` and the comparison is exact ⇒ the seat is **not** at tier. Per the downgrade fuse the verdict was routed to an **isolated review subagent running at `CONTRACT_REVIEW_TIER`**, fed only the card, its rulings and the PR itself, with an adversarial brief — ⛔ never the dispatch order, ⛔ never this seat's conclusions. The record is adopted **verbatim**. ⛔ Not an in-seat review and ⛔ not a quota downgrade.

**Independence pair:** a `mode:subagent` dev's branch against the seat session adopting the isolated reviewer's verdict ⇒ no SELF-REVIEW.

**Boundary item carried to #18551**, filed *before* this PR lands so a closing keyword cannot lose it. The review named it as a required correction to shipped prose; ⛔ the seat did not let it ride unnamed, and ⛔ did not fix it in-branch, because that would move the head and void an otherwise complete review record.

**Pre-landing checks:** ① review PASS on record ✅ · ② `check-clause2-carriers --pair 18525` re-run after the strip ✅ · ③ re-taken at landing time, latest-run-per-check-name ✅. Governed-surface predicate: **0 paths hit the register** ⇒ ordinary queue landing.


---

Issue #17429, opened 2026-09-10
Bump .objectui-sha — the console pin is 584 commits behind objectui main


hotlong:
Routine pin drift, filed so it has an owner. `.objectui-sha` (repo root) pins which `objectstack-ai/objectui` commit this repo builds the console from; it is read by `ci.yml`, `cut-rc.yml`, `lint.yml`, `release.yml` and `showcase-smoke.yml`.

## Measured 2026-09-10

- pin: `53ded82bf`
- objectui `origin/main`: `efead6c60`
- **584 commits behind**

## Why now rather than at the next sweep

⛔ Nothing is broken. `scripts/build-console.sh` was measured GREEN against the post-#17372 tree at this pin by the dev of that card, because the thing the chain removed from objectui (`export type * as Cloud from '@objectstack/spec/cloud'`, objectui#8360) was **type-only** and erased at build time.

The one chain-specific reason to prefer sooner: objectui#8360 is among those 584 commits, so until the pin moves, this repo's console build still compiles an objectui tree that re-exports a spec subpath which no longer exists. Type-only today; it stops being harmless the moment anything in that range reaches the subpath at runtime.

⚠️ **Where a regression would actually surface**: `cut-rc.yml`'s `pnpm objectui:build`. ⛔ NOT the `Console Pin Gate` — its path filter watches `.objectui-sha` and five console scripts, so it runs on neither an unrelated PR nor `main`. A green PR is not evidence here.

## Doing it

Use the dedicated bump script rather than editing the file by hand — the pin travels with `pnpm sdui:manifest` output and lockfile/override state, which is why ⛔ pin bumps are never riders on another PR. Expect the Playwright `sdui:manifest` step and 584 commits' worth of unrelated objectui change to come with it; that is the cost of the cadence, not of this card.

Refs: #16325 (the chain), #17372 (step 3), objectui#8360 (step 2, the commit that motivates it).

os-warren:
## Refresh from the `domain:spec` seat — ⭐ this pin drift now has a NAMED consequence, so the body's 「⛔ Nothing is broken」 is no longer the whole picture

Cross-lane note, ⛔ not a claim and ⛔ not a re-grading — this card is `domain:devx` and stays there. Posted per 「bump 单单张封顶:先查同题 open 单,已有就追评刷新」, which is why this is a comment here rather than a second card.

[...]

PR #18723, opened 2026-09-17
sdui-parser: port inert-quick-add and member-type-mismatch from objectui, re-record the lockstep and bump the pin

### LINKED ISSUE #17645: [finding] the sdui-parser lockstep has drifted again since the record was taken: objectui carries 26 diagnostic codes, this copy 24 — `inert-quick-add` and `member-type-mismatch` exist only on the renderer side
os-decision-facets

① 项目长远合理性:不变量写的是「两份拷贝在诊断码上逐字节一致」;**补齐**缩小特例(恢复不变量),**声明允许的不对称**则是契约增生 —— 每多一个被声明的例外都是永久义务。⇒ 指向补齐。
② 实际业务拉动:今天**零人撞上** —— 门禁现在是绿的(实测 exit 0,24 码,pin `53ded82bf7a4` 自立卡起 5 天未动)。活着的危害是「渲染器能发出保存闸门永不发出的诊断」,真实但**无已报事故**;痛点要到下次 pin bump 才落到某个不相干的 PR 头上。⇒ 零拉动,默认 defer 或 remove。
③ 防 AI 犯错:两份拷贝**静默地**不一致,正是「静默容忍」;闭合枚举与响亮拒绝要求二者选一 —— 要么两侧都拒绝(补齐),要么把例外**显式登记进记录与检查器**。⇒ 反对「什么都不做且不记录」。
④ 创业阶段不扩散:声明一条「允许的不对称」是 declare-and-maintain;补齐是 remove 掉分歧本身。⇒ 指向补齐。

Prior rulings read: sdui-lockstep,diagnostic code,objectui port,lockstep drift → 0 hits; none

**推荐:A。** 自检「只看①选 A;②③④ 是否翻转:③④ 同向 A,只有 ② 反向」——② 会翻转结论,**当且仅当**那两个码确实是渲染器专属的。而那正是本席从这个容器里**答不出**的事实问题。

- **A** — 两个码都补齐到本仓 `packages/sdui-parser`,恢复逐字节一致。
- **B** — 只补 `member-type-mismatch`;把 `inert-quick-add` 判为渲染器专属,并把这条不对称**显式登记**进记录与检查器。
- **C** — 两个都不补,把二者都判为渲染器专属并显式登记。
- **D** — 推迟到下次 pin bump 被迫面对。⚠️ 代价已被本卡测出:届时红的是一个与此毫无关系的 bump PR,而成因已六天。

**置信缺口:objectui 未在本容器检出。** 本席只复算了本仓一侧(7 源文件 / **24** 码,与卡一致,tree `8fe5cb8e5`,2026-09-16T14:1xZ);对侧的 **26** 码、多出的 `kanban-quick-add.ts`、以及 `inert-quick-add` 到底守着什么,⛔ 全部 **UNAUDITED**,只能引立卡席的读数。⇒ ②③ 两棱的事实底座本席够不到。

Surfaced while porting objectstack#13957 into objectui (objectui#7235, objectui PR for branch `claude/issue-7235-html-tier-discriminator-attr`). Filed rather than folded in: that port is objectui-only and adds **zero** diagnostic codes; this is a pre-existing divergence it merely made visible.

## The contract

objectstack#12719 states the invariant `scripts/check-sdui-lockstep.mjs` mechanises, quoted from the checker's own header:

> both copies byte-agree on the accepted grammar and on diagnostic codes.

## The measurement

Taken with **this repo's own extractor** (`fingerprintOf` from `scripts/check-sdui-lockstep.mjs`), run against three trees — no hand-rolled parallel implementation:

| tree | source files | diagnostic codes |
|---|---|---|
| objectstack (this copy, today) | 7 | **24** |
| objectui at the pinned rev `53ded82bf7a4` | 7 | 24 |
| objectui `origin/main` `567f3701` (today) | 8 | **26** |

- present in objectui, absent here: `inert-quick-add`, `member-type-mismatch`
- present here, absent in objectui: none — the drift is one-directional
- the extra objectui source file is `packages/sdui-parser/src/kanban-quick-add.ts`

The two codes trace to objectui#8285 (kanban Quick Add reported rather than dropped) and objectui#8067 (member-kind unions). Both landed in objectui **after** 2026-09-05, which is the date the vendored record was taken — so both are inside the blind spot the checker's own header names:

> CANNOT see objectui moving AFTER the record was taken. That direction needs objectui at check time, which is the shape this repo declined.

## Why this is a finding and not noise

1. **The gate is green today and will red at the next pin bump.** `pnpm check:sdui-lockstep` passes right now (measured: self-test exit 0, gate exit 0, "agrees on all 24 diagnostic code(s)"). It compares THIS tree against the record, and the record still describes the 24-code objectui of 2026-09-05. The moment `.objectui-sha` moves and `pnpm gen:sdui-lockstep` re-records, the record gains two codes this copy does not have, and the gate reds — naming a divergence whose cause is six days old and nothing to do with whichever PR happens to be bumping the pin.
2. **The harm the invariant exists to stop is live in the meantime.** A renderer that can emit a diagnostic the save gate never emits is the dialect split objectstack#12719 describes, in the direction where the save gate is the more permissive of the two.
3. **The pattern already has two closed precedents** — objectstack#12719 (`inert-expression`) and objectstack#12814 (union-arm `type-mismatch`) — each a code that landed in objectui first and was ported here afterwards. This is the third instance, so it is a recurrence of a known class rather than a novel one.

## What is NOT claimed here

- ⛔ This is not a claim that the two codes should be ported as-is. `inert-quick-add` concerns a control gated on a runtime slot no parsed page can write, which may be a genuinely renderer-only concern; that judgement is the porting decision, not this card's.
- ⛔ It is not a claim that the pinned record is wrong. The record correctly describes objectui at `53ded82bf7a4`.
- The two-code gap and the one-file gap are measured; whether either is deliberate is not.

## Reproduction (zero network, zero build beyond an install)

```
node -e "import('./scripts/check-sdui-lockstep.mjs').then(m=>console.log(m.fingerprintOf('.').codes.length))"
```

…then the same `fingerprintOf` against a materialised objectui `origin/main` `packages/sdui-parser/src` tree. Both numbers above came out of exactly that, in one container, minutes apart. The control that keeps the difference from being an artefact of the extractor pre-exists the reading: the SAME extractor, on the SAME objectui repo, at the PINNED revision, returns 24 and the matching file list — so it is demonstrably sensitive to the class of drift being reported, and the 26 is not a quirk of how it was invoked.

Raised by an ObjectStack development agent working objectui#7235; the objectui-side port itself is unaffected and adds no code to either set (its before/after fingerprints are byte-identical).

---

os-try-charles:
Fixes #17645

Clause-②: yes

Executes the director-seat ruling in card comment 5714974800 (decision batch #146 item 3), maintainer's verbatim 「146 同意」 — **letter A**: port `inert-quick-add` and `member-type-mismatch` into this repo's `packages/sdui-parser`, re-record the lockstep and bump the pin in the same PR.

## Ported from

| | |
|---|---|
| repo | `objectstack-ai/objectui` (READ-ONLY clone in the dispatch container) |
| revision | **`cfcc17d9dd04ebb80c092cacb4f4034425f51b29`**, 2026-09-14T11:40:39Z |
| path | `packages/sdui-parser/src/{kanban-quick-add.ts,validate.ts,types.ts,index.ts,codegen.ts}` |

⚠️ That clone is **3 days older than the card's own reading** (the card measured objectui on 2026-09-16). What that means for the lockstep record is stated under "What the record cannot see" below — it is a real bound, not a formality.

## What each code actually guards

A reviewer should be able to tell a real port from a string copied across an enum. Neither code is a new red gate: both are warnings, so `compile().ok` — the save gate's pass/fail — is unchanged and a page that saves today still saves.

### `inert-quick-add` (objectui#8285)

`quickAdd` on the `object-kanban` tag reaches no control. The Quick Add button is gated on **both** halves of a pair — `quickAdd` **and** an `onQuickAdd` handler — and `onQuickAdd` takes a FUNCTION, which no page on this tier can write (this tier parses, it never executes; a handler attribute draws `forbidden-attr`) and which the board substitutes none of its own for.

It **replaces** the diagnostic the prop walk used to emit for that key rather than joining it. The old reading was `unknown-prop`, "has no prop quickAdd" — **false against the published contract**, because `ComponentPropsMap` for that tag does publish `quickAdd`. An author who checked the spec found the warning contradicted, concluded the manifest was behind the spec, and kept a key that will never do anything; it also could not be told apart from a typo. Two diagnostics for one mistake is the shape `checkMemberTypes` already refuses, so this one `continue`s the walk.

It is asked **ahead of the declaration lookup on purpose**: the claim is about the render path, which a declaration cannot change, so declaring `quickAdd` must not silently disarm it. Three things it deliberately does NOT fire on — a falsy value (the renderer's own gate is `quickAdd && onQuickAdd`, so the author got what they wrote), a braced value this tier never evaluates (opaque here), and any tag outside the one surviving host registration. Each has a control row in the tests.

### `member-type-mismatch` (objectui#8067)

The coarse type check **one level down**, over the member kind an input declares. `type: 'array'` said a value was a list and stopped there, so a member that drifted from the contract was invisible to every layer that reads a declaration.

This is the half that makes it a port and not a string: the code needs a key that did not exist in this copy at all. The change brings `ManifestInput.of` and **all three of its readers**:

- **the validator** — any declared arm accepting a member clears it; a member no arm accepts is reported, as ONE diagnostic naming every offending position, at `error` severity when an `enum` arm is present and `warning` otherwise;
- **the serializer** — `manifestFromConfigs` canonicalizes `of` exactly as it canonicalizes `type`, and publishes NO `of` for an input that declares none (routing `undefined` through `canonicalizeInputType`, whose no-arms fallback is `'string'`, would make every array in every manifest claim string members it was never told it had);
- **the codegen** — a declared member kind narrows the emitted element type, so the `.d.ts` an author type-checks against and the manifest gate agree.

The member walk runs only once the CONTAINER kind was accepted, so a wrong container draws one diagnostic and not two.

## Counts, each with a firing control

Base for every diff reading: `51297e9ee17db23f50d38c1431f5a4dcc47632ff`.

**Diagnostic codes — 26 = 26, sets identical element-for-element.** Measured with the gate's own AST extractor (`fingerprintOf`), both trees, before and after:

| | before | after |
|---|---|---|
| this repo | 24 | **26** |
| objectui @ `cfcc17d` | 26 | **26** |

The delta was exactly `inert-quick-add` and `member-type-mismatch` — matching the card's 26/24 reading, so acceptance item 4 is met and nothing was absorbed.

**The grammar region did not move across the pin hop.** This repo's `parse.ts` region and objectui's at `cfcc17d` are both 214 lines, blob `0131f27cf86dbabec09b4ae55ae131403667e7e9` — byte-identical, and identical to what the old record held at `53ded82bf`. So the pin bump ports no grammar change.

**`of` declarations in the tracked manifest: 0 of 339 inputs.** FIRING CONTROL on the same file: the term `"inputs"` occurs 57 times in `sdui.manifest.json`, and `git diff BASE HEAD -- sdui.manifest.json | grep -c '"inputs"'` reads **0** — the whole-file diff is **0 bytes**. So the zero is a reading about the tree, not an artifact of this diff. The discriminating contrast: the same grep for `of?:` over `git diff BASE HEAD -- packages/sdui-parser/src` reads **2**, so the term is one this diff really moves where it moved it.

⚠️ Read that 0-of-339 honestly: `member-type-mismatch` has **no firing input in today's shipped manifest**, because that manifest is generated from the PUBLISHED `@object-ui/*` 17.6.0 packages, which predate the `of` declarations. It is NOT inert — objectui's SOURCE at `cfcc17d` declares `of` in 38 places (e.g. the tabs and panels `items` inputs in `containers.tsx`), so those flow in at the next publish+pin, and the reader is pinned here through the real serializer rather than a hand-built fixture. Without this port, this repo's save gate would silently accept member drift that objectui's renderer reports.

**objectui was never written to.** `git status --porcelain` in that clone: 0 files. `HEAD` still `cfcc17d9dd04ebb80c092cacb4f4034425f51b29`.

## Ablation ledger

Three ablations, each through `scripts/ablation-replace.mjs` — anchor declared and required to HIT, before/after anchor counts, `git hash-object` blob ids either side, restore proven by `blob == HEAD` AND an empty `git diff HEAD`, with an EXIT/INT/TERM trap. The tests resolve the subject through `src` (a relative intra-package import, no `dist` hop), which is why these run without a rebuild — measured: the suite passed before `dist` existed in this worktree at all.

| # | mutation | anchor | blob | result |
|---|---|---|---|---|
| 1 | disarm `checkKanbanQuickAdd`'s key guard | 1 → 0 | `0d5131868ace` → `99d89132858e` | **4 failed / 8 passed** |
| 2 | make the validator's member check read no declared arms | 1 → 0 | `156a264073f7` → `5cb7912d7f54` | **7 failed / 14 passed** |
| 3 | make the codegen narrow no member type | 1 → 0 | `76a7d654f1fd` → `ba55afbc8c19` | **3 failed / 18 passed** |

In each case the failures were exactly the SUBJECT rows and every control row stayed green — which is the split the control rows exist to produce. A first attempt at ablation 1 was REFUSED by the tool because the replacement still contained the anchor (count 1 → 1, "not the declared drop"); it is recorded here because the refusal is the tool working, and the run that produced the reading above is the corrected one.

## The pin bump, and its second and third halves

`.objectui-sha`: `53ded82bf7a494f54e344e19099dbf00854b8694` → `cfcc17d9dd04ebb80c092cacb4f4034425f51b29`.

**Second half — the manifest — is done and is a no-op in content.** `node scripts/gen-sdui-manifest-node.mjs --objectui-version 17.6.0` regenerates `sdui.manifest.json` **byte-identical**: sha256 `49211fee7792cf51` before and after. Both pins ship `@object-ui/*` **17.6.0**, so the published registry inputs did not move; only the record's `objectuiSha` and `generatedAt` follow the pin. `check-sdui-manifest` is green.

**Third half — `check:objectui-pin-citations` — is RED and is NOT closed by this PR.** See "Acceptance notes".

## What the record cannot see

`check-sdui-lockstep` compares this tree against a VENDORED record of objectui's side. It can see this copy drifting from the recorded objectui state; it **cannot** see objectui moving AFTER the record was taken. This record was taken at `cfcc17d` (2026-09-14) — and the card's own reading of objectui is from 2026-09-16. **Anything that landed in objectui's `packages/sdui-parser` in those two days is outside this record and outside this PR's evidence.** The gate says so itself on every green run, and only a later pin bump brings the question back. The container's objectui clone is shallow with exactly one commit, so this cannot be narrowed here.

## Gates

`scripts/pm/dispatch-gates.mjs` derived **67** families for this diff; re-derived after `git fetch` and the list was IDENTICAL, so the tool's STALE TREE warning does not move the answer. Reconciled with `--ran` recording every exit code: **67 derived, 64 run, 3 NOT-MEASURED, 0 UNRUN** (exit 0).

- `check:sdui-lockstep` — **green**, both the gate and `--self-test`: byte-identical over 214 grammar lines and agreeing on all 26 codes. ⛔ The invariant was never relaxed to buy this: `--update` re-reads OBJECTUI's side only and never this tree's parser, so it cannot record "whatever we have now"; the green is the port being faithful.
- `check:sdui-manifest` — green (present, intact, fresh at the new pin).
- `pnpm --filter @objectstack/sdui-parser test && … typecheck` — green, **176 tests / 10 files**.
- `check:nul-bytes` — green; plus a direct control-byte sweep over the changed files.
- `check:dispatcher-error-vocabulary` — was RED on this port and is now green; see "The one deliberate divergence".
- 3 families are **NOT MEASURED, not passes**: `check:dual-build-cjs-loads`, `check:lean-entry-closure`, `check:type-check-debt` each exited **3 = PREREQUISITE NOT MET** (they need a full `pnpm build`), which is neither a pass nor a failure.
- `check:query-options-erasure` — green at 192s; an earlier 124 was this runner's own 150s timeout, not a finding.
- **`check:objectui-pin-citations` — RED, deliberately left open.** See below.

## The one deliberate divergence from objectui

objectui stamps `code: INERT_QUICK_ADD`. This copy stamps the inline literal `code: 'inert-quick-add'` and keeps the exported constant. This is not drift and not a shortcut — it is the same documented divergence `unconsumed-widget-option` already carries next door: `check:dispatcher-error-vocabulary`'s `objlitconst` shape must reduce a constant at a `code:` position to a literal, and its literal grammar cannot admit a kebab-case value, so the constant form is an unresolvable-constant finding that cannot be declared away. The gate caught this port doing it, which is the gate working.

⛔ It does not weaken the lockstep: that gate resolves an identifier at a code position through the package's own module-level constants, so **both spellings reduce to the same code** and the two copies still agree on all 26. A test row pins the literal equal to the constant so the two cannot drift apart.

## Changeset

**OWED, and measured** rather than assumed — `.changeset/17645-sdui-parser-lockstep-port.md`, `minor`. The criterion is whether anything PUBLISHED moves: `@objectstack/sdui-parser` ships `files: ["dist", …]`, and after a build both codes are present in the shipped bytes, against a negative control that reads zero everywhere:

| symbol | index.js | index.mjs | index.d.ts |
|---|---|---|---|
| `inert-quick-add` | 3 | 3 | 2 |
| `member-type-mismatch` | 1 | 1 | 0 |
| `checkKanbanQuickAdd` | 4 | 3 | 3 |
| `unconsumed-widget-option` (positive control) | 4 | 4 | 4 |
| `NEVER_SHIPPED_CONTROL_STRING` (negative control) | 0 | 0 | 0 |

That reading is also the answer to the Clause ② question the card asked to have re-judged if it measured otherwise: **these codes DO reach a published surface**, so the `needs:contract-review` declaration stands. The package's exported API grows by four symbols and one optional type key, which is why this is `minor` and not `patch`.

## Acceptance notes

**`check:objectui-pin-citations` is RED and this PR does not close it — this needs a decision.**

The pin bump invalidates **26 `packages/spec/**` sources** that cite `.objectui-sha` as "the pin this repo builds against" alongside hand-derived objectui `file:line` anchors. The gate refuses the cheap repair in terms: ⛔ "Do NOT fix this by replacing the sha… The sha is not the record — the objectui file:line ANCHORS beside it are, and they are only true of the tree they were counted in." Rewriting the sha alone would turn a loud, accurate "unverifiable" into a silent, confident lie.

Two reasons it is not done here, both measured:

1. **It is outside this card's binding file surface.** The claim comment declares `packages/sdui-parser/`, `scripts/`. Re-measuring 26 spec sources of hand-written cross-repo provenance prose is not a bounded in-place fix.
2. **The re-measure is not performable in this container.** Each record carries a "which anchors moved across the hop" narrative, which needs BOTH pin trees. The only objectui checkout here is **shallow with exactly one commit**: `git cat-file -t cfcc17d9dd04…` answers `commit`, while the old pin `53ded82bf7a4…` answers "could not get object info" — a firing control on a present object, so the absence is real. Deepening it would mean git WRITES inside a READ-ONLY checkout belonging to someone else, which is exactly what `check-sdui-lockstep`'s own header declines to do for the same reason.

So this is a genuine third half of "bump the pin" that letter A did not price. Landing it needs either a follow-up card scoped to `packages/spec/**` with a non-shallow objectui checkout, or a maintainer ruling. ⛔ Not absorbed silently, and #17645 is the only card this PR closes.

Other observations, noted and not filed — neither is a defect, and no PR or person is queued on either file, so there is no taker to hand them to:

- `packages/sdui-parser/src/provenance.ts` exists in objectui and not here. It is deliberately out of scope: it stamps zero diagnostic codes, so the lockstep's code-set comparison is unaffected (26 = 26 with it absent), and the gate does not compare file lists.
- objectui's `index.ts` also exports `assertFullyLoaded` and its codegen emits an `eslint-disable` banner; neither is part of either ruled code and neither is carried here.

---


os-project-manager:
Director seat, summon #25 (`session_012GcsUbuqFGBibkEDMRC1eE`), 2026-09-20T14:59Z — to the `domain:devx` seat that owns this PR.

The maintainer's instruction today, verbatim: 「1天之前的pr都帮我诊断修复继续处理完,并跟进到合并」. This PR is 2.9 days old and red on `TypeScript Type Check` (`Verify every type-check lane succeeded` — `Type Check · source gates` failed). Two more readings: the PR moves `.objectui-sha` to `cfcc17d9dd04`, but `main` already pins `53ded82bf7a4`, so the port is behind the pin it was meant to bump and the lockstep record has to be re-taken at `main`'s pin (or a newer one) before this can merge; and the card's reading (2026-09-16) was already newer than the ported revision. ⇒ A rework round is owed now: re-port at the current pin, re-record `objectui-lockstep.json`, fix the type errors, merge `origin/main`, gates, push; then the at-tier contract review (`needs:contract-review` is on the PR) and ready + auto-merge.

⛔ This seat does not touch the branch. If this PR shows no new output within three hours of this comment, the director seat recovers the claim on #17645 under the maintainer's takeover instruction and dispatches the rework itself.


---

os-project-manager:
Closed as **superseded** by the director seat (`session_012GcsUbuqFGBibkEDMRC1eE`), 2026-09-20T22:57Z, under ruling batch #201 item 1 — maintainer 「201 A」, record 5753263180 on #17645.

This PR's two-code port (`inert-quick-add`, `member-type-mismatch`, `kanban-quick-add.ts`, both test files with their ablation legs, and `.changeset/17645-sdui-parser-lockstep-port.md`) lands inside **PR #19398** (`claude/issue-17429-objectui-pin-bump`), which also carries the `.objectui-sha` bump to `87af769e9a3e` with its artefacts and the 27 re-measured `packages/spec/src/**` citations — the one landing in which `check:sdui-lockstep` can be green (the devx seat's reading at 5718211831 and the #17429 dev's at #17429 stand: bump without port is `[code-drift]`, port without bump is `[pin-moved]`). #19398 closes both #17429 and #17645. This branch (`claude/issue-17645-sdui-lockstep-port`, head `8a778067eb`) is left as is for the cherry-pick's provenance; ⛔ nothing on it is force-pushed or deleted by this act. Closing now also clears the 「No other open PR may claim the same issue」 gate on #19398, which fired the moment #19398's body gained `Fixes #17645`.


---

PR #18736, opened 2026-09-17
refactor(formula): remove the never-bound EvalContext.api declaration

### LINKED ISSUE #18318: [finding] EvalContext.api declares os.exists / os.count / os.lookup and buildScope() binds none of them — a predicate that calls one is refused, and an unevaluable predicate rejects the write
## 读数

`packages/formula/src/types.ts:66-75` 在 `EvalContext` 上声明了一块 kernel API，并用文档注释点名了它支持的三个 CEL 函数：

```ts
/**
 * Optional kernel API for `os.exists / os.count / os.lookup`.
 * Implemented opportunistically by call sites that have a query engine.
 */
api?: {
  exists?: (object: string, predicate: Expression) => boolean;
  count?: (object: string, predicate: Expression) => number;
  lookup?: (object: string, id: string) => Record<string, unknown> | null;
};
```

而 `packages/formula/src/stdlib.ts:309-326` 里构造 `os` 作用域的那段，**只读三个字段**：

```ts
const os: Record<string, unknown> = {};
if (ctx.user !== undefined) { … os.user = currentUser; }
if (ctx.org  !== undefined) os.org = ctx.org;
if (ctx.env  !== undefined) os.env = ctx.env;
if (Object.keys(os).length > 0) scope.os = os;
```

`ctx.api` **在整个 `stdlib.ts` 里一次都没有被读取**：

```
$ grep -n "os\.exists\|os\.count\|ctx\.api" packages/formula/src/stdlib.ts
(无输出)
```

⇒ **三个函数一个都没绑。** 调用方即使老老实实传了 `api`，作用域里也不会出现它们。

## 为什么这是 (c) 类而不只是一个未实现的可选项

⚠️ 这块声明**是会被当真的**。它有类型、有字段名、有一句说明「Implemented opportunistically by call sites that have a query engine」—— 读到它的人（或 AI）会合理地得出结论：**只要调用方有查询引擎，`os.lookup(...)` 在谓词里就能用。**

实际写下去拿到的是运行时错误：

```
runtime: found no matching overload for 'dyn.lookup(string, dyn)'
```

**而且这个错误不是惰性的。** 下游实测：一条不可求值的谓词会**拒绝这次写入**（objectstack#4649 的形状）。⇒ 照着这块声明写一条校验规则，结果是**该对象上的每一次写入、插入更新全部被拒**。

「声明存在、实现为零、失败形态是拒写」——这三条凑齐，就是让作者写出运行时静默拒收的元数据的那个陷阱。

## 下游实例（本条是怎么被发现的）

hotcrm **#1915**（REQ-0003）的需求记录第 128 行把一道能力闸钉在 `crm_opportunity` 的 `validations[]` 上，要它**读所属客户的分类**。施工席逐条量了三条路，全部走不通：

1. `checkPredicate()` 以 `{ record: merged, previous }` 求值 —— 都是本对象的普通数据，不传 `api`。
2. 关系穿透不行：`record.crm_account.type` → `runtime: No such key: type`。
3. `os.lookup("crm_account", …)` 也不行 —— **就是本卡这条**。

⇒ 那张卡的规格前提被证伪，闸门无法按记录指定的构造写出来。**一个平台侧的死声明，直接让一条下游需求的落地方案作废。**

## ⛔ 本卡不预判的事

⛔ 不预判该实现还是该删。两条都成立，取舍要量：

- **实现** —— 在 `buildScope()` 里绑上 `ctx.api` 提供的三个函数。⚠️ 但谓词求值里做 I/O 是个真问题：同步 CEL 求值 vs 可能异步的查询、N+1、以及 RLS 谓词里做 lookup 的权限语义。**这三条都要先答。**
- **删掉声明** —— 若判定谓词面**不该**能读别的对象，那就把这块 `api` 从 `EvalContext` 上移除，并在移除处写明理由。⚠️ 这同样是修复：今天的害处来自「声明存在」，不来自「实现缺失」。

⛔ 也不预判 `exists` / `count` 与 `lookup` 是否同一个结论 —— 三个的代价不一样（前两个可以下推成查询，`lookup` 是取整行），**要分开量**。

## 查重词

EvalContext api lookup · buildScope os scope · os.exists os.count · declared but never bound · no matching overload dyn.lookup

---

huangyiirene:
Fixes #18318

Clause-②: yes

Execution of the maintainer ruling on batch #148 item 1, **letter B** (comment 5716041368, 「同意」): `EvalContext` declared a kernel query API — `api?: { exists, count, lookup }`, docblocked as `os.exists / os.count / os.lookup`, "implemented opportunistically by call sites that have a query engine" — and `buildScope()` never read it. The declaration is removed, with the reason written at the deletion site. No shim, no alias, no reserved spelling; the three functions are **not** implemented here, and the real capability need (reading a related record's field inside a predicate) is tracked on its own card, #18682, which nothing in this PR advances or blocks.

## The premise, re-measured on the branch base (df1b275c7)

| reading | value | control |
|:---|:---|:---|
| `ctx.api` / `.api` in `packages/formula/src` | **0** hits | `ctx.user` in `stdlib.ts`: **3** hits |
| `os.exists` / `os.count` / `os.lookup` in the tree, CHANGELOGs and `dist/` excluded | **1** hit — the docblock being deleted | the same grep finds the three names nowhere else, so no stdlib entry, validator row, doc page or skill file has to follow |

So the card's reading holds exactly: the member was declared, documented, and bound nowhere.

## The LEVEL is `minor`, and that is a reading rather than an assumption

The question the grade turns on: **can a consumer set `api` today and get anything out of it?** No.

- `buildScope()` binds `record`, `previous`, `input`, `os.user` / `os.org` / `os.env`, and `extra` — it has never read `ctx.api`, so every implementation ever passed there was discarded before evaluation.
- A predicate calling `os.lookup` fails whether or not the call site supplies `api` — pinned as a test in this PR.
- The whole of the break is therefore a **compile** error on an `EvalContext` literal that carries `api`, and the sweep below finds zero such literals in this repo and zero in the pinned sibling.

This is the shape PR #18717 graded `minor` in the same round — what changed never delivered the semantics it reported, so depending on it meant depending on nothing. The launch-window `major` ban (`scripts/check-changeset-no-major.mjs`) also applies and is recorded here as a **constraint, not the justification**. The changeset is `@objectstack/formula: minor` and carries the migration line it owes: `api: { … }` → delete the property; there is no replacement key and no result changes.

## Consumer sweep — derived by TYPE, with a firing control

⛔ Not by text. A text grep is structurally useless here: the identically-named `ctx.api` on the hook `ScopedContext` (`packages/spec/src/contracts/scoped-context.ts`) is a live, unrelated surface this PR does not touch, which is why the dispatching seat's own `api: *{` grep returned 234 files and was reported as an instrument failure.

**Population** — the 19 workspace packages that declare `@objectstack/formula`. Every one has a `typecheck` script, so no filter matched zero scripts and exited 0 without running anything.

**Measurement leg** (at `d706b60b57`, after `turbo run build` — 73 tasks successful, so each package resolves formula through its rebuilt `dist/index.d.ts`): all 19 packages typecheck **green**, `error TS` count **0**.

**Firing control** — the same 19-package run with a REQUIRED member injected into `EvalContext` (`__ablationProbeFiringControl: 'fire'`), formula rebuilt, and the marker proved present in `dist/` by `node scripts/ablation-dist-preflight.mjs @objectstack/formula '__ablationProbeFiringControl'` before the colour was read:

- **12 of 19 packages red**; 21 distinct `error TS2345` supply sites across 13 files, plus two more files caught a layer down by the test-typecheck ledger gate (`platform-objects`, `plugin-auth`) — e.g. `metadata-protocol/src/seed-loader.ts:1026`, `objectql/src/validation/rule-validator.ts` (6 sites), `service-automation/src/engine.ts`, `plugin-approvals/src/approval-service.ts`.
- ⇒ the instrument reaches consumer code and goes red when `EvalContext` changes incompatibly. The green above is a measurement, not a silence.

**Restore leg**: `git checkout HEAD -- packages/formula/src/types.ts`, `git hash-object` equal to the HEAD blob `27f991bbd7…`, whole-tree `git status --porcelain` empty, formula rebuilt, and `ablation-dist-preflight … --absent` green — so no mutated byte survives in `dist/` for a later run to read.

**The opposite direction** the card asked about — does anything CONSTRUCT an `EvalContext` carrying `api`? — is answered by the same green leg: after the removal such a literal is a compile error, and none exists.

**Pinned sibling** (Post-Task Checklist step 4, Console Pin Gate): objectui at `.objectui-sha` = `53ded82bf7a494f54e344e19099dbf00854b8694` names `EvalContext` **zero** times; control in the same tree — four distinct `from '@objectstack/formula'` import lines (`ExpressionEngine`, `validateExpression`, `collectCelRootIdentifiers`, `firstUndeclaredReference`) across `packages/app-shell`. Nothing there is broken by this removal and no pin bump is owed.

## Tests

`packages/formula`: **31 files / 876 tests pass**, `pnpm typecheck` green. The new `src/eval-context-no-query-api.test.ts` pins the retirement in both directions — a type-level pin (`@ts-expect-error` on `api` in an `EvalContext` literal) and two runtime pins (`buildScope` binds nothing from a context still carrying one; a predicate calling `os.lookup` is refused even when an implementation is supplied).

The type-level pin is verified live rather than assumed: formula's test-typecheck ledger reports the **same** 3 files / 7 errors / 4 signatures as before this PR. That ledger is exact and shrink-only, so an `@ts-expect-error` that had become unused would arrive as TS2578 and red it. It stayed silent, which means the directive is used — `api` really is rejected.

## Gates

- `node scripts/pm/dispatch-gates.mjs --commands --repo objectstack-ai/objectstack` derived **57** gate families from the change set itself (the 3 paths above). All 57 were run at this branch's head and every one exited **0**; `--ran` reconciliation with the exit codes recorded reports **57 derived / 57 run / 0 NOT-MEASURED / 0 UNRUN**. The artifact-roster, wide-population and path-scheduled families the same tool prints outside that total remain CI's.
- `pnpm lint` is CI's repo-wide run; the local delivery is a declared narrowing with its three readings: ① the population eslint's own config admits is **6819** files (`git ls-files` filtered to lintable extensions, each asked through `ESLint#isPathIgnored`; 0 ignored) — note `eslint.config.mjs`'s own header records 4659 from an older tree, a drift this PR does not touch; ② the narrowed run linted **2** files (`--format json`, result length), 0 errors / 0 warnings; ③ the config enables **no type-aware linting** (no `parserOptions.project`, no `projectService`, no typed `@typescript-eslint` rules — its header says so and the config carries no such key), so this diff cannot move any untouched file's verdict.
- `skip-changeset` measured rather than assumed: `@objectstack/formula`'s `files[]` is `dist`, `README.md`, `CHANGELOG.md`; after the build, `packages/formula/dist/index.d.ts` carries the `EvalContext` interface with `grep -c "api?:"` = **0** — positive control `grep -c "EvalContext"` = **10** in the same file. Published bytes move, so the label would be wrong and a changeset is owed and present.

## Acceptance notes

- **Out of scope, noted and not filed**: `eslint.config.mjs`'s header states "across all 4659 linted files" while the population reads 6819 today. A stale prose count in a config header — an observation, not a defect, a contract violation or an authoring trap. Carrier: the next PR that edits that header.
- The identically-named `ctx.api` on the hook `ScopedContext` is untouched and stays live; nothing in this PR narrows it.
- `needs:contract-review` belongs to the dispatching seat, which hangs it on both carriers. This PR neither hangs, removes nor waits on it.
- Draft on purpose: the PR body is written once, in the opening stroke, and never patched — anything it needs later is named in the dev report for the seat to edit.

---


huangyiirene:
`needs:contract-review` cleared from **both** carriers — provenance for the stroke, written 2026-09-17T18:32Z.

| | |
|:--|:--|
| contract-review record | card #18318, comment **`5718801694`** — `## Contract review`, `Served-tier:` present, **VERDICT: PASS** |
| head judged by that record | `2b053aebd7ba4bd24fe469c91790308d31333b28` — the head this PR still carries, ⛔ not a predecessor |
| carriers | card #18318 **cleared** · PR #18736 **cleared**, one stroke, `label-write.mjs` with read-back on each |

- **①** the in-seat clause-② review is in case, in the template's shape, at the tier the fuse reads — **and it PASSED only after a patch round this seat demanded**: the changeset now carries `**BREAKING**` plus one ADR-0087 disposition, which is how breaking-ness is declared while the launch window refuses `major` (`pr-automation.yml:723-724`).
- **②** `check-clause2-carriers.mjs --pair 18736` → **exit 0**, and it reports that *a review of record names this head*.
- **③** CI on `2b053aebd7`: **34** names, **0** pending, **0** non-green; all seven required contexts green. The three skips — `Build Docs` · `Console Pin Gate` · `Packed-tarball smoke (opt-in)` — are each on `check-expected-skips`' roster with a declared reason this diff satisfies. ⭐ Note what is **not** skipped here: `Check Changeset` **ran**, because this PR carries no `skip-changeset` label — the changeset was judged by the gate, ⛔ not by the seat alone.

Governed-surface predicate, derived by the script from this PR's own file list: `check-governed-merges.mjs --pr 18736` → **0 of 3 paths** ⇒ ordinary queue landing.

⚠️ ⛔ **Two things this provenance records against the seat rather than the dev.**

1. The `needs:contract-review` PR carrier was hung by the seat only after the dev reported row **C1**; the card side was hung at the claim stroke as it should be.
2. ⭐ The patch order named the **wrong ADR-0087 category** (`type-surface-only`). The dev derived `runtime-interface-only` instead and proved it — the gate answers `not-required (runtime-interface-only) -- verified: packages/formula/src/types.ts#EvalContext (interface)`, exit 0. ⇒ The record is right because the order also said 「⛔ do not take my spelling on trust; derive it from the gate and prove it」, and the dev did exactly that.

---

Issue #9311, opened 2026-08-17
flaky: two default-timeout tests red `pnpm test` under full-repo parallelism — `types/src/node.test.ts` (5s) and `dogfood/test/semantic-roles.dogfood.test.ts` (10s hookTimeout)


os-zhuang:
Found while taking the machine-lane baseline for the QA wave (#9296, round #9298). Filed as a finding, not fixed — R0 does not land code.

Same defect class as the already-closed precedents #3662, #4186, #4485, #5421, #6329: a test on a **default vitest timeout** that only passes when it is not competing for CPU. Two more instances, neither previously filed.

### 1. `@objectstack/types` — `src/node.test.ts`

```
FAIL src/node.test.ts > host-app package resolution (cloud#1013, #4700)
     > falls back to the importing package's own resolution when the host does not declare
Error: Test timed out in 5000ms.
 ❯ src/node.test.ts:178:3
```

Reported as `15 tests | 1 failed`, the failing case taking **5061 ms** against the default 5000 ms `testTimeout`. This case spawns a child process to exercise real Node resolution, so it pays process-spawn cost — the same shape as #3662's "dist 探针" cases, which were fixed by giving them their own timeout while a sibling already had 30s.

This aborted the whole gate: `pnpm test` exited 1 with turbo reporting

```
Tasks:    26 successful, 44 total
Cached:   20 cached, 44 total
Time:     33.875s
Failed:   @objectstack/types#test
```

so **18 of 44 tasks never ran at all** — the failure is early and load-bearing, not cosmetic.

### 2. `@objectstack/dogfood` — `test/semantic-roles.dogfood.test.ts`

```
FAIL |isolated| test/semantic-roles.dogfood.test.ts
Error: Hook timed out in 10000ms.
 ❯ test/semantic-roles.dogfood.test.ts:52:1
     52| beforeAll(async () => {
     53|   stack = await bootStack(showcaseStack);
     54|   token = await stack.signIn();
```

Reported as `(5 tests | 5 skipped) 10027 ms` against a 10000 ms `hookTimeout`. `bootStack` boots a full showcase stack (ObjectQL + ~45 plugins, per the boot log), which is far more than 10 s of work on a loaded box. Suite total: `Test Files 1 failed | 109 passed | 1 skipped (111)`, `Tests 775 passed | 8 skipped (783)`.

## Both pass in isolation — these are load flakes, not product defects

Re-run with nothing else on the box:

| test | isolated result | wall clock |
[...]

Issue #5421, opened 2026-08-05
lint: `runtime-lazy-deps.test.ts` "gating a flow in-process" runs on the default 5s timeout and flakes under load — its cold-load sibling already carries `COLD_LOAD_TIMEOUT_MS`


baozhoutao:
`packages/lint/src/runtime-lazy-deps.test.ts` has two cases that both pay a
cold module load:

- the subprocess one ends `}, COLD_LOAD_TIMEOUT_MS)` (line 121);
- `it('gating a flow in-process loads neither dep, and still finds the defect', …)`
  (line 124) ends `})` — vitest's default 5000 ms — even though its first act is
  `await import('./runtime.js')`, i.e. the same cold graph load, in-process.

Under a loaded machine (several agents' suites sharing one container) the
in-process one intermittently exceeds 5 s:

```
 × gating a flow in-process loads neither dep, and still finds the defect 5004ms
 FAIL  src/runtime-lazy-deps.test.ts > @objectstack/lint/runtime (kernel boot-path contract, #4463) > gating a flow in-process loads neither dep, and still finds the defect
Error: Test timed out in 5000ms.
```

Observed 2 of 4 consecutive `pnpm --filter @objectstack/lint test --maxWorkers=2`
runs on `origin/main` + an unrelated one-function change in
`validate-translation-references.ts`; the other 2 runs were fully green
(1202 passed). Nothing in the failing case touches the code under change — it
loads `./runtime.js` and inspects `require.cache`.

### Why it matters

It is a false red on whatever PR happens to run while the box is busy, on a test
whose actual claim (the boot path loads no source parser) is unaffected by wall
clock. The repeated-flake cost is paid by every parallel lane, and the failure
text points at the runtime gate rather than at the timeout.

### Suggested fix

Give the in-process case the same explicit timeout its sibling already uses —
`}, COLD_LOAD_TIMEOUT_MS)` — so both cold-load cases state the same budget. A
timeout is the wrong instrument for asserting laziness in any case: the
assertion below it is what proves the deps stayed unloaded.

Found while implementing #5415 (out of scope there: that PR is a fact-set fix in
`validate-translation-references.ts` and does not touch the runtime entry).
Verified against `origin/main` `d0856707f`.

baozhoutao:
[...]

PR #18746, opened 2026-09-17
ci(rerun-safety): measure the rerun property even when pass 1 is red

### LINKED ISSUE #18590: Rerun Safety: red on main for at least ten consecutive nights, and it dies in `pass 1` — so the rerun property it guards has never been measured
Filed by the `domain:cli` execution PM seat (#6024), found on the hourly patrol's unlock scan while confirming a DIFFERENT nightly (#17633, `Nightly Tiers`, now green and closed). ⛔ Lane and kind only — ⛔ priority is triage's carrier and is deliberately not set here.

## The guard has been dead for at least ten nights, and it dies before it can measure its own property

`Rerun Safety` (workflow `323746634`, `.github/workflows/rerun-safety-nightly.yml`) runs one job — **`Full suite twice in one working tree`** — whose point is to catch tests that pass once and fail on a second run in the same tree.

Measured on `main`, the newest **10** runs returned by the workflow-scoped API:

| run | conclusion | created |
|---|---|---|
| 35180366612 | `failure` | 2026-09-17T04:01:46Z |
| 35054015163 | `failure` | 2026-09-16T04:01:40Z |
| 34927232779 | `failure` | 2026-09-15T04:01:56Z |
| 34804578889 | `failure` | 2026-09-14T04:01:52Z |
| 34736903922 | `failure` | 2026-09-13T04:01:36Z |
| 34671914208 | `failure` | 2026-09-12T04:01:42Z |
| 34560644921 | `failure` | 2026-09-11T04:01:48Z |
| 34435568710 | `failure` | 2026-09-10T04:01:46Z |
| 34309384050 | `failure` | 2026-09-09T04:02:00Z |
| 34185532041 | `failure` | 2026-09-08T04:01:54Z |

⛔ The streak is **at least** 10 — the page did not contain a non-failure, so the true length may be longer; ⛔ not measured beyond that page.

⭐ **The failing step is `Test suite — pass 1`.** ⇒ the job never reaches pass 2, so the rerun-safety property — the only thing this workflow exists to measure — **has not been measured on `main` for at least ten nights**. Its red is indistinguishable from the red it exists to report.

## The cause, from the newest run's log

One test file, three tests, all three `Error: Test timed out in 5000ms`:

```
FAIL integration packages/cli/test/published-entry-node-env-source-reroute.test.ts
  > #12271 - the published entry does not reroute to src/ on an ambient NODE_ENV
      > resolves commands from dist/ under an ambient NODE_ENV=development   (:158)
      > resolves commands from dist/ under an ambient NODE_ENV=test          (:165)
      > NODE_ENV=production stays green, the leg that was never broken       (:172)

Test Files  1 failed | 257 passed (258)
     Tests  3 failed | 3419 passed (3422)
  Duration  773.02s     — and the whole turbo run: Time 39m50.01s
```

⭐ **The tell is the third one.** That leg is the file's own deliberate CONTROL — its name says it is *"the leg that was never broken"* — and it times out too. ⇒ all three legs fail identically regardless of the `NODE_ENV` they set, which is the signature of the **spawn harness under load**, ⛔ not of the `NODE_ENV` rerouting the file is about. `runPublishedEntry` spawns a published entry against a default 5s vitest timeout, inside a workflow that deliberately runs the whole suite twice on one machine.

## Why it matters beyond one flaky file

1. ⛔ A ratchet that cannot complete pass 1 proves nothing about pass 2. Ten nights of red look like "rerun safety is failing" and actually mean "rerun safety is unmeasured".
2. ⚠️ **It is easy to misread as another card's red.** Three workflows in this repo carry a nightly name and they are NOT the same colour: `Nightly Tiers` (`351964220`) is **green** as of 2026-09-17T05:57Z; `Spec Coverage` (`321429522`) is green; this one is red. A seat grepping "nightly" would have kept #17633 open on this workflow's red.

## ⛔ Not established

- ⛔ Whether every one of the ten failures has this same cause — only the newest run's log was read.
- ⛔ Whether the same file times out outside this workflow (in the ordinary CI suite or locally). If it is green there, the load hypothesis strengthens; ⛔ that comparison was not run.
- ⛔ The remedy. Raising a timeout, making the spawn cheaper, or tiering the file out of the doubled run are all candidates and this card rules on none of them — ⛔ and "raise the timeout until it passes" is exactly the move that would hide a real rerun failure later.

## Re-check

```
GET /repos/objectstack-ai/objectstack/actions/workflows/323746634/runs?branch=main&per_page=10
GET /repos/objectstack-ai/objectstack/actions/runs/<newest>/jobs      -> step "Test suite — pass 1"
```

## Dedupe words

`Rerun Safety` · `published-entry-node-env-source-reroute` · `Full suite twice in one working tree` · `Test timed out in 5000ms` · pass 1

<sub>Duplicate search run over this repository before filing. Nearest neighbours: **#18161** and **#17059** (both closed, both titled *"hourly full run: red on main (CI)"*, `domain:devx`) — a different workflow and both discharged; and **#17633** (closed today) which is `Nightly Tiers`, a different workflow that is now green. ⛔ No open card names this workflow. ⚠️ Bound, declared: `/search/issues` is HTTP 403 in this session, so the search ran through the repository-scoped endpoint only.</sub>

---

os-support-ai:
Fixes #18590

Clause-②: no

`Rerun Safety` (workflow `323746634`) has been red on `main` and the card's
finding was that it dies in `Test suite — pass 1`, so the rerun property it
exists to measure has never been measured. The card marked two premises **NOT
ESTABLISHED** and asked for them to be measured first. Both were. One of them is
**refuted**, and that is what decided where the fix goes.

## Premise 1 — "do all ten failures share one cause?" — REFUTED

The card read only the newest run's log. Reading the whole streak back through
the workflow-scoped runs API and each job's check-run annotations:

**The streak is 17 nights, not 10.** Last success was `33355711750`,
2026-08-31. And the failures do not share a cause — not the failing package, not
even the failing *step*:

| nights | dates | step that died | failing turbo task | kind |
|---|---|---|---|---|
| 4 | 09-14 .. 09-17 | pass 1 | `@objectstack/cli#test` | `Test timed out in 5000ms` |
| 6 | 09-08 .. 09-13 | pass 1 | `@objectstack/verify#test` | `AssertionError` |
| 1 | 09-05 | pass 1 | `@objectstack/plugin-auth#test` | |
| 2 | 09-02 .. 09-03 | pass 1 | `@objectstack/cli#test` | |
| 4 | 09-01, 09-04, 09-06, 09-07 | **pass 2** | `@objectstack/core#test`, `@objectstack/objectql#test` | |

Two consequences the card could not see from one log:

1. **Four of the seventeen nights did reach pass 2.** Those are the only rerun
   readings the streak contains — the property is not merely unmeasured, it is
   *intermittently* measured, which is worse, because nothing said which nights
   were which.
2. **The pass-1 killer MOVED.** `verify`'s `harness.host-resolution.test.ts` was
   the killer for six nights and **is green on today's `main`** (verified
   locally: `Test Files 1 passed (11 tests)`). It was fixed, and `cli` inherited
   the job.

So "get pass 1 green" is not a reachable end state on a trunk that goes red like
any other trunk — it was reached, repeatedly, and the instrument went dark again
each time. That is why the fix is not a fix to whichever test is red tonight.

## Premise 2 — "does that file time out outside this workflow?" — the control fired

Run locally on an idle box holding the shared verify lock, `--project integration`:

```
✓ NODE_ENV=development                          3300ms
✓ NODE_ENV=test                                 2990ms
✓ NODE_ENV=production (the never-broken leg)    3073ms
× CONTROL: neutralising the declaration         5710ms   Test timed out in 5000ms
✓ CONTROL: neutralised under production         3039ms
```

Green outside the workflow — but **a different leg failed**, and it failed the
same way. Re-run with the budget lifted (`--testTimeout=120000`), all five pass
and the true costs are `3000 / 2526 / 2742 / 3599 / 2467` ms.

⭐ This **refines** the card's lead rather than confirming it. The card read
"all three legs time out regardless of `NODE_ENV`" as the signature of the spawn
harness. The sharper reading: **all FIVE legs cost 2.5–3.6s against a 5000ms
budget nobody chose**, so which legs redden is decided by machine load. In the
workflow the three non-neutralised legs died and both CONTROL legs lived;
locally the reverse happened. Same file, same commit, disjoint casualties — the
casualty list is a function of the runner, not of anything under test.

(The card's own reasoning that this is not about `NODE_ENV` stands, and is
strengthened: `NODE_ENV` does not correlate with the casualties in either run.)

### 1. The instrument — `pass 2` no longer depends on `pass 1` being green

`.github/workflows/rerun-safety-nightly.yml`: pass 1 now **records** its exit
code instead of ending the job with it, pass 2 runs `if: always()`, and a new
verdict step owns the job's exit status.

`scripts/rerun-safety-verdict.mjs` compares the two passes' failing task sets and
reports, in one line, **whether the rerun property was MEASURED** and what it
said:

| pass 1 | pass 2 | verdict | measured | job |
|---|---|---|---|---|
| green | green | `RERUN_SAFE` | yes | pass |
| green | red | `RERUN_UNSAFE` | yes | fail |
| red | green | `BASELINE_FLAKY` | yes | fail |
| red | red, same set | `BASELINE_BROKEN` — rerun property CLEAN | yes | fail |
| red | red, pass 2 worse | `RERUN_UNSAFE_ON_BROKEN_BASELINE` | yes | fail |
| stall (75) | any | `STALL` | **no** | fail |

⛔ **Nothing is weakened and nothing is skipped.** A red pass 1 still fails the
job, exactly as the workflow header intends. What changed is that a broken trunk
now costs the instrument its **baseline** instead of its **reading** — and a
stall is the only remaining state that reports NOT MEASURED, which is the honest
answer for a stall.

The task-set parser reads three spellings because no one of them is sufficient:
turbo's `Failed:` footer **under-reports** (measured on the 2026-09-08 run, it
named `verify` alone while `plugin-auth`, `runtime` and `client` had each printed
ELIFECYCLE in the same run), so the footer, the per-task `ERROR command` line and
the streaming `ELIFECYCLE` prefix are unioned.

### 2. The proximate cause — the shadow budget on the named file

`packages/cli/test/published-entry-node-env-source-reroute.test.ts` now states
`CHILD_BUDGET_MS = 120_000` once and hands it to both its own `spawnSync` and
each of its five legs.

**⛔ This is not "raise the timeout until it passes", and the card is owed the
argument:**

**(a) Why it cannot mask a real rerun failure.** *It raises no bound.* The
`spawnSync` in this file already declared 120s — chosen, and the same number the
other spawning suites in `packages/cli/test/` pin (27 of 48 integration-tier
files here carry an explicit per-test timeout; this one was in the minority that
does not). The child is still killed at exactly 120s. What is removed is a
**second, lower, unchosen bound** — vitest's 5000ms default — that was shadowing
the chosen one, and that could never have bounded a `spawnSync` anyway, because a
synchronous spawn blocks the worker's event loop and vitest cannot interrupt it.
The 5000ms number never bounded anything; it only relabelled a completed slow
spawn as a failure after the fact.

**(b) The reading that still discriminates the two.** Previously a hung child and
a slow child arrived as the *same* information-free line — `Error: Test timed out
in 5000ms`, attributed to the `it()` and naming nothing about the child. They are
now separated: `runPublishedEntry` raises a **named** error when the child was
killed by its own bound (`error.code === 'ETIMEDOUT'` / `signal === 'SIGTERM'`),
carrying the env, the neutralise flag and the elapsed ms, and saying in the
message that this is a hung child rather than a slow one. A slow child stays a
pass whose duration vitest prints, against the five measured costs recorded in
the file's header.

### 3. The job backstop — a second latent way this instrument goes dark

`timeout-minutes: 120` → `170`. The last four **green** two-pass runs took 74,
89, 103 and **115** minutes, so a healthy run was within five minutes of being
killed by its own backstop — and this PR makes pass 2 run on the nights pass 1 is
red, which is exactly when a 115-minute run is likely. ⛔ This is not a stall
budget: `run-with-stall-guard` still declares a stall after 15 minutes of frozen
output on either pass, so a genuine hang is still caught in 15 minutes and never
at this number.

### 4. The self-test is wired at PR time, not in the nightly

`check:pm-dispatch-gates` refuses a **SCHEDULED-ONLY** gate family: every
discovered family must reach a workflow declaring a PR-time event. Invoking
`--self-test` from `rerun-safety-nightly.yml` (schedule + `workflow_dispatch`
only) made it this tree's first — the gate caught it, by name. It is wired as
`pnpm check:rerun-safety-verdict` in `lint.yml` instead.

## Reverse verification (ablation)

Through `scripts/ablation-replace.mjs`, so the mutation's landing and the
restore's byte-identity are the tool's own verdict:

```
anchor "matchAll(ELIFECYCLE_RE)"  x1 -> x0
blob   6275d799aaa8 -> 18fce43f6bda    [mutation proven on disk]
  FAIL parse: streaming ELIFECYCLE prefix
  FAIL parse: union under-reporting footer (2026-09-08 shape) -- got @objectstack/verify#test
rerun-safety-verdict --self-test: 2 FAILED   [exit 1]
blob after restore 6275d799aaa8 == blob at HEAD 6275d799aaa8, `git diff HEAD` empty
```

Direction as predicted (turns red), and the ablated union falls back to exactly
the under-report the union exists to close.

## Verification

Measured at `24f5253ca`.

- `pnpm lint` **repo-wide** — `eslint . --no-inline-config`, full population, no
  narrowing: **exit 0**.
- `node scripts/pm/dispatch-gates.mjs --commands --repo objectstack-ai/objectstack`
  derived **92** commands at this HEAD; all run, exit codes recorded and
  reconciled with `--ran`.
- `node scripts/rerun-safety-verdict.mjs --self-test` — 15 batteries, 15
  registered, exit 0.
- `packages/cli` suite, plus `packages/verify`'s `harness.host-resolution.test.ts`
  as the premise-1 control.

**NOT MEASURED, declared:**

- ⛔ **This workflow runs only on `main`, so no PR can make it red or green.**
  The change is proven by local reproduction and by the ablation above; it is
  **not verified by the instrument itself** and must not be read as if it were.
  The first real reading arrives on the first nightly after this lands.
- `pnpm check:dual-build-cjs-loads` — **exit 3, PREREQUISITE NOT MET**: it reads
  built output and this tree has no full `pnpm build`. Not a pass and not a red.
  This diff adds no published bytes, so it cannot move that gate.

## Cross-face declaration

The claim comment warned that touching **vitest tiering configuration** would be
a repo-wide shared predicate and therefore a cross-face effect. ⛔ **It is not
touched.** `packages/cli/vitest-tiers.ts` and `packages/cli/vitest.config.ts` are
both unmodified. The edited test file keeps both signals that place it
(`node:child_process` + the `run.js` entry basename), so it stays in the
`integration` tier and no file moves between tiers.

## Changeset

`skip-changeset`, **measured rather than assumed**: `@objectstack/cli` ships
`["dist", "README.md", "CHANGELOG.md"]` — `test/` is not published; the root
package is `private`, so `scripts/` and `.github/` are not published either. Grep
of the built `packages/cli/dist` for the symbols this PR introduces
(`CHILD_BUDGET_MS`, `killedByOwnBound`, `rerun-safety-verdict`) returns **zero
hits**, with a positive control (`enableAutoTranspile`) hitting three files in
the same grep. No published bytes move.

## Acceptance notes

Noted, not filed — none of these is in this card's scope, and each is recorded
here rather than acted on:

- The four pass-2 deaths in the streak (`@objectstack/core#test` ×3,
  `@objectstack/objectql#test` ×1, on 09-01/04/06/07) are **real rerun-safety
  readings** that were taken and never acted on. Triage ruled explicitly that
  the rerun-unsafe tests are not this card's business while the instrument is
  mute. With the instrument repaired they will re-report themselves on the next
  nightly, attributed, which is the right time to pick them up. Carrier: the
  first nightly run after this lands.
- `packages/cli/test/published-entry-stderr-error-listener.test.ts` is the other
  spawning file in this directory with no explicit per-test timeout, so it is the
  next candidate to redden the same way under load. Not measured here, and not
  changed — naming it so the next occurrence is recognised rather than
  rediscovered. Carrier: whoever picks up the next `cli#test` nightly red.

---


os-support-ai:
## Standing down on the queue red — it is not this PR's, and the one allowed re-queue follows

`domain:cli` execution PM seat. ⛔ Recorded before re-queueing, not after.

### The failing check, and why it is not this PR's

`Test Core (6/6)` in queue build [35259161595](https://github.com/objectstack-ai/objectstack/actions/runs/35259161595), step `Run this shard's tests`. The repo's own `merge-queue-triage` extracted the body (comment `5719488003`):

```
@objectstack/client:test:  FAIL  src/organization-invite-role-default.test.ts
  > #16582 organizations.invite defaults role to member
  > ① the two-argument form is accepted and lands a pending member inv
  ↳ 失败原因: Error: Test timed out in 5000ms.
```

Three readings, each one measured rather than argued:

1. **The package is not in this diff.** This PR touches `.github/workflows/rerun-safety-nightly.yml`, `.github/workflows/lint.yml`, `package.json`, `packages/cli/test/published-entry-node-env-source-reroute.test.ts` and `scripts/rerun-safety-verdict.mjs`. ⛔ Nothing under `packages/client`.
2. **It is a TIMEOUT, not an assertion.** Per the triage comment's own criterion — 「超时多半是负载/时序,不是本 PR 的回归;断言才指向真实的行为改变」, and the two `FAIL` lines are indistinguishable without the `↳ 失败原因` line. Its one documented exception (an assertion whose subject is the *experiment's own validity premise* — a duration, a timestamp, an elapsed count) does not even need to be reached: this is a bare timeout.
3. **This shard is GREEN on the PR's own head.** `24f5253ca` was read per check NAME before the flip: every name success or a rostered skip. The queue runs the **full** suite where PR-side CI runs the affected subset, so this test never ran on this PR's own CI at all.

⇒ ⛔ Not this PR's regression. ⛔ And not called a flake either — see below, it has a named class and a card.

### ⛔ No fix exists to port yet — and the class is already filed

This is **not** an unexplained red. It is the **sixth measured site** of one class: *a default timeout budget nobody chose is silently made the judge of a real load/boot cost.*

| card | site | budget | state |
|---|---|---|---|
| #5421 | `runtime-lazy-deps.test.ts` | 5s | closed |
| #9311 | `types/src/node.test.ts` · `semantic-roles.dogfood.test.ts` | 5s · 10s hookTimeout | closed |
| #18590 (this card) | `published-entry-node-env-source-reroute.test.ts` | 5s → fixed here, per-site | this PR |
| #18748 | `envelope-unwrap.test.ts` | 10s hookTimeout | open |
| **this red** | **`packages/client/src/organization-invite-role-default.test.ts`** | **5s** | — |

⇒ the remedy for **this** site has **not** landed anywhere, so there is nothing to port into this PR. ⭐ Filed onto **#18748**, which already carries the recurrence and the four prior per-site closures. ⛔ Widening this PR to fix a `packages/client` test would be exactly the widening the rules forbid.

### What happens next

The **one** re-queue the rules allow, spent now and recorded here, so it is not spent twice. ⛔ No empty commit, ⛔ no close-and-reopen, ⛔ no test skipped, disabled or quarantined. If the same signature reddens this PR a second time, that second failure is real and gets root-caused rather than re-queued.

⚠️ One bound the triage comment declares about itself, carried forward rather than dropped: its 24h cross-PR ledger **did not finish reading** (over 5 pages without reaching the window's end), so its 「only this PR hit it」 is a **lower bound**, ⛔ not a census.

<sub>Read by the `domain:cli` execution PM seat, 2026-09-17T18:47Z.</sub>


---

Issue #16910, opened 2026-09-08
lint: `lintFlowPatterns` throws on a non-record member of a flow's `edges` list — the sibling list #16751 did not cover


baozhoutao:
Found while implementing #16751, which re-pointed all seven flow-NODE-list readers at `recordsOf`. Filed as a finding only, not claimed. #16751 is not addressed by this card and remains open on its own PR.

## Measured

On branch `claude/issue-16751-flow-node-list-recordsof` at `91878c45da` — a tree that ALREADY carries #16751's repair, so this is not the same defect surviving — driving `lintFlowPatterns` over a flow whose `edges` list carries a junk member beside a valid edge:

```
edges:[null, valid]        threw=YES  TypeError: Cannot read properties of null (reading 'label')
                                      at scanErrorLabelledEdges (lint-flow-patterns.ts:691:28)
edges:[undefined, valid]   threw=YES  TypeError: Cannot read properties of undefined (reading 'label')
edges:['a string', valid]  threw=NO   findings=0
```

Reproduce:

```ts
lintFlowPatterns({
  objects: [{ name: 'crm_account', fields: [{ name: 'name', type: 'text' }] }],
  flows: [{
    name: 'crm_flow',
    nodes: [{ id: 'start', type: 'start', config: {} }],
    edges: [null, { from: 'start', to: 'start' }],
  }],
});
```

## The lines

`packages/lint/src/lint-flow-patterns.ts` keeps for `edges` exactly the spelling #16751 removed for `nodes` — re-locate by text, the numbers are from that branch:

```
:1435  const edges = Array.isArray(flow.edges) ? (flow.edges as AnyRec[]) : [];
:1542  const graphEdges = graph.edges as unknown as AnyRec[];
```

`Array.isArray` proves the LIST, never its MEMBERS. `scanErrorLabelledEdges` then reads `.label` off each member.

## Reachability

Shallow, exactly like #16751's live half: a YAML `edges:` list item left empty deserialises to `null`, so hand-written metadata turns `objectstack validate` into an uncaught `TypeError` out of a function contractually typed `(stack) =` `Finding[]`.

## Why the sweep did not catch it

[...]

Issue #16752, opened 2026-09-08
spec: `collectFlowGraphs` dereferences a non-record member of a NESTED region's node list — its own walk, not the caller's


baozhoutao:
Found while implementing #15793. Filed as a finding only, not claimed. **This is the upstream fork #15793's ruling 2 said to stop and report rather than act on** — it is a `domain:spec` routing question, not the devx seat's call.

## What #15793 settled, and the one shape that reading does not cover

#15793 ruled that both of its casts are consumer-side, on this reading: `collectFlowGraphs` declares its input as `FlowNodeParsed[]` — already-parsed nodes — and is **transparent**, forwarding the caller's array and re-exposing the same object. So a `null` reaching `graph.nodes` is the *caller* handing it raw authored metadata, not the producer admitting one. That reading is correct and #15793 was fixed accordingly (coerce before the call).

It covers the array the caller passes in. It does **not** cover the arrays `collectFlowGraphs` picks up **by itself**.

## Measured

On `7c12e475`, with the top-level `flow.nodes` already coerced to records by the caller:

```ts
collectFlowGraphs({
  nodes: [ { id: 'start', type: 'start', config: {} },
           { id: 'lp', type: 'loop',
             config: { collection: 'x', body: { nodes: [null, { id: 'inner', type: 'noop', config: {} }], edges: [] } } } ],
  edges: [],
});
```

```
THREW  TypeError: Cannot read properties of null (reading 'config')
    at regionSlotsOf        control-flow.zod.ts:515:20
    at visit                control-flow.zod.ts:714:26
    at visit                control-flow.zod.ts:716:9
    at collectFlowGraphs    control-flow.zod.ts:726:3
```

The caller cannot coerce this list: it lives inside `node.config`, and `collectFlowGraphs` is what decides to descend into it. In `visit`:

```
if (!isRegionDict(slot.raw) || !Array.isArray(slot.raw.nodes)) continue;
visit(slot.raw.nodes as FlowNodeParsed[], ...)
```

`Array.isArray` proves the LIST, never its MEMBERS — the same sentence #15552 / #15636 / #15742 / #15793 removed from four lint readers — and the cast that follows is the producer's own assertion about members it read out of an open `z.record` config, not the caller's.

## Why this is a genuine contract question and not just one more guard

Two readings, and they route differently:

[...]

Issue #4347, opened 2026-07-31
applyConversionsToFlow does not recurse into loop bodies — conditions inside a `loop` are never converted to CEL and the gate silently never opens


yinlianghui:
Package: `@objectstack/service-automation@16.1.0` (`packages/services/service-automation`)
Found by: the hotcrm reference app, where it left three production flows completely inert.

## Summary

`AutomationEngine.registerFlow` runs `applyConversionsToFlow`, which rewrites a bare string `condition` into a `{ dialect: 'cel', source }` envelope. That pass walks a flow's **top-level `edges` only** — it does not recurse into the ADR-0031 structured control-flow regions (`loop.config.body`).

A condition left as a bare string inside a loop body falls through to the engine's legacy template path, which substitutes `{var}` templates (a CEL expression has none) and then **string-compares the leftover expression text**:

```
'row.shouldRun == true'  →  'row.shouldRun' === 'true'  →  false
```

## Why this is worse than an ordinary bug

The failure is silent and indistinguishable from success:

- the flow run reports `success: true`
- the sweep queries correctly and selects exactly the right records
- it then does nothing at all
- nothing is logged

There is no way to tell "this sweep had no work to do" from "this sweep's gate is broken" without asserting on side effects.

In hotcrm this meant three flows shipped completely inert: the stalled-deal sweep found every stalled deal and nudged nobody, the renewal sweep never booked a renewal, and the campaign action enrolled no leads. It was caught only by adding runtime tests that assert on records written.

The asymmetry also means the *same expression* behaves differently depending on where it sits in the graph, which is not something a flow author would expect.

## Minimal reproduction

Two flows, identical apart from how the loop-body edge condition is written. Both are handed data that makes the condition true.

```js
// node repro.mjs   — run where @objectstack/* resolves
import { AutomationEngine, installBuiltinNodes } from '@objectstack/service-automation';

const silent = { info(){}, warn(){}, error(){}, debug(){}, trace(){}, child(){ return silent; } };

function makeData() {
  const store = { src: [{ id: 'row1', shouldRun: true }], out: [] };
  let seq = 0;
  return {
    store,
    async find(o) { return store[o] ?? []; },
    async findOne(o) { return (store[o] ?? [])[0] ?? null; },
    async insert(o, data) { const r = { id: `${o}_${++seq}`, ...data }; (store[o] ??= []).push(r); return r; },
    async update() { return { modified: 0 }; },
  };
}

const flowWith = (condition) => ({
[...]

PR #18906, opened 2026-09-18
fix(spec): filter the edge members collectFlowGraphs hands out

### LINKED ISSUE #18102: spec: `collectFlowGraphs` hands out a `FlowGraph` whose `edges` can hold a non-record — the node side #16752 filtered, the edge side it did not
Found while implementing #16910, which re-pointed the flow `edges` readers in `packages/lint` at `recordsOf`. Filed as a finding only, not claimed. #16910 is not addressed by this card.

## The declared contract

`collectFlowGraphs` (`packages/spec/src/automation/control-flow.zod.ts`) declares what it hands out:

```ts
function visit(
  nodes: readonly FlowNodeParsed[],
  edges: readonly FlowEdgeParsed[],
  ...
```

`FlowEdgeParsed` is a parsed edge record. The walk does not hold to it for `edges`.

## Measured — the node side is filtered, the edge side is forwarded

Around `:773`, the NODE side drops a member it cannot read (#16752):

```ts
const kept = nodes.filter((node) => isRegionDict(node));
graphs.push({ scope, path, nodes: kept.length === nodes.length ? nodes : kept, edges });
```

`edges` is pushed through untouched. And a nested region's edge list is admitted on `Array.isArray` alone, around `:786`:

```ts
Array.isArray(slot.raw.edges) ? (slot.raw.edges as FlowEdgeParsed[]) : [],
```

`Array.isArray` proves the LIST, never its MEMBERS. So a `body: { edges: [null, ...] }` — what a YAML `edges:` list item left empty deserialises to — becomes a `FlowGraph` whose `edges` array holds `null` while its declared element type says it cannot.

Reproduce, on `main`:

```ts
collectFlowGraphs({
  nodes: [
    { id: 'start', type: 'start', config: {} },
    { id: 'lp', type: 'loop', config: { collection: 'x', body: { nodes: [], edges: [null] } } },
  ],
  edges: [],
});
// the region graph's `edges` is [null] — declared FlowEdgeParsed[]
```

## Why this is not already closed by #16752

#16752 repaired this exact shape for `nodes`, at this exact producer, for this exact reason: a declared element type that lies is a contract the consumer cannot defend against, because no coercion at a call site can reach a list the producer reads out of a container's open `z.record` config. The repair was applied to one of the two lists the walk hands out. The edge side was not touched and still carries the producer's word about its members rather than a check.

## Reachability today — latent, not live

⛔ Do not read this as a crash report. #16910 re-pointed both `graph.edges` consumers in `packages/lint` (`lintFlowPatterns` and `validateStackExpressions`) at `recordsOf`, and the producer itself never dereferences an edge member — it only forwards them. So there is no reachable throw on the tree today, and the `packages/lint` family sweep is green in both directions at this addressing depth.

What is filed here is the contract lie, and the reason it matters is the one #15793 stated and #16752 acted on: a producer that hands out a member its declared type excludes puts every present and future consumer one unguarded dereference away from the crash, and the consumer coercion that currently saves us is belt, not braces. Remove either and the class re-opens — which is exactly the reading #16751's own note asks the reader not to collapse.

## Notes for whoever takes it

- The shape is settled by #16752's own repair one list over: filter the edge members the same way the node members are filtered, in the same walk, so `FlowGraph.edges` means what it says.
- ⚠️ Check what `analyzeRegion` (`:442`, `for (const e of edges)` reading `e.source` / `e.target` / `e.id`) does with a non-record member — it is a second consumer inside `packages/spec` itself, and it was not part of #16910's `packages/lint` surface.
- The pin belongs beside the existing ones: `packages/lint/src/non-record-object-entry.test.ts` already carries a `flows[].nodes[].config.body.edges` arm (added by #16910), so a producer-side repair is already observable from there; a `packages/spec`-side pin next to #16752's is the one still missing.

---

os-elon-musk:
Fixes #18102

Clause-②: no

`collectFlowGraphs` handed out a `FlowGraph` whose `edges` could hold a non-record. The node side of that same walk has been member-filtered since the repair #16752 landed; the edge side was forwarded untouched, four lines away. One filter per list, side by side, in the same walk — so `FlowGraph.edges` means what its declared type says.

## What the card claimed, re-derived on this tree

All three readings hold, and the line numbers were re-derived rather than copied.

| card's reading | re-derived at | verdict |
|---|---|---|
| node side does `nodes.filter(isRegionDict)` around `:773` | `control-flow.zod.ts:773` | holds, exact |
| `edges` forwarded untouched in the same `graphs.push` | `:774` | holds |
| a nested region's edge list admitted on `Array.isArray` alone around `:786` | `:786` | holds, exact |

The reproduction from the card body, run on `main` at `0b31d90fb`:

```
collectFlowGraphs({ nodes: [start, loop{ body: { nodes: [], edges: [null] } }], edges: [] })
  graph[0] scope=""                 path=[]                              edges=[]
  graph[1] scope="loop 'lp' body"   path=["nodes",1,"config","body"]     edges=[null]
  NON-RECORD EDGE MEMBERS HANDED OUT: [null]
```

## The shape of the repair

- **A drop, in the producer.** `edges.filter(isRegionDict)`, through the same single predicate the node side uses — extended in this PR to name its third caller, so the two lists the walk hands out cannot drift from each other. Copy-on-write per list: a well-formed flow is handed back the very same arrays, by identity, and pays nothing.
- **Not a refusal, and not a looser signature.** No authoring door's accept set moves. `FlowSchema.safeParse` still returns an envelope rather than throwing (the walk runs inside the parse, so the repair has to stay a drop and a skip), the region `safeParse` refusal in `validateControlFlow` still owns and still reports the malformed region, and `FlowGraph.path` still indexes the RAW node list so a Zod issue stays anchored where the author wrote it. The declared element type is unchanged; it is now true.
- **No consumer indexes `graph.edges` positionally.** Both `packages/lint` readers key findings by `edge.id`; the edge-id uniqueness rule in `flow.zod.ts` reads `flow.edges` raw, never `graph.edges`. So a dropped member changes the index and never whether an edge was judged.

## Latent, not live — measured, and NOT for the reason the card gave

The card's reachability argument rests on "#16910 re-pointed both `graph.edges` consumers in `packages/lint`". There are **three** `graph.edges` consumers on this tree, not two.

| consumer | guard |
|---|---|
| `packages/lint/src/lint-flow-patterns.ts:1560` | `recordsOf(graph.edges)` |
| `packages/lint/src/validate-expressions.ts:1462` | `recordsOf(graph.edges)` |
| `packages/services/service-automation/src/engine.ts:9244` | **none** — reads `edge.id` / `edge.source` / `edge.target` straight off each member |

The third one is shielded only by call ORDER: inside `registerFlow`, `canonicalizeStoredFlow` runs `validateControlFlow(parsed)` at `engine.ts:4021`, which refuses the malformed region before `validateFlowExpressions` is reached at `:4058`. Measured:

```
validateControlFlow(the card's repro)
  THROWS Error: loop 'lp' body: invalid region — Too small: expected array to have >=1 items;
                                                Invalid input: expected object, received null
```

So the conclusion "no reachable throw today" survives — by one belt more than was counted, and a belt made of validator ordering rather than of a member check. That is exactly the asymmetry this card was filed about, one layer up. After this change the declared type carries it, and the next consumer of this list needs neither a coercion nor a call-order argument. ⛔ Not written up as a crash report: no throw was measured on the tree.

## `analyzeRegion` — the card's open question, answered

`analyzeRegion` (`:412`, `for (const e of edges)` at `:442`) **throws** on a non-record edge member, measured on all five shapes:

```
null       THROWS TypeError: Cannot read properties of null (reading 'source')
undefined  THROWS TypeError: Cannot read properties of undefined (reading 'source')
a string   no throw — errors: ["edge 'undefined' source 'undefined' is not a region node", …]
a number   no throw — same
an array   no throw — same
```

**But it is not a consumer of `FlowGraph.edges`, and the card's "second consumer" framing does not hold.** Nothing on this tree routes producer output into it. Its three in-repo entry points:

- `validateControlFlow:624` hands it `parsed.data` from a *successful* `slot.schema.safeParse` — the failure branch throws first, so its `edges` members are always parsed records;
- `findRegionEntry` ← `AutomationEngine.runRegion(region: FlowRegionParsed)` — parsed;
- `bpmn-mapping.ts:111` casts a raw `cfg(c).body` region in, which is a pre-existing raw-cast surface of that file and untouched here.

It reads an edge list; it does not read `FlowGraph.edges`. Its declared parameter type is `FlowEdgeParsed[]`, so a throw on input that type excludes is not a defect at that door — hardening it would be the lenient-consumer direction Prime Directive #12 refuses, and the direction #16752's own note refuses ("the repair is a drop, never a looser signature"). Left alone deliberately.

## Where the pin went

Beside #16752's, in `packages/spec/src/automation/region-normalization.test.ts` — the `packages/spec`-side pin the card said was the one still missing. Both halves of that claim were verified first:

- `packages/lint/src/non-record-object-entry.test.ts` already carries the `flows[].nodes[].config.body.edges` arm (`underNestedRegionEdges`, in `SWEPT_COLLECTIONS`), so no arm is added there and that file is untouched;
- #16752's own spec-side pin is the `describe('#16752 — a non-record member never reaches a returned graph')` block in `region-normalization.test.ts`, not in `control-flow.test.ts`. The new `describe('#18102 — …')` sits directly after it, under the same `#4347 — collectFlowGraphs` parent.

31 new cases: five shapes × (three nesting depths + an unguarded-dereference arm + an anti-vacuity arm), plus the card's own reproduction and a per-list identity arm.

## Verification

Union run at `369c4cca5` (the merge commit; `origin/main` merged in at `600b1e2ab` before opening this PR).

**Reverse verification — the pin can fail.** Mutation through `scripts/ablation-replace.mjs`, anchor hit asserted on disk, restore proven by blob equality with HEAD:

```
anchor  "edges: keptEdges.length === edges.length ? edges : keptEdges,"  x1 -> x0
replace "edges,"                                                        x0 -> x1
blob    38cf36cb5fdc -> b293f7924509        (the mutation landed)
result  23 failing tests, all inside the #18102 describe, zero elsewhere
restore blob == HEAD (38cf36cb5fdc) and `git diff HEAD` empty
```

Direction: **red**, as expected. The suite resolves `./control-flow.zod.js` relatively inside its own package, so the mutation reddened it with no rebuild — this ablation has no `dist/` leg to prove.

**Tests.**

| run | result |
|---|---|
| `pnpm --filter @objectstack/spec test` | 488 files / 14159 tests passed |
| `pnpm --filter @objectstack/spec typecheck` | pass (test layer compiles; debt ledger unmoved) |
| `pnpm --filter @objectstack/lint test` | 104 files / 3909 tests passed |
| `... service-automation exec vitest run src/nested-region-parity.test.ts` | 22 tests passed |

The `packages/lint` run is the load-bearing one for the cross-package direction: `non-record-object-entry.test.ts` keeps `RESIDUAL_THROWS` and `RESIDUAL_INVENTED` exact in BOTH directions, so a producer change that flipped a row there would red it. Both stay empty.

**Gates.** Derived from the actual changed paths, never from a named list:

```
node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack --commands
  -> 81 families
81 run, every exit code captured to a file BEFORE any pipe, all 0
node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack --ran ran.txt
  -> 81 derived, 81 run, 0 NOT-MEASURED, 0 UNRUN
     (a DERIVED zero — all 81 recorded an exit code and none of them is 3)
```

Two conditions were cleared rather than reported as reds, because neither measured anything: `check-plugin-teardown-shape --self-test` exited 3 on a shallow clone that could not reach its pinned fixture (cleared by `git fetch --unshallow`, as its own text prescribes), and four gates exited 3 with `PREREQUISITE NOT MET` before the package closure was built (cleared by `turbo run build` over `./packages/*`). The whole set was then re-run at the final `369c4cca5`, so no ratchet reading predates the last commit.

⛔ Derived families are not a complete account of CI on this PR: the 50 artifact-roster families, the 11 wide-population families, the 5 path-scheduled CI jobs and the always-runs tail are each outside that total, and a green from a `--self-test`-only family grades the checker, not this diff.

**Lint, as a proven narrowing rather than a repo scan.** `pnpm lint` is `eslint . --no-inline-config`, a repo-wide scan CI owns. The three facts that make the narrow run a measurement:

1. **Population, read from `eslint.config.mjs` itself** — its file-matching objects declare `**/*.{ts,tsx,mts,cts,js,jsx,mjs,cjs}` and narrower per-root subsets of it, minus `NEVER_LINTED`. A `.md` path matches no config object, which eslint states in its own report for the changeset (`File ignored because no matching configuration was supplied`).
2. **Count, read from `--format json`** — 3 paths handed in, 3 in the report, 2 linted: 0 errors, 0 warnings, 0 suppressed. Exit 0.
3. **Invariance** — `eslint.config.mjs` states, with its own measured positive control, that this repo "never enables type-aware linting (no `parserOptions.project`, no typed `@typescript-eslint` rules) for ANY file, test or not". With no rule reading a type, nothing in this diff can move the verdict on a file this diff does not contain.

## Acceptance notes

Observations, none of them filed — each is noted here for whoever next touches the file.

- **Two `packages/lint` comments now misstate the producer.** `lint-flow-patterns.ts:1555` and `validate-expressions.ts:1459` each say the edge side of this walk "is not" member-filtered, as the justification for their `recordsOf` coercion. Their conclusion stays right (belt and braces, ⛔ not one repair written twice) — only the parenthetical is now stale. Left untouched because the dispatch's declared file surface does not include those two files and a stale comment is not one of the three filing classes. Exact edit needed: in both, "the edge side is not" becomes "both sides are since #18102".
- **`non-record-object-entry.test.ts:378` carries the same stale sentence** in `underNestedRegionEdges`' docblock. Same reasoning; the arm's behaviour is unaffected and the file is untouched.
- **The edge-id uniqueness rule is top-level only.** `flow.zod.ts:1164` walks `flow.edges` and no region's, so two region edges may share an id where two top-level edges may not. A coverage asymmetry, not an error: no example fails, nothing is mis-documented. Not filed.
- **`bpmn-mapping.ts` casts raw region config into `analyzeRegion`.** `cfg(c).body as Region` at `:313` (and the `try`/`catch`/`branches` siblings) is a bare cast off an open `z.record`, so a malformed region reaches `analyzeRegion` as a `TypeError` where that file's own located diagnostic belongs. Reachability was NOT measured, so this stays an un-rehearsed observation rather than a card. Successor: none — no open PR touches that file.
- #16910 is not addressed here, as the card and triage both state.

## 维护者速读(草稿)

**改了什么。** `packages/spec` 里那个把一张 flow 拆成「每个子图」的函数,原先对节点列表做了成员过滤、对边列表没做。这次给边列表补上同一个过滤,一条路一个过滤器,并排放着。

**为什么改。** 它对外声明「我交出去的边一定是一条边」,实际可能交出一个 `null` —— 作者在 YAML 里把 `edges:` 下面一项写空了,反序列化出来就是 `null`。同一个函数、同一次遍历,节点那一半修了、边这一半没修,声明与事实不符的那一半就还在。今天没人被它崩到,但每个现在和将来的消费者都只差一次不设防的取值。本次实测还发现:挡住它的其实不止卡上写的那两道消费端强制转换,还有第三个消费者靠「校验顺序」侥幸躲开 —— 也就是说,原来的安全边界比卡上记的更薄。

**风险与代价(含回滚)。** 行为变化只发生在「声明里本来就说不可能存在」的输入上;结构良好的 flow 按引用原样返回,零代价。不新增任何拒收,作者能写的东西一点没少。回滚 = revert 本 PR 单个 commit,无数据迁移、无配置、无兼容窗口。

**席位意见。**

**你要做的。** 确认这里只补过滤、不新增拒收(这是本卡的裁决前提),然后按常规流程合。若要顺手清掉上面 Acceptance notes 里那三处已过期的注释,那是另一个 PR 的事。

---


os-elon-musk:
**Seat provenance: `needs:contract-review` was mis-hung on this pair by the seat and has been cleared from BOTH carriers. This red was the seat's, not the diff's.**

`domain:spec` seat 3, `session_019srGWGCBBCBHqcDoRZpQRh`, R1.

## What happened, in order

1. At 2026-09-18T05:12Z this seat added `needs:contract-review` to PR #18906 and card #18102.
2. `Check Changeset` then went **red**. Its log names the mechanism: it read `Clause-②: no` from the PR body, and the carrier label made it require at least `minor` for a widening — 「the `needs:contract-review` carrier is a PR label and the `Clause-②:` line is a PR-body line」.
3. ⭐ The decisive reading: an **earlier** `Check Changeset` run on **this same commit** `369c4cca58` **succeeded** (job `105489421788`), and the failing run (`105489658889`) came after the label write. The diff did not change between them. ⇒ the label was the variable.
4. At 2026-09-18T05:20Z the label was removed from PR #18906 and card #18102, **two seconds apart**, deliberately: 「一次合法的清除留下两条相隔数秒的移除,一次剥标只留一条」 — a one-sided removal is indistinguishable from a strip.

## The seat's error, stated precisely

`needs:contract-review` is the carrier for a **`Clause-②: yes`** declaration, and the discipline says ⛔ 不前瞻预挂. This pair declares `Clause-②: **no**`.

What actually applies here is a different rule: the clause-② **path limb** fires because the diff touches `packages/spec/src/**`, and that limb requires an in-seat clause-② review **PASS on record** before the PR may enqueue. ⇒ The seat conflated "a review is owed" with "the `yes` carrier is owed". Only the first is true. The gate's own route 2 names the remedy exactly — 「the level is right and the DECLARATION is wrong -> correct it at the producer: the `needs:contract-review` carrier is the review seat's to place and to clear」 — and ⛔ it warns against adding a tolerance to route around a declaration its author did not mean, which is why nothing about the changeset or the PR body was touched.

## What is NOT being changed, and why

- ⛔ The changeset stays `@objectstack/spec: patch`. The level is correct: this filters a member the producer forwards and moves no authoring door's accept set, which is a narrowing. Raising it to `minor` to silence the gate would be paying for the seat's mistake with a wrong release level.
- ⛔ `Clause-②: no` stays. It is correct on the diff.
- ⛔ No re-run was spent. Removing a label fires its own event, so the enforcing run arrives on its own.

## Still owed before this PR may enqueue

An **isolated at-tier clause-② review** is running against head `369c4cca58` (the path limb's requirement). Its PASS will be recorded before any enqueue, and ⛔ this seat does not self-review it: the seat measured its own served model **below** `CONTRACT_REVIEW_TIER`.

<sub>Written 2026-09-18T05:20Z. The two job ids, their conclusions and the gate's log text were read at 2026-09-18T05:19Z. Seat post: #18883.</sub>


---

PR #18952, opened 2026-09-18
feat(spec)!: publish the two named refinement patterns the runtime already enforces

os-elon-musk:
Part of #18670 (item 2 only — see "What is left" below; item 1 landed as #18729).

Clause-②: yes (narrowing)

Director ruling batch #154 item 3, letter **C** (comment 5725370614, maintainer 「同意」): 「the projection emits a refinement only where the rule is a complete, mechanically derivable JSON Schema pattern — banned keys, required-one-of, non-blank — one ledger row at a time; everything else stays annotated as `x-dropped-refinements`」.

## What this does

`z.toJSONSchema()` has no arm for a `custom` check. On zod 4.4.3 — the version `packages/spec` resolves — a plain record, the same record with a `.refine()`, and the same record with an **aborting** `.refine()` project byte-identically. So every rule written as a refinement reached the runtime and not `packages/spec/json-schema/**`: the published file was **wider** than the contract it is generated from, which is the direction where an author's (or an AI's) validator answers PASS right up to the moment the platform answers NO.

Two of the ruling's four named patterns now project, and only those two:

| pattern | emitted as | sites | why it is EXACT, not approximate |
|:---|:---|--:|:---|
| `required-one-of` | `anyOf` of one `required` per key, conjoined through `allOf` | **137** | A key absent from a JSON object is the only way for its value to read `undefined`, so `required` and `!== undefined` name the same set of documents. A key present with any JSON value, `null` included, satisfies both. |
| `non-blank-string` | `minLength: 1` plus the pattern `\S` | **60** | `String.prototype.trim` removes exactly ECMA-262 WhiteSpace ∪ LineTerminator, and `\S` is the complement of that same set. |

`system/TraceSamplingConfig.json` — the card's own named specimen — now carries no `x-dropped-refinements` at all, and `shared/Expression.json` states the source-or-ast rule, so `{ "dialect": "cel" }` is refused by the published file exactly as the runtime already refused it.

## Proof of work: the shrink-only ledger

`packages/spec/dropped-refinements.baseline.json`, measured by the generator itself:

| reading | before | after |
|:---|--:|--:|
| `publishedSchemasWithDroppedRefinements` | **246** | **201** |
| `droppedRefinementSites` | **750** | **553** |
| `refinementSitesThatDidProject` | **0** | **197** |
| `refinementSitesWithNoJsonFormToCompare` | 3 | 3 |

45 rows deleted outright, 75 rows shrunk, **197 sites closed, 0 sites added anywhere**. The edit was not typed by hand: a throwaway auditor parsed the gate's own "corrected entries, in full" output and refused to write unless the file round-tripped byte-identically through `JSON.stringify(obj, null, 2)`, every removed site was one the post-change census classifies `projected`, no entry gained a site, and the per-entry arithmetic closed. Independently re-proved at JSON level against `git show HEAD:...`: 0 keys added, 45 keys deleted, 0 sites added, 197 sites removed, header equals body.

The generator also prints the closed population **per pattern** on every run, with a line of its own for a site that projects with no declared pattern — that bucket reads **0**:

```
🔇 553 refinement site(s) across 201 published schema(s) reach the RUNTIME and not the published JSON Schema
     Also measured this run: 197 refinement site(s) DID reach the file, 3 had no JSON form on either side to compare.
📣 197 refinement site(s) DO reach the published JSON Schema, by declared pattern:
      137  required-one-of
       60  non-blank-string
```

## The contract: nothing the runtime accepts becomes refused, MEASURED

The ruling is explicit that this is a correction of the machine-readable declaration and ⛔ not a behaviour change. That is a reading here, not an assertion. A probe parses one shared corpus of **6027 documents** across **12 schemas** — `ExpressionSchema`, `EvaluatedExpressionSchema`, the four input unions, `PredicateSchema` / `PredicateInputSchema`, `UpdateAiConversationRequestSchema`, and three deep composers (`FlowSchema.edges[].condition`, `ObjectSchema.titleFormat`, `CronScheduleSchema.expression`) — recording per document the success bit and every issue as a sorted `code@path`. Run in two worktrees, at the merge base and at this head, over the same corpus file:

```
merge-base 46559f61c   cases=12 documents=6027 accepted=1873 refused=4154
this head   eae168e20   cases=12 documents=6027 accepted=1873 refused=4154
cmp exit 0 · sha256 15a714e471f1ba43ec0c0c8773ac345c5c0d03974855ed15116ae2014119008a (both files)
```

Byte-identical, so the accept set, the refusal set and every refusal message and path are unchanged.

⭐ **And the probe can fire.** Under a lit control that weakens `NON_BLANK_STRING` from `source.trim().length > 0` to `source.length > 0` — one token — **732 of the 6027 documents move**. So the zero above is a measurement, not a vacuous pass.

## The two `packages/spec/src/**` files, and why each had to move

The projection cannot READ a predicate's meaning: `.superRefine()` and `.check()` carry no readable function at all (their check def holds only `{ check: 'custom' }`, where `.refine()`'s holds `{ type, check, fn }`), and a projection turning on `fn.toString()` would be a source-text parser. So the pattern is **declared at the refinement's own call site**, and `requiredOneOf` builds its predicate **from** that declaration, so the published `anyOf` and the enforced rule cannot name different keys.

Each change replaces the predicate expression handed to an **existing** `.refine()` and nothing else — no schema shape, no key, no message, no `.strict()`, no optionality, nothing added or removed:

- `src/shared/expression.zod.ts` — `e => e.source !== undefined || e.ast !== undefined` becomes `requiredOneOf(['source', 'ast'])`; three copies of `(source) => source.trim().length > 0` become the shared `NON_BLANK_STRING` (one is inside `typedExpressionStringArm`, so it covers both the cron and the template slot).
- `src/api/protocol.zod.ts` — `p => p.title !== undefined || p.metadata !== undefined` becomes `requiredOneOf(['title', 'metadata'])`, plus its import.

The parse-equivalence reading above is the evidence that both substitutions are behaviour-preserving.

## ⚠️ One place the spelling differs from the ruling's prose, and why

The ruling names 「`anyOf` + `required` for required-one-of」. Emitted as a **top-level** `anyOf` beside the node's own `type: object` and `properties`, that is valid JSON Schema and correct to a validator — and it degraded the reference pages. `scripts/lib/format-type.ts` tests `anyOf` before `properties`, so the node stopped rendering as its object shape: `content/docs/references/system/tracing.mdx`'s `condition` cell went from

```
Record[string, any] | string | { dialect: Enum[...]; source?: string; ast?: any; meta?: object }
```

(angle brackets written as square ones throughout this paragraph — the platform's body sanitizer eats a short angle-bracket fragment even inside a fence, so the real cell reads with the usual generic spelling)

to `Record[string, any] | string | any | any`, and **26 reference pages moved the same way — 182 insertions / 182 deletions**. Those pages are the ADR-0033 authoritative input for AI authors, so that would be a second machine-readable lie traded for the first one. The same `anyOf` + `required` is therefore conjoined through `allOf`: identical to a validator, and `gen:docs` then produces a **zero-line diff**. Both the measurement and the absence are pinned (`⛔ never writes a TOP-LEVEL anyOf`). The PM accepted this reading as more faithful to the ruling than its literal nesting; ⛔ teaching `format-type.ts` to skip pure-`required` branches was declined by name — a shared renderer is the wrong blast radius for an equivalent emission.

## Ablations — three, each restored with proof

Every leg ran through `scripts/ablation-replace.mjs`, so the mutation is proved on disk (anchor count and blob hash) and the restore by blob equality with HEAD plus an empty `git diff HEAD`.

| mutation | expected | observed |
|:---|:---|:---|
| `required-one-of` emits nothing | the ledger's row deletions must fail | `check:authorable-surface` exit 1 — **45 undeclared + 75 miscounted**, exactly the rows this PR deleted and shrank |
| the detector's differential drops the generator's `override` | the ledger can no longer see a closed site | same 45 + 75 red: the coupling is load-bearing, not decoration |
| `NON_BLANK_PATTERN` corrupted to `.` | the equivalence pin must fail | 3 cases red, including every ECMA-262 blank code point |

The second leg is the one that matters for the ruling's mechanism: without it, a site whose rule the file already states would read `dropped` for ever, and 「every site it closes deletes its ledger row」 would be unreachable.

## What is left, and why this is `Part of` rather than a closing keyword

Two of the ruling's four named patterns are **not** taken here, and the census says why rather than leaving it to judgement:

- **banned keys** (`propertyNames` / `not`) — **zero** clean candidates. The nearest sites judge a banned *value* on a string, or an allowed key set that is data-dependent (`ai.paramHints` against the action's own `params`), which is not mechanically derivable.
- **`dependentRequired`** — exactly **one** candidate, 2 sites: `data/SSLConfig`'s `hasCert === hasKey`, which is precisely `dependentRequired: { cert: ['key'], key: ['cert'] }`. Sound and small; deliberately not taken in this round so the verification surface stays two arms wide, per the dispatch's 「landing one or two patterns with the ledger shrinking measurably beats four half-done ones」.

So this request does not carry a closing keyword for the card: the remaining named-arm worklist above is a real remainder, and whoever takes it starts from these two measurements rather than a fresh census. The 553 sites still in the ledger are the ruling's intended terminal state for refinements outside the closed list — they stay dropped and annotated.

## Verification

- `pnpm --filter @objectstack/spec build` · `check:generated` — "All 15 generated artifacts are up to date".
- `pnpm --filter @objectstack/spec test` — **489 files / 14209 tests**, green. `typecheck` green (`tsc --noEmit` + `check:scripts-typecheck` + `check:test-typecheck`: 54 files / 259 errors / 144 pinned signatures, the standing ledger unchanged).
- `pnpm --filter @objectstack/spec exec vitest run --project repo` over the five generator-facing files (`build-schemas-check-mode`, `check-generated-ledger`, `schema-tree-freshness`, `dist-freshness`, `def-key-collisions`) — **5 files / 141 tests**, green.
- `npx eslint . --no-inline-config --format json` — the **whole** repo, no narrowing: **6858 files, 0 errors, 0 warnings**, taken at `eae168e20`.
- Gate families derived by `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack --commands` from the merge base and re-run **in full on this head**: **84 derived, 79 exit 0, 5 NOT MEASURED, 0 UNRUN** under `--ran`. The five: `check:doc-formula-expressions`, `check:dual-build-cjs-loads`, `check:lean-entry-closure` and `check:type-check-debt` each answer PREREQUISITE NOT MET (exit 3 — they sweep the built output of the whole workspace, which is CI's Build Core job), and `check:pm-dispatch-gates` was **killed by the container's foreground cap** at both 200s and 560s without reaching a verdict of its own. ⛔ None of the five reads as a pass.
- `origin/main` merged three times through `scripts/pm/os-regen-merge.sh`, regeneration committed after each merge. The branch delta against `origin/main` is exactly the 11 paths of this change; `control-flow.zod.ts`, `check-duration-unit-keys.ts`, `check-widening-tells.mjs` and `check-adr-0087-registration.mjs` all diff empty against main, so the merges took main's content intact.

## Acceptance notes

⛔ Observations only — nothing below is addressed here, and none of them is filed.

- The `x-dropped-refinements` annotation still does not reach `content/docs/references/**`: the docs renderer drops unknown `x-` keys, so a reference page states a narrowed rule only where it became real JSON Schema keywords. Carrying the annotation onto the page is a docs-surface change. Successor: whoever takes the remaining named arms.
- `packages/spec/json-schema/**` is gitignored and untracked while being shipped through `files[]`, so the narrowing is visible in a review only through the ledger, the reference pages and the generator log — never as a diff of the artefact itself.
- Four other `z.toJSONSchema()` callers inside `packages/spec/src/**` (approval node config, schemaless node config, driver common, metadata-type schemas) serve Studio's SchemaForm at runtime and were deliberately left alone: narrowing those would change what a form refuses, which is the behaviour change the ruling forbids. The override is exported so they can adopt it under a decision of their own.
- The five sandbox builders in `build-schemas-check-mode.test.ts` still mount the committed package-root ledgers from a hand-kept list; nothing holds that list equal to the set the generator actually reads. Carried over from item 1, unchanged here.

## 维护者速读(草稿)

**改了什么** —— 已发布的 `packages/spec/json-schema/**` 过去对「写成 `.refine()` 的规则」一字不提:作者或 AI 拿这些文件校验 metadata,校验通过,平台随后拒收。本 PR 让两条规则真正出现在文件里:「source 与 ast 至少有一个」和「字符串去空白后非空」。共 **197 个站点**从「运行时有、文件没有」变成「两边都有」,只降不升的台账从 **246 个 schema / 750 个站点**缩到 **201 / 553**。卡片点名的样本 `TraceSamplingConfig` 现在一条缺口都不剩。

**为什么改** —— 裁决(批次 #154 第 3 项 C,维护者「同意」)把方向定死:只在规则是**完整且可机械推导**的命名模式时收窄,其余保持注解。这不是行为变化,而是把一份机器可读的声明改成它一直描述的那个运行时。

**风险与代价(含回滚)** —— 风险集中在一处:收窄已发布产物,理论上可能让昨天通过的文档今天不通过。这一点是**实测**排除的,不是论证排除的:6027 份文档、12 个 schema,在合并基与本分支上跑出**逐字节相同**的接受/拒收结果与错误码;并用一条对照变异证明这个探针能发现差异(732 份文档会移动)。代价是新增两个模块与一套等价性 pin。回滚成本低:两个新模块与 5 处调用点的替换是可逆的,台账回退到 246/750 即恢复原状;但回滚会让文件重新对机器撒谎。

**席位意见** ——

**你要做的** —— ① 确认「已发布 JSON Schema 可以向运行时方向收窄」这件事按裁决执行无误(裁决已定,此处仅复核落地与裁决一致)。② 决定剩下两条命名模式的去向:banned keys 实测零个干净候选,`dependentRequired` 只有 `data/SSLConfig` 一个 2 站点候选 —— 是另起一单,还是就此收尾。③ 本 PR 未携带关卡关键词,卡片的开闭由你或 PM 决定。

---


os-elon-musk:
## ACCEPT — contract review PASS, clause-② gate PASS

Reviewed by an isolated at-tier review subagent, dispatched 2026-09-18T0739Z, returned 0755Z. Verdict **PASS WITH FINDINGS**, **BLOCKING: none**, **CLAUSE-② REVIEW: PASS**.

**Tier attestation.** `CONTRACT_REVIEW_TIER` is `claude-fable-5-1` (`scripts/pm/dispatch-gates.mjs:11899`). This seat serves below that tier, so the review was routed to an isolated at-tier subagent rather than done in seat. The reviewer's own first line claimed the tier; that claim is not the evidence, because a subagent cannot self-attest — its `get_session` reads the parent session. The tier was verified instead from the reviewer's own transcript: **85 per-turn model stamps, all 85 `claude-fable-5-1`, zero stamps of any other model**. Read 2026-09-18T0756Z.

**Review footprint.** Branch read via the explicit ref `origin/claude/issue-18670-project-expressible-refinements`, main via `origin/main`, no `FETCH_HEAD` and no working-tree reads. Head re-read fresh from the PR object three times (0740Z, 0751Z, 0752Z), unchanged at `eae168e204b94a99d018779e66101ee1acdbab46`. Two `git fetch`es, no repo writes, no worktree. Executable measurements ran in the reviewer's own scratchpad against blob-identical copies of the branch modules.

### Independent measurements the review added

Beyond judging the author's evidence, the reviewer measured four things itself:

- **`allOf` vs the ruling's literal nesting, with a real validator.** ajv 8.20.0 (draft 2020-12) compiled the emitted node and a clone rewritten to the ruling's literal top-level `anyOf`; over 300 instances (`dialect` present/absent × 7 `source` values × 7 `ast` values including `null`, `''`, `'   '`, objects × 3 `meta` states, plus 6 non-object instances) the verdicts were **identical 300/300**, 60 accepted by both. Lit control: with the rule deleted, `{dialect:'cel'}` is accepted, and both spellings refuse it. 0751Z.
- **The two `packages/spec/src/**` edits, re-classified line by line.** 18 changed lines = 6 comment, 2 import, 10 predicate lines forming 5 `-`/`+` pairs; for every pair the text before `.refine(` and after the predicate argument (including the `{ message: … }` object) is byte-identical, so only the first argument moved. Per-file token counts main→branch are flat: `expression.zod.ts` `.refine(` 4→4, `.optional(` 5→5, `.strict(` 0→0, `z.object(` 2→2, `message:` 5→5; `protocol.zod.ts` 1→1, 184→184, 1→1, 141→141, 10→10. Lit control: the shape-token regex hits 14 context lines of the same diff. 0746Z. This independently reproduces the seat's own 0-schema-shape-token reading.
- **The ledger at JSON level.** keys 246→201, keys added **0**, deleted 45; among the 201 survivors sites added **0**, rows grown **0**, shrunk 75, unchanged 126; sites removed 197 = 60 via deleted rows + 137 via shrunk rows; header equals body on both sides; `measured.zod` 4.4.3 and `refinementSitesWithNoJsonFormToCompare` 3 both unchanged; both files round-trip byte-identically through 2-space serialisation. Shrink-only in the strong sense. 0745Z.
- **`Part of` vs `Fixes`.** 0 closing-keyword hits over body and title (close/closes/closed/fix/fixes/fixed/resolve/resolves/resolved, optional `owner/repo` prefix); lit control on the synthetic string `Fixes #18670` hits 1. 0741Z.

### The two declared deviations, as resolved

**1. `allOf` conjunction instead of the ruling's literal top-level `anyOf`.** Accepted by this seat before the PR opened, on the measurement that the top-level spelling degraded 26 reference pages (182/182 lines) because `scripts/lib/format-type.ts` tests `anyOf` before `properties`; the wider alternative — teaching `format-type.ts` to skip pure-`required` branches, a shared renderer — was declined by name. The review's scope on this was exactly one question, validator-equivalence, and it came back identical 300/300.

The review also surfaced a stronger reason than the one the acceptance rested on: for a node that **already carries a union `anyOf`** (exercised at `packages/spec/scripts/refinement-projection.test.ts:179`), the literal spelling would *overwrite that union*, so `allOf` is not merely equivalent there but the only correct spelling. The acceptance is firmer than when it was made.

**2. File surface beyond the claim's declared set** (`src/shared/expression.zod.ts`, `src/api/protocol.zod.ts`). Necessary rather than incidental: the pattern must be declared where the refinement is written, because it cannot be read back out of a predicate. Independently re-verified predicate-only above. Stated separately in the PR body as this seat asked.

### Non-blocking findings — disposition

Findings 1 and 2 are **latent with no live instance on this head** (verified at source: each declared call site is the only custom check on its node, and `safeExtend` adds none). They become **required work for the next arm on this card**, which is a named successor rather than a hope: #18670 stays open with two arms outstanding, and whoever takes them edits these exact files.

1. **Detector verdict granularity is per node, not per check.** `verdictFor`/`projectOrNull` compares the node with all custom checks against the node with none, so any declared arm on a node marks the whole node `projected`. Measured on the branch's own `collectDroppedRefinements`: `z.string().refine(NON_BLANK_STRING).refine(s => s.startsWith('x'))` yields `dropped: []` and `projected: [{ count: 2, declaredPatterns: ['non-blank-string'] }]`; lit control, the undeclared rule alone reads `dropped`. A second `.refine()`/`.superRefine()` chained onto any of the five declared nodes would then be neither in the ledger nor annotated, with the ratchet green and the generator's UNDECLARED line blind to it (it reports only sites with zero declared patterns) — silently violating the ruling's own `A refinement that is not one of these named patterns stays dropped and annotated`. Fix is small: treat a node as `projected` only when `customs.length === declaredPatterns.length`, else `dropped` conservatively, and/or add a report line for `count > declaredPatterns.length`.

2. **Generator/detector coupling is by convention, not construction** — the higher-risk of the two, because its trigger is broader than the next arm. `build-schemas.ts` (three `toJSONSchema` calls, lines 503/512/531) and `projectOrNull` each pass the override independently. Dropping the `override:` argument on the generator side alone — a plausible merge-conflict resolution — leaves all 197 sites reading `projected` with a green ledger and a green gate while the published file goes wide again with no annotation: the item-1 silence restored, now behind a green ratchet. No ablation covers this leg (a mutates the shared emitter, b the detector side), and no test reads generator *output* for the projection (`build-schemas-check-mode.test.ts` has 0 hits for `allOf|minLength|refinementProjection`; control: 0 tests mention `x-dropped-refinements` either). Fix options: one shared projection helper imported by both, or a sandbox-builder pin asserting `shared/Expression.json` carries `allOf[].anyOf[].required`.

Why these are carried rather than required in this PR: there is no reachable instance today, the guard only becomes load-bearing once a mixed node or a divergent generator call can exist, and the next arm both creates that possibility and regenerates these files anyway. Requiring them here would buy a full re-verification cycle on an `size/xl` PR for protection against an edit nobody has made. Recorded as required work on the card instead, so the next dispatch inherits them as obligations and not as suggestions.

3. **Literal shortfall against the ruling, recorded.** The ruling asks for a changeset `its body naming the patterns projected and the rows retired`. The changeset names both patterns and gives counts (45 deleted / 75 shrunk / 197 sites) plus two examples, but does not name the 45 retired rows. The names are in the ledger diff in the same PR, so the record exists; the at-tier reviewer judged the shortfall non-blocking, and on a question of contract faithfulness that judgement is the one that counts here, since this seat is below tier. The next arm's changeset should name rows.

4. `\S` exactness is ECMA-262-relative, and the review made it concrete: strings consisting solely of U+0085 (NEL) or U+001C–U+001F are accepted by the runtime and by ajv, and refused by a Python `re` reading of `\S`. Lit controls: U+0020 refused by both, `a` accepted by both. Disclosed by the author as a specification appeal; no portable spelling exists (ECMA lacks `\x{…}`, RE2 lacks `\uXXXX`).
5. No validator-backed pin ships in the PR although ajv is already a workspace devDependency (`packages/lint`, `packages/objectql`). The reviewer ran one; the repo does not.
6. `packages/spec/scripts/build-openapi.ts:95` projects nine hand-listed API contract schemas into `json-schema/openapi.json`, shipped in the same `files[]` directory, without the override. None of the nine carries a declared refinement today (`UpdateAiConversationRequest` is not among them: 0 hits, control `BaseResponse` 1), so nothing diverges now; a future arm touching one of the nine would ship two disagreeing machine-readable descriptions in one directory.

### Reviewer's own instrument failures, recorded rather than buried

A lockfile control (`grep -c '/zod@'` over `origin/main:pnpm-lock.yaml`) returned 0 while the target `zod@4.4.3` returned 21 — a control pattern wrong for lockfile-v9 syntax. The reviewer discarded that pair and read the importer entry positively instead (`packages/spec: zod: version: 4.4.3`). A dead control discarded is worth more than a reading kept.

### Landing

`Part of #18670`, no closing keyword — deliberate, and the reason is on record: two of the ruling's four named arms are not taken (banned keys, measured at zero clean candidates; `dependentRequired`, one candidate at `data/SSLConfig`, 2 sites). The card therefore does **not** auto-close at merge, and the partial-landing sequence applies in one write: return #18670 to `pm:queue`, clear the assignee, and post a `Release:` line naming what landed and where the remainder goes, together with findings 1 and 2 as required work for the next arm.

CI at 2026-09-18T0756Z on head `eae168e204b94a99d018779e66101ee1acdbab46`: 35 distinct checks (raw 35, so no superseded runs), 31 success, 3 skipped, 1 in progress (`Lint & Repo Gates`), **0 failures**, every latest run bound to the current head. `TypeScript Type Check` is **success** on this head, together with all four `Type Check · *` constituents including `workspace` — the earlier red on the superseded head `34aeb4261` is resolved by the author's final `origin/main` merge, not waived. Enqueue follows the last gate job reaching `completed: success`; ⛔ not an aggregate reading.


---

os-elon-musk:
## Clause-② carrier stripped from both carriers — record cited, and the seat's own error named

**Why this comment exists.** `Governed Surface Queue Guard` refused the merge group containing this PR at 2026-09-18T08:24:50Z and `github-merge-queue[bot]` ejected it. The guard's two legs separate cleanly in its log, and only the second one fired:

- Governed-surface leg: **✅ CLEAR** — `0 governed pull request(s), 0 unattributed governed commit(s)`. The pre-enqueue reading (`check-governed-merges.mjs --pr 18952`, 0 of 11 paths, exit 0, 2026-09-18T0759Z) was correct and is not what failed.
- Contract-Review Carrier leg: **⛔ REFUSED** — `#18952 — CARRIES needs:contract-review — this pull request may not be in the queue`, exit 6, citing 〈入队与落地〉「双肢命中任一 ⇒ 无席内条款②复核 PASS 在案 ⛔ 禁止入队」 and `references/contract-review.md` 〈载体纪律〉「开着的载体恒 = 真实待审」.

**This was the dispatching seat's error, not a gate defect and not a flake.** The protocol step is 「PASS ⇒ 同席剥标并引记录、ready、auto-merge」 — the strip comes FIRST. This seat had the PASS on record and went straight to ready + auto-merge with the carrier still hung, which is precisely the state the queue leg exists to refuse. The guard did its job; the ejection cost one queue cycle and nothing else.

**The verdict this strip rests on was already on record before the enqueue, not manufactured after the refusal.** Comment `5727034823` on this PR, posted 2026-09-18T07:58:44Z — i.e. 4 minutes BEFORE the enqueue at 08:02:49Z:

- Isolated at-tier contract review, dispatched 0739Z, returned 0755Z. Verdict **PASS WITH FINDINGS**, **BLOCKING: none**, **CLAUSE-② REVIEW: PASS**.
- Tier verified from the reviewer's own transcript — **85 per-turn model stamps, all 85 `claude-fable-5-1`**, zero stamps of any other model — because a subagent cannot self-attest (its `get_session` reads the parent session). `CONTRACT_REVIEW_TIER` is `claude-fable-5-1` (`scripts/pm/dispatch-gates.mjs:11899`) and this seat serves below it, which is why the review was routed out rather than done in seat.
- The clause-② substance: the published artefact narrows only toward what the runtime already refuses, measured with a conforming validator (ajv 8.20.0, draft 2020-12) — 0 runtime-vs-file disagreements across a 300-instance presence/value lattice and 22 whitespace code points — and the runtime itself is unchanged (five predicate substitutions, each textually identical to what it replaced).

⛔ This is not a strip-to-pass. The guard's own text says the label leg cannot tell a carrier stripped before any PASS from one never hung, and that whether a verdict EXISTS is `check-clause2-carriers.mjs`'s question — so that instrument is re-run after this strip and its reading is posted rather than asserted.

**Both carriers were written, 9 seconds apart**, each with the four-step write and a read-back that MATCHED the target: this PR (labels now `documentation`, `size/xl`, `tests`, `tooling`, `domain:spec`) and card #18670 (now `priority:p1`, `pm:dispatched`, `domain:spec`, assignee retained). A one-sided removal is indistinguishable from a strip, which is why neither was left alone.

Nothing about the change itself moved: head is still `eae168e204b94a99d018779e66101ee1acdbab46`, 35 distinct checks with 0 failures, and the ejection was a queue-admission refusal rather than a finding against the diff.


---

os-elon-musk:
## Contract review

Served-tier: 85/85 `CONTRACT_REVIEW_TIER`
Head-sha: `eae168e204b94a99d018779e66101ee1acdbab46`

Tier was verified from the reviewing subagent's OWN transcript, not from its self-report: 85 per-turn model stamps read, all 85 equal to the constant above, zero stamps of any other value. A subagent cannot self-attest, because its `get_session` reads the parent session rather than itself. This seat serves below that constant, which is why the review was routed to an isolated subagent rather than done in seat.

### ① Derived judgments

The published artefact narrows only toward what the runtime already refuses, and the runtime itself is unchanged.

- **File side, conforming validator.** ajv 8.20.0 (draft 2020-12) compiled the emitted node and a clone rewritten to the ruling's literal top-level `anyOf`. Over 300 instances (`dialect` present/absent, 7 `source` values, 7 `ast` values including null, empty and blank strings and objects, 3 `meta` states, plus 6 non-object instances) the two verdicts were identical **300/300**, 60 accepted by both. Lit control: with the rule deleted `{dialect:'cel'}` is accepted, and both spellings refuse it. A further 22 whitespace code points gave 0 runtime-versus-file disagreements.
- **Runtime side, unchanged by construction.** Five predicate substitutions across two contract sources, each textually identical to the predicate it replaced. 18 changed lines classify as 6 comment, 2 import, 5 minus/plus predicate pairs whose text before `.refine(` and after the predicate argument is byte-identical. Per-file shape-token counts are flat: `expression.zod.ts` `.refine(` 4 to 4, `.optional(` 5 to 5, `.strict(` 0 to 0, `z.object(` 2 to 2, `message:` 5 to 5; `protocol.zod.ts` 1 to 1, 184 to 184, 1 to 1, 141 to 141, 10 to 10. Lit control: the same token regex hits 14 context lines of that diff.
- **Author's contract reading, judged for coverage rather than re-run.** 6027 documents across 12 schemas parse byte-identically at the merge base and at this head (`cmp` exit 0, matching sha256), with a lit control moving 732 of the 6027 when the non-blank predicate is weakened by one token. All five source-level `.refine()` call sites from which the 197 closed published sites descend are exercised by those 12 schemas.
- **Ledger monotonicity.** Keys 246 to 201, **0 added**, 45 deleted; among the 201 survivors **0 sites added**, **0 rows grown**, 75 shrunk; header equals body on both sides; the measured zod version and the no-JSON-form count unchanged.
- **Deviation accepted before the PR opened.** The ruling names `anyOf` plus `required`; the same pair is emitted conjoined through `allOf`, because the top-level spelling degraded 26 reference pages (182 of 182 lines) via `scripts/lib/format-type.ts` testing `anyOf` before `properties`. The wider alternative, teaching that shared renderer to skip pure-`required` branches, was declined by name. Review scope on it was validator-equivalence alone: identical 300/300. The review additionally found that on a node already carrying a union `anyOf` (exercised at `packages/spec/scripts/refinement-projection.test.ts:179`) the literal spelling would overwrite that union, so `allOf` is the only correct spelling there.

### ② Semver level

`@objectstack/spec` **minor**, with the BREAKING annotation — the documented combination for a clause-② narrowing rather than a major. ADR-0087 disposition `not-required (no-migration-prescription)`, in the gate's own spelling, accepted by `check-adr-0087-registration` as BREAKING plus clause-②-narrowing. Recorded shortfall against the ruling's literal ask: the changeset gives counts (45 deleted, 75 shrunk, 197 sites) rather than naming the 45 retired rows; the names are in the ledger diff in the same PR, and the at-tier reviewer judged the shortfall non-blocking. The next arm's changeset should name rows.

### ③ Boundary flags

- Clause-② narrowing: **yes**, declared line-initial on both carriers.
- Governed surface: **none** — `check-governed-merges.mjs --pr 18952` derived 11 paths three-dot and hit 0 of 5 surfaces, exit 0.
- Release-owned surface: **untouched** — 0 paths under `content/docs/releases/`, control 16 files present there on main.
- Card closure: `Part of`, **no** closing keyword bound to the card (0 hits over body and title, lit control 1 on a synthetic `Fixes` string), because two of the ruling's four named arms are not taken.
- Public API surface: unchanged — the two new modules are exported from no barrel.
- Carried to the next arm, neither with a live instance on this head: detector verdict granularity is per node rather than per check, so a second refinement on a declared node would be neither ledgered nor annotated while the ratchet stayed green; and the generator/detector override coupling is by convention only, so a merge-conflict resolution dropping it generator-side would leave a green ledger over a wide file, a leg none of the three ablations covers.

Implemented-by: `claude/issue-18670-project-expressible-refinements`

**VERDICT: PASS**

Full reviewer report and its per-item table: comment `5727034823`. Carrier clear and the seat error behind it: comment `5727351098`.


---

PR #18946, opened 2026-09-18
chore(gates): retire check-type-source-resolution (maintainer ruling E on #18373)

os-try-charles:
Fixes #18373

Clause-②: no

`skip-changeset` — measured, not asserted; see Verification below.

Executes the maintainer ruling of 2026-09-18 on this card (batch #153 item 3, **letter E**):
retire `check-type-source-resolution`. This PR is EXECUTION — it does not re-argue A/B/C/D.

## What the ruling ordered, and where each piece landed

| the ruling's words | landed as |
|:--|:--|
| delete `scripts/check-type-source-resolution.mjs` and its self-test | file deleted. Its "self-test" was the file's OWN `--self-test` dispatch, not a separate file, so it went with it (`git ls-files` matched exactly one path for the name). |
| `KNOWN_DIST_RESOLVED_TYPE_IMPORTS` with it | that registry lived inside the deleted file; the identifier now has **0** occurrences anywhere in the tree. |
| the `check:type-source-resolution` entry in the root `package.json` | removed (one line). |
| the 「Type-source resolution gate」 step in `.github/workflows/lint.yml` | step removed, together with the 23-line comment block that exists only to explain it. |
| `AGENTS.md` and any doc that names the gate as a standing check | **`AGENTS.md` names it zero times — re-measured here, see §1. This diff does not touch the governed surface.** |
| `docs/audits/gate-census-2026-09.md:215` verdict | rewritten to the ruling's exact text, see §2. |
| PR #18708 | not touched. |
| `check:test-source-alias` (the vitest-axis sibling) | not touched, and its ledger is not touched. |

## 1. `AGENTS.md`: the ruling's sentence describes a sentence that does not exist

Re-measured independently of the dispatch, whitespace-**flattened** first so wrapped prose
cannot give a false zero, every zero paired with a control drawn from the same flattened
population:

```
type-source-resolution        0        check:test-source-alias   1   (control, hits)
Type-source resolution        0        check:                   47   (control, hits)
type source resolution        0        test-source-alias         1   (control, hits)
TYPE SOURCE RESOLUTION        0
KNOWN_DIST_RESOLVED           0
check-type-source-resolution  0        file 78,857 bytes / flattened 76,881 bytes
```

My reading **agrees with the dispatch's**. So the ruling's `AGENTS.md` clause has no
referent, no line was hunted for there, and `AGENTS.md` / `CLAUDE.md` / `.claude/**` /
`skills/**` / `docs/adr/**` are all absent from this diff.

[...]

PR #18903, opened 2026-09-18
skills(pm-dispatch): key the clause-② contract review by lane — spec and skills owe it on every round, other lanes owe none

os-elon-musk:
Fixes #18536

Clause-②: no

## What this lands

The maintainer's lane rule, restated by the director record on the card (comment 5717169811), quoted verbatim and untranslated:

> 「曾经要求只有 spec 和 skills 需要 fable,其他 opus 就够了,理论上其他车道不需要契约复审」「即使项目经理跑在 opus 上,也可以支持起 fable 子 agent 复审,这样最快。除非子 agent 用 fable 启动失败,也就是没有 fable」

carried into the governed text and the two machine readers:

1. **Spec and skills lanes** — every delivered round gets the contract review at `CONTRACT_REVIEW_TIER`: in-seat when the seat's served tier is that tier, otherwise by the at-tier review subagent the seat spawns (the 09-17 route stays). The record is the same-shape comment with `Served-tier:`, on the PR or the card.
2. **Every other lane** — no contract review. The clause-② limbs (path `packages/spec/src/**`, declaration `Clause-②: yes`) survive unchanged and now answer the LANE question: a limb hit is spec-lane work and moves there; a `Clause-②: no` PR outside the contract surface lands on the three pre-checks and the gates. No default-tier self-review record is demanded, and no other lane spawns the at-tier subagent.
3. **Tier unavailable** (the at-tier subagent cannot start) — the review cannot be produced: the PR stays draft and out of the queue, and the maintainer's own review is the only bypass, by their word each time.

The lane key is the 2026-09-10 key restored (PR #17294) with the 2026-09-17 subagent route kept (PR #18511); the 2026-09-16 tier key (PR #18363) is superseded. Not a byte-revert: both spec AND skills owe the review, and 「余席默认档自审加门禁」 is gone — other lanes owe nothing.

## Charter (commit 1) — net 0 lines per file

| file | before → after | ratchet ceiling |
|:--|:--|:--|
| `.claude/skills/pm-dispatch/SKILL.md` | 812 → 812 | 812 |
| `references/contract-review.md` | 60 → 60 | 60 |
| `references/core-rules.md` | 151 → 151 | 151 |

Every edited line is at most 120 bytes; `check:pm-skill-ratchet` is green on `bc0c2ec41`. Rule text carries no issue numbers (`check:pm-skill-id-lint` green); the four-axis frame block is untouched (`check:skill-frame-sync` / `-freshness` green).

[...]

PR #18889, opened 2026-09-18
feat(spec): ship a per-release section in spec-changes.json, verified against both tarballs

os-litant:
Fixes #17080

Clause-②: yes (widening)

Ruled **A** (issue comment 5643392133, decision batch #119 item 2, maintainer 「同意」 2026-09-12):
minors ship real, machine-readable change data, with a correctness gate. All four execution
clauses land here.

## The defect, re-measured from the published tarballs

`npm pack @objectstack/spec@17.3.0 @objectstack/spec@17.4.0`, then the repo's own export-surface
row convention (`ENTRY: NAME (KIND)` — entry point, export name, export kind — the spelling
`build-spec-changes.ts` flattens to):

| | 17.3.0 | 17.4.0 |
|---|---|---|
| export-surface rows | 5259 | 5433 |
| **true delta** | — | **225 added, 51 removed** |
| `spec-changes.json` → `aggregate.added` / `.removed` | 0 / 0 | **0 / 0** |
| `protocolVersion` | 17.0.0 | 17.0.0 |
| `perMajor[16→17]` | converted 58, migrated 77 | converted 57, migrated 77 |
| `[REMOVED]` prescriptions in `json-schema/**` | 156 | 221 |

Every figure the card reports reproduced. The shipped manifest answered `0 / 0` over a release
that moved 276 exports, and a consumer reading semver sees a minor.

## What lands

**1 · A per-release section, shipped in the tarball.** `spec-changes.json` gains `release` —
`fromVersion` → `toVersion` at package-version resolution, with `added` / `removed` (the exports
that arrived and left, each named) and `converted` / `migrated` (the ADR-0087 D2/D3 entries first
registered in that release). Generated at **publish time only**, from the previously published
tarball: the committed copy stays the deterministic registry projection and `check:spec-changes`
is green against it, which is how the determinism concern is answered rather than traded away.

**2 · The correctness gate, as acceptance.** Before anything reaches npm, the release lane packs
the artifact it is about to publish and recomputes the delta from **the two tarballs**, with its
own reader — `scripts/check-release-spec-changes.mjs` deliberately does not import the generator's
flattening, because an instrument that shares the code it audits reports agreement with itself.
A mismatch fails the release, naming the disagreeing exports and the **direction** of each
disagreement (claimed-but-not-real, real-but-unclaimed, per array) plus the command that
regenerates the section. A held release arrives with its own diagnosis.

**3 · `os validate --json` reads the same data.** New key `specReleaseChanges` — `fromVersion`,
[...]

PR #18688, opened 2026-09-17
feat(spec)!: a structured region body refuses a pause-capable node and an 'end' node

os-litant:
Part of #15646

Clause-②: yes

The flow accept set shrinks for five node types inside region bodies — shapes the runtime never honoured. Ruling D clause 4 states it verbatim.

⚠️ **The runtime half is NOT in this PR.** A region-contained node that *durably suspends* must fail the run with a named error — only the run can know that — and ruling D assigns it to a separate `domain:services` card. ⇒ this PR lands with `Part of`, ⛔ not `Fixes`; **#15646 stays open until the runtime half lands.**

---

## ⚠️ SEAT BANNER — ruling **D** superseded the route this body argues for

Everything below the horizontal rule was written when the card was ruled **C**, and it argues for **route A** ("Recommendation: A, as implemented"). ⛔ **That is no longer what this PR does.** It is kept unedited as the record of how the decision was reached — ⛔ deleting it would erase the evidence the later ruling was made on.

**What this PR does NOW**, per ruling D (batch #153 item 1, comment `5724940095`, maintainer 「其他同意」):

- Inside `loop` / `parallel` branch / `try_catch` (try **and** catch) bodies at any depth, `FlowSchema.superRefine` refuses **five** node types: `screen`, `wait`, `approval`, `approval_revise`, and `end`.
- ⛔ **`map` and `subflow` are NOT refused by type.** They pause exactly when the child flow their `config.flowName` names pauses — a *different metadata record*, not in hand at parse. Refusing them by type would also refuse `loop { map(synchronous child) }`, which runs correctly today.
- `packages/spec` keeps its published identifier `FLOW_PAUSE_CAPABLE_NODE_TYPES`; only its **contents** narrow.
- **`packages/services` is untouched by this round** — measured, zero paths. The 5-tests-in-3-files cost the section below describes **does not occur**: the whole package runs **138 files / 1652 tests, all passing**, with a false-green control proving the test read the rebuilt artifact and not a stale `dist`.

[...]

PR #18971, opened 2026-09-18
feat(spec): pin every export by its .d.ts declaration text, and retire the 27 signature hashes

### LINKED ISSUE #16045: [finding] @objectstack/spec pins 5309 exports by name-and-kind but only 27 by signature — 99.5% of the ratified surface passes green through a shape change
Filed by the `domain:cli` execution PM seat (#6024) from a measurement handed back by the dev agent on #15630, using [Claude Code](https://claude.ai/code). Not dispatched: the fix is a design decision about a much larger surface, not a mechanical repair.

## The finding

`@objectstack/spec` carries the same defect class that #15630 repaired in `@objectstack/cli`'s `./hook-body` pin, at a very different scale:

- `api-surface/*.json` pins **5309 exports across 17 entry points** by **name and kind**.
- `api-surface-signatures.json` pins **27 of them** by **signature hash** — **0.5%**.

A signature change, a renamed interface field, or a dropped union member moves neither the name nor the kind. So for the other **5282** exports, the surface behaves exactly like the one #15630 repaired: the pin is green through a breaking shape change to a ratified public type.

## Why this is worth a decision rather than a patch

#15630 established the shape of a real fix and what it costs. Its repair compiles a real consumer fixture with a real `tsc`, from the consumer directory, against the **packed** `.d.ts` reached through the `exports` map — chosen over a signature snapshot deliberately, on the grounds that a compile asks *whether a dependent can still build* while a snapshot only asks *whether the shipped bytes moved*, and that a snapshot's known failure mode is that a red one gets accepted rather than investigated.

That reasoning does not obviously scale from 4 exports to 5309. Which of these is right for `@objectstack/spec` is a design call:

- widen `api-surface-signatures.json` toward full coverage (cheap to run, weak instrument, accept-the-snapshot failure mode at scale);
- a consumer-compile conformance fixture per entry point (strong instrument, 17 fixtures, real build cost);
- something scoped — e.g. signature pins only for types ADR-0087 treats as ratified.

⛔ This seat is not grading which. Filing so the choice is made deliberately rather than by the gap persisting.

## Two neighbouring surfaces, measured in the same pass

- **`@objectstack/console` does NOT share the gap** — this answers a question #15630's card itself left as NOT MEASURED. It exports `./package.json` and nothing else, declares no types, and has no type surface behind any name. The spelling is identical; there is nothing there to widen.
- **`@objectstack/cli`'s own `./console` subpath has no surface pin at all** — neither names nor shapes — and it points straight at an internal barrel with 13 exports, so every export that module gains is published the moment it lands. The existing pin asserts only that `./console` is a declared subpath. Filed separately as its own card.

Taken together: for the surfaces #15630's card named, its repair is **one fix, not a family** — but `@objectstack/spec` is a genuinely separate and much larger instance.

## Provenance

Measured on the `#15630` worktree at the head that became PR #16043; handed back rather than acted on, per that dispatch's instruction to return out-of-scope findings with their measurement.

⚠️ Not verified by this seat independently — this is the dev agent's measurement, relayed. The counts (5309 / 17 / 27) should be re-derived before anyone acts on them.

os-elon-musk:
Fixes #16045

Clause-②: yes (widening)

Ruled at `5560224701` (director batch #60, 2026-09-06, maintainer verbatim 「同意」), re-affirmed by triage at `5724532096`: option A, a readable declaration-text snapshot, ⛔ not a hash. The card body's three mutually exclusive routes predate that ruling and were not re-litigated here.

`@objectstack/spec` pinned its public surface on one axis. `api-surface/` records each export as `name (kind)`, and a signature change, a renamed interface field and a dropped union member move **none** of those rows. The only shape pin was `api-surface-signatures.json`: 27 rows, and reference-level even there. This adds `api-surface-declarations/`, the declaration text the packed build actually emits for every export of every published entry point, and retires the 27 hashes it subsumes.

## The counts, re-derived on this head before the first generation

The ruling asks for this by name; the card's own numbers were self-declared unverified and 12 days old.

| Number | Card | This head (`b33898f5d`) | Unit, and what would make it something else |
|---|---|---|---|
| entry points | 17 | **17** | type entry points in the `exports` map — those whose `require.types` ends in `.d.ts`. Adding or removing one such subpath. |
| `exports` map entries | (not stated) | **19** | every key in the map. The extra two are `./openapi.json` and `./package.json` — asset subpaths with no declaration at all, filtered out by the same `.d.ts` test `build-api-surface.ts` has always applied. ⇒ premise 1 resolved: **17 is right and the map did not grow**; 19 counts two things that were never entry points. |
| pinned rows | 5309 | **5336** | `name (kind)` rows summed over the 17 `api-surface/` shards. +27 since the card. Ratio unmoved: 27/5336 = 0.51%, so the headline 99.5% stands. |
| distinct exported names | (not stated) | **5200** | (entry, name) pairs. The gap to 5336 is dual-declared names, which are two rows by design. |
| signature hashes | 27 | **27** | top-level keys of `api-surface-signatures.json`. Bright control: the first value really is a `sha256:` string, so this counts signature entries and not empty objects. |

Premise 3 also holds: all 17 packed `.d.ts` files exist and resolve through the map (3,215,437 bytes for the root entry down to 13,081 for `./integration`). No entry point lacks a packed declaration, so the gap the dispatch reserved for itself did not open.

## What the artefact costs — premise 4, which nobody had costed

| | |
|---|---|
| shards | 17, one per entry point |
| declaration blocks | 5336 |
| bytes | **12,661,943 (12.08 MiB)** |
| lines | **237,706** |
| gzipped | **1,071,825 (1.02 MiB)** — against this package's ~17.57 MiB compressed `dist`, so about **+5.8%** of tarball |
| largest shard | `system.txt`, 3,592,701 bytes / 73,283 lines |
| median declaration | **81 bytes** |
| skew | the 20 largest declarations hold **~65%** of all bytes; four exceed 20,000 lines each (`EnvironmentArtifactSchema` 21,868, `ObjectStackDefinitionSchema` and `ObjectStackSchema` 21,851, `ChangeSetSchema` 20,395) |

Stated plainly, as the dispatch asks, and ⛔ not as a veto: the packed `.d.ts` is a tsup dts rollup, so a Zod schema's declaration is its **fully expanded** structural type. That expansion is exactly what makes an inner field rename visible — and it is also why a single schema can produce a 21,000-line diff. The ruling's stated reason for choosing text over a hash is that the contract-review seat reads the diff; that reasoning holds per declaration and is worth a second look at the top twenty. One reading, for whoever wants it: 31% of declarations hold 97.7% of the bytes, so nothing cheap is available by trimming the tail.

## Both instruments, measured on one tree at one commit

The card's thesis is that the old pin cannot fail on a shape change. Not argued — ablated, with the mutation proven on disk by blob hash and the mutation proven to have reached `dist/` before any verdict was read.

**A. the source-level control — a renamed interface field, the card's own class.** `JobRunOutcome.reason?` renamed to `degradationReason?` in `packages/spec/src/contracts/job-service.ts` (blob `363443e2` to `d4b1520c`), spec rebuilt, `ablation-dist-preflight` exit 0 confirming the marker reached the built artefact:

```
check:api-surface              exit=0    "public API surface unchanged"      [BLIND]
check:api-surface-declarations exit=1    "~ JobRunOutcome (interface)"       [SEES IT]
```

Restore leg: blob back to `363443e2`, rebuilt, `ablation-dist-preflight --absent` exit 0 (marker gone from all 214 built files), `git diff HEAD` clean, gate back to exit 0.

**B. the gate can fail on its own artefact.** One field renamed inside `qa.txt` by hand (blob `3f5efb04` to `5b86fec2`, injected occurrences 1, deleted text 0): exit **1**, attributed to `TestSuiteSchema (const)`, failure text naming the regenerate command. Restored to the HEAD blob, `git diff HEAD` empty: exit **0**.

## The retirement, and the coverage proof the ruling demands

All **27** signature names resolve to a declaration block in `api-surface-declarations/root.txt`, **0 missing** — enumerated from `defineAction` through `defineWebhook`, each as `(function)`.

One honest qualification, because the subsumption is not uniform. For those 27 factory declarations the text is `declare function defineAction(config: z.input of ActionSchema): ActionParsed;` — a type **reference**, exactly as blind to an inner-key narrowing as `typeToString` was. What is gained is not sharper text on the 27; it is the **5309 other declarations**, including `ActionSchema` itself, whose own expanded block is where such a narrowing shows up. So the retirement is a strict superset of pinned declarations, not an equal trade. Nothing published read the retired file — it was never in this package's `files[]`.

## Where it lands, and why there

- Generator: `packages/spec/scripts/build-api-surface-declarations.ts`, beside the eight sibling artefact generators, reading the same input through the same `collectEntries` logic. The ruling says "one generator script under `scripts/`"; this reads that as the directory the whole family lives in, because the artefact reads the **built dist** and only the lane that builds spec can run its gate.
- Artefact: `packages/spec/api-surface-declarations/ENTRY.txt`, a sibling **directory** of `api-surface/`. Not inside it: `listShardNames` throws on any file in that directory that is not a `NAME.json` shard, so `api-surface/` is closed by construction. No existing `api-surface/*.json` is regenerated by this PR (`check:api-surface` green throughout), which keeps it clear of PR #18688 and PR #18319.
- Gate: `check:api-surface-declarations`, a step in lint.yml's `Type Check · consumer gates` lane after the two build steps, with `check:api-surface` and the other dist-reading gates. **No new required context** — a step in an existing lane. Registered in the `check:generated` ledger, in `REGEN_ARTIFACTS`, and in `.gitattributes` as `merge=os-regen`.
- Sharded per entry point from day one, for the reason its neighbour is: the merge queue rebuilds server-side where no custom driver runs, so two PRs sharing one generated file evict the second. Pit 1 from `5715457322` is answered by the layout rather than by an assumption — and `check:merge-driver`, which reconciles `.gitattributes` against `REGEN_ARTIFACTS` in both directions, is green over the swap.
- Published, with the reason the gate demands. `check:published-files` refuses a `files[]` entry that carries none; the registered line says what a consumer does with it — read two published tarballs and see *which declared shape* moved between releases, the question `api-surface` cannot answer. If 1.02 MiB of tarball is judged too much, one line of `files[]` removes it without touching anything else.

Three registries had to learn about the new gate, each because it discovered the gate on its own rather than because a list named it:

- `check:published-files` — demanded the reason above.
- `scripts/pm/dispatch-gates.mjs` — its live manifest edge gave the new gate a population before anything listed it, which is the eighth member of a class whose seventh was recorded the same way. Declared as `CLASS_EIGHTH`, with a case asserting the edge really reaches it.
- `scripts/pm/check-widening-tells.mjs` — `PUBLISHED_SURFACES` is derived from `REGEN_ARTIFACTS`, so retiring the signatures row dropped it off that surface and reddened two self-test cases. Both are retargeted to state the retirement as a counterfactual (the surface follows the table, not a literal); ⛔ the new artefact is **not** added to that surface, because the ruling assigns "is a snapshot diff a Clause-② signal" to the skills seat by name and out of this card's scope. Both directions are now pinned, so the boundary is declared rather than forgotten. 483 cases pass, up from 481.

## Verification

- **Gate families**: derived with `node scripts/pm/dispatch-gates.mjs --repo objectstack-ai/objectstack --commands` from the merge base, 120 commands, every exit code redirected to a file and read back. **All 120 green.** Four returned exit **3** PREREQUISITE NOT MET on first pass (`check:doc-formula-expressions`, `check:dual-build-cjs-loads`, `check:lean-entry-closure`, `check:type-check-debt`); each names a build, each was built and re-run green, and none is recorded as a finding. Reconciled with `--ran`.
- **Tests**: `@objectstack/spec` local project **488 files / 14,182 tests passed**; the tooling suites that name the edited scripts, both projects, **10 files / 220 tests passed** (`sharded-artifacts`, `check-generated-ledger`, `dist-freshness`, `dist-freshness-adoption`, `api-surface-dual-kind-rows.pin`, `build-schemas-check-mode`, `def-key-collisions`, `root-index`, `export-list`, `docs-import-surface`). `pnpm --filter @objectstack/spec typecheck` green.
- **eslint, the union rather than a narrowing**: `eslint . --no-inline-config --format json` at `b33898f5d` examined **6856 files**, **0 errors, 0 warnings**, exit 0. The population is eslint's own config resolution and the count is read from its JSON output; type-aware linting is not enabled in `eslint.config.mjs` (no `parserOptions.project`, no typed rules), so this diff cannot move an untouched file's verdict either way.
- **Control bytes**: `check:nul-bytes` green over 8906 files, plus a direct scan of all 31 changed paths for the wider control-byte class — no matches.
- `scripts/check-single-claim-paths.mjs` in the diffstat is **not mine**: it arrived with the one-commit `origin/main` merge (`16cb493d5`) this PR carries.

## Acceptance notes

- `.claude/skills/spec-property-retirement/SKILL.md` line 124 lists `api-surface-signatures` as an instance of a retirement shape, and that row goes stale with this landing. ⛔ Left untouched on purpose: `.claude/**` is a governed surface, so editing it would make this whole PR maintainer-landed for a one-word prose nit. Noted, not filed.
- `packages/spec/scripts/build-schemas.ts` line 830 carries the same stale mention. Left untouched because PR #18952 holds that file; noted, not filed, with the later lander as the natural carrier.
- Three files in this diff are held by open PRs and were edited anyway because the retirement forces it, not by choice: `scripts/pm/check-widening-tells.mjs` (PR #18948), `scripts/pm/dispatch-gates.mjs` (PR #18903) and `.github/workflows/lint.yml` (PRs #18946, #18889, #18414). All are hand-written files where a text conflict is visible rather than silent, and all three of my hunks are small and far from theirs. Whoever lands second resolves.
- The top-20 skew above is a reading, not a finding: no gate is wrong and nothing is unenforced. It is recorded here because the ruling's own justification for text over hash is per-declaration readability, and at 21,000 lines a declaration that argument thins out.

## 维护者速读(草稿)

**改了什么。** `@objectstack/spec` 从今天起为它的**每一个**公开导出留一份"形状快照" —— 不是哈希,而是打包后 `.d.ts` 里那段声明原文,按入口点分成 17 个文件签入仓库,并配一道 CI 闸门:重新生成后对不上就红,失败信息里直接给出重新生成的命令。同时退休了旧的 27 条签名哈希文件。

**为什么改。** 原来的 pin 只记"某个名字还在不在",5336 行里只有 27 行能看出"形状变没变"。也就是说:把一个接口字段改名、砍掉一个联合成员、改一个函数签名 —— 这些都是会让客户升级后编译失败的破坏性改动 —— 全部一路绿灯。本次 PR 里有实测:改了 `JobRunOutcome` 的一个字段名之后,旧闸门 `check:api-surface` 退出码 **0**(看不见),新闸门退出码 **1**(点名了那个 interface)。路线是 2026-09-06 决策批次 #60 里您逐字「同意」的那一条。

**风险与代价(含回滚)。** 代价是体积:12.08 MiB 文本、23.7 万行,压缩后 1.02 MiB,相当于 npm 包增长约 5.8%。更值得注意的是分布极不均匀 —— 最大的 4 个 schema 各自超过 2 万行声明文本,一旦它们变动,复核席位面对的是一份 2 万行的 diff;而裁决选"文本不选哈希"的理由恰恰是"diff 可读"。这一点我按实测如实报告,未自行改动路线。回滚成本很低:从 `files[]` 去掉一行即可停止随包发布;整道闸门回滚就是撤销本 PR,不留任何数据迁移。

**席位意见。**

**你要做的。** 只有一件事需要您判断:12 MiB / 23.7 万行这个量级,以及最大 4 个 schema 的 diff 可读性,是否仍符合当初选 A 方案时的预期。若认为需要收窄,那是裁决层面的一次增补,不是本 PR 的返工。其余部分已按裁决落地并自证。

---


os-elon-musk:
## `Test Core (1/6)` failed on this head — measured, and the single re-run is being held deliberately

Failing check: **`Test Core (1/6)`**, head `b33898f5d8d0dc8eff732eb8ac26452506b8622e`, job `105540309254`, read 2026-09-18T0854Z.

The failure itself, from the job log:

```
FAIL packages/qa/vitest-filter-preflight/test/config-wiring-sweep.test.ts
     > packages/cli/vitest.config.ts
     > ⭐ CONTROL — refuses NOTHING when no override is named
Error: Test timed out in 5000ms.   (test/config-wiring-sweep.test.ts:306:3)
Test Files  1 failed | 2 passed (3)
     Tests  1 failed | 110 passed (111)
```

### Why this is very likely not this pull request's

Measured on the diff rather than assumed:

- The 31 changed files contain **0 files under `packages/qa/**`**, and **0 changed paths matching `vitest`**. Lit control on the same instrument: **23** changed files under `packages/spec/`, so the zero is a real absence and not a dead query.
- The failing subject is `packages/cli/vitest.config.ts`, which this diff does not touch either.
- The failure is a **5000ms timeout on a test that spawns a real vitest child process**. In the SAME run, the same-named `⭐ CONTROL` leg passed for other configs at **2098ms** and **1442ms** — so this family habitually runs at roughly 30–40% of its own limit, and a loaded runner crossing 5s is the timing shape rather than a behavioural change.
- 110 of 111 tests in that package passed.

### Two grounds I checked and did NOT get to rely on, stated so this is not read as hand-waving

- **Not red on the base branch.** On main tip `89c6ec52b5`, `Test Core (1/6)` is `success`, and main has 0 failing checks. The "red on base too" limb does not hold here.
- **It did not pass earlier on this exact commit.** There is exactly **one** run of that check name on this head. The names with repeated runs on this head are `Auto Label`, `Check PR Size`, `Packed-tarball smoke (opt-in)` and `Check Changeset` — not `Test Core`.

So the only remaining ground for "not this PR's" is that it reproduces identically on one re-run, and that re-run has not been spent.

### Why the re-run is being held rather than fired now

At the time of reading, **16 checks were still in progress** on this head. The standing rule allows at most one re-run in total, so spending it on a single job while other jobs could still fail would waste it. It is held until the run converges, then spent once across the whole failure set.

⛔ Not done and not on the table: skipping, disabling or quarantining the test; an empty commit or a close/reopen to kick CI; editing the check.

### One more reading worth recording

The same job log shows `@objectstack/spec:test` was **scheduled but never reached** — "the run stopped before it" — along with 10 other packages (`check-test-completeness: OK (1 of 12 scheduled package(s) reported … 11 never reached)`). This shard therefore has **not yet exercised this pull request's own package at all**, so it is no evidence either way about the change itself.

### What happens next

- Re-run reproduces identically ⇒ the failure is not this PR's by the standing definition, and the timeout gets its own card rather than being ignored: whoever finds a flake fixes it or files it, never routes around it.
- Re-run passes ⇒ flake confirmed on a timing-sensitive subprocess test, and it still gets that card, because a green re-run is not a root cause.

Either way this PR is not landed on a red head, and the contract review for it is proceeding separately.


---

os-elon-musk:
## Contract review

Served-tier: 88/88 `CONTRACT_REVIEW_TIER`
Head-sha: `b33898f5d8d0dc8eff732eb8ac26452506b8622e`

Tier was verified from the reviewing subagent's OWN transcript, not from its self-report: 88 per-turn model stamps read, all 88 equal to the constant above, zero stamps of any other value. A subagent cannot self-attest, because its `get_session` reads the parent session rather than itself. This seat serves below that constant, which is why the review was routed to an isolated subagent instead of being done in seat.


### ① Derived judgments

The ruling (`5560224701`, director batch #60, maintainer verbatim 「同意」) required a readable declaration-text snapshot of every export of every published entry point, ⛔ not a hash. Independently re-measured by the reviewer against the branch blobs:

- **Text, not a hash.** First-token histogram over all **5,336** blocks: `type` 2,376 · `declare const` 2,076 · `declare function` 447 · `interface` 430 · `declare class` 7; **0** blocks fail to start with a declaration keyword, **0** `(other)`/`(namespace)` kinds, **0** inline `import(...)` types. No digest anywhere in the artefact or the check path — the verdict is a byte compare of regenerated against on-disk shard text.
- **Coverage is total and symmetric.** Per-shard set comparison of the `api-surface` JSON rows against the `api-surface-declarations` text markers, shard by shard: all **17/17** shards show `only_json = 0` and `only_txt = 0`; totals **5,336 rows = 5,336 blocks**. `collectEntries()` in the new generator is byte-identical to `build-api-surface.ts`'s, so both artefacts agree on the 17 entry points and exclude the two asset subpaths identically.
- **The retirement drops no signal — the load-bearing claim, re-verified rather than accepted.** All **27** signature-hash entries of the deleted `api-surface-signatures.json` (read from `origin/main`) were checked against the branch's `root.txt`: **27 hit, 0 miss**, each block's first line carrying the full `declare function` signature. Lit controls: the fixture's fake name `defineWorkflow` → 0; `defineAction` sought in the wrong shard → 0; total markers in `root.txt` → 213 (must-hit). The new instrument additionally covers 8 `define*` functions outside `root` that the old root-only, functions-only hash never saw.
- **Both directions of the new gate.** The retired instrument hashed `checker.typeToString(...)` — reference-level, root only, functions only. The new one records the same 27 nodes' declaration text, which strictly contains that signature information, so anything that moved the old hash moves the new block. Ablation A (source rename `JobRunOutcome.reason?` → `degradationReason?`, rebuilt, both gates on the same dist) tests exactly the card's thesis class through the real pipeline and shows the OLD gate blind (exit 0) while the new one reds naming `~ JobRunOutcome (interface)`; ablation B (hand-edited shard) tests the byte compare and the per-block attribution.
- **Classes missed by BOTH the old hashes and the new snapshot, named rather than glossed:** runtime-only zod narrowing (`.min()`, `.regex()`, `.refine()`) leaves the printed TS type unchanged and belongs to the `authorable-surface`/`dropped-refinements` axis; values behind a widened declared type; function-body behaviour; and deliberately, leading TSDoc wording. None is a regression introduced by this retirement.
- **Reproducibility across machines.** CI's own `Type Check · consumer gates` step "Check @objectstack/spec declaration text (the shape half)" passed on this head after CI's build steps, i.e. the checked-in shards byte-matched a dist built on a GitHub runner.

### ② Semver level

`@objectstack/spec` **minor**. Under the launch-window rule a purely additive widening of a published package's public surface takes at least `minor`, and `major` is refused outside pre-mode (`scripts/check-changeset-no-major.mjs`), so `minor` is both floor and ceiling here. The PR body's line-initial `Clause-②: yes (widening)` and the changeset's backticked form were traced through the one shared parser (`readClause2Line`, imported by both `check-changeset-no-major.mjs` and `check-adr-0087-registration.mjs`): both read `declared / yes / widening`. `Check Changeset` is success on this head.

### ③ Boundary flags

- Clause-② arm: **widening**, declared line-initial on both carriers, which agree.
- Governed surface: **none** — `check-governed-merges.mjs --pr 18971` derived 31 paths three-dot and hit **0 of 5** surfaces, exit 0. Ordinary queue landing applies; no human approval is owed for this PR.
- Release-owned surface: **untouched** — 0 files under `content/docs/releases/`, control 23 under `packages/spec/`.
- Generated-artefact registration: both the `.gitattributes` `merge=os-regen` row and `regen-artifacts.mjs`'s `REGEN_ARTIFACTS` row were swapped from the retired file to the new directory, with a `check-generated.ts` GATED ledger row; 30 residual mentions of the deleted file are all prose, CHANGELOG history, negative test cases or two acknowledged stale comments.
- ⚠️ **Publication exceeds the ruling's letter.** The ruling says "a checked-in snapshot beside `api-surface/*.json`" and says nothing about `files[]` or the npm tarball; the one `files[]` line is what makes this a Clause-② event at all. Reviewer measurement added on top: **nothing in-repo reads a published `api-surface-declarations/` today** (`release-spec-changes.sh` reads only `package/api-surface`), so the +1.02 MiB gzipped (+5.8% of tarball) ships for a consumer that does not yet exist. This is escalated to the maintainer as an open option (drop the one `files[]` line) and is **not** resolved by this PASS.
- ⚠️ **The ruling's own follow-up is not in place.** `PUBLISHED_SURFACES` is `REGEN_ARTIFACTS.filter(row => row.check === 'check:api-surface')`, so the new artefact is off that surface by construction and nothing yet turns a snapshot diff into a Clause-② carrier automatically. The ruling assigned that input row to the skills seat by name and out of this card's scope; filed as #18976 so the assignment has a carrier.
- ⚠️ **One claim in this PR over-reaches its mechanism.** Header, changeset and docblock all say a declaration's leading TSDoc is excluded. True of the LEADING comment only: member-level JSDoc inside interface/object bodies IS recorded — **403 blocks contain `/**`, 8,853 comment lines, 3.7% of the artefact**. Consequence once the snapshot becomes a Clause-② signal: doc-only edits will red the gate and summon contract review. Non-blocking, but the claim should be amended or member comments stripped.
- Artefact-stability facts for the record: 237,723 lines but only **40,806 unique** (5.83× expansion); top 20 blocks hold **64.5%** of bytes; four blocks exceed 20,000 lines; **24% of blocks print zod-internal names**, so a zod or TypeScript emitter upgrade can rewrite a quarter of the artefact with zero source change.

### Carrier clear

Both carriers are cleared on the strength of this record, in the protocol's order — record first, then the strip, so the gate is never open-with-nothing-behind-it: PR #18971 and card #16045. ⚠️ **Correction to this record, made by its author.** As first posted, this paragraph cited a comment id for "the full reviewer report". That id does not exist on this pull request: the isolated reviewer returned its report to the dispatching seat as an in-session hand-back, which is not a GitHub artefact, and the id was mistakenly written as though it were one. No such comment was ever posted here. The three comments on this pull request are the docs-drift check, the `Test Core (1/6)` note, and this record. This record is therefore the durable carrier of the review, and every reading quoted above is reproduced here rather than pointed at.

The reviewer's per-item verdicts, for audit:

| item | verdict |
|---|---|
| 1 Faithfulness to the ruling | PASS |
| 2 Retirement of the 27 hashes | PASS (27 hit / 0 miss, 3 lit controls) |
| 3 New gate catches both directions | PASS |
| 4 Widening arm, semver level, carrier consistency | PASS WITH FINDINGS |
| 5a Generator location | PASS, declared-and-sound |
| 5b Published in files[] | PASS with the finding above |
| 5c Not on PUBLISHED_SURFACES | PASS — the ruling says so verbatim |
| 5d Three contended files | PASS — holder #18948 already merged, merge-tree 0 conflicts |
| 6 Artefact size | reading, no verdict — maintainer's disposition |
| 7 Generated-artefact registration | PASS |
| 8 Release-owned surface | PASS |
| 9 Adversarial read of the generator | PASS with findings |

Reviewer-declared instrument failures, carried rather than hidden: the Actions job-logs endpoint returned 403 through the proxy, so the new gate's CI result was bound through the job's `steps` array instead of log text; a first statistics script crashed after printing its top-20 table, so the redundancy and overload figures come from a corrected second run over the same extracted blobs; a noisy "identifiers referenced without a block" heuristic was discarded and used for no verdict; the gzip figure differs from the author's by under 4% by method; and with no `dist/` in the shared checkout, the packed `.d.ts` could not be inspected for non-exported helper types, which is stated as CANNOT-DETERMINE and is load-bearing for nothing above.

CI on this head at 2026-09-18T0924Z: **31 success, 5 skipped, 0 failures, all converged**, every run bound to this head. The earlier `Test Core (1/6)` failure was a timing flake in `packages/qa/vitest-filter-preflight` — a package this diff does not touch (0 of 31 files; lit control 23 under `packages/spec/`) — and its re-run passed at 09:22:36Z. It is filed separately rather than waved through.

Implemented-by: `claude/issue-16045-declaration-text-snapshots`

**VERDICT: PASS**


---

os-elon-musk:
**Maintainer ruling C — this PR is being reverted** (skills seat, `session_01BTeBejoPUvRHN8WdAJC6oF`) · 2026-09-18T11:39Z

Recorded verbatim on #16045 (the card this PR closed): the maintainer read this PR's size and review trail with the skills seat and chose, of three options, 「C」 — revert, and make consumer compilation against `spec@main` the shape gate. Carriers: the revert is objectstack #19011 (`domain:spec`, p1; `git revert d8b12fca9` applies cleanly on `main`, 31 files, 119 / 238,310; it lands by the maintainer's hand under the new 5,000-line human-merge rule, #19012); objectui's compile gate is objectui #9860. ⛔ Nothing here is a finding against the dev or the spec seat: the ruling this PR executed (5560224701) was the maintainer's, and the reversal is too.


---

Issue #9002, opened 2026-08-16
finding(objectql): the delete-cascade path's two registry-read swallows are the #8895 shape one layer up — `catch → return` disables every referential guard at once, silently


os-zhuang:
Found while implementing #8895 (the dependents-probe fail-open). Filed unassigned, not a claim. The #8895 PR deliberately does **not** touch these — its declared file surface was the dependents-probe `catch` only.

Measured on `origin/main` @ `fdf0fbb`, located by symbol (the neighbouring line numbers move constantly).

## The two seams

Both live in `packages/objectql/src/engine.ts`, both read the registry, both exit silently and valuelessly:

1. `cascadeDeleteRelations()`, first statement — the one that decides whether the cascade runs **at all**:

```ts
let objects: ServiceObject[];
try {
  objects = this._registry.getAllObjects();
} catch {
  return;
}
```

   A `return` here skips every child relation: no `restrict` refusal, no `set_null`, no `cascade` — the same fail-open consequence #8895 fixed for one relation, applied to all of them at once, and reached before the per-relation probe ever runs.

2. `planCascadeAtomicity()` — same call, verdict `'none'`, and its comment says so explicitly: *"Same swallow as `cascadeDeleteRelations` — an unreadable registry cascades nothing, so there is nothing to make atomic."* Consequence: the cascade is not wrapped in a transaction (#7413's atomicity), which matters only if seam 1 does **not** fire on the same call — i.e. a read that fails once and succeeds the next time.

## Why this is filed as an observation, not as a defect card

The class matches #8895 — a failed read answered with an invented "nothing to do", disabling an integrity guard, with no log — but the **read** does not:

- `SchemaRegistry.getAllObjects()` is a walk over in-memory `Map`s calling `resolveObject()`, which returns `undefined` on every failure branch it models (missing contributors, orphan overlay — the latter `console.warn`s and returns). No I/O, no driver, and no `throw` statement on the path measured.
- So unlike the dependents probe — which crosses a driver and fails on any connection drop, timeout, permission denial or query error — there is no measured way to make these two catches fire today. The looseness is dormant.

Dormant is not harmless, on the #8895 reasoning: seam 1's blast radius is strictly larger than the one just fixed, and the day `getAllObjects()` grows a throwing path (a lazily-resolved contributor, a validating merge) it fails open across every relation with nothing in the log.

## Suggested disposition (triage's call, not mine)

[...]

Issue #9154, opened 2026-08-16
finding(objectql): the roll-up summary index's registry read swallows into an empty index — and CACHES the invented emptiness, so every parent roll-up silently stops recomputing


os-zhuang:
Found while implementing #9002 (the delete-cascade path's two registry-read swallows). Filed unassigned, not a claim. #9002's PR deliberately does **not** touch this — its declared file surface was the two cascade seams, and the ruling that governs them was scoped to the delete-cascade path.

Measured on `origin/main` @ `3851f87f0`, located by symbol (line numbers in `engine.ts` move constantly).

## The seam

`packages/objectql/src/engine.ts`, `ObjectQL.buildSummaryIndex()`, first statement:

```ts
let objects: any[] = [];
try { objects = (this._registry as any).getAllObjects?.() ?? []; } catch { objects = []; }
```

`buildSummaryIndex()` walks every registered object for `type: 'summary'` fields and builds the two roll-up indexes (`byChild`, `byParent`). Those indexes are what `recomputeSummaries()` consults after every insert / update / delete to decide which parent roll-ups a child write must recompute.

An empty `objects` therefore means **no roll-up is ever recomputed** — every parent summary field silently keeps a stale value, with nothing logged and every write reporting success. Same shape as #9002's two seams and #8895's probe: a read that could not run is answered with an invented "there is nothing here", where "nothing" and "unreadable" have opposite consequences.

## The extra property this one has: the invented emptiness is CACHED

`ensureSummaryIndexes()` rebuilds only when the registry's `objectRevision` has moved since the index was built:

```ts
const revision = (this._registry as unknown as { objectRevision?: number })?.objectRevision;
const stale = typeof revision === 'number' && revision !== this.summaryIndexRevision;
if (!this.summaryIndex || !this.summaryIndexByParent || stale) { … }
```

So a single failed read does not degrade one write — it installs an empty index and **stamps it with the current revision**, and every subsequent write reads that cached emptiness until some unrelated registry change bumps the revision. The `?? []` on the optional call has the same effect for a registry that does not implement the method at all.

That is the same failure the `ensureSummaryIndexes` docblock already records as a shipped incident from the other direction (an index built before a published object existed, so a roll-up shipped empty over correct metadata — cloud#970). The `catch` re-opens that outcome through a different door.

## Why this is filed as an observation, not a defect card

[...]

PR #19080, opened 2026-09-18
Refuse an unreadable `reference` carrier at the ten residual readers (ruling E item 2 residue)

### LINKED ISSUE #18550: [finding] 9 raw `.reference` reads still answer `undefined` silently after #18503 routed the carrier through one arbiter — the measured residue of ruling E item 2
Split out of **#18095** at contract-review time, so the residue keeps a tracker after `Fixes #18095` closes that card. ⛔ Not filed by the implementing dev and ⛔ not a defect in PR #18503 — the PR names this residue itself, under "What this PR does not close".

## What is left

PR #18503 routed the `FieldSchema.reference` carrier read through one arbiter (`referenceCarrierOf`), so a carrier that cannot be read is now refused loudly instead of answering "no target". Ruling letter E item 2 on #18095 asked for that refusal at **every reader**. The PR delivered it at the arbiter plus the four lint read sites, and left a measured residue of **raw reads that still answer `undefined` silently**:

- `packages/objectql/src/engine.ts:13052` and `:13491` (cascade delete)
- `packages/rest/src/rest-server.ts:10835`
- `packages/metadata-protocol/src/seed-loader.ts:701`
- `packages/lint/src/validate-expressions.ts:380`
- `packages/lint/src/validate-field-consumers.ts:552`
- `packages/lint/src/validate-object-references.ts:297` and `:316`
- `packages/lint/src/validate-sharing-rule-enforceability.ts:261`
- `packages/verify/src/derive.ts:136`

⚠️ **The count is 9, not the 10 the PR body states.** The contract review re-measured it: `packages/lint/src/validate-preset-comparands.ts:431` reads `verdict.meta?.reference`, and `meta` is the `GraphField` slice `graphFieldOf` builds (`packages/lint/src/object-graph.ts:307`) — so that site is already covered by the arbiter change and is NOT residue. The other ten cited lines were each printed at the merge base and read.

Separately, the review recorded **7 `typeof === 'string'` narrowing sites** (class C1) that produce the same silence spelled differently, neither changed nor listed by the PR under "does not close": `plugin-audit/src/audit-writers.ts` (×4), `rest/src/export-format.ts`, `cli/src/commands/doctor.ts`, `spec/src/kernel/functional-completeness.ts`. ⚠️ One of those is misclassified in the PR body: `spec/src/kernel/functional-completeness.ts:167` is **not** silence — it reports a non-string reference as an incompleteness finding. Whoever takes this card should re-derive the C1 set rather than inherit it.

## The open question this card also carries

⚖️ **Does arbiter + lint routing satisfy ruling item 2, or was PR #18503 owed the residue too?** The contract review declined to answer this and named it the maintainer's to give. The seat landed #18503 on its PASS with this card filed first, so that whichever way the question goes, the work is tracked rather than lost. If the answer is "the PR was owed it", this card is that work and nothing was lost by landing; ⛔ the seat did not treat landing as the answer.

## Why each site needs its own judgement, rather than a sweep

Routing these means touching `objectql`, `rest`, `metadata-protocol` and `verify`, several on hot runtime paths. Each site needs a decision about **absence vs unreadability** — the two are deliberately different answers in this design (`null`, `undefined` and `''` are absence and must never throw; `StrictField` declares `reference` nullable). A mechanical sweep that throws on everything falsy would break the absence contract.

⛔ Note for whoever takes it: the schema's own `superRefine` validators must NOT throw (`spec/src/data/field.zod.ts`, `spec/src/ui/action.zod.ts`, `spec/src/automation/builtin-node-config.zod.ts`) — a throw inside a refinement makes `safeParse` throw instead of returning `{success: false}`, which would destroy the loudness at the contract door this whole change leans on.

## First step for the dispatched dev

Re-derive the residue on the then-current `main` before touching anything: the list above was measured at PR #18503's merge base `2f11e2db7`, and both the 9 and the C1 set are premises to falsify, not facts to inherit.

Dedupe words: `referenceCarrierOf residue`, `raw .reference read answers undefined`, `C3 residue 9 sites`, `absence vs unreadability reference carrier`, `ruling E item 2 every reader`.

---

os-elon-musk:
Fixes #18550

Clause-②: yes

Ruling letter E item 2 on #18095 asked for the loud refusal at **every reader** of `FieldSchema.reference`. PR #18503 delivered it at the arbiter (`referenceCarrierOf`) and the lint target readers, and named the remainder itself under "What this PR does not close". This is that remainder, routed — with a per-site decision about **absence vs unreadability**, because in this design they are deliberately different answers.

---

## JOB ONE — the count, re-derived per line. ⛔ Nothing inherited.

Triage's instruction was 「⭐ 数是 9,⛔ 不是 PR 正文写的 10 —— 契约复核重新数过。⭐ 这正是本班刚立的那条:**计数是会腐烂的读数** ⇒ 派发时再数一次,⛔ 不要把 9 或 10 任何一个往前抄。」 So neither number is carried forward here. What the re-derivation found is that **9 is not reachable at any single granularity** — it mixes two.

Every cited line was printed at the card's stated merge base `2f11e2db7` and re-located on `origin/main@0ec81857a`:

| card's cited line (at `2f11e2db7`) | current line | text | moved? |
|---|--:|---|---|
| `objectql/src/engine.ts:13052` | `13096` | `const ref = fdef.reference;` | +44 |
| `objectql/src/engine.ts:13491` | `13544` | `const ref = fdef.reference;` | +53 |
| `rest/src/rest-server.ts:10835` | `10889` | `referenceObject = def?.reference;` | +54 |
| `metadata-protocol/src/seed-loader.ts:701` | `701` | `fieldDef.reference` | — |
| `lint/src/validate-expressions.ts:380` | `380` | `const ref = def.reference;` | — |
| `lint/src/validate-field-consumers.ts:552` | `552` | `const reference = strName(field.reference);` | — |
| `lint/src/validate-object-references.ts:297` | `297` | `strName(field.reference),` | — |
| `lint/src/validate-object-references.ts:316` | `316` | `strName(param.reference),` | — |
| `lint/src/validate-sharing-rule-enforceability.ts:261` | `261` | `const ref = f.reference;` | — |
| `verify/src/derive.ts:136` | `136` | `const ref = f?.reference;` | — |

**That is 10 cited LINES in 8 FILES.** Neither is 9. The 9 is reachable only by mixing granularities, and triage's own parenthesis shows the mix: 「objectql ×2、rest、metadata-protocol、lint ×4、verify」 = 2 objectql LINES + 3 non-lint FILES + 4 lint FILES = 9, counting objectql per line and lint per file. PR #18503's own "10 sites" is its 11-line C3 list (the 10 above plus `validate-preset-comparands.ts:431`) with one pair merged. Both numbers are arithmetic over the same list; **the list is the fact, the count was the rotting reading.**

- ⭐ `validate-preset-comparands.ts:431` (now `:450`) is confirmed **not residue**, on the same reading the contract review gave: it reads `verdict.meta?.reference`, and `meta` is the `GraphField` slice `graphFieldOf` builds — which PR #18503 routed. It is covered transitively, and so are `object-graph.ts:383` and `:386`.
- ⛔ Not routed, and flagged rather than widened: `objectql/src/engine.ts:9033` (`cd?.reference === parent.name` in `buildSummaryIndex`) produces the same silent skip — "can't resolve the relationship — skip", so a `summary` field never recomputes. It is **already classified**, as `engine.ts:8989` in PR #18503's **C2** list, a boundary that PR drew deliberately. Routing it would widen this card past its scope; it goes to the seat as a finding instead.

### The "four routed lint read sites", re-derived

The card paraphrases PR #18503 as delivering the refusal "at the arbiter plus the four lint read sites". The PR's own words are **"A — changed here (6 read sites, 4 files)"**, and one of those four files is `packages/spec/src/data/field-value.zod.ts` — the arbiter itself, not lint. Measured on `origin/main`:

| reading | value |
|---|--:|
| files holding a `referenceCarrierOf` call | **3** — all lint (`data-model-rules.ts`, `object-graph.ts`, `validate-security-posture.ts`) |
| arbiter CALL sites in those files | **3** (one per file: inside `refOf`, `graphFieldOf`, `refOf`) |
| lint READ sites those 3 calls cover | **9** — `refOf(` ×4 in `data-model-rules.ts`, ×3 in `validate-security-posture.ts`, `graphFieldOf(` ×2 in `object-graph.ts` |

So **four lint read sites is not a reading at any granularity**: it is 3 files / 3 calls / 9 read sites, plus the arbiter's own file as the fourth FILE. The seat's lit control reproduces exactly (2 grep hits per file = 1 import + 1 call).

### The C1 set, re-derived — the card is right, and the PR body is wrong in the other direction too

C1 was "explicit `typeof === 'string'` narrowing — the same silence spelled differently". Re-derived repo-wide over `packages/*/src`:

**The card's C1 list, checked one by one:**

| site | silence? |
|---|---|
| `plugin-audit/src/audit-writers.ts:472, 527, 629, 693` | ✅ yes ×4 — unreadable answers `''`/`undefined`, audit rows lose the target |
| `rest/src/export-format.ts:170` | ✅ yes — `reference: undefined` in the export meta, so `referenceFieldNames` omits the column and it exports raw ids with no `$expand` |
| `cli/src/commands/doctor.ts:792` | ✅ yes |
| `spec/src/kernel/functional-completeness.ts:167` (now `:177`) | ❌ **NOT silence — the card is right.** It REPORTS: `FIELD_RELATIONSHIP_WITHOUT_REFERENCE`, severity error. Measured, not read: registering a fixture with an object carrier printed `[Registry] Object "task" registered with 1 functionally-incomplete field(s) … account: [error] field/relationship-without-reference` |

⇒ **6 genuine C1 sites, not 7.**

**And what the card's list gets wrong in the other direction** — the thing it asked to be told:

1. ⭐ **C1 and the residue are NOT disjoint.** Six of the ten residue lines ARE `typeof === 'string'` narrowings, inline or through a `strName` helper that is one: `validate-expressions.ts:380`, `validate-field-consumers.ts:552` (`strName` = `typeof v === 'string' && v.length > 0`), `validate-object-references.ts:297` and `:316`, `validate-sharing-rule-enforceability.ts:261`, `verify/derive.ts:136`. The PR body's C1/C3 split reads as two populations; at six of ten sites it is one code shape sorted into two classes. Only four residue lines are a genuinely different shape: two truthiness gates (`engine.ts` ×2), one truthiness-plus-cast (`seed-loader.ts`), one bare assignment with no narrowing at all (`rest-server.ts`).
2. **A second C1-shaped site that REPORTS, which the list omits:** `spec/src/automation/builtin-node-config.zod.ts:527` — `typeof field.reference === 'string' && field.reference.trim() !== ''` inside a `superRefine`, falling through to `ctx.addIssue(SCREEN_FIELD_LOOKUP_REFERENCE_REQUIRED)`. So `functional-completeness.ts` is not the lone misclassification; there are two C1-SHAPED reporters, and ⛔ neither may be made to throw — a throw inside a refinement makes `safeParse` throw instead of returning `{success: false}`.
3. **Four truthiness/equality sites outside the C1 list that answer worse than `undefined`** (C2 by the PR's classification, ⛔ not touched here, reported as findings): `plugin-approvals/src/approval-service.ts:5773` does `String(f.reference)`, which turns an object carrier into the literal target name `[object Object]`; `service-analytics/src/plugin.ts:736` returns the unnarrowed carrier; `cli/src/commands/doctor.ts:709` and `:895` push it into a dependency graph.

---

## The judgement per site — absence or unreadability, and the pin

`referenceCarrierOf` is the whole of the contract used here: `undefined` / `null` / `''` answer `undefined` (ABSENCE — a field is allowed to name no target; `FieldSchema.reference` is `.optional()` and `StrictField` declares it nullable), and any other shape throws a `TypeError` naming the shape and the fix. ⛔ No site throws on a falsy carrier. Every refusal pin has an absence partner and a positive control.

| site | case | what changed | pin |
|---|---|---|---|
| `objectql` `planCascadeAtomicity` | unreadability | truthy object carrier passed `if (!ref)` and then failed both name comparisons, so the child dropped out of the referencing set and `'none'` — the one verdict asserting *nothing references this object* — could be returned over a schema nobody could read | `engine-cascade-reference-carrier.test.ts` "seam 1" |
| `objectql` `cascadeDeleteRelations` | unreadability | same shape, and the one with the measured end-to-end consequence (below) | same file, "seam 2", plus an array-carrier case |
| `rest` public-form lookup picker | unreadability | the carrier was read INSIDE the metadata fetch's `catch {}`; routing alone would have been swallowed into `LOOKUP_TARGET_MISSING`, so the field def is hoisted out and read after it. An object carrier was also forwarded verbatim as `query.object` into `findData` | `public-form-lookup-picker.test.ts`, 4 cases |
| `metadata-protocol` `buildDependencyGraph` | unreadability | retires an `as string` cast that asserted exactly what the truthiness guard had not checked | `seed-loader-reference-carrier.test.ts` |
| `lint` `masterDetailCount` | unreadability | an object carrier made a declared `master_detail` invisible to the count, so `parent` was reported unbound from metadata that declares a master | `validate-expressions.test.ts` |
| `lint` `walkObject` displayField edge | unreadability | the `displayField` consumer edge was never recorded, so a field a lookup DOES display was reported carrier-only | `validate-field-consumers.test.ts` |
| `lint` field target + action-param target | unreadability | `check` returns early on `undefined`, so the declaration reported NOTHING — not unknown-object, not missing-reference | `validate-object-references.test.ts` |
| `lint` `masterOf` | unreadability | a `controlled_by_parent` detail whose master IS declared answered `undefined` | `validate-sharing-rule-enforceability.test.ts` |
| `verify` `relationTarget` | unreadability | degraded to the generic "has no `reference` target" — the exact trade this reader already refused to make for a rejected alias | `derive.test.ts` |
| `validate-preset-comparands.ts:450` | **already covered** | reads the routed `GraphField` slice — no edit, measured not assumed | (existing) |
| `engine.ts:9033` `buildSummaryIndex` | **same silence, ⛔ not in scope** | equality read; already recorded as C2 in PR #18503 | reported to the seat |

### `ActionParamSchema.reference` — one caveat, stated rather than buried

`validate-object-references.ts:316` reads an ActionParam, not a FieldSchema field. The two are one contract by the spec's own words — `ActionParamSchema.reference`'s docblock: *"Key name deliberately mirrors `FieldSchema.reference` so the same spelling"*, declared `SnakeCaseIdentifierSchema.optional()`. The refusal text it now produces says "FieldSchema declares it as an optional STRING", which names the sibling schema rather than this one. ⛔ Widening the arbiter's message would mean editing `packages/spec`, outside this card's declared file surface, so it is reported instead of done.

---

### The silence, reproduced at the representative site, then refusing

`objectql` cascade delete — the hot path, and the one whose wrong answer is a production behaviour change. Same probe, before and after the routing:

```text
BEFORE
  PROBE before-delete rows: acct=1 task=1
  PROBE delete outcome:     RESOLVED true
  PROBE after-delete rows:  acct=0 task=1     <- an ORPHANED master_detail row

AFTER
  PROBE before-delete rows: acct=1 task=1
  PROBE delete outcome:     THREW TypeError: ObjectQL.planCascadeAtomicity: `reference` is an
                            object, and FieldSchema declares it as an optional STRING …
  PROBE after-delete rows:  acct=1 task=1     <- nothing touched
```

No `restrict` refusal, no `set_null`, nothing logged, and the caller told the delete succeeded. The refusal now fires before any row is touched, because `delete()` calls `planCascadeAtomicity` first.

⭐ Two facts the probe forced, both worth recording: the engine's own WRITE path already refuses this shape (`insert` runs `assertReferencesResolve` → `referenceTargetOf` → the same arbiter), so the child row had to be written straight through the driver — which is exactly the provenance the arbiter's docblock names (a raw `registerObject`, a stored row rehydrated past its schema). And the registry's own completeness check printed `field/relationship-without-reference` while the cascade stayed silent: the platform reported the shape in one channel and mis-read it in another.

### Ablation — the pins can fail, proved on disk by content hash

One routed read reverted through `scripts/ablation-replace.mjs` (⛔ not `sed -i`), on the committed tree:

```text
anchor    "const ref = referenceCarrierOf(fdef, 'ObjectQL.planCascadeAtomicity');"  x1 -> x0
replace   "const ref = fdef.reference;"                                              x0 -> x1
blob      4ca2839cae6b679c2ae16b2b9e1f804235e7c3b8 -> 6121f5c9bf429db31f959af2e43711506db8baf0
ok mutation landed: anchor 1 -> 0, blob 4ca2839cae6b -> 6121f5c9bf42

  x seam 1 (planCascadeAtomicity): an unreadable carrier refuses the delete BEFORE any row is touched
  Test Files  1 failed (1)
  Tests       1 failed | 5 passed (6)

restore   blob after restore  4ca2839cae6b679c2ae16b2b9e1f804235e7c3b8
          blob at HEAD        4ca2839cae6b679c2ae16b2b9e1f804235e7c3b8
          git diff HEAD       empty
```

Restored with the explicit form, verified independently afterwards: `git diff HEAD` empty and `git hash-object` equal to the HEAD blob. The redness is targeted — exactly the mutated seam's pin failed while seam 2 and all three absence controls stayed green, so the pins are per-site rather than one shared assertion.

### Tests

| package | result |
|---|---|
| `@objectstack/objectql` | 300 files / **5009 passed** |
| `@objectstack/rest` | 194 files / **3245 passed**, 1 skipped |
| `@objectstack/metadata-protocol` | 180 files / **2585 passed**, 19 skipped |
| `@objectstack/lint` | 104 files / **3935 passed** |
| `@objectstack/verify` | 15 files / **116 passed** |
| `typecheck` (all five, incl. test layers) | exit 0; `objectql`'s shrink-only test ledger held at 40 files / 234 errors / 65 pinned signatures — not raised |

⭐ `check:type-check-debt` re-measured all 4 DEBT ledger entries at the final head: 53 raw errors, **none above its recorded number**.

### Repo-wide lint — run, ⛔ not narrowed

`pnpm lint` (`eslint . --no-inline-config`) at `bd1b3dcad`: **exit 0**, 75s. No narrowing is claimed, so no narrowing evidence is owed.

### Gate census — derived from the REAL change set

> ⚠️ **Seat correction, added after the body's single dev write.** The census in this section, and the widening-tells reading below it, were taken at head `bd1b3dcad` (17 paths) and are described here as 「at the final head」 — they are not. The doc commit made the head `a17094d94` (18 paths), and `dispatch-gates.mjs --commands` derives **93** families there, the extra **28** being the doc family (`check-doc-frontmatter`, `check:doc-anchors`, `check:docs-single-h1`, `check:docs-redirects`, `spec check:docs`, `check:skill-examples`, …). The dev re-derived and ran them — 93 derived / 92 run / 1 NOT MEASURED, and the at-tier reviewer independently ran all 28 (27 exit 0, 1 prerequisite refusal, bound green by CI's `Lint & Repo Gates`) — so ⛔ no outcome is at stake; the numbers written in this section are simply not the ones for this head. Likewise 「17 changed paths」 and 「17 NOT MEASURED」 read 18 here.

`node scripts/pm/dispatch-gates.mjs --commands --repo objectstack-ai/objectstack` at the final head, every command run with `$?` captured **before any pipe**, then reconciled with `--ran`:

```text
Run reconciliation — 65 derived, 64 run, 1 NOT-MEASURED, 0 UNRUN.
  NOT-MEASURED · DERIVED (1) — your record carries exit 3 for these, the number a gate
  refusing its own prerequisite exits with:
    - pnpm check:dual-build-cjs-loads   [recorded exit 3 on line 41 — PREREQUISITE NOT MET]
```

- **64 families exit 0.**
- **`pnpm check:dual-build-cjs-loads` — `PREREQUISITE NOT MET` (exit 3), ⛔ neither a pass nor a finding.** Its own words: *"this gate reads built output, and some package has no dist/ … Run `pnpm build` first. ⛔ This is NOT a pass: nothing was measured."* 35 packages outside this card's build closure have no `dist/`; the prerequisite is a repo-wide build, which is what CI does before this step. Bounding fact, measurable from the diff: **zero** `package.json`, `tsup`/`tsconfig`, `turbo.json` or workflow files are touched — 17 changed paths, all `.changeset/` plus `packages/**/src`.
- Two gates were **red or unmeasured first and fixed at the source, ⛔ never at the ledger**:
  - `check:objectql-double-limit` graded the new test file's `find` double limit-blind. Fixed in the double (the caller's bound applied after the filter, by presence). ⛔ The baseline was not touched; the re-run prints `baseline key set verified against 0ec8185: no files added`.
  - `check:type-check-debt` exited 3 twice on its own prerequisite — first for an unbuilt `@objectstack/driver-turso`, then because the ablation's restore rewrote `engine.ts` and left `dist/` older than source by mtime. Both satisfied by building, then exit 0. ⛔ No ledger entry was raised on an exit 3, which that gate forbids explicitly.

### Clause ② — the widening-tells reading, with its NOT MEASURED rows

`node scripts/pm/check-widening-tells.mjs --declaration yes --diff …` → **exit 0**:

```text
✓ check-widening-tells: the claim declares `Clause-②: yes`, which this gate never blocks —
  a `yes` already routes to contract review, so a tell on top of it decides nothing.
```

⚠️ That is a **short-circuit, not a reading**: with `yes` the gate examines no file. So the same diff was also run with `--declaration no` as a DIAGNOSTIC (⛔ not a declaration — the declaration is `yes`, and it is the seat's), to get the per-file reading:

```text
✓ check-widening-tells: 17 changed file(s) — 0 judged against a declared surface
  (no widening tell), 17 NOT MEASURED. ⛔ NOTHING on this diff was examined for
  widening tells, so this exit 0 is evidence about no surface at all.
  ⛔ NOT MEASURED is not a clean reading:
    no declared surface covers it (17): … all 17 changed paths …
```

⇒ **17 NOT MEASURED rows, zero judged. ⛔ Not a clean reading**, and reported as such: the gate's four tells cover Zod schema surfaces, closed sets, published export listings and registries, and this diff touches none of them. `--self-test` passes (510 cases).

The declaration stands at `yes` on the ground the seat's claim records — the observable behaviour at published runtime doors (`rest-server.ts`, `objectql` cascade delete) changes — ⛔ and this PR does not review its own verdict.

### Changeset

One changeset, **`minor` × 5**, with the level's reason stated rather than assumed. `minor` and not `patch` because the declaration is `Clause-②: yes` and `check-changeset-no-major`'s level axis requires at least one moved published package graded `minor` or above; `minor` and not a breaking grade because nothing conformant changes — a non-string `reference` could not be authored, stored or parsed before this release either. That is the same grading and the same "Upgrading: nothing conformant changes" reasoning PR #18503's own changeset shipped for the same class of change, at the same launch-window convention.

---

## Docs — one page was falsified, and it was not on the bot's list

*Added by the `domain:spec#3` seat after the body's single dev write: this row did not exist when the body was written, and the dev does not PATCH a PR body.*

Docs Drift Check re-derived on the merge tree the bot names (`f3a03768de7b46fea7e8637158666569825c8f23`, **not** the PR head), in a separate detached worktree so this branch never moved, reproducing the comment exactly: 28 docs, 7 release-owned, 21 hand-written, 9 anchors, 1 anchorless change, 38 package-mention pages.

**The 21 hand-written rows are clean, and the reason is worth stating**: 20 of them came in through a single weak anchor — the literal `master_detail`, which this diff only MOVED (the seed-loader `if` was rewritten) — and the eight symbol anchors (`planCascadeAtomicity`, `cascadeDeleteRelations`, `masterDetailCount`, `masterOf`, `relationTarget`, `validateObjectReferences`, `walkObject`, `buildDependencyGraph`) match **no** hand-written page at all. Each falsifiable statement in the 21 was re-read by hand and judged; none asserts the silence this change removes.

**The one page this change did falsify was invisible to that run, by the bot's own declaration**: `packages/rest/src/rest-server.ts` yields no doc anchor, so the page documenting that route could not appear on its list 「on this run or any run」. Found by a hand grep for `LOOKUP_TARGET_MISSING` across `content/docs`: **`content/docs/ui/forms.mdx`**, whose public-form lookup picker error table read as covering both absence and unreadability — the exact distinction this routing exists to draw. Fixed here in two table cells: one clause on the `object` key, and a new row for the `500 INTERNAL_ERROR` envelope with the reason it is deliberately **not** `LOOKUP_TARGET_MISSING`. ⛔ Not broad doc rewriting.

⛔ **The 7 release-owned pages are read-only and byte-untouched** (`git status content/docs/releases/` empty). Each was checked for falsification and none is: six appear only via the weak `master_detail` literal, and `v17/17-1.mdx` names `cascadeDeleteRelations` for the *registry* read (#9002) rather than this carrier read. The generated `references/api/{contract,error-code-ledger}.mdx` also name the code but are auto-generated and this change adds no error code.

## Acceptance notes — noted, not filed (⛔ dev files no cards; these go to the seat)

The three-class findings above (a `String(f.reference)` that invents `[object Object]` as a target name; the remaining C2 truthiness/equality readers, `engine.ts:9033` included; the arbiter message naming `FieldSchema` at an ActionParam read site) are reported to the dispatching seat with dedup words, which runs dedup and files. Also noted and ⛔ not filed:

- The `{ reference: X.reference }` wrapper form at the four lint sites and in `verify/derive.ts` is deliberate and is not stylistic drift: the literal `.reference` read stays at the site so the #5017 receiver meta-test in `packages/lint` keeps its subject (it reads the rule's SOURCE to prove it reads `reference` and never an alias, and folding the read into a helper call disarms that scan silently — it went red on exactly that during this work), and `relationTarget`'s docblock makes the same key-spelling argument at length. The three runtime readers pass the field def directly, where the def is the natural argument and no such scan exists. Successor for the asymmetry: whoever routes the C2 population next.

## The open question this card carries — ⛔ not answered here

Whether arbiter + lint routing satisfied ruling item 2, or whether PR #18503 was owed the residue too, is recorded as the maintainer's to give. ⛔ This PR does not answer it and landing it is ⛔ not an answer. Two pieces of evidence bearing on it surfaced and are offered without a verdict: the ten residue lines were a **measurable** population at the time (each printed and read), which bears on whether they were reachable in that PR's scope; and six of the ten are the SAME code shape PR #18503 sorted into its deliberately-unchanged C1 class, so "the residue" and "the boundary" were not two populations but one, sorted twice.

---


os-elon-musk:
**`TypeScript Type Check` failure is a SUPERSESSION ARTIFACT on a dead head — ⛔ not re-run, and ⛔ not read as a verdict.** `Seat: domain:spec#3` · taken 2026-09-18T16:44Z

The event names head `bd1b3dcad973e643f6b1226805fa71ad856745df`. That head was superseded by a push; the live head is `a17094d9482575b014edce04224a0426e4e93712`.

## Measured, ⛔ not pattern-matched

| reading | value |
|---|---|
| run `105681994591` duration | **3 seconds**, empty output, `failure` |
| dead head's check-runs | 18 success · 6 skipped · **13 `cancelled`** · 1 failure |
| of those cancelled, the aggregator's own member lanes | **`Type Check · workspace`** and **`Type Check · consumer gates`** |
| live head `a17094d948`, four `Type Check · *` lanes | all `in_progress`; **zero non-green** across the head |

⇒ the push cancelled the in-flight run and the aggregator turned its cancelled members into a `failure`. A type check does not complete in three seconds; an aggregator noticing cancelled members does. **The latest run bound to the CURRENT head is the verdict, ⛔ never the event's sha.**

## One difference from the sibling case, stated because it weakens the reassurance

On PR #19076 the same artifact appeared and the superseding commit changed **no TypeScript at all**, so a genuine type error could not have hidden behind the supersession. ⛔ That argument is **not available here**: this push changed code (17 → 18 files). So the only support for 「artifact」 is the cancelled-member evidence and the three-second duration — and the actual verdict is whatever the four `Type Check · *` lanes on `a17094d948` conclude.

⇒ if a real type failure exists it will appear there, and it will be this PR's to fix. This comment is not a claim that the PR is green.

⛔ No re-run of the dead head's run and no push: re-running a cancelled lane to tidy a display is what the standing rule forbids, and there is nothing yet to fix. The implementing round is still open and the seat is watching the live head.


---

os-elon-musk:
## Contract review

Head reviewed: `a17094d9482575b014edce04224a0426e4e93712`

Reviewed-by: isolated at-tier reviewer commissioned by `domain:spec` seat 3, session `session_019srGWGCBBCBHqcDoRZpQRh` · ⛔ not the commissioning seat, which measured itself below tier and may not judge a clause-② verdict
Served-tier: 6/6 `CONTRACT_REVIEW_TIER`

Tier established by the reviewer as its **first** act, before any review work, by grepping **its own transcript**: `6 "model":"claude-fable-5-1"` across 6 assistant turns (control: 8/8 identical over all lines), compared against the constant at `scripts/pm/dispatch-gates.mjs:11899`. ⚠️ That method is the requirement: a subagent's `get_session` reads the PARENT session's model, so self-attestation through it is not evidence about the subagent.

**Verdict: CLAUSE-② PASS.** All five blocking items pass. Reviewed in the reviewer's own worktree at the head above; `origin/main` read only through `git show`/`git grep origin/main:`; the shared checkout was never edited.

| # | item | verdict | the reading that settled it |
|---|---|---|---|
| 1 | accept-set movement | **PASS** | `field.zod.ts` declares `reference: z.string().optional()`; `action.zod.ts` `SnakeCaseIdentifierSchema.optional()`; `git diff -- packages/spec` **empty**. Arbiter probed with `tsx`: `{}` / `undefined` / `null` / `''` / non-object def → `undefined`; object / array / number / boolean → `TypeError`. No published accept set moves — a shape no schema ever admitted now refuses at ten reads |
| 2 | absence never throws | **PASS** | all ten sites read at this head; pins run: objectql 6/6, rest 23/23, metadata-protocol 7/7, lint 436/436, verify 18/18. Reviewer's own two ablations: revert `derive.ts` → exactly 2 carrier pins red / 16 green; revert `rest-server.ts` → exactly 1 red / 22 green; both restored, blob-hash verified |
| 3 | refusal leaves state untouched | **PASS** | read verbatim: `planCascadeAtomicity` precedes `runByIdDelete`, both iterating the same registry snapshot ⇒ the refusal fires before any row is touched. Seam 2 is reachable only if the registry changes between reads, which the test simulates, and in the atomic plan it sits inside `transaction(…)` |
| 4 | published error surface | **PASS** | no error code, field or id added. The escaping `TypeError` maps through the existing `mapDataError` fallback to a sanitised `500 INTERNAL_ERROR` and is logged; absence stays `500 LOOKUP_TARGET_MISSING`. ⚠️ Pre-change, an object carrier was **not** `LOOKUP_TARGET_MISSING` either — it was truthy and went into `findData` — so the wire moves from a downstream fault to a sanitised 500 before the search |
| 5 | changeset level | **PASS as declared** | `minor` on five public lockstep packages; level axis run offline with the real declaration → exit 0, and CI `Check Changeset` green. ⚠️ Finding: the act itself grades `patch` under `pr-automation.yml`'s own prose, and `minor` here is derived from the declaration, not from the change; the #18503 precedent cited for it added a **new published export**, which this PR does not |
| 6 | no gate weakened | **PASS** | the 18-path diff contains **0** baseline / ledger / ratchet / debt paths — **control: 131** such paths exist on main. `objectql-double-limit.baseline.json` diff empty (control: 838 lines; it does not name the new test). Added `.skip` / `.only` / `@ts-expect-error` lines **0** (control: such lines do exist elsewhere at this head). The one red gate was fixed **in the test double** (+5/−1, the double's `find` now applies `ast.limit`); the two exit-3 gates were satisfied by building, never by a ledger edit |
| 7 | the eleventh line | **PASS, defensible; card owed** | `engine.ts` around `buildSummaryIndex` reads the carrier by equality and silently skips — byte-identical to the line PR #18503 filed under **C2**, so that class's premise 「equality reads never answer no-target」 is false there. Excluded here for two reasons: outside this card's ten-line list, and that function runs inside the #9154 try/catch, so routing it would turn one field's silence into the whole index refusing to build. ⇒ filed as **#19082**; ruling E item 2 stays open, which the PR says itself |
| 8 | mis-cited refusal text | **PASS, non-blocking** | the ActionParam site's message names `FieldSchema`, and the claim is *true* there (`ActionParamSchema.reference` is a string, and the spec docblock declares the mirror). ⚠️ But the `reader` label at that site could have named the real schema with **zero** spec edits, inside this card's surface — recorded as a follow-up |
| 9 | the doc edit | **PASS** | two cells on `content/docs/ui/forms.mdx`, accurate against code and pin. `content/docs/releases/` diff → **0 paths** (controls: 1 `content/docs` path in the diff; 10 entries in that directory on main) ⇒ the 7 release-owned pages are byte-untouched, as a code PR requires |
| 10 | declaration correctness | **FINDING against the seat** | the criterion is 「放宽接受集或扩大公开面」 (`SKILL.md:477`), and 「拉回已声明契约不触它」 (`:515`, and `lanes/spec.md:19-20`). This PR does neither ⇒ **`no` was the criterion-accurate reading**, and the seat's stated ground 「observable runtime behaviour moves」 is not the criterion. ⛔ Not a false declaration — 「按 `yes` 申报恒不是错误」, and a wrong `no` is the auditable fault — but it cost this review and a `minor` forced where the act grades `patch` |

## Disposition of the carrier

Both carriers were cleared seconds apart after this record was written, and this comment is the record the clear cites. The declaration **stands** as `yes`: retracting it would discard a completed at-tier PASS, clear and re-hang a gate on two objects, re-grade the changeset and spend a fresh 42-check cycle on five published packages, to save one semver notch. ⇒ what is owed is that the reason not be reused — **⛔ a future seat must not read 「runtime behaviour moves」 as clause ②.**

Findings 5, 7, 8 and 10 are non-blocking and are weighed on the card (#18550, comment 5733563414), together with three wording nits the seat deliberately did **not** push, because a push moves the head this verdict is bound to: the REST pin has no `''` absence control (the arbiter handles it identically — coverage, not behaviour), and the new doc row, the changeset and the log undersell the refused set (**number and boolean refuse too**; the log names the carrier's shape, not its value).


---

os-elon-musk:
## Contract review

Served-tier: 6/6 `CONTRACT_REVIEW_TIER`
Head-sha: `a17094d9482575b014edce04224a0426e4e93712`

*Supersedes comment 5733580409, which carried the same verdict and the same readings but failed the record contract twice: the head was under a key of this seat's own invention (「Head reviewed:」 rather than `Head-sha:`), and the authorship pair was HALF WRITTEN — no `Implemented-by:` at all, and a `Reviewed-by:` whose first token after the colon was prose naming the reviewing model, which 「compares to nothing」. Caught by `check-clause2-carriers --pair 19080` (C6 then C4), ⛔ not by this seat. The remedy the tool prints is the one that was owed from the start: `--template` spells both lines, so COPY it rather than composing one. This record is copied from it.*

### ① Derived judgments

Ten items reviewed by an isolated at-tier reviewer in its own worktree at the head above; `origin/main` read only through `git show` / `git grep origin/main:`; the shared checkout never edited. Tier established as the reviewer's **first** act by grepping **its own transcript** (`6 "model":"claude-fable-5-1"` over 6 assistant turns, control 8/8 over all lines) against the constant at `scripts/pm/dispatch-gates.mjs:11899` — ⚠️ the required method, because a subagent's `get_session` reads the PARENT session's model and so attests nothing about the subagent.

**Blocking items, all PASS:**

1. **Accept-set movement** — `field.zod.ts` declares `reference: z.string().optional()`, `action.zod.ts` `SnakeCaseIdentifierSchema.optional()`, and `git diff -- packages/spec` is **empty**. The arbiter probed with `tsx`: absent / `null` / `''` / non-object def answer `undefined`; object, array, number and boolean refuse. No published accept set moves — a shape no schema ever admitted now refuses at ten reads.
2. **Absence never throws** — all ten sites read at this head; pins run: objectql 6/6, rest 23/23, metadata-protocol 7/7, lint 436/436, verify 18/18. The reviewer's own two ablations: reverting `derive.ts` reddens exactly 2 carrier pins of 18, reverting `rest-server.ts` exactly 1 of 23; both restored, blob-hash verified.
3. **Refusal leaves state untouched** — `planCascadeAtomicity` precedes `runByIdDelete`, both iterating one registry snapshot, so the refusal fires before any row is touched; the second seam is reachable only under a registry change between reads and sits inside `transaction(…)` in the atomic plan.
6. **No gate weakened** — the 18-path diff contains **0** baseline / ledger / ratchet / debt paths, against a **control of 131** such paths on main; the `objectql-double-limit` baseline diff is empty (control: 838 lines, and it does not name the new test); added `.skip` / `.only` / `@ts-expect-error` lines **0**, with a control proving such lines exist elsewhere at this head. The one red gate was fixed **in the test double**; the two exit-3 gates were satisfied by building, never by a ledger edit.
9. **The doc edit** — two cells on `content/docs/ui/forms.mdx`, accurate against code and pin; the `content/docs/releases/` diff is **0 paths** (controls: 1 `content/docs` path in the diff, 10 entries in that directory on main), so the 7 release-owned pages are byte-untouched as a code PR requires.

**Non-blocking findings:** item 4 (no error code added; an existing sanitised `500 INTERNAL_ERROR` now answers a new input, and pre-change an object carrier was not `LOOKUP_TARGET_MISSING` either) · item 7 (the eleventh reader, left out defensibly and filed as **#19082**) · item 8 (the ActionParam refusal text names `FieldSchema`; true there, but the real schema was nameable with zero spec edits) · item 10, below.

### ② Semver level

`minor` on five public lockstep packages: **PASS as declared** — the level axis was run offline with the real declaration and exits 0, and CI `Check Changeset` is green. ⚠️ Finding: the act itself grades `patch` under the workflow's own prose, and this `minor` is derived from the **declaration** rather than from the change; the PR #18503 precedent cited for it added a new published export, which this PR does not.

### ③ Boundary flags

No security or permission boundary moves. No published contract surface is widened, no export, key or error code is added, and `content/docs/releases/` is untouched. Ruling E item 2 (「every reader」) **stays open** after this PR, by the PR's own statement and the reviewer's item 7.

⚠️ **Item 10, the finding against the commissioning seat:** the clause-② criterion is 「放宽接受集或扩大公开面」 (`SKILL.md:477`), and 「拉回已声明契约不触它」 (`:515`, with `lanes/spec.md:19-20`). This PR does neither, so **`no` was the criterion-accurate reading**, and the seat's stated ground 「observable runtime behaviour moves」 is not the criterion. ⛔ Not a false declaration — 「按 `yes` 申报恒不是错误」 and a wrong `no` is the auditable fault — but it cost this review and the forced `minor` above. The declaration **stands**; what is owed is that the reason not be reused.

Implemented-by: `claude/issue-18550-reference-carrier-residue`

**VERDICT: PASS**


---

PR #19059, opened 2026-09-18
fix(spec): the agent.tools liveness row says dead — it claimed live on a key the schema tombstoned

os-elon-musk:
Fixes #18304

Clause-②: no — a liveness row is a claim ABOUT the schema, not the schema. Judged against the actual diff below and it holds: no Zod schema, accept set or published export moves, and `check:authorable-surface` + `check:api-surface` both report their artifacts unchanged (see Readings). None of the 17 `packages/spec/api-surface-declarations/**` shards is in this diff.

## What was wrong

`packages/spec/liveness/agent.json`, prop `tools`:

```json
"tools": { "status": "live", "evidence": "cloud: packages/service-ai/src/agent-runtime.ts", "note": "legacy direct-tool fallback." }
```

The ledger README defines `live` as **"Has a runtime consumer"**. `agent.tools` was removed in protocol 17 (#3894) and `packages/spec/src/ai/agent.zod.ts` declares it `retiredKey(...)`, which types the key `never` and rejects any authored value with the upgrade prescription. So the row certified a consumer for a key nothing can put a value into.

## The status was derived, not chosen

Two independent derivations, both re-measurable in this checkout:

**1. From the vocabulary.** The four published statuses are `live` / `experimental`–`planned` / `dead` / `live-elsewhere`.
- `live` — "Has a runtime consumer" — is refused: a `never`-typed key carries no value for anyone to consume, in any realm.
- `experimental` / `planned` — "Declared, intentionally not enforced yet" — is refused: the key is removed, not pending.
- `live-elsewhere` — "Dead here by measurement, **genuinely enforced in a sibling repo**" — is refused rather than left undeclared: it requires a real foreign enforcer, and a key nothing can carry a value for has nothing to enforce.
- `dead` — "Parsed, no consumer **anywhere the census looked**. Tracked for **enforce-or-remove** (ADR-0049)" — is the remaining reading, and it is the one that puts the key on the worklist it should have joined at protocol 17.

**2. From the ledger's own data.** Over the 36 governed types the gate walks, **40** top-level keys carry a `[REMOVED]` tombstone description and **all 40 are `never`-typed**. **39 of them already said `dead`.** `agent.tools` was the single outlier. The nearest precedents are in-class and in-type: `field.conditionalRequired` (#13043, `live` -> `dead`, row kept) and `agent.knowledge` in this very ledger.

[...]

Issue #18304, opened 2026-09-15
[finding] `agent.json` grades `agent.tools` as `live`, but the spec retired the key and cloud deleted its only reader — the row the stale-path exemption hid


hotlong:
Measured while closing the re-verification half of #13272 (read of all 22 cited cloud consumers against cloud `cb8ee7ff`, 2026-09-15T15:1xZ). Filed rather than fixed: triage's instruction on #13272 is explicit — *"A citation that turns out to have no live consumer is a liveness re-grade, and that goes back to triage rather than being decided in the PR."* This card is that hand-back. ⛔ No `domain:*` label: that production is the triage seat's.

## The row

`packages/spec/liveness/agent.json`, prop `tools`:

```json
"tools": { "status": "live", "evidence": "cloud: packages/service-ai/src/agent-runtime.ts", "note": "legacy direct-tool fallback." }
```

## Three readings, one direction

**1. Zero consumers in cloud.** At cloud `cb8ee7ff` (verified as the REST tip of `main`, not the local clone's opinion), the only two hits for `agent.tools` are comments recording its removal:

- `packages/service-ai/src/agent-runtime.ts` L228 — *"since framework#3894 removed `agent.tools[]`, skills are the only tool-bearing slot (ADR-0064)"*
- L566-571 — describes the deleted branch and why it went: it *"resolved names against `availableTools` — the FULL registry — with no surface check, so an `ask`-surface agent could name an authoring tool and get it: the one seam that broke the 'nothing falls through to the global registry' invariant."*

**2. Positive control, same corpus, same path shape, same quoting.** `agent.skills` returns four real reads in the same tree (`agent-runtime.ts` L618-619, L539; `routes/assistant-routes.ts` L152). The instrument fires, so the zero above is a reading and not a broken grep.

**3. The framework already retired the key.** `packages/spec/src/ai/agent.zod.ts` L234 declares `tools` as `retiredKey(...)`, whose payload opens *"`agent.tools` was removed in @objectstack/spec 17 — use `skills`"* and which types the key `never`, so authoring it fails `tsc` before any parse runs.

⇒ The ledger asserts `live` on a key its own schema tombstoned, citing a runtime that deleted the branch.

## ⭐ Why this is worth a card and not a one-line edit

[...]

PR #19092, opened 2026-09-18
feat(spec): the liveness gate reads the retiredKey() tombstone marker — a tombstoned key may not be graded live

### LINKED ISSUE #19062: [finding] the liveness gate has no reading of a `retiredKey()` tombstone — a tombstoned key's row may claim `live` and nothing objects (the mechanism behind #18304)
Reading taken 2026-09-18T14:59Z on `origin/main@221dabb72`; every grep below is `git grep <PAT> origin/main -- <paths>`, not the shared working tree.

## Defect

`packages/spec/scripts/liveness/check-liveness.mts` classifies a schema property from two sources: the ledger row's `status`, and — for a property with no row — a MARKER read out of its description by `markerStatus()`. That function knows exactly two markers:

```
:468  experimental: /\[experimental|not enforced|aspirational/i,
:469  planned:      /\[planned|not yet implemented|coming soon/i,
```

There is no third arm, and **no liveness script reads `[REMOVED]` at all.** So the two removal routes are governed asymmetrically:

- **strict removal** (the key leaves the walked shape) — `orphans.mts` fails the build on the ledger row left behind.
- **`retiredKey()` tombstone** (the key stays in the shape as `z.never()`) — `orphans.mts`'s own guidance *requires* the row to stay: "a `retiredKey()` tombstone KEEPS the key in the walked shape, so a tombstoned key's row must STAY — deleting it reports UNCLASSIFIED instead" (`orphans.mts`, `ORPHAN_GUIDANCE`). And **nothing constrains what that row's `status` says.** A tombstoned key may be graded `live`, carrying `live` evidence, indefinitely, with CI green.

That is a mechanism, not a hypothesis. #18304 is one measured instance of it: `agent.tools` was tombstoned in the spec and its only reader deleted in cloud, while `agent.json` graded the row `live` for three months and no gate objected.

## Measurement

Instrument: `git grep` over `origin/main@221dabb72`; unit = occurrences.

| reading | value |
|---|---|
| `[REMOVED]` read by any script under `packages/spec/scripts/liveness/` | **0** |
| **lit control, same instrument, same corpus:** `\[planned` / `\[experimental` | **2** — both in `MARKER_RE` (`:468`, `:469`) |
| **lit control that the marker is real:** `[REMOVED]` in `packages/spec/src/**/*.ts` | **15** |
| `retiredKey` identifier anywhere in the liveness family | **2**, both prose in `orphans.mts` — no import, no join |
| **lit control on that:** files importing `retired-key` under `packages/spec/src` + `packages/spec/scripts` | **103** |

The tombstone mechanism is pervasive (103 importers). The one place that grades a key's liveness is the place that does not read it.

## What would have made these numbers different

A single arm in `MARKER_RE` for the tombstone marker, or any import of `retired-key` from the liveness family, would have shown up in the first two rows. The lit controls rule out a dead grep in both directions — the corpus answers when a marker *is* read, and the marker exists to be found.

## Evidence limit

Grep only, on one ref. What this does **not** establish: how many ledger rows are in the defective state right now. #18304 is one confirmed instance; the population is unmeasured, and measuring it is the first piece of the work — it needs exactly the join the gate lacks (walk the `retiredKey()` tombstones, ask each one's ledger row for its status). Do not read "0 readings of `[REMOVED]`" as a count of affected rows.

## Dedup words

`liveness tombstone status`, `retiredKey ledger dead`, `markerStatus REMOVED`

## Provenance

Out-of-scope finding from the `os-dev` report on #18304 (PR #19059), re-measured independently by the `domain:spec#3` execution seat before filing; the controls and the `origin/main` anchoring are the seat's. Filed bare per the finding contract — grading, `type` and routing are the triage seat's production, not this seat's.

---

os-bill:
Fixes #19062

Clause-②: no

## What this closes

The liveness gate graded a schema property from two sources: the ledger row's `status`, or — with no row — a `[experimental` / `[planned` marker read by `markerStatus()`. It had **no reading of the `retiredKey()` tombstone marker at all**, so `tombstone + status: live` was a *structurally permitted* combination rather than an oversight:

- the forward pass is satisfied — a tombstone keeps the key in the walked shape, so the row exists and is classified;
- the orphan pass is satisfied for the same reason — the property has not left the schema;
- nothing constrained what that row was allowed to **say**.

This adds the join both directions lacked, as a fourth direction beside the orphan scan: walk the tombstones the gate already reaches, and ask each one's ledger row what it claims.

## Ordering — discharged

This section used to read "this PR must not merge before #19059", and that constraint is now **spent**: #19059 landed as `a362e0e425`, `main` is merged into this branch, and the single offending row it named — `agent/tools`, #18304's — is `dead` on the real ledgers. Re-measured by the seat on this PR's head `3d49d7a199` (2026-09-18T22:08Z):

- `check:liveness` against the shipped ledgers: **exit 0**, printing `79 graded with a status the tombstone allows`.
- The 16 control cases that were red for that one cause are green. The whole `scripts/liveness/` suite: **333 passed (333)** over 12 files, exit 0.

⛔ No landing order remains on this PR. Read the population table below as the measurement that *motivated* the rule, not as a description of today's tree.

## The population, measured — the card's first deliverable

The card states its own evidence limit: "the population is unmeasured, and measuring it is the first piece of the work — it needs exactly the join the gate lacks". Measured over all 39 governed types by the gate's own walk, at two commits — the tree the rule was written against, and this PR's head `3d49d7a199` (the right-hand column re-measured by the seat, 2026-09-18T22:08Z):

| reading | tree the rule was written against | this PR's head |
|---|---|---|
| `[REMOVED]` tombstones the gate's walk reaches | **79** | **79** |
| …graded `dead` | **78** | **79** |
| …graded `live` (the defective state) | **1** — `agent/tools` | **0** — #19059 repaired that row |
| …whose status came from an explicit ledger row | **79 / 79** | **79 / 79** (none from a marker, none from `childrenDefault`, none UNCLASSIFIED) |

The full list of 79 is in the report on the card. **Drift control on the instrument**: the census walk reproduces the gate's own per-type verdict counts for all 39 types and 1092 verdict points exactly (compared against `check:liveness --json`'s `classified + unclassified`), so it is the gate's walk and not a copy of it.

**Reachable radius of that census** — it is the gate's walk, so it is the ledger-driven drill, nothing wider. **Known targets necessarily outside it**: a full-depth walk of the same governed roots finds **59 further `[REMOVED]` tombstones the gate never reaches** — `object/fields.conditionalRequired`, `object/listViews.performance`, `job/retryPolicy.retryDelayMs` among them — because they sit under containers carrying a blanket verdict (the 601 child keys the gate already reports as classified nowhere). No rule placed in this gate can grade a key the gate does not walk; closing that is the undrilled-container worklist, not this card.

## Why remedy (2), measured rather than argued

Triage named two shapes and refused to pick. Their side-effect surfaces were measured, each applied alone to the base gate with the mutation proven on disk before any verdict was read:

- **(1) a third `[REMOVED]` arm in `markerStatus()`** — on the acceptance sample this leaves the gate at **exit 0**. `markerStatus()` is consulted only when `led?.status` is falsy, and all 79 tombstones the walk reaches take their status from an explicit row, so this arm changes the verdict of **0 of the 79** today. What it does change is the no-row case, which the gate currently reports as UNCLASSIFIED and fails on, and which `ORPHAN_GUIDANCE` tells authors never to create — so (1) alone would *silence* an existing loud failure. It cannot satisfy the acceptance.
- **(2) a rule beside the tombstone-row rule** — implemented here. The forbidden set is a named constant holding exactly `live`, which is what was ruled. `planned` / `experimental` / `live-elsewhere` on a tombstone are equally false, and their population today is zero, so widening the set is a separate measurement rather than a rider on this one.

## Acceptance

Every leg runs the real gate the way CI invokes it, via `--ledger-root` against a copied ledger.

The carrier the legs flip is **derived per run from the gate's own `report.tombstones` enumeration**, not named: top-level, not a `BOUND_PROOF_PATHS` coordinate, and currently `dead`. A fixture pinned to one row's *current verdict* is pinned to a moving fact, which is how this block first broke — it named `agent/tools`, that row went `dead` upstream, and the "this would be a no-op" guard took four cases red. The seat re-derived the same filter chain independently (2026-09-18T22:08Z): **79 enumerated → 42 top-level → 42 unbound → 42 `dead`**, carrier `action/bulkEnabled`. ⚠️ Today only the top-level filter bites — the bound-path and `dead` filters each drop **0** — so those two are shape guards holding a property, not measured reductions, and the block's own assertion is on *eligible > 0*, never on a filter biting.

The red and green samples differ in **exactly one status value** (plus the arithmetic the generated count artifact requires, so a stale count cannot mask the verdict), and "one value apart" is now checked as bytes by its own case rather than asserted in prose.

| leg | before | after |
|---|---|---|
| **main** — tombstone + row `status: live` | exit **0** (green — the gate had no reading) | exit **1**, `<derived carrier> -> "live"`, with exactly one `✗` block in the output |
| **lit control** — same sample, legal status | exit 0 | exit **0** |
| **dark control** — ordinary non-tombstoned `live` rows | exit 0 | exit **0**, and the whole gate output is identical except the two lines added here |

The dark control is the strong form: a full diff of the gate's output on a legal ledger, before versus after, differs only by the new `tombstoned keys:` census line and the new clause in the summary sentence. All 39 per-type verdict lines and every other check's output are byte-identical, so nothing else was tightened.

The before-legs were taken by reverting both files to the base commit after the fix was committed, with the mutation proven on disk (blob hashes moved, the new symbols counted 0 occurrences) and the restore proven by `git diff HEAD` being empty and the blob hash matching `HEAD`'s.

## Non-vacuity

`0 forbidden` reads identically whether every row is honest or the marker drifted and the scan reached nothing, so the gate prints the population it asked on every run, pass or fail. On this PR's head `3d49d7a199` against the shipped ledgers (2026-09-18T22:08Z) it prints the green form:

```
tombstoned keys: 79 `[REMOVED]` tombstone(s) reached by the walk, 79 graded with a status the tombstone allows.
```

and on the red leg the same line carries the finding:

```
tombstoned keys: 79 `[REMOVED]` tombstone(s) reached by the walk, 78 graded with a status the tombstone allows, 1 FORBIDDEN.
```

and `orphans.test.ts` holds `TOMBSTONE_MARKER` equal to what `retiredKey()` actually writes, read from the producer rather than restated.

## Changeset

`skip-changeset`, measured rather than assumed. `@objectstack/spec`'s `files[]` is `dist · json-schema · liveness · prompts · llms.txt · README.md · src/**/*.zod.ts · CHANGELOG.md · api-surface · api-surface-declarations · spec-changes.json`; `scripts/` is not in it, and the build's entries are src modules. Each of the three symbols this PR adds occurs in **0** files under those shipped paths; positive controls through the same instrument and the same path set: `retiredKey` in 102, `ObjectSchema` in 30. Nothing this diff moves is shipped by any package.

## Acceptance notes

Noted, not filed — observations, with no reproducible defect, contract violation or authoring trap behind them:

- PR #19059 landed `scripts/liveness/tombstoned-row-status.test.ts` as `a362e0e425` (2026-09-18T22:08Z), a contract test asserting the same invariant against the **real** ledgers at **top-level granularity only**. It is complementary rather than duplicative — that one pins today's ledgers in `Test Core`, this one is a gate rule over every depth the walk reaches and over any ledger root. Left untouched here, per triage's `⛔ 不折叠`. This PR is the second of the two, and both files were written without sight of the other.
- The gate's `markerStatus()` still has no `[REMOVED]` arm, deliberately: with a row present it is never consulted, and without one the UNCLASSIFIED failure is the louder answer. Recorded so the next reader does not read its absence as an oversight twice.

---


os-bill:
## Seat review — re-measured against the tree, not the report

`domain:spec` seat 2. The three stale statements the dev flagged are now corrected in the body above. Below is what the seat measured **itself**, on head `3d49d7a199`, before landing (2026-09-18T22:08Z).

| reading | instrument | result — all taken 2026-09-18T22:08Z |
|---|---|---|
| the ordering blocker | #19059 merged as `a362e0e425`; `main` merged into this branch | **discharged** |
| the gate on the real ledgers | `tsx scripts/liveness/check-liveness.mts` | **exit 0** — `79 … 79 graded with a status the tombstone allows` |
| the suite | `vitest run scripts/liveness/` | **333 passed (333)**, 12 files, exit 0 |
| the census | the gate's own `--json` `report.tombstones` | 79 enumerated, **all 79 graded `dead`**, 0 taking status from anything but an explicit row; 37 of the 79 sit below the top level, spanning 21 of the 39 types |
| the derivation | the seat's own filter chain, written independently against `BOUND_PROOF_PATHS` and the ledgers | 79 → 42 top-level → 42 unbound → 42 `dead`, carrier `action/bulkEnabled` — reproduces the block's own figures exactly |
| `merge=os-regen` | `git check-attr merge` on each of the 4 changed files | `unspecified` on all four — nothing here resolves silently |
| governed surface | `check-governed-merges --pr 19092` | 0 of 4 paths hit the register — ordinary queue landing |
| clause-② | `check-clause2-carriers --pair 19092` | exit 0; `Clause-②: no` on both the card claim (5733893029) and the body. A new forbidden combination **narrows** the accepted set, which is not clause-②. |

**One observation, recorded rather than filed.** Of the derivation's three filters only the first bites today: after the top-level filter takes 79 to 42, the `BOUND_PROOF_PATHS` filter and the `dead` filter each drop **0**. That is not a defect — they hold a property of the *shape*, and the block asserts on `eligible > 0` rather than on any filter biting — but "three filters" should not be read as three measured reductions, so the body now says which one is load-bearing today.

**Radius of that re-measurement.** It is the gate's own walk, so it is silent about the 59 further `[REMOVED]` tombstones under blanket-verdict containers the walk never reaches. Known target necessarily outside it: #19118 — strict-route retirements leave no `[REMOVED]` marker at all, so they are invisible to `report.tombstones` by construction. The 79 is a correct answer to a narrower question than "how many retirements exist".

Ready flipped; auto-merge mounted on SQUASH.


---

PR #19097, opened 2026-09-18
fix(spec): route build-openapi through the published-projection choke point, and pin the choke point to the tree

### LINKED ISSUE #19020: [finding] projectPublishedJsonSchema's docblock claims it is the ONE published-projection call, but build-openapi.ts:95 is a second override-less z.toJSONSchema path writing the published openapi.json
Surfaced by the isolated at-tier clause-② review of PR #19005 (card #18670, item 2's third arm) as its NON-BLOCKING finding 2. Filed by the `domain:spec` seat 3 because a reviewer, like a dev, does not POST issues. `finding` only — no `domain:*`, no `type`, no `priority:*`.

## Class

**(b) — violates a declared contract.** The contract is a docblock claim, and the measurement contradicts it.

PR #19005 routes the generator and the detector through ONE shared helper, `projectPublishedJsonSchema`, so that the dropped-refinements override cannot be dropped on one side. That is sound, and the review confirmed it: the publication path went from **4** direct `z.toJSONSchema(` calls at base to **0** in `build-schemas.ts` and **0** in `dropped-refinements.ts`.

But the helper's own docblock describes itself as the ONE call anywhere the published projection is produced, and that is wider than what was measured. `packages/spec/scripts/build-openapi.ts:95` is a **second, override-less `z.toJSONSchema` path**, and it writes `packages/spec/json-schema/openapi.json` — a file in the same published directory, exported from the package as `./openapi.json`.

⇒ the choke point is real for the surfaces the ledger covers, and the docblock claims a scope the choke point does not have.

## Why it is NOT urgent, stated so nobody prices it as a live bug

Measured at PR #19005's head: `openapi.json` is **byte-identical base to head**, and embeds **0** `SSLConfig`, **0** `dialect`, **0** `\S`, **0** `dependentRequired`. So **no declared arm is affected today** — the second path currently produces nothing the override would have changed. The defect is that nothing prevents that from changing: the ledger does not read that population, so a future refinement reaching `openapi.json` would be projected wide and silently, which is exactly the failure class #18670 exists to close.

## The two candidate shapes, neither of them chosen here

1. Route `build-openapi.ts` through `projectPublishedJsonSchema` too, and widen the ledger's population to cover `openapi.json`.
2. Narrow the docblock to the population the choke point actually governs, and record `openapi.json` as deliberately outside it.

⚠️ These are not equivalent: (1) changes what gets published if any refinement ever reaches that file; (2) changes only prose. ⛔ Do not treat this as a mechanical tidy — whichever is right, the choice is about how wide the published projection guarantee is meant to be, which is #18670's own subject.

## Evidence limits

Everything above is from the reviewer's reading at head `6007a484a50cf8a985390a795cb69832f3df6af3`, and this seat has re-derived none of it independently. The load-bearing readings to re-take before pricing: that `build-openapi.ts:95` really does call `z.toJSONSchema` without the override, and that `openapi.json` really embeds none of the declared arms. Both are cheap greps; neither was done by me.

## Dedup words

`build-openapi second toJSONSchema` · `projectPublishedJsonSchema choke point scope` · `openapi.json override-less projection` · `published projection population ledger` · `json-schema openapi wide refinement`

Dedup **was run** before filing: a semantic search including closed cards returned 13 results whose only same-family hit was #18670 itself, this card's parent. Limit: the words are the vocabulary of the generator, so a card filed from the consumer side ("the OpenAPI document accepts metadata the runtime refuses") would share none of them.

---

os-bill:
Fixes #19020

Clause-②: no

Triage set the deliverable order and this PR keeps it: ① measure whether today's published `openapi.json` diverges from the override-projected one, ② only then choose a remedy. ⭐ Every reading below was taken in a dedicated worktree at base `86b9000da`, on a `json-schema/openapi.json` **this run built**. That matters: the tree is gitignored, so no `git grep` on any ref can read it, and a stale copy on disk reads 0 for every probe word while saying nothing at all about today's source.

## ① Divergence measurement — 0 named schemas, 0 named fields

`build-openapi.ts` projects exactly nine names. Three legs, one process each:

| leg | reading |
| --- | --- |
| published `openapi.json` vs `projectPublishedJsonSchema`, per named schema | **9 of 9 byte-identical** |
| the old `build-openapi.ts:95` call vs the helper, same nine | **9 of 9 byte-identical** |
| refinement census over the nine | **0 dropped / 0 projected / 0 undecidable** |

The third leg is the cause of the first two: **there is no `custom` check anywhere in the nine's reachable graphs today**, so the override has nothing to emit and the two projections cannot differ.

⭐ **LIT CONTROL** — the same comparator, on schemas the override provably rewrites:

```
DIVERGENT  api/UpdateAiConversationRequestSchema (requiredOneOf)   /allOf        added by override
DIVERGENT  shared/ExpressionSchema (requiredOneOf + non-blank)     /allOf        added by override
DIVERGENT  data/SSLConfigSchema (dependentRequired)   /dependentRequired         added by override
```

⭐ **DARK CONTROL** — schemas unrelated to the override, byte-identical on both sides:

```
IDENTICAL  api/AiAgentCapabilitiesSchema
IDENTICAL  api/CreateAiConversationRequestSchema
IDENTICAL  synthetic plain z.object
```

⚠️ **Instrument radius, and one known target necessarily outside it.** The comparator reaches exactly the nine names in `build-openapi.ts`'s own `contractSchemas` table, each projected whole (subschemas included). `src/api` exports 402 names ending in `Schema`; nine are projected here, so **393 are outside the radius** — `UpdateAiConversationRequestSchema` is one of them, and it is a lit target, which is why "0" here is a statement about the nine and about nothing else.

⚠️ **A control the instrument needed.** An early run of the same comparator reported 2 divergences (`ListRecordResponse` and `BulkRequest`, both at an items-level `description`). They were an artefact of the probe setting `OS_EAGER_SCHEMAS=1`, which `gen:openapi` does not set: re-running the file against today's **bare** call in the same process reproduced the same 2, proving the difference was process-vs-file and not override-vs-bare. Without that third leg the escalation condition would have been reported hit on an instrument artefact.

⇒ **The escalation condition is NOT met.** Triage's wording is "已发布的 `openapi.json` 与经 override 的投影在任一处不一致(具名 schema + 具名字段)⇒ 升 `p1`". Measured: no such place exists today.

## ② Remedy — shape 1, routed, and why shape 2 was not enough

`build-openapi.ts:95` now calls `projectPublishedJsonSchema`. The published artefact is **byte-identical across the change** (sha256 `34b1dc9c…` on both sides, `cmp` exit 0), because of the census above.

Why routing and not prose, in one measurement. An injection ablation, both legs through the real generator, restore proven against the HEAD blob:

```
LEG A  routed call + one injected declared refinement on BulkRequest
       openapi.json  BulkRequest.allOf = [{"anyOf":[{"required":["records"]},{"required":["allOrNone"]}]}]   498 bytes
LEG B  the OLD bare call + the SAME injection
       openapi.json  BulkRequest.allOf = null                                                                424 bytes
restore  blob == HEAD (04db95741ff4) and `git diff HEAD` empty
```

So on the day any one of the nine grows a `.refine()` on the closed list, the old call published the **wide** document — the rule stated in `api/BulkRequest.json` and silently absent from the same schema's copy in `openapi.json`, two artefacts shipped in one tarball disagreeing about which documents the contract accepts. Narrowing the docblock would have left that live.

**On widening the ledger population — a reading that corrects the card's framing.** The card says "the ledger does not read that population". That is true of the FILE and not of the SCHEMAS: all nine are already members, as `api/ApiError`, `api/BaseResponse`, `api/BulkRequest`, `api/BulkResponse`, `api/CreateRequest`, `api/DeleteResponse`, `api/ListRecordResponse`, `api/SingleRecordResponse` and `api/UpdateRequest` in the tracked `json-schema.manifest/api.json` (441 entries). So the population needed no widening at schema granularity — what needed fixing was the projection CALL, which is what this PR does. What remains outside is the `x-dropped-refinements` ANNOTATION on the `openapi.json` rendering; with a census of 0 it would emit nothing today, and it is a decision about how wide the published-projection guarantee reaches, which is #18670's own subject.

## ③ The choke point stops being a convention

The reading that missed `build-openapi.ts` was a grep over a hand-picked file list. `packages/spec/scripts/published-projection-choke-point.test.ts` replaces the list with the tree: every non-test source under `scripts/`, comments and string literals masked through the shared `scripts/js-comment-mask.mjs`, with a per-file allowance table whose rows each carry their reason. It found three call sites the hand-picked grep had not — and the count it first declared for one of them was wrong, which it also caught.

Reverse verification: reverting `build-openapi.ts:95` to the bare call turns the pin **red** with `build-openapi.ts: 1 direct call(s), declared 0`; restoring returns the blob to HEAD with an empty `git diff HEAD`.

## Verification

- `pnpm --filter @objectstack/spec exec vitest run scripts/published-projection-choke-point.test.ts scripts/openapi-self-consistency.test.ts scripts/refinement-projection.test.ts scripts/dropped-refinements.test.ts` — 4 files / 97 tests, exit 0
- `pnpm --filter @objectstack/spec typecheck` — exit 0 (test layer included)
- `pnpm --filter @objectstack/spec test:repo` — 32 files / 543 tests, exit 0
- `pnpm --filter @objectstack/spec test` — 491 files passed / 1 skipped, 14488 tests passed / 1 skipped, exit 0
- 56 gate commands derived by `node scripts/pm/dispatch-gates.mjs --commands --repo objectstack-ai/objectstack`, each exit code captured before any pipe and reconciled: **56 derived, 51 run, 5 NOT-MEASURED, 0 UNRUN**. Two of the 51 were real catches this PR then fixed — `check:comment-mask-adoption` (a private comment stripper, now routed through the shared mask) and `check:cross-package-test-inputs` (the pin reads outside its package, now declared in `vitest.repo-tests.json`).
- The 5 NOT-MEASURED answered **exit 3 — PREREQUISITE NOT MET**, which is neither a pass nor a finding: `check:dts-closure`, `check:dual-build-cjs-loads`, `check:lean-entry-closure`, `check:sourcemap-no-sources-content`, `check:type-check-debt` each need a full repo build this run did not take. Left to CI. ⚠️ And the reconciliation answers one link only — the 50 artifact-roster families, the 11 wide-population families, the 9 pending-changeset families and the 5 path-scheduled CI jobs are all outside that 56.

## Why `skip-changeset`, measured rather than argued

"Published" is what each package's `files[]` actually ships.

- `packages/spec` `files[]` does not carry `scripts`, and this diff is `scripts/` only plus `vitest.repo-tests.json`.
- `dist` is built from `src`: **0** non-test files under `packages/spec/src` import anything under `scripts/` (positive control — the same grep over all src files, tests included, reads **26**, every one a `.test.ts`).
- `json-schema/openapi.json` is byte-identical before and after (same sha256, `cmp` exit 0).
- Neither changed symbol appears in any shipped tree that exists without a build: `projectPublishedJsonSchema` **0**, `published-projection-choke-point` **0** (positive control — `BulkRequestSchema` reads **4** over the same paths).

## Acceptance notes

⚠️ **Findings for the seat to file — not fixed here, and out of this card's scope.**

1. **`os generate` writes an author-facing JSON Schema through an override-less `z.toJSONSchema`, measured WIDER than the runtime at seven named sites.** `packages/cli/src/commands/generate.ts` lines 3050 / 3057 / 3064 project `ObjectStackDefinitionSchema` — the schema an author's editor validates their metadata against — with no refinement override. Measured (input direction, `unrepresentable: 'any'`): bare 1,430,616 bytes vs override-projected 1,452,594 bytes, census 39 dropped / **7 projected** / 1 undecidable. The seven the override would emit and the CLI's file does not: `jobs.element.schedule.options[0].expression.options[0].in` and `.options[1]`, `objects.element.titleFormat.options[0].in` and `.options[1]`, `sharingRules.element.condition.options[0].in`, `.options[1]` and `.options[1].source`. This is the class-(c) trap exactly: the editor accepts a blank expression or an Expression carrying neither `source` nor `ast`, and the runtime refuses it. Dedupe words: `cli generate json schema override-less` · `ObjectStackDefinitionSchema published projection` · `author-facing schema wider than runtime` · `os generate refinement dropped` · `editor accepts what the platform refuses`.
2. **`build-react-blocks-contract.ts` projects into the published skills catalog through an override-less call**, and 1 of its 3 block schemas is DIVERGENT under the override at projection level. ⚠️ Radius: measured on the PROJECTION, not on the rendered `skills/objectstack-ui/references/react-blocks.md` — whether the markdown prop table changes was NOT measured. Its output is a governed `skills/**` file, so routing it is a governed-surface decision, not a tidy. Dedupe words: `react-blocks contract override-less projection` · `skills catalog published projection` · `gen:react-blocks refinement dropped`.
3. **`gen:openapi` output depends on `OS_EAGER_SCHEMAS`.** With the variable set, two of the nine (`ListRecordResponse`, `BulkRequest`) gain an items-level `description` that the default path drops — `RecordDataSchema` is a `lazySchema()` whose inner `.describe()` does not survive lazily. `gen:schema` sets the variable, `gen:openapi` does not, so two generators disagree about one source. Repro: run `gen:openapi` twice, once with `OS_EAGER_SCHEMAS=1`, and diff. Dedupe words: `OS_EAGER_SCHEMAS published artefact differs` · `lazySchema describe lost` · `gen:openapi env-dependent output`.

**noted, not filed:** the review's "4 direct calls to 0" reading was never turned into a gate, which is why this card exists at all; this PR's pin is that gate, so there is nothing left to file. Taker: this PR.

**A correction to the card, stated because it was load-bearing.** The card's Evidence-limits section calls both re-readings "cheap greps". The second is not one, and the PM had already recorded that on this card: `packages/spec/json-schema/` is gitignored at `.gitignore:63`, 0 tracked files, so reading `openapi.json` requires a build. This run took that build. The first half of the card stands exactly as written.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JbZnqu8bt6YqfJsr9vaFb3)_


os-bill:
**重跑结论:我判「不是本 PR 的失败」被证实了。** 取数时刻 2026-09-18T19:54Z

`Test Core (1/6)` 现在**绿**。⏱️ 2026-09-18T19:54Z 判读 head `40b2cf772d37fd9dd3fec4a35b75a617ae433e45`:**0 failure · 0 pending · 7 个必过 context 全绿**。

## 为什么这次重跑是**唯一**的一次,而不是「再掷一次骰子」

章程只允许在两种情形下重跑,且总共一次。我在按下它之前先把判据写完了:

1. 失败点名的包是 `packages/qa/vitest-filter-preflight`,而本 PR 的 4 个文件里**没有一个**在 `packages/qa/` 或 `packages/cli/` 下(`git diff --name-only` 实读为空);
2. 同一次跑里**同一条断言**对 `core` / `objectql` / `rest` / `runtime` / `spec` / `types` 六个 config 全绿,耗时 1028 / 1723 / 1181 / 1625 ms —— 这条腿要 spawn 一个真 vitest 子进程,常态耗时距 5000ms 上限只剩约 3 倍余量;
3. `origin/main` 同一门实读 **success**(两次跑都是)⇒ ⛔ 不是基线红;
4. 本地在**本 PR 的 head** 上跑同一个文件:**65 passed**,exit 0。

⇒ 四条都指向同一个方向,重跑才被花掉。⚠️ 如果重跑仍红,按章程它就**是**本 PR 的,我回去根因 —— 现在不必了。

⚠️ **一条自纠,免得这条记录被当成干净的**:我第一次本地跑读到 `2 failed`,失败的还是**另一个** config。那是**我的台架** —— worktree 只给 `packages/*` 链了 `node_modules`,漏了 `packages/qa/*` 这种再深一层的;补齐到 74 个之后同一 head 读 65 passed,与真检出的对照完全一致。那一轮的「失败」⛔ 不是读数。

## 入队前的门

- `check-clause2-carriers --pair 19097` → **exit 0**,两侧载体一致读 `Clause-②: no`。
- `check-governed-merges --pr 19097` → **exit 0**,`0 of 4 path(s) hit the register` ⇒ ⛔ 非受管面,普通队列落地。⭐ 这一读是**对最终文件表**重跑的,⛔ 不是复用本轮更早的那次 —— 登记表两天内长过几次,早先的读数是回忆不是读数。

⇒ 布防。


---

Issue #16431, opened 2026-09-07
[finding] Five filter operators ($gt/$gte/$lt/$lte/$between) reach NO published reference page — `build-schemas.ts` skips their whole schema over an unrepresentable `z.date()`, and the skip is silent


huangyiirene:
Filed by the `domain:spec` dev seat (`session_01T6HeZvT9wdSJD1ZxJb5Eno`) while implementing #15059 (PR #16432). Out of scope there by that card's own fence — #15059 is prose on `.describe()`, and the remedy here is a change to the generator or to a schema shape, which #15059 forbids. Unassigned, bare, for triage.

## Measured on `origin/main` @ `0a038cc06d92ba3ae663f9f1e12c52e72dbc7d39`

`packages/spec/scripts/build-schemas.ts` emits one JSON Schema per exported `z.ZodType`. When `z.toJSONSchema()` throws for a known-unrepresentable type it retries with `io: 'input'`, and when that throws too it **skips the export with a `console.warn` and continues**. Four exports of `packages/spec/src/data/filter.zod.ts` take that path. From my own run of `pnpm --filter @objectstack/spec build`:

```
⊘ Data.ComparisonOperatorSchema: Date cannot be represented in JSON Schema (skipped)
⊘ Data.FieldOperatorsSchema: Date cannot be represented in JSON Schema (skipped)
⊘ Data.NormalizedFilterSchema: Date cannot be represented in JSON Schema (skipped)
⊘ Data.RangeOperatorSchema: Date cannot be represented in JSON Schema (skipped)
```

The `Date` is real and deliberate: `orderingComparandSchema` is `z.union([z.number(), z.date(), z.string(), FieldReferenceSchema])`, and `rangeEndpointSchema` is `z.union([z.number(), z.date(), z.string()])`. Nothing is wrong with the schemas; the generator simply cannot project them.

`content/docs/references/**` renders from `packages/spec/json-schema/`, so a skipped export has no section. Measured with lit controls on the same tree:

```
grep -c 'ComparisonOperator' content/docs/references/data/filter.mdx   -> 0
grep -c 'RangeOperator'      content/docs/references/data/filter.mdx   -> 0
grep -c 'FieldOperator'      content/docs/references/data/filter.mdx   -> 0
grep -c 'SpecialOperator'    content/docs/references/data/filter.mdx   -> 2   (control: the file IS live and the pattern CAN hit)

[...]

PR #18618, opened 2026-09-17
fix(spec): judge a nested RETIRED_KEYS_BY_MAJOR row instead of ignoring it

os-bill:
Fixes #17969

Clause-②: yes

## What was broken, and the reading that says so

`check:authorable-surface` builds `currentKeys` from `schema.properties` **one level deep**
(`packages/spec/scripts/build-schemas.ts`). Check (b2) — the guard that refuses a
`RETIRED_KEYS_BY_MAJOR` entry naming a key the build still emits as LIVE — consults that map, so a
**dotted** row matched nothing in it and was silently ignored rather than judged. 48 of the shipped
rows are dotted.

Reproduced on this branch before the fix, by ablation, with the card's own controls:

| probe | pre-fix `check:authorable-surface` |
|:--|:--|
| fabricated nested row `system/SchemaLevelIsolationStrategy:performance.zzNotARealKey9999` | **exit 0**, zero failures |
| lit control: `system/SchemaLevelIsolationStrategy:strategy`, a live top-level key | **exit 1**, refused |

## The coupling, measured before the fix was written

Three lines below `currentKeys` the same map is `currentEntries`, the **published**
authorable-surface baseline. The dispatch asked for a measurement rather than an argument, so:

1. **Does widening `currentKeys` in place move `authorable-surface/**`?** Measured, not reasoned:
   the one-level loop was replaced with a recursive walk, `gen:schema` was run, and the diff read
   **+14,376 lines across all 14 shards** (`ai 320, api 6099, automation 496, data 1393,
   identity 83, integration 170, kernel 1997, marketplace 86, qa 93, security 90, shared 6,
   studio 187, system 1648, ui 1709`). The script was then restored byte-identical
   (`git hash-object` back to `b4f0ac58c00086a60528b88acdf1c45f13115387`, `git diff HEAD` empty).
   So yes: widening in place publishes the whole nested key space as a ratcheted, author-facing
   contract.
2. **Is a nested view consulted only by the checker sufficient?** Yes, and that is what shipped.
   `currentKeys` and `currentEntries` are untouched; the nested state is resolved for the
   **registered rows only** and reaches nothing that emits. Deliberately no nested map is
   materialised — the hazard here is a nested key space lying around for a later author to pipe
   into `currentEntries`, so there is nothing to pipe.
3. **Does the published surface have to move?** No. `git diff --exit-code -- packages/spec/authorable-surface/`
   is clean after a full `gen:schema` on this branch, and `check:generated` reports all 15 generated
[...]

PR #19135, opened 2026-09-18
feat(spec): give the published `x-unprojectable-branches` record a reader

### LINKED ISSUE #17107: [finding] union-branch projection (PR #17085): the 「refuse a marked node anywhere but a direct anyOf/oneOf member」 clause is pinned only at the root — loosening the guard to root-only passes 19/19 and would publish `FlowFunctionEntry` with an `x-os-unprojectable` marker leak; `x-unprojectable-branches` has no reader
Filed by the director seat (summon #18 segment 3, `session_017Js5kTpTtxieBjPyScgxJ3`, 2026-09-09) from the post-merge tier audit of PR #17085 (5599052798, findings F2 + F3; card #16431 option (a)). ⛔ Not graded, no `domain:*` — triage's; landing point `packages/spec/scripts/lib/union-branch-projection.ts` + its test, and `build-docs.ts` for the second half. Class (a): incomplete pin — an assertion never observed to fail.

## F2 — the nested-mark refusal is not pinned

`packages/spec/scripts/lib/union-branch-projection.ts:297` is the guard that refuses the projection when a marked (unprojectable) node sits anywhere except as a direct member of an `anyOf` / `oneOf`. `packages/spec/scripts/union-branch-projection.test.ts` pins it only at the root. Mutations executed on a scratch copy of the merged head `3b78f35e70`:

| mutation | result |
|:--|:--|
| M2 — projection disabled | 8 fail |
| M3 — strict re-conversion removed | 4 fail |
| M1 — guard at `:297` deleted | 1 fail (only `refuses a union whose every branch is unprojectable`, `:157`) |
| **M1b — guard loosened to `=== '#'` (nested marks tolerated)** | **19/19 pass** |

The property / record / array pins at `:152-154` and the `PersistenceAdapterSchema` pin at `:229` pass under M1b vacuously via the `pruned.length === 0` early-out (`:300`) — their schemas contain no union, so the guard is never reached. Under M1b the `FlowFunctionEntry` shape publishes `{"anyOf":[{"type":"object","properties":{"handler":{"x-os-unprojectable":"function"}},"required":["handler","name"],…},{"type":"string"}]}` — a marker leak plus a required annotation-only property, the failure the PR body itself calls load-bearing. The only backstop is indirect: the ledger's `repaired` ratchet (`build-schemas.ts:2774-2779`) refuses the build when `Automation.FlowFunctionEntrySchema` starts emitting, contingent on that ledger line staying.

Ask: a unit pin `projectByPruningUnionBranches(z.union([z.function(), z.object({ handler: z.function() }), z.string()]))` → `null`, and a pin that no emitted `json-schema/**` file contains `x-os-unprojectable`.

## F3 — `x-unprojectable-branches` has zero readers

`packages/spec/scripts/build-schemas.ts:486-498` writes the record; `git grep x-unprojectable-branches origin/main` finds only the writer and the changeset. `build-docs.ts` does not render it, so `content/docs/references/data/filter.mdx` shows Type `number | string | { $field … }` beside prose reading "a number, a Date, a string", and `hook.mdx` shows `handler: string` beside "or inline function (pre-build)", with nothing on the page saying a branch was dropped. Not an ADR-0078 breach (an `x-` annotation on a generated artifact, not an authorable key), but the PR body's "recorded ON the artifact" is the whole record. Ask, same card: `build-docs.ts` emits a one-line note from `x-unprojectable-branches`, or the module docblock states that artifact-only is the intended reach.

Re-check: `node --test` / `pnpm --filter @objectstack/spec exec vitest run scripts/union-branch-projection.test.ts` after applying M1b locally (change the guard's comparison at `:297` to `=== '#'`) → all green today; `git grep -n "x-unprojectable-branches" origin/main -- packages/spec/scripts` → writer only.

Dedupe: `search/issues` for 「union-branch-projection unprojectable pin」 returned no result set at filing time (query answered empty); the audit that found this is minutes old and PR #17085 merged at 08:05Z, so no earlier card can name it. Related: #16431, PR #17085, #16906 (repaired as a side effect of the same PR), #17040 (record incident row 4).

os-bill:
Fixes #17107

Clause-②: no

F3 of #17107, the card's only live half — F2 landed in #17804 and was verified on `origin/main`. `x-unprojectable-branches` is written onto every JSON Schema the per-branch projection publishes, and `packages/spec/package.json` ships `json-schema/` in `files[]`, so it is a **published** field. It had **one writer and no reader**: nothing in this repo could tell a faithful record from one naming branches the projection never dropped.

## The premise, re-derived (hits printed, with a lit control)

Tree: `objectstack-ai/objectstack` at `a675ad4ef0` (`origin/main`, 2026-09-18T21:59Z). `git grep -n` over the whole tree.

**Subject — `x-unprojectable-branches`, 5 hits, 0 of them a reader:**

```
.changeset/filter-operator-schema-projection.md:33   changeset prose
packages/spec/scripts/build-schemas.ts:435           comment
packages/spec/scripts/build-schemas.ts:546           THE WRITER
packages/spec/scripts/build-schemas.ts:554           comment
packages/spec/scripts/lib/dropped-refinements.ts:24  docblock prose
```

**Lit control, same instrument — the sibling marker `x-os-unprojectable`, which really is read:**

```
packages/spec/scripts/lib/union-branch-projection.ts:71  export const UNPROJECTABLE_MARK = 'x-os-unprojectable';
packages/spec/scripts/union-branch-projection.test.ts:208
```

The control fires and resolves to a named constant the projection branches on, so the zero beside it is a reading, not a dead instrument. This matches the table in comment 5732887703 that the triage adjudication (5733358188) adopted in place of the card's own now-inexact sentence; `dropped-refinements.ts:24` is prose, added after filing, and is not a reader.

## Which kind of reader — and why the docs page was the wrong one for this card

Triage ruled verbatim that a reader need **not** be a docs page: reading route A as "render it in the docs" is the card's two named pages, not the definition of the word. Both kinds were available. This PR builds the **checking** reader, for four reasons that are properties of this card rather than preferences:

1. **A docs note cannot fail.** Rendering the record into `filter.mdx` / `hook.mdx` prints whatever the record says. If the record named a branch the projection never dropped, the page would print that too — faithfully. The card's class is (a), *an incomplete pin: an assertion never observed to fail*, so a reader that cannot disagree with the writer does not close it.
2. **`content/docs/references/**` carries `merge=os-regen`** (`.gitattributes:153`), where a conflict resolves silently. ⛔ **This PR regenerates nothing under that tree** — the diff is three files under `packages/spec/scripts/`, and `content/docs/references/` is untouched.
3. **`build-docs.ts` is outside the file surface this card was dispatched with.** The dispatch cleared `union-branch-projection.ts`, its test, and `build-schemas.ts`.
4. It strengthens an existing gate rather than adding a required one: the audit runs inside `gen:schema`, which every spec build already runs.

The docs half of the card's original ask stays unbuilt and is called out in the acceptance notes rather than silently dropped.

## What the reader is

`auditUnprojectableBranchRecord` reads the record **off the artifact, by its key**, and checks it in two directions — either alone passes a record that lies:

- **Replay.** Removing exactly the branches the record names from the unpruned projection must reproduce the published file. Only root **annotation** is allowed on top (`$id`, and `x-` keywords, which every validator ignores); anything structural stays in the comparison. This catches a pointer whose index shifted — the hazard `pruneMarkedUnionBranches` records the pre-removal index to avoid — an entry for a branch that was never dropped, and a missing entry for a union that collapsed whole.
- **Completeness.** Every marked direct union member of the unpruned projection must appear under the type it was marked with. The replay cannot see this on its own: a collapsed union is removed wholesale, so the members inside it reproduce the published file whether the record names them or not.

`build-schemas.ts` calls it in the one place that still holds the tree the record describes, and the build **fails** on a disagreement. The key also moves from a bare literal at the writer to `UNPROJECTABLE_BRANCHES_KEY`, declared beside the prune that produces the record, so writer and reader cannot drift to two spellings.

`BranchProjection` gains `unpruned`, a clone taken before the prune mutates the tree in place — the record states what is *missing* from the published schema, and the only tree that can answer whether it states it correctly is the one the branches came out of.

## Ablation — the reader really fails when the marker and the real dropped branches disagree

Three legs, all against the real generator on the real corpus (1541 schemas, 5 of them branch-pruned, 31 recorded branches). Mutations landed through `scripts/ablation-replace.mjs`, which proves each write on disk (anchor count 1 to 0, `git hash-object` blob moved) and proves each restore against `HEAD`. No `dist` leg: `gen:schema` runs `build-schemas.ts` through tsx from source, so `dist/` is not on the resolution path.

**Leg A — record tampered (`prunedBranches.slice(1)`), reader present ⇒ `gen:schema` exits 1:**

```
ablation-replace: ok mutation landed: anchor 1 -> 0, blob e5f42c7917a1 -> 4efb4df32134
❌ 5 `x-unprojectable-branches` record(s) disagree with the schema they annotate:
   ⤷ Data.ComparisonOperatorSchema: #/properties/$gt/anyOf/1 was dropped as "date" and the record does not say so
   ⤷ Data.FieldOperatorsSchema: #/properties/$gt/anyOf/1 was dropped as "date" and the record does not say so
   ⤷ Data.NormalizedFilterSchema: #/properties/$and/items/anyOf/0/additionalProperties/properties/$gt/anyOf/1 was dropped as "date" and the record does not say so
   ⤷ Data.RangeOperatorSchema: #/properties/$between/prefixItems/0/anyOf/1 was dropped as "date" and the record does not say so
   ⤷ Data.HookSchema: x-unprojectable-branches is [], not the non-empty array of dropped branches this projection has
Exit status 1
ablation-replace: ok restored: blob == HEAD (e5f42c7917a1) and `git diff HEAD` is empty
```

**Leg B — the same tamper with the reader deleted, i.e. today's code ⇒ `gen:schema` exits 0 and the wrong record ships unremarked:**

```
ablation-replace: ok mutation landed: anchor 1 -> 0, blob e5f42c7917a1 -> 61086c23e26a
✅ Successfully generated 1541 schemas.
data/ComparisonOperator.json carries:
  [{"at":"#/properties/$gte/anyOf/1","type":"date"},
   {"at":"#/properties/$lt/anyOf/1","type":"date"},
   {"at":"#/properties/$lte/anyOf/1","type":"date"}]
ablation-replace: ok restored: blob == HEAD (e5f42c7917a1) and `git diff HEAD` is empty
```

Leg B is the defect itself, executed: `$gt`'s dropped Date branch vanishes from the record, the file publishes anyway, and nothing anywhere objects. Leg A is the same run with the reader in place.

**Restore leg — unmutated tree, real generator ⇒ exit 0, correct record back, whole tree clean:**

```
✅ Successfully generated 1541 schemas.
record now: [{"at":"#/properties/$gt/anyOf/1","type":"date"}, {"at":"#/properties/$gte/anyOf/1","type":"date"},
             {"at":"#/properties/$lt/anyOf/1","type":"date"}, {"at":"#/properties/$lte/anyOf/1","type":"date"}]
git status --porcelain: (empty)
```

## Tests

`packages/spec/scripts/union-branch-projection.test.ts` — 21 tests before, **28 after**, all passing. The seven new ones pin: the record `build-schemas.ts` actually writes, for all four exports #16431 measured (asserted non-vacuous — each has a record to audit); a shifted pointer; an under-reporting and an inventing record; a record that renames what was dropped; a malformed record; the annotation carve-out and a structural addition beside it; and the collapsed-union case, where the completeness direction is the only one that can report the loss.

## Local verification

Gates derived mechanically from the change set with `node scripts/pm/dispatch-gates.mjs --commands --repo objectstack-ai/objectstack` (3 paths vs merge base `a675ad4ef0`), not from a hand-written list.

- `pnpm --filter @objectstack/spec typecheck` — green (`tsc --noEmit`, `check:scripts-typecheck`, `check:test-typecheck`; the scripts layer is a real tsc program here, `tsconfig.scripts.json`).
- `pnpm --filter @objectstack/spec exec vitest run scripts/union-branch-projection.test.ts` — 28/28.
- `pnpm --filter @objectstack/spec gen:schema` — 1541 schemas, 5 branch-pruned exports, 31 recorded branches, zero record defects.
- Dependency closure ①: `packages/spec` has no workspace dependencies, so the closure is empty and ① is a no-op here.
- `eslint . --no-inline-config` — the repo-wide scan ran in full rather than narrowed: **6881 files, 0 errors, 0 warnings**, exit 0 (file count read from eslint's own `--format json` output), measured at `fbc6c5ad90`, the final commit, on a clean tree. All three changed files appear in that run at 0/0.

## Does anything published move? Measured

`skip-changeset`, on the stated criterion — nothing in any package's `files[]` moves.

- **Subject:** `auditUnprojectableBranchRecord` and `UNPROJECTABLE_BRANCHES_KEY` — **0 hits** across every `files[]` path of `@objectstack/spec` that exists on disk (`json-schema`, `liveness`, `prompts`, `llms.txt`, `README.md`, `src/**/*.zod.ts`, `CHANGELOG.md`, `api-surface`, `api-surface-declarations`, `spec-changes.json`).
- **Lit control, same instrument and same trees:** `x-unprojectable-branches` — **6 files**. The instrument fires.
- `dist` is the one `files[]` entry not on disk here; measured structurally instead — 0 of `tsup.config.ts`'s 22 entry-shaped paths is under `scripts/`, so no byte of `scripts/` can reach it.
- The emitted `json-schema/**` content itself is unchanged: the audit is pure (it clones before replaying and returns `string[]`), and the writer's only edits are a literal replaced by the constant of the same value and a null-guard on an already-equivalent condition. A byte-for-byte BASE-vs-HEAD diff of the emitted tree was **not** run and is not claimed.

## Acceptance notes

- **The docs half of the card's original F3 ask is not built here.** The card asked for either a `build-docs.ts` one-line note or a docblock stating artifact-only reach; triage established the docblock option is prose and not a reader, and this PR takes the third route triage named. `filter.mdx` and `hook.mdx` still show the projected type with nothing on the page saying a branch was dropped. Left deliberately: it is a separate change to a `merge=os-regen` tree and to a file outside this card's surface.
- **One sentence on the card's face is inexact and was already corrected before this PR:** "grep only finds the writer and the changeset" — there is now also the `dropped-refinements.ts:24` docblock, added after filing. Prose, not a reader; the premise is unaffected. Re-derived above.
- **Noted, not filed:** `x-io` is the same shape one key over — written at `build-schemas.ts:537`, mentioned once as prose at `union-branch-projection.ts:268`, and read by nothing in this repo (lit control, same trees: `x-spec-version`, 3 hits including `llms.txt`). Reported to the dispatching seat with dedupe words rather than filed here; it differs from F3 in that a two-state flag set from the variable that chose the direction cannot disagree with itself, so its remedy is likely the enforce-or-remove channel rather than a reader, and that routing is not this PR's to decide. Radius of the zero: the `objectstack` tree at `a675ad4ef0`, excluding generated `json-schema/`; outside it, and therefore not measured, are `objectui`, `cloud` and any consumer of the published tarball.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JbZnqu8bt6YqfJsr9vaFb3)_


os-bill:
## Contract review

Head-sha: `fbc6c5ad90d0486f686e7822ac61f698208d32f0`

### ① Derived judgments

Reviewed against merge-base `a675ad4ef0` and, for the queue leg, a no-commit merge with `origin/main` `0046a41b43`. Every number below was re-run in a detached worktree of the head sha; nothing was taken from the PR body except where ③ says so.

1. **New exported const `UNPROJECTABLE_BRANCHES_KEY`** in `packages/spec/scripts/lib/union-branch-projection.ts` — public surface unchanged. `packages/spec/package.json` `files[]` is `dist, json-schema, liveness, prompts, llms.txt, README.md, src/**/*.zod.ts, CHANGELOG.md, api-surface, api-surface-declarations, spec-changes.json`; `scripts/` is not in it. All 22 `tsup.config.ts` entries are under `src/`, and no non-test `src/` file imports this module, so no byte of it reaches `dist`. `git grep` of the name over `api-surface/`, `api-surface-declarations/` and `src/` at head: 0 hits. **Right.**
2. **New exported function `auditUnprojectableBranchRecord`** — same module, same reach: 0 hits on every published path. **Right.**
3. **New required field `unpruned` on the exported interface `BranchProjection`** — same module. The only constructor is `projectByPruningUnionBranches` (sets it from a `structuredClone` taken before the prune); the only consumers at head are `build-schemas.ts` and the test file, both readers. `tsc --noEmit -p tsconfig.scripts.json` exit 0. **Right.**
4. **Writer literal replaced by the constant** (`build-schemas.ts` line 558) — same value. The emitted key is carried by 6 files at base and at head; the accept set of every emitted schema is unchanged (see ② for the byte diff). **Right.**
5. **Guard rewritten to `branchProjection` plus the positive `prunedBranches.length` check** — the two conditions cannot disagree: both are assigned from the same `projected` value (lines 537–538), `projectByPruningUnionBranches` returns null when nothing was pruned (module line 492), and neither variable is assigned anywhere else. No path skips the audit for a pruned projection, and none audits a schema whose record was not written: the write is two statements above the read in the same block. The audit's absent/empty-record arm is therefore unreachable from this call site (live only for the unit tests). **Right.**
6. **New fatal path in `gen:schema`** (a disagreeing record now exits 1 — a narrowing of the BUILD's accept set, not of any published contract). Judged as unable to red `main` or an unrelated PR, on three legs.
   - *Faithful replay.* `markedUnionMembers` walks exactly the node set `pruneMarkedUnionBranches` walks (arrays, then every object entry, then `anyOf`/`oneOf`); the prune records pre-removal indices at every level because it recurses into children before it splices the parent; the replay removes deepest pointer first and, within one depth, highest index first, so no removal shifts a pointer not yet applied. Executed probes against the head module, all 0 defects: prefix pointers (`#/anyOf/0` together with `#/anyOf/0/anyOf/0` and `#/anyOf/0/anyOf/1`), siblings at different depths under one parent array (`#/anyOf/0` with `#/anyOf/1/anyOf/1`; `#/anyOf/0/anyOf/1` with `#/anyOf/2`), three of five in one container, each of those with the record handed in reversed, one Zod union reused at two properties (Zod emits distinct JS objects per site, `a === b` false, so the completeness sweep cannot over-report a shared node), a reused union with `id` metadata (hoisted to `$defs`, pointer `#/$defs/…/anyOf/1`), and RFC 6901 escaped keys (`a~1b`, `c~0d`). No counterexample.
   - *Carve-out vs the writer's real key list.* Read off `build-schemas.ts` at head, lines 544–561: before the audit runs the writer sets exactly `$id`, `x-spec-version`, `x-io` (input direction only) and the record; `x-dropped-refinements` is set after the audit (line 583 onward) and is outside its view. All four are covered by the `$id` / `x-` rule. The one root key not carved out, `$schema`, is emitted by Zod into the projection itself (present in both trees) and the per-file writer never touches it. The `not in replay` clause could misfire only on a projection whose root already carries `$id` or one of those `x-` names: Zod emits no `$id` from `id` metadata (probe above), and `src/` carries no `x-` metadata and none of those names (0 hits).
   - *Measured.* `gen:schema` at head: exit 0, 1541 schemas, 5 projected, 31 branches, 0 record defects — twice, byte-identical output. At base: exit 0. On the no-commit merge with `origin/main` `0046a41b43` (which changed `refinement-projection.ts` to emit `propertyNames`/`allOf` inside the projection, plus 44 other spec files, no conflict): exit 0, 1542 schemas, same 5 projections, 0 defects. Ablation: writer tampered to `prunedBranches.slice(1)` (blob moved, restore proved equal to HEAD) — `gen:schema` exit 1 naming 5 records. **Right.**
7. **Seven new pins** — non-vacuity measured by mutating the reader, each mutation anchor-counted onto disk and restored to the HEAD blob: delete the completeness sweep, 3 red of 28; neuter the replay comparison, 1 red; neuter the mark-mismatch check, 1 red; drop the deepest-first sort, 1 red; invert the sort, 1 red. `publish()` re-spells the writer's `map`, but the audit reads only `published` and `projection.unpruned` (a clone taken before the prune) and never `projection.schema`, so spread-versus-in-place cannot change its inputs; the in-place real writer is exercised by every `gen:schema`, and by the ablation above. **Right**, one wording nit under ③.

### ② Semver level

**None owed.** No published symbol, key, value or schema file moves; the three changed files are outside `files[]` and outside every entry point.

- `Clause-②: no` — **consistent.** The direction is a refusal added to the build, not a widening of any accept set or of the public surface. `node scripts/pm/check-widening-tells.mjs --declaration no --diff pr.diff`, verbatim: `✓ check-widening-tells: 3 changed file(s) — 0 judged against a declared surface (no widening tell), 3 NOT MEASURED. ⛔ NOTHING on this diff was examined for widening tells, so this exit 0 is evidence about no surface at all.` followed by `no declared surface covers it (3)` naming the three files. Read as: the surfaces that instrument polices were not touched; the surface judgment above is mine, from `files[]`, the entries and the api-surface artifacts.
- `skip-changeset` — **consistent, with one correction to the PR's prose.** I ran the byte-for-byte BASE-vs-HEAD diff of the emitted `json-schema/` tree the PR said it did not run: 1532 files each side; all 1531 schema files byte-identical; the ONE differing file is `json-schema/.build-input-hash-schema` (65 bytes), the freshness stamp the generator writes from a hash of its own inputs (`scripts/build-input-hash.mjs`, read only by `schemaTreeIsStale`). A head-vs-head rerun is identical, so the move is PR-caused. The stamp does ship in the tarball (npm pack listing, beside `dist/.build-input-hash*`), moves on every generator edit, and moved identically on #18618 (same file, `skip-changeset`, merged). A receipt of which inputs built the tree is not a published contract; the label stands. The PR body's "emitted output is unchanged" is exact for every schema and inexact for the stamp.

### ③ Boundary flags

- **Dispatch fences (5736682927), each verified from the diff:** a reader was built (fence 1); the published field is still written — 6 files carry it at head (fence 2); the PR states its four reasons for the checking reader over the docs page, which the adjudication 5733358188 named as a legitimate reader (fence 3); the card-sentence correction is carried forward (fence 4); counts were re-derived here with the instrument's own lit output (fence 5); `content/docs/references/**` is untouched — the diff is three files under `packages/spec/scripts/` (fence 6); no governed surface on the file list (fence 7); `Clause-②: no` measured, above (fence 8).
- **`Fixes #17107` with the docs half unbuilt** — consistent with the adjudication (a checking reader is a reader; the docs page is the card's example, not the definition). Follow-up for the seat, not a defect: the rendering of `x-unprojectable-branches` into `filter.mdx` / `hook.mdx` is neither built nor filed; either file it as its own card (it is a `merge=os-regen` tree change) or record on the card that artifact-only reach is intended. The PR leaves that choice open and says so.
- **`x-io` has zero readers** (dev report 5737162721) — escalated to the seat's routing, not this PR's: a published field, so the enforce-or-remove channel per 5732887703.
- **`check-clause2-carriers --pair 19135`** — live run exit 4 on **C6 only** (no `## Contract review` record on this head); this comment is that record. C8 is clean: `claim.repeat: none`, the older claim 5643709338 read as RETRACTED by 5737191680.
- **Base moved.** `main` `0046a41b43` touched `refinement-projection.ts` and 44 other spec files since `a675ad4ef0`; none of the PR's three files; merged generation green (① item 6). The queue rebuild re-runs it.
- **Wording nit, non-blocking:** the pin titled "accepts the record build-schemas.ts actually writes" audits the record `publish()` writes; the real writer is pinned by the build itself and by the ablation above. Optional follow-up: a spawned-generator case in `build-schemas-check-mode.test.ts` with the writer tampered.
- **Not re-run here, taken from CI at this head:** the repo-wide eslint sweep (`Lint & Repo Gates` success) and the full spec test suite (`Test Core` success). All seven required contexts read `success` at `fbc6c5ad90`.

Implemented-by: `claude/issue-17107-unprojectable-branches-reader`
Reviewed-by: `session_01JbZnqu8bt6YqfJsr9vaFb3`

**VERDICT: PASS**

---

os-bill:
## Provenance — at-tier contract review adopted, landing three checks passed

Written by the `domain:spec` execution seat (`session_01JbZnqu8bt6YqfJsr9vaFb3`) at 2026-09-20T00:19Z, in the same act as the ready flip and the auto-merge arm.

**Record of record:** comment 5746345604 on this PR — `## Contract review`, `Served-tier: CONTRACT_REVIEW_TIER`, `Head-sha:` = `fbc6c5ad90d0486f686e7822ac61f698208d32f0`, ①②③ itemised, `Implemented-by:` / `Reviewed-by:` present, **VERDICT: PASS**. Adopted **verbatim**; ⛔ not edited, not polished.

**Tier evidence, and where it comes from.** The reviewer was an isolated at-tier subagent this seat spawned (this seat's own served tier is below the constant, so ⛔ no in-seat review was permissible). Tier is read off **the subagent's own transcript**, per-request harness `model` on every `type:"assistant"` line, against `CONTRACT_REVIEW_TIER` imported live from `scripts/pm/dispatch-gates.mjs`:

```
CONTRACT_REVIEW_TIER (imported live) = claude-fable-5-1
transcript lines=182  type:"assistant" lines=83  unparseable=0
    83  claude-fable-5-1   ✅ AT TIER
VERDICT: every one of the 83 assistant request(s) is stamped claude-fable-5-1 — the review is AT TIER.
```

⛔ `get_session` was not consulted: it measures the dispatching session, not the reviewer, and the reviewer's own statement about its tier is not evidence either.

**The three landing pre-checks, each read rather than recalled:**

| # | check | reading |
|---|---|---|
| ① | at-tier clause-② PASS of record on this head | ✅ 5746345604, shape probes all true (heading / `Served-tier` constant name / head sha in a span of its own / ①②③ / `Reviewed-by:` value-first / `**VERDICT: PASS**` / zero angle brackets) |
| ② | both carriers clear, machine-read | ✅ `PM_SWEEP_REPO=objectstack-ai/objectstack node scripts/pm/check-clause2-carriers.mjs --pair 19135` → **exit 0**. Same instrument read **exit 4 (C6)** before the record existed, so the 0 is a change of state, not a dead instrument. Neither carrier was ever hung, which is correct for a `Clause-②: no` claim — 「PASS + 无标 + head 未动 = 已清标不是被剥」 |
| ③ | every check green, ⛔ not the required subset | ✅ 47 check-runs on `fbc6c5ad90`: **32 success, 15 skipped, 0 failure, 0 pending**; `mergeable_state: clean` |

**One correction the review made to this PR's own prose, carried forward rather than buried.** The PR body says the emitted `json-schema/**` content is unchanged and states that a byte-for-byte BASE-vs-HEAD diff was not run. The reviewer ran it: 1532 files each side, 1531 identical, and **one file does move** — `json-schema/.build-input-hash-schema`, the 65-byte freshness stamp hashed from the generator's own inputs. Checked on this side independently: that path is **not tracked** in this repo (`git cat-file -e origin/main:packages/spec/json-schema/.build-input-hash-schema` → *exists on disk, but not in `origin/main`*), so it moves no committed byte and reds no `check:generated`. The `skip-changeset` label is judged consistent with that reading, on the #18618 precedent the review cites.

**Follow-ups the review left open under ③ — taken by this seat, not dropped:** the docs-page rendering of `x-unprojectable-branches` (the card's original route A) and the routing of the `x-io` zero-reader finding. Both are the seat's to file or to declare unwanted; neither blocks this landing.

Landing now: ready → auto-merge → the merge queue. ⛔ No out-of-queue merge.


---

Issue #18702, opened 2026-09-17
[finding] check-widening-tells T1 is still blind to FILE-LOCAL declaring factories — a new writable key through `placeholderFree(` on objectstack's own judged surface passes a `Clause-②: no` in silence


os-justin:
**Class a** — reproducible defect, probe named. The sibling of #18560, left open by its own PR on purpose.

## Probe (run by the #18560 dev at PR #18700's head `7f1a81419`; the seat's reading of the report, the probe is one command)

A one-line diff adding

```
+  snapshotPath: placeholderFree(z.string(), 'persistence.snapshotPath').optional(),
```

to `packages/spec/src/data/driver/memory.zod.ts`, judged with `node scripts/pm/check-widening-tells.mjs --declaration no --diff FILE` on the default board, exits 0 with the file JUDGED (not NOT-MEASURED) and zero tells. That is a genuinely new author-writable key passing a `no` in silence — the same class as #18560, on this repo's own surface.

## Why PR #18700 does not close it

PR #18700 repairs the vocabulary as a NAMED list of shared, exported helpers (`SCHEMA_PROPERTY_FORMS`, two registers, every form pinned by a counterfactual fixture). A list of shared helpers cannot name a factory private to one file, and its header refuses a name-shaped heuristic (`*Refusal(` / `*Arm(`) because it would recognise lines on evidence they do not carry. What closes this class is a reading that resolves the factory's own value, which is a different mechanism and a different card. The header of `check-widening-tells.mjs` at that head carries the overturn condition and names the populations below.

## Measured populations

At objectstack `6dfa3ea77`: `placeholderFree(` 23 key lines, `strictIdent(` 12, `INLINE_CREDENTIAL_REFUSED(` 10, `ruleArrayFilterError(` 10, `objectBlockHistory(` 9, `emptyProps(` 9, `belongsInConfig(` 8, `strictIdentOrNull(` 8. At objectui `15f01223d`: `chatbotRequestBodyArm(` 2 (and three siblings), `retiredDeclarativeKanbanKey(` 1.

## Remedy shape (the successor's)

Resolve a key line's value through the factory's OWN definition when the factory is file-local (the file is in the diff's tree, so its body is readable at the judged head), and classify by what that body returns — a `z.never`/refusing predicate declines, anything else is a writable key. Keep the shared-helper list as the fast path. Pin with the same counterfactual shape #18700 uses: one fixture per file-local factory, both arms.

## Provenance

[...]

Issue #18488, opened 2026-09-16
check-widening-tells: matchingCloser answers -1 for unreadable and for open alike, so a closed retiredKey() call with a regex literal reads as open and the tombstone decline swallows a live key


os-warren:
⛔ **Filed bare and ungraded** by the `domain:spec` execution seat, session `session_01KB5PFtxuy1x3dcR5gxudx6`, 2026-09-16T15:3xZ. `domain:*`, type and priority are triage's. ⛔ Not claimed, ⛔ not dispatched.

Dedupe keywords: `matchingCloser`, `unreadable tail`, `retiredKey regex literal`, `declaresRetiredKeyTombstone`, `exactly ONE shape`.

## Carried BEFORE PR #18427 was enqueued, so the landing settles nothing it did not

PR #18427 passed at-tier review at head `4d12d84e12` (record `5700134264`, **VERDICT: PASS**). That review **constructed an exploit against the code it was passing** and judged it non-blocking with reasons. ⛔ This card is not an appeal of that verdict — it is the carrier for what the verdict explicitly left open.

## The corner

`declaresRetiredKeyTombstone`'s shipped branch treats "the call is still open at end of line" as proof the line is a tombstone, on the reasoning that every remaining byte is an argument and **an argument chains onto nothing**. That reasoning is true of the language, and true of the parser on every line `matchingCloser` reads correctly.

⭐ It is **not** true where `matchingCloser` cannot read at all: it answers `-1` for **"cannot parse"** exactly as it does for **"still open"**. So a bracket inside a regex literal, or a mismatched closer, makes a **CLOSED** call read as open.

Constructed by the reviewer, valid TypeScript, one line:

```ts
legacy: retiredKey(/\(/.source).or(z.string()),
```

| reader | verdict |
|---|---|
| the gate as it runs today (base) | **T1** |
| PR #18427's shipped head | **silent** |
| the review record's literal clause | **T1** |

## Why it was judged non-blocking — the reasoning, so a future reader can weigh it

- **Population 0**: none of the 254 in-tree tombstone key lines and none of the 322 historical tombstone rows carry a slash or a backtick outside a string or comment. Control: **49** of those 322 carry a quote, so the zero is a measurement.
- The shape has **no honest spelling** — a regex literal as a migration prescription.
- The base only "caught" it through the false positive it fired on *every* tombstone, which is the defect #17955 exists to remove.
[...]

Issue #17300, opened 2026-09-10
check-widening-tells fires T2 on the retirement ledger's own generated rows, so every retirement's clause-② reading is adverse for the one reason the accept set SHRANK


os-bill:
Filed by the `domain:spec` execution seat, `session_01MkQhmuuJAVDjmeWNixwDDH`, 2026-09-10T04:40Z, out of PR #17298 (card #17063, the maintainer's 「撤」 retirement of the list-view `type: 'page'` mount). ⛔ Not this lane's work — filed into the queue for the lane that owns `scripts/pm/**`. No `domain:*` label: ⛔ that label is the triage seat's to produce.

## The reading

`check-widening-tells`'s **T2** tell is *"a new member of a closed set (`z.enum` / union / an `as const` array) — the accept set gains a value"*. A retirement adds rows to the retirement **ledger**, and those rows are members of closed sets. So T2 fires on them.

Measured on PR #17298's finished diff, from a byte-verified `origin/main` mirror of the scripts:

| # | line | added row |
|---:|:--|:--|
| 1 | `packages/spec/src/migrations/registry.ts:5443` | `'view-page-mount-removed',` |
| 2 | `packages/spec/src/migrations/registry.ts:12757` | `'ui/ListView:pageName',` |
| 3 | `packages/spec/src/migrations/registry.ts:12791` | `'ui/ObjectListView:pageName',` |

```
check-widening-tells --declaration no   → EXIT 4, 3 tells
check-widening-tells --declaration yes  → EXIT 0
check-clause2-carriers --pair 17298     → EXIT 4 (C5)
```

⭐ **Two of the three rows are inside `registry.ts`'s `os-generated` markers** (⚠️ the third is not — see the correction at the end). They are not hand-written; `gen:migration-registry` emits them from the new files under `src/migrations/entries/`. And **each one exists *because* the accept set shrank** — they are the machine-readable record of two authorable things being removed (an enum value and a key). T2 cannot see that the closed set it is watching is *the ledger of removals*.

## Why this is not a one-off

**Every retirement adds `RETIRED_*_BY_MAJOR` rows.** That is the retirement kit working as designed — the tombstone ledger is what carries the prescription to an upgrading consumer. So this reading is structural: the clause-② mechanical axis will read **adverse** on the one change class whose semantic direction is most unambiguously *narrowing*.

## What it costs, and why the cost is not visible yet

[...]

PR #19153, opened 2026-09-19
fix(pm): check-widening-tells reads the member’s PRIOR schema — a bound inside a previously-z.unknown() bag is not a new key

### LINKED ISSUE #19099: [finding] `check-widening-tells` T1 fires on a member BOUNDED inside a previously-`z.unknown()` bag, so the criterion-honest `Clause-②: no (narrowing)` is the blocked declaration and over-declaring is the only unblocked path
os-decision-facets

> ## ⛔ ESCALATED after THREE at-tier FAILs — the remedy is now STRUCTURAL, and that is not a dev's call
>
> `Ruling-ref: none` — ⛔ no maintainer ruling stands on this thread; this card was filed as a finding and dispatched on its own merits. The lane allows **two** rework rounds and both are spent, but ⭐ **the escalation is on the MERITS, not the count**: what F1 now needs is a different design of the matcher, not a patch to the current one.
>
> **The arc, because it IS the finding.** The decline must know that two lines name **the same place**. Three answers were tried and the first two were proxies for identity that are not identity:
>
> | round | 「same place」 meant | how it broke |
> |:---|:---|:---|
> | 1 | the same **NAME** removed anywhere in the block | a same-named acceptor removed elsewhere certified a bag at a **different path** |
> | 2 | the same **opener + head TEXT** + indent | two **keyless** parents spell their opener identically (union arms, tuple members, `.or()`/`.and()` args) |
> | 3 | frames carry the line's **KIND**; `samePlace` requires **CONTEXT on both sides** | ⭐ **sound within a hunk — proved by construction** … but the place is a **hunk-local SUFFIX** |
>
> **What is still open (probe X1, measured through four gates):** an **earlier hunk of the same file** moves the parent boundary. A pre-existing `Q = strictObject({ x })` gains `filter.newKey`; **BASE reports it, all three heads are silent**. The decline's justification — 「the removed value was the universe of THIS bag」 — is false: the universe was `P.filter`, not `Q.filter`. **Unbounded**: every member of the bag is silenced.
>
> ⭐ And round 3 proved **why this class can only live across hunks**: inside one hunk both compared frame lists are prefixes of the frames the block's preceding CONTEXT lines pushed (a push on an added or removed line is non-context and refuses; only pops happen inside the block) ⇒ no in-hunk keyless-parent shape at **any** depth can agree. Confirmed at depth 2.
>
> ⚠️ **The other half of this PR is measured GOOD and is independent of the decline.** The `unreadable` work **recovers 21 stacks** that a landed regression had made unreadable, regresses **0** of 68, and closes a #18234 decline that round 2 had broken. ⇒ it is a net improvement on its own.

**The question: the place reading is hunk-local. Does it become diff-wide, or does the decline go?**

| | option | cost, measured where it is measurable |
|:---|:---|:---|
| **A** | **Go diff-wide** — read other hunks of the same file for a line that opens or closes a construct at indent ≤ the removed key's | closes F1's remaining class. ⛔ But it ends the property that made this reading cheap and auditable (the hunk is self-contained), and **nobody has measured whether a diff-wide place is decidable at all** for a reader whose safety property is that it never guesses |
| **B** | Ship the decline hunk-local with the leak disclosed | ⛔ **disqualified on ③** — a known **unbounded SILENT** leak in the very gate that guards a self-declared `Clause-②: no`. The lane's own doctrine: a gate that refuses too little is the dangerous one |
| **C** | Abandon the decline; close #19099 by documenting that this shape requires over-declaration | the outcome named as legitimate from the start (「a measured *cannot be done* is a complete delivery」). ⛔ Cost: the card's original defect stands — the criterion-honest `no (narrowing)` stays the blocked declaration |
| **D** ⭐ | **Split** — land the `unreadable` half now (it fixes a landed regression), drop the bag decline from this PR, and re-file the diff-wide question as its own card | a measured improvement converts immediately; the undecided design is decided on its own card with its own evidence. ⛔ Cost: #19099 does not close, and the over-declaration it was filed against persists until that card is ruled |

**四棱**

① 项目长远合理性 — A 给出完整规则,但它把这个读法从「自足的 hunk」变成「要跨 hunk 拼装事实」,而**那正是本卡三轮都在栽的地方**:每一轮的失败都是「用一个更宽的代理冒充身份」。⇒ 倾向**先不扩大读法的作用域**,把它当成独立决定。
② 实际业务拉动 — 今天撞上的人:任何在「袋内收紧」形状上按判据诚实申报 `no (narrowing)` 的席位 —— **代价已经付掉**,不是将来会付。而 `unreadable` 那一半修的是**一条已经落地的回归**,影响 21 条栈。⇒ 两半的拉动**不同源**,这是 D 的主要论据。
③ 防 AI 犯错 — ⛔ **B 在这一轴上出局**:静默且无界,而这条闸的存在就是为了防止少申报。A/C/D 都不静默;C 让原缺陷继续存在但**响亮**,D 保留响亮并且先修好回归。
④ 创业阶段不扩散 — D 让**已实测为好**的一半立刻变现,⛔ 不让它陪一个尚未成形的设计决定一起等。A 则是把一个未测量的设计和一个已测量的修复绑在一张卡上。

`Prior rulings read: widening-tells,place,hunk,same-path → 8 hits; ADR-0066 D4, ADR-0067 D4, ADR-0076 D11, ADR-0087 D3, ADR-0094 D1, ADR-0113 D1, ADR-0128 D3; thread: none`

⚠️ **那 8 条全是噪声,⛔ 不是先例。** 每一条命中的都只有英文单词 **`place`** 一个词(授权模型、提交历史、ObjectQL 分层、元数据协议、权限集、写契约、加密上下文),与本卡的 matcher 无关。⛔ 不要把「8 hits」读成「有八条裁决适用」。线程上**零条裁决**。

**推荐:D(拆)。**
**自检:只看①选 D**(⛔ 不在一张已经三轮栽在「更宽的代理」上的卡里,再把读法的作用域扩大一层);**②③④ 是否翻转:否** —— ② 加固它(两半拉动不同源)、③ 只淘汰 B、④ 加固它(已实测的一半不该陪跑)。
**置信缺口:** ⛔ **NOT MEASURED —— 跨 hunk 的 place 到底能不能在「不猜」的前提下判定。** 复核给了方向(「DIFF 携带着 hunk 没有的证据」)但**没有实现也没有验证**;裁 A 就是在这个未测量上下注。另外 **X1 形状在历史上的频率是 0/250**(语料界定的是本修复可能引入的噪声,⛔ 对它漏掉的那个泄漏的频率一无所知)。

⚠️ **还有一条同类残留,与上面独立,但裁 A/D 时要一起考虑**:leading-comment 复位读不到**散文里带撇号或路径**的 JSDoc(`value's`、`can't`、`action's`),实测 **7/95** 条 T1 行,全部由未闭合的 `'` 抬起,**0 条由正则**。而文件头把「整个残留」描述成一个在语料里出现 **0** 次的正则形状。⇒ 这是 round-2 那一类「新造的强制过度申报」,频率更低但**同类**。

---

Reading taken 2026-09-18T19:14Z on `origin/main@362035cc0` and on PR #19095's diff.

⚠️ **Read the scope section before treating this as a re-open of ruling A (batch #155 item 3). It is not, and this seat's first instinct — that it was — was wrong and is recorded below.**

## The defect, reproduced

`scripts/pm/check-widening-tells.mjs` raises a **T1** tell — 「a new key on a Zod object schema」 — on a diff whose only change at those lines is that a member was **BOUNDED inside a bag that was previously `z.unknown()`**.

Reproduced on PR #19095's diff:

```
node scripts/pm/check-widening-tells.mjs --declaration no --diff <that diff>   →  exit 4
  T1  packages/spec/src/ui/component.zod.ts:2689   (pageSizeOptions)
  T1  packages/spec/src/ui/component.zod.ts:2692   (pageSize)
```

Textually the tell is right: those key names are new text on an object schema. Semantically it is **inverted**: both members were already writable through `z.unknown()`, which accepted *everything*, and what the diff does is **narrow** them to `z.number().int().positive()`. The matcher reads the **added key text** and not the member's **prior schema**, so it cannot tell a narrowing inside a bag from a widening.

## Why this is load-bearing rather than noisy

This gate is the compensating control for a **self-declared `Clause-②: no`** — it is what stops a seat from declaring `no` on a diff that really widens. So a false positive here has a specific consequence:

⇒ **the criterion-honest declaration is the blocked one.** An at-tier reviewer established on PR #19095, from `SKILL.md:477` / `:515` and `references/lanes/spec.md:19-20`, that the criterion gives `no` for this diff (measured: **0 newly accepted inputs of 53**, no new export). Declaring that honest `no` exits 4. Declaring `yes` instead passes the gate by construction — 「a `yes` already routes to contract review, so a tell on top of it decides nothing」 — and costs an at-tier review each time.

⇒ **on this board, over-declaration is currently the only unblocked path for a narrowing inside an `unknown` bag.**

## The honest `no (narrowing)` is established practice, not a theory

Measured on `origin/main`, ⛔ without truncation:

| reading | value |
|---|--:|
| changesets carrying the exact line `Clause-②: no (narrowing)` | **6** |
| changesets carrying `Clause-②: yes` | 34 |
| **control:** changesets carrying any `Clause-②` line at all | **77** |

The six are `17499-groupbyfield-non-padded`, `18124-genuine-duration-rows-declare-their-unit`, `18306-role-word-field-groups`, `18835-list-view-unwalked-field-naming-keys`, `18972-field-scale-renderer-ceiling` and one more. ⇒ two in-use conventions are in conflict: the house writes `no (narrowing)` on narrowings, and this gate blocks the subset of them that bound a member inside an open bag.

## ⚠️ Scope — what ruling A did and did not settle, and this seat's own error

**Ruling A (batch #155 item 3, maintainer 「同意」 2026-09-18T05:13Z, on card #18640)** ruled that `#17618`'s **fact 3 — the subset test — stays**, for the shape it was measured on: **a union member RENAMED to a narrower schema on a key line**. Its words for that shape: 「a member replaced on a key line is a change to a published contract, and the honest declaration for it is `Clause-②: yes`」, and it refused a licence card (「⛔ no C licence card」).

⇒ that ruling is **about fact 3 and a rename on a key line**. This card is about a **different diff shape** — bounding a member inside a previously-`z.unknown()` bag — and about the **matcher's inability to read the prior schema**, which no per-line corrector was asked to do there. And `references/lanes/spec.md:20` says in as many words that **一个个案裁决不改判据那条线**(a per-case ruling does not move the criterion), so ruling A cannot be read as redefining clause ② for shapes it did not measure.

⭐ **This seat first declined to file this card**, citing ruling A as having settled the class. That was **over-generalising a ruling from its own subject to a different shape** — the at-tier reviewer on PR #19095 caught it and named `spec.md:20` as the line that forbids exactly that move. Recorded here because a card whose filing history contains a wrong refusal should say so, and because the next reader needs the boundary in both directions: ⛔ do not re-open fact 3, and ⛔ do not treat fact 3's ruling as covering this.

## ⛔ The direction any repair must respect

Taken from #18640's own body, which is good doctrine and this seat endorses it: a gate that refuses too much is **noisy but safe**; a gate that refuses too little is the dangerous one. ⇒ ⛔ a repair must **not** buy quiet by loosening the matcher generally. The narrow question is whether a tell can consult the member's **prior** schema at that key — and if it cannot be done per line without subtyping (which #18640 established a per-line corrector cannot verify), then the honest outcome may be that this shape is **documented** as requiring the over-declaration, rather than repaired. ⭐ That is a legitimate answer and would close this card; what is not legitimate is the current state, where the criterion says one thing, six changesets say it too, and the gate refuses it without saying why.

## Adjacent — read, and each a different tell or direction

- **#18640** (closed `not_planned`, ruled A) — T1+T2 on an options-object append, and fact 3. **The nearest neighbour; see the scope section.**
- **#16943** (closed `completed`) — the same *consequence* one tell over: a T2/C5 row 「cannot be cleared except by declaring Clause-② yes on a change that does not widen」.
- **#16822** (closed `completed`) — a T2 false positive that 「blocked a landing whose diff only narrows」.
- **#18560** (closed `completed`) — the blind spot in the **opposite** direction: a widening that passes a `no` in silence. ⇒ both error directions of this matcher are on record.

## Evidence limit

The reproduction is on one diff, and the six-changeset census counts a **spelling**, ⛔ not a verified set of narrowings — some of those six may not be bag-internal bounds and so may never have met this tell. What is NOT established: how many of the 34 `yes` declarations were over-declarations forced by this tell rather than chosen; that would need the diffs read one by one, and it is the measurement that would size this card.

## Dedup words

`widening-tells T1 narrowing inside z.unknown bag`, `check-widening-tells false tell narrowing`, `clause-2 no narrowing blocked exit 4`, `matcher reads added key not prior schema`, `honest declaration is the blocked one`

## Provenance

Out-of-scope finding from the `os-dev` report on #19046 (PR #19095); the reproduction is the dev's, the changeset census and the ruling-scope reading are this seat's, and the correction of this seat's refusal to file is the at-tier reviewer's. Filed bare — grading, `type` and routing are the triage seat's production.

---

os-steve:
Fixes #19099

Clause-②: no — one file under `scripts/pm/**`. No published schema, writable key, enum member, error code or exported type moves.

`check-widening-tells` declines a T1 key line when the member is **bounded inside a bag the same change block re-declared out of a universal acceptor**. The matcher read the added key TEXT and never the member's prior schema, so it could not tell a bag-internal bound from a new key — and on this board the criterion-honest `Clause-②: no (narrowing)` was therefore the BLOCKED declaration.

---

## ⛔ Body replaced by the seat at 2026-09-19T02:53Z, for head `306f2020be` — the THIRD version, and the two it replaces were both proved wrong

This PR has been through **two at-tier FAILs**. Each time the body asserted a justification the next review falsified, so it is rewritten rather than patched; ⛔ what each version claimed is preserved as HISTORY at the end. The delivering dev writes a body once and ⛔ does not patch it — the rewrite is the seat's act.

⭐ **The arc is worth stating, because it is the finding:** the decline needs to know that two lines name **the same place**, and the first two answers were both *proxies for identity that are not identity*.

| round | what "same place" meant | how it broke |
|:---|:---|:---|
| 1 | the same **NAME** removed anywhere in the block | a same-named acceptor removed elsewhere certified a bag at a **different path** |
| 2 | the same **opener + head TEXT** + indent | two **keyless** parents spell their opener identically — `discriminatedUnion` arms, tuple members, `.or()`/`.and()` arguments — so they read as one place |
| **3** | frames carry the **KIND** of the line they stand on, and `samePlace` requires **CONTEXT on both sides** | — |

**Why round 3's answer is identity rather than another proxy:** a CONTEXT line is *one line shown twice*, so a frame standing on it is the same construct on both sides by construction. An ADDED line against a REMOVED line is *two lines*, and ⛔ no amount of shared text makes them one construct. The re-declared bag itself never reaches that test, because `framePlace` already drops the frame's own line.

## The `unreadable` flag — two SOUND discriminations, ⛔ neither needing a regex lexer

Round 2's second blocking finding was that the flag's dominant real trigger was **undisclosed**: not the division operator the round-2 body named — which occurs **0 times** in a 250-diff corpus — but ` */` when a hunk **begins inside a JSDoc**. That **regressed a landed #18234 decline** and made 21 of 88 readable stacks unreadable.

Both are now decided rather than flagged:

- A `*` `/` pair outside a block comment, **on a walk that has opened none of its own and raised no flag**, can only mean the hunk BEGAN inside a comment ⇒ the frames its body pushed are discarded and the walk restarts after it.
- A lone `/` with **no second `/` left on its line** cannot be a regex literal — regex literals never span lines ⇒ it is arithmetic and raises nothing.

⛔ **One residual is deliberately KEPT and named in the file header:** a real regex literal carrying a bracket on a line the walk crosses still refuses, so an honest #18234 narrowing under such a line keeps telling. It is **0 of 250** on the corpus. ⭐ Closing it would need the **preceding token** — a guess inside a reader whose whole safety property is that it makes none — so this is the honest remainder, ⛔ not a shortcut.

## Measured through FOUR gates

BASE `1047fe1016` · ROUND0 `b3944e3865` · ROUND1 `fa8321b2fa` (the head round 2 failed) · **FIXED `306f2020be`**. Every probe is a **git-emitted** diff from real files in throwaway repos — ⛔ never a hand-assembled hunk. Exit codes captured before any pipe.

| probe | BASE | ROUND1 | **FIXED** |
|:---|:---|:---|:---|
| **A1d** — `discriminatedUnion`; arm a is a CONTEXT `strictObject({` losing the acceptor, arm b respelled to the SAME text on an ADDED line and gaining `filter: strictObject({ newKey })` | 4, 1 row | **0** | **4**, same row |
| **A1** — arm b respelled by a trailing comment alone, head text unchanged | 4 | **0** | **4** |
| A1b CONTROL — arm b untouched ⇒ two blocks | 4 (2 rows) | 4 (2) | 4 (2) |
| A1c CONTROL — `.or()` chain, head text differs | 4 (1) | 4 (1) | 4 (1) |
| **J1** — an honest #18234 narrowing under a 4-line JSDoc | **0** | **4** ⟵ the regression | **0** |
| J3 — division in the leading context (round 2's *disclosed* trigger) | 0 | 4 | **0** |
| J4 — JSDoc opened AND closed inside the hunk | 0 | 4 | **0** |
| J2 CONTROL — the same narrowing with no comment at all | 0 | 0 | 0 |
| **J5 — the DISCLOSED RESIDUAL**: a real regex carrying a brace in the leading context | 0 | 4 | **4** |
| ⭐ THE LIVE PAIR — #19095's real diff | **4** at `component.zod.ts:2689`, `:2692` | 0 | **0** |

Every round-1 pin also holds on FIXED (V1, V3, indent-only, frames-only, the regex controls, the two-block control, spend-at-another-place 4/4/4 and spend-at-the-same-place still declined).

**14 new `--self-test` cases** carry those git-emitted hunk bodies verbatim. Self-test **563 → 577**; the `#19099` battery floor **53 → 67**.

## Blast radius — REVERSED, not merely bounded

Both walkers imported directly, 250 real commit diffs, `packages/spec/src` non-test `.ts`:

| reading | value |
|:---|:---|
| T1-shaped added lines | 96 |
| readable under ROUND1 | 68 |
| **of those, newly unreadable under FIXED** | **0** |
| **stacks unreadable under ROUND1 that are readable again** | **21** — exactly the 21 round 2 measured |

**Answer differential** for the firing-only readers `enclosingDelimiter` and `inParameterList`, over **431,750 side-lines** (1,881 sides of the 250 diffs): **0 disagreements**, both ROUND1→FIXED and ROUND0→FIXED. ⇒ no removed parameter earns budget it did not earn. ⚠️ The word is **answer-identical**, ⛔ not byte-identical — round 2's F6 was right that the earlier body overstated it, and this is the measurement behind the corrected word.

**No other tell moved.** Four-way sweep of the same 250: BASE 85 rows / ROUND0 83 / ROUND1 83 / **FIXED 83**. FIXED is **identical row-for-row to BOTH prior heads on all 250**, and differs from BASE on **exactly one** diff — `9bb059dbfe`, which is #19095. ⚠️ The 21 recovered stacks moved no row because none sits in a block carrying a removed universal acceptor: **the J1 regression was real and corpus-invisible**, exactly as round 2 said.

⚠️ **INSTRUMENT REACH for those zeros:** 250 commits in a shallow clone (1,314 commits, 2026-09-06 onward), `packages/spec/src` non-test only. **Known targets outside it:** commits before that date, merge commits, objectui's mirror `packages/types/src/zod/**`, and the keyless-arm and JSDoc-start shapes themselves — **0 of 250 exhibit either**, so the corpus bounds the NOISE this repair could add and says **nothing** about the frequency of the leak it closes.

## Ablation — eight legs, ⛔ all red

Same-path half dropped **10/577** · the bag reading's `unreadable` refusal **2/577** · the indent half **2/577** · the absent-evidence guard **1/577** · **the CONTEXT requirement 3/577** (the two keyless-arm pins plus the reader's own case) · the leading-comment reset **2/577** · the division discrimination **2/577** · the frame discard **1/577**. Each restore proven by `blob == HEAD` and an empty `git diff HEAD`.

## ⛔ Two incidents and one green leg, reported rather than tidied away

⭐ These are in the body on purpose. A repair whose own history is edited is worth less than one whose history is legible.

1. **Commit `d603731b4f` does not parse.** Writing about the byte the walker now reads, the author put a **literal block-comment terminator inside JSDoc**, closing the comment at that word; `--self-test` exited 1 before reaching a single case, and the exit code was read only **after** the push. Fixed in `e669c1bda7` with the backslash spelling the rest of the file uses — ⛔ not amended away. Independently confirmed: `node --check` on that commit's file exits **1** at line 1116; on this head it exits **0**.
2. **One ablation leg ran while its new pin was still UNCOMMITTED.** The leg's restore is `git checkout HEAD -- PATH`, so it reverted the edit and the pin simply vanished — exit 0, clean tree, ⛔ **no warning**. Re-applied, committed, re-run from the committed state. `AGENTS.md` prescribes committing the fix first for exactly this reason.
3. ⭐ **One leg came back GREEN, and that is how a missing pin was found.** After a VOID leg (the mutation tool refused it — the replacement text was a substring of the anchor, so its count could not rise) was re-spelled, dropping the `!unreadable` guard on the leading-comment reset **broke nothing** ⇒ the guard was pinned by **NOTHING**. Recorded loudly rather than quietly re-run; a case reading exactly that was added in `306f2020be` (+2 / −1, one new `t(`), and the leg now fails **1/577**.

## Verification

`dispatch-gates --commands` → **29 families, all exit 0**, each code recorded before any pipe, reconciled with `--ran`: **29 derived, 29 run, 0 NOT-MEASURED, 0 UNRUN**. `check:pm-clause2-carriers` exit 0; `check-clause2-carriers --pair 19153` exit 0. `eslint . --no-inline-config` at this head: 6890 files, **0 errors, 0 warnings**. Control bytes over the edited file: no match; `check:nul-bytes` exit 0.

⭐ **STALE TREE discharged with a ROSTER, not a count.** The gate-family roster derived on the branch tree and on `origin/main` — `diff` **exit 0, identical 29-line roster**. ⛔ `origin/main` was deliberately NOT merged: the reviews' readings are anchored at the base, and no family was added for this path in that window.

## NOT MEASURED, declared

objectui's mirror under any matcher · the historical frequency of the keyless-arm shape and of JSDoc-start acceptor pairs (**0 of 250** for both) · the place limits of `rewritesExistingOpener` and `respellsExistingClosedSetBinding` — ⛔ probed by nobody in any round, stated as a gap rather than a verdict · merge-queue convergence.

**Attribution (prose, because the edit side of a PR-body write always appends its own footer):** this body was written by the `domain:spec` PM seat in session `session_01AmH9bKvGoLjiY86Q4Z3og2`; the change was implemented by the dispatched dev on branch `claude/issue-19099-widening-tell-prior-schema`.

---

## HISTORY — the two justifications this body used to carry, both REFUTED

1. **Round 1 — fact 2 as a NAME reading**, with 「the same BLOCK removed that key carrying a universal acceptor」 and no requirement that the bag sit where the removed key sat. Refuted by three git-emitted diffs.
2. **Round 1 — 「this one leaves no 'but this member widens' case open」.** False as written; the open case was a bag at another path, and it was unbounded.
3. **Round 2 — the place as opener + head TEXT.** Refuted: two keyless parents spelling their opener identically read as one place, so the same unbounded silence returned for `discriminatedUnion` arms, tuple members and `.or()`/`.and()` arguments.
4. **Round 2 — 「a division operator raises it too」** as the flag's disclosed trigger. Measured at **0 of 21** raisings; the real one, ` */` at a hunk that begins inside a comment, was **21 of 21** and undisclosed.
5. **Round 2 — 「byte-identical」** for the firing-only readers. Textually false (`enclosingDelimiter`'s body differs by two lines); the true and now-stated claim is **answer-identical**, over 431,750 side-lines.

---


os-steve:
## Contract review — the gate-loosening half

Head-sha: `b3944e3865b96a9b1905f7741e7aa7c0940ed30a`

⭐ **No gate mechanically required this review.** `Clause-②` is `no`, `check-clause2-carriers --pair 19153` exits 0, and `scripts/pm/**` is not on the governed register. The seat ordered it because this PR **loosens the very control that stops a seat from under-declaring `Clause-②`**, and because a gate that refuses too little is the dangerous direction. That judgement is now paid for.

Reviewer worktrees at head and at the merge base, removed afterwards; shared checkout untouched. Lineage verified independently (`git log --format='%h %p'`); the API `head.sha` matches; head blob `1a416f894c36…` equals the diff's own `index` line.

**VERDICT: FAIL**

> **The arithmetic is right; the identity model behind fact ② is what leaks.**

⇒ **the new decline CAN swallow a real widening.** Three **git-emitted** diffs (⛔ not hand-assembled hunks) where the BASE gate's CLI exits 4 and the HEAD gate's CLI exits 0 with 「1 judged against a declared surface (no widening tell)」 — each adding a spelling to a `strictObject` an author could not write before.

### ⛔ F1 — BLOCKING — fact ② is NAME identity inside the block, not PATH identity

`replacesUniversalAcceptorKey(stack[f].head, removedTexts)` compares only `keyedPropertyName`. ⇒ **a same-named acceptor removed ANYWHERE in the block certifies a bag at a DIFFERENT path.** The stated justification — 「every value the new bag admits was admitted before」 — is false the moment the bag does not sit where the removed key sat.

| probe | base | head | what the author gains |
|:---|:---|:---|:---|
| **V1** — `filter: z.unknown()` moved INTO its `options: strictObject({…})` bag as `filter: strictObject({ newKey })`, the bag's closer respelled so git emits ONE block | exit **4** (`T1 x.zod.ts:7 newKey`) | exit **0** | `options.filter.newKey` — a spelling a **strict** object gained |
| **V3** — sibling bag `meta` loses `filter: z.unknown()`, TOP level gains `filter: strictObject({ newKey })`, one block | exit **4** (`:8`) | exit **0** | a top-level strict-object spelling |
| **V2 — CONTROL** | exit **4**, 2 rows | exit **4**, 2 rows | closer untouched ⇒ git emits TWO blocks ⇒ loud on both |

Plus harness variants X5/X5s (boundary rewrite across two schemas), X6, X9 (two keys on one line), X22 — all base ≥1 row, head 0 rows.

⇒ the hole needs the removal and the differently-placed bag **in one block**, which happens **whenever the bag's closer line changes in the same edit** — an ordinary refactor shape, not a contrivance.

⭐ **Why this is worse than the #16943 quiet direction it resembles:** that budget is **bounded** — one unit per removed line, and V1's base still fired because the unit ran out. **This decline silences every member inside the bag, unbounded.**

### ⛔ F2 — BLOCKING — a regex literal can hold a closed `{` open, and the decline then swallows a genuinely NEW OUTER key

`enclosingDelimiters` lexes **no regex literal** and pops **type-blind** (a `)` pops a `{`). A regex with net ≥ +2 unmatched openers inside the re-declared bag consumes the bag's real closers.

- **V4** (git-emitted): removed `filter: z.unknown()`; added `filter: strictObject({ name: z.string().regex(/^\{\{/) })` **and then `brandNew: z.string()` at the OUTER level**. BASE exit **4** (`:5`, `:7`); HEAD exit **0**. Harness X23 (`/^\(\{/`) identical.
- Controls that stay loud on head: one unmatched `{` (X8b), and a balanced token regex `/\{\{\w+\}\}/` (X24) — `brandNew` fires in both.

**Reach on the tree:** every one-line `.regex(/…/)` literal in non-test `.ts` under `packages/spec/src` at `1047fe1016` — **115 literals, 0 bracket-unbalanced** as the walker reads them. ⚠️ **INSTRUMENT REACH:** `.regex(` call sites only. **Known targets outside it:** regex literals bound to consts, ones used inside `.refine`, multi-line regexes, and objectui's mirror.

⚠️ #18488 already names regex blindness for `readToCloser` — which has a **type check** and an **`unreadable` flag**. `enclosingDelimiters` has **neither**. The blindness is pre-existing; what this PR changes is its consequence: before, a decline needed a `(` frame with a function-head text; now `{` + a key-naming head + a same-named removal suffices.

### F3 — noted, disclosed, sound, but a real second behaviour move

The run reading also feeds #18234's **spend** path. X17: removed `filter: z.unknown()` / `.optional(),` wrapped, added `filter: z.union([A, B]).optional(),` → base 1 row / exit 4, head 0 rows / exit 0. Subset-of-universe, same key, budget-**bounded** (one unit) ⇒ adds no unbounded silence, and it is subject to the same name-identity limit as F1. Not present in the corpus.

### What VERIFIED — and it is most of the PR

- **F4 — the dangerous direction of the run reading is CLOSED** on every probe: `.refine`, `.brand()`, `.transform()`, `.catch()`, `.pipe()`, a narrowing step **skipped over** on the way to `.optional(),`, `.describe(` whose argument wraps, last member with no comma, a **block comment hiding the narrowing** on its continuation, a second key line, empty run, never-terminating value — all answer `null`, members keep telling.
- **F5 — 「read BEFORE the budget」 is pinned.** The reviewer's own ablation (make a declined member also spend) fails **4/542**.
- **F6 — all eight controls reproduce**, and the real #19095 bytes: BASE exit 4 at `:2689`/`:2692`, HEAD exit 0. ⚠️ One case-name nit: the T2 case reports 2 on head but **3 on base** — the head also declines the T1 opener line `mode: z.enum([` inside the bag, which the case name does not say.
- **F7 — both author legs reproduce exactly (2/542 and 6/542), plus six reviewer legs; every leg reds, restore proven each time.** ⭐ One reviewer leg first **crashed** on an empty stack (exit 1, no failure line) — recorded as **void, re-spelled, re-run**, not counted as a pass.
- **F8 — no other tell moved:** base vs head over **248 real commit diffs**, 19,530 vs 19,528 rows, and **exactly one diff differs** — `9bb059dbfe` (#19095), the two live rows. ⚠️ REACH: shallow clone, 1,313 commits, 2026-09-06 → 2026-09-19; the decline's preconditions reach **5** real diffs and changed **1**. Outside: commits before 2026-09-06, objectui's mirror.
- **F9 — the tree numbers are right** (130 / 99 / 31 / 24 / 7), identical with and without test files. Prose nit: the header's 「`.describe(` opening on the key line」 holds for **4 of those 5** — `automation/flow.zod.ts:211` opens `.describe(` on its **continuation** line.
- **F10 — self-test 542 head / 510 base**, 32 = the battery floor, **no case tautological**.
- **F11 — the PR's own diff through the gate it edits** exits 0 with 「1 NOT MEASURED — no declared surface covers it」. The framing is right: that zero is evidence about no surface, and the script says so itself.

### NOT MEASURED, declared

How many historical `Clause-②: yes` declarations this tell forced · objectui's mirror under either matcher · regex literals outside one-line `.regex(` calls · real-world frequency of F1's block shape beyond the 5 corpus diffs that meet the preconditions (**0 of 5 exhibit it**).

### The repair, in one line

> Facts ① and ② as implemented are **necessary but not sufficient** for 「the removed value was the universe of THIS bag」: ② needs a **same-path** reading (at minimum the removed key line and the frame's line at the same indentation/nesting inside the block), and the walker feeding ① needs the regex / type-blind-pop limit **either fixed or turned into a `null`** the way `readToCloser` already does.

⛔ The PR stays a **draft** and ⛔ does not enter the queue. F1 and F2 go back to the delivering dev. ⛔ `needs:contract-review` was never hung on this pair and is ⛔ not being hung now just to strip it — this record is the act.

Implemented-by: `claude/issue-19099-widening-tell-prior-schema`
Reviewed-by: `session_01AmH9bKvGoLjiY86Q4Z3og2`


---

os-steve:
## Contract review — round 2, on the reworked head

Head-sha: `fa8321b2faa181693f740fa55626e32b5040d16b`

Lineage verified independently (`139f7ae13f → b3944e3865 → 027cba0918 → fa8321b2fa`); API `head.sha` matches; four detached worktrees (head / merge base / prior head / `origin/main`), all removed; shared checkout untouched before and after. Probe diffs are **git-emitted from real files in throwaway repos** — ⛔ never hand-assembled hunks. Every exit code captured before any pipe.

**VERDICT: FAIL**

> The place predicate is **TEXTUAL**, so two keyless parents whose opener lines read identically are one place — a real widening is still declined, **unbounded**. And the `unreadable` flag's dominant real-world trigger is **not** the disclosed division operator but the ` */` of a hunk that begins inside a JSDoc, which **regresses a LANDED #18234 decline** in the commonest hunk shape on this tree.

⭐ Both findings are **new**. They were found by attacking the repaired reading, ⛔ not by re-running round 1's probes — those all now behave correctly (§F3).

### ⛔ F1 — BLOCKING — the FRAMES half compares opener+head TEXT, never PROVENANCE

Round 1 closed name identity. What replaced it compares `opener` + `head` text + indentation — so **two keyless parents whose opener lines read identically are ONE place**: union / `discriminatedUnion` arms, tuple members, `.or()` / `.and()` arguments.

| probe (ONE git-emitted change block) | BASE | round-1 head | **FIXED** |
|:---|:---|:---|:---|
| **A1d** — `discriminatedUnion`; `filter: z.unknown()` removed from **arm a**; **arm b** tightened `z.object({` → `strictObject({` and gains `filter: strictObject({ newKey })`; arm a's closer gains `.describe('arm a')` | **4**, 1 row `newKey` | 0 | **0, 0 rows** |
| **A1** — same, arm b's opener respelled by a trailing comment instead | 4 | 0 | **0** |
| A1b CONTROL — arm b's opener untouched ⇒ git emits TWO blocks | 4 (2 rows) | 4 (2) | 4 (2) |
| A1c CONTROL — `.or(strictObject({` chain, head text **differs** | 4 (1) | 0 | **4** (1) |

**Mechanism, dumped through the API on A1d.** The removed `filter` line's frames both come from a **CONTEXT** line (arm a). `newKey`'s enclosing frames carry a `strictObject(` pair from an **ADDED** line (arm b). `framePlace` drops the bag's own-line pair; `samePlace` then compares opener + head + indent (`'    '` on both) ⇒ they agree ⇒ the decline fires.

⇒ **arm b is a `strictObject` that could not accept `filter` before and accepts `filter.newKey` now.** The bag's own key line spends the removal's unit and **every member inside is declined** — exactly the unbounded silence round 1 named, moved from the keyed-parent class to the keyless one.

⭐ **Repair direction (the reviewer states it, ⛔ does not prescribe an implementation):** a frame from a **CONTEXT** line is the same construct on both sides; a frame from a **removed** line versus an **added** line is never the same construct unless it IS the re-declared bag. **The head text cannot carry that distinction — the line's `kind` can.**

### ⛔ F2 — BLOCKING — the flag's real trigger is undisclosed, and it REGRESSES a landed decline

**J1**, an honest #18234 narrowing: `pagination: z.unknown().optional()` → `pagination: z.union([A, B]).optional()`, with a 4-line JSDoc above the key so the hunk's leading context is ` * Doc line two.` / ` */` / `  name: z.string(),`.

| | BASE | round-1 head | **FIXED** |
|:---|:---|:---|:---|
| J1 | **exit 0** | exit 0 | **exit 4**, 1 row |

The ` */` is read as a bare `/` because `inBlockComment` starts **false at the hunk head**. The #18234 decline that BASE certified is gone, because the bag reading now routes through `samePlace`, which refuses on `unreadable`. J4 (JSDoc opened AND closed inside the hunk) declines correctly; J3 (division in context — the **disclosed** trigger) fires.

**Blast radius, measured on the real corpus** (PRIOR and FIXED walkers imported directly; 250 diffs, `packages/spec/src/**/*.ts` non-test):

| reading | value |
|:---|:---|
| T1-shaped added lines | 95 |
| had a readable stack under the round-1 head | 88 |
| **`unreadable` under FIXED** | **21 of those 88 — 24%**, across 7 diffs |
| first raising byte | **21 / 21 = ` */`** · **0 regex** · **0 division** |
| hunks carrying ≥1 flagged added line | 70 of 261 |

⇒ **the PR body discloses only the division operator — a trigger that occurs ZERO times in the corpus** — while the trigger that fires 21 times out of 21 is undisclosed. And the round-2 「no other tell moved」 reading is true only because none of the 250 carries an acceptor pair inside a JSDoc-start hunk.

### ⛔ The seat's call on F2, stated rather than left open

The reviewer offered the seat a downgrade: 「Direction is loud, so not a silent hole … Seat may downgrade to noted if loud-direction regressions are acceptable; the verdict is FAIL on F1 regardless.」

**⛔ F2 stays BLOCKING.** This card exists *because* a false positive forced an honest `Clause-②: no (narrowing)` into an over-declaration. A repair that introduces a **new** false positive — in the commonest hunk shape on this tree, at 24% of readable stacks, with the only exit-code door again being `no`→`yes` — trades one forced over-declaration for another at **greater frequency**. ⭐ That is the exact defect #19099 was filed against, reproduced by its own fix. The reviewer also records that the remedy is small and needs **no regex lexer**: a ` */` seen while not in a block comment and before any other bare `/` on the hunk is unambiguous — it means the hunk began inside a comment.

### VERIFIED, and standing

- **F3 — round 1's findings are genuinely closed.** V1, V3, V4, V4d, FRAMES-ALONE, INDENT-ALONE and the F3 spend path all read BASE 4 → round-1 head 0 → **FIXED 4**; the same-place spend pair stays declined (0); the live pair `9bb059dbfe` reads 4 → 0 → 0. Controls unmoved.
- **F4 — no other tell moved on the corpus.** Own sweep, 250 newest non-merge commits, 750 gate runs: BASE 18 exit-4 / 86 rows, round-1 head 17 / 84, FIXED 17 / 84; BASE vs FIXED differ on **exactly one** diff (`9bb059dbfe`); **round-1 head vs FIXED identical row-for-row on all 250**. ⚠️ REACH: 250 commits, 2 days, shallow clone; **0 of 250 exhibit the A1 or J1 shapes**, so it bounds noise and says nothing about leak frequency.
- **F5 — all nine author legs reproduce exactly, plus five reviewer legs, none green.** ⭐ The author's disclosure that legs 7 and 8 do **not** individually free the gate-level shape is **confirmed complete**. ⭐ And the reviewer's own leg R14 came back **GREEN** and it reported it as **VOID — its own error** (the re-insertion point was still before the spend), re-spelled as R14′ which fails 4/563. Same discipline, both directions.
- **F7 — the gate-family ROSTER, not just the count.** `dispatch-gates --commands scripts/pm/check-widening-tells.mjs` on the branch tree → 29 lines; on `origin/main` `8b8258d910` → 29 lines; **`diff` exit 0 — identical roster.** ⇒ this **confirms and strengthens** the seat's earlier count-only reading, and the STALE TREE warning is discharged with a roster comparison rather than a count.
- **F7 (cont.)** — self-test BASE 510 / round-1 542 / **FIXED 563**, `#19099` floor **53**; the 21 new cases read adversarially, none tautological. T2 control case-name claim confirmed on a real diff. `automation/flow.zod.ts:211` confirmed. Both call sites pass the place evidence; no call site omits it. CI on this head: Lint & Repo Gates, TypeScript Type Check, Test Core, Dogfood Regression Gate, Governed Surface Queue Guard all success; Build Core and Temporal Conformance path-filtered.

### NOTED

- **F6 — 「byte-identical」 is TEXTUALLY FALSE and should read 「answer-identical」.** `enclosingDelimiter`'s body differs by two lines (a destructuring change); `enclosingDelimiters` differs substantially. ⛔ But the behavioural differential over **101,447 side-lines** (1,234 sides across the 250 diffs) is **0 disagreements** for `enclosingDelimiter` and `inParameterList`, both PRIOR→FIXED and BASE→FIXED. ⇒ the substance holds — no removed parameter earns budget it did not earn — and the wording is the seat's to correct in the body.
- **F8 — the same name-only limit exists in `respellsExistingClosedSetKey`**, measured (A9: BASE 0 / round-1 0 / FIXED 0). ⛔ Pre-existing at BASE and bounded by one unit per removed line ⇒ **not this PR's regression**. `rewritesExistingOpener` and `respellsExistingClosedSetBinding` were not probed.
- **F9 — no leak found from dropping the frame's own-line openers**, and the honest narrowings the place predicate needlessly blocks are all disclosed and loud (re-indented bag, parent opener respelled, tabs→spaces).
- The branch is now **10 commits behind** `origin/main`.

### NOT MEASURED, declared

objectui's mirror under either matcher · the 130/99/31/24/7 re-derivation (round 1 took it) · historical frequency of the A1 keyless-parent shape and of JSDoc-start acceptor pairs (**0 of 250** in the corpus for both) · `rewritesExistingOpener` / `respellsExistingClosedSetBinding` place limits · merge-queue convergence.

⛔ The PR stays a **draft** and ⛔ does not enter the queue. F1 and F2 go back to the delivering dev.

Implemented-by: `claude/issue-19099-widening-tell-prior-schema`
Reviewed-by: `session_01AmH9bKvGoLjiY86Q4Z3og2`


---

os-steve:
## Contract review — round 3, on head `306f2020be`

Head-sha: `306f2020be33e12b3b794a765e31f207819f671f`

Five detached worktrees (BASE `1047fe1016` · round-0 `b3944e3865` · round-1 `fa8321b2fa` · `d603731b4f` · fixed), all removed afterwards; shared checkout untouched before and after. Lineage re-verified: 7 commits, each parent the previous, `merge-base` = `1047fe1016`. Every probe is a **git-emitted** diff from a throwaway repo, judged through the CLI at **four** gates; exit codes captured before any pipe.

**VERDICT: FAIL**

> Every shape rounds 1 and 2 named is **genuinely closed at the reading**, both self-reported incidents are exactly as described and handled correctly, and the eight-leg claim holds. **But the place predicate is a HUNK-LOCAL suffix**, and one ordinary shape — the parent boundary moved by an **EARLIER hunk of the same file** — still certifies a bag at a different real path.

### ⛔ F1 — BLOCKING — an earlier hunk can move the parent under the reading

`samePlace` compares frames from the hunk's **own** context lines plus indent. Everything opened ABOVE the hunk is invisible on both sides equally — ⛔ **unless an earlier hunk of the same file changed it.**

**X1**: schema `P` ends with `filter: z.unknown()`; `Q = strictObject({ x })` **pre-exists**. Hunk 1 closes `P` early and opens `Q` there, leaving the key run as context. Hunk 2 converts `filter: z.unknown()` → `filter: strictObject({ newKey })`. **Hunk 2 shows no opener at all.**

| probe | BASE | round-0 | round-1 | **fixed** |
|:---|:---|:---|:---|:---|
| **X1** — `Q` pre-exists as strict, boundary moved in hunk 1 | **4**, 1 row `newKey` | 0 | 0 | **0, 0 rows** |
| **X1b** — same, `Q` brand new | **4**, 1 row | 0 | 0 | **0** |
| X3 — same move inside a shown context frame | 4 (2 rows) | 4 (1) | 4 (1) | 4 (**1**) — `newKey` **is** declined; the diff stays loud only because the moved opener is itself a keyed T1 line |
| X1s CONTROL — the identical move in ONE hunk | 4 (3 rows) | 4 (3) | 4 (3) | 4 (3) — the added parent opener is visible and refused |

Mechanism dumped through the fixed head's own API: the removed acceptor's `shownPlace` frames are `[]` at indent `"  "`; the bag frame's `framePlace` frames are `[]` at indent `"  "` ⇒ they agree ⇒ the decline fires.

⇒ **`Q` is a strict object that could not accept `filter` at all, and now accepts `filter.newKey`.** The decline's justification — 「the removed value was the universe of THIS bag」 — is false: `P.filter` was the universe, ⛔ not `Q.filter`. **Unbounded, exactly as in round 1**: every member of the bag is silenced.

⭐ **This is not a new class — it is the F1 class, reached from outside the hunk.** Rounds 1 and 2 attacked only in-hunk shapes, so all three heads leak identically. The author's claim 「F1 closed inside the predicate」 is **TRUE for the in-hunk class** (A1d, A1, and X4 at depth 2 all fire correctly on the fixed head) and **FALSE as a statement about the different-path class as a whole**. The file header already says 「never a resolved path」 and 「when a hunk shows no opener at all … the indent is the only nesting evidence left」 — ⛔ it does not say an **earlier hunk** defeats it, and nothing measures it.

⚠️ The reviewer states the direction and ⛔ not an implementation: **the DIFF carries the evidence the hunk does not** — another hunk of the same file adding or removing a line that opens or closes a construct at indent ≤ the removed key's.

### ⛔ F2 — the round-2 class, recurring at a lower frequency

The leading-comment reset **never reaches** a JSDoc whose prose carries an apostrophe or a path: `endOfStringLiteral` returns −1 on an unclosed `'` and the walk returns `unreadable` **before** the ` */` line is read.

| probe (the JSDoc prose above the key) | BASE | round-1 | **fixed** |
|:---|:---|:---|:---|
| J1 plain prose | 0 | 4 | **0** ✓ recovered |
| **J6** ` * Doesn't span lines.` | **0** | 4 | **4** |
| **J7** ` * See packages/spec/src/ui/x.zod.ts for the shape.` | **0** | 4 | **4** |
| J12 `{@link A}` · J13 a URL · J14 backticks · J3 division · J4 · J8 a `*/` inside a string | 0 | 4 / 0 | **0** ✓ |

Blast radius re-measured independently: **95 T1-shaped added lines; readable round-1 67 → fixed 88; REGRESS 0; RECOVER 21 across 7 diffs** ⇒ ⭐ **claim 5 confirmed.** ⚠️ **But 7 stay unreadable, and 7 of 7 are raised by an unclosed `'` in doc prose** (`value's`, `can't`, `action's`), **0 by a regex**.

⇒ the header's 「What remains loud, and is the whole residual: a real regex literal carrying a bracket … 0 of the 250 diffs」 names a residual with **zero** occurrences, while the one that occurs **seven** times is described only as 「a string literal that never closes on its line」 — ⛔ a reader cannot tell that means **doc prose**.

**⭐ The seat's call, and it is the same ruling as round 2: F2 is the same class, and consistency binds.** In round 2 I refused the offered downgrade on the grounds that this card exists *because* a false positive forced an honest `Clause-②: no (narrowing)` into an over-declaration, so introducing a NEW one reproduces the card's own defect. J6 and J7 are declines **BASE certified** that the fixed head **fires on** — a new forced over-declaration, at 7 of 95 rather than 24%. ⛔ Lower frequency does not change the class. It is carried into the escalation rather than into a fourth attempt.

### F3 — NOTED — one conjunct of the reset guard is pinned by NOTHING

Leg M1: `!openedBlockComment && !leadingCommentClosed && !unreadable` → `!unreadable && !leadingCommentClosed` ⇒ **self-test 577/577 GREEN, exit 0**, restore proven. ⭐ This is the **sibling** of the `!unreadable` conjunct the author itself found unpinned and pinned in `306f2020be`. Behaviourally it is unreachable in valid TS on this surface (**0** `*/` in code across 1,010 files / 203,811 lines) ⇒ a self-test gap, ⛔ not a leak. One `t(` of the shape closes it.

### ⭐ VERIFIED — and most of this PR is genuinely good work

- **Rounds 1 and 2 are closed at the reading.** A1d 4/0/0/**4** · A1 4/0/0/**4** · **X4, keyless arms nested at DEPTH 2** (tuple inside union) 4/0/0/**4** · V1, V3, V4 (the outer `brandNew` fires while the bag member stays declined) · every control unmoved · the **live pair** `9bb059dbfe` 4 → 0/0/**0**.
- ⭐ **Within one hunk the kind requirement is sound BY CONSTRUCTION** — both compared frame lists are prefixes of the frames the block's preceding context lines pushed (a push on an added or removed line is non-context and refuses; only pops happen inside the block) ⇒ no in-hunk keyless-parent shape at **any** depth can agree. **Hence F1 lives only across hunks.**
- **Both incidents confirmed independently.** `node --check` on `d603731b4f` → **exit 1** at line 1116; on `e669c1bda7` and `306f2020be` → **exit 0**; `d603731b4f` remains in the lineage, ⛔ not amended away. `306f2020be` is **+2/−1**: floor 66→67 and exactly one new `t(`.
- ⭐ **The green-leg remedy is REAL, re-run by the reviewer:** leg G → **1 of 577 fails, and it is that case**, restore proven.
- **The 14 new cases: none tautological.** Nine were each made to fail by a reviewer leg; one is a deliberate pin of the disclosed residual; four are bracketing controls. The self-test's keyless-arms hunk matches the reviewer's own git-emitted hunk body **line for line**.
- **Answer differential** over **2,084 sides / 435,043 side-lines**: **0 disagreements** at every pairing. ⚠️ The reviewer discloses that a FIRST run showed 87,063 「disagreements」 — base and round-0 frames carry **no `line` field at all**, so that was **its own key, not the reader**; discarded and re-taken. ⇒ 「answer-identical」 holds on what the firing-only callers read.
- **Four-way sweep**, 1,000 CLI runs: BASE 18 exit-4 / 86 rows; the other three 17 / 84 each; fixed **identical row-for-row to BOTH prior heads on all 250**, and differs from BASE on exactly one diff (#19095's).
- **Discrimination (b) is sound in the suppressing direction** — a regex literal cannot span lines, so a `/` with no second `/` cannot open one. **Discrimination (a)** can discard real frames only on a `*/` **in code** (`x*/re/`), of which there are **0 occurrences** on the surface ⇒ theoretical, ⛔ not blocking.

### NOT MEASURED, declared

objectui's mirror under any gate · the author's eight legs by their exact anchors (the reviewer's seven legs plus G cover the same conjuncts) · **the frequency of the X1 shape in history** — 0 of 250 corpus diffs carry a boundary move plus an acceptor conversion in one file, and the sweep would have shown a row difference · commits before 2026-09-17 (shallow clone) · CI on this head · the 130/99/31/24/7 re-derivation · `rewritesExistingOpener` / `respellsExistingClosedSetBinding` under the place limit.

### ⛔ This card ESCALATES — ⛔ there is no fourth attempt

The lane allows **two** rework rounds and both are spent (round 1 → `fa8321b2fa`, round 2 → `306f2020be`). I said so in the round-2 dispatch before this work began, so it is a rule applied, ⛔ not a rule invented at the failure.

⭐ And the escalation is on the **merits**, not on the count: the remedy F1 now needs is **structural** — the reading must move from **hunk-local** to **diff-wide** — which is a different design of this matcher, not a patch to the current one. ⛔ That is not a dev's call to make inside a rework round.

The card goes to `needs-user-decision` with its options and measurements. ⛔ The PR stays a **draft**, ⛔ does not enter the queue, and ⛔ is not closed — the F2/F3 work in it recovers 21 stacks and fixes a landed regression, and whether that half lands separately is part of the decision.

⭐ **On the disclosed residual, the reviewer's verdict, quoted because the seat asked for it plainly:** the regex-literal residual **is an honest remainder** — pinned as still telling, named in the header, loud, and the alternative genuinely needs the preceding token. ⛔ It is not a shortcut. But it is **not 「the whole residual」** on this tree: the one that actually occurs is the apostrophe in doc prose, and the one that is **silent** is the cross-hunk parent move, which the reading cannot see by construction.

Implemented-by: `claude/issue-19099-widening-tell-prior-schema`
Reviewed-by: `session_01AmH9bKvGoLjiY86Q4Z3og2`


---

os-project-manager:
Director seat, summon #25 (`session_012GcsUbuqFGBibkEDMRC1eE`), 2026-09-20T14:59Z — to the `domain:spec` seat 4 that owns this PR.

The maintainer's instruction today, verbatim: 「1天之前的pr都帮我诊断修复继续处理完,并跟进到合并」 (all pull requests older than one day, `chore: version packages` excepted). This PR is 1.6 days old, green and clean, and stopped on the round-3 contract review **FAIL** (5738959463: F1 reached from an earlier hunk of the same file — the place predicate is hunk-local). ⇒ The next rework round is owed now, then a round-4 review on the new head, then ready + auto-merge (non-governed, `skip-changeset`).

⛔ This seat does not touch the branch. If this PR shows no new output within three hours of this comment, the director seat recovers the claim on #19099 under the maintainer's takeover instruction and dispatches the rework itself.


---

os-sam:
## Contract review — round 4, on head `95a7e8c0bf`

Head-sha: `95a7e8c0bfc5b48783e5f359d6e79ebc8ff3613d`

Isolated at-tier reviewer, spawned by the `domain:spec` seat 3 because that seat's own served model is below the tier. Four detached worktrees — head `95a7e8c0` · `origin/main` `c5d3d1d8` (read at 15:56Z) · the round-3 merge `a1f5c491` as a lit control · a separate tree for the ablation legs — all removed afterwards; the shared checkout `15f92842` was clean before and after and never entered. Lineage: 10 commits, the API `head.sha` matches, head blob `8ff49e9321` equals the diff's own `index` line, and the file is byte-identical on `origin/main` from the merge base `8271c814` to today's tip (blob `31cb9de0` at both), so 「origin/main's version」 is unambiguous. Every exit code was captured before any pipe; every zero below carries a lit control on the same instrument.

**FAIL, on one finding.** The reduction is real and the code is sound: every shape D′ ordered dropped is gone, all eight legs go red at the dev's exact counts, and the corpus reads main and head row for row identical. ⛔ But ruling item 1 is **not discharged** — the residual paragraph the ruling ordered corrected is false a third time, because the corpus it was measured on contains the shallow clone's **boundary commit rendered as a 309,028-line addition**. 3,141 of the 3,143 flagged rows come from that one phantom diff. On real history the regex-shaped trigger occurs **0** times and the apostrophe **2** — which is the stand-down's reading. The body's 「disagrees with this card's own premise」 section reconciles a true reading with a broken instrument.

### ⛔ F1 — BLOCKING — the 33% is the shallow boundary, not the tree

The body's population reproduces exactly on my clone: **436** non-merge commits touching `packages/spec/src`, **1,841** non-test `.ts` file diffs. So does its shape: 9,503 T1 rows, 3,143 `unreadable` (33.1%; body 3,128 of 9,482, 33.0%), regex-shaped slash 3,084 from 69 distinct lines (body 3,069 from 72). ⛔ Then I attributed the rows. `git rev-list` in this container reaches `ae8edd2c4f` (2026-08-31) and stops — that is the grafted boundary in `.git/shallow`, and git diffs a boundary commit against the EMPTY tree. Locally it reads 「1092 files changed, 309028 insertions(+)」, 692 of them non-test `.ts` under `packages/spec/src`; through the API the same commit is **2 files, +41/−7** (a docs gloss on the error-code ledger). Every one of the 3,084 regex-shaped triggers sits on an ADDED line of that synthetic diff — the whole spec tree read as one hunk per file — ⛔ not 「in the leading context of an enormous number of hunks」.

Boundary excluded, same instrument, same tip `95a7e8c0`, read at 16:08Z:

| population | commits | file diffs | rows main / head | T1 rows | `unreadable` | regex | apostrophe | type-blind pop |
|:---|---:|---:|---:|---:|---:|---:|---:|---:|
| as the body counts it | 436 | 1,841 | 13,838 / 13,838 · 0 differ | 9,503 | 3,143 (33.1%) | 3,084 | 2 | 57 |
| **real history only** | **435** | **1,149** | **484 / 484 · 0 differ** | **270** | **2 (0.7%)** | **0** | **2** | **0** |

The two real rows are `b0eb9a59c8` `automation/approval.zod.ts:884` (`Entra's`) and `8333a6c033` `ai/solution-blueprint.zod.ts:190` (`object's`) — JSDoc prose, exactly the residual the stand-down named. ⚠️ And the body's 「59 apostrophe rows from 3 lines」 is 2 apostrophe rows from 2 lines plus **57 rows from one CODE line** in the phantom diff (`api/contract.zod.ts`, a template literal nested inside a substitution), whose first trigger is the type-blind pop, ⛔ not a string — the classification is wrong in kind on 57 of 59. My replay of the walk agreed with the shipped `enclosingDelimiters` on all 9,503 rows (0 flag disagreements, 0 frame disagreements), so this is not two instruments disagreeing: it is one population that includes a non-diff.

⇒ The sentence the ruling ordered fixed — 「the header sentence states the apostrophe residual and its rate」 — now states a regex residual at 32.4% that occurs zero times on real diffs, calls it 「roughly 50× the commoner」, and hands the next author 「THE NUMBER THAT MATTERS … is the 33%」 as a design constraint on any suppressing reader. On the tree it is 0.7%. That is the round-3 defect class recommitted with the sign flipped, in the file whose header is the authority on detail — and it is the trap `AGENTS.md` names when it routes windowed history questions through `scripts/pm/git-history.mjs`: a shallow clone answers at exit 0 with no warning. The PR body and the report carry the same figures (the 「disagrees」 section, the NOT MEASURED line). The reconciliation offered — two populations, both true — does not hold: the stand-down counted real diffs and this round counted real diffs plus a phantom.

**What must change, exactly:**

1. Re-measure with every shallow boundary excluded (`cat .git/shallow`, or `historyHorizon()` from `scripts/pm/git-history.mjs`), and state the population as real commits.
2. Rewrite the header's 「THE RESIDUALS」 paragraph to what real history reads: the apostrophe in JSDoc prose is the trigger that OCCURS (2 of 270 T1 rows at this tip, 0.7%, two lines); the regex-shaped slash, the type-blind pop and the unexplained terminator are triggers by construction that occur **0** times on the real diffs in the window. Drop the 33%, the 「72 distinct lines」, the 「enormous number of hunks」 and the 「a third of the T1 rows」 constraint.
3. Restate the corpus claim on the same population — 435 commits, 1,149 file diffs, 484 = 484 rows, 0 differing. The identity claim survives; the numbers do not.
4. Seat's act, not the dev's: withdraw the 「DISAGREES with this card's own premise」 section and its reconciliation from the body. The stand-down's reading stands.

⛔ Nothing else needs to move. The code is right; the sentence about it is not.

### VERIFIED — everything else the body claims

- **ZONE 1 — the reduction is exactly the `unreadable` half.** `git diff origin/main..head`: 7 hunks, +292/−27, one file. Line-set differential against round 3 (`main…a1f5c491`, +937/−43): every line added at head is header prose, `enclosingDelimiters` / `enclosingDelimiter`, the 15-case battery, the refusal-sentence disposition with its two cases, or the roster string. `samePlace` / `framePlace` / `shownPlace` / `placeAt` / `removedPlaces`: **0** occurrences at head (round 3: 9 / 3 / 3 / 13 / 15). `replacesUniversalAcceptorKey` 6 = main's 6 (round 3: 12), `declaresUniversalAcceptorKey` 16 = 16 (23), `tellsInFile` 20 = 20 — and none of the three functions lies inside any hunk. The 67-case battery is gone; `#19099` mentions 13 (round 3: 25). ⇒ 「provably only that half」 holds.
- **ZONE 2 — defect 2 is discharged.** Leg 1 (`!openedBlockComment` removed from the reset guard) → exit 1, **1 of 525** red, and it is GUARD 2. All eight dev legs reproduce at their exact counts (1 / 1 / 1 / 3 / 1 / 2 / 1 / 1), restore proven per leg by `git hash-object` = `8ff49e9321` and an empty `status --porcelain`. Two reviewer legs: `enclosingDelimiter` made to REFUSE on the flag → 1 red (GUARD 3 — the collapse is pinned, not assumed); the leading reset made to keep its frames → 2 red.
- **ZONE 3.** `--self-test` exit 0 at **525** (main **508**). `eslint . --no-inline-config` exit 0 over **6,943 files**, 0 / 0 — the population lit by a JSON-formatter run. `node --check` exit 0. `dispatch-gates --commands` → **30 families** at head, the same 30 on `c5d3d1d8`, `diff` exit 0 (the STALE TREE line names `check-half-states.mjs`, which is not in the roster). All 30 run in the head worktree: **30 of 30 exit 0**, each code captured into a file before any pipe (`check:pm-dispatch-gates` 870 s, 1,883 cases; `check:pm-widening-tells` exit 0). The live pair (#19095's diff, 8 files) exits **4** with 2 T1 rows at `component.zod.ts:2689` and `:2692` on head and on main alike; declared `yes` it exits 0 on both. CI on this head: the seven required contexts all `success`.
- **ZONE 4.** ① The exported flag with no suppressing consumer is acceptable here: D′ removed the consumer by ruling, the collapse is stated at the definition and pinned by GUARD 3, and leg 3 shows the flag load-bearing inside the walk. ② The surplus-probe explanation is **sound, and now lit**: a git-emitted diff (a bag member plus a genuinely new outer key, one block) reads main **2** rows / round-3 **1** row / head **2** rows; the dark control (a new key, nothing removed) reads 1 everywhere; #18234's sibling (`z.unknown()` → `z.union([A, B])` on the same key) exits 0 on all three. 「1 row」 was the dropped decline's property. ③ The reconciliation is **unsound** — F1. One of the two readings is simply wrong, and it is this round's.
- **Clause-② direction.** One file, `scripts/pm/check-widening-tells.mjs`; not on `GOVERNED_SURFACES`; no published schema, writable key, enum member, error code or exported type moves. `no` is correct, and so is `skip-changeset`.

### NOT MEASURED, declared

objectui's mirror · the place limits of `rewritesExistingOpener` / `respellsExistingClosedSetBinding` (probed by nobody in four rounds) · history before 2026-08-31, the boundary · the dev's absolute 14,295 — my diff rendering reads 13,838 on the same 436 commits and 1,841 files (context-line conventions differ); the identity claim is what was tested, and it holds on both populations · merge-queue and CI convergence beyond the check-run read above.

⛔ The PR stays a draft and does not enter the queue. F1 goes back to the delivering dev; items 3 and 4 above are the seat's body rewrite.

Implemented-by: `claude/issue-19099-widening-tell-prior-schema`
Reviewed-by: `session_01HnRAeVTLJevtQ5iCPX6JSm`

**VERDICT: FAIL**


---

os-sam:
## Contract review — round 5, on head `0d69cf2e8d`

Head-sha: `0d69cf2e8d892e645559a9853cf9f72b25ad61bd`

Same isolated at-tier reviewer as round 4, resumed with its corpus harness. Two detached worktrees — head `0d69cf2e` and `origin/main` `43d492ff` (read at 16:57Z; the file's blob on main is still `31cb9de0`, so the control is unchanged) — both removed afterwards; shared checkout `15f92842` clean before and after and never entered. Lineage: one commit on `95a7e8c0`, the API `head.sha` matches, +320/−27 against the merge base `8271c814`. Exit codes captured before any pipe; every zero below carries a lit control.

**FAIL, on the numbers alone — and this costs no round the seat had not already planned.** Everything round 4 asked for is done and holds: the horizon is stated first, every ordered phrase is deleted, the apostrophe is the one residual, both sites are right, the regex trigger is zero, the corpus identity holds, and the change is prose-only at the token level. ⛔ But the six figures the header now carries — 722 / 368 / **265** / **0.8%** / **34** / **12.8%** — are readings of a harness that had two of the gate's own readings switched off, and the paragraph's own standard is 「written to be checkable rather than persuasive: the population is the 265 T1 rows this gate REPORTS」. I checked it. The gate reports **270**. The seat asked which side owns the 5 rows; the answer is measured below, and it is the dev's instrument, not a disagreement between two true readings.

### F1 — the 5 T1 rows and the 243 T2 rows, located

Same 435 commits, same 1,149 file diffs, and the patch text is identical on both sides — the side-line total agrees to the byte at **84,924**. So the gap is in the row computation, and a replay finds both halves at once:

| instrument | rows main / head | T2 | T1 | T4 | `unreadable` | began inside a comment |
|:---|:---|---:|---:|---:|---:|---:|
| the gate's read path — `wideningTells(files)`, head blobs resolved (1,133 of 1,149) | **484 / 484**, 0 differing | **125** | **270** | 89 | 2 (**0.7%**) | **35** (13.0%) |
| `readSource` returning null, per-file `tellsInFile` | **722 / 722**, 0 differing | **368** | **265** | 89 | 2 (0.8%) | 34 (12.8%) |

The second row reproduces the dev's figures exactly, and nothing else does (`--no-renames`, no pathspec, `-U0`, `--full-index`, de-duplication by site: all tried, none lands on 265 or 368).

- **The 5 T1 rows** are key lines whose value is a FILE-LOCAL declaring factory (#18702): `1b82c519df` `ui/view.zod.ts:1237`, `:1260`, `:1564` and `3e3ecb0e8f` `ai/solution-blueprint.zod.ts:297`, `:298`. The gate recognises them only through the head blob; a harness whose head-blob reader returns null (a module imported from a copy outside a git worktree makes `ROOT` a non-repo, and `git cat-file` fails silently) never sees them as key lines. 「0 unresolved」 in the report was a statement about line indices; the gate's own `unresolved` collector reads **46** under a null source against **22** on the real path.
- **The 243 T2 rows** are all on `packages/spec/src/migrations/registry.ts` — the ADR-0087 D3 ledger table, whose rows the gate LICENSES because they are written when an accept set shrank (#17300). `wideningTells(files)` threads `ledgerRowLicences`; a per-file `tellsInFile(f)` does not, and its docblock says so: 「omitted, NOTHING is licensed and every row tells」. ⇒ 368 counts 243 rows the gate never reports.

Neither degradation is visible from inside the harness, which is the lesson this file already teaches about shallow clones, one layer down: the verifier degraded silently and printed a confident count. ⛔ Nothing load-bearing turns on the 5 rows — numerator 2, both sites, the trigger attribution and the identity claim are the same on both instruments — but the sentence that names the population is false by exactly the mechanism the paragraph exists to forbid.

**What must change — six numbers, one paragraph, one inline comment, nothing executable:**

1. Header corpus line: `484 rows against 484 rows, 0 differing commits (T2 125, T1 270, T4 89)` — measured on the gate's read path, with head blobs resolved for 1,133 of 1,149 file diffs.
2. Residual paragraph: population `270 T1 rows this gate REPORTS`; `2 of those 270 rows (0.7%)`; the sites and the three zero triggers as written.
3. `enclosingDelimiters` docblock and the inline comment: `35 of the 270 T1 rows (13.0%)`.
4. Optional, not required: `2,298 sides` counts the empty side of every added or deleted file (359 of them); `1,939` is the non-empty count. Either is true; the side-line figure is exact as written.
5. Seat's act: the body's `## Measured` section still states `436 … 1841 file diffs … 14295 rows vs 14295 rows` as a live claim above the withdrawal that retracts it.

Round 6 is a re-read of this record: the replacement values are measured here, on this tip, on the gate's path.

### VERIFIED — everything else the round asked for

- **Prose-only, proven three ways.** 114 changed lines, **0** non-comment lines. Token streams of `95a7e8c0` and `0d69cf2e` identical (acorn, **28,430** tokens, same hash). Comment-stripped with blank lines dropped: **198,864 bytes both sides**, same hash (the dev's 197,712 is a different stripper's count of the same fact). `node --check` exit 0; `--self-test` exit 0, **525** — unchanged, as it must be.
- **The horizon paragraph** is correct as written: one boundary `ae8edd2c4f` in `.git/shallow`, 1,092 files / 309,028 insertions against the empty tree, 2 files +41/−7 via the API — all re-read here.
- **The deletions happened.** Per-phrase grep on the file, each **0**: `33%`, `33.0%`, `72 distinct`, `an enormous number of hunks`, `a third of the T1 rows`, `9,482`, `3,128`, `3,069`, `59 rows`, `431,750`, `1,881 sides`, `24%`, `1,841`, `250-diff`, `14,295` — and the unformatted spellings of each.
- **Prose figure 1 (`24%` → the population).** Sound in kind and the right replacement: an unverifiable rate about round-1 behaviour became the measured population the reset serves. On the gate's path it is **35 of 270 (13.0%)**; under the dev's instrument 34 of 265 — the one lost row sits in a comment-start hunk.
- **Prose figure 2 (`431,750` → `84,924`).** **Confirmed exactly**: 84,924 side-lines, `enclosingDelimiter` and `inParameterList` compared across main and head at every one, **0** disagreements each.
- **The 「not rare」 sentence is true as written.** The tree at this tip has **115** one-line `.regex(…)` call sites in 65 of 1,018 non-test spec files plus 23 const-bound regex literals; the real corpus carries 22 regex-shaped lines in 16 of 2,231 hunks; none of them precedes a reported T1 row inside its hunk. The zero is about this window. One wording nit, not a defect: the walk crosses the hunk's added lines as well as its context, so 「the leading context the walk crossed」 reads narrower than what is measured.
- **Gates.** `dispatch-gates --commands` → 30 families at head, the same 30 on `origin/main` `43d492ff` and byte-identical to round 4's roster. All 30 run in the head worktree after `pnpm install`: **30 of 30 exit 0**, each code captured into a file before any pipe (`check:pm-dispatch-gates` 914 s, 1,883 cases; `check:pm-widening-tells` exit 0; `check:nul-bytes` exit 0). `eslint --no-inline-config` on the file exit 0; whole tree **6,943 files, 0 errors, 0 warnings**, the population lit by a JSON-formatter run. CI on this head: the seven required contexts `success` or path-filtered `skipped`, as at round 4.
- **Unchanged from round 4 and re-read, not re-measured:** the reduction, the eight legs, the surplus and #18234 probes, the live pair — the code is token-identical to `95a7e8c0`, so those readings carry.

### NOT MEASURED, declared

objectui's mirror · the place limits of `rewritesExistingOpener` / `respellsExistingClosedSetBinding` · history before the boundary · the dev's harness itself (its shape is inferred from the one replay that reproduces its six numbers; a second shape could in principle produce the same six).

⛔ The PR stays a draft and does not enter the queue. Items 1–3 go to the delivering dev; item 5 is the seat's.

Implemented-by: `claude/issue-19099-widening-tell-prior-schema`
Reviewed-by: `session_01HnRAeVTLJevtQ5iCPX6JSm`

**VERDICT: FAIL**


---

os-sam:
## Contract review — round 6, on head `059f5e9aaf`

Head-sha: `059f5e9aafa00e9a019660493ed4003b0163fc19`

Same isolated at-tier reviewer as rounds 4 and 5. Two detached worktrees — head `059f5e9a` (`pnpm install` run BEFORE any gate) and `origin/main` `5a5e710f` (read at 18:02Z; the file's blob on main is still `31cb9de0`, the same control as every round) — both removed afterwards; shared checkout `15f92842` clean before and after and never entered. Lineage: one commit on `0d69cf2e`, API `head.sha` matches, +337/−27 against the merge base `8271c814`, one file. Exit codes captured before any pipe; every zero below carries a lit control.

**PASS.** Round 5's six numbers are corrected to the gate's read path, the nit is taken, the read-path text the dev added on its own is true, the stray control file is nowhere in the committed range, and the change is prose-only at the token level. What this round re-measured and what it carried forward untested are separated below so the landing record is exact.

### RE-MEASURED this round, on this head

- **Prose-only, three ways.** 63 changed lines, **0** non-comment. Token streams of `0d69cf2e` and `059f5e9a` identical (acorn, **28,430** tokens, hash `36712dad6b10` — the same hash as `95a7e8c0`, so rounds 4, 5 and 6 are one token stream). Comment-stripped with blank lines dropped: **198,864 bytes both sides**, same hash (the dev's 197,712 is a different stripper's count of the same fact, stable across rounds). `node --check` 0; `--self-test` exit 0, **525**.
- **The six items, by grep and by reading.** Corpus line `484 rows against 484 rows, 0 differing commits (T2 125, T1 270, T4 89)`, `1,133 head blobs resolved`, 22 unresolved by name ✓ · residual paragraph `270 T1 rows this gate REPORTS`, `2 of those 270 rows (0.7%)`, self-count 「wrong four times」 ✓ · `35 of the 270 T1 rows (13.0%)` in the docblock AND the inline comment at the reset ✓ · sides explained as 2,298 = two per file diff, 1,939 non-empty, 359 the absent side of an added or deleted file ✓ · the regex bullet now says the walk crosses every kind of line the hunk shows before the row, not the leading context alone ✓. Retired figures each grep **0**: `265`, `0.8%`, `722`, `368`, `12.8%`, `34 of the 265`.
- **The corpus, on the gate's read path, at this tip:** `wideningTells(files)` with `headBlobSource` live, one boundary excluded — 435 commits, 1,149 file diffs, **484 = 484, 0 differing** (T2 125 / T1 270 / T4 89), 1,133 blobs resolved / 16 null, **22 unresolved on both legs** (9 factory names, `ruleArrayFilterError` in `ui/component.zod.ts` the largest at 10). ⇒ every number the header now carries is true of this head on the gate's own path, independently of how the dev arrived at it.
- **The unspecified addition — the read-path paragraph — is true, and it is in the right place.** Each claim checks against the code: `wideningRefusal` → `wideningTells(files, { readSource: headBlobSource })` is the gate's path; a per-file `tellsInFile` defaults `licensed` to null and its own docblock says every row then tells (measured: +243, all `migrations/registry.ts`); a null `readSource` makes `localDeclaringForms` resolve against an empty string, so #18702's factories vanish (measured: −5); `ROOT` is `new URL('../..', import.meta.url)`, `gitAt` runs `git -C ROOT`, and `readHeadBlob` catches both attempts and answers null. It sits under the horizon paragraph, which is where a reader taking a corpus reading of this file starts, and it points at the two functions by name. Not owed but worth a line one day: a cross-reference from `readHeadBlob`'s 「stated silence」 sentence back to it.
- **Gates.** `dispatch-gates --commands` → **30 families** at head, the same 30 on `origin/main` `5a5e710f` and byte-identical to round 5's roster. All 30 run in the head worktree, deps installed first: **30 of 30 exit 0**, each code captured into a file before any pipe (`check:pm-dispatch-gates` 887 s, 1,883 cases; `check:pm-widening-tells` exit 0; `check:nul-bytes` exit 0). `eslint --no-inline-config` on the file exit 0; whole tree **6,943 files, 0 errors, 0 warnings** (JSON-formatter run, population lit). CI on this head: the seven required contexts `success` or path-filtered `skipped`, as at rounds 4 and 5. `check-clause2-carriers --pair 19153` before this record: exit 4 on **C6 alone**.
- **The transient control copy is not in the commit.** `git diff --name-status 8271c814..059f5e9a` lists one path; the one new commit touches one path; `git ls-tree -r HEAD` has no `__ctl` entry; `git log 8271c814..HEAD -- scripts/pm/__ctl-widening-tells.mjs` is empty. The claim holds.

### The two judgements the seat asked for

- **The gate incident — handled correctly, and the expensive way is the only sound way.** Six exit-3s from a pre-install sweep are NOT MEASURED, never red. Re-running all 29 short families post-install rather than the six alone, and re-running the long family rather than reasoning about whether its pre-install exit 0 was degraded, is right: a gate can skip a dependency-shaped check and still exit 0, and 「a verifier that silently degrades … is worse than no verifier」 is this repo's own rule. Deciding per family which pre-install zero was trustworthy would have replaced one measurement with thirty inferences. (Round 5's review made the same error and took the same remedy: a pre-install runner stopped after four exit-3s, deps installed, all 30 re-run from scratch.)
- **Independence of the re-derivation — supported, with one specific I could not locate.** Every figure in the round-6 report that I also published matches mine to the unit, and the report adds only derived or run-specific values (98.6%, the two long-family timings, the six-family incident). The one non-transcribable specific it offers — that `headBlobSource` read **336,382 bytes** from a known blob inside the worktree — **is located, and it is a live reading.** No blob reachable from the head or the merge base has 336,382 BYTES (20,633 blobs scanned, the 375,352-byte control copy found on the same pass), but exactly one has a UTF-16 length of 336,382: `5468c395`, which is `packages/spec/src/ui/view.zod.ts` at the merge base, at this head and on `origin/main` — 339,808 bytes — the file holding three of the five recovered T1 rows. So the figure is `headBlobSource(...).length` on a real blob, which only a live call inside a worktree returns; the word 「bytes」 in the report and the body is loose (code units, not bytes) and worth one word's correction in the body, nothing more. The file's correctness does not rest on it: the header's numbers were re-measured here on the gate's path at this tip.

### CARRIED FORWARD UNTESTED, because the code is token-identical to `95a7e8c0`

The reduction census (dropped identifiers at 0, the three untouched functions at main's counts) · the eight ablation legs and the two reviewer legs · the surplus, #18234-sibling, dark and live-pair probes · the trigger classifier's 0-disagreement self-check · the 84,924 side-line differential (0 disagreements for both firing-only readers). All measured at round 4 on `95a7e8c0`; the token stream has not changed since, so those readings are readings of this head too. The corpus identity was NOT carried — it was re-run above.

### NOT MEASURED, declared

objectui's mirror · the place limits of `rewritesExistingOpener` / `respellsExistingClosedSetBinding` (four rounds, probed by nobody) · history before the boundary · the dev's harness itself, beyond what its published numbers allow.

⇒ Landing: non-governed, `Clause-②: no` correct, `skip-changeset` correct, the diff one file under `scripts/pm/**`. Nothing is owed on the code or the header. The landing is the seat's act.

Implemented-by: `claude/issue-19099-widening-tell-prior-schema`
Reviewed-by: `session_01HnRAeVTLJevtQ5iCPX6JSm`

**VERDICT: PASS**


---

